PR #1045, opened 2017-02-09
Yyuen/root cause analysis link

justYves:
### What's new: 
- Hooked up the Anomalies result `root cause analysis` button
- Configure the Root Cause Analysis page to use the Hash Params and Hash Services 
- Root Cause Analysis now parses the hash correctly 
- Added custom end point (temporary) @cyenjung 

### Gif:
mr3wd55ikd

### How this was tested: 
- Manually tested locally with prod data

### To do (in follow up PR): 
- metric name, granularity, dimension and filters are not prepopulated after a redirect. (no regression for default behavior/clicks)
- d3/c3 details panel not render correctly after a redirect
- Unify Hash Params for Root Cause Anomalies page so that it doesn't use both `params.metric` and `params.metricId`

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/DataResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/DataResource.java
@@ -97,6 +97,14 @@ public class DataResource {
     this.anomaliesResoure = new AnomaliesResource();
   }
 
+//------------- endpoints to metric config -------------
+  @GET
+  @Path("data/metricId")
+  public List<MetricConfigDTO> getMetricsByName(@QueryParam("name") String name) {
+    List<MetricConfigDTO> metricConfigDTOs = metricConfigDAO.findByMetricName(name);
+    return metricConfigDTOs;
+  }
+
   //------------- endpoints to fetch summary -------------
   @GET
   @Path("summary/metrics")
--- a/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
@@ -157,9 +157,7 @@ DataService.prototype = {
         name,
       };
 
-      return this.getDataAsynchronous(url, data).then(([data]) => {
-        return data.id;
-      });
+      return this.getDataAsynchronous(url, data).then(([data]) => data.id);
     },
 
     fetchTimeseriesCompare: function (metricId, currentStart, currentEnd, baselineStart, baselineEnd,
--- a/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
@@ -154,7 +154,7 @@ DataService.prototype = {
     fetchMetricId(name) {
       const url = "/data/data/metricId";
       const data = {
-        name,
+        name
       };
 
       return this.getDataAsynchronous(url, data).then(([data]) => data.id);
--- a/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
@@ -148,7 +148,7 @@ DataService.prototype = {
     },
 
     fetchTimeseriesCompare: function (metricId, currentStart, currentEnd, baselineStart, baselineEnd,
-        dimension, filters, granularity) {
+        dimension = "ALL", filters = {}, granularity = "DAYS") {
       var url = "/timeseries/compare/" + metricId + "/" + currentStart + "/" + currentEnd + "/"
           + baselineStart + "/" + baselineEnd + "?dimension=" + dimension + "&filters="
           + JSON.stringify(filters) + "&granularity=" + granularity;
@@ -158,7 +158,7 @@ DataService.prototype = {
     },
 
   fetchHeatmapData: function (metricId, currentStart, currentEnd, baselineStart, baselineEnd,
-      filters) {
+      filters = {}) {
     var url = "/data/heatmap/" + metricId + "/" + currentStart + "/" + currentEnd + "/"
         + baselineStart + "/" + baselineEnd + "?filters=" + JSON.stringify(filters);
     console.log("heatmap data fetch URL ----> ");
--- a/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
@@ -30,7 +30,7 @@ DataService.prototype = {
       spinner.spin(target);
 
       console.log("request url:", url)
-      $.ajax({
+      return $.ajax({
         url: url,
         data: data,
         type: 'get',
@@ -43,7 +43,11 @@ DataService.prototype = {
         }
       }).done(function(data) {
         spinner.stop();
-        callback(data);
+        if (callback) {
+          callback(data);
+        } else {
+          return data;
+        }
       });
     },
 
@@ -145,6 +149,17 @@ DataService.prototype = {
     fetchFiltersForMetric : function(metricId) {
       var url = "/data/autocomplete/filters/metric/" + metricId;
       return this.getDataSynchronous(url);
+    },
+
+    fetchMetricId(name) {
+      const url = "/data/data/metricId";
+      const data = {
+        name,
+      };
+
+      return this.getDataAsynchronous(url, data).then(([data]) => {
+        return data.id;
+      });
     },
 
     fetchTimeseriesCompare: function (metricId, currentStart, currentEnd, baselineStart, baselineEnd,
--- a/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashParams.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashParams.js
@@ -34,6 +34,7 @@ function HashParams() {
   this.DASHBOARD_ANOMALY_SUMMARY_CONTROLLER = "anomalySummary";
   this.DASHBOARD_METRIC_SUMMARY_CONTROLLER = "metricSummary";
   this.DASHBOARD_WOW_SUMMARY_CONTROLLER = "wowSummary";
+
 }
 
 
@@ -104,5 +105,24 @@ HashParams.prototype = {
 
       console.log('hash Params init');
       console.log(this.controllerNameToParamNamesMap);
-    }
+    },
+
+    isSame(key, currentValue, newValue) {
+      switch(key){
+        case this.ANALYSIS_CURRENT_START:
+        case this.ANALYSIS_CURRENT_END :
+        case this.ANALYSIS_BASELINE_START :
+        case this.ANALYSIS_BASELINE_END :
+        case this.ANOMALIES_START_DATE :
+        case this.ANOMALIES_END_DATE:
+          return currentValue.isSame(newValue, 'day');
+          break;
+        case this.ANALYSIS_FILTERS:
+          return JSON.stringify(currentValue) == newValue;
+          break;
+        default:
+          return currentValue == newValue;
+      }
+    },
+
 };
--- a/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashParams.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashParams.js
@@ -14,6 +14,15 @@ function HashParams() {
   this.ANOMALIES_DASHBOARD_ID = 'dashboardId';
   this.ANOMALIES_ANOMALY_IDS = 'anomalyIds';
 
+  this.ANALYSIS_METRIC_ID = 'metricId';
+  this.ANALYSIS_CURRENT_START= 'currentStart';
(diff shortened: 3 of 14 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/DataResource.java
```diff
@@ -97,6 +97,14 @@ public DataResource() {
     this.anomaliesResoure = new AnomaliesResource();
   }
 
+//------------- endpoints to metric config -------------
+  @GET
```
justYves: Kindly provided by @cyenjung. Thank you again! 🎉

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
```diff
@@ -147,8 +151,19 @@ DataService.prototype = {
       return this.getDataSynchronous(url);
     },
 
+    fetchMetricId(name) {
+      const url = "/data/data/metricId";
+      const data = {
+        name,
+      };
+
+      return this.getDataAsynchronous(url, data).then(([data]) => {
```
justinpark: this can be shorten`return this.getDataAsynchronous(url, data).then(([data]) => data.id);`

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnalysisView.js
```diff
@@ -8,7 +8,8 @@ function AnalysisView(analysisModel) {
 }
 
 AnalysisView.prototype = {
-  init: function () {
+  init({metricId=""}) {
```
justinpark: Single quote please

justinpark: In this case, `init` must be called with `{}` and will throw an error in `init()` case.
I suggest to use like
```
init(params = {}) {
  const {metricId} = params;
  this.viewParams.metricId = metricId || '';
}
```

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnalysisView.js
```diff
@@ -39,18 +40,26 @@ AnalysisView.prototype = {
...
-      self.viewParams['metric'] = {id: metricId, alias: metricAlias, allowClear:true, name:metricName};
+    analysisMetricSelect.on("change", (e) => {
+      const selectedElement = $(e.currentTarget);
+      const selectedData = selectedElement.select2("data");
+      let metricId;
+      let metricAlias;
+      let metricName;
+
+      if (selectedData.length) {
+        metricId = selectedData.map(function (e) {return e.id})[0];
```
justinpark: It looks iterating the entire selectedData array but it only needs the first one.
Let's pick the first index data and then spread the values.
```
const { id, text, name } = selectedData[0];
metricId = id;
metricAlias = text;
metricName = name;
```

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnalysisView.js
```diff
@@ -39,18 +40,26 @@ AnalysisView.prototype = {
...
-    analysisMetricSelect.on("change", function(e) {
-      var selectedElement = $(e.currentTarget);
-      var selectedData = selectedElement.select2("data");
-      var metricId = selectedData.map(function (e) {return e.id})[0];
-      var metricAlias = selectedData.map(function (e) {return e.text})[0];
-      var metricName = selectedData.map(function (e) {return e.name})[0];
-      self.viewParams['metric'] = {id: metricId, alias: metricAlias, allowClear:true, name:metricName};
+    analysisMetricSelect.on("change", (e) => {
+      const selectedElement = $(e.currentTarget);
+      const selectedData = selectedElement.select2("data");
```
justinpark: Single quote please too :)

justYves: Yes sir! Sorry, it was mostly copy pasta :) I'll make sure to refactor as I go

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/models/DimensionTreeMapModel.js
```diff
@@ -26,6 +26,11 @@ DimensionTreeMapModel.prototype = {
         this.metricId = params.metric.id;
         this.metricName = params.metric.name;
       }
+
+      if (params.metricId) {
```
justYves: I think we can run into the case where both params.metricId and params.metric are undefined. so `params.metric.id` will throw a TypeError.

Ideally, I'd like to only keep `params.metricId` and deprecate `params.metric`. This will require a deeper dive into the code. I'll keep this section of code as-is, and will update the 'to-dos'

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/DataService.js
```diff
@@ -147,8 +151,17 @@ DataService.prototype = {
       return this.getDataSynchronous(url);
     },
 
+    fetchMetricId(name) {
+      const url = "/data/data/metricId";
+      const data = {
+        name,
```
theseyi: We don't use trailing commas as a convention

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashParams.js
```diff
@@ -81,9 +91,38 @@ HashParams.prototype = {
...
+        case this.ANALYSIS_CURRENT_END :
+        case this.ANALYSIS_BASELINE_START :
+        case this.ANALYSIS_BASELINE_END :
+        case this.ANOMALIES_START_DATE :
+        case this.ANOMALIES_END_DATE:
+          return currentValue.isSame(newValue, 'day');
+          break;
+        case this.ANALYSIS_FILTERS:
+          return JSON.stringify(currentValue) == newValue;
+          break;
```
theseyi: You have a` return` statement above, the `break` is unreachable code and should be removed

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashParams.js
```diff
@@ -81,9 +91,38 @@ HashParams.prototype = {
...
+        case this.ANALYSIS_CURRENT_START:
+        case this.ANALYSIS_CURRENT_END :
+        case this.ANALYSIS_BASELINE_START :
+        case this.ANALYSIS_BASELINE_END :
+        case this.ANOMALIES_START_DATE :
+        case this.ANOMALIES_END_DATE:
+          return currentValue.isSame(newValue, 'day');
+          break;
+        case this.ANALYSIS_FILTERS:
+          return JSON.stringify(currentValue) == newValue;
```
theseyi: Should this be an identity `===` comparison? This is preferred as a default except you're intentionally seeking the implicit type conversion with `==`, in that case it is better to be explicit with a type conversion operation e.g. using `String()` or concatenating with an empty string literal `'' + ...`

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashService.js
```diff
@@ -42,11 +42,19 @@ HashService.prototype = {
...
+      case HASH_PARAMS.ANALYSIS_CURRENT_END:
+      case HASH_PARAMS.ANALYSIS_BASELINE_START:
+      case HASH_PARAMS.ANALYSIS_BASELINE_END:
         value = moment(value);
         break;
       case HASH_PARAMS.ANOMALIES_PAGE_NUMBER:
         value = Number(value);
         break;
+      case HASH_PARAMS.ANALYSIS_FILTERS:
+        if (typeof value === "string") {
```
theseyi: Hmm.. this may not actually achieve what you want... `JSON.parse("")` will throw. Personally, if i'm uncertain of the input into `JSON.parse` I would encase in a try-catch block then subsequently check that value is valid

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashService.js
```diff
@@ -57,7 +65,7 @@ HashService.prototype = {
     console.log('hash service.update');
     console.log(paramsToUpdate)
     for (var key in paramsToUpdate) {
-      this.params[key] = paramsToUpdate[key];
```
theseyi: `Object#hasOwnProperty()` should be used here if you're sticking with `for..in`, or `Object#keys` on paramsToUpdate if you want to iterate with  `Array#forEach`

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashService.js
```diff
@@ -79,11 +87,11 @@ HashService.prototype = {
       }
       var defaultValue = paramNamesToDefaultValuesMap[paramName];
       var value = this.get(paramName);
-      if (value == undefined && defaultValue != undefined) { //if default value is present, use that
+      if (!value && defaultValue) { //if default value is present, use that
```
theseyi: You can condense these lines 89-95 into one succinct `default` operation:

```javascript
var value = this.get(paramName) || this.set(paramName, defaultValue);
```

or  even 88 -95

```javascript
var value = this.get(paramName) || this.set(paramName, paramNamesToDefaultValuesMap[paramName]);
```

if the var `defaultValue` is not used anywhere else. `Ember.set` returns the passed in value

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/models/AnomalyResultModel.js
```diff
@@ -149,5 +149,11 @@ AnomalyResultModel.prototype = {
     default:
       return feedbackType;
     }
-  }
+  },
+  getMetricIdFromName(name) {
+    if (!name) {
```
theseyi:
````javascript 
return name && dataService.fetchMetricId(name);

File: thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnalysisView.js [hunk 14]
```diff
@@ -39,18 +41,27 @@ AnalysisView.prototype = {
...
     });
 
-    analysisMetricSelect.on("change", function(e) {
-      var selectedElement = $(e.currentTarget);
-      var selectedData = selectedElement.select2("data");
-      var metricId = selectedData.map(function (e) {return e.id})[0];
-      var metricAlias = selectedData.map(function (e) {return e.text})[0];
-      var metricName = selectedData.map(function (e) {return e.name})[0];
-      self.viewParams['metric'] = {id: metricId, alias: metricAlias, allowClear:true, name:metricName};
+    analysisMetricSelect.on('change', (e) => {
```
theseyi: You don't need parenthesis around `e`

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnalysisView.js
```diff
@@ -39,18 +41,27 @@ AnalysisView.prototype = {
...
+      const selectedElement = $(e.currentTarget);
```
theseyi: We generally by convention prefix $ elements with $, so `$selectedElement`, so it's clear to the reader later in the context that this is $ element as opposed to a DOM element

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashParams.js
justYves: good catch!

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashService.js
justYves: so the initial intention was that since this property can either be set from the controller as an object `{}` , or  taken from the hash params as a stringified object `"{}"`, I would parse it only in the latter case, the default value being `"{}"`
But I agree that a try catch might be cleaner.

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashService.js
justYves: Thanks! Will refactor. Wish it was an ember app though :p

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnalysisView.js
justYves: Airbnb's guideline is to omit parenthesis when braces are also omitted, otherwise keep them for clarity https://github.com/airbnb/javascript#arrows--one-arg-parens
Are we doing this differently ?

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/HashService.js
theseyi: Seen, in that case since you have complete control over this input value, maybe a comment here indicating that, just incase another dev works on this codebase and might think this is a bug on first pass

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/DataResource.java
cyenjung: You're welcome :-)

PR #1048, opened 2017-02-10
ThirdEye: Multi metrics anomaly detection

apucher:
* add functionality for multi-metric retrieval

* implement sample anomaly function with two metrics (RatioOutlierFunction)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/RawAnomalyResultDTO.java
```diff
@@ -22,8 +22,9 @@ public void setFunction(AnomalyFunctionDTO function) {
     this.function = function;
   }
 
+  // TODO: rename to getMetric and update the column name in DB?
   public String getMetric() {
-    return function.getMetric();
+    return function.getTopicMetric();
```
puneetjaiswal: just use get/set metric instead of "topicMetric" and have a javadoc explaining what metric is.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/function/RatioOutlierFunction.java
```diff
@@ -0,0 +1,202 @@
...
+      min = Double.valueOf(props.getProperty(MIN_VAL));
+    }
+
+    Double max = null;
+    if (props.containsKey(MAX_VAL)) {
+      max = Double.valueOf(props.getProperty(MAX_VAL));
+    }
+
+    // Metric
+    String topicMetric = getSpec().getTopicMetric();
```
puneetjaiswal: just use getSpec().getMetric().. I dont see a reason to overrider plain simple getters/setters.

PR #1049, opened 2017-02-11
Auto-fix segment consumption if idealstate update fails during LLC segment completion

mcvsubbu:
When a segment completes, we update zk with two changes:
1. Update proprtystore with a new LLC segment
2. Update idealstate to mark old segment in ONLINE and add a new segment in CONSUMING
These two updates cannot be done in a transaction. It is possible that the controller
fails after step 1 but before step 2. In that case, we had called the code to fix
things on a controller getting leadership.

It is also possible that the idealstate update fails due to concurrent updates to
the idealstate znode (e.g. retention manager, deletes by operator, etc.). In these
cases, we have a retry, but no guarantees.

With this change, we call the same logic that we call on leadership change.

Also changed the logic that detects and fixes the condition to use priority
queues to make the code more readable.

Existing unit tests cover this case already

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -23,7 +23,7 @@ import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -46,6 +46,7 @@ import org.apache.helix.store.zk.ZkHelixPropertyStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import com.google.common.base.Function;
+import com.google.common.collect.MinMaxPriorityQueue;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.linkedin.pinot.common.config.AbstractTableConfig;
 import com.linkedin.pinot.common.config.TableNameBuilder;
@@ -549,6 +550,7 @@ public class PinotLLCRealtimeSegmentManager {
     try {
       lock.lock();
       updateHelixIdealState(realtimeTableName, newInstances, committingSegmentNameStr, newSegmentNameStr);
+      LOGGER.info("Changed {} to ONLINE and created {} in CONSUMING", committingSegmentNameStr, newSegmentNameStr);
     } finally {
       lock.unlock();
     }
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -620,69 +622,76 @@ private void completeCommittingSegments() {
...
+    Comparator<LLCSegmentName> comparator = new Comparator<LLCSegmentName>() {
+      @Override
+      public int compare(LLCSegmentName o1, LLCSegmentName o2) {
+        return o2.compareTo(o1);
       }
-      if (curPartition < 0) {
-        break;
+    };
+
+    Map<Integer, MinMaxPriorityQueue<LLCSegmentName>> partitionToLatestSegments = new HashMap<>(16);
```
jfim: Hardcoded `16`?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -620,69 +622,76 @@ private void completeCommittingSegments() {
...
       return;
     }
 
-    Collections.sort(segmentNames, Collections.reverseOrder());
+    completeCommittingSegments(realtimeTableName, segmentIds);
+  }
 
-    int curPartition = segmentNames.get(0).getPartitionId();  // Current kafka partition we are working on.
-    final int nSegments = segmentNames.size();
+  private void completeCommittingSegments(String realtimeTableName, Map<Integer, MinMaxPriorityQueue<LLCSegmentName>> partitionToLatestSegments) {
```
jfim: Avoid overloading methods

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -620,69 +622,76 @@ private void completeCommittingSegments() {
...
+        LOGGER.info("{}: Assigning segment {} to {}", realtimeTableName, segmentId, newInstances);
+        // TODO Re-write num-partitions in metadata if needed.
+        // If there was a prev segment in the same partition, then we need to fix it to be ONLINE.
+        LLCSegmentName prevSegmentName = entry.getValue().pollLast();
+        String prevSegmentNameStr = null;
+        if (prevSegmentName != null) {
+          prevSegmentNameStr = prevSegmentName.getSegmentName();
         }
-        curPartition--;
+        updateHelixIdealState(realtimeTableName, newInstances, prevSegmentNameStr, segmentId);
```
jfim: Can we rename `updateHelixIdealState` to better describe what it does? I assume it's doing something close to `addNewSegmentToIdealState`

mcvsubbu: It is updating the ideal state. We prefer to do that in one zk write operation, so we set up whatever needs to be done beforehand, and update the idealstate (with back-off) in this method.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: Not sure why you bring this up, I think we use (and the standard libraries do as well) this pattern all over the place.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: Removed.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: Changed the private method to be completeCommitingSegmentsInternal() as per our conversation

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: changed it to updateIdealState() as per discussion and added comments as per discussion

PR #1050, opened 2017-02-11
Performance improvements for no-dictionary group key generator.

mayankshriv:
1. Separated out implementations of no-dictionary group key generators
for single column and multiple columns.

2. For group-by multiple columns, create on-the-fly dictionary and use
array of dictionaryId's as key , as opposed to string concatinated values.


mayankshriv:
@Jackie-Jiang Added unit test for NoDictionarySingleColumnGroupKeygenerator.

PR #1057, opened 2017-02-13
[PINOT-4491] Re-throw exception when we fail to get a schema

mcvsubbu:
We were swallowing the exception when the schema fetch fails, and were running into another
exception as we process the generic record.

Added more information in the logs.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaAvroMessageDecoder.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaAvroMessageDecoder.java
@@ -95,20 +95,13 @@ public class KafkaAvroMessageDecoder implements KafkaMessageDecoder {
     if (md5ToAvroSchemaMap.containsKey(md5String)) {
       schema = md5ToAvroSchemaMap.get(md5String);
     } else {
-      URL url;
       final String schemaUri = schemaRegistryBaseUrl + "/id=" + md5String;
       try {
-        url = new URL(schemaUri);
-      } catch (Exception e) {
-        LOGGER.error("Could not parse URI {}", schemaUri, e);
-        throw new RuntimeException(e);
-      }
-      try {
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaAvroMessageDecoder.java
```diff
@@ -101,7 +99,8 @@ public GenericRow decode(byte[] payload, int offset, int length, GenericRow dest
         md5ToAvroSchemaMap.put(md5String, schema);
       } catch (Exception e) {
         schema = defaultAvroSchema;
-        LOGGER.error("error fetching schema from md5 String", e);
+        LOGGER.error("error fetching schema from md5 String from url {}", schemaRegistryBaseUrl, e);
```
jfim: It'd be good to have the MD5 as well as the full URL that is used to fetch the schema. Can this be changed to output something like `Error fetching schema from URL http://.../id=1234`?

Also, since we throw an exception, `schema = defaultAvroSchema` does nothing, as `schema` is a local.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaAvroMessageDecoder.java
```diff
@@ -111,7 +110,7 @@ public GenericRow decode(byte[] payload, int offset, int length, GenericRow dest
               length - HEADER_LENGTH, null));
       return avroRecordConvetrer.transform(avroRecord, schema, destination);
     } catch (IOException e) {
-      LOGGER.error("Caught exception while reading message", e);
+      LOGGER.error("Caught exception while reading message using schema {}", schema.getName(), e);
```
jfim: It doesn't look like it's possible for `schema` to be `null` at this point in time, but can you guard against it in case it happens in a code change/refactoring?

PR #1065, opened 2017-02-15
Calculate percentage of replicas which are up and emit a statistic

rven1:


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -181,6 +181,7 @@ public void runSegmentMetrics() {
       ExternalView externalView = helixAdmin.getResourceExternalView(helixClusterName, tableName);
 
       final int nReplicasIdeal = Integer.parseInt(idealState.getReplicas());
```
mcvsubbu: we don't need this variable anymore, right?

rven1: We need the variable for the initial value

PR #1069, opened 2017-02-15
Introduce multimetrics to new anomaly function framework

cyenjung:
Introduce multi-metrics to new anomaly function framework. Currently, the metrics besides the main metric are used for providing additional information for performing transformation and detection on the main metric.

Tested with new unit test.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/context/AnomalyDetectionContext.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/context/AnomalyDetectionContext.java
@@ -2,7 +2,9 @@ package com.linkedin.thirdeye.anomalydetection.context;
 
 import com.linkedin.thirdeye.anomalydetection.function.AnomalyDetectionFunction;
 import com.linkedin.thirdeye.anomalydetection.model.prediction.PredictionModel;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * The context for performing an anomaly detection on the sets of time series from the same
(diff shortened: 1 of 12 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/context/AnomalyDetectionContext.java
```diff
@@ -16,16 +18,16 @@
...
+  private Map<String, TimeSeries> current = new HashMap<>();
+  private Map<String, List<TimeSeries>> baselines = new HashMap<>();
 
   //TODO: Add DAO for accessing historical anomalies or scaling factor
 
   // The followings are intermediate results and are appended during anomaly detection
-  private TimeSeries transformedCurrent;
-  private List<TimeSeries> transformedBaselines;
+  private Map<String, TimeSeries> transformedCurrent = new HashMap<>();
+  private Map<String, List<TimeSeries>> transformedBaselines = new HashMap<>();
```
bl44: Is it possible to treat baselines as named <string, timeseries>? your current choice is a direct extension from previous case where you have current  + baselines (list). But if you have current already as a list, there isnot strong reason you treat them differently. you just need the hash-name to impose the relationship between current and baselines, this relationship can be achieved with a real tag (or how you construct the transformer).

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/context/AnomalyDetectionContext.java
```diff
@@ -44,29 +46,29 @@ public void setTimeSeriesKey(TimeSeriesKey key) {
   /**
    * Returns the current (observed) time series, which is not transformed.
    */
-  public TimeSeries getCurrent() {
-    return current;
+  public TimeSeries getCurrent(String metricName) {
```
bl44: related to previous comments in AnomalyContext. you need to use metricName anyway, so under this new situation, its better to treat current and baseline equally. Since both are modeled as timeseries right now. It should be fine to have 'un-equal' length for baseline and current. In general, as long as we remove the seasonality, you do not need the baseline and current always aligned in length. Usually the transformer do not mix current and baseline.

PR #1072, opened 2017-02-17
Switch to first come first served scheduled query executor

antumbde:
This change is mainly about switching to the scheduled query executor
to allow us to use custom query scheduling algorithms. Current implementation
will use FCFS as the only supported algorithm. This commit has following
important changes:

1. Switch to scheduled query executor that processes queries on different
executor service freeing up netty threads quickly. Freeing up netty is less beneficial
right now but we can potentially lower threads and support async queries from broker in future.

2. Query worker threads are moved from TableDataManager control to query scheduler. This
means we have single query worker pool instead of per table pool - much less threads -
which will impact existing multitenant clusters.
3. Number of threads in each thread pool are configurable.

Performance:
There is no performance impact across multiple use cases.


pinot-common/src/main/java/com/linkedin/pinot/common/query/QueryExecutor.java
```diff
@@ -24,13 +25,13 @@
 
 
 public interface QueryExecutor {
-  public void init(Configuration queryExecutorConfig, DataManager dataManager, ServerMetrics serverMetrics) throws ConfigurationException;
+  void init(Configuration queryExecutorConfig, DataManager dataManager, ServerMetrics serverMetrics) throws ConfigurationException;
```
mcvsubbu: pls add some javadocs with what these do and how the args are used

pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QueryScheduler.java
```diff
@@ -73,18 +81,43 @@ public QueryScheduler(@Nullable QueryExecutor queryExecutor) {
...
+    Preconditions.checkNotNull(queryExecutor);
+
     numQueryRunnerThreads = schedulerConfig.getInt(QUERY_RUNNER_CONFIG_KEY, DEFAULT_QUERY_RUNNER_THREADS);
     numQueryWorkerThreads = schedulerConfig.getInt(QUERY_WORKER_CONFIG_KEY, DEFAULT_QUERY_WORKER_THREADS);
-    queryRunners = MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(numQueryRunnerThreads));
-    queryWorkers = MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(numQueryWorkerThreads));
+    LOGGER.info("Initializing with {} query runner threads and {} worker threads", numQueryRunnerThreads,
+        numQueryWorkerThreads);
+    ThreadFactory queryRunnerFactory = new ThreadFactoryBuilder().setDaemon(false)
+        .setPriority(7)
```
mcvsubbu: pls use static final int and explain what 7 means and why 7 ?

pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QuerySchedulerFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.base.Preconditions;
+import com.linkedin.pinot.common.query.QueryExecutor;
+import javax.annotation.Nonnull;
+import org.apache.commons.configuration.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class QuerySchedulerFactory {
+  private static final String DEFAULT_QUERY_SCHEDULER_ALGORITHM = "fcfs";
```
mcvsubbu: static string for fcfs will be good

pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QuerySchedulerFactory.java
```diff
@@ -0,0 +1,46 @@
...
+  private static Logger LOGGER = LoggerFactory.getLogger(QuerySchedulerFactory.class);
+
+  public static @Nonnull  QueryScheduler create(@Nonnull Configuration schedulerConfig,
+      @Nonnull QueryExecutor queryExecutor) {
+    Preconditions.checkNotNull(schedulerConfig);
+    Preconditions.checkNotNull(queryExecutor);
+
+    String algorithm = schedulerConfig.getString("algorithm", DEFAULT_QUERY_SCHEDULER_ALGORITHM).toLowerCase();
```
mcvsubbu: pls turn "algorithm" into static as well

pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QuerySchedulerFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+    if (algorithm.equals("fcfs")) {
+      LOGGER.info("Using FCFS query scheduler");
+      return new FCFSQueryScheduler(schedulerConfig, queryExecutor);
+    }
```
mcvsubbu: adda a warning if scheduler name is not recognized

mcvsubbu: it may be useful to check if a class can be loaded here (a class with name of algoritthm)

antumbde:
Addressed all comments

PR #1073, opened 2017-02-17
Handle case where there are no replicas in ideal state

rven1:


pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/SegmentStatusCheckerTest.java
```diff
@@ -367,4 +367,47 @@ public void missingEVPartitionPushTest() throws Exception {
     segmentStatusChecker.stop();
   }
 
+  @Test
+  public void noReplicas() throws Exception {
```
jfim: Rename test to `testZeroReplicas`

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -242,8 +242,10 @@ public void runSegmentMetrics() {
       // Synchronization provided by Controller Gauge to make sure that only one thread updates the gauge
       _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.NUMBER_OF_REPLICAS,
           nReplicasExternal);
-      _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.PERCENT_OF_REPLICAS,
-          nReplicasExternal*100/nReplicasIdealMax);
+      if (nReplicasIdealMax != 0) {
+        _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.PERCENT_OF_REPLICAS,
```
antumbde: We won't emit a value for nReplicasIdealMax = 0. Seems there should be one. What happens if we've alerts on this and no value is emitted for a while ?

rven1: We have set it to 0 in setStatusToDefault

antumbde: that happens only on startup (or for non-leaders). what about on a continuous basis for leader ?

PR #1077, opened 2017-02-20
In Pinot Broker, use a separate SPECTATOR to process non-instance messages.

Jackie-Jiang:
Fix the issue that broker keep processing external view changes and block other listeners.
Use spectator Helix manager to manage non-instance messages.
    
Tested in a cluster with about 1,000 tables. With the change, messages got processed immediately, compare to other brokers that take several minutes to get the message.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/HelixBrokerStarter.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/HelixBrokerStarter.java
@@ -22,6 +22,7 @@ import com.linkedin.pinot.common.metrics.BrokerMeter;
 import com.linkedin.pinot.common.utils.CommonConstants;
 import com.linkedin.pinot.common.utils.ControllerTenantNameBuilder;
 import com.linkedin.pinot.common.utils.NetUtil;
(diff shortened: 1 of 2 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -122,18 +122,21 @@ public HelixBrokerStarter(String helixClusterName, String zkServer, Configuratio
...
 
-
     // Register the service status handler
     ServiceStatus.setServiceStatusCallback(
         new ServiceStatus.IdealStateAndExternalViewMatchServiceStatusCallback(_helixAdmin, helixClusterName, brokerId));
 
     ClusterChangeMediator clusterChangeMediator = new ClusterChangeMediator(_helixExternalViewBasedRouting,
         _brokerServerBuilder.getBrokerMetrics());
-    _helixManager.addExternalViewChangeListener(clusterChangeMediator);
     _helixManager.addInstanceConfigChangeListener(clusterChangeMediator);
```
kishoreg: InstanceConfig and LiveInstance change listener must also be registered to spectatorManager

PR #1078, opened 2017-02-21
Fix the bug where trailing empty strings in group key gets ignored

Jackie-Jiang:


kishoreg:
Can we add a test case.

pinot-core/src/test/java/com/linkedin/pinot/query/aggregation/groupby/AggregationGroupByTrimmingServiceTest.java
```diff
@@ -0,0 +1,94 @@
...
+  private List<String> _groups;
+  private AggregationGroupByTrimmingService _serverTrimmingService;
+  private AggregationGroupByTrimmingService _brokerTrimmingService;
+
+  @BeforeClass
+  public void setUp() {
+    Set<String> groupSet = new HashSet<>(NUM_GROUPS);
+    while (groupSet.size() < NUM_GROUPS) {
+      groupSet.add(
+          RandomStringUtils.randomAlphanumeric(RANDOM.nextInt(10)) + '\t' + RandomStringUtils.randomAlphanumeric(
```
mayankshriv: Could you also explicitly add groups with '\t'?
Also use random instead of randomAlphanumeric.

Jackie-Jiang: Added both.

PR #1079, opened 2017-02-21
Adding event counters as metric reported through jmx

puneetjaiswal:
added dbCallCounter, detectionTaskCounter, detectionTaskSuccessCounter.


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyApplication.java
```diff
@@ -37,11 +38,19 @@
   private AutoLoadPinotMetricsService autoLoadPinotMetricsService = null;
   private DataCompletenessScheduler dataCompletenessScheduler = null;
 
+  public static final Counter detectionTaskCounter =
```
cyenjung: Should we capital the name of "public static final" variables?

puneetjaiswal: conventionally yes.

PR #1080, opened 2017-02-21
Compute maximum number of replicas from the zookeeper state instead of using value of Replicas from zookeeper, which may be stale.

rven1:


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -237,15 +236,16 @@ public void runSegmentMetrics() {
...
       }
       // Synchronization provided by Controller Gauge to make sure that only one thread updates the gauge
       _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.NUMBER_OF_REPLICAS,
           nReplicasExternal);
-      if (nReplicasIdealMax != 0) {
-        _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.PERCENT_OF_REPLICAS,
-            nReplicasExternal * 100 / nReplicasIdealMax);
-      }
+      _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.PERCENT_OF_REPLICAS,
+          (nReplicasIdealMax > 0) ? nReplicasExternal * 100 / nReplicasIdealMax : 0);
```
mcvsubbu: should be 100 if there are no segments in idealstate

PR #1081, opened 2017-02-22
Decoupling sequence Id and Postfix for segment naming

jenniferdai:


pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
```diff
@@ -285,6 +287,14 @@ public void setTimeColumnName(String timeColumnName) {
     _segmentTimeColumnName = timeColumnName;
   }
 
+  public int getSequenceId() {
+    return _sequenceId;
+  }
+
+  public void setSequenceId(int sequenceId) {
```
mcvsubbu: please add a javadoc here saying that this method should be used instead of setPostFix if it is the sequece number that is to be added. thanks. also add a javadoc in setPostFix to use setSequenceNumber() instead, if sequence number is needed at the end of segment name

pinot-core/src/main/java/com/linkedin/pinot/core/segment/DefaultSegmentNameGenerator.java
```diff
@@ -23,7 +23,13 @@
   private final String _segmentName;
   private final String _timeColumnName;
   private final String _tableName;
+  /*
+    The postfix is used if segments have different granularities while the sequenceId is used for numbering segments.
```
mcvsubbu: remove this comment body entirely.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/SegmentNameBuilder.java
```diff
@@ -16,13 +16,19 @@
 package com.linkedin.pinot.common.utils;
 
 public class SegmentNameBuilder {
```
mcvsubbu: Please fold in the methods of this class as (protected) methods in DefaultSegmentNameGenerator and remove this class altogether. There are no users of these methiods

pinot-core/src/main/java/com/linkedin/pinot/core/segment/DefaultSegmentNameGenerator.java
```diff
@@ -41,11 +48,12 @@ public DefaultSegmentNameGenerator(final String segmentName) {
...
-  public DefaultSegmentNameGenerator(String timeColumnName, String tableName, String segmentNamePostfix) {
+  public DefaultSegmentNameGenerator(String timeColumnName, String tableName, String segmentNamePostfix, int sequenceId) {
     _timeColumnName = timeColumnName;
     _tableName = tableName;
     _segmentNamePostfix = segmentNamePostfix;
     _segmentName = null;
+    _sequenceId = sequenceId;
   }
 
   @Override
```
mcvsubbu: Add a javadoc here giving examples of what the segment name would look like with various combinations.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/DefaultSegmentNameGenerator.java
```diff
@@ -56,6 +51,23 @@ public DefaultSegmentNameGenerator(String timeColumnName, String tableName, Stri
...
+   * The sequenceId is used for numbering segments while the postfix is used to append a string to the segment name.
+   *
+   * Some examples:
+   *
+   * If the time column name exists, _segmentNamePostfix = "postfix", and _sequenceId = 1, the segment name would be
+   * tableName_minDate_maxDate_postfix_1
+   *
+   * If there is no time column, _segmentNamePostfix = "postfix" and _sequenceId = 1, the segment name would be
+   * tableName_postfix_1
+   *
```
mcvsubbu: also add example when postfix is not specified, or seq id is not specified

mcvsubbu: also, you should set postfix to "" if it is null? Otherwise you may end up getting "null" in the name? Just verify please.
add a test in any case. thanks

PR #1086, opened 2017-02-23
Filter optimization to merge range predicates on time columns.

mayankshriv:
Predicates on time column of type (time > t1 and time <= t2)
can be optimized into a single range predicate with range (t1, t2].
This avoids potential scanning/searching of same data multiple times.

1. While this optimization can be done generically for all predicates,
   this PR does it for 'time' column only as broker does not have
   data type information for other columns.

2. Range intersection is performed for 'AND'. No merging is performed
   for 'OR' (union) in this PR (may be addressed in future PR's
   if need be).

3. If after merging a node is left with just one child with
   range predicate, the node is replaced by its child. But time
   range predicates are not pulled up in presence of other
   predicates, in this PR.

4. Added unit test for the same.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
@@ -282,18 +282,22 @@ public class BrokerRequestHandler {
       // At least one table matches the broker request.
       BrokerRequest offlineBrokerRequest = null;
       BrokerRequest realtimeBrokerRequest = null;
+
+      // We get timeColumnName from time boundary service currently, which only exists for offline table.
+      String timeColumnName = (offlineTableName != null) ? getTimeColumnName(offlineTableName) : null;
+
       if ((offlineTableName != null) && (realtimeTableName != null)) {
         // Hybrid table.
-        offlineBrokerRequest = getOfflineBrokerRequest(brokerRequest);
-        realtimeBrokerRequest = getRealtimeBrokerRequest(brokerRequest);
+        offlineBrokerRequest = _optimizer.optimize(getOfflineBrokerRequest(brokerRequest), timeColumnName);
+        realtimeBrokerRequest = _optimizer.optimize(getRealtimeBrokerRequest(brokerRequest), timeColumnName);
       } else if (offlineTableName != null) {
         // Offline table only.
         brokerRequest.getQuerySource().setTableName(offlineTableName);
-        offlineBrokerRequest = _optimizer.optimize(brokerRequest);
+        offlineBrokerRequest = _optimizer.optimize(brokerRequest, timeColumnName);
       } else {
         // Realtime table only.
         brokerRequest.getQuerySource().setTableName(realtimeTableName);
-        realtimeBrokerRequest = _optimizer.optimize(brokerRequest);
+        realtimeBrokerRequest = _optimizer.optimize(brokerRequest, timeColumnName);
       }
 
       ReduceService reduceService = _reduceServiceRegistry.get(responseType);
@@ -301,6 +305,19 @@ public class BrokerRequestHandler {
       return processOptimizedBrokerRequests(brokerRequest, offlineBrokerRequest, realtimeBrokerRequest, reduceService,
           scatterGatherStats, null, requestId);
     }
+  }
+
+  /**
+   * Returns the time column name for the table name from the time boundary service.
+   * Can return null if the time boundary service does not have the information.
+   *
+   * @param tableName Name of table for which to get the time column name
+   * @return Time column name for the table.
+   */
+  @Nullable
+  private String getTimeColumnName(@Nonnull String tableName) {
+    TimeBoundaryInfo timeBoundary = _timeBoundaryService.getTimeBoundaryInfoFor(tableName);
+    return (timeBoundary != null) ? timeBoundary.getTimeColumn() : null;
   }
 
   /**
@@ -316,7 +333,7 @@ public class BrokerRequestHandler {
     String offlineTableName = TableNameBuilder.OFFLINE_TABLE_NAME_BUILDER.forTable(hybridTableName);
     offlineRequest.getQuerySource().setTableName(offlineTableName);
     attachTimeBoundary(hybridTableName, offlineRequest, true);
-    return _optimizer.optimize(offlineRequest);
+    return offlineRequest;
   }
 
   /**
@@ -332,7 +349,7 @@ public class BrokerRequestHandler {
     String realtimeTableName = TableNameBuilder.REALTIME_TABLE_NAME_BUILDER.forTable(hybridTableName);
     realtimeRequest.getQuerySource().setTableName(realtimeTableName);
     attachTimeBoundary(hybridTableName, realtimeRequest, false);
-    return _optimizer.optimize(realtimeRequest);
+    return realtimeRequest;
   }
 
   /**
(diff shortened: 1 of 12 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -289,11 +289,11 @@ public BrokerResponse processBrokerRequest(@Nonnull BrokerRequest brokerRequest,
       } else if (offlineTableName != null) {
         // Offline table only.
         brokerRequest.getQuerySource().setTableName(offlineTableName);
-        offlineBrokerRequest = _optimizer.optimize(brokerRequest);
+        offlineBrokerRequest = _optimizer.optimize(brokerRequest, getTimeColumnName(offlineTableName));
       } else {
         // Realtime table only.
         brokerRequest.getQuerySource().setTableName(realtimeTableName);
-        realtimeBrokerRequest = _optimizer.optimize(brokerRequest);
+        realtimeBrokerRequest = _optimizer.optimize(brokerRequest, getTimeColumnName(realtimeTableName));
```
mayankshriv: Fixed

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -332,7 +345,7 @@ private BrokerRequest getRealtimeBrokerRequest(@Nonnull BrokerRequest brokerRequ
     String realtimeTableName = TableNameBuilder.REALTIME_TABLE_NAME_BUILDER.forTable(hybridTableName);
     realtimeRequest.getQuerySource().setTableName(realtimeTableName);
     attachTimeBoundary(hybridTableName, realtimeRequest, false);
-    return _optimizer.optimize(realtimeRequest);
+    return _optimizer.optimize(realtimeRequest, getTimeColumnName(realtimeTableName));
```
mayankshriv: Fixed

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -304,6 +308,19 @@ public BrokerResponse processBrokerRequest(@Nonnull BrokerRequest brokerRequest,
...
+   * Returns the time column name for the table name from the time boundary service.
+   * Can return null if the time boundary service does not have the information.
+   *
+   * @param tableName Name of table for which to get the time column name
+   * @return Time column name for the table.
+   */
+  @Nullable
+  private String getTimeColumnName(@Nonnull String tableName) {
+    TimeBoundaryInfo timeBoundary = _timeBoundaryService.getTimeBoundaryInfoFor(tableName);
+    return (timeBoundary != null) ? timeBoundary.getTimeColumn() : null;
```
jfim: Not 100% sure, but IIRC there's no time boundary for offline-only tables nor for the realtime part of a hybrid table. In that case, that would mean that most of the time this optimization wouldn't apply.

pinot-broker/src/test/java/com/linkedin/pinot/broker/requesthandler/FilterOptimizerTest.java
```diff
@@ -134,7 +135,7 @@ public void testNegative() throws Exception {
     int numOps;
 
     req = pql2Compiler.compileToBrokerRequest("SELECT * FROM T WHERE (A = 4 AND (B = 5 OR D = 9))");
-    tree = RequestUtils.generateFilterQueryTree(_optimizer.optimize(req));
+    tree = RequestUtils.generateFilterQueryTree(_optimizer.optimize(req, null /* timeColumn */));
```
jfim: I believe you can use `timeColumn` here too

pinot-broker/src/test/java/com/linkedin/pinot/broker/requesthandler/RangeMergeOptimizerTest.java
```diff
@@ -0,0 +1,183 @@
...
+    expectedTree = buildFilterQueryTree("select * from table where foo = 'bar'", false);
+    compareTrees(actualTree, expectedTree);
+  }
+
+  /**
+   * Helper method to compare two filter query trees
+   * @param actualTree Actual tree
+   * @param expectedTree Expected tree
+   */
+  private void compareTrees(FilterQueryTree actualTree, FilterQueryTree expectedTree) {
```
jfim: This part of the test is very similar to the one in `MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizerTest`, might be a good idea to refactor the tests to avoid duplicated code.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/RangeMergeOptimizer.java
```diff
@@ -0,0 +1,152 @@
...
+      if (upper1 == Long.MAX_VALUE) { // lower1 == lower2
+        stringBuilder.append(RangePredicate.UNBOUNDED + RangePredicate.UPPER_EXCLUSIVE); // * always has ')'
+      } else {
+        stringBuilder.append(lower1);
+        stringBuilder.append(
+            (predicate1.includeUpperBoundary() && predicate2.includeUpperBoundary()) ? RangePredicate.UPPER_INCLUSIVE
+                : RangePredicate.UPPER_EXCLUSIVE);
+      }
+    }
+    return Collections.singletonList(stringBuilder.toString());
```
jfim: Does this actually work correctly in the case of degenerate ranges (eg. `a < 30 AND 35 < a`)?

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
mayankshriv: Yes, in the PR the optimization will be enabled for hybrid usecases only, and we pick up the time column from offline part of the hybrid usecase.

pinot-broker/src/test/java/com/linkedin/pinot/broker/requesthandler/RangeMergeOptimizerTest.java
mayankshriv: Will address that in a separate PR.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/RangeMergeOptimizer.java
```diff
@@ -0,0 +1,165 @@
...
+    // For OR, we optimize all its children, but do not propagate up.
+    if (current.getOperator() == FilterOperator.OR) {
+      int length = children.size();
+      for (int i = 0; i < length; i++) {
+        children.set(i, optimizeRanges(children.get(i), timeColumn));
+      }
+      return current;
+    }
+
+    List<FilterQueryTree> newChildren = new ArrayList<>();
```
jfim: Can you add a comment here that says that after this point, since we have children, we're guaranteed to be an AND node (as well as an assert or if check, so that if we implement something like XOR that logic won't break).

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/RangeMergeOptimizer.java
```diff
@@ -0,0 +1,165 @@
...
+package com.linkedin.pinot.broker.requesthandler;
+
+import com.linkedin.pinot.common.request.FilterOperator;
+import com.linkedin.pinot.common.utils.request.FilterQueryTree;
+import com.linkedin.pinot.core.common.predicate.RangePredicate;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.jar.Pack200;
```
jfim: I don't think you really use the Pack200 import :)

mayankshriv: Sorry, it was an unused 'auto' import, will remove it.

PR #1090, opened 2017-02-24
Fix the bug that group by metrics can pass startree check.

Jackie-Jiang:


Jackie-Jiang:
@mayankshriv Will address in another PR.

PR #1099, opened 2017-02-28
Add support for Server side Time segment pruning.

mayankshriv:
1. Server will now prune segments based on start/end time in the segment
metadata and time filter predicate(s) on the query. Observed QPS
improvement from 700 to 1400 on one server with 40GB of data.

2. Pruning out segments results in their docs not being counted towards
totalRawDocs in the query response. Fixed it by couting the totalRawDocs
before pruning.

3. Added api's on SegmentMetadata to get TimeColumn, TimeUnit &
Start/EndTime.

4. Added unit test for the same.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
@@ -15,11 +15,16 @@
  */
 package com.linkedin.pinot.core.query.pruner;
 
-import org.apache.commons.configuration.Configuration;
-import org.joda.time.Interval;
-
 import com.linkedin.pinot.common.request.BrokerRequest;
+import com.linkedin.pinot.common.request.FilterOperator;
+import com.linkedin.pinot.common.segment.SegmentMetadata;
+import com.linkedin.pinot.common.utils.request.FilterQueryTree;
+import com.linkedin.pinot.common.utils.request.RequestUtils;
+import com.linkedin.pinot.core.common.predicate.RangePredicate;
 import com.linkedin.pinot.core.indexsegment.IndexSegment;
+import java.util.List;
+import javax.annotation.Nonnull;
+import org.apache.commons.configuration.Configuration;
 
 
 /**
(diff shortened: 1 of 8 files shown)


pinot-core/src/test/java/com/linkedin/pinot/query/pruner/TimeSegmentPrunerTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @BeforeClass
+  public void setup() {
+    _compiler = new Pql2Compiler();
+  }
+
+  @Test
+  public void test() {
+    // Equality predicate
+    String query = "select count(*) from table where time = 10";
```
antumbde: Add tests for disjoint range  < 10 or > 20 and not time column predicate

also for time < 0 or time > Long.Max

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
```diff
@@ -49,4 +123,60 @@ public void init(Configuration config) {
...
+   * Helper method that generates TimeInterval from a RangePredicate string.
+   *
+   * @param rangeString Range String
+   * @return TimeInterval for the rangeString
+   */
+  private static TimeInterval getIntervalFromString(String timeColumn, List<String> rangeString) {
+    RangePredicate range = new RangePredicate(timeColumn, rangeString);
+    String lowerString = range.getLowerBoundary();
+
+    // It is OK to add/subtract '1' without worrying about long overflow, as time column is guaranteed to be
```
antumbde: But this is user input (query) that we are parsing. So, we need to protect ?

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
```diff
@@ -49,4 +123,60 @@ public void init(Configuration config) {
...
+    // within the range TimeUtils.VALID_MIN_TIME_MILLIS and TimeUtils.VALID_MAX_TIME_MILLIS.
+    long lower = lowerString.equals(RangePredicate.UNBOUNDED) ? 0 : Long.parseLong(lowerString);
+    lower = range.includeLowerBoundary() ? lower : lower + 1;
+
+    String upperString = range.getUpperBoundary();
+    long upper = upperString.equals(RangePredicate.UNBOUNDED) ? Long.MAX_VALUE : Long.parseLong(upperString);
```
antumbde: protect from upperString == 0 or logic may break

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
```diff
@@ -396,6 +413,26 @@ public String getIndexType() {
   }
 
   @Override
+  public String getTimeColumn() {
+    return _timeColumn;
+  }
+
+  @Override
+  public long getStartTime() {
+    return _segmentStartTime;
```
mayankshriv: No it cannot be reused, because its semantics does not match start/end of segment metadata. Start and end are both inclusive in segment metadata, whereas for Interval class, the end is exclusive. Besides, it is inefficient because it requires unnecessary conversion to millis.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
```diff
@@ -33,11 +38,80 @@
...
+   * @param filterQueryTree Filter tree for the query.
+   * @param segmentInterval Time interval of the segment to be pruned.
+   * @return True if segment can be pruned out, false otherwise.
+   */
+  public static boolean pruneSegment(@Nonnull FilterQueryTree filterQueryTree, @Nonnull TimeInterval segmentInterval,
+      @Nonnull String timeColumn) {
+    List<FilterQueryTree> children = filterQueryTree.getChildren();
+
+    // Leaf Node
+    FilterOperator operator = filterQueryTree.getOperator();
```
mayankshriv: Conceptually, yes. But based on profiling results, no it does not buy anything at all.

antumbde: How many segments and what depth of query tree ? Why use slower algorithm that susceptible to query predicate sizes ?

mayankshriv: It takes < 1 microseconds to DFS on a filter query tree with 20 nodes and depth of 5.
Also, extracting the time interval range requires merging of multiple predicates (this pruner should not assume that just one predicate on time column), handling AND vs OR.

antumbde: ok,  <1micros is very small. 

Extracting time range does not have to assume single predicate. Instead of recursively calling pruneSegment, it can call "CombineRange" to get the larger time range. I generally feel we should avoid algorithms that are susceptible to user inputs.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
mayankshriv: Can you elaborate what will break?

pinot-core/src/test/java/com/linkedin/pinot/query/pruner/TimeSegmentPrunerTest.java
mayankshriv: Added tests for disjoint range negative time ranges etc. Cannot test absence of time column in this unit test as the api it calls has timeColumn as Nonnull. But made sure it is covered elsewhere.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
```diff
@@ -49,4 +123,60 @@ public void init(Configuration config) {
...
+    String upperString = range.getUpperBoundary();
+    long upper = upperString.equals(RangePredicate.UNBOUNDED) ? Long.MAX_VALUE : Long.parseLong(upperString);
+    upper = range.includeUpperBoundary() ? upper : upper - 1;
+
+    return new TimeInterval(lower, upper);
+  }
+
+  /**
+   * Utility class to represent time interval, with start and end both inclusive.
+   * The segment metadata's endTime is inclusive, where as
```
mayankshriv: Corrected.

pinot-core/src/test/java/com/linkedin/pinot/query/pruner/TimeSegmentPrunerTest.java
antumbde: I meant query without time predicate in it.

mayankshriv: Added

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
antumbde: Let me flip it and ask, if it can be done with similar code in O(Tree size + # segments), what is a good reason to do it in O(# segments * Tree size) cost ?

I believe <1 microsecond is for 1 segment. For 500 segments of a table (we are in that ballpark with some use-cases) it's about 400-500 microseconds. If clients happen to construct bad query programmatically with nested OR/ANDs with multiple of those having time columns, Tree_size in above formula increases. Numbers are very small. Latency impact is ~5-10% for low latency use cases.

mayankshriv: These ranges are strings of form (*, 100], [10, *), etc. The combine logic becomes a bit non-trivial in that case. Sorry, I am not at all convinced this can ever have any kind of noticeable impact on latency regardless of number of segments or query complexity. With large segments and/or query complexity, there are other parts of the code that take a hit as well. At this level of cpu times, I would always choose code simplicity and over premature optimization.

PR #1102, opened 2017-03-01
AlertFilterTaskConfig

cecilynie:
Fix AnomalyApplicationEndToEndTest Error with alert filter factory


thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/integration/AnomalyApplicationEndToEndTest.java
```diff
@@ -72,6 +75,8 @@
   private int id = 0;
   private String dashboardHost = "http://localhost:8080/dashboard";
   private String functionPropertiesFile = "/sample-functions.properties";
+  private String alertFilterPropertiesFile = "/sample-alertfilter.properties";
+  private String alertFilterAutoTunePropertiesFile = "/sample-alertfilterautotune.properties";
```
npawar: Is alertFilterAutotunePropertiesFile requires? Looks like it is not being used anywhere

cecilynie: It's not. Now put it as a place holder. The autotune job is now triggered only by endpoint. However I wonder in the future we will need to put it to worker. Thus I initiated it here.

PR #1103, opened 2017-03-02
Yyuen/import front end plugin

justYves:
Pre Work for a phantom JS integration 

### What's new: 
- added node to the build process
- running a babel grunt task to transpile our JS
- added watcher task (dev)
- some more minor clean up (relative path, inline scripts and missing var declaration)

### To do: 
- import sass
- watcher for templates and css files


thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/app.js
```diff
@@ -0,0 +1,18 @@
+// app entry point
+
```
justYves: copy pasta from inline scripts. Ideally we want to use es6 modules import exports and not rely on globals. (todo)

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/app.js
```diff
@@ -0,0 +1,18 @@
...
+var dataService = undefined;
+var constants = undefined;
+var HASH_SERVICE = undefined;
+var HASH_PARAMS = undefined;
+$(document).ready(function() {
```
theseyi: You could also just do
````javascript
$(function () {
//stuff happens here...
});
````
synonymous with doc ready

justYves:
@theseyi The whole point of this pr was the transpiling. One of the big step in the email flow experience is including a screenshot of the graph anomaly. Unfortunately, the current version of phantomJS doesn't handle es6 at all. (they have a new beta that does, but it seems to be buggy on OSX). Thanks for reviewing!

PR #1105, opened 2017-03-02
Fix test failure due to arbitrary test order of TestNG

cyenjung:
Observed issue: The database for testing would randomly disappear or be wiped out during the execution of test classes.

Issues that induce this problem:
1. We only have one global reference to test classes' DB.

2. We only have one global DAO registry for storing the references to different tables in the DB.

3. TestNG could interleave two different classes' @BeforeClass, @AfterClass methods. Assume that we have two class A and B. The following test order could happen: A.BeforeClass, A.tests, B.BeforeClass, A.AfterClass, B.tests, B.AfterClass. Furthermore, the database is wiped and created in @BeforeClass and database is wiped out again in @AfterClass. Due to this behavior, one test's DB could be wiped out by its subsequent or preceding test. Consequently, both tests fail at the end.

Solution:
1. Each test class has its own DB and its own reference to its DB.
2. Each test class has its own DAO Registry during its execution.

Test has done:
Local mvn test is passed.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/autoload/pinot/metrics/AutoloadPinotMetricsServiceTest.java
+++ b/thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/autoload/pinot/metrics/AutoloadPinotMetricsServiceTest.java
@@ -45,7 +45,7 @@ public class AutoloadPinotMetricsServiceTest  extends AbstractManagerTestBase {
    */
   @BeforeMethod
   void beforeMethod(){
-    DAORegistry.overrideSingletonDAORegistryForTesting(testUtils.getTestDaoRegistry());
+    DAORegistry.overrideSingletonDAORegistryForTesting(testDBResources.getTestDaoRegistry());
   }
 
   @Test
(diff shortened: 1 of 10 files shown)


thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/autoload/pinot/metrics/AutoloadPinotMetricsServiceTest.java
```diff
@@ -36,6 +37,17 @@ void afterClass() {
...
   }
 
+  /**
+   * This test class needs to use a class that accesses the global singleton DAO registry. Therefore,
+   * we need to copy the local DAO registry, which is created for this test, to the global one.
+   * Otherwise, the global singleton one may reference to an arbitrary DB that is created for this
+   * test.
+   */
+  @BeforeMethod
+  void beforeMethod(){
```
npawar: Don't we need to do this for all the other tests (TestAnomalyFunctionManager, TestAlertCOnfigManager etc)?

cyenjung: We only need this method for tests that use classes which use DAORegistry.getInstance(). Otherwise, they don't need to copy their local DAO Registry to the global singleton one.

We will work on another patch to remove the global singleton DAO Registry from the application, which is a really bad practice.

PR #1109, opened 2017-03-03
Ensure time column can be parsed as an INT/LONG before running time based pruning.

mayankshriv:
In case of bad input where time column is not really a number (ie cannot
be parsed as an INT/LONG), skip running the time based pruner.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.core.query.pruner;
 
+import com.linkedin.pinot.common.data.FieldSpec;
 import com.linkedin.pinot.common.request.BrokerRequest;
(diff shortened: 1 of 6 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/TimeSegmentPruner.java
```diff
@@ -35,19 +38,31 @@
...
+        return false;
+      }
+
+      TimeInterval segmentInterval = new TimeInterval(metadata.getStartTime(), metadata.getEndTime());
+      FilterQueryTree filterQueryTree = RequestUtils.generateFilterQueryTree(brokerRequest);
+
+      return (filterQueryTree != null && pruneSegment(filterQueryTree, segmentInterval, timeColumn));
+    } catch (Exception e) {
+      // If anything goes wrong during pruning, just return false to be safe.
+      LOGGER.error("Unexpected exception during pruning segment '{}' on time.", segment.getSegmentName(), e);
```
antumbde: this may be very noisy printing for every non-numeric-timestamp segment of every query

mayankshriv:
Closing the PR, we need a better fix for this.

PR #1110, opened 2017-03-03
Migrate min max anomaly function

cyenjung:
Implement min max anomaly function using the new anomaly function framework.

Add an unit test for min max anomaly function.

Tested with unit test and local integration test.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/detection/MinMaxThresholdDetectionModel.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/detection/MinMaxThresholdDetectionModel.java
@@ -38,7 +38,7 @@ public class MinMaxThresholdDetectionModel extends AbstractDetectionModel {
     // Compute the bucket size, so we can iterate in those steps
     long bucketMillis = anomalyDetectionContext.getBucketSizeInMS();
     Interval timeSeriesInterval = timeSeries.getTimeSeriesInterval();
-    long numBuckets = (timeSeriesInterval.getStartMillis() - timeSeriesInterval.getEndMillis()) / bucketMillis;
+    long numBuckets = Math.abs(timeSeriesInterval.getEndMillis() - timeSeriesInterval.getStartMillis()) / bucketMillis;
 
     // avg value of this time series
     averageValue /= numBuckets;
@@ -55,7 +55,7 @@ public class MinMaxThresholdDetectionModel extends AbstractDetectionModel {
         anomalyResult.setEndTime(timeBucket + bucketMillis); // point-in-time
         anomalyResult.setDimensions(dimensionMap);
         anomalyResult.setScore(averageValue);
-        anomalyResult.setWeight(Math.abs(deviationFromThreshold)); // higher change, higher the severity
+        anomalyResult.setWeight(deviationFromThreshold); // higher change, higher the severity
         anomalyResult.setAvgCurrentVal(value);
         String message =
             String.format(DEFAULT_MESSAGE_TEMPLATE, deviationFromThreshold, value, min, max);
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/detection/MinMaxThresholdDetectionModel.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/detection/MinMaxThresholdDetectionModel.java
@@ -1,0 +1,85 @@
+package com.linkedin.thirdeye.anomalydetection.model.detection;
+
+import com.linkedin.thirdeye.anomalydetection.context.AnomalyDetectionContext;
+import com.linkedin.thirdeye.anomalydetection.context.TimeSeries;
+import com.linkedin.thirdeye.api.DimensionMap;
+import com.linkedin.thirdeye.datalayer.dto.RawAnomalyResultDTO;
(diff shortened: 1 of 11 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/detection/MinMaxThresholdDetectionModel.java
```diff
@@ -0,0 +1,85 @@
...
+    if ((min != null && currentValue < min && min != 0d)) {
+      return calculateChange(currentValue, min);
+    } else if (max != null && currentValue > max && max != 0d) {
+      return calculateChange(currentValue, max);
+    }
+    return 0;
+  }
+
+  protected static double calculateChange(double currentValue, double baselineValue) {
+    return (currentValue - baselineValue) / baselineValue;
```
bl44: Do we need to consider divided by error? I also write this tiny function for sign test. It might make sense to make it as a common util function. Since many anomaly results might carry currentValue, baselineValue and change.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/detection/MinMaxThresholdDetectionModel.java
```diff
@@ -0,0 +1,85 @@
...
+      double deviationFromThreshold = getDeviationFromThreshold(value, min, max);
+
+      if (deviationFromThreshold != 0) {
+        RawAnomalyResultDTO anomalyResult = new RawAnomalyResultDTO();
+        anomalyResult.setProperties(properties.toString());
+        anomalyResult.setStartTime(timeBucket);
+        anomalyResult.setEndTime(timeBucket + bucketMillis); // point-in-time
+        anomalyResult.setDimensions(dimensionMap);
+        anomalyResult.setScore(averageValue);
+        anomalyResult.setWeight(deviationFromThreshold); // higher change, higher the severity
```
bl44: Can't really compare the 'negative' severity with 'positive' severity here, even we use absolute(severity). The definition of deviation from threshold might not be a good 'severity' to use.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/merge/AverageAnomalyMergeModel.java
```diff
@@ -28,18 +28,26 @@ public void update(AnomalyDetectionContext anomalyDetectionContext,
...
       weight += rawAnomaly.getWeight();
       score += rawAnomaly.getScore();
+      current += rawAnomaly.getAvgCurrentVal();
+      baseline += rawAnomaly.getAvgBaselineVal();
     }
     if (rawAnomalyResultDTOs.size() != 0) {
       double size = rawAnomalyResultDTOs.size();
       weight /= size;
       score /= size;
+      current /= size;
```
bl44: there computation is not correct (not consistent with raw weight, score, current, baseline definition) if you have different lengths for two anomalies to merge.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/merge/MinMaxThresholdMergeModel.java
```diff
@@ -0,0 +1,67 @@
+package com.linkedin.thirdeye.anomalydetection.model.merge;
```
bl44: I am wondering if the mergemodel is properly defined. 1) can this be called independent of 'AnomalyDetectionFunction'. this use AnomalyDetectionContext to update a merged anomaly, which seems odd. Do we actually store this context in our anomaly results? 2) I did not see where you provide this MergedAnomalyResult.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/merge/AverageAnomalyMergeModel.java
cyenjung: The assumption for this model is that the length of raw anomalies are the same, e.g., the anomalies generated by splitline regression, and does a simple average across all raw anomalies.

If the length of the raw anomalies are different, this model provides an estimation. In this case, the developer should choose a different merge model.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/merge/MinMaxThresholdMergeModel.java
cyenjung: Merged anomaly are one of the inputs of Update() method. The updated results will be written back. The input AnomalyDetectionContext should provide the necessary time series in order to update the given merged anomaly.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/model/detection/MinMaxThresholdDetectionModel.java
cyenjung: I think the original design of the weight of Min Max function is that if the time series swings below and above the min max thresholds, then the weight should be countered.

I agree that the weight calculation for this function should be redesigned, which is out of the scope of this migration.

PR #1113, opened 2017-03-03
Handle tables with no ideal state in SegmentStatusChecker.

rven1:


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -175,6 +175,8 @@ public void runSegmentMetrics() {
       }
       IdealState idealState = helixAdmin.getResourceIdealState(helixClusterName, tableName);
       if (idealState == null) {
```
antumbde: does helix return null or empty set if idealstate is empty ? Null idealstate seems to be error condition.

For empty idealstate, # of up replicas is # of replicas for the table

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/SegmentStatusCheckerTest.java
```diff
@@ -303,7 +303,9 @@ public void missingIdealTest() throws Exception {
     Assert.assertEquals(controllerMetrics.getValueOfTableGauge(tableName,
         ControllerGauge.SEGMENTS_IN_ERROR_STATE), 0);
     Assert.assertEquals(controllerMetrics.getValueOfTableGauge(tableName,
-        ControllerGauge.NUMBER_OF_REPLICAS), 0);
+        ControllerGauge.NUMBER_OF_REPLICAS), 1);
```
antumbde: Need test case for 0 segments in the table. Ideal state exists but number of partitions of the resource are 0.

rven1: Added.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -175,6 +175,9 @@ public void runSegmentMetrics() {
...
```
mcvsubbu: Change this condition to cover all cases.

idealState == null || idealState.getPartitionSet().isEmpty()

PR #1115, opened 2017-03-03
Adding AnomalyEvent provider

puneetjaiswal:
- Adding AnomalyEventProvider
- Added end point to upload custom events


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
```diff
@@ -35,6 +34,7 @@
 import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import static com.linkedin.thirdeye.anomaly.utils.ThirdeyeMetricUtil.*;
```
cyenjung: Use "import com.linkedin.thirdeye.anomaly.utils.ThirdeyeMetricUtil;" like the other files does?

puneetjaiswal: does not really matter. 
eg I would not have to write ThirdeyeMetricUtil.detectionTaskCounter, rather I would just write detectionTaskCounter while accessing it.

PR #1121, opened 2017-03-07
Adding informed query util class to query informed API

puneetjaiswal:
https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Event+Store+Design+and+Access+API
https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/EventAccess+End+Points

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/InformedQueryUtil.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/InformedQueryUtil.java
@@ -68,8 +68,9 @@ public class InformedQueryUtil {
         }
         //        informedPosts.sortInAscendingOrder();
         return informedPosts;
-
-  }return new InFormedPosts();}
+    }
+   return new InFormedPosts();
+  }
 
   public static class InFormedPosts {
     public List<InFormedPost> data = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/InformedQueryUtil.java
```diff
@@ -0,0 +1,174 @@
...
+      if (informedPosts.status.equals("success")) {
+        if (LOG.isTraceEnabled()) {
+          for (InFormedPost post : informedPosts.data) {
+            LOG.trace(post.toString());
+          }
+        }
+        //        informedPosts.sortInAscendingOrder();
+        return informedPosts;
+
+  }return new InFormedPosts();}
```
cyenjung: The format of this line is off

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/InformedQueryUtil.java
```diff
@@ -0,0 +1,174 @@
...
+    this.queryPrefix = informedUrl + INFORMED_QUERY;
+  }
+
+  public InFormedPosts retrieveInformedEvents(String informedEventName, DateTime startTimeInclusive,
+      DateTime endTimeExclusive) {
+    long startMillis = startTimeInclusive.getMillis() / TIME_PRECISION_IN_MILLIS;
+    long endMillis = endTimeExclusive.getMillis() / TIME_PRECISION_IN_MILLIS;
+    String fetchUrl =
+        new StringBuilder(String.format(queryPrefix, informedEventName)).append("&start=")
+            .append(startMillis).append("&end=").append(endMillis).toString();
```
cyenjung: Could you confirm that the entTime on InFormed is also exclusive?

puneetjaiswal: Not sure if it is. But it would matter only when we start storing these event to our local db, at that point we would not want to store duplicates.

PR #1125, opened 2017-03-08
Fix NPE and log message when rebuilding broker resource

mcvsubbu:


jenniferdai:
Thanks for fixing this.

PR #1129, opened 2017-03-09
Enable DetectionJob to send back job id

cyy0714:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobScheduler.java
```diff
@@ -161,10 +161,11 @@ public void run() {
    * @param anomalyFunction
    * @param detectionStatusToUpdate
    */
-  private void runAnomalyFunctionAndUpdateDetectionStatus(List<Long> startTimes, List<Long> endTimes,
+  private long runAnomalyFunctionAndUpdateDetectionStatus(List<Long> startTimes, List<Long> endTimes,
       AnomalyFunctionDTO anomalyFunction, List<DetectionStatusDTO> detectionStatusToUpdate) {
+    long jobExecutionId = 0l;
```
npawar: Let's return null here, similar to what you've done in runBackfill. Will help keep consistency in what is returned when detection doesn't run. Also will avoid situations where we might actually have a jobId=0L

PR #1135, opened 2017-03-11
Add server side column min/max value pruner

Jackie-Jiang:
1. Add a ColumnMinMaxValuePruner to prune segments based on the min/max value of the column.
2. Pruner works on EQUALITY and RANGE, can handle nested AND and OR.
3. Pruner can detect invalid RANGE predicate (e.g. BETWEEN 10 AND 0).
4. Add GenerateColumnMinMaxValueMode: NONE, TIME, NON_METRIC, ALL to config the columns to add min/max value to, mode can be configured through TableCustomConfig, default TIME.
5. Column min/max values are added at loading time, and flushed to segment metadata on disk.
6. Add 2 unit tests, one for generating min/max, one for pruner.


pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/columnvalue/ColumnValueSegmentPruner.java
```diff
@@ -0,0 +1,212 @@
...
+ */
+public class ColumnValueSegmentPruner implements SegmentPruner {
+
+  @Override
+  public void init(Configuration config) {
+  }
+
+  @Override
+  public boolean prune(@Nonnull IndexSegment segment, @Nonnull BrokerRequest brokerRequest) {
+    FilterQueryTree filterQueryTree = RequestUtils.generateFilterQueryTree(brokerRequest);
```
mayankshriv: We are generating the same filterQueryTree per segment. Not a blocker for this PR, but we should change it so that we do this once per query.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/columnvalue/ColumnValuePruneMode.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.core.query.pruner.columnvalue;
+
+/**
+ * Mode for {@link ColumnValueSegmentPruner}.
```
mayankshriv: Can you elaborate why we need these modes, instead of having just one MinMaxPruner? If min/max is available for a column in  the metadata, why would we not want to prune based on that? If you think it is needed, I'd prefer to have separate pruners (one for time and one for all other columns) as opposed to one overloaded pruner.

Jackie-Jiang: This is used to generate min/max value for columns. Renamed to match the actual use case.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/columnvalue/ColumnValueSegmentPruner.java
Jackie-Jiang: This needs interface change. Will address this in another PR if this cause any performance issue.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/ColumnValueSegmentPruner.java
```diff
@@ -0,0 +1,211 @@
...
+      // Leaf Node
+
+      // Skip operator other than EQUALITY and RANGE
+      if ((filterOperator != FilterOperator.EQUALITY) && (filterOperator != FilterOperator.RANGE)) {
+        return false;
+      }
+
+      ColumnMetadata columnMetadata = columnMetadataMap.get(filterQueryTree.getColumn());
+      if (columnMetadata == null) {
+        // Should not reach here after DataSchemaSegmentPruner
```
mayankshriv: As per your comment this is not expected, so should this be an error?

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/ColumnValueSegmentPruner.java
```diff
@@ -0,0 +1,211 @@
...
+      }
+
+      Comparable minValue = columnMetadata.getMinValue();
+      Comparable maxValue = columnMetadata.getMaxValue();
+
+      if (filterOperator == FilterOperator.EQUALITY) {
+        // EQUALITY
+
+        // Doesn't have min/max value set in metadata
+        if (minValue == null) {
```
mayankshriv: Please add a check for maxValue as well, just to be safe.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -414,6 +420,10 @@ public static void removeColumnMetadataInfo(PropertiesConfiguration properties,
     properties.clearProperty(getKeyFor(column, TOTAL_NUMBER_OF_ENTRIES));
     properties.clearProperty(getKeyFor(column, IS_AUTO_GENERATED));
     properties.clearProperty(getKeyFor(column, DEFAULT_NULL_VALUE));
+    properties.clearProperty(getKeyFor(column, DERIVED_METRIC_TYPE));
```
mayankshriv: Seems these two are independent bug you found, and not related to this PR, right?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/ColumnMetadata.java
```diff
@@ -122,6 +125,36 @@ public static ColumnMetadata fromPropertiesConfiguration(String column, Properti
       builder.setDerivedMetricType(derivedMetricType);
     }
 
+    // Set min/max value if available.
+    String minString = config.getString(getKeyFor(column, MIN_VALUE), null);
+    if (minString != null) {
```
mayankshriv: Lets add a check for maxString as well.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/columnminmaxvalue/ColumnMinMaxValueHandler.java
```diff
@@ -0,0 +1,137 @@
...
+import com.linkedin.pinot.core.segment.index.readers.LongDictionary;
+import com.linkedin.pinot.core.segment.index.readers.StringDictionary;
+import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
+import com.linkedin.pinot.core.segment.store.ColumnIndexType;
+import com.linkedin.pinot.core.segment.store.SegmentDirectory;
+import java.io.File;
+import org.apache.commons.configuration.PropertiesConfiguration;
+
+
+public class ColumnMinMaxValueHandler {
```
mayankshriv: Java doc to explain what the class is doing.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/columnminmaxvalue/GenerateColumnMinMaxValueMode.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Mode for generating column min/max value.
+ * <ul>
+ *   <li>NONE: do not generate on any column</li>
+ *   <li>TIME: generate on time column only</li>
+ *   <li>NON_METRIC: generate on time/dimension columns</li>
+ *   <li>ALL: generate on all columns</li>
+ * </ul>
+ */
+public enum GenerateColumnMinMaxValueMode {
```
mayankshriv: s/GenerateColumnMinMaxValueMode/ColumnMinMaxValueGeneratorMode

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/ColumnValueSegmentPruner.java
Jackie-Jiang: There is no guarantee on the order of pruner right now, so we just prune the segment when reaching here.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/ColumnValueSegmentPruner.java
Jackie-Jiang: Fixed.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
Jackie-Jiang: Yes.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/ColumnMetadata.java
Jackie-Jiang: Fixed.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/columnminmaxvalue/GenerateColumnMinMaxValueMode.java
Jackie-Jiang: Fixed.

PR #1136, opened 2017-03-12
Fix the bug where temporary segments are not properly cleanned up

Jackie-Jiang:
Delete the file after write it to data file.
When the file is copying to instead of renaming to the new file, the tmp file might be left in the dir.
Add a warning when multiple items are uploaded.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
@@ -15,51 +15,11 @@
  */
 package com.linkedin.pinot.controller.api.restlet.resources;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.linkedin.pinot.common.metadata.segment.RealtimeSegmentZKMetadata;
-import com.linkedin.pinot.common.utils.helix.HelixHelper;
-import com.linkedin.pinot.controller.helix.core.PinotHelixSegmentOnlineOfflineStateModelGenerator;
-import java.io.File;
-import java.io.IOException;
-import java.io.UnsupportedEncodingException;
-import java.net.InetAddress;
-import java.net.URLDecoder;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import org.apache.commons.configuration.ConfigurationException;
-import org.apache.commons.fileupload.FileItem;
-import org.apache.commons.fileupload.disk.DiskFileItemFactory;
-import org.apache.commons.io.FileUtils;
-import org.apache.helix.HelixAdmin;
-import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang.Validate;
-import org.apache.helix.ZNRecord;
-import org.apache.helix.model.ExternalView;
-import org.apache.helix.model.IdealState;
-import org.apache.helix.store.zk.ZkHelixPropertyStore;
-import org.joda.time.Interval;
-import org.json.JSONArray;
-import org.json.JSONException;
-import org.restlet.data.MediaType;
-import org.restlet.data.Status;
-import org.restlet.ext.fileupload.RestletFileUpload;
-import org.restlet.representation.FileRepresentation;
-import org.restlet.representation.Representation;
-import org.restlet.representation.StringRepresentation;
-import org.restlet.resource.Delete;
-import org.restlet.resource.Post;
-import org.restlet.util.Series;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import com.alibaba.fastjson.JSONObject;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.linkedin.pinot.common.config.OfflineTableConfig;
 import com.linkedin.pinot.common.config.TableNameBuilder;
 import com.linkedin.pinot.common.metadata.ZKMetadataProvider;
-import com.linkedin.pinot.common.metadata.segment.OfflineSegmentZKMetadata;
 import com.linkedin.pinot.common.metrics.ControllerMeter;
 import com.linkedin.pinot.common.restlet.swagger.HttpVerb;
 import com.linkedin.pinot.common.restlet.swagger.Parameter;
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -15,31 +15,52 @@
  */
 package com.linkedin.pinot.controller.api.restlet.resources;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.linkedin.pinot.common.metadata.segment.RealtimeSegmentZKMetadata;
+import com.alibaba.fastjson.JSONObject;
```
mayankshriv: Probably you did not change this, just optimized imports. But seems like different json libraries being used here? I thought we were using jackson?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -361,13 +352,17 @@ public Representation post(Representation entity) {
...
-              fi.write(dataFile);
+          for (FileItem fileItem : items) {
+            if (!found) {
+              if (fileItem.getFieldName() != null) {
+                found = true;
+                dataFile = new File(tempDir, fileItem.getFieldName());
+                fileItem.write(dataFile);
+              }
+            } else {
+              LOGGER.warn("Get extra file item: " + fileItem.getFieldName());
```
mayankshriv: s/Get/Got. Add more information, eg what was the code doing when this happened.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -361,13 +352,17 @@ public Representation post(Representation entity) {
           // list of FileItems
           items = upload.parseRequest(getRequest());
 
-          for (final Iterator<FileItem> it = items.iterator(); it.hasNext() && !found;) {
-            final FileItem fi = it.next();
-            if (fi.getFieldName() != null) {
-              found = true;
-              dataFile = new File(tempDir, fi.getFieldName());
-              fi.write(dataFile);
+          for (FileItem fileItem : items) {
```
mayankshriv: Can multiple pushes of the same file at the same time cause any concurrency issue here?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: Seems not just this file use this json library. Will clean up in the future if needed.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: Should not. While we need to try this on a controller to verify that.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: Fixed

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -361,13 +352,17 @@ public Representation post(Representation entity) {
...
+            if (!found) {
+              if (fileItem.getFieldName() != null) {
+                found = true;
+                dataFile = new File(tempDir, fileItem.getFieldName());
+                fileItem.write(dataFile);
+              }
+            } else {
+              LOGGER.warn("Got extra file item while pushing segments: " + fileItem.getFieldName());
             }
+            fileItem.delete();
```
mcvsubbu: From what i read on commons io and fileupload packages, and from what we have seen, the temporary file should be cleaned up. The documentation reads that the file is cleaned up when the object is garbage collected. I am not sure if deleting the file actively will cause something else to break. 

I saw some code in newer version of fileupload utils that deletes the file in their code.

mcvsubbu: Also, it will be a good idea to field any exceptions and log a warning so that the upload does not fail if exceptions are thrown for any reason

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
antumbde: Be careful with this. Some annotations may break unless you explicitly intend to use this library.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -361,12 +352,24 @@ public Representation post(Representation entity) {
...
+              LOGGER.warn("Got extra file item while pushing segments: " + fileItem.getFieldName());
+            }
+
+            // TODO: remove the try-catch after verifying it will not throw any exception
+            try {
+              // Remove the temp file
+              // When the file is copied to instead of renamed to the new file, the temp file might be left in the dir
+              fileItem.delete();
+            } catch (Exception e) {
+              LOGGER.error("Caught exception while deleting the temp file, should not reach here", e);
```
antumbde: log file name

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -361,12 +352,24 @@ public Representation post(Representation entity) {
...
+                fileItem.write(dataFile);
+              }
+            } else {
+              LOGGER.warn("Got extra file item while pushing segments: " + fileItem.getFieldName());
+            }
+
+            // TODO: remove the try-catch after verifying it will not throw any exception
+            try {
+              // Remove the temp file
+              // When the file is copied to instead of renamed to the new file, the temp file might be left in the dir
```
antumbde: why not rename instead ? That will be more efficient.

Jackie-Jiang: The internal implementation of write() will first try to rename, if rename failed, it will copy instead. We don't have control on that.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: If delete throws exception (which should not happen), get file name from fileItem might throw another exception which break the file upload.

PR #1137, opened 2017-03-13
Use the smallest value between current and baseline bucket number

cyenjung:
Problem: Exception when showing time over time comparison between two weeks and one of the week contains DST changes.

Fix: Ensure the dashboard does not go out of boundary of both current and baseline time series

Tested on local.


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/client/comparison/TimeOnTimeResponseParser.java
```diff
@@ -72,7 +72,7 @@ public TimeOnTimeResponseParser(ThirdEyeResponse baselineResponse, ThirdEyeRespo
 
     metricFunctions = baselineResponse.getMetricFunctions();
     numMetrics = metricFunctions.size();
-    numTimeBuckets = baselineRanges.size();
+    numTimeBuckets = Math.min(currentRanges.size(), baselineRanges.size());
```
kishoreg: Add a log when the ranges are not equal and chose the smaller among the two. Explicitly state that this can happen during DST

cyenjung: Done!

PR #1142, opened 2017-03-14
[PINOT-4603] Re-factor parameter passing in Segment Completion Protocol

mcvsubbu:
This is the first check-in towards adding a lease-based segment build time.
Changing the parmeters to be a class makes it easier to add additional parameters
and features around it

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
@@ -43,6 +43,7 @@ import com.alibaba.fastjson.JSONObject;
  * https://github.com/linkedin/pinot/wiki/Low-level-kafka-consumers
  */
 public class SegmentCompletionProtocol {
+
   /**
    * MAX_HOLD_TIME_MS is the maximum time (msecs) for which a server will be in HOLDING state, after which it will
    * send in a SegmentConsumedRequest with its current offset in kafka.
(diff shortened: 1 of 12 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -144,17 +146,65 @@ protected String getUri(String msgType) {
         PARAM_INSTANCE_ID + "=" + _instanceId +
           (_reason == null ? "" : ("&" + PARAM_REASON + "=" + _reason));
     }
+
+    public static class Params {
+      private static final String UNKNOWN_SEGMENT = "UNKNOWN_SEGMENT";
+      private static final long UNKNOWN_OFFSET = -1L;
+      private static final String UNKNOWN_INSTANCE = "UNKNOWN_INSTANCE";
+
+      long _offset;
```
jfim: Fields should be private

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -225,8 +271,46 @@ public long getOffset() {
...
+      builder.append("{\"" + STATUS_KEY + "\":" + "\"" + _status.name() + "\"," + "\"" + OFFSET_KEY + "\":" + _offset);
+      builder.append("}");
+      return builder.toString();
+    }
+
+    public static class Params {
+      private static ControllerResponseStatus DEFAULT_RESPONSE_STATUS = ControllerResponseStatus.FAILED;
+      private static long DEFAULT_RESP_OFFSET = -1L;
+
+      ControllerResponseStatus _status;
```
jfim: `private`

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -144,17 +146,65 @@ protected String getUri(String msgType) {
...
+      public Params reset() {
+        resetInternal();
+        return this;
+      }
+      private void resetInternal() {
+        _offset = UNKNOWN_OFFSET;
+        _segmentName = UNKNOWN_SEGMENT;
+        _instanceId = UNKNOWN_INSTANCE;
+      }
+      public Params setOffset(long offset) {
```
jfim: Since this is a fluent interface-style class and not a Javabean setter, can you _not_ use the Javabeans names for setters (`setX`) and rather use a name that doesn't overlap with the Javabeans convention.

I skimmed over this code assuming it was standard Javabean property setters and was surprised to see it used later on in a fluent way. I suggest `withX` as the convention for this (eg. `Params.withFoo(1).withBar(2)`) which is readable and doesn't overlap with other conventions.

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
mcvsubbu: ok

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
mcvsubbu: ok

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
mcvsubbu: ok

PR #1147, opened 2017-03-14
[TE] Yyuen/post enchancements for launch

justYves:
### What's new: 
- included InGraph link (when available) 
- hooked up new user feedback 
- added data sets to search results and investigate page 
- Fixed a bug where c3 graphs will take awhile to resize:
  - the above was caused by the fetching blocking the c3 rerender call 
  - Reimplemented data caching so that it works with hashParams
  - calling destroy/rerender on `shouldTransition` 
- Added spinner to investigation page and fix the anomaly page's spinner

### Screenshots: 

screen shot 2017-03-14 at 3 05 37 pm
screen shot 2017-03-14 at 3 08 29 pm
screen shot 2017-03-14 at 3 13 26 pm

### JIRA: 
- https://jira01.corp.linkedin.com:8443/browse/THIRDEYE-1120
- https://jira01.corp.linkedin.com:8443/browse/THIRDEYE-1121
- https://jira01.corp.linkedin.com:8443/browse/THIRDEYE-1123 
- https://jira01.corp.linkedin.com:8443/browse/THIRDEYE-1145
- https://jira01.corp.linkedin.com:8443/browse/THIRDEYE-1160

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/models/AnomalyResultModel.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/models/AnomalyResultModel.js
@@ -30,6 +30,22 @@ function AnomalyResultModel() {
 }
 
 AnomalyResultModel.prototype = {
+  hasSameParams(params) {
+    if (!params.anomaliesSearchMode || !this.anomaliesSearchMode) {
+      return false;
+    }
+    const paramKeys = Object.keys(params);
+    for (let index in paramKeys){
+      let key = paramKeys[index];
+      if (key !== 'tab' && key !== 'rand') {
+        if (!!this[key] && !HASH_PARAMS.isSame(key, params[key],this[key])) {
+          return false;
+        }
+      }
+    }
+    return true;
+  },
+
   reset : function() {
     console.log('Reset of anomaly result model');
     this.metricIds = null;
@@ -40,9 +56,8 @@ AnomalyResultModel.prototype = {
 
   },
   // Call setParams every time there is a change to the model
-  setParams : function() {
+  setParams : function(params) {
     console.log("Set params of Anomaly Result Model");
-    var params = HASH_SERVICE.getParams();
     if (params != undefined) {
       console.log("params");
       if (params[HASH_PARAMS.ANOMALIES_ANOMALIES_SEARCH_MODE] != undefined) {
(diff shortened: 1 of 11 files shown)


thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnomalyResultView.js
```diff
@@ -210,6 +213,10 @@ AnomalyResultView.prototype = {
 
 
   },
+
+  destroy() {
+    $("#anomaly-results-place-holder").html('');
```
justinpark: remove would be better to detach the event binders.
If you'd like to keep the `#anomaly-results-place-holder` element, `.children().remove()` would be fine.

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/InvestigateView.js
```diff
@@ -28,31 +30,43 @@ InvestigateView.prototype = {
...
+    if (anomalyFeedback) {
+      $(`input[name=feedback-radio][value=${anomalyFeedback}]`).prop('checked', true);
+    }
+    this.renderAnomalyChart(this.investigateModel.getAnomaly());
     // this.setupListenerOnViewContributionLink();
     this.setupListenerOnUserFeedback();
   },
 
+  destroy() {
+    $("#investigate-place-holder").html('');
```
justinpark: same here

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnomalyResultView.js
justYves: Thanks for the tip! I'll make the change

thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/models/AnomalyResultModel.js
```diff
@@ -30,6 +30,22 @@ function AnomalyResultModel() {
 }
 
 AnomalyResultModel.prototype = {
+  hasSameParams(params) {
+    if (!params.anomaliesSearchMode || !this.anomaliesSearchMode) {
+      return false;
+    }
+    const paramKeys = Object.keys(params);
+    for (let index in paramKeys){
```
theseyi: I'd use `Array#every` in this case, it's more declarative than imperatively going through with a `for...in` since you're trying to check that each element meets a condition. Also allows for easier to read constructs, less lines and more maintainable. I would also add a comment so future you knows exactly what the intention of this code is and the "why".

It looks like you are trying to check, in the positive, that?
- each `param` in `params` is a *tab* or a *rand*
- each `param` is `HASH_PARAMS.isSame` as `this[param]`


If so, you can do:

````javascript
// **You can modify this comment, not fully aware of the context here
// Check that each param is either a tab or a rand
// and is the same as the HASH_PARAMS
return Object.keys(params).every((param) => {
  const isParamTabOrRand = ['tab', 'rand'].includes(param);
  const isParamSameAsHash = HASH_PARAMS.isSame(param, params[param], this[param]);

  return isParamTabOrRand && isParamSameAsHash;
});

justYves: Yeah, this code needs comment as I can see it's a bit hard to understand what the intention is. Don't think .every would work for me here, but I'll refactor and make it cleaner. Thank you!

PR #1153, opened 2017-03-15
Update merge anomaly info even it has only one raw anomaly

cyenjung:
The avgBaselineVal and avgCurrentVal are not updated when the merged anomaly has only one raw anomaly. In the original code, the information is updated through default calculation, which is not working for unknown reasons.

This PR adds another default calculator which copies the information from the raw anomaly to the merged anomaly if there is only one raw anomaly exists.

Tested on local


justYves:
Thank you for fixing this! 👍  (Not comfortable enough with back end to approve this though)

PR #1154, opened 2017-03-15
Add TableRetentionValidator to validate the retention policy in table config, and the start/end timestamp in segment metadata

Jackie-Jiang:
Check on 'segmentPushType', retention duration on APPEND push type for table config.
Check time unit, start/end timestamp in segment metadata for APPEND push type table.
Add it into the pinot-tools collection.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
@@ -1,0 +1,195 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.controller.util;
+
+import com.linkedin.pinot.common.config.AbstractTableConfig;
+import com.linkedin.pinot.common.config.SegmentsValidationAndRetentionConfig;
+import com.linkedin.pinot.common.metadata.ZKMetadataProvider;
+import com.linkedin.pinot.common.metadata.segment.OfflineSegmentZKMetadata;
+import com.linkedin.pinot.common.utils.time.TimeUtils;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.apache.helix.PropertyPathConfig;
+import org.apache.helix.PropertyType;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.manager.zk.ZKHelixAdmin;
+import org.apache.helix.manager.zk.ZNRecordSerializer;
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The <code>TableRetentionValidator</code> class validates the retention policy in table config, and the start/end
+ * timestamp in segment metadata.
+ * <p>Will validate the followings:
+ * <ul>
+ *   <li>
+ *     Table Config
+ *     <ul>
+ *       <li>"segmentsConfig" is set.</li>
+ *       <li>"segmentPushType" is set to APPEND or REFRESH.</li>
+ *       <li>Retention setting is valid for APPEND push type.</li>
+ *     </ul>
+ *   </li>
+ *   <li>
+ *     Segment Metadata
+ *     <ul>
+ *       <li>For APPEND push type, offline segment start/end time and time unit is valid.</li>
+ *     </ul>
+ *   </li>
+ * </ul>
+ */
+public class TableRetentionValidator {
+  public static final long DEFAULT_DURATION_IN_DAYS_THRESHOLD = 365;
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(TableRetentionValidator.class);
+
+  private final String _clusterName;
+  private final ZKHelixAdmin _helixAdmin;
+  private final ZkHelixPropertyStore<ZNRecord> _propertyStore;
+
+  private String _tableNamePattern = null;
+  private long _durationInDaysThreshold = DEFAULT_DURATION_IN_DAYS_THRESHOLD;
+
+  public TableRetentionValidator(@Nonnull String zkAddress, @Nonnull String clusterName) {
+    _clusterName = clusterName;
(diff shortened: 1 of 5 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
```diff
@@ -0,0 +1,176 @@
...
+ *     <ul>
+ *       <li>For APPEND push type, offline segment start/end time and time unit is valid.</li>
+ *     </ul>
+ *   </li>
+ * </ul>
+ */
+public class TableRetentionValidator {
+  private static final Logger LOGGER = LoggerFactory.getLogger(TableRetentionValidator.class);
+
+  private static final String TABLE_CONFIG_PATH = "/CONFIGS/TABLE/";
```
antumbde: remove. Use ZKMetadataProvider

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
```diff
@@ -0,0 +1,176 @@
...
+      throws Exception {
+    // Get all resources in cluster
+    List<String> resourcesInCluster = _helixAdmin.getResourcesInCluster(_clusterName);
+
+    for (String tableName : resourcesInCluster) {
+      // Skip non-table resources
+      if (!tableName.endsWith("_OFFLINE") && !tableName.endsWith("_REALTIME")) {
+        continue;
+      }
+      LOGGER.info("Table: {}", tableName);
```
antumbde: info statements here are very verbose if we are doing for all tables

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+        long endTimeInMillis = segmentTimeUnit.toMillis(offlineSegmentMetadata.getEndTime());
+        if (!TimeUtils.timeValueInValidRange(endTimeInMillis)) {
+          LOGGER.error("ERROR: Segment: {} has invalid end time in millis: {}", segmentName, endTimeInMillis);
+        }
+      }
+    }
+  }
+
+  private AbstractTableConfig getTableConfig(String tableName)
```
antumbde: Use ZKMetadataProvider here and for method below to read segment metadata

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
```diff
@@ -0,0 +1,176 @@
...
+ *   </li>
+ *   <li>
+ *     Segment Metadata
+ *     <ul>
+ *       <li>For APPEND push type, offline segment start/end time and time unit is valid.</li>
+ *     </ul>
+ *   </li>
+ * </ul>
+ */
+public class TableRetentionValidator {
```
antumbde: If the validator relies on logging, it's better to move this to tools instead of controller

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
```diff
@@ -0,0 +1,176 @@
...
+        LOGGER.error("ERROR: null push type");
+        continue;
+      } else if (segmentPushType.equalsIgnoreCase("REFRESH")) {
+        LOGGER.info("Push type: REFRESH");
+        continue;
+      } else if (segmentPushType.equalsIgnoreCase("APPEND")) {
+        LOGGER.info("Push type: APPEND");
+      } else {
+        LOGGER.error("ERROR: Invalid push type: {}", segmentPushType);
+        continue;
```
antumbde: For this and some errors below that validate retention time/unit, it's better to proceed further to get error information on segment metadata. Otherwise we'll have to run this multiple times after fixing each problem

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
Jackie-Jiang: The reason I put it in controller is that controller can integrate this into the validator in the future. When you run it from PinotToolLauncher, it can log properly.

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
Jackie-Jiang: Fixed.

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
Jackie-Jiang: Will make it an one line error log for failed tables.

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
Jackie-Jiang: If push type is not APPEND, we will skip checking the segments in RetentionManager, which means the segment metadata can be valid even without start/end timestamp (e.g. a table without time column).

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
Jackie-Jiang: Fixed.

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
antumbde: ok, makes sense for push type. But if retention time/unit validation fails below, it's better to validate segment metadata also to have complete data in one run.

Jackie-Jiang: While there is a fairly big chance that all segments in that table are invalid, which might pollute the logs.

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableRetentionValidator.java
```diff
@@ -0,0 +1,195 @@
...
+      List<String> errorMessages = new ArrayList<>();
+      for (String segmentName : segmentNames) {
+        OfflineSegmentZKMetadata offlineSegmentMetadata = getOfflineSegmentMetadata(tableName, segmentName);
+        TimeUnit segmentTimeUnit = offlineSegmentMetadata.getTimeUnit();
+        if (segmentTimeUnit == null) {
+          errorMessages.add("Segment: " + segmentName + " has null time unit");
+          continue;
+        }
+        long startTimeInMillis = segmentTimeUnit.toMillis(offlineSegmentMetadata.getStartTime());
+        if (!TimeUtils.timeValueInValidRange(startTimeInMillis)) {
```
antumbde: Some clusters have time in format yyyymmdd which is 20160401 for example. That's a "valid" time as per this check but not valid practically. For our usecase, this is important. Maybe we can parse timestamp as yyymmdd format and see if those values look valid and within timerange ? Another option is to check that the time is within last, say, 5 years

Jackie-Jiang: In SegmentZKMetadata, we never use yymmdd format as start/end timestamp. We already parsed them when generating the timestamp.

PR #1157, opened 2017-03-16
Refactor segment creation stats gathering

jfim:
Refactor the stats gathering part of segment creation by moving the
stats gathering logic into a SegmentCreationDataSource. This allows
data sources to provide stats from auxiliary information (eg. existing
dictionaries for realtime segments or other file metadata)

Other minor cleanups as well.


jfim:
Checked with Mayank, he's okay with these changes as well.

PR #1162, opened 2017-03-17
Fix IndexOutOfBoundsException while generating minmaxvalue for empty segment

Jackie-Jiang:


antumbde:
Please test this on one server before cluster rollout

PR #1165, opened 2017-03-17
[PINOT-4637] Configure the average number of multi-values per row.

mcvsubbu:
For multi-valued columns in realtime segments, we allocate memory initial memory assuming that the average
number of values in any row for that column is 10. Changed the default to 2, and made it configurable
for individual server instances if needed.

Modifed the tests to print the seed value in case of exceptions

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/config/TableDataManagerConfig.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/config/TableDataManagerConfig.java
@@ -15,6 +15,12 @@
  */
 package com.linkedin.pinot.core.data.manager.config;
 
+import java.util.List;
+import org.apache.commons.configuration.Configuration;
+import org.apache.commons.configuration.ConfigurationException;
+import org.apache.commons.configuration.PropertiesConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import com.linkedin.pinot.common.config.AbstractTableConfig;
 import com.linkedin.pinot.common.config.IndexingConfig;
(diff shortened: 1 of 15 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/config/TableDataManagerConfig.java
```diff
@@ -41,6 +41,8 @@
   private static final String READ_MODE = "readMode";
   private static final String TABLE_DATA_MANAGER_DATA_DIRECTORY = "directory";
   private static final String TABLE_DATA_MANAGER_NAME = "name";
+  private static final String REALTIME_AVG_MULTI_VALUE_COUNT = "avgMvCount";
```
jfim: Do we want count here or length? Would "realtime average multi value entries per row" be a more appropriate description?

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixInstanceDataManagerConfig.java
```diff
@@ -33,6 +32,9 @@
 public class HelixInstanceDataManagerConfig implements InstanceDataManagerConfig {
   private static final Logger LOGGER = LoggerFactory.getLogger(HelixInstanceDataManagerConfig.class);
 
+  // Average number of values in multi-valued columns in any table in this instance.
+  // This value is used to allocate initial memory for multi-valued columns in realtime segments in consuming state.
+  private static final String AVERAGE_MV_COUNT = "realtime.avgMultiValueCount";
```
jfim: Whatever name we pick from earlier, let's make sure it's also consistent and understandable for the end user.

PR #1174, opened 2017-03-20
[PINOT-4603] Added server side code for extending lease for build time

mcvsubbu:


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/SegmentBuildTimeLeaseExtender.java
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * Adds a segment for periodic lease request.
+   * The first lease request is sent before {@param initialBuildTimeMs} exipres. Subsequent lease requests are sent
+   * within two minutes.
+   * @param segmentId is the name of he segment that is being built
+   * @param initialBuildTimeMs is the initial time budget that SegmentCompletionManager has allocated.
+   * @param offset The offset at which this segment is being built.
+   */
+  public void addSegment(String segmentId, long initialBuildTimeMs, long offset) {
+    final long initialDelayMs = initialBuildTimeMs * 9 / 10;
```
rven1: need to Sanity check initial build time. This cannot be negative, or larger than a specified value.
Should be sending lease requests to prevent waiting for a dead server longer than 2 minutes.

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -207,8 +208,8 @@ public Params withWaitTimeMillis(long waitTimeMillis) {
         _waitTimeMillis = waitTimeMillis;
         return this;
       }
-      public Params withExtTimeSec(int extTimeSec) {
-        _extTimeSec = extTimeSec;
+      public Params withExtraTimeSec(int exttraTimeSec) {
```
jfim: 1 `t` for extra

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -257,7 +267,7 @@ public SegmentCommitRequest(Params params) {
     }
     @Override
       public String getUrl(final String hostPort) {
```
jfim: Weird indent, should be the same level as the `@Override`

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -233,12 +234,21 @@ public long getBuildTimeMillis() {
...
     }
   }
 
+  public static class ExtendBuildTimeRequest extends Request {
+    public ExtendBuildTimeRequest(Params params) {
+      super(params);
+    }
+    @Override
+    public String getUrl(final String hostPort) {
+      return "http://" + hostPort + getUri(MSG_TYPE_EXTEND_BUILD_TIME);
```
jfim: Can we have a default implementation of this, it looks they're all the same except for the message type anyway.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/SegmentBuildTimeLeaseExtender.java
mcvsubbu: i will add code to sanity check, good point.

I disagree with the other comment about sending every 2 minutes even if the controller lease time is high. Unless needed, we don't want to send lease req messages. We will evolve this code to learn the build time so that no config will be needed.

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
mcvsubbu: will fix

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
mcvsubbu: will do

PR #1179, opened 2017-03-21
[TE] unbreak logging in unit tests

apucher:
* unbreaks logging when running tests, etc. (i.e. logging only worked in main dashboard and application)


apucher:
all metrics enabled again.
mvn clean test: log output from 2363 lines (before) back up to 5908 lines (after)

PR #1181, opened 2017-03-22
Support for partitioned data during segment generation and query execution.

mayankshriv:
1. Segment Generation: Added config to specify the column with
partitioned data. Config contains:
   - Name (and arguments) for the partition function.
   - Expected partition range(s) for the segment.
   Segment generator checks if values for partitioned columns are within
specified ranges. If yes, then partition info is written to metadata,
otherwise not.

2. Partition based segment pruning: If there are predicates on
partitioned columns, then segments for which the predicates cannot be
satisfied are pruned out from query planning and execution.

3. Unit tests for new code.


pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/ColumnMetadata.java
```diff
@@ -157,9 +164,27 @@ public static ColumnMetadata fromPropertiesConfiguration(String column, Properti
       }
     }
 
+
+    String partitionFunctionString = config.getString(getKeyFor(column, V1Constants.MetadataKeys.Column.PARTITION_FUNCTION));
```
mayankshriv: If you think of it as an attribute of the column that indicates values of this column are expected to follow a certain mathematical constraint, then it seems more appropriate to be property of a column. Also, are there any existing segment level properties with column as key (I could not find one using example segments).

antumbde: Partitioning is a segment property with mathematical constraint applied on a set of columns. Putting this on column metadata structurally impacts composite partitioning keys.

The property should be segment.partitioning.columns=a,b,c. There are plenty of examples for that.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
```diff
@@ -535,16 +547,14 @@ void buildIndexCreationInfo()
   @Override
   /**
    * Returns the name of the segment associated with this index creation driver.
-   */
-  public String getSegmentName() {
+   */ public String getSegmentName() {
```
mcvsubbu: please put back the newline here

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
```diff
@@ -535,16 +547,14 @@ void buildIndexCreationInfo()
...
     return segmentName;
   }
 
   @Override
   /**
    * Returns the path of the output directory
-   */
-  public File getOutputDirectory() {
+   */ public File getOutputDirectory() {
```
mcvsubbu: same

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/stats/AbstractColumnStatisticsCollector.java
```diff
@@ -120,4 +141,48 @@ Object addressNull(Object entry, DataType e) {
...
+   *
+   * @param value Column value to check.
+   */
+  protected void checkPartition(Object value) {
+    if (partitionFunction != null) {
+      int partition = partitionFunction.getPartition(value);
+
+      for (IntRange partitionValue : partitionValues) {
+        if (partition >= partitionValue.getMinimumInteger() && partition <= partitionValue.getMaximumInteger()) {
+          return;
```
mcvsubbu: Do you want to add a counter for number of items not in range?
Or, do you want to log the situation differently if the value was out of range?

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/PartitionFunctionFactory.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * This method generates and returns a partition function based on the provided string.
+   *
+   * @param partitionFunctionString String containing function name and its arguments.
+   * @return Partition function
+   */
+  public static PartitionFunction getPartitionFunction(@Nonnull String partitionFunctionString) {
+    String[] tokens = partitionFunctionString.split(PARTITION_FUNCTION_DELIMITER);
```
antumbde: has test ?

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Helper method to convert an array of ranges in string form (eg [2 3]) into a list
+   * of {@link IntRange}. Expects each string range to be formatted correctly.
+   *
+   * @param inputs Array of ranges in string form.
+   * @return List of IntRange's for the given input.
+   */
+  public static List<IntRange> rangesFromString(String[] inputs) {
+    List<IntRange> ranges = new ArrayList<>(inputs.length);
+    for (String input : inputs) {
```
antumbde: regex match will be simpler I guess (minor and optional)

UT ?

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
```diff
@@ -0,0 +1,125 @@
...
+  public ColumnPartitionConfig() {
+
+  }
+
+  /**
+   * Constructor for the class.
+   *
+   * @param functionName Name of the partition function, includes delimiter separated arguments. For example:
+   *                     "module\t\t20": Function name is module, with argument 20 (divisor).
+   * @param partitionValues Comma separated ranges expected for column values (e.g. [1 2], [3 5], [7 9]).
```
antumbde: means values can not have spaces ? We should handle that to either prevent, ignore or allow

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
```diff
@@ -0,0 +1,125 @@
...
+   */
+  public ColumnPartitionConfig(@Nonnull String functionName, String partitionValues) {
+    _functionName = functionName;
```
antumbde: validate nonnull input since other code doesn't tolerate that. 
PartitionValues should also be nonnull

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentPartitionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.apache.commons.lang.math.IntRange;
+import org.codehaus.jackson.annotate.JsonIgnoreProperties;
+
+
+@SuppressWarnings("unused") // Suppress incorrect warning, as methods are used for json ser/de.
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class SegmentPartitionConfig {
+  Map<String, ColumnPartitionConfig> _columnPartitionMap;
```
antumbde: final

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentPartitionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+    return _columnPartitionMap;
+  }
+
+  /**
+   * Returns the partition function for the given column, null if there isn't one.
+   *
+   * @param column Column for which to return the partition function.
+   * @return Partition function for the column.
+   */
+  @Nullable
```
antumbde: Thumbs up for adding nullable annotations here. Thanks!

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/AbstractSegmentPruner.java
```diff
@@ -0,0 +1,101 @@
...
+      case OR:
+        for (FilterQueryTree child : children) {
+          if (!pruneSegment(child, columnMetadataMap)) {
+            return false;
+          }
+        }
+        return true;
+
+      default:
+        throw new IllegalStateException("Unsupported filter operator: " + filterOperator);
```
antumbde: Should it be silent instead of throwing ? (not sure of all possible operators we can get here

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/StatsCollectorConfig.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.commons.lang.math.IntRange;
+
+
+/**
+ * Config class for Stats collector, contains all the configs and parameters required to build
+ * the stats collector.
+ */
+public class StatsCollectorConfig {
+
+  private Schema _schema;
```
antumbde: final

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/StatsCollectorConfig.java
```diff
@@ -0,0 +1,66 @@
...
+  private SegmentPartitionConfig _segmentPartitionConfig;
+
+  /**
+   * Constructor for the class.
+   * @param schema Data schema
+   * @param segmentPartitionConfig Segment partitioning config
+   */
+  public StatsCollectorConfig(@Nonnull Schema schema, @Nullable SegmentPartitionConfig segmentPartitionConfig) {
+    _schema = schema;
```
antumbde: check inputs not null

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/stats/AbstractColumnStatisticsCollector.java
```diff
@@ -120,4 +141,48 @@ Object addressNull(Object entry, DataType e) {
...
+        }
+      }
+
+      LOGGER.warn("Column '{}' not partitioned as specified for value: '{}', dropping partition metadata.", column,
```
antumbde: should it throw because warnings will be lost and it will result in problems downstream

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Map;
+import org.apache.commons.configuration.Configuration;
+import org.apache.commons.lang.math.IntRange;
+
+
+/**
+ * Implementation of {@link SegmentPruner} that uses partition information to perform pruning.
+ */
+public class PartitionSegmentPruner extends AbstractSegmentPruner {
+  @Override
```
antumbde: Add UT

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
```diff
@@ -0,0 +1,125 @@
...
+    return ranges;
+  }
+
+  /**
+   * Helper method to convert a list of {@link IntRange} to a delimited string.
+   * The delimiter used is {@link #PARTITION_VALUE_DELIMITER}
+   * @param ranges List of ranges to be converted to String.
+   * @return String representation of the lis tof ranges.
+   */
+  public static String rangesToString(List<IntRange> ranges) {
```
antumbde: UT

pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
```diff
@@ -0,0 +1,215 @@
...
+   * Unit test for {@link PartitionSegmentPruner}.
+   * <ul>
+   *   <li> Generates queries with equality predicate on partitioned column with random values. </li>
+   *   <li> Ensures that column values that are in partition range ([0 5]) do not prune the segment,
+   *        whereas other values do. </li>
+   *   <li> Ensures that predicates on non-partitioned columns do not prune the segment. </li>
+   * </ul>
+   */
+  @Test
+  public void testPruner() {
```
antumbde: Add positive and negative test. 
Also cover AND and OR clauses

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
```diff
@@ -0,0 +1,125 @@
...
+import org.codehaus.jackson.annotate.JsonIgnore;
+import org.codehaus.jackson.annotate.JsonIgnoreProperties;
+
+
+@SuppressWarnings("unused") // Suppress incorrect warnings as methods used for ser/de.
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ColumnPartitionConfig {
+  public static final String PARTITION_VALUE_DELIMITER = ",";
+
+  String _functionName;
```
mayankshriv: Will make it private. Cannot be final because we need empty constructor for json de-serialization, until we eliminate segment generation config being read from json.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -0,0 +1,103 @@
...
+  public boolean prune(IndexSegment segment, BrokerRequest brokerRequest) {
+    FilterQueryTree filterQueryTree = RequestUtils.generateFilterQueryTree(brokerRequest);
+    if (filterQueryTree == null) {
+      return false;
+    }
+
+    // For realtime segment, this map can be null.
+    Map<String, ColumnMetadata> columnMetadataMap =
+        ((SegmentMetadataImpl) segment.getSegmentMetadata()).getColumnMetadataMap();
```
mayankshriv: Because SegmentMetadata is in common and ColumnMetadata is in core.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
mayankshriv: That seems accidental, thanks for catching, will fix.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
mayankshriv: Ok, figured out why this happened. Since the @Override annotation is above the javadoc, auto formatting in Intellij did that. Fixed by moving the annotation below javadoc.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/stats/AbstractColumnStatisticsCollector.java
mayankshriv: Yes, the log is a few lines below. This return is the happy path when value is in range, ie no action needed. When value is not in range, we clear the partition function from stats collector, so it does not get written to metadata.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ModuloPartitionFunction.java
```diff
@@ -0,0 +1,59 @@
...
+   * Constructor for the class.
+   * <ul>
+   *   <li> Requires the divisor to be passed in via args. </li>
+   * </ul>
+   *
+   * @param args Arguments for the function
+   */
+  public ModuloPartitionFunction(String[] args) {
+    Preconditions.checkArgument(args.length == 1);
+    _args = args;
```
mayankshriv: Will update.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/PartitionFunctionFactory.java
mayankshriv: Covered under SegmentPartitionTest, and we only have one partition function defined currently.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
mayankshriv: This is the same syntax as RangePredicates. We should extract out the logic to parse ranges that should handle spaces etc and use it here as well as RangePredicate. Will address that in separate PR.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
mayankshriv: Added nonnull for partitionValues. Adding a separate check is flagged as redundant in IDE.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentPartitionConfig.java
mayankshriv: Made private. Cannot be final because empty constructor required for json de-serialization

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/AbstractSegmentPruner.java
mayankshriv: Currently we only have two operators. When we add a new operator (for example XOR), this exception will help us identify this location and take necessary action (eg write code for XOR based pruning).

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/StatsCollectorConfig.java
mayankshriv: done

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/StatsCollectorConfig.java
mayankshriv: IDE flags check as warning due to annotation.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/stats/AbstractColumnStatisticsCollector.java
mayankshriv: Had a discussion with @mcvsubbu on the same topic. The current consensus was for stats collector to provided an api to indicate if partitioning passed/failed. And if failed, the caller (segment generation) can decide what needs to be done.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
mayankshriv: Checkout SegmentPartitionTest.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
antumbde: It can be if use following C'tor signature. That way you need only one. 
(good to have; not absolutely mandatory)

public ColumnPartitionConfig(@Nonnull @JsonProperty("functionName") String functionName, @JsonProperty("partitionValues") String partitionValues)

pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
mayankshriv: Positive/negative tests already present. Added tests for AND/OR.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
mayankshriv: Added

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
mayankshriv: Added

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/ColumnMetadata.java
mayankshriv: I agree that it is more likely that partitioning is performed on a composite key, but supporting that would require a lot of changes to existing code. Until then I'd lean towards, allowing partitioning on multiple columns, as it has added benefit of pruning when any of the partitioned columns has predicates (as opposed to all of the columns in composite partitioning).

antumbde:
Make sure CI passes before pushing

PR #1183, opened 2017-03-22
Calculate offset for baselines using Period instead of millis, so that DST is handled correctly

npawar:
This fixes the wow, wo2w, wo3w chart on Investigate page of anomalies, which was showing 0s for daily anomalies due to this issue
This also fixes how scheduler calculates its new buckets

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobSchedulerUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobSchedulerUtils.java
@@ -7,6 +7,7 @@ import java.util.concurrent.TimeUnit;
 
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+import org.joda.time.Period;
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobSchedulerUtils.java
```diff
@@ -117,14 +118,28 @@ public static long getBucketSizeInMSForDataset(DatasetConfigDTO datasetConfig, A
...
       }
     } else {
-      bucketMillis = TimeUnit.MILLISECONDS.convert(1, dataUnit);
+      bucketSizePeriod = getBucketSizePeriodForUnit(dataUnit);
     }
-    return bucketMillis;
+    return bucketSizePeriod;
+  }
+
+  private static Period getBucketSizePeriodForUnit(TimeUnit unit) {
```
cyenjung: I suggest that we add a method (e.g., toPeriod) in TimeGranularity.java that returns the period according to time unit and size of the time granularity. We can also take care of lines 119-127 in toPeriod method.

We also need to consider the case of MINUTES.

PR #1192, opened 2017-03-23
Adding regex predicate support

kishoreg:
Adding support for the REGEXP_LIKE operator. 

REGEXP_LIKE operator supports java style regular expressions. All comparisons are case insensitive by default. One can turn on/off flags by using [Embedded Flag Expressions](https://docs.oracle.com/javase/tutorial/essential/regex/pattern.html) 

e.g queries
```
select l_comment from tpch_lineitem where regexp_like (l_comment, '.*fox.*')
```

Note the initial commit supported MySQL  LIKE semantics but it was getting complicated to convert a LIKE expression to java regex expression and hence dropped it in the subsequent commit. Note the pattern does not have to match the entire string. If users want to match the entire string, they can use ^ at the start and $ at the end of the expression.


pinot-common/src/main/antlr4/com/linkedin/pinot/pql/parsers/PQL2.g4
```diff
@@ -87,6 +88,9 @@ comparisonOperator: '<' | '>' | '<>' | '<=' | '>=' | '=';
 betweenClause:
   expression BETWEEN expression AND expression;
 
+regexClause:
+  expression (LIKE | REGEXP ) '(' literal (',' literal)* ')';
```
jfim: It would be good to not require parentheses for this. For example, in pretty much every SQL database, the syntax `foo LIKE 'a%'` is valid.

This also copies the (non-standard) `REGEXP` keyword from MySQL. Unfortunately, that extension doesn't support adding regexp flags (eg. case insensitive matching, multiline match, etc.). I took a quick look at how other databases do it:

- MySQL: `foo REGEXP 'a.*'`, no support for regular expression flags
- PostgreSQL: `foo ~ 'a.*'` for case sensitive match, `foo ~* 'a.*'` for case insentitive match, 
- MSSQL: Not supported
- Oracle: `REGEXP_LIKE(foo, 'a.*', 'i')` with flag support (eg. `'i'` for case insensitive match)
- H2: Same as Oracle
- SQLite: Same as MySQL

Even though I normally like how PostgreSQL does things, in this case I think it's a bit weird and prefer the Oracle syntax (implement it as a function, with arguments, which also allows regexp flags).

Alternatively, if we use the MySQL (or PostgreSQL) syntax, we could encode the flags within the expression itself, just like Java does (eg. `foo(?im)` is a case-insensitive multiline match). Let's figure out which one we want, because once it's in the grammar we'll be stuck with it for a long time.

pinot-controller/src/main/resources/webapp/js/init.js
```diff
@@ -98,6 +98,7 @@ var HELPERS = {
 
   executeQuery: function(query, traceEnabled, callback) {
     var url = "/pql?pql=" + query +"&trace=" + traceEnabled;
+    url = encodeURI(url);
```
jfim: Does that actually work? I would think you'd need to encode `query` instead

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/RegexPredicateAstNode.java
```diff
@@ -0,0 +1,83 @@
...
+
+    for (AstNode astNode : getChildren()) {
+      if (astNode instanceof LiteralAstNode) {
+        LiteralAstNode node = (LiteralAstNode) astNode;
+        String expr = node.getValueAsString();
+        //convert LIKE semantics to java semantics, this is not fully compatible with SQL semantics
+        if (matchType == MatchType.LIKE) {
+          expr = expr.replace(".", "\\.");
+          expr = expr.replace("_", ".");
+          expr = expr.replaceAll("%", ".*?");
```
jfim: You also need to escape regular expression atoms, eg. `foo like 'abc (%)'` won't expand correctly.

jfim:
For the integration test, you can use `REGEXP_LIKE` in H2 (http://www.h2database.com/html/functions.html#regexp_like) as part of the query comparison/generation. I can help you with that part.

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/RegexPredicateAstNode.java
kishoreg: you mean escape ( [ ^ etc?.  I am wondering if we should just quote all segments before and after characters % and _?

pinot-controller/src/main/resources/webapp/js/init.js
kishoreg: It seemed to work but I did not test the traceEnabled part. But good point, I will encode the query part.

jfim: Oh, interesting, `encodeURI` doesn't actually encode special characters like & and ?. In your case, it worked fine, but it means that if there is a query that contains a question mark, the question mark won't get encoded properly.

What you want in this case is `encodeURIComponent(query)` which will properly escape the query contents:

```
encodeURI('foo=bar?baz%')
"foo=bar?baz%25"
encodeURIComponent('foo=bar?baz%')
"foo%3Dbar%3Fbaz%25"
```

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/RegexPredicateAstNode.java
jfim: Yeah, escape the ( [ and such atoms.

kishoreg:
I had trouble getting the same results as H2 in our integration test. REGEXP_LIKE in H2 uses pattern.compile.find() method instead of pattern.compile.matches(). I feel matches is the right thing to do but for now I have used find() to match with H2. Its not clear whats the behavior in other DB's.

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/RegexPredicateAstNode.java
```diff
@@ -0,0 +1,73 @@
...
+  public RegexPredicateAstNode() {
+  }
+
+  @Override
+  public FilterQueryTree buildFilterQueryTree() {
+    if (_identifier == null) {
+      throw new Pql2CompilationException("IN predicate has no identifier");
+    }
+
+    TreeSet<String> values = new TreeSet<>();
```
Jackie-Jiang: Any specific reason why we use TreeSet instead of HashSet?

pinot-core/src/main/java/com/linkedin/pinot/core/common/predicate/RegexPredicate.java
```diff
@@ -18,19 +18,26 @@
...
 import java.util.List;
 
+import com.google.common.base.Preconditions;
 import com.linkedin.pinot.core.common.Predicate;
-import com.linkedin.pinot.core.common.Predicate.Type;
 
 
 public class RegexPredicate extends Predicate {
-
+  String regex;
```
Jackie-Jiang: We have decided to use the convention _regex for member variable.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/predicate/RegexPredicateEvaluator.java
```diff
@@ -0,0 +1,82 @@
...
+  public int[] getNonMatchingDictionaryIds() {
+    throw new UnsupportedOperationException("Returning non matching values is expensive for predicateType:" + predicate.getType());
+  }
+
+  @Override
+  public boolean apply(int[] dictionaryIds, int length) {
+    for (int i = 0; i < length; i++) {
+      int dictId = dictionaryIds[i];
+      String value = dictionary.getStringValue(dictId);
+      if (pattern.matcher(value).matches()) {
```
Jackie-Jiang: matches() or find()?

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/OfflineClusterIntegrationTest.java
```diff
@@ -51,6 +51,7 @@
 /**
  * Integration test that converts Avro data for 12 segments and runs queries against it.
  */
+@Test
```
Jackie-Jiang: We don't need this annotation to mark it as a test class

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/QueryGenerator.java
```diff
@@ -799,7 +799,44 @@ public QueryFragment generatePredicate(String columnName) {
...
+            sqlRegexBuilder.append(".*");
+          } else {
+            pqlRegexBuilder.append(value.charAt(i));
+            sqlRegexBuilder.append(value.charAt(i));
+          }
+        }
+
+        String pql = String.format(" REGEXP_LIKE(%s, '%s')", columnName, pqlRegexBuilder.toString());
+        String sql = String.format(" REGEXP_LIKE(%s, '%s', 'i')", columnName, sqlRegexBuilder.toString());
+        System.out.println("Value:" + value + " pql:" + pql + " sql:" + sql);
```
Jackie-Jiang: Do not do sout because it might polute the test log and break Travis-CI for the number of output lines limit.

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/RegexPredicateAstNode.java
kishoreg: not really. changing it to hashset

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/predicate/RegexPredicateEvaluator.java
kishoreg: it was changed to find. Not sure why the PR is not reflecting it. will update the PR again.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/OfflineClusterIntegrationTest.java
kishoreg: for some reason, eclipse needs that.

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/RegexPredicateAstNode.java
```diff
@@ -0,0 +1,72 @@
...
+
+import com.linkedin.pinot.common.request.FilterOperator;
+import com.linkedin.pinot.common.utils.StringUtil;
+import com.linkedin.pinot.common.utils.request.FilterQueryTree;
+import com.linkedin.pinot.pql.parsers.Pql2CompilationException;
+
+public class RegexPredicateAstNode extends PredicateAstNode {
+  private String _identifier;
+
+  public RegexPredicateAstNode() {
```
mayankshriv: Is this needed?

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/RegexPredicateAstNode.java
```diff
@@ -0,0 +1,72 @@
...
+        values.add(expr);
+      }
+    }
+    if(values.size() > 1) {
+      throw new Pql2CompilationException("Matching more than one regex is NOT supported currently");
+    }
+
+    String[] valueArray = values.toArray(new String[values.size()]);
+    FilterOperator filterOperator = FilterOperator.REGEX;
+    List<String> value = Collections.singletonList(StringUtil.join("\t\t", valueArray));
```
mayankshriv: static final for "\t\t"

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/utils/PQLParserUtils.java
```diff
@@ -124,4 +124,25 @@ public static void decorateWithMapReduce(JSONObject jsonObj,
       throw new AssertionError("Should not reach this");
     }
   }
+
+  /**
+   * Converts MySQL LIKE expression to java Regex
+   * @param expr
+   * @return
+   */
+  public static String convertLikeExpressionToJavaRegex(String expr) {
```
mayankshriv: Does this have unit test?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/predicate/RegexPredicateEvaluator.java
```diff
@@ -0,0 +1,82 @@
...
+import com.linkedin.pinot.core.common.predicate.RegexPredicate;
+import com.linkedin.pinot.core.segment.index.readers.Dictionary;
+
+public class RegexPredicateEvaluator implements PredicateEvaluator {
+
+  private RegexPredicate predicate;
+  private Dictionary dictionary;
+  private Pattern pattern;
+
+  public RegexPredicateEvaluator(RegexPredicate predicate, Dictionary dictionary) {
```
mayankshriv: Just a heads up, there's another PR in-flight which requires no-dictionary support for predicate evaluators.

kishoreg: Good point. Should be easy for regex since we don't have to differentiate based on datatype

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/utils/PQLParserUtils.java
kishoreg: I will delete this method as I removed the support for LIKE syntax

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/RegexPredicateAstNode.java
kishoreg: removed

kishoreg:
I realized that we had REGEX enum already defined in request thrift file. Since no one has used this yet, I thought this was the right time to fix that as well. I changed REGEX to REGEXP_LIKE and renamed all predicate classed to reflect the same.

PR #1196, opened 2017-03-24
Add partition function to mimic Kafka's default partitioner.

mayankshriv:
1. Added partition function that mimics Kafka's default partitioner.
This will be used to validate parittioning in LLC. For validated
segments, partitioning info will be written out in the metadata,
partition based pruner can pick it up during query execution.

2. Added unit test for the same.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/PartitionFunctionFactory.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/PartitionFunctionFactory.java
@@ -16,6 +16,8 @@
 package com.linkedin.pinot.core.data.partition;
 
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 import javax.annotation.Nonnull;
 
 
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/PartitionFunctionFactory.java
```diff
@@ -27,7 +27,8 @@
 
   // Enum for various partition functions to be added.
   public enum PartitionFunctionType {
-    MODULO
+    modulo,
+    defaultkafkapartitioner
```
rven1: Camel Case?

pinot-core/src/test/java/com/linkedin/pinot/core/data/partition/PartitionFunctionTest.java
```diff
@@ -43,10 +43,42 @@ public void testModulo() {
...
+   *   <li> Tests that toString returns expected string. </li>
+   * </ul>
+   */
+  @Test
+  public void testDefaultKafkaPartitioner() {
+    long seed = System.currentTimeMillis();
+    Random random = new Random(seed);
+
+    for (int i = 0; i < 1000; i++) {
+      int divisor = Math.abs(random.nextInt());
```
rven1: Returns a pseudorandom, uniformly distributed int value between 0 (inclusive) and the specified value (exclusive), drawn from this random number generator's sequence.
Need to handle divide by 0 or exclude 0

mayankshriv: Fixed.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/PartitionFunctionFactory.java
mayankshriv: I wanted to support case insensitive naming, ie any casing of name can be specified in config. Matching enum with any kind of casing on string is easier with this approach.

PR #1200, opened 2017-03-24
[PINOT-4475] Handle SimpleDateFormat correctly when creating segments

mcvsubbu:
If time column is of SimpleDateFormat type, the we need to set the start and end time
correctly in the segment.

Input files to pinot segment creator (for pure offline use cases) may contain time column values
like 20170324 to indicate Mar 24 2017. If we take this value to be from epoch, we will get year 57194.

Modified SegmentGeneratorConfig to take additional configuration for time type. The default
is epoch, so that the default behavior remains unchanged. Setting the time format correctly
will now re-adjust the start/end time of the segment if needed, so that the start and end time
in the segment metadata is reflected correctly.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
@@ -52,6 +52,10 @@ import com.linkedin.pinot.core.startree.hll.HllConfig;
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class SegmentGeneratorConfig {
+  public enum TimeColumnType {
+    EPOCH,
+    SIMPLE_DATE
+  }
   private static final Logger LOGGER = LoggerFactory.getLogger(SegmentGeneratorConfig.class);
 
   private Map<String, String> _customProperties = new HashMap<>();
@@ -84,6 +88,8 @@ public class SegmentGeneratorConfig {
   private SegmentNameGenerator _segmentNameGenerator = null;
   private SegmentPartitionConfig _segmentPartitionConfig = null;
   private int _sequenceId = -1;
+  private TimeColumnType _timeColumnType = TimeColumnType.EPOCH;
+  private String _simpleDateFormat = null;
 
   public SegmentGeneratorConfig() {
   }
@@ -140,6 +146,19 @@ public class SegmentGeneratorConfig {
   public void setCustomProperties(Map<String, String> properties) {
     Preconditions.checkNotNull(properties);
     _customProperties.putAll(properties);
+  }
+
+  public void setSimpleDateFormat(String simpleDateFormat) {
+    _timeColumnType = TimeColumnType.SIMPLE_DATE;
+    _simpleDateFormat = simpleDateFormat;
+  }
+
+  public String getSimpleDateFormat() {
+    return _simpleDateFormat;
+  }
+
+  public TimeColumnType getTimeColumnType() {
+    return _timeColumnType;
   }
 
   public boolean containsCustomProperty(String key) {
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
```diff
@@ -455,6 +459,36 @@ private void handlePostCreation()
     LOGGER.info("Driver, indexing time : {}", totalIndexTime);
   }
 
+  private void updateSegmentStartEndTimeIfNecessary(ColumnStatistics timeColumnStats) {
+    switch (config.getTimeColumnType()) {
+      case EPOCH:
+        break;
```
antumbde: shouldn't we set start/end time for epoch also if the user didn't provide ?

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
```diff
@@ -142,6 +148,19 @@ public void setCustomProperties(Map<String, String> properties) {
     _customProperties.putAll(properties);
   }
 
+  public void setSimpleDateFormat(String simpleDateFormat) {
+    _timeColumnType = TimeColumnType.SIMPLE_DATE;
+    _simpleDateFormat = simpleDateFormat;
+  }
+
+  public String getSimpleDateFormat() {
```
antumbde: @Nullable for return annotation

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
```diff
@@ -424,6 +427,7 @@ private void handlePostCreation()
       throws Exception {
     final String timeColumn = config.getTimeColumnName();
     segmentName = config.getSegmentNameGenerator().getSegmentName(segmentStats.getColumnProfileFor(timeColumn));
+    updateSegmentStartEndTimeIfNecessary(segmentStats.getColumnProfileFor(timeColumn));
```
antumbde: timeColumn can be null or empty. Code needs to handle that

We check for != null in this file but reading code it timeColumn will not be null but empty if not specified

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
```diff
@@ -455,6 +459,36 @@ private void handlePostCreation()
...
+    return convertSDFToMillis(minTimeStr);
+  }
+
+  public long convertEndTimeSDFToMillis(ColumnStatistics timeColumnStats) {
+    final String maxTimeStr = timeColumnStats.getMaxValue().toString();
+    return convertSDFToMillis(maxTimeStr);
+  }
+
+  private long convertSDFToMillis(final String colValue) {
+    final String sdfFormatStr = config.getSimpleDateFormat();
```
antumbde: protect here. format can be null (bad coding) and colValue can be empty if there is no time column which will trigger exception

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
mcvsubbu: That is automatically handled in the code. Time column name is returned as "" if time column is null. See RawIndexCreatorTest

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
mcvsubbu: Epoch start/end time is set by default. We are overwriting things if the time format is simpledate

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
mcvsubbu: I will add the protection in setTimeFormat() method of SegmentGeneratorConfig

mcvsubbu:
@jfim adding a unit test

PR #1202, opened 2017-03-25
Reuse in-memory dictionary during stats gathering

jfim:
Reuse the in-memory data structures when writing realtime segment to
disk instead of iterating over the in-memory segment row-by-row to
build the column dictionaries and determine the per-column statistics.
Almost doubles the performance of realtime segment write.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
@@ -1,0 +1,156 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.linkedin.pinot.core.realtime.converter.stats;
+
+import com.linkedin.pinot.common.data.FieldSpec;
+import com.linkedin.pinot.core.common.Block;
+import com.linkedin.pinot.core.common.BlockMultiValIterator;
+import com.linkedin.pinot.core.data.partition.PartitionFunction;
+import com.linkedin.pinot.core.io.reader.SingleColumnSingleValueReader;
+import com.linkedin.pinot.core.operator.blocks.RealtimeSingleValueBlock;
+import com.linkedin.pinot.core.realtime.impl.datasource.RealtimeColumnDataSource;
+import com.linkedin.pinot.core.realtime.impl.dictionary.MutableDictionaryReader;
+import com.linkedin.pinot.core.segment.creator.ColumnStatistics;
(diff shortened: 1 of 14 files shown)


jfim:
Did some quick testing with the realtime cluster integration test, the time used to convert the segments during the tests:

```
Without this optimization
conversionTimeMillis = 2791
conversionTimeMillis = 2134
conversionTimeMillis = 1994
conversionTimeMillis = 2004
conversionTimeMillis = 1941

With this optimization
conversionTimeMillis = 1845
conversionTimeMillis = 1462
conversionTimeMillis = 1263
conversionTimeMillis = 1255
conversionTimeMillis = 1246
```

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -0,0 +1,156 @@
...
+    // If this is a single value, then by definition the data is sorted
+    final int blockLength = _block.getMetadata().getLength();
+    if (blockLength <= 1 || getCardinality() <= 1) {
+      return true;
+    }
+
+    // Iterate over all data to figure out whether or not it's in sorted order
+    SingleColumnSingleValueReader singleValueReader = ((RealtimeSingleValueBlock) _block).getReader();
+
+    int docIdIndex = _sortedDocIdIterationOrder != null ? _sortedDocIdIterationOrder[0] : 0;
```
mcvsubbu: will be useful to factor this logic into a method.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -0,0 +1,156 @@
...
+
+  @Override
+  public int getCardinality() {
+    return _dictionaryReader.length();
+  }
+
+  @Override
+  public int getLengthOfLargestElement() {
+    // Length of longest string
+    int maximumStringLength = 0;
```
mcvsubbu: should this not be set to the length of the first element by default?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -0,0 +1,156 @@
...
+
+    // If this column is a string column, iterate over the dictionary to find the maximum length
+    if (_dataSource.getDataSourceMetadata().getDataType() == FieldSpec.DataType.STRING) {
+      final int length = _dictionaryReader.length();
+      for (int i = 0; i < length; i++) {
```
mcvsubbu: we should already have this information as we input the rows, right?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
jfim: It's equivalent and less verbose to do it this way. If I try to set it to the length of the first element, I need to add a check for the dictionary being empty.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
jfim: The String dictionary does not keep that information.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
jfim: If it's extracted into a method, then the optimizer can't notice that this check is actually a loop invariant without inlining that method.

PR #1204, opened 2017-03-27
Updating table configuration should ensure that realtime cluster is setup

antumbde:
PUT call to update table configuration should ensure that realtime cluster
is setup to maintain backwards compatibility.


antumbde:
Tested for backward compatibility with realtime cluster - add llc, remove hlc, add back hlc

PR #1207, opened 2017-03-27
Unpack znodes from client API if necessary

jfim:
Unpack compressed znodes when reading the external view from the
client API. This allows clients with large clusters to be able to read
the brokerResource external view correctly when serving several
hundreds of tables.

### DIFF
--- a/pinot-api/src/main/java/com/linkedin/pinot/client/ExternalViewReader.java
+++ b/pinot-api/src/main/java/com/linkedin/pinot/client/ExternalViewReader.java
@@ -15,6 +15,9 @@
  */
 package com.linkedin.pinot.client;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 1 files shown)


pinot-api/src/main/java/com/linkedin/pinot/client/ExternalViewReader.java
```diff
@@ -114,4 +120,27 @@ public ExternalViewReader(ZkClient zkClient) {
     return tableToBrokersMap;
   }
 
+  private static byte[] unpackZnodeIfNecessary(byte[] znodeContents) {
```
kishoreg: This is a temporary workaround right, we should be using Helix API to get the EV - that would have handled compression automatically.

jfim: The thing is that if we do that, it pulls in all of Helix into the Pinot client.

PR #1209, opened 2017-03-28
Add classifier config

cyenjung:
Tested with unit test


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/bao/jdbc/ClassificationConfigManagerImpl.java
```diff
@@ -0,0 +1,43 @@
...
+      return configs;
+    } else {
+      return Collections.emptyList();
+    }
+  }
+
+  @Override
+  public ClassificationConfigDTO findByName(String name) {
+    Map<String, Object> filters = new HashMap<>();
+    filters.put("name", name);
```
npawar: Use Predicate instead of filters map. Check DetectionStatusIndex for example.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/ClassificationConfigBean.java
```diff
@@ -0,0 +1,42 @@
+package com.linkedin.thirdeye.datalayer.pojo;
+
+import java.util.List;
+
+public class ClassificationConfigBean extends AbstractBean {
```
npawar: Can we have some java doc explaining this new config and these fields? mainFunctionId isn't too self explanatory.

PR #1211, opened 2017-03-28
Separate IndexLoadingConfig from TableDataManagerConfig

Jackie-Jiang:
This PR is a prerequisite of RELOAD message.
When loading a segment, we should pick up the latest table config.
Old behavior: when we create the TableDataManager, we set the IndexLoadingConfig and it becomes immutable, which causes new segment unable to pick up the latest table config.
New behavior: create a new IndexLoadingConfig per segment loading so that segment loading config is in sync with latest table config.

Fixed a bug where SegmentMetadata.getIndexDir() return inconsistent result for v3 format when loading with different methods.
Added integration test to test triggering inverted index creation.


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/offline/TableDataManagerProvider.java
```diff
@@ -15,54 +15,49 @@
...
 public class TableDataManagerProvider {
-  private static ServerMetrics SERVER_METRICS;
-
-  private static Map<String, Class<? extends TableDataManager>> keyToFunction =
-      new ConcurrentHashMap<String, Class<? extends TableDataManager>>();
-
-  static {
-    keyToFunction.put("offline", OfflineTableDataManager.class);
-    keyToFunction.put("realtime", RealtimeTableDataManager.class);
+  private TableDataManagerProvider() {
```
mcvsubbu: if all methods are static, please make this a singleton class and have one static method to get the instance.

Jackie-Jiang: This is a util class, which is not allowed to instantiate (The method is defined private to prevent instantiate).

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/IndexLoadingConfig.java
```diff
@@ -0,0 +1,195 @@
...
+      if (invertedIndexColumns != null) {
+        _invertedIndexColumns.addAll(invertedIndexColumns);
+      }
+
+      List<String> noDictionaryColumns = indexingConfig.getNoDictionaryColumns();
+      if (noDictionaryColumns != null) {
+        _noDictionaryColumns.addAll(noDictionaryColumns);
+      }
+
+      String tableSegmentVersion = indexingConfig.getSegmentFormatVersion();
```
mcvsubbu: see other PR

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/Loaders.java
```diff
@@ -28,123 +26,101 @@
...
 
+
 public class Loaders {
   private static final Logger LOGGER = LoggerFactory.getLogger(Loaders.class);
 
   public static class IndexSegment {
-    public static com.linkedin.pinot.core.indexsegment.IndexSegment load(File indexDir, ReadMode readMode)
+
+    /**
+     * For tests only.
```
mcvsubbu: move this method to a test class.,

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/offline/TableDataManagerProvider.java
```diff
@@ -15,54 +15,49 @@
...
+  private static ServerMetrics _serverMetrics;
+
+  public static TableDataManager getTableDataManager(@Nonnull TableDataManagerConfig tableDataManagerConfig,
+      @Nullable String serverInstance) {
+    Preconditions.checkNotNull(_serverMetrics);
+
+    String tableDataManagerType = tableDataManagerConfig.getTableDataManagerType().toLowerCase();
+    TableDataManager tableDataManager;
+    switch (tableDataManagerType) {
+      case "offline":
```
mayankshriv: Can we have enum/constant for tableDataManagerType?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
```diff
@@ -121,15 +123,18 @@ public HLRealtimeSegmentDataManager(final RealtimeSegmentZKMetadata segmentMetad
         this.sortedColumn = null;
       }
     }
-    //inverted index columns
-    invertedIndexColumns = indexingConfig.getInvertedIndexColumns();
-    if (sortedColumn != null && !invertedIndexColumns.contains(sortedColumn)) {
+
+    // Inverted index columns
+    Set<String> invertedIndexColumns = indexLoadingConfig.getInvertedIndexColumns();
+    if (sortedColumn != null) {
```
mayankshriv: I know this is not part of your PR, but not sure if I understand the motivation of adding inv index when sortedColumn != null. Could you add an item to investigate this further (outside of this PR)?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/config/TableDataManagerConfig.java
```diff
@@ -85,94 +65,21 @@ public static TableDataManagerConfig getDefaultHelixTableDataManagerConfig(
...
-      LOGGER.info("Loading table: {} with server configured segment format version: {}", tableName,
-          serverConfigVersion);
-    }
+    return new TableDataManagerConfig(defaultConfig);
   }
 
-  public IndexLoadingConfigMetadata getIndexLoadingConfigMetadata() {
-    return new IndexLoadingConfigMetadata(_tableDataManagerConfig);
+  public void overrideConfigs(@Nonnull AbstractTableConfig tableConfig) {
+    // Override table level configs
```
Jackie-Jiang: Added comments

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
Jackie-Jiang: Find the reason. The motivation is that, when we seal the realtime in memory segment into offline segment, we use sortedColumn's inverted index to maintain the order of records. Will add comments.

kishoreg:
Will this cause additional zk read for every segment load?. This might
impact the startup time rt?

Jackie-Jiang:
@kishoreg No, there will be no additional zk read because we always read table config but do not use it...

PR #1218, opened 2017-03-28
Control the creation of shaded jars using a profile called build-shaded-jar. 

rven1:


rven1:
Control the creation of shaded jars using a profile called build-shaded-jar. Only build shaded jars for pinot-server, pinot-broker, pinot-controller and pinot-hadoop. Build time reduced from 2:50 min to 1:04 for building without shaded jars, and 1:20 for building only the specified shaded jars with the profile build-shaded-jar.

PR #1230, opened 2017-04-01
Add support for partitioners in LLC realtime.

rven1:


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -525,13 +531,12 @@ private boolean buildSegmentInternal(boolean forCommit) {
     // Build a segment from in-memory rows.If buildTgz is true, then build the tar.gz file as well
     // TODO Use an auto-closeable object to delete temp resources.
     File tempSegmentFolder = new File(_resourceTmpDir, "tmp-" + _segmentNameStr + "-" + String.valueOf(now()));
-
+    int nPartitions = _consumerWrapper.getPartitionCount(_kafkaTopic, /*maxWaitTimeMs=*/5000L);
```
mcvsubbu: Not clear what happens if there is a timeout. 

Also, the use cases that do not care to get the number of partitions will experience the delay as well. We have a  budget for segment building now. I would like to see if we can restrict this call to Kafka for use cases that absolutely need it.

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
```diff
@@ -120,6 +121,14 @@ public void setNoDictionaryColumns(List<String> noDictionaryColumns) {
     _noDictionaryColumns = noDictionaryColumns;
   }
 
+  public Map<String, String> getPartitioners() {
```
mcvsubbu: Nullable

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
```diff
@@ -120,6 +121,14 @@ public void setNoDictionaryColumns(List<String> noDictionaryColumns) {
...
+    return _partitioners;
+  }
+
+  public void setPartitioners(Map<String, String> partitioners) {
```
mcvsubbu: Nullable

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
```diff
@@ -39,6 +39,7 @@
   private String _starTreeFormat;
   private String _columnMinMaxValueGeneratorMode;
   private List<String> _noDictionaryColumns;
+  private Map<String, String> _partitioners;
```
mcvsubbu: Nullable

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -840,6 +845,13 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
     // Create field extractor
     _fieldExtractor = FieldExtractorFactory.getPlainFieldExtractor(schema);
     makeConsumerWrapper();
+
+    // Partitioners
+    _partitioners = indexingConfig.getPartitioners();
+    _nPartitions = _consumerWrapper.getPartitionCount(_kafkaTopic, /*maxWaitTimeMs=*/5000L);
```
mcvsubbu: Why are we getting tis in 2 places?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/MutableDictionaryReader.java
```diff
@@ -186,4 +202,63 @@ public void print() {
...
+    if (partitionFunction != null) {
+       int partition = partitionFunction.getPartition(value);
+
+      for (IntRange partitionValue : partitionValues) {
+        if (partition >= partitionValue.getMinimumInteger() && partition <= partitionValue.getMaximumInteger()) {
+          return;
+        }
+      }
+
+      LOGGER.warn("Column '{}' not partitioned as specified for value: '{}', dropping partition metadata.", column,
```
mcvsubbu: if one value falls out of partition, will we be printing this message for every row thereafter? 
Also, wil we be printing this message for use cases that do not have a partition funciton?
Isnt it better to add a null chekc up front and return right away? This code is in the consumption path, right?

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/LLCRealtimeClusterIntegrationTest.java
```diff
@@ -36,18 +38,22 @@
...
 public class LLCRealtimeClusterIntegrationTest extends RealtimeClusterIntegrationTest {
   private static int KAFKA_PARTITION_COUNT = 2;
+  private static final String KAFKA_PARTITIONING_KEY = "AirlineID";
 
   protected void setUpTable(String tableName, String timeColumnName, String timeColumnType, String kafkaZkUrl,
       String kafkaTopic, File schemaFile, File avroFile) throws Exception {
     Schema schema = Schema.fromFile(schemaFile);
     addSchema(schemaFile, schema.getSchemaName());
+    Map<String, String> map = new HashMap<String, String>();
+    map.put(KAFKA_PARTITIONING_KEY, PartitionFunctionFactory.PartitionFunctionType.ByteArray.toString());
```
mcvsubbu: With this change, we are always testing with partition key?
Can we please add a separate test for testing partition key so that we still have other tests that do not use partition key but still use llc?

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
mayankshriv: Add comment to describe the key and value for this map.

pinot-common/src/main/java/com/linkedin/pinot/common/request/helper/ControllerRequestBuilder.java
```diff
@@ -122,7 +124,18 @@ public static JSONObject buildCreateRealtimeTableJSON(String tableName, String s
         String timeColumnName, String timeType, String retentionTimeUnit, String retentionTimeValue, int numReplicas,
     String segmentAssignmentStrategy, JSONObject streamConfigs, String schemaName, String sortedColumn,
         List<String> invertedIndexColumns, String loadMode, boolean isHighLevel, List<String> noDictionaryColumns)
-    throws JSONException {
+      throws JSONException {
```
mayankshriv: Why does the partitioners argument need to be empty map as opposed to null?

pinot-common/src/main/java/com/linkedin/pinot/common/request/helper/ControllerRequestBuilder.java
```diff
@@ -157,6 +170,9 @@ public static JSONObject buildCreateRealtimeTableJSON(String tableName, String s
     if (noDictionaryColumns != null) {
       tableIndexConfig.put("noDictionaryColumns", noDictionaryColumns);
     }
+    if (partitioners != null) {
```
mayankshriv: Since we are passing empty map in one of the calls, it is going to write out an empty config? I'd rather have nothing written at all in that case.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mayankshriv: To add, is this additional call actually required? I thought LLC segment names have partition number encoded in the name, which would imply that we already queried the partition once. Can we not use the same value? Or does that happen in the controller?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mayankshriv: +1

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -16,40 +16,39 @@
...
  * Implementation of {@link ModuloPartitionFunction} that mimics Kafka's
  * {@link org.apache.kafka.clients.producer.internals.DefaultPartitioner}.
  *
  */
-public class DefaultKafkaPartitionFunction implements PartitionFunction {
-  private static final String NAME = "DefaultKafkaPartitioner";
+public class ByteArrayPartitionFunction implements PartitionFunction {
+  private static final String NAME = "ByteArray";
   private final int _divisor;
+  public ByteArrayPartitioner defaultPartitioner;
```
mayankshriv: Change the variable name to match the type name.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -16,40 +16,39 @@
...
   @Override
-  public int getPartition(Object value) {
-    if (value instanceof String) {
-      return (Utils.murmur2(((String) value).getBytes()) & 0x7fffffff) % _divisor;
-    } else {
-      throw new IllegalArgumentException(
-          "Illegal argument for partitioning, expected Integer, got: " + value.getClass().getName());
-    }
+  public int getPartition(Object valueIn) {
+    byte[] value = valueIn.toString().getBytes();
```
mayankshriv: this is not used in the function?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: I have added a try catch and also made it conditional on having a partitioner

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -16,40 +16,39 @@
 package com.linkedin.pinot.core.data.partition;
 
 import com.google.common.base.Preconditions;
-import org.apache.kafka.common.utils.Utils;
+import kafka.producer.ByteArrayPartitioner;
 
 
 /**
  * Implementation of {@link ModuloPartitionFunction} that mimics Kafka's
```
mayankshriv: Fix the typo in previous checkin, it is implementation of ParititionFunction, not ModuloPartitionFunction

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/MurmurPartitionFunction.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Implementation of {@link ModuloPartitionFunction} that mimics Kafka's
+ * {@link org.apache.kafka.clients.producer.internals.DefaultPartitioner}.
+ * However, this is not general purpose. It assumes that the partition key
+ * was a string as used by the Tracking Producer. This is only a reference
+ * implementation and users should create their own for their specific partitioner.
+ *
+ */
+public class MurmurPartitionFunction implements PartitionFunction {
```
mayankshriv: Since you added ByteArrayPartitionFunction, do we really need this? Also, is ByteArrayPartitionFunction internally doing the same math as this one?

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/AbstractSegmentPruner.java
```diff
@@ -31,7 +32,7 @@
  */
 public abstract class AbstractSegmentPruner implements SegmentPruner {
 
-  public abstract boolean pruneSegment(FilterQueryTree filterQueryTree, Map<String, ColumnMetadata> columnMetadataMap);
+  public abstract boolean pruneSegment(IndexSegment segment, FilterQueryTree filterQueryTree, Map<String, ColumnMetadata> columnMetadataMap);
```
mayankshriv: I think the lower level functions only need partition info. So instead of passing index segment and columnMetadaMap all over, may be it would be better to extract out that information upfront, and just pass that around.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -79,18 +90,30 @@ public boolean pruneSegment(FilterQueryTree filterQueryTree, Map<String, ColumnM
 
     // Leaf node
     String column = filterQueryTree.getColumn();
-    ColumnMetadata metadata = columnMetadataMap.get(column);
-    if (metadata == null) {
-      return false;
+    List<IntRange> partitionRanges;
```
mayankshriv: Refactor this out so that the high level recursive function is concise and clear to read.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -16,40 +16,39 @@
...
 
   @Override
-  public int getPartition(Object value) {
-    if (value instanceof String) {
-      return (Utils.murmur2(((String) value).getBytes()) & 0x7fffffff) % _divisor;
-    } else {
-      throw new IllegalArgumentException(
-          "Illegal argument for partitioning, expected Integer, got: " + value.getClass().getName());
-    }
+  public int getPartition(Object valueIn) {
```
mayankshriv: Is there a unit test for this?

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
rven1: Done

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/MutableDictionaryReader.java
rven1: We only print the first time we fall out of the partition after which the partionfunction will be set to null and no further checks will be done.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: Mayank: The number of partitions can change at any time. The current kafka partitiion is fixed.

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
rven1: Done

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/LLCRealtimeClusterIntegrationTest.java
rven1: Moved changes to a new test.

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
rven1: Done

pinot-common/src/main/java/com/linkedin/pinot/common/request/helper/ControllerRequestBuilder.java
rven1: Done

pinot-common/src/main/java/com/linkedin/pinot/common/request/helper/ControllerRequestBuilder.java
rven1: ok. Done.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/MurmurPartitionFunction.java
rven1: ByteArrayPartitioner is needed to force the local partition. Murmur is for the tracking in  aggregate which uses open source kafka partitioning.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/AbstractSegmentPruner.java
rven1: It is better to only extract the metadata which is needed. There could be metadata for other pruners which we don't want to extract. Extracting is work which is done once for limited use.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
rven1: Done

pinot-common/src/main/java/com/linkedin/pinot/common/request/helper/ControllerRequestBuilder.java
jfim: Should be empty map, not `null`. This way you avoid having to add a null check. See Effective Java Item 43: Return empty arrays or collections, not nulls.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -840,6 +845,19 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
...
+
+    // Partitioners
+    _partitioners = indexingConfig.getPartitioners();
+    if (_partitioners != null && _partitioners.size() > 0) {
+      try {
+        _nPartitions = _consumerWrapper.getPartitionCount(_kafkaTopic, /*maxWaitTimeMs=*/5000L);
+        _segmentPartitionConfig = createPartitionConfig();
+        _realtimeSegment.setSegmentPartitionConfig(_segmentPartitionConfig);
+      } catch (Exception e) {
+        segmentLogger.info("Couldn't get number of partitions in 5s, not using partition config");
```
jfim: Should be warn, not info.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -840,6 +845,19 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
     // Create field extractor
     _fieldExtractor = FieldExtractorFactory.getPlainFieldExtractor(schema);
     makeConsumerWrapper();
+
+    // Partitioners
+    _partitioners = indexingConfig.getPartitioners();
+    if (_partitioners != null && _partitioners.size() > 0) {
```
jfim: `!_partitioners.isEmpty()`

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/MutableDictionaryReader.java
```diff
@@ -186,4 +202,63 @@ public void print() {
...
+   * Checks if the given value lies within one of the partition ranges. If the value is not within
+   * any of the partition ranges, partitioning is dropped. The effect of that is:
+   * <ul>
+   *   <li> Subsequent values will not be checked against partition ranges. </li>
+   *   <li> Partition information will not be written out to the metadata for this column. </li>
+   * </ul>
+   *
+   * @param value Column value to check.
+   */
+  protected void checkPartition(Object value) {
```
jfim: Partitioning is not a dictionary concern and should be placed in either `RealtimeSegmentStatsContainer` or a related class. It makes no sense design-wise to have a partition function on a dictionary, especially considering partitioning only makes sense on a segment basis.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -169,6 +172,9 @@ public boolean isFinal() {
   private final String _tableName;
   private final List<String> _invertedIndexColumns;
   private final List<String> _noDictionaryColumns;
+  private final Map<String, String> _partitioners;
```
mcvsubbu: rename it as _column2PartitionerMap ?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/MutableDictionaryReader.java
mcvsubbu: Try to see if you can get the segment name in the log message. Otherwise, with multiple segments being consumed simultaneously, you will have no way of knowing which segment did not fall into the partition correctly.
Alternatively, add a log when the segment is being built as to whether the partition column fitted right or not.

pinot-common/src/main/java/com/linkedin/pinot/common/request/helper/ControllerRequestBuilder.java
mayankshriv: Sorry, I didn't ask the question correctly. My concern was to just make sure that if there's no partition config, nothing gets written out. It seemed that with empty map, the code would have written out the empty partition config, as opposed to not writing it at all.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -20,29 +20,44 @@
...
+  public RealtimeColumnStatistics(RealtimeColumnDataSource dataSource, int[] sortedDocIdIterationOrder, ColumnPartitionConfig columnPartitionConfig) {
     _dataSource = dataSource;
     _sortedDocIdIterationOrder = sortedDocIdIterationOrder;
     _dictionaryReader = dataSource.getDictionary();
     _block = dataSource.getNextBlock();
+    if (columnPartitionConfig != null) {
+      partitionFunction = columnPartitionConfig.getPartitionFunction();
+      partitionRanges = columnPartitionConfig.getPartitionRanges();
+      if (partitionFunction != null) {
+        checkPartition();
```
mayankshriv: Can we move checkPartition() into the Partition function interface?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -146,11 +161,42 @@ public boolean hasNull() {
...
+   * For each element in the dictionary for a column
+   * Checks if the given value lies within one of the partition ranges. If the value is not within
+   * any of the partition ranges, partitioning is dropped. The effect of that is:
+   * <ul>
+   *   <li> Subsequent values will not be checked against partition ranges. </li>
+   *   <li> Partition information will not be written out to the metadata for this column. </li>
+   * </ul>
+   *
+   */
+  void checkPartition() {
```
mayankshriv: This should be moved to within PartitionFunction.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/ColumnStatistics.java
```diff
@@ -76,4 +76,8 @@
     PartitionFunction getPartitionFunction();
 
     List<IntRange> getPartitionRanges();
+
+    /* A partition function was specified but some rows did not satisfy it */
+    boolean getPartitionMismatch();
```
mayankshriv: May be renamed to isPartitioned and change the return value accordingly?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -928,4 +945,14 @@ public String getSegmentName() {
   public int getMaxTimeForConsumingToOnlineSec() {
     return _maxTimeForConsumingToOnlineSec;
   }
+
+  private SegmentPartitionConfig createPartitionConfig() {
+    Map<String, ColumnPartitionConfig> map = new HashedMap();
+    for (String column : _column2PartitionersMap.keySet()) {
+      map.put(column, new ColumnPartitionConfig(
+          _column2PartitionersMap.get(column) + ColumnPartitionConfig.PARTITIONER_DELIMITER + _nPartitions,
+          "[" + _kafkaPartitionId + " " + _kafkaPartitionId + "]"));
```
jfim: It would be good to have this logic separated somewhere else instead of passing Strings around. For example, it would make better sense design-wise to have something like `new ColumnPartitionConfig(partitioner, partitionCount, PartitionRange.build(kafkaPartitionId, kafkaPartitionId))` instead of expecting a specific string format.

Since there's already a billion comments on the PR, we can do it in a different PR.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -0,0 +1,53 @@
...
+  private static final String NAME = "ByteArray";
+  private final int _divisor;
+  public ByteArrayPartitioner _byteArrayPartitioner;
+
+  /**
+   * Constructor for the class.
+   * @param args Arguments for the partition function.
+   */
+  public ByteArrayPartitionFunction(String[] args) {
+    Preconditions.checkArgument(args.length == 1);
```
jfim: Same idea, have proper typed arguments instead of passing specially-formatted strings around.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -146,11 +161,42 @@ public boolean hasNull() {
...
+    // Iterate over the dictionary to check the partitioning
+    final int length = _dictionaryReader.length();
+    for (int i = 0; ((i < length) && (partitionFunction != null)); i++) {
+      int partition = partitionFunction.getPartition(_dictionaryReader.get(i));
+      for (IntRange partitionValue : partitionRanges) {
+        if (partition < partitionValue.getMinimumInteger() || partition > partitionValue.getMaximumInteger()) {
```
jfim: IntRange has a `containsInteger` method, this would be more readable as `!partitionValue.containsInteger(partition)` (especially if you rename partitionValue to partitionRange).

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -840,6 +844,19 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
     // Create field extractor
     _fieldExtractor = FieldExtractorFactory.getPlainFieldExtractor(schema);
     makeConsumerWrapper();
+
+    // Partitioners
+    _column2PartitionersMap = indexingConfig.getPartitioners();
+    if (_column2PartitionersMap != null && !_column2PartitionersMap.isEmpty()) {
+      try {
+        _nPartitions = _consumerWrapper.getPartitionCount(_kafkaTopic, /*maxWaitTimeMs=*/5000L);
```
rven1: The realistic latency is in the order of 10's of ms. The 5s limit was in keeping with the existing code.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -840,6 +844,19 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
...
+        _segmentPartitionConfig = createPartitionConfig();
+        _realtimeSegment.setSegmentPartitionConfig(_segmentPartitionConfig);
+      } catch (Exception e) {
+        segmentLogger.warn("Couldn't get number of partitions in 5s, not using partition config");
```
rven1: Formally setting it to null.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.core.data.partition;
+
+import com.google.common.base.Preconditions;
+import kafka.producer.ByteArrayPartitioner;
```
rven1: IDE only gives kafka.producer.ByteArrayPartitioner, but I changed it to org.apache.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -0,0 +1,53 @@
...
+
+
+/**
+ * Implementation of {@link Byte array partitioner}
+ * {@link org.apache.kafka.producer.ByteArrayPartitioner}.
+ *
+ */
+public class ByteArrayPartitionFunction implements PartitionFunction {
```
rven1: This is a reasonable partitioner to use which guarantees consistency between producer and consumer. It could be used by open source.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/ColumnStatistics.java
rven1: Talked with Mayank. Need this for mismatch checking.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: I agree, Will do it in a separate PR.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
rven1: I agree. Will do it in a separate PR.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
rven1: Will do as a follow-up

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
rven1: Will do as a follow-up

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -840,6 +844,20 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
     // Create field extractor
     _fieldExtractor = FieldExtractorFactory.getPlainFieldExtractor(schema);
     makeConsumerWrapper();
+
+    // Partitioners
+    _column2PartitionersMap = indexingConfig.getPartitioners();
+    if (_column2PartitionersMap != null && !_column2PartitionersMap.isEmpty()) {
+      try {
+        _nPartitions = _consumerWrapper.getPartitionCount(_kafkaTopic, /*maxWaitTimeMs=*/5000L);
```
mcvsubbu: How do we make sure that the variable _nPartitions does not get used for some other purpose if there is an exception here? Do you need it as a member variable? Why not leave it local to this method?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mcvsubbu: +1 to that.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -840,6 +844,20 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
...
+        _segmentPartitionConfig = createPartitionConfig();
```
mcvsubbu: Do you need _segmentPartitionConfig as a member variable ? If it is not used anywhere else then make it local

rven1: I agree. Will make the change separately.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -840,6 +844,20 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
...
+        _realtimeSegment.setSegmentPartitionConfig(_segmentPartitionConfig);
+      } catch (Exception e) {
+        segmentLogger.warn("Couldn't get number of partitions in 5s, not using partition config");
```
mcvsubbu: Please call makeConsumerWrapper() again when the exception is fielded. The consumer wrapper is a stateful object, and we have found it better to create a new one when experiencing transient Kafka failures. In this case, whether the exception is transient or permanent, it is better to create a new consumer wrapper. See other uses of this method and the conditions in which they are called.

rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: Fixed by setting segment partition config to null in which case number of partitions is not relevant.
It makes sense to make npartitions a local variable. Will do it as a separate PR.

rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: Done

PR #1235, opened 2017-04-04
Remove keyword 'LIKE' from PQL grammar.

mayankshriv:
Removing the keyword 'LIKE' from PQL grammar as it is not supported
currently.


Jackie-Jiang:
WHY DO WE NEED THIS PR？？？

mayankshriv:
@Jackie-Jiang Please refer to the changes in the PR. The 'LIKE' keyword is being removed. What is your concern?

PR #1237, opened 2017-04-04
Log info instead warning for bad queries

Jackie-Jiang:


pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -164,7 +164,7 @@ public BrokerResponse handleRequest(@Nonnull JSONObject request)
     try {
       brokerRequest = REQUEST_COMPILER.compileToBrokerRequest(pql);
     } catch (Exception e) {
-      LOGGER.warn("Parsing error on requestId {}: {}", requestId, pql, e);
+      LOGGER.info("Parsing error on requestId {}: {}", requestId, pql, e);
```
mayankshriv: Since we made it info, can we just print e.getMessage() instead of the entire stack trace. Consider where applicable.

Jackie-Jiang: Fixed.

PR #1238, opened 2017-04-04
Fix the NPE for realtime segment

Jackie-Jiang:
In realtime consuming segment, we first add value to dictionary then update inverted index.
If we query a column with inverted index, it is possible that the inverted index has not been generated.
The fix is filter out all null inverted index.
This will not affect the accuracy of the results because we just ignore a single under-indexing record.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BitmapBasedFilterOperator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BitmapBasedFilterOperator.java
@@ -15,10 +15,6 @@
  */
 package com.linkedin.pinot.core.operator.filter;
 
-import org.roaringbitmap.buffer.ImmutableRoaringBitmap;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.linkedin.pinot.core.common.Block;
 import com.linkedin.pinot.core.common.BlockId;
 import com.linkedin.pinot.core.common.DataSource;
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BitmapBasedFilterOperator.java
```diff
@@ -86,11 +86,26 @@ public BaseFilterBlock nextFilterBlock(BlockId BlockId) {
...
+      if (bitmap != null) {
+        bitmaps.add(bitmap);
+      }
     }
-    bitmapBlock = new BitmapBlock(dataSource.getOperatorName(), dataSourceBlock.getMetadata(), startDocId, endDocId, bitmaps, exclusion);
+
+    // Log size diff to verify the fix
+    int numBitmaps = bitmaps.size();
+    if (numBitmaps != length) {
+      LOGGER.info("Not all inverted indexes are generated, numDictIds: {}, numBitmaps: {}", length, numBitmaps);
```
mcvsubbu: This log may appear for every query to every realtime consuming segment, is that right? Can we change this to debug? Do we want to add a metric if this is important?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BitmapBasedFilterOperator.java
```diff
@@ -86,11 +86,26 @@ public BaseFilterBlock nextFilterBlock(BlockId BlockId) {
...
         throw new UnsupportedOperationException("Regex is not supported");
     }
-    ImmutableRoaringBitmap[] bitmaps = new ImmutableRoaringBitmap[dictionaryIds.length];
-    for (int i = 0; i < dictionaryIds.length; i++) {
-      bitmaps[i] = invertedIndex.getImmutable(dictionaryIds[i]);
+
+    // For realtime use case, it is possible that inverted index has not yet generated for the given dict id, so we
+    // filter out null bitmaps
+    int length = dictionaryIds.length;
+    List<ImmutableRoaringBitmap> bitmaps = new ArrayList<>(length);
```
mcvsubbu: hopefully no perf impact here?

Jackie-Jiang: Compare to the actual filter / aggregation, this overhead is negligible.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BitmapBasedFilterOperator.java
Jackie-Jiang: No, this info will only appear when the current code throw NPE. This is just for bug verification.

kishoreg:
Why should we start the indexing immediately. Can we not start that from
the next segment.

kishoreg:
I think the planning phase should look at the metadata of consuming segment
and not use inverted index.

Jackie-Jiang:
@kishoreg Not using inverted index for consuming segment might cause performance issue and high query latency.

PR #1242, opened 2017-04-05
Log info instead of error for NumberFormatException

Jackie-Jiang:


pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -146,18 +147,24 @@ public DataTable processQuery(final QueryRequest queryRequest, ExecutorService e
...
 
       // Update the total docs in the metadata based on un-pruned segments.
       dataTableMetadata.put(DataTable.TOTAL_DOCS_METADATA_KEY, String.valueOf(totalRawDocs));
       return dataTable;
     } catch (Exception e) {
       _serverMetrics.addMeteredQueryValue(instanceRequest.getQuery(), ServerMeter.QUERY_EXECUTION_EXCEPTIONS, 1);
-      LOGGER.error("Exception processing requestId {}", requestId, e);
+
+      // Do not log error for NumberFormatException because it's caused by bad query
+      if (e instanceof NumberFormatException) {
```
mcvsubbu: May be better to add a new RuntimException called (say) BadQueryException.
If we explicitly raise this exception, then we can use it for multiple different errors

jfim:
Why do we want to log the exception as info instead of warn/error?

Jackie-Jiang:
@jfim Because the NumberFormatException is caused by bad query (e.g. select * from table where int_column = "abc"). We should not log warn/error for client wrong behavior. I will add a new Exception type to handle the Bad Request.

PR #1243, opened 2017-04-05
Fixing race condition between segment push/delete.

jenniferdai:
Logging an info message for the NullPointerException caused by race condition between segment being pushed and either segment being deleted or table being deleted.


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/LLCSegmentConsumed.java
```diff
@@ -51,7 +51,11 @@ public Representation get() {
       return new StringRepresentation(SegmentCompletionProtocol.RESP_FAILED.toJsonString());
     }
     SegmentCompletionProtocol.Request.Params reqParams = new SegmentCompletionProtocol.Request.Params();
-    reqParams.withSegmentName(segmentName).withInstanceId(instanceId).withOffset(Long.valueOf(offset));
+    try {
+      reqParams.withSegmentName(segmentName).withInstanceId(instanceId).withOffset(Long.valueOf(offset));
```
mcvsubbu: Hmm.. this statement (reqParams.with()...) should never throw an NPE.

PR #1246, opened 2017-04-05
When validating schema, log info instead of error

Jackie-Jiang:


pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
```diff
@@ -255,54 +255,54 @@ public boolean validate(Logger ctxLogger) {
...
+                break;
+            }
+            break;
+          default:
+            ctxLogger.info("Unsupported field type: {} for field: {}", dataType, fieldName);
+            isValid = false;
+            break;
         }
+      } catch (Exception e) {
+        ctxLogger.info("Caught exception while validating {} field {} dataType {}", fieldType, fieldName, dataType, e);
```
mcvsubbu: e.getMessage()

PR #1248, opened 2017-04-06
[TE] Add data filter for removing noisy time series

cyenjung:
1. Introduce a data filter stage before anomaly detection in order to remove noisy time series, e.g., ratio time series that is generated when traffic or total count is low.

2. Implement an Average Threshold data filter which average the values of the given time series and test if it exceeds the threshold. The filter provides two advanced usages:
  I. A live zone could be set on the values of bucket. For instance, we could ignore all buckets with very small values, which could be noise of the time series. A live zone could be defined by a min and/or a max value.
  II. User could define the override threshold for certain dimensions. For example, we could override the threshold for all time series whose country=US.

3. Add roll up option to disable/enable rolling up low volume time series. The roll up functionality was designed for UI; however, this function could swallow time series for anomaly detection. Now we could disable this function through this flag.

Tested with new unit tests and local integration test, which backfills anomalies for one month.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/datafilter/AverageThresholdDataFilter.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/datafilter/AverageThresholdDataFilter.java
@@ -11,22 +11,34 @@ import java.util.NavigableMap;
 import java.util.TreeMap;
 import org.apache.commons.lang3.StringUtils;
 
+
+/**
+ * The data filter determines whether the average value of a time series passes the threshold.
+ *
+ * Advance Usage 1: A bucket of the time series is taken into consider only if its value is located inside the live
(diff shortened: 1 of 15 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/datafilter/AverageThresholdDataFilter.java
```diff
@@ -0,0 +1,174 @@
...
+import org.apache.commons.collections.MapUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The data filter determines whether the average value of a time series passes the threshold.
+ *
+ * Advance Usage 1: A bucket of the time series is taken into consider only if its value is located inside the live
+ * zone, which is specified by minLiveZone and maxLveZone. In other words, if a bucket's value is smaller than
```
npawar: minor nit: typo at maxLveZone in 2 places

PR #1255, opened 2017-04-07
Follow-up from partitioning pruner changes

rven1:
Addressed some outstanding comments from https://github.com/linkedin/pinot/pull/1230

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
@@ -19,6 +19,7 @@ import com.linkedin.pinot.core.data.partition.PartitionFunction;
 import com.linkedin.pinot.core.data.partition.PartitionFunctionFactory;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import javax.annotation.Nonnull;
 import org.apache.commons.lang.math.IntRange;
 import org.codehaus.jackson.annotate.JsonIgnore;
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
```diff
@@ -48,6 +49,11 @@ public ColumnPartitionConfig(@Nonnull @JsonProperty("functionName") String funct
     _partitionValues = partitionValues;
   }
 
+  public ColumnPartitionConfig(Map<String, String> map, String column, int partitionId, int nPartitions) {
+    _functionName = map.get(column) + PARTITION_VALUE_DELIMITER + nPartitions;
+    _partitionValues = "[" + partitionId + "]";
```
mcvsubbu: should we not have 2 values here?

rven1: Made the change and pushed new version.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/ColumnPartitionConfig.java
```diff
@@ -48,6 +49,11 @@ public ColumnPartitionConfig(@Nonnull @JsonProperty("functionName") String funct
     _partitionValues = partitionValues;
   }
 
+  public ColumnPartitionConfig(Map<String, String> map, String column, int partitionId, int nPartitions) {
+    _functionName = map.get(column) + PARTITIONER_DELIMITER + nPartitions;
+    _partitionValues = "[" + partitionId + " " + partitionId + "]";
```
mcvsubbu: isn't it safer to call the rangesToString() method that is defined in this class?

PR #1256, opened 2017-04-07
Add JSON string serialization/de-serialization apis for SegmentPartitionConfig.

mayankshriv:
1. Added api's in SegmentPartitionConfig for
serialization/de-serialization to/from JSON string.

2. Added unit test for SegmentPartitionConfig class.

### DIFF
--- a/pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
+++ b/pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
@@ -36,6 +36,7 @@ import com.linkedin.pinot.core.segment.index.SegmentMetadataImpl;
 import com.linkedin.pinot.core.segment.index.loader.Loaders;
 import com.linkedin.pinot.pql.parsers.Pql2Compiler;
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
```diff
@@ -220,16 +221,67 @@ public void testStringRangeConversions() {
...
+        builder.append("[").append(j).append(" ").append(j + 1).append("]");
+        if (j < numValues - 1) {
+          builder.append(ColumnPartitionConfig.PARTITION_VALUE_DELIMITER);
+        }
+      }
+      expectedMap.put(partitionColumn, new ColumnPartitionConfig(partitionFunction, builder.toString()));
+    }
+
+    SegmentPartitionConfig expectedConfig = new SegmentPartitionConfig(expectedMap);
+    SegmentPartitionConfig actualConfig = SegmentPartitionConfig.fromJsonString(expectedConfig.toJsonString());
```
mcvsubbu: will be useful to add a compatibility test as well. Have a json that has some new fields in it, and make sure it is deserialzed correctly.

mayankshriv: Yes, it is already there in the same test, check a few lines below.

PR #1258, opened 2017-04-08
Derive partition ranges during indexing instead of relying on config.

mayankshriv:
1. Compute the partition range during segment generation instead of
getting it from segment generation config. We will generate only one
range currently, instead of a list of ranges. Metadata and pruner still
support a list of partition ranges.

2. The new code will now always write out a partition range as computed
while indexing. This is different from the existing code that would not
write the partition range metadata if it is not as expected.

3. Updated tests to reflect this change.


pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/PartitionFunctionFactory.java
```diff
@@ -84,4 +84,28 @@ public static PartitionFunction getPartitionFunction(@Nonnull String partitionFu
...
+   * @param functionName Name of partition function
+   * @param arguments Arguments for partition function
+   * @return Partition function string.
+   */
+  public static String buildPartitionFunctionString(String functionName, Object... arguments) {
+    StringBuilder builder = new StringBuilder(functionName);
+    builder.append(PARTITION_FUNCTION_DELIMITER);
+
+    for (int i = 0; i < arguments.length; i++) {
+      builder.append(arguments[i].toString());
```
Jackie-Jiang: builder.append(PARTITION_FUNCTION_DELIMITER);
builder.append(arguments[i].toString());

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -166,37 +167,30 @@ public PartitionFunction getPartitionFunction() {
...
-   *   <li> Partition information will not be written out to the metadata for this column. </li>
-   * </ul>
+   * Update partition ranges based on column values.
    *
    */
-  void checkPartition() {
+  void updatePartition() {
     // Iterate over the dictionary to check the partitioning
     final int length = _dictionaryReader.length();
     for (int i = 0; ((i < length) && (partitionFunction != null)); i++) {
```
Jackie-Jiang: No need for the null check

PR #1260, opened 2017-04-08
Fix to recompute routing table when instance config changes in Helix

kishoreg:
currently, we never update Routing on instanceConfigChange callback because of the following code in HelixExternalViewBasedRouting.isRoutingTableRebuildRequired.
```
      // If it's the same znode, don't bother comparing the contents of the instance configs
      if (previousInstanceConfig.getRecord().getVersion() == currentInstanceConfig.getRecord().getVersion()) {
        continue;
      }
```
Pinot expected Helix to populate the version field but that does not seem to be the case. Will file a separate Helix ticket to fix this in Helix. 
For now, since we already read the versions in processInstanceConfig method, the fix was to set the version in the instanceConfig

### DIFF
--- a/pinot-transport/src/main/java/com/linkedin/pinot/routing/HelixExternalViewBasedRouting.java
+++ b/pinot-transport/src/main/java/com/linkedin/pinot/routing/HelixExternalViewBasedRouting.java
@@ -651,6 +651,7 @@ public class HelixExternalViewBasedRouting implements RoutingTable {
     }
 
     long statFetchStart = System.currentTimeMillis();
+    Map<String, Stat> instanceConfigStatMap = new HashMap<>();
     Stat[] instanceConfigStats =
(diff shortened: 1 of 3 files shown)


pinot-transport/src/main/java/com/linkedin/pinot/routing/HelixExternalViewBasedRouting.java
```diff
@@ -688,6 +690,13 @@ public void processInstanceConfigChange() {
     if (!affectedTables.isEmpty()) {
       long icFetchStart = System.currentTimeMillis();
       List<InstanceConfig> instanceConfigs = helixDataAccessor.getChildValues(propertyKeyBuilder.instanceConfigs());
+      // Helix does not set the version field, we need to set it explicitly
```
jfim: Can you add a Helix bug ID so that we can remove this when it's fixed in Helix?

jfim:
I've added an integration test for this.

PR #1261, opened 2017-04-10
Max data time endpoint, topk in heatmap, rounding off double values

npawar:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/util/ThirdEyeUtils.java
```diff
@@ -390,4 +394,25 @@ public static MetricConfigDTO getMetricConfigFromNameAndDataset(String metricNam
...
+  /**
+   * Get rounded double value, according to the value of the double.
+   * For values gte 0.1, use ##.## (eg. 123, 2.5, 1.26, 0.5, 0.162)
+   * For values lt 0.1 and gte 0.01, use ##.### (eg. 0.08, 0.071, 0.0123)
+   * For values lt 0.01 and gte 0.001, use ##.#### (eg. 0.001, 0.00367)
+   * This function ensures we don't prematurely round off double values to a fixed format, and make it 0.00 or lose out information
+   * @param value
+   * @return
+   */
+  public static String getRoundedValue(double value) {
```
kishoreg: We should select the format looking at all the values right. Having a different format for each value is not good.

PR #1262, opened 2017-04-10
Change the log level for transient exceptions

jfim:
Change the log level to info for transient exceptions (eg. current
Kafka broker is not leader for this Kafka partition, no such
topic-partition present on this Kafka broker, etc.) for the retriable
part of the Kafka consumption loop.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -61,6 +61,7 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
+import javax.annotation.Nonnull;
 import kafka.message.MessageAndOffset;
 import org.apache.commons.collections.map.HashedMap;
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -249,13 +250,13 @@ private boolean endCriteriaReached() {
...
 
-  private void handleTransientKafkaErrors(Exception e) throws  Exception {
+  private void handleTransientKafkaErrors(@Nonnull Exception e) throws Exception {
     consecutiveErrorCount++;
     if (consecutiveErrorCount > MAX_CONSECUTIVE_ERROR_COUNT) {
       segmentLogger.warn("Kafka transient exception when fetching messages, stopping consumption after {} attempts", consecutiveErrorCount, e);
       throw e;
     } else {
-      segmentLogger.warn("Kafka transient exception when fetching messages, retrying (count={})", consecutiveErrorCount, e);
+      segmentLogger.info("Kafka transient exception when fetching messages, retrying (count = {}, cause = {})", consecutiveErrorCount, e.toString());
```
mcvsubbu: should be warn, but we don't need to log the exception stack.
ideally, we should log an INFO when we get back from the transient error, and the complete stack and error if we cannot.

PR #1263, opened 2017-04-10
[PINOT-4748] Retry on temporary kafka failures in the controller

mcvsubbu:
Kafka failures when we try to get number of partitions, or the current offset from Kafka
are usually temporary, and a single retry should fix things. Added retry logic for these
cases, logging error only when all retries fail.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
@@ -15,10 +15,10 @@
  */
 package com.linkedin.pinot.controller.helix.core;
 
-import com.linkedin.pinot.core.realtime.impl.kafka.KafkaLowLevelStreamProviderConfig;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import org.apache.commons.compress.utils.IOUtils;
 import org.apache.helix.HelixAdmin;
 import org.apache.helix.ZNRecord;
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
```diff
@@ -283,4 +283,52 @@ private static String getGroupIdFromRealtimeDataTable(String realtimeTableName,
...
+    private int getPartitionCount() {
+      return _partitionCount;
+    }
+
+    private Exception getException() {
+      return _exception;
+    }
+
+    @Override
+    public Boolean call() throws Exception {
```
mayankshriv: Can some of the information be cached across retries (eg not creating new SimpleConsumerWrapper per retry, etc)

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
```diff
@@ -283,4 +283,52 @@ private static String getGroupIdFromRealtimeDataTable(String realtimeTableName,
...
+      if (bootstrapHosts == null || bootstrapHosts.isEmpty()) {
+        throw new RuntimeException("Invalid value for " + Helix.DataSource.Realtime.Kafka.KAFKA_BROKER_LIST);
+      }
+      SimpleConsumerWrapper consumerWrapper = SimpleConsumerWrapper.forMetadataConsumption(
+          new KafkaSimpleConsumerFactoryImpl(), bootstrapHosts,
+          PinotTableIdealStateBuilder.class.getSimpleName() + "-" + kafkaTopicName, KAFKA_CONNECTION_TIMEOUT_MILLIS);
+
+      try {
+        _partitionCount = consumerWrapper.getPartitionCount(kafkaTopicName, /*maxWaitTimeMs=*/5000L);
+        if (_exception != null) {
```
mayankshriv: Just to confirm, each instance of this class is executed only by one thread?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
mcvsubbu: No, we actually want to create a new SimpleConsumerWrapper if it throws transient exception. (Currently) it does not reset its state correctly.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
mcvsubbu: Correct.

mcvsubbu:
Regarding getting the number of partitions in the server, I had already discussed this with @rven1 We do not need a retry there, since on any exception, we choose not to have the partition information in the segment metadata. It does not affect consumption in any way.

PR #1264, opened 2017-04-10
Enhance IndexingConfig to have SegmentPartitionConfig.

mayankshriv:
The current indexing config has a map of column to partition function
name, which is not sufficient information. Also, it imposes the burden
on client of IndexingConfig to parse this info and build
SegmentPartitionConfig. This change embeds SegmentPartitionConfig inside
of IndexingConfig, thus solving both the problems.

1. Added partition config inside of indexing config.
2. Separated out number of partitions as a first class member of
partition config, as opposed to embeded in partition function string.
3. Moved the config classes to config, as IndexingConfig resides there.
4. Enhanced unit tests to cover the changes.


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -844,12 +843,12 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
...
-    if (_column2PartitionersMap != null && !_column2PartitionersMap.isEmpty()) {
+    SegmentPartitionConfig segmentPartitionConfig = indexingConfig.getSegmentPartitionConfig();
+    if (segmentPartitionConfig != null) {
       try {
         int nPartitions = _consumerWrapper.getPartitionCount(_kafkaTopic, /*maxWaitTimeMs=*/5000L);
-        _realtimeSegment.setSegmentPartitionConfig(createPartitionConfig(nPartitions));
+        segmentPartitionConfig.setNumPartitions(nPartitions);
+        _realtimeSegment.setSegmentPartitionConfig(segmentPartitionConfig);
       } catch (Exception e) {
         segmentLogger.warn("Couldn't get number of partitions in 5s, not using partition config");
```
mcvsubbu: can u please add e.getMessage() to this warning? thanks. segmentLogger automatically adds segmentName (that has table name and partition ID)

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -21,33 +21,36 @@
 
 /**
  * Implementation of {@link Byte array partitioner}
- * {@link org.apache.kafka.producer.ByteArrayPartitioner}.
  *
  */
 public class ByteArrayPartitionFunction implements PartitionFunction {
```
mcvsubbu: Since this class is used only for testing, I would suggest moving it to some test utils. I thought I had already asked @rven1 to do this, but then maybe I missed.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/MurmurPartitionFunction.java
```diff
@@ -29,26 +29,30 @@
  */
 public class MurmurPartitionFunction implements PartitionFunction {
```
mcvsubbu: I would also add an ALERT to the comment block above.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
mcvsubbu: Ah,I take that back. You need this in the factory, I guess.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
```diff
@@ -15,6 +15,21 @@
  */
 package com.linkedin.pinot.core.indexsegment.generator;
 
+import com.google.common.base.Preconditions;
```
mcvsubbu: no other changes to this file other than re-ordering imports?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/ColumnIndexCreationInfo.java
```diff
@@ -37,11 +37,12 @@
   private final PartitionFunction partitionFunction;
   private final List<IntRange> partitionRanges;
   private final Object defaultNullValue;
+  private int numPartitions;
```
mcvsubbu: Not sure if this is always number of Kafka partitions. If so, then can we just be explicit with numKafkaPartitions? if not, then maybe we can think of another name?

pinot-common/src/main/java/com/linkedin/pinot/common/config/SegmentPartitionConfig.java
```diff
@@ -51,9 +52,32 @@ public SegmentPartitionConfig(
...
+  }
+
+  /**
+   * Set the number of partitions for the specified column.
+   *
+   * @param column Column for which to set the number of partitions.
+   * @param numPartitions Number of partitions to set.
+   */
+  @JsonIgnore
+  public void setNumPartitions(String column, int numPartitions) {
```
mcvsubbu: Is it possible for each column to have a different number of partitions? If not, can we remove this method (seems like not used as far as I could see, but I may have missed something). 
If yes, then does json package handle the two set methods correctly? Does it make sense to rename it as setNumPartitionsForColumn()?

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/ByteArrayPartitionFunction.java
mayankshriv: Yes.

pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/MurmurPartitionFunction.java
mayankshriv: Will add a TODO.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
mayankshriv: Moving SegmentPartitionConfig from core to common now requires adding the following explicitly.
import com.linkedin.pinot.common.config.SegmentPartitionConfig

That is the change in this file.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/ColumnIndexCreationInfo.java
mayankshriv: For offline it is not necessarily kafka partitions. I think numPartitions is fine.

pinot-common/src/main/java/com/linkedin/pinot/common/config/SegmentPartitionConfig.java
mayankshriv: If code is designed to handle different number of partitions per column. There are calls for both functions.

PR #1269, opened 2017-04-11
Handle invalid content type while reading input schema

atumbde:


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSchemaRestletResource.java
```diff
@@ -145,7 +146,18 @@ public Representation post(Representation entity) {
   })
   private Representation uploadNewSchema()
       throws Exception {
-    File dataFile = getUploadContents();
+    File dataFile;
+    try {
+      dataFile = getUploadContents();
+    } catch (FileUploadBase.InvalidContentTypeException e) {
+      LOGGER.info("Invalid content type while adding new schema");
```
mcvsubbu: log any additional info here. URI?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSchemaRestletResource.java
```diff
@@ -145,7 +146,18 @@ public Representation post(Representation entity) {
...
+      return errorResponseRepresentation(Status.CLIENT_ERROR_BAD_REQUEST,
+          e.getMessage());
+    } catch (Exception e) {
+      LOGGER.error("Error reading request body while adding new schema", e);
```
mcvsubbu: Log URI if possible

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSchemaRestletResource.java
```diff
@@ -207,7 +219,18 @@ public Representation put(Representation entity) {
   private Representation uploadSchema(
       @Parameter(name = "schemaName", in = "path", description = "The name of the schema to get")
       String schemaName) throws Exception {
-    File dataFile = getUploadContents();
+    File dataFile;
+    try {
+      dataFile = getUploadContents();
+    } catch (FileUploadBase.InvalidContentTypeException e) {
+      LOGGER.info("Invalid content type while updating schema {}", schemaName);
```
mcvsubbu: ditto

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSchemaRestletResource.java
```diff
@@ -207,7 +219,18 @@ public Representation put(Representation entity) {
...
+      return errorResponseRepresentation(Status.CLIENT_ERROR_BAD_REQUEST,
+          e.getMessage());
+    } catch (Exception e) {
+      LOGGER.error("Error reading request body while updating new schema: {}", schemaName, e);
```
mcvsubbu: ditto, and can this code be merged to be in one place?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSchemaRestletResource.java
antumbde: URIs are not useful. They just say /schema which we know already

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSchemaRestletResource.java
antumbde: as above. URIs are not useful

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSchemaRestletResource.java
antumbde: as above but this one has schema name which is logged (we don't have schema name above)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSchemaRestletResource.java
antumbde: Making separate method will require different exception to allow early return. It also needs if-else to handle optional schemaName in logs. So, there is no real benefit.

We can make a common method that handles schema update that is called by both PUT and POST apis. But there are subtle differences and no UTs. So, I want to limit the scope of this change.

PR #1272, opened 2017-04-11
Implement parsing function from and to JSON String

cyy0714:
1. Implement the JSON parser from and to AnomalyTimelinesView to JSON.
2. As long as the getter and setter is properly set, the ObjectMapper can deal with all the variables in the class.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/views/AnomalyTimelinesView.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/views/AnomalyTimelinesView.java
@@ -1,13 +1,23 @@
 package com.linkedin.thirdeye.anomaly.views;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/views/AnomalyTimelinesView.java
```diff
@@ -1,13 +1,23 @@
...
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 
 public class AnomalyTimelinesView {
+  private static ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  public static String TIME_BUCKETS = "timeBuckets";
```
npawar: Where are these variables being used?

cyy0714: It's not used. Will remove them

PR #1287, opened 2017-04-12
Add a blocking shutdown API to NettyServer to ensure server is correctly shutdown in tests

Jackie-Jiang:


pinot-transport/src/main/java/com/linkedin/pinot/transport/netty/NettyServer.java
```diff
@@ -170,6 +170,29 @@ public void shutdownGracefully() {
   }
 
   /**
+   * Shutdown completely.
+   * <p>This is a blocking call to wait until the shutdown is completely done.
+   */
+  public void shutdownCompletely() {
```
mcvsubbu: Take the time here as an argument.
Suggest rename method as waitForShutdown(long msecsToWait), or shutdownSynchronously(long msecsToWait);

Jackie-Jiang: Fixed.

PR #1288, opened 2017-04-13
Bugfix for server instance shutdown state

jfim:
Fix the bug where marking an instance as shutting down would correctly
remove it from the routing table, but marking it as not shutting down
would not add it back to the routing table if there was more than one
server node in the cluster. Add integration test that catches this
particular edge case.


jfim:
The change in the test is that we added it to the multiple node cluster integration test and added a test that disables/re-enables a single random instance multiple times.

PR #1289, opened 2017-04-13
Fix != in query predicates

jfim:
Fix != in query predicates. The previous grammar did not support
C-style inequalities. For some reason, antlr4 is configured by default
to not reject invalid tokens when lexing the input string (!) and
simply logs the offending tokens to the console. Since ! is not a
valid token, a query that had a clause with a C-style inequality (eg.
WHERE foo != 1234) got silently converted to an equality query (eg.
WHERE foo = 1234).

Add a lexer error listener that correctly throws an exception if there
is an unexpected token in the lexing phase. Add a unit test for this
case.

Since C-style inequalities are supported by some databases, add
C-style inequalities as an alias for <>. Add a unit test for the !=
operator.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
@@ -21,11 +21,16 @@ import com.linkedin.pinot.pql.parsers.pql2.ast.AstNode;
 
 import org.antlr.v4.runtime.ANTLRInputStream;
 import org.antlr.v4.runtime.BailErrorStrategy;
+import org.antlr.v4.runtime.BaseErrorListener;
 import org.antlr.v4.runtime.CharStream;
 import org.antlr.v4.runtime.CommonToken;
 import org.antlr.v4.runtime.CommonTokenFactory;
+import org.antlr.v4.runtime.RecognitionException;
+import org.antlr.v4.runtime.Recognizer;
 import org.antlr.v4.runtime.TokenStream;
 import org.antlr.v4.runtime.UnbufferedTokenStream;
+import org.antlr.v4.runtime.misc.NotNull;
+import org.antlr.v4.runtime.misc.Nullable;
 import org.antlr.v4.runtime.tree.ParseTree;
 import org.antlr.v4.runtime.tree.ParseTreeWalker;
 import org.apache.commons.lang3.exception.ExceptionUtils;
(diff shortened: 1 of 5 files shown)


kishoreg:
does this mean we will support both != and <> for NOT EQUALS?

Jackie-Jiang:
I think his change will accept '!=' as '<>', but reject invalid identifier such as '?='

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
```diff
@@ -42,6 +47,14 @@ public BrokerRequest compileToBrokerRequest(String expression) throws Pql2Compil
       CharStream charStream = new ANTLRInputStream(expression);
       PQL2Lexer lexer = new PQL2Lexer(charStream);
       lexer.setTokenFactory(new CommonTokenFactory(true));
+      lexer.removeErrorListeners();
+      lexer.addErrorListener(new BaseErrorListener() {
+        @Override
+        public void syntaxError(@NotNull Recognizer<?, ?> recognizer, @Nullable Object offendingSymbol, int line,
+            int charPositionInLine, @NotNull String msg, @Nullable RecognitionException e) {
+          throw new Pql2CompilationException(msg);
```
kishoreg: can we add more information in the exception message - exact fragment with error  using the offendingSymbol, line, charPositionInLine

pinot-common/src/test/java/com/linkedin/pinot/pql/parsers/Pql2CompilerTest.java
```diff
@@ -93,4 +94,18 @@ private void testTopZeroFor(Pql2Compiler compiler, String s, final int expectedT
...
+  @Test
+  public void rejectInvalidLexerToken() {
+    assertCompilationFails(new Pql2Compiler(), "select foo from bar where baz ?= 2");
+  }
+
+  @Test
+  public void testCStyleInequalityOperator() {
+    Pql2Compiler compiler = new Pql2Compiler();
+
+    BrokerRequest brokerRequest = compiler.compileToBrokerRequest(
```
kishoreg: what happens for <  > and ! =.  if there is space between the characters? 
Also =! should throw parsing error?

jfim: Done, added to tests.

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
jfim: Done.

PR #1290, opened 2017-04-13
Logging an info message when segment upload interrupted - 93588,97051,97051

jenniferdai:


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -350,7 +351,12 @@ public Representation post(Representation entity) {
 
           // 3/ Request is parsed by the handler which generates a
           // list of FileItems
-          items = upload.parseRequest(getRequest());
+          try {
+            items = upload.parseRequest(getRequest());
+          } catch (FileUploadBase.IOFileUploadException e) {
```
mcvsubbu: I believe you are catching the wrong FileUploadExcepiton here. Should be org.apache.commons.fileupload.FileUploadException. Can you please verify?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -350,7 +351,12 @@ public Representation post(Representation entity) {
...
+            LOGGER.info("Exception uploading segment: " + e);
```
mcvsubbu: may make sense to log the full exception here, as an INFO though.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
jenniferdai: According to the EKG tickets, all of them throw a IOFileUploadException
2015/12/09 04:33:40 ERROR [PinotSegmentUploadRestletResource] [qtp111819772-648954] [pinot-controller] Caught exception in file upload
org.apache.commons.fileupload.FileUploadBase.IOFileUploadException: Processing of mul

jenniferdai: I see an IOFileUploadException in the stack trace!

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
antumbde: Use comma instead of +. That gives full exception stack. That said, this IOException seems to be error condition. We shouldn't turn this to info
I believe there are other exception types thrown here like InvalidContentTypeException that should return 415.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -350,7 +351,12 @@ public Representation post(Representation entity) {
...
+            return new StringRepresentation("Exception uploading segment: " + e);
```
antumbde: this will return 200 with error message. Make sure you return correct error code.
Use  errorRepresentation method from base class for convenience

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
mcvsubbu: This exception will be caused if the incoming request was bad, right? (say, the post did not conform to http protcool, or the connection was dropped and EOF received early). So, an INFO is fine, it seems. Why do you think it should be an ERROR?

antumbde: parseRequest doc says it throws FileUploadException which can indicate bad request format, early connection termination or local disk issue.  
For bad request format  it throws FileUploadBase.InvalidContentType exception that should be info logged with correct return status and message.
IOFileUploadException indicates issue reading network request or writing to local disk. That should be error. Even for connection termination it's hard to say if that's client issue.

mcvsubbu: Discussed this offline, and it looks like the way API handles it, it could be a local error or a remote error, we cannot say. So, this should remain as an error with the stack, I suppose.

PR #1291, opened 2017-04-13
Handling user errors in quota checker API

jenniferdai:
* 149869

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/StorageQuotaChecker.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/StorageQuotaChecker.java
@@ -63,12 +63,22 @@ public class StorageQuotaChecker {
   public QuotaCheckerResponse isSegmentStorageWithinQuota(@Nonnull File segmentFile, @Nonnull String tableNameWithType,
       @Nonnull String segmentName,
       @Nonnegative int timeoutMsec) {
-    Preconditions.checkNotNull(segmentFile);
-    Preconditions.checkNotNull(tableNameWithType);
-    Preconditions.checkNotNull(segmentName);
-    Preconditions.checkArgument(timeoutMsec > 0, "Timeout value must be > 0, input: %s", timeoutMsec);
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/StorageQuotaChecker.java
```diff
@@ -63,12 +63,22 @@ public StorageQuotaChecker(AbstractTableConfig tableConfig, TableSizeReader tabl
...
       @Nonnull String segmentName,
       @Nonnegative int timeoutMsec) {
-    Preconditions.checkNotNull(segmentFile);
-    Preconditions.checkNotNull(tableNameWithType);
-    Preconditions.checkNotNull(segmentName);
-    Preconditions.checkArgument(timeoutMsec > 0, "Timeout value must be > 0, input: %s", timeoutMsec);
-    Preconditions.checkArgument(segmentFile.exists(), "Segment file: %s does not exist", segmentFile);
-    Preconditions.checkArgument(segmentFile.isDirectory(), "Segment file: %s is not a directory", segmentFile);
+    try {
+      Preconditions.checkNotNull(segmentFile);
```
antumbde: These function params have @Nonnull annotations meaning they don't expect null values to be passed in. So, if a caller passes nulls that is an error. Add error messages to Preconditions if you want but for NPE there is little additional info to report.  These should definitely not be inside try-catch.

Caller should correctly handle these bad values returning appropriate error responses before calling this method.

jenniferdai: Okay, so for user errors, I should continue to allow the api to throw an exception? My thought process was that for user errors, we should always log infos; is that incorrect? I can remove the try catch for the null preconditions, but should I leave them for illegal argument exceptions?

antumbde: Log info for user errors but this method is not right place to check that. Caller should handle bad input values before calling this method.

Preconditions explicitly say throw exception if that pre-requisite is not met. So, there should not be try-catch block around it. If you want to change semantics, chage those to if-statements

PR #1292, opened 2017-04-13
[TE]FixSingleFunctionReplayUnNotified

cecilynie:
Make single function replay un-notified to send initial alert

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
@@ -755,7 +755,9 @@ public class DetectionJobResource {
   }
 
   /**
-   *
+   * Single function Reply to generate unnotified anomalies for user to label.
+   * Given anomaly function Id, start time, end time, it clones a function with same configurations and replays from start time to end time
+   * The replayed anomalies are being kept, with notified label = false
    * @param id FunctionId
    * @param replayStartTimeIso replay start time in ISO format, e.g. 2017-02-27T00:00:00.000Z
    * @param replayEndTimeIso replay end time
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
```diff
@@ -768,6 +770,7 @@ public Response replayAnomalyFunctionByFunctionId(@PathParam("functionId") long
     DateTime replayStart = ISODateTimeFormat.dateTimeParser().parseDateTime(replayStartTimeIso);
     DateTime replayEnd = ISODateTimeFormat.dateTimeParser().parseDateTime(replayEndTimeIso);
     FunctionReplayRunnable functionReplayRunnable = new FunctionReplayRunnable(detectionJobScheduler, anomalyFunctionDAO, mergedAnomalyResultDAO, rawAnomalyResultDAO, new HashMap<String, String>(), id, replayStart, replayEnd, false);
+    functionReplayRunnable.setForceBackfill(false);
```
bl44: It is hard for me to understand how you setForceBackfill to 'false'  related to 'notified or unnotified'. Is it true that any generated anomalies is treated as 'notified label = false' by default. What the setForceBackfill really do?

cyenjung: Set forceBackfill to true would ignore any existing anomalies in the interval to be backfilled; otherwise, the system would abort backfilling to prevent unwanted backfilling.

I don't think changing this forceBackfilling would achieve what you want. You could make backfill function takes as input a notification flag and set notified variable to that flag. (Currently, the flag is always true.)

cecilynie: Thanks for your suggestions @bl44 and @cyenjung. I'll close this pr and create a new one to generate initial anomalies report. - i.e. That's our actual goal. :)

cecilynie: Issue is resolved in https://github.com/linkedin/pinot/pull/1297

PR #1300, opened 2017-04-14
Shorten hostnames in server response time stats logs

mcvsubbu:
Right now we get FQDNs in the logs while logging server response time. It is useful to shorten this to
just the leading part of the domain name, so as to reduce the log size.


pinot-transport/src/main/java/com/linkedin/pinot/transport/scattergather/ScatterGatherImpl.java
```diff
@@ -166,17 +166,17 @@ public ScatterGatherImpl(KeyedPool<ServerInstance, PooledNettyClientResourceMana
           new ArrayList<KeyedFuture<ServerInstance, ByteBuf>>();
       for (SingleRequestHandler h : handlers) {
         responseFutures.add(h.getResponseFuture());
-        String serverName = h.getServer().toString();
+        String shortServerName = h.getServer().getShortHostName();t commit
```
Jackie-Jiang: Fix this line

PR #1304, opened 2017-04-15
Fixing service status to return GOOD after current state matches with…

kishoreg:
… Idealstate

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
@@ -18,12 +18,22 @@ package com.linkedin.pinot.common.utils;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+
 import org.apache.helix.HelixAdmin;
+import org.apache.helix.HelixDataAccessor;
+import org.apache.helix.HelixManager;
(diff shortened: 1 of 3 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
```diff
@@ -63,6 +71,106 @@ public static Status getServiceStatus() {
...
+
+    public IdealStateAndCurrentStateMatchServiceStatusCallback(HelixManager helixManager, String clusterName, String instanceName,
+        List<String> resourcesToMonitor) {
+      _helixManager = helixManager;
+      _clusterName = clusterName;
+      _instanceName = instanceName;
+      _helixAdmin = _helixManager.getClusterManagmentTool();
+      _resourcesToMonitor = resourcesToMonitor;
+
+      LOGGER.info("Monitoring resources {} for start up of instance {}", _resourcesToMonitor, _instanceName);
```
mcvsubbu: if there are 100s of tables hosting partitions on a server, will this log all the table names? Would it be better to log this as debug, or just log the number of reosurces that need to be monitored?

kishoreg: This happens only at start up and should log only once when the ServiceStatus is inited.

PR #1305, opened 2017-04-15
[TE] Root-cause search framework prototype

apucher:
* root-cause search framework

* search pipelines

* sample implementation

* dataframe extensions


apucher:
split into two parts

PR #1307, opened 2017-04-17
Removed parameterization of K for KeyedFuture interface.

mcvsubbu:
We always use this with ServerInstance (other than for tests), so hard-coded ServerInstance in to make
things easier while navigating. We can also use the interfaces exposed by ServerInstance specifically
in the Future implementations now. (coming in another PR)

### DIFF
--- a/pinot-transport/src/main/java/com/linkedin/pinot/transport/common/KeyedFuture.java
+++ b/pinot-transport/src/main/java/com/linkedin/pinot/transport/common/KeyedFuture.java
@@ -22,6 +22,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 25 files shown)


pinot-transport/src/main/java/com/linkedin/pinot/transport/common/KeyedFuture.java
```diff
@@ -39,10 +40,9 @@
  * request dispatch with speculative (duplicate) request dispatching.
  *
  *
- * @param <K> Key type to identify a result/error
  * @param <V> Response Type
  */
-public interface KeyedFuture<K, V> extends ListenableFuture<Map<K, V>> {
+public interface KeyedFuture<V> extends ListenableFuture<Map<ServerInstance, V>> {
```
antumbde: I think this should be generic <k,v> given the name KeyedFuture. It can have wider use.

Better option is to have a thin derived impl ServerResponseFuture<V> (or something similar) that use serverInstance  as key

PR #1310, opened 2017-04-17
Unify Performance Evaluation Method in ThirdEye

cyy0714:
1. Move AlertFilterEvaluationUtil to PrecisionRecallEvaluator
2. Provide a precision and recall evaluator to all
3. Change all usage of precision recall calculation to this unified method


thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/detector/email/filter/TestAlertFilterUtil.java
```diff
@@ -15,27 +15,26 @@
   @Test
   public void testPrecisionAndRecall() throws Exception{
     AlertFilter dummyAlertFilter = new DummyAlertFilter();
-    AlertFilterEvaluationUtil evaluator = new AlertFilterEvaluationUtil(dummyAlertFilter);
     // test data with 1 positive feedback, 1 negative feedback, other NA feedbacks
     List<MergedAnomalyResultDTO> anomalies = getMockMergedAnomalies(7,8);
```
kishoreg: We should probably rename the test class since the original class name is renamed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/PerformanceEvaluationUtil.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.List;
+import java.util.Properties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Utility class to evaluate the performance of a list of merged anomalies
+ */
+public class PerformanceEvaluationUtil {
```
kishoreg: The name of the class is too generic. Why not call it XXXEvaluator. Util classes generally have a bunch of static methods.

What is the purpose of this class

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/detector/email/filter/TestAlertFilterUtil.java
cyy0714: Good idea. I will change the name.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/PerformanceEvaluationUtil.java
cyy0714: That's a good catch. I'll rename the class.
This is to provide a class to evaluate precision and recall in ThirdEye. There are two versions of precision and recall calculation in ThirdEye: one is in AlertFilterEvaluationUtils, and the other is in EmailReportGenerator. I want to have a tool to calculate the precision and recall for a past time region, but I cannot use either of them. Therefore, I want to unify the calculation of precision and recall.

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/detector/email/filter/TestPrecisionRecallEvaluator.java
```diff
@@ -10,32 +10,31 @@
...
 
     // test data with 0 positive feedback, 1 negative feedback and others are NA feedbacks
-    anomalies = getMockMergedAnomalies(-1,6);
-    try{
-      evaluator.updateWeighedPrecisionAndRecall(anomalies);
-      fail("Should throw exception");
-    } catch (Exception e) {
-      assertEquals("No true labels in dataset. Check data", e.getMessage());
-    }
+//    anomalies = getMockMergedAnomalies(-1,6);
```
cyenjung: Remove the dead code or put comment to explain why this section of test is disabled.

cyy0714: Sure. Thanks

PR #1311, opened 2017-04-17
Shorten host names in response time logs

mcvsubbu:


mcvsubbu:
@kishoreg the hashcode/equals methods and their contents are as they are before. They just surfaced because I changed KeyedFuture<K, V> to KeyedFuture<V> (and hard-coded K to ServerInstance, since we never use it any other way).

See https://github.com/linkedin/pinot/pull/1307

I will fix the hashcode and equals methods

mcvsubbu:
@kishoreg like we discussed it seems the equals and hashcode methods are both present in ServerInstance. Also, they take into account the port.

PR #1313, opened 2017-04-17
Implement methods for self-onboard daily detection function

cyy0714:
1. Enable FunctionReplayRunnable and DetectionJobResource not to include original performance. If the evaluation is based on offline analysis, the performance is not accurate.
2. Implement an endpoint to apply autotune_config to original function or a cloned function.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
@@ -569,6 +569,10 @@ public class DetectionJobResource {
    * ex: {"baselineLift": [0.9, 0.95, 1, 1.05, 1.1], "baselineSeasonalPeriod": [2, 3, 4]}
    * @param goal
    * the expected performance assigned by user
+   * @param includeOrigin
+   * to include the performance of original setup into comparison
+   * If we perform offline analysis before hand, we don't get the correct performance about the current configuration
+   * setup. Therefore, we need to exclude the performance from the comparison.
    * @return
    * A response containing all satisfied properties with their evaluation result
    */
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
```diff
@@ -591,6 +592,7 @@ public Response anomalyFunctionReplay(@PathParam("id") @NotNull long functionId,
       @QueryParam("speedup") @DefaultValue("true") boolean speedup,
       @QueryParam("tune") @DefaultValue("{\"pValueThreshold\":[0.05, 0.01]}") String tuningJSON,
       @QueryParam("goal") @DefaultValue("0.05") double goal,
+      @QueryParam("includeOriginal") @DefaultValue("true") boolean includeOrigin,
```
cecilynie: Can you be more explicit on what does "includeOrigin" do?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -440,6 +444,50 @@ public Response updateAnomalyFunction(@NotNull @QueryParam("id") Long id,
...
+    }
+
+    // Update function configuration
+    targetFunction.updateProperties(autotuneConfigDTO.getConfiguration());
+    targetFunction.setActive(true);
+    anomalyFunctionDAO.update(targetFunction);
+
+    // Dis-activate original function
+    if(isCloneFunction) {
+      originalFunction.setActive(false);
```
cecilynie: How do we deal with the anomalies from original function? It varies with the usecase:
1. The cloned function is for replay and replace original function. You don't care about the anomalies from original function. You can erase the historical anomalies from original function
2. If original function already have labeled anomalies, then you should always have "isCloneAnomalies = true"

PR #1315, opened 2017-04-18
Fix the problem of int overflow when trying to generate inverted index on huge number of entries

Jackie-Jiang:
Added checks before generating inverted index.
If the check failed, log a warning and skip creating the inverted index.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -167,14 +167,38 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
     }
 
     for (String column : config.getInvertedIndexCreationColumns()) {
-      if (!schema.hasColumn(column)) {
-        LOGGER.warn("Skipping enabling index on column:{} since its missing in schema", column);
+      // Validation check
+      FieldSpec fieldSpec = schema.getFieldSpecFor(column);
+      if (fieldSpec == null) {
+        LOGGER.warn("Skip creating inverted index for segment: {}, column: {} because column does not exist in schema",
+            segmentName, column);
+        continue;
+      }
+      if (totalDocs > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES) {
+        LOGGER.warn(
+            "Skip creating inverted index for segment: {}, column: {} because totalDocs: {} exceeds the limit: {}",
+            segmentName, column, totalDocs, OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES);
         continue;
       }
       ColumnIndexCreationInfo indexCreationInfo = indexCreationInfoMap.get(column);
-      int uniqueValueCount = indexCreationInfo.getDistinctValueCount();
-      OffHeapBitmapInvertedIndexCreator invertedIndexCreator = new OffHeapBitmapInvertedIndexCreator(file,
-          uniqueValueCount, totalDocs, indexCreationInfo.getTotalNumberOfEntries(), schema.getFieldSpecFor(column));
+      int cardinality = indexCreationInfo.getDistinctValueCount();
+      if (cardinality > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES) {
+        LOGGER.warn(
+            "Skip creating inverted index for segment: {}, column: {} because cardinality: {} exceeds the limit: {}",
+            segmentName, column, cardinality, OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES);
+        continue;
+      }
+      int totalNumberOfEntries = indexCreationInfo.getTotalNumberOfEntries();
+      if ((!fieldSpec.isSingleValueField())
+          && (totalNumberOfEntries > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES)) {
+        LOGGER.warn(
+            "Skip creating inverted index for segment: {}, multi-value column: {} because totalNumberOfEntries: {} exceeds the limit: {}",
+            segmentName, column, totalNumberOfEntries, OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES);
+        continue;
+      }
+
+      OffHeapBitmapInvertedIndexCreator invertedIndexCreator =
+          new OffHeapBitmapInvertedIndexCreator(file, cardinality, totalDocs, totalNumberOfEntries, fieldSpec);
       invertedIndexCreatorMap.put(column, invertedIndexCreator);
     }
   }
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
@@ -71,6 +71,11 @@ public class OffHeapBitmapInvertedIndexCreator implements InvertedIndexCreator {
   private static final Logger LOGGER =
       LoggerFactory.getLogger(OffHeapBitmapInvertedIndexCreator.class);
 
+  // Number of bytes required to store an integer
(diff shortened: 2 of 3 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -167,14 +167,38 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
     }
 
     for (String column : config.getInvertedIndexCreationColumns()) {
-      if (!schema.hasColumn(column)) {
-        LOGGER.warn("Skipping enabling index on column:{} since its missing in schema", column);
+      // Validation check
+      FieldSpec fieldSpec = schema.getFieldSpecFor(column);
+      if (fieldSpec == null) {
+        LOGGER.warn("Skip creating inverted index for segment: {}, column: {} because column does not exist in schema",
```
mcvsubbu: is it possible to log the segment or schema name?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -167,14 +167,38 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
...
+            segmentName, column);
+        continue;
+      }
+      if (totalDocs > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES) {
+        LOGGER.warn(
```
mcvsubbu: pls log the segment or schema name

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -167,14 +167,38 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
...
         continue;
       }
       ColumnIndexCreationInfo indexCreationInfo = indexCreationInfoMap.get(column);
-      int uniqueValueCount = indexCreationInfo.getDistinctValueCount();
-      OffHeapBitmapInvertedIndexCreator invertedIndexCreator = new OffHeapBitmapInvertedIndexCreator(file,
-          uniqueValueCount, totalDocs, indexCreationInfo.getTotalNumberOfEntries(), schema.getFieldSpecFor(column));
+      int cardinality = indexCreationInfo.getDistinctValueCount();
+      if (cardinality > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES) {
+        LOGGER.warn(
+            "Skip creating inverted index for segment: {}, column: {} because cardinality: {} exceeds the limit: {}",
```
mcvsubbu: pls log the segment or schema name

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -167,14 +167,38 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
...
+            segmentName, column, cardinality, OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES);
+        continue;
+      }
+      int totalNumberOfEntries = indexCreationInfo.getTotalNumberOfEntries();
+      if ((!fieldSpec.isSingleValueField())
+          && (totalNumberOfEntries > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES)) {
+        LOGGER.warn(
```
mcvsubbu: pls log the segment or schema name

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
Jackie-Jiang: We already logged the segment name.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
```diff
@@ -91,7 +91,31 @@ public void createInvertedIndices()
...
     String column = columnMetadata.getColumnName();
+    int totalDocs = columnMetadata.getTotalDocs();
+    if (totalDocs > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES) {
+      LOGGER.warn(
+          "Skip creating inverted index for segment: {}, column: {} because totalDocs: {} exceeds the limit: {}",
+          segmentName, column, totalDocs, OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES);
+      return;
+    }
+    int cardinality = columnMetadata.getCardinality();
+    if (cardinality > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES) {
```
antumbde: Duplicated code logic. Move this to OffheapBitmapInvertedIndexCreator.isConstructionPossible(cardinality, totalDocs)  static method. That will keep all the logic in one place

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -71,6 +71,11 @@
   private static final Logger LOGGER =
       LoggerFactory.getLogger(OffHeapBitmapInvertedIndexCreator.class);
 
+  // Number of bytes required to store an integer
+  private static final int INT_SIZE = Integer.SIZE / Byte.SIZE;
+  // Maximum number of entries for each buffer
+  public static final int MAX_NUM_ENTRIES = Integer.MAX_VALUE / INT_SIZE;
```
antumbde: This is not accounting the size correctly. Each entry takes Integer.size + bitmap serialization size. Bitmap serialization size is unpredictable but it's atleast 1 byte but most likely more. So, say, atleast 5 bytes per entry.
So, max entries can not exceed (Integer.max_value / (int_size + 1) - 1). For -1 see class comment

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
Jackie-Jiang: I did it this way for minor performance gain (which is definitely unnecessary :-P), will address this in a separate PR

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
Jackie-Jiang: The actual size calculation is trickier. For multi-value columns, we allow this number of entries; for number of total docs, we allow this number of total docs. But we don't allow this high cardinality, which is what actually stored in the final bitmap file.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
```diff
@@ -91,7 +91,31 @@ public void createInvertedIndices()
 
   private void createInvertedIndexForColumn(ColumnMetadata columnMetadata)
       throws IOException {
+    // Validation check
     String column = columnMetadata.getColumnName();
+    int totalDocs = columnMetadata.getTotalDocs();
+    if (totalDocs > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES) {
+      LOGGER.warn(
```
Jackie-Jiang: I agree it's hard to track this performance problem if inverted indexes are not created. Actually adding a counter (metric) here does not help that much, and it's not very possible to pass the ServerMeter all the way down to this method. Should we just throw the exception, and let CRC check catch this?

PR #1316, opened 2017-04-18
Throw exception when failed to load segment in both retry and no retry case

Jackie-Jiang:
Note: Helix message itself does not have default retry unless explicitly set.
For customized message, will trigger the onError() callback when handleMessage() throws exception.

### DIFF
--- a/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
+++ b/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
@@ -201,11 +201,10 @@ public class SegmentFetcherAndLoader {
             }
           }
         }
-        if (_segmentLoadMaxRetryCount <= retryCount) {
-          String msg =
-              "Failed to download segment " + segmentId + " (table " + tableName + " after " + retryCount + " retries";
-          LOGGER.error(msg);
-          throw new RuntimeException(msg);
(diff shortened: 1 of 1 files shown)


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
```diff
@@ -201,11 +201,10 @@ public void addOrReplaceOfflineSegment(String tableName, String segmentId, boole
             }
           }
         }
-        if (_segmentLoadMaxRetryCount <= retryCount) {
-          String msg =
-              "Failed to download segment " + segmentId + " (table " + tableName + " after " + retryCount + " retries";
-          LOGGER.error(msg);
-          throw new RuntimeException(msg);
+        if (retryCount == maxRetryCount) {
```
mcvsubbu: I suppose this logic applies to whether we are throwing an exception during segment refresh vs. segment load. In the latter case, we want to retry multiple times (preferably retry only if download fails, not if load fails). 

So, can we try to name the variables accordingly? Is this a download or a refresh? Is this a network failure or a load failure? It seems that if segments fail to load, it does not make sense to retry.

Jackie-Jiang: Will address that in separate PR. This one is just to fix the problem where reload failure does not log any error.

PR #1319, opened 2017-04-18
Add ability to read from external pipelines

npawar:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/PipelineUtils.java
```diff
@@ -0,0 +1,42 @@
...
+      props.load(input);
+    } catch (Exception e) {
+      LOG.error("Exception in loading pipelines file", e);
+    }
+
+    for (Entry<Object, Object> entry : props.entrySet()) {
+      try {
+        String className = (String) entry.getValue();
+        Constructor<?> constructor = Class.forName(className).getConstructor();
+        Pipeline pipeline = (Pipeline) constructor.newInstance();
```
kishoreg: the pipelines might need additional configuration. It will be good to come up with a standardized way to pass in the configuration needed for each pipeline.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/PipelineUtils.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Map.Entry;
+import java.util.Properties;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.linkedin.thirdeye.common.ThirdEyeConfiguration;
+import com.linkedin.thirdeye.rootcause.Pipeline;
+
+public class PipelineUtils {
```
kishoreg: rename to PipelineInstantiator or PipelineLoader - anything other than utils.

PR #1324, opened 2017-04-19
[TE] Add alert grouping logic

cyenjung:
1. Add a grouping logic to alerter in order to group anomalies that happen in the same time frame. 

2. Implement a dimensional grouper to group anomalies by user-specified dimensions. Moreover, each group  could have a list of auxiliary email recipients.

Integration test on local setup.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/alert/grouping/AlertGrouperFactory.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/alert/grouping/AlertGrouperFactory.java
@@ -29,7 +29,7 @@ public class AlertGrouperFactory {
     }
 
     GrouperType filterType = GrouperType.DUMMY;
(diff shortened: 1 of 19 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/alert/grouping/AlertGrouperFactory.java
```diff
@@ -0,0 +1,48 @@
...
+    return alertGrouper;
+  }
+
+  private static AlertGrouper fromStringType(String type) {
+    if (StringUtils.isBlank(type)) { // backward compatibility
+      return DUMMY_ALERT_GROUPER;
+    }
+
+    GrouperType filterType = GrouperType.DUMMY;
+    for (GrouperType enumFilterType : GrouperType.class.getEnumConstants()) {
```
npawar: Use GrouperType.values() instead of class.getEnumConstants() ?

PR #1333, opened 2017-04-20
[TE] fix pagination ajax call

justYves:
### Bug Fix: 
Fixing ajax calls for multiple page search where page 2+ would strip the metricName/Time filters.


justYves:
@cyy0714 yes, those are cached by the HASH_SERVICE

PR #1339, opened 2017-04-21
Add support for Star Tree realtime segment generation.

mayankshriv:
1. Added support in IndexingConfig to specify StarTreeIndexSpec.
2. Added code that picks up star tree index spec from IndexingConfig
   and passes it to segment generation.
3. Only enabled for LLC, as HLC will be phased out.
4. Added unit test for serialization/deserialization of star tree
   spec using IndexingConfig.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -21,6 +21,7 @@ import com.google.common.util.concurrent.Uninterruptibles;
 import com.linkedin.pinot.common.config.AbstractTableConfig;
 import com.linkedin.pinot.common.config.IndexingConfig;
 import com.linkedin.pinot.common.data.Schema;
+import com.linkedin.pinot.common.data.StarTreeIndexSpec;
 import com.linkedin.pinot.common.metadata.instance.InstanceZKMetadata;
 import com.linkedin.pinot.common.metadata.segment.LLCRealtimeSegmentZKMetadata;
 import com.linkedin.pinot.common.metadata.segment.RealtimeSegmentZKMetadata;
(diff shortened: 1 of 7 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -533,7 +535,7 @@ private boolean buildSegmentInternal(boolean forCommit) {
     RealtimeSegmentConverter converter =
         new RealtimeSegmentConverter(_realtimeSegment, tempSegmentFolder.getAbsolutePath(), _schema,
             _segmentZKMetadata.getTableName(), _segmentZKMetadata.getSegmentName(), _sortedColumn,
-            _invertedIndexColumns, _noDictionaryColumns);
+            _invertedIndexColumns, _noDictionaryColumns, _starTreeIndexSpec);
```
kishoreg: this constructor has become quite big. We need to clean up this class - it can be part of another PR

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -695,4 +698,12 @@ public void setSegmentPartitionConfig(SegmentPartitionConfig segmentPartitionCon
   public SegmentPartitionConfig getSegmentPartitionConfig() {
     return segmentPartitionConfig;
   }
+
+  public void setStarTreeIndexSpec(StarTreeIndexSpec starTreeIndexSpec) {
```
mcvsubbu: the methods to set and get start tree index spec in this class don't seem to be used. Or, did I read it wrong?

mayankshriv: You are right, by the time the code was implemented these methods ended up being unused. I have removed them.

PR #1340, opened 2017-04-23
Autometrics config setup

npawar:
Creating a new table called autometrics config for storing autometrics. This will replace the ingraph_dashboard and ingraph_metrics tables. Created a new one instead of modifying the existing ones, so that this jar can be used in the webapp MP during development, without disturbing the existing setup, and also because the metrics store is actually called autometrics.
This is the table which will be used by the Autometrics client, to get additional information about the metrics it has identified to be from autometrics (which will happen via dataset config flag, or by checking presence of dataset and metric name in autometrics table)

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/AutometricsConfigBean.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/AutometricsConfigBean.java
@@ -1,0 +1,121 @@
+package com.linkedin.thirdeye.datalayer.pojo;
+
+import java.util.Objects;
+
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+
+/**
+ * Class to represent the autometric definition in thirdeye database
+ */
+public class AutometricsConfigBean extends AbstractBean {
+
+  /** Metric name as in MetricConfig **/
+  private String metric;
+  /** Dashboard name in external dashboard, and dataset name in datasetConfig **/
+  private String dashboard;
+  private String rrd;
+  private String metricDataType;
+  /**  COUNTER or GAUGE **/
+  private String autometricsType;
+  private String container;
+  private String fabricGroup;
+  private boolean active = true;
+
+  public AutometricsConfigBean() {
+  }
+
+  public String getMetric() {
+    return metric;
+  }
+
+  public void setMetric(String metric) {
+    this.metric = metric;
+  }
+
+  public String getDashboard() {
+    return dashboard;
+  }
+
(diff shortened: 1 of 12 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/AutometricsConfigBean.java
```diff
@@ -0,0 +1,121 @@
...
+        && Objects.equals(getRrd(), ac.getRrd())
+        && Objects.equals(getMetricDataType(), ac.getMetricDataType())
+        && Objects.equals(getAutometricsType(), ac.getAutometricsType())
+        && Objects.equals(getContainer(), ac.getContainer())
+        && Objects.equals(getFabricGroup(), ac.getFabricGroup())
+        && Objects.equals(isActive(), ac.isActive());
+  }
+
+  @Override
+  public String toString() {
```
cyenjung: The same toString method is defined in its parent class: AbstractDTO.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/AutometricsConfigBean.java
```diff
@@ -0,0 +1,121 @@
...
+public class AutometricsConfigBean extends AbstractBean {
+
+  /** Metric name as in MetricConfig **/
+  private String metric;
+  /** Dashboard name in external dashboard, and dataset name in datasetConfig **/
+  private String dashboard;
+  private String rrd;
+  private String metricDataType;
+  /**  COUNTER or GAUGE **/
+  private String autometricsType;
```
cyenjung: Should we use enum class instead of String for types?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/AutometricsConfigBean.java
npawar: Thanks for pointing that out!

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/AutometricsConfigBean.java
npawar: I did not add enum here, because that would mean putting autometrics specific logic in open source

cyenjung: I see. Then leave it String type should be fine.

PR #1343, opened 2017-04-24
Derive table name from input segment for ColumnarToStarTreeConverter.

mayankshriv:
The tool to convert columnar to star tree format was not writing out the
table name in the metadata. Fixed it by reading the table name from
input segment, and setting it into output segment.

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/ColumnarToStarTreeConverter.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/ColumnarToStarTreeConverter.java
@@ -15,12 +15,16 @@
  */
 package com.linkedin.pinot.tools.segment.converter;
 
+import com.linkedin.pinot.common.segment.SegmentMetadata;
 import com.linkedin.pinot.common.utils.TarGzCompressionUtils;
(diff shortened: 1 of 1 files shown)


mayankshriv:
@Jackie-Jiang Good catch, updated PR with the change to load segment name from metadata as well.

pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/ColumnarToStarTreeConverter.java
```diff
@@ -113,7 +115,11 @@ private void convertSegment(File columnarSegment)
     config.setOutDir(_outputDirName);
     config.setStarTreeIndexSpecFile(_starTreeConfigFileName);
     config.setOverwrite(_overwrite);
-    config.setSegmentName(columnarSegment.getName());
+
+    // Read the segment and table name from the segment's metadata.
```
mcvsubbu: Please use segmentNameGenerator instead of setSegmentName. that is the direction in whcih we are evloving

mayankshriv: This tool is for existing segments for which we want to preserve the input segments name. Also refresh breaks if segment names are different.

PR #1353, opened 2017-04-25
Refine performance evaluation method

cyy0714:
1. Precision and recall calculation is now based on the top metric. The original calculation take the average of downstram metrics.

2. Function replay takes the time and duration instead of a fix start and end time.

3. ReplayRunnable will clone anomalies before the monitoring window.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
@@ -7,6 +7,7 @@ import com.linkedin.thirdeye.anomalydetection.alertFilterAutotune.AlertFilterAut
 import com.linkedin.thirdeye.anomalydetection.alertFilterAutotune.AlertFilterAutotuneFactory;
 import com.linkedin.thirdeye.anomalydetection.performanceEvaluation.PerformanceEvaluateHelper;
 import com.linkedin.thirdeye.anomalydetection.performanceEvaluation.PerformanceEvaluationMethod;
+import com.linkedin.thirdeye.api.TimeGranularity;
 import com.linkedin.thirdeye.client.ThirdEyeCacheRegistry;
 import com.linkedin.thirdeye.client.ThirdEyeClient;
 import com.linkedin.thirdeye.datalayer.bao.AutotuneConfigManager;
@@ -560,8 +561,10 @@ public class DetectionJobResource {
    * the id of the target anomaly function
    * @param replayStartTimeIso
    * the start time of the anomaly function replay in ISO format, e.g. 2017-02-27T00:00:00.000Z
-   * @param replayEndTimeIso
-   * the end time of the anomaly function replay in ISO format
+   * @param replayDuration
+   * the duration of the replay ahead of the replayStartTimeIso
+   * @param durationUnit
+   * the time unit of the duration, DAYS, HOURS, MINUTES and so on
    * @param speedup
    * whether we speedup the replay process
    * @param tuningJSON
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
```diff
@@ -579,20 +582,34 @@ public Response evaluateAlertFilterByAutoTuneId(@PathParam("autotuneId") long id
   @POST
   @Path("replay/function/{id}")
   public Response anomalyFunctionReplay(@PathParam("id") @NotNull long functionId,
-      @QueryParam("start") @NotNull String replayStartTimeIso, @QueryParam("end") @NotNull String replayEndTimeIso,
+      @QueryParam("time") String replayTimeIso, @QueryParam("duration") @DefaultValue("30") int replayDuration,
```
bl44: Does this mean that you will fix the end to now? This is less flexible than the original one. Why making this choice?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/performanceEvaluation/PrecisionByTimePerformanceEvaluation.java
```diff
@@ -52,13 +52,15 @@ public double evaluate() {
...
-      }
-      avgPrecision += (double) dimensionToOverlapTimeLength.get(dimensions) / dimensionToDetectedAnomalyTimeLength.get(dimensions);
+      totalDetectedAnomalyTimeLength += dimensionToDetectedAnomalyTimeLength.get(dimensions);
+      totalDimensionOverlapTimeLength += dimensionToOverlapTimeLength.get(dimensions);
+    }
+    if (totalDetectedAnomalyTimeLength == 0) {
+      return Double.NaN;
     }
-    return avgPrecision / ((double) dimensionToDetectedAnomalyTimeLength.size());
+    return (double) totalDimensionOverlapTimeLength / totalDetectedAnomalyTimeLength;
```
bl44: Is your definition of F1 score need to change too?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/lib/FunctionReplayRunnable.java
```diff
@@ -64,7 +74,6 @@ public FunctionReplayRunnable(DetectionJobScheduler detectionJobScheduler, Anoma
     setForceBackfill(isForceBackfill);
     setTuningParameter(tuningParameter);
     setFunctionAutotuneConfigId(functionAutotuneConfigId);
-    setSpeedUp(true);
```
bl44: So there will be no SpeedUp?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
cyy0714: No. User can assign different start time in its query. My modification here is to set-up a default start time.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/performanceEvaluation/PrecisionByTimePerformanceEvaluation.java
cyy0714: It is not required. The definition of F1 score is not changed. It is a different calculation strategy among different sub-metrics.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/lib/FunctionReplayRunnable.java
cyy0714: The default speedup value is set in the base constructor. This value here is a duplicate.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
```diff
@@ -579,20 +582,34 @@ public Response evaluateAlertFilterByAutoTuneId(@PathParam("autotuneId") long id
...
+      @QueryParam("durationUnit") String durationUnit,
```
cecilynie: If we have a default duration value, why we don't have a default duration unit?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/lib/FunctionReplayRunnable.java
```diff
@@ -118,22 +127,23 @@ public void run() {
       functionName.append(entry.getValue());
     }
     try {
-      clonedFunctionId = onboardResource.cloneAnomalyFunctionById(tuningFunctionId, functionName.toString(), false);
-      this.lastClonedFunctionId = clonedFunctionId;
+      clonedFunctionId = onboardResource.cloneAnomalyFunctionById(tuningFunctionId, functionName.toString(), cloneAnomaly);
+      this.clonedFunctionId = clonedFunctionId;
```
cecilynie: The clonedFunctionId will always be overrided to the new clonedFunctionId, and in the end you only keep the very last cloned function Id. Shall we keep the name of "lastClonedFunctionId" or some what change to a better name to specify this logic?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/lib/FunctionReplayRunnable.java
```diff
@@ -176,14 +188,16 @@ public void run() {
 
     // clean up and kill itself
     if(selfKill) {
-      onboardResource.deleteExistingAnomalies(Long.toString(clonedFunctionId), replayStart.getMillis(),
+      onboardResource.deleteExistingAnomalies(Long.toString(clonedFunctionId), 0,
           replayEnd.getMillis());
       anomalyFunctionDAO.deleteById(clonedFunctionId);
```
cecilynie: If is selfKipp, clonedFunctionId is also being deleted. What about the clonedFunctionId contained in Runnable?

PR #1354, opened 2017-04-26
Change single-line whitelist to multilines

cyenjung:
1. Change single-lined whitelist to multilined list to easy of management.
2. Remove inappropriate testing code.

Tested on local dashboard and controllers.


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/common/ThirdEyeConfiguration.java
```diff
@@ -21,8 +22,8 @@
   private String configStoreType = "FILE";
   private String implMode = "hibernate";
 
-  private String whitelistCollections = "";
-  private String blacklistCollections = "";
+  private List<String> whitelistCollections;
+  private List<String> blacklistCollections;
```
npawar: Does no default value mean we always have to give blacklistCollections as well?

cyenjung: Fixed! Good catch.

PR #1345, opened 2017-04-24
[TE] rootcause DAG pipeline

apucher:
* DAG execution engine for rootcause search

* unification of aggregator and pipeline

    * partial rewrite of existing pipelines

* dimension rewriter pipeline

* topk pipeline

* unit tests

PR #1356, opened 2017-04-26
[TE] Rootcause DAG framework

apucher:
* Clean split of DAG framework and implementation of #1345

* only RCA framework


apucher:
again, breaking into pieces

PR #1364, opened 2017-04-27
Fix segment status checker to report the best available number when no segments are present

mcvsubbu:
When 0 segments are present (e.g. a table is provisioned), SegmentStatusChecker reports 1
as the number of replicas. This may raise alerts if the number of replcias on the table
is expected to be higher than 1.

### DIFF
--- a/pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/SegmentStatusCheckerTest.java
+++ b/pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/SegmentStatusCheckerTest.java
@@ -15,12 +15,6 @@
  */
 package com.linkedin.pinot.controller.helix;
 
-import com.linkedin.pinot.common.metrics.ControllerGauge;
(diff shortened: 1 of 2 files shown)


pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/SegmentStatusCheckerTest.java
```diff
@@ -470,10 +470,11 @@ public void noIdealState() throws Exception {
   @Test
   public void noSegments() throws Exception {
     final String tableName = "myTable_REALTIME";
+    final int nReplicas = 5;
     List<String> allTableNames = new ArrayList<String>();
     allTableNames.add(tableName);
     IdealState idealState = new IdealState(tableName);
-    idealState.setReplicas("0");
+    idealState.setReplicas(Integer.toString(nReplicas));
```
rven1: Would also like to test for the case when replicas is not set or is 0

PR #1365, opened 2017-04-27
[TE] onboarding add endpoints to activate or deactivate anomaly detection function in batch

bl44:
[te] onboarding add endpoints to adtivate or deactive anomaly detection function in batch

Query Parameters: functionIds=ALL means all functions, "1,3,5" means activate/deactivate function with id 1, 3, 5


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/OnboardResource.java
```diff
@@ -47,20 +41,64 @@ public OnboardResource() {
...
+        dto.setActive(tag);
+        anomalyFunctionDAO.update(dto);
+      }
+      return;  // early return
+    }
+
+    if (StringUtils.isNotBlank(functionIds)) {
+      String[] tokens = functionIds.split(",");
+      for (String token : tokens) {
+        functionIdsList.add(Long.valueOf(token));
```
cyenjung: Handle NumberFormatException for Long.valueOf.

If we want to stop the entire procedure whenever there is an exception, then please add a comment to mention that the behavior (unhandled exception) is expected.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/OnboardResource.java
```diff
@@ -47,20 +41,64 @@ public OnboardResource() {
...
+      }
+    }
+    for (long id:functionIdsList) {
```
cyenjung: Add space in front of and after colon.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/OnboardResource.java
```diff
@@ -47,20 +41,64 @@ public OnboardResource() {
...
+    return functionIds;
+  }
+
+  /**
+   * toggle anomaly functions to active and inactive
+   *
+   * @param functionIds string comma separated function ids, if
+   * @param tag boolean true or false, set function as true or false
+   */
+  private void toggleFunctions(String functionIds, boolean tag) {
```
bl44: Sure.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/OnboardResource.java
```diff
@@ -47,20 +41,64 @@ public OnboardResource() {
...
+    List<Long> functionIdsList = new ArrayList<>();
+    if (functionIds.equals("ALL")) {
```
bl44: Agree that we do need some level of access control for using such endpoints (which I assume is a common problem for many other products). Put it in endpoints make our experiment so much easier that I do want to have some of this endpoints. I will change the magic world 'ALL' for now, we can think of better way for security.

bl44: During the experiment process I found myself using these activate/deactivate function so many times. Its essentially just of way communicating with the application database. Java do not support good 'cli' tools that's why I prefer endpoints over directly invoke the toggle function.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/OnboardResource.java
bl44: fixed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
```diff
@@ -748,23 +748,26 @@ public Response anomalyFunctionReplay(@PathParam("id") @NotNull long functionId,
...
    */
   @POST
   @Path("/replay/singlefunction")
   public Response replayAnomalyFunctionByFunctionId(@QueryParam("functionId") Long functionId,
       @QueryParam("autotuneId") Long autotuneId,
-      @QueryParam("start") @NotNull String replayStartTimeIso, @QueryParam("end") @NotNull  String replayEndTimeIso)
-       {
+      @QueryParam("start") @NotNull String replayStartTimeIso,
+      @QueryParam("end") @NotNull  String replayEndTimeIso,
+      @QueryParam("speedup") @DefaultValue("true") boolean speedup) {
```
cyenjung: Rename to speedUp? Because we use setSpeedUp at line 782.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/OnboardResource.java
cyenjung: I would suggest to move activeFunction, deactiveFunction, and toggleFunctions to RunAdhocDatabaseQueriesTool.java because these 3 functions should not belong to regular onboard procedure. There is a similar method, toggleAnomalyFunction(), in that file. You can extend it for your need.

We typically put experiment and DB related functions in RunAdhocDatabaseQueriesTool.java. There are two advantages: 1. we don't need to start controller in order to change the value in the database. 2. The modifiers to the database are not exposed.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/OnboardResource.java
```diff
@@ -47,20 +41,65 @@ public OnboardResource() {
...
 
   @GET
   @Path("function/{id}")
   public AnomalyFunctionDTO getAnomalyFunction(@PathParam("id") Long id) {
     return anomalyFunctionDAO.findById(id);
   }
 
+  @POST
+  @Path("function/activate")
+  public String activateFunction(@QueryParam("functionIds") String functionIds) {
```
cyy0714: The activate function is implemented in AnomalyResource.activateAnomalyFunction(). Should we modify that one?

PR #1368, opened 2017-04-27
Fix segment deletion manager to handle deleted directory correctly

snleee:
When deleted segment directory does not exist, segment deletion manager throws the NullPointerException because it does not check if the deleted directory path exists.
- Added check for deleted directory path
- Added unit test for checking when deleted segment directory does not exist


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
```diff
@@ -205,21 +205,32 @@ protected void removeSegmentFromStore(String tableName, String segmentId) {
    */
   public void removeAgedDeletedSegments(int retentionInDays) {
     if (_localDiskDir != null) {
-      File targetDir = new File(_localDiskDir, DELETED_SEGMENTS);
+      File deletedDir = new File(_localDiskDir, DELETED_SEGMENTS);
+      // Check that the directory for deleted segments exists
+      if (!deletedDir.exists() || !deletedDir.isDirectory()) {
```
Jackie-Jiang: No need to call exists() because isDirectory() will check if it exists.

PR #1370, opened 2017-04-28
[TE] robustness for dimension analysis, URN parsing and pipeline defaults

apucher:
* improve robustness and performance for dimension analysis (do less work)

* improve parsing of entity URNs (ignore trailing double colons)

* sensible defaults for pipeline properties, or explicit error message


apucher:
* improve robustness and performance for dimension analysis (do less work)

* improve parsing of entity URNs (ignore trailing double colons)

* sensible defaults for pipeline properties, or explicit error message

PR #1372, opened 2017-04-28
[TE] Modify single function replay runnable to threading

cecilynie:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionJobResource.java
```diff
@@ -739,7 +739,8 @@ public Response anomalyFunctionReplay(@PathParam("id") @NotNull long functionId,
   @Path("/replay/singlefunction")
   public Response replayAnomalyFunctionByFunctionId(@QueryParam("functionId") Long functionId,
       @QueryParam("autotuneId") Long autotuneId,
-      @QueryParam("start") @NotNull String replayStartTimeIso, @QueryParam("end") @NotNull String replayEndTimeIso) throws IllegalArgumentException{
+      @QueryParam("start") @NotNull String replayStartTimeIso, @QueryParam("end") @NotNull String replayEndTimeIso)
+      throws IllegalArgumentException, InterruptedException {
```
cyy0714: For the exceptions, we should expose this to users through response.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/utils/AnomalyUtils.java
```diff
@@ -117,7 +118,7 @@ public MetaDataNode(MergedAnomalyResultDTO anomaly){
       this.startTimeISO = new Timestamp(anomaly.getStartTime()).toString();
       this.endTimeISO = new Timestamp(anomaly.getEndTime()).toString();
       this.functionName = anomaly.getFunction().getFunctionName();
-      this.feedback = anomaly.getFeedback();
+      this.feedback = (anomaly.getFeedback() == null)? "null" : String.valueOf(anomaly.getFeedback().getFeedbackType());
```
cyy0714: Why store the feedback as String instead of FeedbackType?

cecilynie: Since feedbackType cannot be null

cecilynie: There seems a gap between Feedback and FeedbackType. For feedbackType there is "NO_FEEDBACK", but if anomaly.getFeedback() is null, it then cannot access .getFeedbackType() thus as always we cannot utilize "NO_FEEDBACK" as an anomaly feedback?

PR #1375, opened 2017-04-29
Prevent checking for the field 'replication' while adding LLC tables.

mcvsubbu:
LLC tables do not pay attention to repliction, only to replicasPerPartition.
In case of HLC, replication must equal to (or be a multiple of) the number of
realtime instances, checking it in PinotTableRestletResource.java and setting it to
some arbitrary configured value does not help.

For offline tables, the check is correct but should be done conditionally (i.e. table type is
OFFLINE)

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotTableRestletResource.java
@@ -117,16 +117,14 @@ public class PinotTableRestletResource extends BasePinotControllerRestletResourc
       int requestReplication;
       try {
         requestReplication = segmentsConfig.getReplicationNumber();
+        if (requestReplication < configMinReplication) {
+          LOGGER.info("Creating table with minimum replication factor of: {} instead of requested replication: {}",
+              configMinReplication, requestReplication);
+          segmentsConfig.setReplication(String.valueOf(configMinReplication));
+        }
       } catch (NumberFormatException e) {
         throw new PinotHelixResourceManager.InvalidTableConfigException("Invalid replication number", e);
       }
-      if (requestReplication < configMinReplication) {
-        LOGGER.info("Creating table with minimum replication factor of: {} instead of requested replication: {}",
-            configMinReplication, requestReplication);
-        segmentsConfig.setReplication(String.valueOf(configMinReplication));
-      }
-    } else {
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotTableRestletResource.java
```diff
@@ -92,33 +93,46 @@ private void addTable(AbstractTableConfig config) throws IOException {
...
-        if (replicasPerPartition < configMinReplication) {
-          LOGGER.info(
-              "Creating table with minimum replicasPerPartition of: {} instead of requested replicasPerPartition: {}",
-              configMinReplication, replicasPerPartition);
-          segmentsConfig.setReplicasPerPartition(String.valueOf(configMinReplication));
+        kafkaStreamMetadata = new KafkaStreamMetadata(config.getIndexingConfig().getStreamConfigs());
+      } catch (Exception e) {
+        throw new PinotHelixResourceManager.InvalidTableConfigException("Invalid tableIndexConfig or streamConfigs", e);
+      }
+      if (kafkaStreamMetadata.hasSimpleKafkaConsumerType()) {
```
antumbde: we don't handle HLC replication < configReplication with this change.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotTableRestletResource.java
```diff
@@ -88,25 +89,51 @@ public Representation post(Representation entity) {
...
+      } catch (NumberFormatException e) {
+        throw new PinotHelixResourceManager.InvalidTableConfigException("Invalid replication number", e);
+      }
+      if (requestReplication < configMinReplication) {
+        LOGGER.info("Creating table with minimum replication factor of: {} instead of requested replication: {}",
+            configMinReplication, requestReplication);
+        segmentsConfig.setReplication(String.valueOf(configMinReplication));
+      }
+    } else {
+      throw new PinotHelixResourceManager.InvalidTableConfigException("Invalid table type: '" + config.getTableType() + "'");
```
jfim: Maybe I'm missing something, but doesn't that mean that this will always throw an exception if the table type is LLC-only (`verifyReplication=false`)

mcvsubbu: yes, will fix.

PR #1376, opened 2017-04-30
[TE] return possible values for status, metric, dataset and dimensions in …

kishoreg:
…the anomaly search results.

will update the description with expected output format.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
@@ -1,0 +1,228 @@
+package com.linkedin.thirdeye.dashboard.resources.v2.pojo;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.linkedin.thirdeye.anomalydetection.context.AnomalyFeedback;
+import com.linkedin.thirdeye.api.DimensionMap;
+import com.linkedin.thirdeye.datalayer.dto.MergedAnomalyResultDTO;
+
+public class SearchFilters {
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static final Logger LOG = LoggerFactory.getLogger(SearchFilters.class);
+
+  Map<String, List<Long>> statusFilterMap;
+
+  Map<String, List<Long>> functionFilterMap;
+
+  Map<String, List<Long>> datasetFilterMap;
+
+  Map<String, List<Long>> metricFilterMap;
+
+  Map<String, Map<String, List<Long>>> dimensionFilterMap;
+
+  public SearchFilters() {
+
+  }
+
+  public SearchFilters(Map<String, List<Long>> statusFilterMap, Map<String, List<Long>> functionFilterMap, Map<String, List<Long>> datasetFilterMap,
+      Map<String, List<Long>> metricFilterMap, Map<String, Map<String, List<Long>>> dimensionFilterMap) {
+    super();
+    this.statusFilterMap = sortByValue(statusFilterMap, true);
+    this.functionFilterMap = sortByValue(functionFilterMap, true);
+    this.datasetFilterMap = sortByValue(datasetFilterMap, true);
+    this.metricFilterMap = sortByValue(metricFilterMap, true);
+    this.dimensionFilterMap = new TreeMap<>();
+    for (String dimensionName : dimensionFilterMap.keySet()) {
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
```diff
@@ -0,0 +1,185 @@
...
+import com.linkedin.thirdeye.anomalydetection.context.AnomalyFeedback;
+import com.linkedin.thirdeye.api.DimensionMap;
+import com.linkedin.thirdeye.datalayer.dto.MergedAnomalyResultDTO;
+
+public class SearchFilters {
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static final Logger LOG = LoggerFactory.getLogger(SearchFilters.class);
+
+  TreeMap<String, Integer> statusFilterMap;
```
apucher: Is there a specific reason to type for on "TreeMap" rather than just "Map"?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
```diff
@@ -0,0 +1,228 @@
...
+      if (dimensionFilterMap != null) {
+        DimensionMap dimensions = mergedAnomalyResultDTO.getDimensions();
+        for (String dimensionName : dimensions.keySet()) {
+          if (dimensionFilterMap.containsKey(dimensionName)) {
+            String dimensionValue = dimensions.get(dimensionName);
+            passed = passed && checkFilter(dimensionFilterMap.get(dimensionName), dimensionValue);
+          }
+        }
+      }
+      if (passed) {
```
cyenjung: Should we make the check fails fast? When an anomaly does not pass a filter, we skip to the next anomaly immediately.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
```diff
@@ -0,0 +1,228 @@
...
+import com.linkedin.thirdeye.anomalydetection.context.AnomalyFeedback;
+import com.linkedin.thirdeye.api.DimensionMap;
+import com.linkedin.thirdeye.datalayer.dto.MergedAnomalyResultDTO;
+
+public class SearchFilters {
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static final Logger LOG = LoggerFactory.getLogger(SearchFilters.class);
+
+  Map<String, Integer> statusFilterMap;
```
cyenjung: It would be helpful to indicate that the value (integer) is the count of appearances of keys.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+
+  public static <K, V extends Comparable<? super V>> Map<K, V> sortByValue(Map<K, V> map) {
+    return sortByValue(map, false);
+  }
+
+  public static <K, V extends Comparable<? super V>> Map<K, V> sortByValue(Map<K, V> map, final boolean descending) {
+    final int sortMultipler = (descending) ? -1 : 1;
+    List<Map.Entry<K, V>> list = new LinkedList<Map.Entry<K, V>>(map.entrySet());
```
cyenjung: We could use arraylist here since we are not going to insert or remove elements in/from the middle of the list.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
```diff
@@ -0,0 +1,228 @@
...
+      if (searchFiltersJSON != null) {
+        return OBJECT_MAPPER.readValue(searchFiltersJSON, SearchFilters.class);
+      }
+    } catch (Exception e) {
+      LOG.warn("Exception while parsing searchFiltersJSON  '%s'", searchFiltersJSON, e);
+    }
+    return null;
+  }
+
+  // @Override
```
cyenjung: Deadcode :-)

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
kishoreg: I did not want to have too many if statements in the code. once passed becomes false, the remainder of the AND expression should not get invoked. There should be no significant performance penalty

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/pojo/SearchFilters.java
kishoreg: I missed this comment, will make this change in another PR

PR #1381, opened 2017-05-02
Update idealstate when replication is changed for a table.

mcvsubbu:
When we change the replication for a table, we need to update the idealstate as well,
so that helix sends the right number of messages to the serving instances. Also,
a subsequent rebalance will do the right thing with the number of replicas per partition

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -15,6 +15,41 @@
  */
 package com.linkedin.pinot.controller.helix.core;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections.Predicate;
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
```diff
@@ -206,6 +206,8 @@ public static void buildLowLevelRealtimeIdealStateFor(String realtimeTableName,
     if (idealState == null) {
       idealState = buildEmptyKafkaConsumerRealtimeIdealStateFor(realtimeTableName, nReplicas);
       create = true;
+    } else {
+      idealState.setReplicas(Integer.toString(nReplicas));
```
Jackie-Jiang: I think the nReplicas will be set in PinotLLCRealtimeSegmentManager.java updateIdealState() line 371, can you please check that? If yes, revert the change for this file.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1139,6 +1140,18 @@ public void setExistingTableConfig(AbstractTableConfig config, String tableNameW
     } else if (type == TableType.OFFLINE) {
```
jfim: Should this also apply to realtime tables?

mcvsubbu: For HLC, no because we have only one replica per segment.

For LLC, I tried to make a change but then backed it out as per comments, because a much bigger change is needed to increase number of replicas.

PR #1382, opened 2017-05-02
Allow loading offline table schema with offline table name or raw table name

Jackie-Jiang:
Added APIs in ZKMetadataProvider to load schema using schema name or table name
Change existing code to load schema using the new APIs


pinot-controller/src/main/java/com/linkedin/pinot/controller/util/AutoAddInvertedIndex.java
```diff
@@ -154,8 +154,16 @@ private void runQueryStrategy()
       }
       LOGGER.info("Table: {} matches the table name pattern: {}", tableName, _tableNamePattern);
 
+      // Get the table type
```
mcvsubbu: Please use TableNameBuilder.getTableTypeFromTableName()

Jackie-Jiang: Fixed.

PR #1384, opened 2017-05-02
Fix bug causing incorrect values for wow in derived metrics

npawar:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/AnomaliesResource.java
```diff
@@ -188,14 +188,15 @@ private double calculateChange(double currentValue, double baselineValue) {
...
   }
 
-  double getTotalFromTimeSeries (MetricTimeSeries mts, boolean isAdditive) {
+  double getTotalFromTimeSeries (MetricTimeSeries metricTimeSeries, boolean isAdditive) {
     double total = 0.0;
-    for (Number num : mts.getMetricSums()) {
-      total += num.doubleValue();
-    }
+
+    Number[] metricTotals = metricTimeSeries.getMetricSums();
```
kishoreg: please add a comment that we can get multiple values if the metric is a ratio

PR #1387, opened 2017-05-03
Update the highest kafka offset metric correctly so that we can use it for alerting

mcvsubbu:
We currently have the capability to raise an alert if the metric kafka number of rows consumed goes to 0
in order to indicate a fault in the system that can cause us to stop consuming from Kafka.
However, this metric does not give the correct picture when there are no events in the kafka
pipeline, or when the kafka pipeline is burstly (e.g. pipelines from Samza jobs).

The idea here is to use the Kafka Highest Offset Consumed metric to detect these conditions.

In order for us to use this metric effectively, we need to make sure that this metric is set to 0
when we are not consuming any rows at all.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -259,15 +259,15 @@ public class LLRealtimeSegmentDataManager extends SegmentDataManager {
     } else {
       segmentLogger.warn("Kafka transient exception when fetching messages, retrying (count={})", consecutiveErrorCount, e);
       Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
-      makeConsumerWrapper();
+      makeConsumerWrapper("Too many transient errors");
     }
   }
 
   protected boolean consumeLoop() throws Exception {
     _fieldExtractor.resetCounters();
     final long idlePipeSleepTimeMillis = 100;
-    final int maxIdleCountBeforeStatUpdate = 6000;  // 1 minute
-    long lastUpdatedOffset = 0L;  // so that we always update the metric when we enter this method.
+    final int maxIdleCountBeforeStatUpdate = 180_000;  // 3 minutes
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -498,6 +508,8 @@ public void run() {
             ServerGauge.LAST_REALTIME_SEGMENT_COMPLETION_DURATION_SECONDS,
             TimeUnit.MILLISECONDS.toSeconds(now() - initialConsumptionEnd));
       }
+      // Set the highest offset consumed to 0, the next consumer will restart it.
+      _serverMetrics.setValueOfTableGauge(_metricKeyName, ServerGauge.HIGHEST_KAFKA_OFFSET_CONSUMED, 0L);
```
rven1: For the duration of the state transition, this metric will be 0, which is not correct.  I think you are repurposing HIGHEST_OFFSET_CONSUMED for other purposes. Perhaps a new metric will be more appropriate.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -259,12 +259,16 @@ private void handleTransientKafkaErrors(Exception e) throws  Exception {
...
       Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
-      makeConsumerWrapper();
+      makeConsumerWrapper("Too many transient errors");
     }
   }
 
   protected boolean consumeLoop() throws Exception {
     _fieldExtractor.resetCounters();
+    final long idlePipeSleepTimeMillis = 100;
+    final int maxIdleCountBeforeStatUpdate = 180_000;  // 3 minutes
```
jfim: Isn't that 180k times * 100ms (5 hours), assuming we're fully idle?

mcvsubbu: good catch, fixed.

PR #1388, opened 2017-05-03
[TE] Shade or exclude commonly used jars in other projects from TE Hadoop fat jar

cyenjung:
Shade old version of Thrift library, which is a dependency of pinot-core, to allow other libraries that depends on ThirdEye jar could use newer version of Thrift.


kishoreg:
why do we need thrift? we can simply exclude it from the jar. I think it's coming transitively from Pinot.

cyenjung:
Amended the pom file to exclude thrift and shade fastjackson's databind.

PR #1402, opened 2017-05-05
[TE] Hotfix for limiting thread creation in MergedAnomalyResultManagerImpl

cyenjung:
Hotfix for reducing the number of threads being created. The root cause of too many thread pools being created in this class is that DAO registry's getMergedAnomalyResultDAO() method keeps creating a new manager whenever it is invoked.

Tested on local controller.


apucher:
I think i have some overlap in the full fix #1403

PR #1405, opened 2017-05-05
[PINOT-4885] Change FixedByteSingleColumnMultiValueReaderWriter to expand header section dynamically

mcvsubbu:
Changed FixedByteSingleColumnMultiValueReaderWriter so that the header sections also expands like the data
section, so that we don't have to allocate data for max number of rows in the beginning.

For now, the same allocation is maintained so that we don't see any change in the behavior in production.

This, and the next few PRs should help us limit consumption by memory (instead of by number of rows) or time,
making it easier to configure real-time service in pinot.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
@@ -15,7 +15,6 @@
  */
 package com.linkedin.pinot.core.io.readerwriter.impl;
 
-import java.io.IOException;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
 import java.util.List;
@@ -24,19 +23,47 @@ import com.linkedin.pinot.core.io.readerwriter.BaseSingleColumnMultiValueReaderW
 import com.linkedin.pinot.core.io.writer.impl.FixedByteSingleValueMultiColWriter;
 import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
 
-
 /**
- * Supports both reads and writes using the same data structure.<br>
- * Writes must be strictly sequential, while reads can be random <br>
- * It is very similar to the SingleColumnMultiValue format representation <br>
- * except that the variable size data buffer size is not known up front in case FixedByteSingleColumnMultiValueReaderWriter
- * This class allocates extra memory in chunks as needed.
+ * This class provides expandable off-heap implementation to store a multi-valued column across a number of rows.
+ * The maximum number of values in any row must be known while invoking the constructor. Other than that, this class
+ * allocates additional memory as needed to accommodate any number of rows.
+ *
+ * Writes into the data structure are strictly sequential, but reads can be random.
+ *
+ * Writes are of type:
+ *
+ *   setIntArray(int rowNumber, int[] values)
+ *
+ * It is expected that rowNumber increments by 1 on each invocation, and that it is decided ahead of time that
+ * the class is used to store a certain type of data structure (int arrays, or char arrays, etc.) Mix & match is not
+ * allowed.
+ *
+ * Two kinds of data structures are used in this class.
+ *
+ * 1. A header (essentially an index into the other data structure) that has one entry per row. The entry has 3 integers
+ *   - data buffer ID
+ *   - offset in the data buffer where column values start
+ *   - length (number of values in the multi-valued column).
+ *
+ *   New header structures are added as new rows come in. Each header class holds the same number of rows (for easy lookup)
+ *
+ * 2. A data buffer that has the values for the column that the header points to. Data buffers are added as needed,
+ *    whenever we reach a limitation that we cannot fit the values of a column in the current buffer.
+ *
+ * Note that data buffers and headers grow independently.
+ *
  * Data format
  * <code>
- *  HEADER SECTION
+ *  HEADER SECTION 0
  *    bufferId startIndex EndIndex
  *    bufferId startIndex EndIndex
  *    bufferId startIndex EndIndex
+ *    ...
+ *  HEADER SECTION 1
+ *    bufferId startIndex EndIndex
+ *    bufferId startIndex EndIndex
+ *    bufferId startIndex EndIndex
+ *    ...
  *  Data BUFFER SECTION 0
  *    [set of values of row 0] [set of values of row 1]
  *     .....
@@ -55,7 +82,6 @@ import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
 
 public class FixedByteSingleColumnMultiValueReaderWriter extends BaseSingleColumnMultiValueReaderWriter {
 
-  public static final int DEFAULT_MAX_NUMBER_OF_MULTIVALUES = 1000;
   /**
    * number of columns is 1, column size is variable but less than maxNumberOfMultiValuesPerRow
    * @param rows
@@ -68,8 +94,10 @@ public class FixedByteSingleColumnMultiValueReaderWriter extends BaseSingleColum
 
   private PinotDataBuffer headerBuffer;
   private List<PinotDataBuffer> dataBuffers = new ArrayList<>();
-  private FixedByteSingleValueMultiColWriter headerWriter;
-  private FixedByteSingleValueMultiColReader headerReader;
+  private List<PinotDataBuffer> headerBuffers = new ArrayList<>();
+  private List<FixedByteSingleValueMultiColReader> headerReaders = new ArrayList<>();
+  private List<FixedByteSingleValueMultiColWriter> headerWriters = new ArrayList<>();
+  private FixedByteSingleValueMultiColWriter curHeaderWriter;
   private List<FixedByteSingleValueMultiColWriter> dataWriters = new ArrayList<FixedByteSingleValueMultiColWriter>();
   private List<FixedByteSingleValueMultiColReader> dataReaders = new ArrayList<FixedByteSingleValueMultiColReader>();
   private FixedByteSingleValueMultiColWriter currentDataWriter;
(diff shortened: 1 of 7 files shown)


pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteSingleColumnMultiValueReaderWriterTest.java
```diff
@@ -31,9 +31,19 @@ public void testIntArray() {
     final long seed = r.nextLong();
```
antumbde: tests should cover for all types if we have copy-pasted code

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -79,46 +107,52 @@
...
+  private int prevRowStartIndex = 0;  // Offset in the databuffer for the last row added.
+  private int prevRowLength = 0;  // Number of values in the column for the last row added.
 
-  public FixedByteSingleColumnMultiValueReaderWriter(int rows, int columnSizeInBytes, int maxNumberOfMultiValuesPerRow,
-      int avgMultiValueCount)
-      throws IOException {
-    int initialCapacity = Math.max(maxNumberOfMultiValuesPerRow, rows * avgMultiValueCount);
+  public FixedByteSingleColumnMultiValueReaderWriter(int maxNumberOfMultiValuesPerRow, int avgMultiValueCount, int rowCountPerChunk,
+      int columnSizeInBytes) {
+    int initialCapacity = Math.max(maxNumberOfMultiValuesPerRow, rowCountPerChunk * avgMultiValueCount);
```
antumbde: Preconditions to assert these inputs are > 0. Definitely needed for rowCOuntPerChunk to ensure headerSize > 3 * sizeof(int)

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteSingleColumnMultiValueReaderWriterTest.java
```diff
@@ -43,7 +53,8 @@ public void testIntArray(final long seed)
     int columnSizeInBytes = Integer.SIZE / 8;
     int maxNumberOfMultiValuesPerRow = 2000;
     readerWriter =
-        new FixedByteSingleColumnMultiValueReaderWriter(rows, columnSizeInBytes, maxNumberOfMultiValuesPerRow, 2);
+        new FixedByteSingleColumnMultiValueReaderWriter(maxNumberOfMultiValuesPerRow, 2, rows/2, columnSizeInBytes);
+        //new FixedByteSingleColumnMultiValueReaderWriter(1000, columnSizeInBytes, maxNumberOfMultiValuesPerRow, 2);
```
antumbde: remove

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteSingleColumnMultiValueReaderWriterTest.java
```diff
@@ -62,4 +73,62 @@ public void testIntArray(final long seed)
     }
     readerWriter.close();
   }
+
+  public void testIntArrayFixedSize(int multiValuesPerRow, long seed)
+      throws IOException {
+    FixedByteSingleColumnMultiValueReaderWriter readerWriter;
+    int rows = 1000;
```
antumbde: rows  = random(1, multiValuesPerRow) to test chunking. I think the caller goes up to multiValuesPerRow of 1000 whcih may not test chunking

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -79,46 +107,52 @@
...
+    headerReaders.add(curHeaderReader);
   }
 
   /**
    * This method automatically computes the space needed based on the columnSizeInBytes
    * @param rowCapacity Additional capacity to be added in terms of number of rows
    * @throws RuntimeException
    */
-  private void addCapacity(int rowCapacity) throws RuntimeException {
+  private void addDataBuffers(int rowCapacity) throws RuntimeException {
```
antumbde: [saying aloud; no action needed] using FixedByteSingleValueMultiColWriter and writing one value at a time seems suboptimal for writing array of values. For nativeByte order, we should be able to use PinotDataBuffer and readFrom method for bulk write but layered interfaces are hurting us....

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -234,9 +282,11 @@ public void setBytesArray(int row, byte[][] bytesArray) {
 
   @Override
   public int getCharArray(int row, char[] charArray) {
-    int bufferIndex = headerReader.getInt(row, 1);
-    int startIndex = headerReader.getInt(row, 1);
-    int length = headerReader.getInt(row, 2);
+    FixedByteSingleValueMultiColReader headerReader = getCurrentReader(row);
```
antumbde: [again speaking aloud] we should be able to bulk copy....
Another point is, for our existing data, most of the time we write one value (inputArray.length == 1). We should special case that with if-else because setting loops is costlier. Not sure how jvm optimizes that at runtime. Needs benchmarking

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -24,19 +23,47 @@
...
+ * The maximum number of values in any row must be known while invoking the constructor. Other than that, this class
+ * allocates additional memory as needed to accommodate any number of rows.
+ *
+ * Writes into the data structure are strictly sequential, but reads can be random.
+ *
+ * Writes are of type:
+ *
+ *   setIntArray(int rowNumber, int[] values)
+ *
+ * It is expected that rowNumber increments by 1 on each invocation, and that it is decided ahead of time that
```
mcvsubbu: done

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -24,19 +23,47 @@
...
+ * 2. A data buffer that has the values for the column that the header points to. Data buffers are added as needed,
+ *    whenever we reach a limitation that we cannot fit the values of a column in the current buffer.
+ *
+ * Note that data buffers and headers grow independently.
+ *
  * Data format
  * <code>
- *  HEADER SECTION
+ *  HEADER SECTION 0
+ *    bufferId startIndex EndIndex
```
mcvsubbu: done

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
mcvsubbu: done

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -79,46 +107,52 @@
...
             new int[] { SIZE_OF_INT, SIZE_OF_INT, SIZE_OF_INT });
-    headerReader =
-        new FixedByteSingleValueMultiColReader(headerBuffer, rows, new int[] { SIZE_OF_INT, SIZE_OF_INT, SIZE_OF_INT });
-    //at least create space for million entries, which for INT translates into 4mb buffer
-    this.incrementalCapacity = incrementalCapacity;
-    addCapacity(initialCapacity);
+    FixedByteSingleValueMultiColReader curHeaderReader =
+        new FixedByteSingleValueMultiColReader(headerBuffer, rowCountPerChunk, new int[] { SIZE_OF_INT, SIZE_OF_INT, SIZE_OF_INT });
+    headerBuffers.add(headerBuffer);
+    headerWriters.add(curHeaderWriter);
```
mcvsubbu: yes

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -142,28 +176,42 @@ public void close() {
       dataBuffer.close();
     }
     dataBuffers.clear();
-    headerBuffer.close();
+    for (PinotDataBuffer headerBuffer : headerBuffers) {
```
mcvsubbu: we don't want to call close on the buffer reader/writer since they use the same buffer that we use. Otherwise, we will get double-close.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -142,28 +176,42 @@ public void close() {
...
+    if (row >= headerBuffers.size() * rowCountPerChunk) {
+      addHeaderBuffers();
     }
+    curHeaderWriter.setInt(getRowInCurrentHeader(row), 0, dataWriterIndex);
+    curHeaderWriter.setInt(getRowInCurrentHeader(row), 1, startIndex);
+    curHeaderWriter.setInt(getRowInCurrentHeader(row), 2, length);
+  }
+
+  private final FixedByteSingleValueMultiColReader getCurrentReader(int row) {
+    return headerReaders.get(row/ rowCountPerChunk);
```
mcvsubbu: fixed

mcvsubbu: Yes, that is why i have it in a separate method. I will add a TODO. Right now, it is not clear yet how the rowsCountPerChunk will be decided. Most likely it will be outside this class, since other classes are also getting re-factored to support multi-tenant realtime

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -234,9 +282,11 @@ public void setBytesArray(int row, byte[][] bytesArray) {
 
   @Override
   public int getCharArray(int row, char[] charArray) {
-    int bufferIndex = headerReader.getInt(row, 1);
-    int startIndex = headerReader.getInt(row, 1);
-    int length = headerReader.getInt(row, 2);
+    FixedByteSingleValueMultiColReader headerReader = getCurrentReader(row);
+    int rowInCurrentHeader  = getRowInCurrentHeader(row);
+    int bufferIndex = headerReader.getInt(rowInCurrentHeader, 1);
```
mcvsubbu: fixed for all the getters. 

This class is not used for anything other than int. Removing the other methods makes the logic crisp and clear, but because of the way inheritance is organized, I could not do it without significant changes (otherwise, that was my first choice).

I also considered adding a 'throws unsupported' on each of the others, but refrained at this time.

I will add a unit test for the other use cases for now, but probably not as extensive as the int[] use case.

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteSingleColumnMultiValueReaderWriterTest.java
mcvsubbu: See above.

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteSingleColumnMultiValueReaderWriterTest.java
mcvsubbu: done

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteSingleColumnMultiValueReaderWriterTest.java
mcvsubbu: I intentionally wanted to check without chunking in this method, but I added a random chunk size in the next test.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
antumbde: Calling close on closed PinotByteBuffer is safe. We should call close() to ensure all writer resources are correctly released

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
mcvsubbu: Will add close of all readers and writers when reader/writer is closed.

pinot-core/src/test/java/com/linkedin/pinot/core/realtime/impl/dictionary/MultiValueDictionaryTest.java
```diff
@@ -15,15 +15,15 @@
...
 import java.util.Random;
 import org.testng.Assert;
 import org.testng.annotations.Test;
+import com.linkedin.pinot.core.io.readerwriter.impl.FixedByteSingleColumnMultiValueReaderWriter;
 
 
 public class MultiValueDictionaryTest {
   private static final int NROWS = 1000;
-  private static final int MAX_N_VALUES = FixedByteSingleColumnMultiValueReaderWriter.DEFAULT_MAX_NUMBER_OF_MULTIVALUES;
+  private static final int MAX_N_VALUES = 1000;
```
jfim: Shouldn't we use the `MAX_MULTI_VALUES_PER_ROW` constant?

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -68,8 +94,10 @@
 
   private PinotDataBuffer headerBuffer;
   private List<PinotDataBuffer> dataBuffers = new ArrayList<>();
-  private FixedByteSingleValueMultiColWriter headerWriter;
-  private FixedByteSingleValueMultiColReader headerReader;
+  private List<PinotDataBuffer> headerBuffers = new ArrayList<>();
```
jfim: Since this is a list of buffers, what's the order of magnitude of the typical length of that list? Is it low (eg. 1-10 entries) or something large like 100+ entries?

You'll want to keep in mind that if this is large, you'll want to coalesce buffers as I believe that large allocations are implemented as anonymous `mmap`s; as such, the process could exhaust `vm.max_map_count` and cause the JVM to crash.

mcvsubbu: As yet, the number of buffers is 1, because we still limit by the number of rows.

When we start limiting by memory (once we have the other pieces of the puzzle fixed to use off-heap) we can work on tuning this to coalesce if necessary. The idea is to try to keep this to 1 or a few at best, by taking hints from the previous completed segment

pinot-core/src/test/java/com/linkedin/pinot/core/realtime/impl/dictionary/MultiValueDictionaryTest.java
mcvsubbu: Why? This one is independent of MAX_MULTI_VALUES_PER_ROW.

PR #1407, opened 2017-05-07
Misc changes for reducing garbage generation during query execution.

mayankshriv:
1. Used thread local memory during query execution, wherever possible.
2. Cached FilterQueryTree inside QueryRequest, to avoid re-creating
   it over and over again.
3. Renamed QueryRequest to ServerQueryRequest to indicate it is
   server-side query request.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/query/ServerQueryRequest.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/query/ServerQueryRequest.java
@@ -22,6 +22,8 @@ import com.linkedin.pinot.common.query.context.TimerContext;
 import com.linkedin.pinot.common.request.BrokerRequest;
 import com.linkedin.pinot.common.request.InstanceRequest;
 import com.linkedin.pinot.common.utils.DataTable;
+import com.linkedin.pinot.common.utils.request.FilterQueryTree;
+import com.linkedin.pinot.common.utils.request.RequestUtils;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -33,10 +35,11 @@ import org.slf4j.LoggerFactory;
  * etc.
  *
  */
-public class QueryRequest {
-  private static final Logger LOGGER = LoggerFactory.getLogger(QueryRequest.class);
+public class ServerQueryRequest {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ServerQueryRequest.class);
 
   private final InstanceRequest instanceRequest;
+  private final FilterQueryTree filterQueryTree;
   // Future to track result of query execution
   private ListenableFuture<DataTable> resultFuture;
   // Timing information for different phases of query execution
@@ -45,10 +48,11 @@ public class QueryRequest {
   private final ServerMetrics serverMetrics;
   private int segmentCountAfterPruning;
 
-  public QueryRequest(InstanceRequest request, ServerMetrics serverMetrics) {
+  public ServerQueryRequest(InstanceRequest request, ServerMetrics serverMetrics) {
     this.instanceRequest = request;
     this.serverMetrics = serverMetrics;
     BrokerRequest brokerRequest = (request == null) ? null : request.getQuery();
+    filterQueryTree = (brokerRequest == null) ? null : RequestUtils.generateFilterQueryTree(brokerRequest);
     timerContext = new TimerContext(brokerRequest, serverMetrics);
   }
 
@@ -71,6 +75,13 @@ public class QueryRequest {
    */
   public BrokerRequest getBrokerRequest() {
     return instanceRequest.getQuery();
+  }
+
+  /**
+   * Returns the filter query tree for the server request.
+   */
+  public FilterQueryTree getFilterQueryTree() {
+    return filterQueryTree;
   }
 
   /**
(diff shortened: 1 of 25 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/common/DataFetcher.java
```diff
@@ -43,11 +43,24 @@
...
+  private static final ThreadLocal<int[]> THREAD_LOCAL_DICT_IDS = new ThreadLocal<int[]>() {
+    @Override
+    protected int[] initialValue() {
+      return new int[DocIdSetPlanNode.MAX_DOC_PER_CALL];
+    }
+  };
 
   // Re-usable array to store MV dictionary id's for a given docId
-  private final int[] _reusableMVDictIds;
+  private static final ThreadLocal<int[]> THREAD_LOCAL_MV_DICT_IDS = new ThreadLocal<int[]>() {
```
Jackie-Jiang: Once this buffer get expanded, it will never get reduced during the whole life cycle, which seems not safe to me.

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -219,7 +219,7 @@ public DataTable processQuery(final QueryRequest queryRequest, ExecutorService e
    * @return Total number of docs across all segments (including the ones that were pruned).
    */
   private long pruneSegments(TableDataManager tableDataManager, List<SegmentDataManager> segments,
-      BrokerRequest brokerRequest) {
+      ServerQueryRequest brokerRequest) {
```
Jackie-Jiang: Please rename "brokerRequest"

pinot-common/src/main/java/com/linkedin/pinot/common/query/ServerQueryRequest.java
```diff
@@ -33,10 +35,11 @@
  * etc.
  *
  */
-public class QueryRequest {
-  private static final Logger LOGGER = LoggerFactory.getLogger(QueryRequest.class);
+public class ServerQueryRequest {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ServerQueryRequest.class);
 
   private final InstanceRequest instanceRequest;
+  private final FilterQueryTree filterQueryTree;
```
Jackie-Jiang: Use new coding style?

mayankshriv:
@Jackie-Jiang Yes this was found when profiling a usecase with throughput requirement of 2k QPS and large number of segments (> 2000 segments, > 3TB data). While for small/medium size usecases these may not matter much, but for large usecases with high qps/low latency requirements, the 99th percentile latency is all of GC. This PR addresses some of the major ones that were found in the query execution path.

pinot-core/src/main/java/com/linkedin/pinot/core/common/DataFetcher.java
mayankshriv: I am open to suggestions, if you have one.

pinot-common/src/main/java/com/linkedin/pinot/common/query/ServerQueryRequest.java
mayankshriv: I refrain from changing member variable naming convention to make it simpler to review the code. Better to have pure clean ups without any functional change as separate PR's.

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
mayankshriv: Done.

pinot-core/src/main/java/com/linkedin/pinot/core/common/DataFetcher.java
antumbde: question is do we even need to lower it if the server is going to serve similar data and queries ? 
Frequency of expand/shrink is important otherwise whole change is not useful.
Two options:
1. If the expanded buffer will be one-off case then use that temporarily
2. Keep accounting information in get() call. Adjust based on values seen in last n accesses. But most of this will be counter-productive for GC goal

antumbde:
@mayankshriv Can you address my comments above ? Why introduce a change that does not impact GC ?

pinot-core/src/main/java/com/linkedin/pinot/core/common/DataFetcher.java
mayankshriv: This is only for MV columns. Even if we had an MV column with 1k values (that is already too large), we will use 4k bytes for memory. So I am not convinced this can cause an issue.

mayankshriv:
@antumbde If it was not clear from my reply to @Jackie-Jiang 's question, I will state it again. This is what I observed in the use case with ~2k QPS on 500 GB of data on one node, where we are trying to improve the 99th percentile latency (which is totally dominated by young gen GC pauses):

1. We create > 10GB of garbage per 15 seconds.
2. About 20% of this comes from issues addressed in this PR. After this PR, that 20% is completely eliminated.

In order to see  improvements in the 99th latency for the use case, we still need to resolve the rest of the GC issues. But in the mean while, I'd say that it is still a pretty good savings in garbage creation, and therefore warrants to be pushed.

antumbde:
Thanks! Your original comment did not have some numbers and if this PR actually solves that.
These numbers indicate about 1.5k garbage per query per second for the changes in PR. Seems too high...

PR #1411, opened 2017-05-08
Reuse metrics during realtime ingestion

jfim:
Add an API to reuse meters instead of allocating new ones on every
metric update. Reuse metrics for the main loop of the realtime
ingestion as to avoid a lot of garbage generated during realtime
indexing.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -64,6 +64,7 @@ import com.linkedin.pinot.core.realtime.impl.kafka.KafkaSimpleConsumerFactoryImp
 import com.linkedin.pinot.core.realtime.impl.kafka.SimpleConsumerWrapper;
 import com.linkedin.pinot.core.segment.index.loader.IndexLoadingConfig;
 import com.linkedin.pinot.server.realtime.ServerSegmentCompletionProtocolHandler;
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -376,15 +380,15 @@ private void processKafkaEvents(Iterable<MessageAndOffset> messagesAndOffsets, L
         transformedRow = _fieldExtractor.transform(decodedRow, transformedRow);
 
         if (transformedRow != null) {
-          _serverMetrics.addMeteredTableValue(_metricKeyName, ServerMeter.REALTIME_ROWS_CONSUMED, 1);
+          realtimeRowsConsumedMeter = _serverMetrics.addMeteredTableValue(_metricKeyName, ServerMeter.REALTIME_ROWS_CONSUMED, 1, realtimeRowsConsumedMeter);
```
mayankshriv: Does it make sense to batch the metric emission every k rows, instead of each row?

jfim: We could do it for every batch of rows that we get from Kafka. I looked at the underlying code and it just does a CAS on a long value (actually it does 4, as it keeps track of the 1/5/15 minute rates as well), so it's not that expensive.

I still need to review our usage of other metrics and update them if necessary to use this newer API, as well as probably change the API a bit for the metrics, so I'll keep that in mind when I revisit that part.

PR #1413, opened 2017-05-09
Added support to PinotHelixResourceManager to handle Minion resources

Jackie-Jiang:
Reordered and added javadoc for part of the methods

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/pojos/Instance.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/pojos/Instance.java
@@ -15,14 +15,13 @@
  */
 package com.linkedin.pinot.controller.api.pojos;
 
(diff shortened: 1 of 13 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/pojos/Instance.java
```diff
@@ -15,15 +15,14 @@
  */
 package com.linkedin.pinot.controller.api.pojos;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
```
mayankshriv: We should use just one of fasterxml/codehaus for Json.

Jackie-Jiang: Not sure if it's safe to change it here. Will address all fasterxml/codehaus problems in a separate PR.

PR #1416, opened 2017-05-09
Adding simple date format to segments config

jenniferdai:


pinot-common/src/test/java/com/linkedin/pinot/common/config/TableConfigTest.java
```diff
@@ -31,7 +32,7 @@ public void testInit()
...
+          "{\"tableName\":\"myTable\",\"quota\":{\"storage\":\"30G\"},\"segmentsConfig\":{\"dateFormat\":{\"simpleDate\":\"yymmdd\"},\"retentionTimeUnit\":\"DAYS\",\"retentionTimeValue\":\"60\",\"segmentPushFrequency\":\"daily\",\"segmentPushType\":\"APPEND\",\"replication\":\"2\",\"schemaName\":\"\",\"timeColumnName\":\"\",\"timeType\":\"\",\"segmentAssignmentStrategy\":\"BalanceNumSegmentAssignmentStrategy\"},\"tableIndexConfig\":{\"invertedIndexColumns\":[],\"loadMode\":\"HEAP\",\"lazyLoad\":\"false\"},\"tenants\":{\"broker\":\"colocated\",\"server\":\"myServer\"},\"tableType\":\"OFFLINE\",\"metadata\":{\"customConfigs\":{\"loadBalancer\":\"myLoadBalancer\"}}}";
```
mcvsubbu: The idea behind this test was to make sure that we do not get any json exceptions whether or not quota is set. You have modified this test to include dateformat, which is ok (Ideally a different independent test is better), but you should also test that the dateformat is null when the table config does not have it, as in lines below where the quota is tested to be null.

as a nit, please change the comments to include "with quota & dateformat" and "no quota and dateformat"

PR #1418, opened 2017-05-09
[TE] make entity mapping type a generic string

apucher:


kishoreg:
what is the purpose of making it string?

apucher:
This allows users to add mappings between custom entities without forking and modifying thirdeye's source code. For example, events and metrics from a proprietary A/B testing framework could be associated with each other.

PR #1422, opened 2017-05-10
In PinotTableRestletResource, for client error, log info instead of error

Jackie-Jiang:


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotTableRestletResource.java
```diff
@@ -285,8 +283,10 @@ private StringRepresentation setTablestate(
     if (tableExists) {
       return new StringRepresentation(ret.toString());
     } else {
+      String error = "Error: Table " + tableName + " not found.";
```
mcvsubbu: can you remove the word "Error" from the message?

Jackie-Jiang: Sure.

PR #1441, opened 2017-05-16
[TE] Fix NULL pointer exceptions in dashboard

cyy0714:
Add missing init() for AnomalyDetectionInputContextBuilder


apucher:
why isn't init() called internally from the constructor?

PR #1442, opened 2017-05-16
[TE] fix warn message

apucher:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/RootCauseResource.java
```diff
@@ -122,7 +122,7 @@ RootCauseEntity applyFormatters(Entity e) {
         try {
           return formatter.format(e);
         } catch (Exception ex) {
-          LOG.warn("Error applying formatter '{}'. Skipping.", ex);
+          LOG.warn("Error applying formatter '{}'. Skipping.", formatter.getClass().getName(), ex);
```
cyenjung: Shouldn't we need to '{}' to print out both class.getName and ex?

apucher: The last Throwable in the varargs is automatically appended by SLF4J convention

cyenjung: I see.

PR #1451, opened 2017-05-17
[TE] support multiple (named) framework in single configuration

apucher:
* add support for multiple named framework configurations

* allows execution of different pipelines per use-case (e.g. only retrieve related metrics, but not events)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/RCAFrameworkLoader.java
```diff
@@ -64,4 +68,10 @@
     return properties;
   }
 
+  private static boolean isEmpty(Map<?, ?> map) {
```
npawar: Could use MapUtils.isEmpty(), isn't it doing the same thing?

kishoreg:
Please add description on why this is required

PR #1452, opened 2017-05-17
Avoid query execution path for simple count(*) queries.

mayankshriv:
For count(*) queries without filters and group by's, we can simply add
up total.raw.docs from all segments, instead of going through the query
planning/execution path which is very expensive for such queries.

This does not help for hybrid usecases, as time filter is added
internally before sending to the OFFLINE/REALTIME tables.


kishoreg:
+1 for idea. However the implementation could be done better. The planning phase can setup operators that can simply work on the metadata instead of actual data. With this, we can extend this logic to min/max and other aggregation functions and it also follows the same execution flow.

mayankshriv:
@kishoreg Agreed, that is a good idea. Filed PR-1460 with an initial ground work that supports serving count(*) using metadata, and makes it easy to add support for other such queries as well.

PR #1458, opened 2017-05-18
[TE] refactor rootcause endpoint to support related metrics query

apucher:
* add REST endpoint to RootCauseResource to query related metrics

* remove redundancy from RootCauseResource

* add dashboard config params


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/ThirdEyeDashboardConfiguration.java
```diff
@@ -11,7 +11,9 @@
   String rcaConfigPath;
   int rcaParallelism;
   List<String> rcaFormatters;
-  String rcaDefaultFramework;
```
npawar: Don't these need default values? Can dashboard never be started without rca framework?

apucher: It's mostly a limitation of how we load configs. If a property is not in the config file, it errors out. We could could change that and make RCA (and other stuff) optional

PR #1462, opened 2017-05-18
[PINOT-4887] Added support for no-dictionary columns in Realtime consuming segment

mcvsubbu:
Queries work on no dictionary columns in the consuming segment, but segment generation
is broken because of assumption of presense of a dictionary while getting row values.

Limiting the hanges in this PR to make it smaller. Will submit another PR to get
the segment generation right, and then enable the feature.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
@@ -250,9 +250,7 @@ public class FixedByteSingleColumnSingleValueReaderWriter extends BaseSingleColu
     if (readers.size() == 1) {
       readers.get(0).readValues(rows, rowStartPos, rowSize, values, valuesStartPos);
       return;
-    }
-    else
-    {
+    } else {
       for (int rowIter = rowStartPos, valueIter = valuesStartPos;
           rowIter < rowStartPos + rowSize;
           rowIter++, valueIter++) {
@@ -267,9 +265,7 @@ public class FixedByteSingleColumnSingleValueReaderWriter extends BaseSingleColu
     if (readers.size() == 1) {
       readers.get(0).readValues(rows, rowStartPos, rowSize, values, valuesStartPos);
       return;
-    }
-    else
-    {
+    } else {
       for (int rowIter = rowStartPos, valueIter = valuesStartPos;
           rowIter < rowStartPos + rowSize;
           rowIter++, valueIter++) {
@@ -284,9 +280,7 @@ public class FixedByteSingleColumnSingleValueReaderWriter extends BaseSingleColu
     if (readers.size() == 1) {
       readers.get(0).readValues(rows, rowStartPos, rowSize, values, valuesStartPos);
       return;
-    }
-    else
-    {
+    } else {
       for (int rowIter = rowStartPos, valueIter = valuesStartPos;
           rowIter < rowStartPos + rowSize;
           rowIter++, valueIter++) {
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.core.io.readerwriter.impl;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.nio.ByteOrder;
 import java.util.ArrayList;
@@ -28,22 +29,100 @@ import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
 // TODO Check if MMAP allows us to pack more partitions on a single machine.
 public class FixedByteSingleColumnSingleValueReaderWriter extends BaseSingleColumnSingleValueReaderWriter {
 
-  FixedByteSingleValueMultiColWriter currentWriter;
-  private List<FixedByteSingleValueMultiColWriter> writers = new ArrayList<>();
-  // For readers it is useful to keep a "chunked-reader" class that is aware of the start and end offset of that buffer.
-  // If we do that then we avoid the % operation on the record on every read.
-  // For the bulk read operation, we can do a loop like this one:
-  //   foreach(row in rows[]) {
-  //     bufferId = row / chunkSize;
-  //     chunked-readers.get(bufferId).getInt(row);
-  //   }
-  // TODO the logic above
-  private List<FixedByteSingleValueMultiColReader> readers = new ArrayList<>();
+  WriterWithOffset currentWriter;
+  private List<WriterWithOffset> writers = new ArrayList<>();
+  private List<ReaderWithOffset> readers = new ArrayList<>();
   private List<PinotDataBuffer> buffers = new ArrayList<>();
   private final long chunkSizeInBytes;
   private final int numRowsPerChunk;
   private final int columnSizesInBytes;
   private long numBytesInCurrentWriter = 0;
+
+  private static class WriterWithOffset implements Closeable {
+    final FixedByteSingleValueMultiColWriter writer;
+    final int startRowId;
+
+    private WriterWithOffset(FixedByteSingleValueMultiColWriter writer, int startRowId) {
+      this.writer = writer;
+      this.startRowId = startRowId;
+    }
+
+    @Override
+    public void close() {
+      writer.close();
+    }
+
+    public void setInt(int row, int value) {
+      writer.setInt(row - startRowId, 0, value);
+    }
+
+    public void setLong(int row, long value) {
+      writer.setLong(row - startRowId, 0, value);
+    }
+
+    public void setFloat(int row, float value) {
+      writer.setFloat(row - startRowId, 0, value);
+    }
+
+    public void setDouble(int row, double value) {
+      writer.setDouble(row - startRowId, 0, value);
+    }
+
+    public void setShort(int row, short value) {
+      writer.setShort(row - startRowId, 0, value);
+    }
+  }
+
+  private static class ReaderWithOffset implements Closeable {
+    final FixedByteSingleValueMultiColReader reader;
+    final int startRowId;
+
+    private ReaderWithOffset(FixedByteSingleValueMultiColReader reader, int startRowId) {
+      this.reader = reader;
+      this.startRowId = startRowId;
+    }
+
+    @Override
+    public void close() {
+      reader.close();
+    }
+
+    public int getInt(int row) {
+      return reader.getInt(row - startRowId, 0);
+    }
+
+    public long getLong(int row) {
+      return reader.getLong(row - startRowId, 0);
+    }
+
+    public float getFloat(int row) {
+      return reader.getFloat(row - startRowId, 0);
+    }
+
+    public double getDouble(int row) {
+      return reader.getDouble(row - startRowId, 0);
+    }
+
+    public short getShort(int row) {
+      return reader.getShort(row - startRowId, 0);
+    }
+
+    public void readValues(int[] rows, int rowStartPos, int rowSize, int[] values, int valuesStartPos) {
+      reader.readIntValues(rows, 0, rowStartPos, rowSize, values, valuesStartPos);
+    }
+
+    public void readValues(int[] rows, int rowStartPos, int rowSize, long[] values, int valuesStartPos) {
+      reader.readLongValues(rows, 0, rowStartPos, rowSize, values, valuesStartPos);
+    }
+
+    public void readValues(int[] rows, int rowStartPos, int rowSize, float[] values, int valuesStartPos) {
+      reader.readFloatValues(rows, 0, rowStartPos, rowSize, values, valuesStartPos);
+    }
+
+    public void readValues(int[] rows, int rowStartPos, int rowSize, double[] values, int valuesStartPos) {
+      reader.readDoubleValues(rows, 0, rowStartPos, rowSize, values, valuesStartPos);
+    }
+  }
 
   /**
    *  @param numRowsPerChunk Number of rows to pack in one chunk before a new chunk is created.
@@ -61,8 +140,10 @@ public class FixedByteSingleColumnSingleValueReaderWriter extends BaseSingleColu
     buffer.order(ByteOrder.nativeOrder());
     buffers.add(buffer);
     FixedByteSingleValueMultiColReader reader = new FixedByteSingleValueMultiColReader(buffer, numRowsPerChunk, new int[]{columnSizesInBytes});
-    currentWriter = new FixedByteSingleValueMultiColWriter(buffer, numRowsPerChunk, /*cols=*/1, new int[]{columnSizesInBytes});
-    readers.add(reader);
+    FixedByteSingleValueMultiColWriter writer  = new FixedByteSingleValueMultiColWriter(buffer, numRowsPerChunk, /*cols=*/1, new int[]{columnSizesInBytes});
+    final int startRowId = numRowsPerChunk * (buffers.size()-1);
+    currentWriter = new WriterWithOffset(writer, startRowId);
+    readers.add(new ReaderWithOffset(reader, startRowId));
     writers.add(currentWriter);
     numBytesInCurrentWriter = 0;
   }
@@ -76,10 +157,10 @@ public class FixedByteSingleColumnSingleValueReaderWriter extends BaseSingleColu
 
   @Override
   public void close() throws IOException {
-    for (FixedByteSingleValueMultiColReader reader : readers) {
+    for (ReaderWithOffset reader : readers) {
       reader.close();
     }
-    for (FixedByteSingleValueMultiColWriter writer : writers) {
+    for (WriterWithOffset writer : writers) {
       writer.close();
     }
     for (PinotDataBuffer buffer : buffers) {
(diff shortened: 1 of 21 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -28,23 +29,101 @@
...
+  WriterWithOffset currentWriter;
+  private List<WriterWithOffset> writers = new ArrayList<>();
+  private List<ReaderWithOffset> readers = new ArrayList<>();
   private List<PinotDataBuffer> buffers = new ArrayList<>();
   private final long chunkSizeInBytes;
   private final int numRowsPerChunk;
   private final int columnSizesInBytes;
   private long numBytesInCurrentWriter = 0;
 
+  private static class WriterWithOffset implements Closeable {
```
mayankshriv: Would this be better as a sub class as opposed to inner class?

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -218,29 +247,66 @@ public void readValues(int[] rows, int rowStartPos, int rowSize, int[] values, i
...
+        values[valueIter] = readers.get(bufferId).getInt(row);
+      }
+    }
+  }
+
+  public void readValues(int[] rows, int rowStartPos, int rowSize, long[] values, int valuesStartPos) {
+    if (readers.size() == 1) {
+      readers.get(0).readValues(rows, rowStartPos, rowSize, values, valuesStartPos);
+      return;
+    }
```
mayankshriv: Format the code, seems 'else' is in new line.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvalsets/RealtimeFixedWidthRawValueSet.java
```diff
@@ -0,0 +1,129 @@
...
+  public void getFloatValues(int[] inDocIds, int inStartPos, int inDocIdsSize, float[] outValues, int outStartPos) {
+    int endPos = inStartPos + inDocIdsSize;
+    int outPos = outStartPos;
+    switch (dataType) {
+      case INT:
+        for (int inPos = inStartPos; inPos < endPos; inPos++) {
+          outValues[outPos++] = reader.getInt(inDocIds[inPos]);
+        }
+        break;
+      case LONG:
```
mayankshriv: Should we allow Long to Float conversion?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -67,21 +68,64 @@ public RealtimeColumnStatistics(RealtimeColumnDataSource dataSource, int[] sorte
 
   @Override
   public Object getMinValue() {
+    if (_dictionaryReader == null) {
```
mayankshriv: I understand this is temporary. But just for my understanding, is there any code that actually needs these values in the 'no-dictionary' flow?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -175,6 +184,59 @@ public RealtimeSegmentImpl(Schema schema, int capacity, String tableName, String
     tableAndStreamName = tableName + "-" + streamName;
   }
 
+  private int getColWidth(FieldSpec.DataType dataType) {
```
mayankshriv: Better to use Byte.size instead of 8. Also, seems these sizes are already available in V1Constants.Numbers.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -300,9 +364,29 @@ public boolean index(GenericRow row) {
     for (String metric : dataSchema.getMetricNames()) {
       FixedByteSingleColumnSingleValueReaderWriter readerWriter =
           (FixedByteSingleColumnSingleValueReaderWriter) columnIndexReaderWriterMap.get(metric);
-      int dicId = dictionaryMap.get(metric).indexOf(row.getValue(metric));
-      readerWriter.setInt(docId, dicId);
-      rawRowToDicIdMap.put(metric, dicId);
+      if (isNoDictionaryFixedWidthColumn(metric)) {
+        switch (dataSchema.getFieldSpecFor(metric).getDataType()) {
+          case SHORT:
```
mayankshriv: Do we support SHORT in schema?

pinot-core/src/test/java/com/linkedin/pinot/core/common/RealtimeNoDictionaryTest.java
```diff
@@ -0,0 +1,269 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import com.linkedin.pinot.common.data.FieldSpec;
+import com.linkedin.pinot.common.data.MetricFieldSpec;
+import com.linkedin.pinot.core.io.readerwriter.impl.FixedByteSingleColumnSingleValueReaderWriter;
+import com.linkedin.pinot.core.operator.BaseOperator;
+import com.linkedin.pinot.core.realtime.impl.datasource.RealtimeColumnDataSource;
+
+
+public class RealtimeNoDictionaryTest {
```
mayankshriv: +1 for adding unit tests. Are there tests that cover the query execution path as well?

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
mcvsubbu: Sub class of what?

For now, I would like to leave this as inner class because it is not used anywhere else. As the implementation proceeds, we can evaluate how it turns out and pull it out if necessary. Agreed?

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
mcvsubbu: fixed

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvalsets/RealtimeFixedWidthRawValueSet.java
mcvsubbu: UnSortedSingleValueSet does it.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
mcvsubbu: SegmentIndexCreationDriverImpl.buildIndexCreationInfo(). I believe this is only for metadata purpose, but that is to be verified and the segment creation driver re-factored appropriately to reflect the correct metadata.

After I added this, I realized that there is more to do to get the realtime segment built with artificial metadata, and I did not want this PR to grow too big. So I submitted it disabling the feature.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
mcvsubbu: will do.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
mcvsubbu: Not sure, but I am removing it for now, since I didn't add support for SHORT elsewhere. I think I saw SHORT in some plasces when I starrtarted off, and then not in the query path, so I will just include the 4 types here.

pinot-core/src/test/java/com/linkedin/pinot/core/common/RealtimeNoDictionaryTest.java
mcvsubbu: Only integration test. I added one of the metrics ("NASDelay") as a no-dictionary column, but it breaks right now, so I have commented it out. This metric is used in all types of queries in the hard-coded query set.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvaliterators/RealtimeSingleValueIterator.java
```diff
@@ -71,6 +71,36 @@ public int nextIntVal() {
   }
 
   @Override
+  public long nextLongVal() {
+    if (!hasNext()) {
+      return Constants.EOF;
+    }
+
+    long ret = reader.getLong(counter);
+    return ret;
```
mcvsubbu: missing the counter increment in nextLong(), nextFloat(), nextDouble(). Will fix, and try to add a unit test as well

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -104,8 +106,9 @@
   private static final int MAX_MULTI_VALUES_PER_ROW = 1000;
 
   public RealtimeSegmentImpl(Schema schema, int capacity, String tableName, String segmentName, String streamName,
-      ServerMetrics serverMetrics, List<String> invertedIndexColumns, int avgMultiValueCount)
-      {
+      ServerMetrics serverMetrics, List<String> invertedIndexColumns, int avgMultiValueCount,
+      List<String> noDictionaryColumns)
```
jfim: We'd need to refactor this at some point so that all the indexing info (inverted index, no dictionary, etc.) is together instead of adding constructor args, WDYT?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -26,15 +31,11 @@
...
-import java.util.Arrays;
-import java.util.List;
-import org.apache.commons.lang.math.IntRange;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 
 /**
  * Column statistics for a column coming from an in-memory realtime segment.
+ * TODO Fix this class after we re-factor segment generation code to get some of the stats later on while going through the values
```
jfim: You shouldn't need to change this class at all. What should be done is to create a `NoDictionaryRealtimeColumnStatistics` instead of `RealtimeColumnStatistics` and build the stats as appropriate when it gets created from `RealtimeSegmentStatsContainer`.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -28,23 +29,101 @@
...
   private final int columnSizesInBytes;
   private long numBytesInCurrentWriter = 0;
 
+  private static class WriterWithOffset implements Closeable {
+    final FixedByteSingleValueMultiColWriter writer;
+    final int startRowId;
+
+    private WriterWithOffset(FixedByteSingleValueMultiColWriter writer, int startRowId) {
+      this.writer = writer;
+      this.startRowId = startRowId;
```
mcvsubbu: Each buffer has a startRowNumber. The offset is how far away from row 0 the start of the buffer is.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -28,23 +29,101 @@
...
+    public double getDouble(int row) {
+      return reader.getDouble(row - startRowId, 0);
+    }
+
+    public short getShort(int row) {
+      return reader.getShort(row - startRowId, 0);
+    }
+
+    public void readValues(int[] rows, int rowStartPos, int rowSize, int[] values, int valuesStartPos) {
+      reader.readIntValues(rows, 0, rowStartPos, rowSize, values, valuesStartPos);
```
mcvsubbu: Yes, that is right. I will remove these methods in favor of getting the reader and invoking it directly on the reader. Otherwise, we will need to adjust each row id into its offset.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -218,29 +247,60 @@ public void readValues(int[] rows, int rowStartPos, int rowSize, int[] values, i
...
     if (readers.size() == 1) {
-      readers.get(0).readIntValues(rows, 0, rowStartPos, rowSize, values, valuesStartPos);
+      readers.get(0).readValues(rows, rowStartPos, rowSize, values, valuesStartPos);
       return;
+    } else {
+      for (int rowIter = rowStartPos, valueIter = valuesStartPos;
+          rowIter < rowStartPos + rowSize;
+          rowIter++, valueIter++) {
+        int row = rows[rowIter];
+        int bufferId = getBufferId(row);
```
mcvsubbu: discussed offline

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvalsets/RealtimeFixedWidthRawValueSet.java
```diff
@@ -0,0 +1,129 @@
...
+    super();
+    this.reader = reader;
+    this.length = length;
+    this.dataType = dataType;
+    this.columnName = columnName;
+  }
+
+  @Override
+  public BlockValIterator iterator() {
+    // throw unimplemented?
```
mcvsubbu: I will remove the comment

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
mcvsubbu: wait on this for the next PR.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -26,15 +31,11 @@
...
  */
 public class RealtimeColumnStatistics implements ColumnStatistics {
```
kishoreg: the default values for statistics when we cannot compute needs to be handled properly in the metadata. Please create an issue to track that. Currently, the caller has no way to know if the max value, cardinality etc can be trusted in the metadata.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -67,21 +68,64 @@ public RealtimeColumnStatistics(RealtimeColumnDataSource dataSource, int[] sorte
...
+        default:
+          return new String[]{""};
+      }
+    }
     return _dictionaryReader.getSortedValues();
   }
 
   @Override
   public int getCardinality() {
+    if (_dictionaryReader == null) {
```
kishoreg: this can break something? Doesn't group by etc depend on this. A better default is the number of rows in the segment?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
kishoreg: +1 on this

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -114,6 +117,12 @@ public RealtimeSegmentImpl(Schema schema, int capacity, String tableName, String
     dictionaryMap = new HashMap<String, BaseOnHeapMutableDictionary>();
     maxNumberOfMultivaluesMap = new HashMap<String, Integer>();
     outgoingTimeColumnName = dataSchema.getTimeFieldSpec().getOutgoingTimeColumnName();
+
+    for (final String column : noDictionaryColumns) {
+      if (canBeNoDictionaryColumnInConsumingSegment(column, invertedIndexColumns)) {
```
kishoreg: what is this trying to achieve, please add comments.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
mcvsubbu: I will create an issue. This will have to be in the ColumnMetadata class where we should throw exception if the metadata we have is not correct.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
mcvsubbu: By itself, this cannot break anything because this code path is executed only during segment creation. Nobody relies on actual values that is returned from this method.

However, like I mentioned before, ColumnMetadata class will need to be tweaked. Even there, I have verified that for no-dictionary columns, this method is not accessed. However, yes, it is good to make that explicit if the metadata is not valid.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
mcvsubbu: Not clear what you are asking for here, but I could not name the method any longer :-)

the method call determines if this column can be a no dictionary column for consuming segment. If yes, then we add it to the list, otherwise we end up creating dictionary for the column in the consuming segment. In the completed segment, this column will continue to have be a no-dictionary column. I will add comments saying so.

PR #1464, opened 2017-05-19
[TE] Refine recipient provider factory for external providers

cyenjung:
Recipient provider factory now could instantiate both external and internal recipient providers. It first check the property file (if it exists) and try to instantiate the provider using the given class. If it fails, then it try to instantiate the provider using internal classes. If both fails, then a dummy provider is used.

Tested on local controller.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/alert/grouping/recipientprovider/AlertGroupRecipientProviderFactory.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/alert/grouping/recipientprovider/AlertGroupRecipientProviderFactory.java
@@ -40,14 +40,14 @@ public class AlertGroupRecipientProviderFactory {
    * The constructor that instantiates a factory that has the configuration file to external classes. The configuration
    * file is given by its path.
    */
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/alert/grouping/recipientprovider/AlertGroupRecipientProviderFactory.java
```diff
@@ -1,32 +1,128 @@
...
+  /**
+   * The default constructor that instantiates a factory that does not have the configuration file to external classes.
+   */
+  public AlertGroupRecipientProviderFactory() { }
+
+  /**
+   * The constructor that instantiates a factory that has the configuration file to external classes. The configuration
+   * file is given by its path.
+   */
+  public AlertGroupRecipientProviderFactory(String AlertGroupRecipientProviderConfigPath) {
```
npawar: minor nit: why does this param name starts with a capital letter?

cyenjung: Good catch. I will fix it. :-)

PR #1467, opened 2017-05-19
For whole table reload, send table level message instead of segment level message

Jackie-Jiang:
Currently for whole table reload, we send segment level messgaes, which is causing problem for big tables.
Added table level reload message, which can significantly reduce the number of messages sent.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -1487,41 +1487,57 @@ public class PinotHelixResourceManager {
     return res;
   }
 
-  public int reloadAllSegments(String tableName) {
-    int numMessagesSent = 0;
+  public int reloadAllSegments(@Nonnull String tableNameWithType) {
+    LOGGER.info("Sending reload message for table: {}", tableNameWithType);
 
-    List<String> segments = getSegmentsFor(tableName);
+    // TODO: Remove this part after all servers get updated
+    int numMessagesSent = 0;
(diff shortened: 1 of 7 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1487,41 +1487,48 @@ public PinotResourceManagerResponse addSegment(final SegmentMetadata segmentMeta
...
-    List<String> segments = getSegmentsFor(tableName);
-    for (String segmentName : segments) {
-      numMessagesSent += sendSegmentReloadMessage(tableName, segmentName);
-    }
+    Criteria recipientCriteria = new Criteria();
+    recipientCriteria.setRecipientInstanceType(InstanceType.PARTICIPANT);
+    recipientCriteria.setInstanceName("%");
+    recipientCriteria.setResource(tableNameWithType);
+    recipientCriteria.setSessionSpecific(true);
+    SegmentReloadMessage segmentReloadMessage = new SegmentReloadMessage(tableNameWithType, null);
```
mcvsubbu: this one is setting the segment name to null, but the server is checking for segment name to be ""?

Jackie-Jiang:
@mcvsubbu You are right, will split it into two parts.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: The default value for partition is "". null means do not set the partition name.

PR #1468, opened 2017-05-22
[PINOT-4887] Fixed min/max values for realtime no-dictionary columns

mcvsubbu:


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeNoDictionaryColStatistics.java
```diff
@@ -0,0 +1,208 @@
...
+    int max = Integer.MIN_VALUE;
+    for (int i = 0; i < _numDocIds; i++) {
+      if (values[i] < min) {
+        min = values[i];
+      }
+      if (values[i] > max) {
+        max = values[i];
+      }
+    }
+    _minValue = new Integer(min);
```
jfim: `Integer.valueOf(min)`, not `new Integer`. Same for other object allocations.

PR #1471, opened 2017-05-23
Returning null to fix LLC segment commit when there is no leader

jenniferdai:


pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ControllerLeaderLocator.java
```diff
@@ -80,12 +80,17 @@ public synchronized String getControllerLeader() {
...
-    return _controllerLeaderHostPort;
+    try {
+      ZNRecord znRecord = dataAccessor.get("/" + _clusterName + "/CONTROLLER/LEADER", stat, AccessOption.THROW_EXCEPTION_IFNOTEXIST);
+      String leader = znRecord.getId();
+      int index = leader.lastIndexOf('_');
+      _controllerLeaderHostPort = leader.substring(0, index) + ":" + leader.substring(index + 1);
+      _refresh = false;
+      return _controllerLeaderHostPort;
+    } catch (Exception e) {
+      LOGGER.error("Could not locate controller leader, exception " + e);
```
mcvsubbu: should be a warn, since we correct the situation by retrying eventually.

Also, instead of +e, please do ", e)

PR #1472, opened 2017-05-23
Fixed ColumnarToStarTreeConverter to correctly load the config file.

snleee:
Because loadConfigFiles() is deprecated, StarTreeIndexSpec is not loaded correctly and the converter would always use the default settings even if the config file for the StarTree index is given.

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/ColumnarToStarTreeConverter.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/ColumnarToStarTreeConverter.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.tools.segment.converter;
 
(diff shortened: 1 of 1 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/ColumnarToStarTreeConverter.java
```diff
@@ -118,6 +120,11 @@ private void convertSegment(File columnarSegment)
     config.setStarTreeIndexSpecFile(_starTreeConfigFileName);
     config.setOverwrite(_overwrite);
 
+    // Load the StarTreeSpec object from the config file.
+    ObjectMapper mapper = new ObjectMapper();
```
Jackie-Jiang: Try to define ObjectMapper as private static final member variable to reduce garbage.

PR #1476, opened 2017-05-23
Add support for partition pruning for in-clauses.

rven1:
This will enable pruning of partitioned segments for which there are no values in the in-clause
satisfying the partitioning function for the segment.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
@@ -16,12 +16,15 @@
 package com.linkedin.pinot.core.query.pruner;
 
 import com.linkedin.pinot.common.query.ServerQueryRequest;
+import com.google.common.base.Splitter;
 import com.linkedin.pinot.common.request.FilterOperator;
 import com.linkedin.pinot.common.utils.request.FilterQueryTree;
+import com.linkedin.pinot.core.common.predicate.RangePredicate;
 import com.linkedin.pinot.core.data.partition.PartitionFunction;
 import com.linkedin.pinot.core.indexsegment.IndexSegment;
 import com.linkedin.pinot.core.segment.index.ColumnMetadata;
 import com.linkedin.pinot.core.segment.index.SegmentMetadataImpl;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import org.apache.commons.configuration.Configuration;
@@ -32,6 +35,7 @@ import org.apache.commons.lang.math.IntRange;
  * Implementation of {@link SegmentPruner} that uses partition information to perform pruning.
  */
 public class PartitionSegmentPruner extends AbstractSegmentPruner {
+  public static final String IN_CLAUSE_DELIMITER = "\t\t";
   @Override
   public void init(Configuration config) {
 
@@ -82,8 +86,9 @@ public class PartitionSegmentPruner extends AbstractSegmentPruner {
       return pruneNonLeaf(filterQueryTree, columnMetadataMap);
     }
 
+    FilterOperator operator = filterQueryTree.getOperator();
     // TODO: Enhance partition based pruning for RANGE operator.
-    if (filterQueryTree.getOperator() != FilterOperator.EQUALITY) {
+    if ((operator != FilterOperator.EQUALITY) && (operator != FilterOperator.IN)) {
       return false;
     }
 
@@ -99,8 +104,35 @@ public class PartitionSegmentPruner extends AbstractSegmentPruner {
       return false;
     }
 
-    Comparable value = getValue(filterQueryTree.getValue().get(0), metadata.getDataType());
     PartitionFunction partitionFunction = metadata.getPartitionFunction();
+
+    switch (operator) {
+      case EQUALITY:
+        return pruneEquality(filterQueryTree, metadata, partitionRanges, partitionFunction);
+      case IN:
+        return pruneIn(filterQueryTree, metadata, partitionRanges, partitionFunction);
+      default:
+        return false;
+    }
+  }
+
+  private boolean pruneIn(FilterQueryTree filterQueryTree, ColumnMetadata metadata, List<IntRange> partitionRanges,
+      PartitionFunction partitionFunction) {
+
+    for (String value : filterQueryTree.getValue().get(0).split(RangePredicate.DELIMITER)) {
+      int partition = partitionFunction.getPartition(getValue(value, metadata.getDataType()));
+      for (IntRange partitionRange : partitionRanges) {
+        if (partitionRange.containsInteger(partition)) {
+          return false;
+        }
+      }
+    }
+    return true;
+  }
+
+  private boolean pruneEquality(FilterQueryTree filterQueryTree, ColumnMetadata metadata,
+      List<IntRange> partitionRanges, PartitionFunction partitionFunction) {
+    Comparable value = getValue(filterQueryTree.getValue().get(0), metadata.getDataType());
     int partition = partitionFunction.getPartition(value);
 
     for (IntRange partitionRange : partitionRanges) {
--- a/pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
+++ b/pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
@@ -125,48 +125,62 @@ public class SegmentPartitionTest {
   /**
    * Unit test for {@link PartitionSegmentPruner}.
    * <ul>
-   *   <li> Generates queries with equality predicate on partitioned column with random values. </li>
+   *   <li> Generates queries with given predicate on partitioned column with random values. </li>
+   *   <li> Predicate can be equality or an in clause. </li>
    *   <li> Ensures that column values that are in partition range ([0 5]) do not prune the segment,
    *        whereas other values do. </li>
    *   <li> Ensures that predicates on non-partitioned columns do not prune the segment. </li>
    * </ul>
    */
   @Test
-  public void testPruner() {
+  public void testPruners() {
+    testPruner(FilterOperator.EQUALITY, 1);
+    testPruner(FilterOperator.IN, 2);
+  }
+
+  public void testPruner(FilterOperator operator, int arguments) {
     Pql2Compiler compiler = new Pql2Compiler();
     PartitionSegmentPruner pruner = new PartitionSegmentPruner();
 
+    int[] columnValues = new int[arguments];
     Random random = new Random(System.nanoTime());
     for (int i = 0; i < 1000; i++) {
-      int columnValue = Math.abs(random.nextInt());
+      for (int j =0 ; j < arguments; j++) {
+        columnValues[j] =  Math.abs(random.nextInt());
+      }
 
       // Test for partitioned column.
-      String query = buildQuery(TABLE_NAME, PARTITIONED_COLUMN_NAME, columnValue);
+      String query = buildQuery(TABLE_NAME, PARTITIONED_COLUMN_NAME, columnValues, operator);
       BrokerRequest brokerRequest = compiler.compileToBrokerRequest(query);
       FilterQueryTree filterQueryTree = RequestUtils.generateFilterQueryTree(brokerRequest);
-
-      Assert.assertEquals(pruner.prune(_segment, filterQueryTree), (columnValue % NUM_PARTITIONS > MAX_PARTITION_VALUE),
-          "Failed for column value: " + columnValue);
+      boolean condition = true;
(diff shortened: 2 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -99,8 +103,41 @@ public boolean pruneSegment(FilterQueryTree filterQueryTree, Map<String, ColumnM
...
+    }
+
+    return pruneIn(filterQueryTree, metadata, partitionRanges, partitionFunction);
+  }
+
+  private boolean pruneIn(FilterQueryTree filterQueryTree, ColumnMetadata metadata, List<IntRange> partitionRanges,
+      PartitionFunction partitionFunction) {
+    Splitter valueSplitter = Splitter.on(IN_CLAUSE_DELIMITER);
+
+    List<String> values = valueSplitter.splitToList(filterQueryTree.getValue().get(0));
```
mayankshriv: Why not use String.split?

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -32,6 +34,7 @@
  * Implementation of {@link SegmentPruner} that uses partition information to perform pruning.
  */
 public class PartitionSegmentPruner extends AbstractSegmentPruner {
+  public static final String IN_CLAUSE_DELIMITER = "\t\t";
```
mayankshriv: Isn't the delimiter defined somewhere in the RangePredicate class? If not it belongs there.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -99,8 +103,41 @@ public boolean pruneSegment(FilterQueryTree filterQueryTree, Map<String, ColumnM
...
     }
 
-    Comparable value = getValue(filterQueryTree.getValue().get(0), metadata.getDataType());
     PartitionFunction partitionFunction = metadata.getPartitionFunction();
+
+    if (operator == FilterOperator.EQUALITY) {
+      return pruneEquality(filterQueryTree, metadata, partitionRanges, partitionFunction);
+    }
+
+    return pruneIn(filterQueryTree, metadata, partitionRanges, partitionFunction);
```
mayankshriv: Protect with if (operator == FilterOperator.IN)

raviarin: Should the code not check for Operator type before calling pruneIn function?

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -99,8 +103,41 @@ public boolean pruneSegment(FilterQueryTree filterQueryTree, Map<String, ColumnM
...
+  private boolean pruneIn(FilterQueryTree filterQueryTree, ColumnMetadata metadata, List<IntRange> partitionRanges,
+      PartitionFunction partitionFunction) {
+    Splitter valueSplitter = Splitter.on(IN_CLAUSE_DELIMITER);
+
+    List<String> values = valueSplitter.splitToList(filterQueryTree.getValue().get(0));
+
+    List<Integer> partitions;
+    partitions = new ArrayList<Integer>();
+
+    for (String value : values) {
```
mayankshriv: This is inefficient. We don't need to get all values upfront. That way we can have early bailout.

pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
```diff
@@ -178,6 +178,62 @@ public void testPruner() {
...
+    Pql2Compiler compiler = new Pql2Compiler();
+    PartitionSegmentPruner pruner = new PartitionSegmentPruner();
+
+    Random random = new Random(System.nanoTime());
+    for (int i = 0; i < 1000; i++) {
+      int columnValue = Math.abs(random.nextInt());
+      int columnValue1 = Math.abs(random.nextInt());
+
+      // Test for partitioned column.
+      String query = buildInQuery(TABLE_NAME, PARTITIONED_COLUMN_NAME, columnValue, columnValue1);
```
mayankshriv: Can we use more appropriate names than columnValue and columnValue1?

pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
```diff
@@ -178,6 +178,62 @@ public void testPruner() {
...
+   * Unit test for {@link PartitionSegmentPruner}.
+   * <ul>
+   *   <li> Generates queries with in predicate on partitioned column with random values. </li>
+   *   <li> Ensures that if any column values that are in partition range ([0 5]) do not prune the segment,
+   *        whereas other values do. </li>
+   *   <li> Ensures that predicates on non-partitioned columns do not prune the segment. </li>
+   * </ul>
+   */
+  @Test
+  public void testPrunerIn() {
```
mayankshriv: This seems like a copy paste from testPruner. We should avoid having duplicate code.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -99,8 +103,41 @@ public boolean pruneSegment(FilterQueryTree filterQueryTree, Map<String, ColumnM
...
+  }
+
+  private boolean pruneIn(FilterQueryTree filterQueryTree, ColumnMetadata metadata, List<IntRange> partitionRanges,
+      PartitionFunction partitionFunction) {
+    Splitter valueSplitter = Splitter.on(IN_CLAUSE_DELIMITER);
+
+    List<String> values = valueSplitter.splitToList(filterQueryTree.getValue().get(0));
+
+    List<Integer> partitions;
+    partitions = new ArrayList<Integer>();
```
mayankshriv: Use diamond operator.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -99,8 +103,41 @@ public boolean pruneSegment(FilterQueryTree filterQueryTree, Map<String, ColumnM
...
     PartitionFunction partitionFunction = metadata.getPartitionFunction();
+
+    if (operator == FilterOperator.EQUALITY) {
+      return pruneEquality(filterQueryTree, metadata, partitionRanges, partitionFunction);
+    }
+
+    return pruneIn(filterQueryTree, metadata, partitionRanges, partitionFunction);
+  }
+
+  private boolean pruneIn(FilterQueryTree filterQueryTree, ColumnMetadata metadata, List<IntRange> partitionRanges,
```
mayankshriv: I am a bit uncomfortable with having to create these string garbages per query per segment. For a recent high throughput case, I found this to be creating a lot of garbage, and I had modified the pruner to take an already created filterQueryTree as opposed to creating it over and over again per segment. I fear this change will cause the same issue.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: We have already returned if it is not one of  EQUALITY or IN operator earlier. I will add another test just to be clearer.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: We are able to do a query with 35 items and 60 features in about 80 ms vs more than 5 seconds earlier

rven1: Changed delimiter to use the one from Rangepredicate

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: Done

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: Done

pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
rven1: Done

pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
rven1: They test two different features with different characteristics even though the flow is similar.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
mayankshriv: There are more than necessary 'if' statements here. And it will complicate the code as more predicates are added. Better to have just one switch-case so that there's no more than one branching per predicate.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
mayankshriv: I don't see how this change can improve latency from 5seconds down to 80ms. Without this change, the only cost to pay is for query planning, where planner should have bailed out determining that the predicate cannot be satisfied at all. Pruner only helps with throughput by avoiding the scheduling of threads for planning. Also, if the the planner did not bail out early, still 5 seconds seems super high, in which case we should understand where the time is being spent and fix that as well.

pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
mayankshriv: What happens when we have more predicates? We wouldn't want to copy-paste-slightly-modify for which of those predicates, right? We should refactor the code so that common parts can be reused, and only test specific code is written within the individual test.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: Added switch

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: This change increases the qps. Currently for a query on a single partition, the query  still spends several milliseconds on machines without the partition.

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/PartitionSegmentPruner.java
rven1: Currently we are calling split for each segment for each in clause in DictionaryBasedInPredicateEvaluator in its call to getInRange. We are just moving all the evaluations to the pruner, so that a single thread does the evaluation freeing up the other threads to do more queries.

pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/creator/SegmentPartitionTest.java
rven1: Refactored.

npawar:

PR #1477, opened 2017-05-23
[PINOT-4887] Ensure that we put in an invalid value for cardinality in metadata, and throw exception when invalid value is present.

mcvsubbu:
When realtime segment has no-dictionary columns, we do not compute the cardinality in RealtimeNoDictionaryColStatistics.
Instead, we return Integer.MIN_VALUE. This value gets written to the segment metadata when the segment is sealed.

When loading the metadata, we want to ensure that the bad number is not used for any processing, so we throw an
exception when it is accessed


pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/defaultcolumn/BaseDefaultColumnHandler.java
```diff
@@ -266,6 +266,7 @@ protected void createColumnV1Indices(String column)
     int dictionaryElementSize = 0;
 
     Object sortedArray;
+    final int cardinality = 1;
```
jenniferdai: I might have missed something, but why is this set to 1? And where do we set the cardinality correctly when it does have dictionary columns? Am I correct in assuming that we just added the cardinality field, and it wasn't there before?

mcvsubbu: Prior to this change, the callers to construct ColumnIndexCreationInfo() had to provide an array of unique values as one of the constructor arguments. Internally, ColumnIndexCreationInfo computed the cardinality as array.length.

With this change, I have introduced an additional argument to the ctor of ColumnIndexCreationInfo -- cardinality.

If you see the lines below, the array of unique values is computed as a one-element array consisting of the default value of the column being added. Since the size of this array is 1, I am hard-setting the cardinality to 1 and passing that value in the constructor. I did not want to hard-code the value in the call to constructor.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/ColumnMetadata.java
```diff
@@ -430,6 +429,13 @@ public String getColumnName() {
     return columnName;
   }
 
+  /**
+   * When a realtime segment has no-dictionary columns, the cardinality for those columns will be
+   * set to Integer.MIN_VALUE.
+   *
+   * @return The cardinality of the column.
+   * @throws UnsupportedOperationException if cardinality is not available.
```
antumbde: remote @throws comment

PR #1482, opened 2017-05-24
Add TableTaskConfig into TableConfig for task config

Jackie-Jiang:
Added TableTaskConfig into TableConfig for task config (used by minion service)
Updated minion integration test to pick up and use the TableTaskConfig

Added TableConfig.Builder inner class to build table config
Modified tests to use the new TableConfig.Builder to build table config
Removed redundant 'lazyload' and 'segmentPushFrequency' configs


pinot-common/src/main/java/com/linkedin/pinot/common/config/TableConfig.java
```diff
@@ -184,22 +198,44 @@ public void setQuotaConfig(@Nullable QuotaConfig quotaConfig) {
...
     JSONObject jsonConfig = new JSONObject();
-    jsonConfig.put(TABLE_NAME_KEY, _tableName);
-    jsonConfig.put(TABLE_TYPE_KEY, _tableType.toString().toLowerCase());
-    jsonConfig.put(VALIDATION_CONFIG_KEY, new JSONObject(OBJECT_MAPPER.writeValueAsString(_validationConfig)));
-    jsonConfig.put(TENANT_CONFIG_KEY, new JSONObject(OBJECT_MAPPER.writeValueAsString(_tenantConfig)));
-    jsonConfig.put(INDEXING_CONFIG_KEY, new JSONObject(OBJECT_MAPPER.writeValueAsString(_indexingConfig)));
-    jsonConfig.put(CUSTOM_CONFIG_KEY, new JSONObject(OBJECT_MAPPER.writeValueAsString(_customConfig)));
-    if (_quotaConfig != null) {
-      jsonConfig.put(QUOTA_CONFIG_KEY, new JSONObject(OBJECT_MAPPER.writeValueAsString(_quotaConfig)));
+    jsonConfig.put(TABLE_NAME_KEY, tableName);
```
mayankshriv: We should avoid creating a JSONObject from strings equivalent of individual fields. It would be better if we can directly use object mapper to generate json object from the TableConfig object. Not a blocker for this PR, because that is likely a bigger change.

pinot-common/src/main/java/com/linkedin/pinot/common/config/helper/TableConfigJSONBuilder.java
```diff
@@ -0,0 +1,227 @@
...
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+
+public class TableConfigJSONBuilder {
```
mayankshriv: Seems like this class is build a JSON object of the table config. Why not directly build the actual TableConfig object?

pinot-tools/src/main/java/com/linkedin/pinot/tools/query/comparison/ClusterStarter.java
```diff
@@ -200,10 +200,16 @@ private void addTable()
     }
 
     String controllerAddress = "http://" + _localhost + ":" + _controllerPort;
```
mayankshriv: This can be hidden inside ControllerRequestURLBuilder.

pinot-common/src/main/java/com/linkedin/pinot/common/config/helper/TableConfigJSONBuilder.java
Jackie-Jiang: I agree, and made the change.

pinot-tools/src/main/java/com/linkedin/pinot/tools/query/comparison/ClusterStarter.java
Jackie-Jiang: This is out of the scope of this PR, will address in another PR if necessary.

PR #1483, opened 2017-05-24
Make auto onboard service generic to onboard from all data sources

npawar:
AutoLoadPinot service has changed to autoLoadService. Each data source config, can specify an auto onboard class. The AutoLoadService will periodically run each data source's autoload.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auto/onboard/AutoOnboardService.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auto/onboard/AutoOnboardService.java
@@ -5,13 +5,12 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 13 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auto/onboard/AutoOnboardService.java
```diff
@@ -0,0 +1,71 @@
...
+          autoOnboardServices.add(autoOnboardConstructor);
+        } catch (Exception e) {
+          LOG.error("Exception in creating autoload constructor {}", autoLoadClassName);
+        }
+      }
+    }
+  }
+
+  public void start() {
+    scheduledExecutorService.scheduleAtFixedRate(this, 0, 15, TimeUnit.MINUTES);
```
apucher: should this be configurable?

npawar: Yes would be nice to have. Fixed

PR #1484, opened 2017-05-24
[TE] Enable Merge For Two Sided SignTest

cecilynie:
Add isMergeable logic to time based merge model. By comparing two anomalies' detection properties and decide if mergeable.


cecilynie:
This PR is a better fix version that can overwrite https://github.com/linkedin/pinot/pull/1454

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/AnomalyDetectionUtils.java
```diff
@@ -12,4 +16,23 @@
   public static long getBucketInMillis(int bucketSize, TimeUnit bucketUnit) {
     return bucketUnit.toMillis(bucketSize);
   }
+
+  private static Splitter SEMICOLON_SPLITTER = Splitter.on(";").omitEmptyStrings();
+  private static Splitter EQUALS_SPLITTER = Splitter.on("=").omitEmptyStrings();
+
+  /**
+   * Decode properties string to into Hashmap
+   * @param propStr: property string which is encoded using com.linkedin.anomalydetection.utils.StringUtil.encodeCompactedProperties
```
bl44: You'd rather move the whole files here. 
StringUtil.java, you do not need to leave a comments like this in a open source project.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
```diff
@@ -132,5 +136,13 @@ private static void populateMergedResult(MergedAnomalyResultDTO mergedAnomaly,
...
     mergedAnomaly.setEndTime(currentResult.getEndTime());
     mergedAnomaly.setCreatedTime(System.currentTimeMillis());
+    // populate current result's property as well, will be used to identify if two anomalies can be merged by comparing their property
+    mergedAnomaly.setProperties(AnomalyDetectionUtils.decodeCompactedPropertyStringToMap(currentResult.getProperties()));
+  }
+
+  // compare if two anomalies have same property when doing anomaly detection, if from same detection configuration then is mergeable
+  private static boolean isMergeable(MergedAnomalyResultDTO anomaly1, MergedAnomalyResultDTO anomaly2){
+    return ((anomaly1.getProperties() == null && anomaly2.getProperties() == null)
+        || (anomaly1.getProperties() != null && anomaly1.getProperties().equals(anomaly2.getProperties()) ));
```
bl44: Only part of the properties need to match, not all of them. Otherwise the properties as a field will be restricted. My original design is  to use the test_config (key, value) as a unique identify of the detection config. You may following this idea and come up with a better one.

cecilynie: I have reduced the "property" to contain only the configurations.  There are two facts that I see when doing integration test:
1. Repeated information are being stored in both anomaly.getProperties() as well as anomaly.getFunction().getProperties(). They actually were storing the same function settings, i.e. signTestPattern:"UP,DOWN". Then I re-designed to put only the one side function configuration into anomaly.getProperties()
2. It is hard to decode test_config (key,value) when do update in merge, we want to use the specific one side configuration rather than the two sided anomaly function property. If we really want to recover tag to lift/shift values, we need anther encoder and decoder for anomaly function config tag.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/MergedAnomalyResultBean.java
```diff
@@ -23,6 +23,7 @@
   private double weight; // change percentage, whose absolute value is severity
   private double impactToGlobal; // the impact of this anomaly to the global metric
   // Additional anomaly detection properties (e.g., patter=UP, etc.)
+  // Being used as identifying if two merged anomalies have same anomaly detection properties, thus can be mergeable
```
bl44: Please remove this comments without actual code change.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
```diff
@@ -71,16 +74,18 @@ public int compare(RawAnomalyResultDTO o1, RawAnomalyResultDTO o2) {
         populateMergedResult(mergedAnomaly, currentResult);
       } else {
         // compare current with merged and decide whether to merge the current result or create a new one
-        if (applySequentialGapBasedSplit
-            && (currentResult.getStartTime() - mergedAnomaly.getEndTime()) > sequentialAllowedGap) {
+        MergedAnomalyResultDTO currMergedAnomaly = new MergedAnomalyResultDTO();
+        populateMergedResult(currMergedAnomaly, currentResult);
+        if ((applySequentialGapBasedSplit
```
bl44: This condition may need some comment to explanation: you defined isMergeable here, but (Not) isMergeable and other two conditions (which defines isOverlap) together defines (shouldBeSeparated).

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/AnomalyDetectionUtils.java
```diff
@@ -12,4 +19,36 @@
...
     return bucketUnit.toMillis(bucketSize);
   }
+
+  private static Splitter SEMICOLON_SPLITTER = Splitter.on(";").omitEmptyStrings();
+  private static Splitter EQUALS_SPLITTER = Splitter.on("=").omitEmptyStrings();
+  private static Joiner SEMICOLON = Joiner.on(";");
+  private static Joiner EQUALS = Joiner.on("=");
+
+  // the following two functions encode and decode properties are moved from class
+  // com.linkedin.thirdeye.controller.mp.function.ThirdEyeAnomalyFunctionUtil
```
cyenjung: You don't need to JavaDoc where is this method moved from; however, please JavaDoc what does this method do.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
```diff
@@ -132,5 +137,17 @@ private static void populateMergedResult(MergedAnomalyResultDTO mergedAnomaly,
     mergedAnomaly.setStartTime(currentResult.getStartTime());
     mergedAnomaly.setEndTime(currentResult.getEndTime());
     mergedAnomaly.setCreatedTime(System.currentTimeMillis());
+    // populate current result's property as well, will be used to identify if two anomalies can be merged by comparing their property
+    mergedAnomaly.setProperties(AnomalyDetectionUtils.decodeCompactedPropertyStringToMap(currentResult.getProperties()));
+  }
+
+  // compare if two anomalies have same property when doing anomaly detection, if from same detection configuration then is mergeable
+  private static boolean isMergeable(MergedAnomalyResultDTO anomaly1, MergedAnomalyResultDTO anomaly2){
+    // If both of anomalies don't have key TEST_CONFIG_TAG, they are mergeable;
```
cyenjung: This is a very dangerous black magic: No one would know isMergeable is enabled unless there is a magic keyword in the properties of anomaly result.

I would create a new method in anomaly function: List<String> getTestIsMergeableProperties(), which returns a list of keys that could be used to retrieve the properties for determining isMergeable. If an anomaly function returns an empty list, then this function's anomalies could be mergeable regardless their properties; otherwise, we compare all values of the test keys.

Please refine the design of this method.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/MergedAnomalyResultBean.java
cyenjung: Not all values in properties are used for determining isMergeable. The values to be used depends on the anomaly function. Please see my comment for the method, isMergeable, and redesign this part.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
```diff
@@ -71,16 +74,18 @@ public int compare(RawAnomalyResultDTO o1, RawAnomalyResultDTO o2) {
         populateMergedResult(mergedAnomaly, currentResult);
       } else {
         // compare current with merged and decide whether to merge the current result or create a new one
-        if (applySequentialGapBasedSplit
-            && (currentResult.getStartTime() - mergedAnomaly.getEndTime()) > sequentialAllowedGap) {
+        MergedAnomalyResultDTO currMergedAnomaly = new MergedAnomalyResultDTO();
```
cecilynie: fixed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/AnomalyDetectionUtils.java
cecilynie: fixed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/MergedAnomalyResultBean.java
cecilynie: fixed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
```diff
@@ -71,16 +74,18 @@ public int compare(RawAnomalyResultDTO o1, RawAnomalyResultDTO o2) {
...
+        populateMergedResult(currMergedAnomaly, currentResult);
+        if ((applySequentialGapBasedSplit
+            && (currentResult.getStartTime() - mergedAnomaly.getEndTime()) > sequentialAllowedGap)
+            || (!isMergeable(mergedAnomaly, currMergedAnomaly))) {
```
cecilynie: Keep the logic as it is here, and change the function isMergeable to isEqualOnMergeableKeys.
This if-statement is doing a comparison on sequentialAllowedGap and if is equal on mergeable keys.
See line 109 is comparing on maxMergedDurationMillis
The AnomalyMergedConfig is fully used in this block of code, to really extract the isMergeable function needs to refine the whole block rather than simply this if-statement, hence changed the function name to specify the specific condition.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
cecilynie: See the comment below

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/MergedAnomalyResultBean.java
cecilynie: See the new commit for re-design.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyMergeConfig.java
```diff
@@ -19,6 +22,17 @@
   private AnomalyMergeStrategy mergeStrategy = AnomalyMergeStrategy.FUNCTION;
   private long sequentialAllowedGap = 30_000; // 30 seconds
   private long maxMergeDurationLength = 12 * 60 * 60 * 1000; // 12 hours
+  private List<String> mergeablePropertyKeys = new ArrayList<>();
+
+  public AnomalyMergeConfig() {
+  }
+
+  public AnomalyMergeConfig(AnomalyMergeConfig other) {
```
bl44: somehow it looks weird to have this kind of constructor. It might be better to define:
mergeConifg(mergeStrategey, gap, length, keys) and use other.* as input when you need to construct a new config with an existing config.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
```diff
@@ -74,18 +78,21 @@ public int compare(RawAnomalyResultDTO o1, RawAnomalyResultDTO o2) {
...
-        populateMergedResult(currMergedAnomaly, currentResult);
+        MergedAnomalyResultDTO currAnomaly = new MergedAnomalyResultDTO();
+        populateMergedResult(currAnomaly, currentResult);
+        // if the merging is applying sequential gap and current anomaly has gap time larger than sequentialAllowedGap
+        // or current anomaly is not equal on mergeable keys with mergedAnomaly
+        // should not merge the two and split from here
         if ((applySequentialGapBasedSplit
             && (currentResult.getStartTime() - mergedAnomaly.getEndTime()) > sequentialAllowedGap)
-            || (!isMergeable(mergedAnomaly, currMergedAnomaly))) {
+            || (!isEqualOnMergeableKeys(mergedAnomaly, currAnomaly, mergeablePropertyKeys))) {
```
bl44: you may want to define a new variable name or function to wrapper the if condition:
isNotMergable(*)

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
```diff
@@ -132,5 +140,35 @@ private static void populateMergedResult(MergedAnomalyResultDTO mergedAnomaly,
...
+    Map<String, String> prop2 = anomaly2.getProperties();
+    // degenerate case
+    if(mergeableKeys.size() == 0 ||
+        (prop1.isEmpty() && prop2.isEmpty())){
+      return true;
+    }
+    // If both of anomalies have mergeable keys and the contents are equal, they are mergeable;
+    // Otherwise it's indicating the two anomalies are detected by different function configurations, they are not mergeable
+    for (String key : mergeableKeys) {
+      if (!prop1.containsKey(key) || !prop2.containsKey(key) || !prop1.get(key).equals(prop2.get(key))) {
```
cyenjung: What if both prop1 and prop2 do not contain "key" or return a null value for "key"? It should return true because the definition at line 162.

I believe this is the if-statement you are looking for:
if (!Objects.equals(prop1.get(key), prop2.get(key))) {
  return false;
}

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyMergeConfig.java
cyenjung: Good suggestion. @cecilynie, could you change this constructor to take variables instead of another config object?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/TimeBasedAnomalyMerger.java
```diff
@@ -89,6 +89,16 @@ public TimeBasedAnomalyMerger(AnomalyFunctionFactory anomalyFunctionFactory) {
       mergeConfig = DEFAULT_TIME_BASED_MERGE_CONFIG;
     }
 
+    // From anomaly function spec initialize the anomaly function, and get its specified mergeable property keys if any
+    // If failed to create anomaly function, default empty mergeable property keys will be used from mergeConfig
+    try {
+      BaseAnomalyFunction anomalyFunction = anomalyFunctionFactory.fromSpec(functionSpec);
+      mergeConfig.setMergeablePropertyKeys(anomalyFunction.getMergeablePropertyKeys());
+    } catch (Exception e) {
+      e.printStackTrace();
```
cyenjung: Log this exception.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/function/BaseAnomalyFunction.java
```diff
@@ -163,4 +163,16 @@ private AnomalyOffset getDefaultOffsets(DatasetConfigDTO datasetConfig) {
...
+
+  /**
+   * Specify the property keys used for merge comparison
+   * Each anomaly detection function will be capable to overwrite this function to return the key lists when merging two anomalies
+   * If two anomalies don't have equal value on the specified keys, they won't be merged
+   * If this function is not being overwritten, empty list will be returned and anomalies won't be compared on mergeable keys when merging
+   * @return A list of keys to used for comparing if two anomalies are equal on mergeable keys
+   */
+  public List<String> getMergeablePropertyKeys(){
+    return new ArrayList<>();
```
cyenjung: Use "return Collections.emptyList();"

cecilynie: fixed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
```diff
@@ -132,5 +140,35 @@ private static void populateMergedResult(MergedAnomalyResultDTO mergedAnomaly,
...
+   * @param anomaly2 The second anomaly result
+   * @param mergeableKeys keys that passed by AnomalyMergeConfig, which is defined by Anomaly Detection Function
+   * @return true if two anomalies are equal on mergeable keys, otherwise return false
+   */
+  private static boolean isEqualOnMergeableKeys(MergedAnomalyResultDTO anomaly1, MergedAnomalyResultDTO anomaly2, List<String> mergeableKeys){
+    Map<String, String> prop1 = anomaly1.getProperties();
+    Map<String, String> prop2 = anomaly2.getProperties();
+    // degenerate case
+    if(mergeableKeys.size() == 0 ||
+        (prop1.isEmpty() && prop2.isEmpty())){
```
cyenjung: This assumption might not always hold. It could be accidentally changed to null at somewhere.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
cyenjung: Change it to warning. Reserve error for operational or production issues.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
cecilynie: Changed to use Map.Utils to avoid cases if props are null

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/merge/AnomalyTimeBasedSummarizer.java
cecilynie: It was set to warning for the latest commit. :)

PR #1485, opened 2017-05-25
[PINOT-4602] Changed servers to re-use segment files if available, when asked to re-commit an LLC segment

mcvsubbu:
It can happen that a servers overshoot the lease (due to GC, for example) and we end up with the controller
asking a server to commit the segment again. In that case, if the offset is still the same, we keep the tar
file around and re-use it when committing the segment.

We take care to remove the tar file when we go ONLINE for a segment, so that old files are cleaned up.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -16,6 +16,21 @@
 
 package com.linkedin.pinot.core.data.manager.realtime;
 
+import java.io.Closeable;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicLong;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.linkedin.pinot.common.config.IndexingConfig;
@@ -50,22 +65,7 @@ import com.linkedin.pinot.core.realtime.impl.kafka.SimpleConsumerWrapper;
 import com.linkedin.pinot.core.segment.index.loader.IndexLoadingConfig;
 import com.linkedin.pinot.server.realtime.ServerSegmentCompletionProtocolHandler;
 import com.yammer.metrics.core.Meter;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicLong;
 import kafka.message.MessageAndOffset;
-import org.apache.commons.io.FileUtils;
-import org.apache.commons.lang3.tuple.Pair;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 
 /**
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -126,6 +126,28 @@ public boolean isFinal() {
...
+    public long getOffset() {
+      return _offset;
+    }
+
+    public String getSegmentFile() {
+      return _segmentFile;
+    }
+
+    @Override
+    public void close() {
```
jfim: I wouldn't use `close` for this, since it's not really releasing a resource. `deleteSegmentFile` is a better name for this.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -582,43 +612,51 @@ private boolean buildSegmentInternal(boolean forCommit) {
...
-    } catch (FileNotFoundException e) {
-      segmentLogger.error("Tar file {} not found", segTarFileName, e);
+    File segTarFile = new File(segTarFileName);
+    if (!segTarFile.exists()) {
+      throw new RuntimeException("Segment file does not exist:" + segTarFileName);
+    }
+    SegmentCompletionProtocol.Response response =  postSegmentCommitMsg(segTarFile);
+    if (!response.getStatus().equals(SegmentCompletionProtocol.ControllerResponseStatus.COMMIT_SUCCESS)) {
+      segmentLogger.warn("Received controller response {}", response);
+      // Save the segment file if commit fails
```
jfim: I'm not sure I follow this logic. If the commit fails, you remove the segment file but you still keep it?

I'm assuming in this case that you're only keeping track of the segment file and offset at commit time. It would be easier to keep track of it at build time (ie. when you build the segment, you keep track of the filename and offset) and have the build logic skip the segment build if it would build the same thing; that way the logic is only in build and half in commit and half in build.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: +1

rven1: +1

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -484,12 +510,12 @@ public void run() {
               _serverMetrics.addMeteredTableValue(_metricKeyName, ServerMeter.LLC_CONTROLLER_RESPONSE_COMMIT, 1);
               _state = State.COMMITTING;
               long buildTimeSeconds = response.getBuildTimeSeconds();
-              success = buildSegment(true, buildTimeSeconds*1000L);
-              if (!success) {
+              final String segmentTarFile = buildSegmentForCommit(buildTimeSeconds * 1000L);
```
rven1: why have final here? Won't this be lost once we leave the switch

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -528,29 +556,31 @@ private File makeSegmentDirPath() {
...
     } finally {
-      if (forCommit) {
-        _leaseExtender.removeSegment(_segmentNameStr);
-      }
+      _leaseExtender.removeSegment(_segmentNameStr);
     }
   }
 
-  private boolean buildSegmentInternal(boolean forCommit) {
+  protected String buildSegmentInternal(boolean forCommit) {
```
rven1: do we ever buildSegmentInternal not forCommit?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
rven1: Just use _segmentFileAndOffset throughout, set it on build, reset on successful commit or consumption

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mcvsubbu: will change it

mcvsubbu: will change

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mcvsubbu: It doesn't matter either way.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mcvsubbu: Yes, lookup method hierarchy in your favorite IDE

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mcvsubbu: Hmm... let me think about this and incorporate it. It does sound simpler.

PR #1488, opened 2017-05-25
Fix the issue where cache does not get cleared properly when setting fields in schema

Jackie-Jiang:


pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
```diff
@@ -58,13 +58,14 @@
...
+
+  // Json ignored fields
   private final Map<String, FieldSpec> _fieldSpecMap = new HashMap<>();
   private final Set<String> _dimensionSet = new HashSet<>();
   private final Set<String> _metricSet = new HashSet<>();
   private final List<String> _dimensionList = new ArrayList<>();
   private final List<String> _metricList = new ArrayList<>();
-
-  private transient String _jsonSchema;
+  private String _jsonSchema;
```
mayankshriv: I feel it is better to just remove this field, because the call to get JSON schema is not on any critical paths (eg query execution, segment generation).

Jackie-Jiang: I felt the same way. Removed the field.

PR #1504, opened 2017-05-31
te_endpoints_anomaly

bl44:
clean up endpoints for get / delete anomalies in time range


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/OnboardResource.java
```diff
@@ -257,11 +283,11 @@ public Boolean cloneAnomalyInstances(Long srcId, Long destId) {
   @POST
   @Path("function/{id}/deleteExistingAnomalies")
   public Map<String, Integer> deleteExistingAnomalies(@PathParam("id") String id,
```
cyy0714: We should also unify the type of id here. Use long instead of String.

PR #1508, opened 2017-06-01
[PINOT-4889] Preparation for off-heap mutable dictionary for Realtime consuming segments.

mcvsubbu:
Introduce a new base class MutableDictionary that gets exported whether the underlying implementation is
on-heap or off-heap.

Added a close() method to the dictionary, since the off-heap structures need to be released when the segment
is destroyed. Added a call to close() in the destroy() method for RealtimeSegmentImpl

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
@@ -50,7 +50,7 @@ import com.linkedin.pinot.core.io.readerwriter.impl.FixedByteSingleColumnMultiVa
 import com.linkedin.pinot.core.io.readerwriter.impl.FixedByteSingleColumnSingleValueReaderWriter;
 import com.linkedin.pinot.core.realtime.RealtimeSegment;
 import com.linkedin.pinot.core.realtime.impl.datasource.RealtimeColumnDataSource;
-import com.linkedin.pinot.core.realtime.impl.dictionary.BaseOnHeapMutableDictionary;
+import com.linkedin.pinot.core.realtime.impl.dictionary.MutableDictionary;
 import com.linkedin.pinot.core.realtime.impl.dictionary.MutableDictionaryFactory;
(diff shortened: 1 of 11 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -520,13 +520,21 @@ public void destroy() {
         LOGGER.error("Failed to close inverted index. Service will continue with memory leaks, error: ", e);
       }
     }
+
+    for (Map.Entry<String, MutableDictionary> entry : dictionaryMap.entrySet()) {
```
Jackie-Jiang: dictonaryMap.values()?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/MutableDictionary.java
```diff
@@ -0,0 +1,89 @@
...
+
+package com.linkedin.pinot.core.realtime.impl.dictionary;
+
+import java.io.Closeable;
+import java.io.IOException;
+import com.linkedin.pinot.core.segment.index.readers.Dictionary;
+import javax.annotation.Nonnull;
+
+
+public abstract class MutableDictionary implements Dictionary, Closeable {
```
Jackie-Jiang: I recommend name it as BaseMutableDictionary so by name we can tell it's an abstract class, up to you.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
mcvsubbu: If close fails, I would like to log an error with the column name. So, it is better to get the entries rather than just values. entry.key() is the column name.

Jackie-Jiang: Oh, you are right, I missed that

PR #1511, opened 2017-06-02
[TE] Fix the decoration for filter set when query mixes non and additive datasets

cyenjung:
Fix the filter decoration when the query mixes metrics from non-additive and additive datasets. Usually, this kind of queries is used for calculating the derived metric.

Problem:
Previous code assumes that all metric comes from the same dataset. Therefore, it applies the same filter decoration among all metrics in the same query.

Fix:
The current fix checks the origin dataset for each metric and decorate the filter accordingly.

Tested on local dashboard and compared the data with rApTOr.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datasource/pinot/PinotThirdEyeDataSource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datasource/pinot/PinotThirdEyeDataSource.java
@@ -1,12 +1,16 @@
 package com.linkedin.thirdeye.datasource.pinot;
 
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datasource/pinot/PinotThirdEyeDataSource.java
```diff
@@ -123,6 +137,49 @@ public PinotThirdEyeResponse execute(ThirdEyeRequest request) throws Exception {
...
+   * dimension filters such that the filter could pick out the "all" value for that dimension.
+   * Example: Suppose that we have a dataset with 3 dimensions: country, pageName, and osName, and the pre-aggregated
+   * keyword is 'all'. Further assume that the original request's filter = {'country'='US, IN'} and GroupBy dimension =
+   * pageName, then the decorated request has the new filter = {'country'='US, IN', 'osName' = 'all'}.
+   *
+   * @param filterSet the original filterSet, which will NOT be modified.
+   *
+   * @return a decorated filter set for the given precomputed dataset.
+   */
+  public static Multimap<String, String> generateDecoratedFilterSetForPrecomputedDataset(
```
npawar: Maybe it's just me, but "decorate filter set" doesn't really convey that we are putting in pre aggregate value. Since you're touching this part, should we rename it? What do you think?

cyenjung: Sure, let me rename it.

PR #1513, opened 2017-06-02
[PINOT-4889] Off-heap mutable dictionary for fixed width columns

mcvsubbu:
Forward mapping of index to value is maintained using the FixedWidth reader-writers.
The reverse mapping is maintained using IntBuffer(s). The dictionary ID is stored as an integer
in the IntBuffer at an offset that is derived by hashing the actual value.

Collisions are handled by creating new IntBuffers, potentially larger than the previous one created.
A small overflow is maintained in heap so that we don't create new buffers on a single collision.

The current perf suite on dictionary performance is not correct for micro-benchmark.
However, a few brif runs and eye-balling results seem to show:
* Overall performance of off-heap implementation does not degrade with cardinality
 (probably because we pre-allocate the table for that cardinality in the reverse map).
* At high cardinality (> 2M) indexing in off-heap is faster (3x?).
* Value fetching time is slower in off-heap (4x?)
* Dictionary id fetching time is a little slower in off-heap for small cardinalities (< 2M) but
  faster for higher cardinality.

That said, we need to have a better perf measuring program for comparisons.

TODO:
    Limit the number of buffers created (and close the segment), so that we minimize performance
    impact on lookup. This will be done in subsequent PRs when the offheap implementation is weaved into
    RealtimeSegmentImpl

    Do some performance studies and set num-columns, overflow-threshold, expansion-multiples etc.


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,411 @@
...
+    }
+    return dictId;
+  }
+
+  protected void indexValue(@Nonnull Object value) {
+    final int hashVal = Math.abs(value.hashCode());
+    ValueToDictId valueToDictId = _valueToDict.get();
+    final List<IntBuffer> iBufList = valueToDictId.getIBufList();
+
+    for (IntBuffer iBuf : iBufList) {
```
mayankshriv: I am probably reading this incorrectly, but seems that the number of iBuf's ~= cardinality? If so, is this a linear search?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,411 @@
...
+public abstract class BaseOffHeapMutableDictionary extends MutableDictionary {
+  // Percent of cardinality that we expect to hold in the overflow hash table.
+  private static final int OVERFLOW_THRESHOLD_PERCENT = 1;
+
+  // expansionMultiple setting as we add new buffers. A setting of 1 sets the new buffer to be
+  // the same size as the last one added. Setting of 2 allocates a buffer twice as big as the
+  // previous one. It is assumed that these values are powers of 2. It is a good idea to restrict
+  // these to 1 or 2, but not higher values. This array can be arbitrarily long. If the number of
+  // buffers exceeds the number of elements in the array, we use the last value.
+  private static final int[] EXPANSION_MULTIPLES = new int[] {1, 1, 2, 2, 2};
```
mcvsubbu: Current thought is NOT to have these come via config, but make this a learning system where one segment "learns" from the previously built segment. I don't have actual mechanisms to do that yet, so I will leave this code as is. If a config is eventually needed, we can plumb it in. 

As for benchmarking, as my check-in remark says, the current benchmark tool is not quite valid, but I have preliminary results based on that. We need to write a benchmark using JMH. I will do that in a different check-in.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,411 @@
...
+  private final int _sizeOfFirstBuf;
+
+  /**
+   * A class to hold all the objects needed for the reverse mapping.
+   */
+  private static class ValueToDictId {
+    // Each iBuf layout is as described in comments above.
+    private final List<IntBuffer> _iBufList;
+    // The map should be a concurrent one.
+    private final Map<Object, Integer> _overflowMap;
```
mcvsubbu: The map (as the name indicates), is for overflow elements. It is sized to be 1% of the cardinality at this time.

I was hoping that the comments in the beginning of the file will help understand this, but let me know how I can improve the comments so that it is evident that the size of this overflow map is bounded.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,411 @@
...
+      return _iBufList;
+    }
+    private Map<Object, Integer> getOverflowMap() {
+      return _overflowMap;
+    }
+  }
+
+  private AtomicReference<ValueToDictId> _valueToDict = new AtomicReference<>();
+
+  protected BaseOffHeapMutableDictionary(int estimatedCardinality) {
```
mcvsubbu: Like I mentioned before, the current plan is to start with a guess and then learn from the previous segment. Mechanisms TBD.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,411 @@
...
+  @Override
+  public void close() throws IOException {
+    doClose();
+
+    ValueToDictId valueToDictId = _valueToDict.get();
+    _valueToDict.set(null);
+    Map<Object, Integer> overflowMap = valueToDictId.getOverflowMap();
+    overflowMap.clear();
+    List<IntBuffer> iBufs = valueToDictId.getIBufList();
+    iBufs.clear();
```
mcvsubbu: In line 249

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
mcvsubbu: Not sure what ~= means, but yes, this is a linear search. In practice, we will expect to have a few iBufs in the first segment, and 1 (max 2) in subsequent segments.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
jfim: As mentioned before, I very strongly disagree with the idea of an on-heap overflow map. It adds complexity for rather dubious gains in memory and slows down both lookups and insertions.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
jfim: Why do we need this? What's the rationale behind adding this complexity and what are the gains/tradeoffs?

mcvsubbu: The idea is to keep the implementation flexible at this point, because we don't have enough data right now. Having the expansion multiple be configurable helps us evaluate different expansion techniques in our benchmark.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
mcvsubbu: The map is small compared to the cardinality. We can also bound the size to some small number (say, 1k elements). Keeping the overflow in the map provides a way to absorb small spikes.

jfim: What's the rate at which these small spikes happen?

mcvsubbu: The rate does not matter, right? It is the number of spikes that we are guarding against.  When we reach a threshold, we create a new off-heap buffer.

As coded, the threshold is set at 1%. I am inclined towards setting the threshold to some hard value (say, 5k), so that we can hard-bound the number of values stored in heap.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,411 @@
...
+
+    private List<IntBuffer> getIBufList() {
+      return _iBufList;
+    }
+    private Map<Object, Integer> getOverflowMap() {
+      return _overflowMap;
+    }
+  }
+
+  private AtomicReference<ValueToDictId> _valueToDict = new AtomicReference<>();
```
Jackie-Jiang: volatile ValueToDictId is good enough and with higher performance right?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,411 @@
...
+    iBufs.clear();
+
+    for (PinotDataBuffer pinotDataBuffer : _pinotDataBuffers) {
+      pinotDataBuffer.close();
+    }
+    _pinotDataBuffers.clear();
+  }
+
+  private IntBuffer expand(final int newSize) {
+    // newSize must be a power of 2
```
Jackie-Jiang: Not newSize but newSize / NUM_COLUMNS right?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
mcvsubbu: I have now changed the threshold to be passed in for overflow. A threshold of 0 will disable overflow.
Performance benchmarks show that the number of ops with overflow is slightly lower than the number of ops without. so there is a marginal perf degradation during insertion. (5811 vs 5812, 7018 vs 7174,  7042 vs 7103 for various initial cardinality settings)
Memory benchmark shows that having a small overflow (I benchmarked with 1000) is definitely advantageous in terms of reducing the number of buffers (and total memory) allocated. I have detailed benchmark results for anyone to see, or you can run the perf and mem benchmarks checked in.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
mcvsubbu: Perhaps so. I don't recollect why I set it to atomic ref. let me check on that.

pinot-core/src/test/java/com/linkedin/pinot/core/realtime/impl/dictionary/DictionaryPerfRunner.java
```diff
@@ -113,14 +112,19 @@ public void run() {
...
   public static void main(String[] args)
-      throws InterruptedException {
-    int cardinality = Integer.parseInt(args[0]);
+      throws InterruptedException, IOException {
+    final int cardinality = Integer.parseInt(args[0]);
+    MutableDictionary dictionary;
 
     System.out.println("--------------------------------------------------------------------------------");
-    readWritePerfOnCardinality(cardinality);
+    dictionary = new IntOnHeapMutableDictionary();
```
mcvsubbu: I have now added jmh based performance suite now.

pinot-core/src/test/java/com/linkedin/pinot/core/realtime/impl/dictionary/ConcurrentReadWriteDictionaryTest.java
```diff
@@ -33,40 +35,130 @@
...
   }
 
+  // A test to verify the functionality of off-heap int dictionary without concurrency
   @Test
-  public void testMultiReadersSingleWriter()
-      throws Exception {
-    Future[] readerFutures = new Future[NUM_READERS];
-    for (int i = 0; i < NUM_READERS; i++) {
-      readerFutures[i] = EXECUTOR_SERVICE.submit(new Reader());
+  public void testOffHeapDictionary() throws Exception {
```
mcvsubbu: Done

pinot-core/src/test/java/com/linkedin/pinot/core/realtime/impl/dictionary/ConcurrentReadWriteDictionaryTest.java
```diff
@@ -79,44 +171,60 @@ public void tearDown() {
...
-        do {
-          dictId = _intOnHeapMutableDictionary.indexOf(i + 1);
-        } while (dictId < 0);
-        Assert.assertEquals(dictId, i);
-        Assert.assertEquals(_intOnHeapMutableDictionary.getIntValue(dictId), i + 1);
-
-        // Fetch value by a random existing dictId
-        int randomDictId = RANDOM.nextInt(i + 1);
-        Assert.assertEquals(_intOnHeapMutableDictionary.getIntValue(randomDictId), randomDictId + 1);
+      try {
```
mcvsubbu: So that if we see a throwable, we can reproduce the test case with the right random seed.

pinot-core/src/test/java/com/linkedin/pinot/core/realtime/impl/dictionary/ConcurrentReadWriteDictionaryTest.java
```diff
@@ -79,44 +171,60 @@ public void tearDown() {
...
         throws Exception {
-      for (int i = 0; i < NUM_ENTRIES; i++) {
-        _intOnHeapMutableDictionary.index(i + 1);
-        Assert.assertEquals(_intOnHeapMutableDictionary.indexOf(i + 1), i);
-
-        // Index a random existing value
-        int randomValue = RANDOM.nextInt(i + 1) + 1;
-        _intOnHeapMutableDictionary.index(randomValue);
-        Assert.assertEquals(_intOnHeapMutableDictionary.indexOf(randomValue), randomValue - 1);
+      try {
```
mcvsubbu: ditto

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
mcvsubbu: Benchmark measurements show that it degrades performance marginally, but helps in memory allocation (because of the multiple factor to allocate off-heap memory). One new buffer would be allocated instead of storing a few values in heap.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,476 @@
...
+  private int nearestPrime(int size) {
+    for (int i = 0; i < PRIME_NUMBERS.length; i++) {
+      if (PRIME_NUMBERS[i] >= size) {
+        return PRIME_NUMBERS[i];
+      }
+    }
+    return PRIME_NUMBERS[PRIME_NUMBERS.length-1];
+  }
+
+  private IntBuffer expand(final int newSize) {
```
mcvsubbu: Actually, it is not the number of rows. It is numRows * NUM_COLUMNS.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,476 @@
...
+    PinotDataBuffer buffer = PinotDataBuffer.allocateDirect(bbSize);
+    _pinotDataBuffers.add(buffer);
+    buffer.order(ByteOrder.nativeOrder());
+    IntBuffer iBuf = buffer.toDirectByteBuffer(0L, bbSize).asIntBuffer();
+    for (int i = 0; i < iBuf.capacity(); i++) {
+      iBuf.put(i, NULL_VALUE_INDEX);
+    }
+    newList.add(iBuf);
+    Map<Object, Integer> newOverflowMap = new ConcurrentHashMap<>(_maxItemsInOverflowHash);
+    if (_maxItemsInOverflowHash > 0) {
```
mcvsubbu: else we won't have anything in the overflow hash, so nothing to do, right?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,476 @@
...
+
+  private int validateSize(int lastCapacity, int expansionMultiple) {
+    long newSize = (long) lastCapacity * expansionMultiple;
+    while (newSize >= Integer.MAX_VALUE) {
+      newSize = newSize/2;
+    }
+    return nearestPrime((int) newSize);
+  }
+
+  protected int nearestPowerOf2(int num) {
```
mcvsubbu: Used by sub-classes

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
antumbde: that's why I suggested numEntries or targetNumEntries (you are also using _numEntries to track uniques)

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
antumbde: makes sense

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,476 @@
...
+    // We know that we had a hash collision beyond the number of columns in the buffer.
+    Map<Object, Integer> overflowMap = valueToDictId.getOverflowMap();
+    if (_maxItemsInOverflowHash > 0) {
+      Integer dictId = overflowMap.get(value);
+      if (dictId != null) {
+        return;
+      }
+    }
+
+    setRawValueAt(_numEntries, value);
```
antumbde: fold this and block below in the if _maxItemsInOverflowHash > 0 block

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
mcvsubbu: working on that. let us see how it turns out.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,476 @@
...
+    final int numBuffers = iBufList.size();
+    if (numBuffers == 0) {
+      return expand(_sizeOfFirstBuf);
+    }
+    final int lastCapacity = iBufList.get(numBuffers-1).capacity();
+    int expansionMultiple = EXPANSION_MULTIPLES[EXPANSION_MULTIPLES.length-1];
+    if (numBuffers < EXPANSION_MULTIPLES.length) {
+      expansionMultiple = EXPANSION_MULTIPLES[numBuffers];
+    }
+    int newSize = validateSize(lastCapacity, expansionMultiple);
```
mcvsubbu: working on this

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/BaseOffHeapMutableDictionary.java
mcvsubbu: we need to set the raw value *before* we update the reverse index, and hence it is here.

PR #1538, opened 2017-06-09
Fix brokerHost argument being ignored

jfim:
When starting the Pinot broker with the brokerHost argument, this
argument was previously ignored and was not used to build the broker
id. This patch fixes this issue.


pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -74,18 +74,23 @@
   private static final String ROUTING_TABLE_PARAMS_SUBSET_KEY =
       "pinot.broker.routing.table";
 
-  public HelixBrokerStarter(String helixClusterName, String zkServer, Configuration pinotHelixProperties)
+  public HelixBrokerStarter(String brokerHost, String helixClusterName, String zkServer, Configuration pinotHelixProperties)
```
kishoreg: keep the old constructor and invoke the new constructor with current hostname

PR #1543, opened 2017-06-09
[TE] expanding daterangepicker max time range

justYves:
Per @bl44's request, increasing the max time limit (from 60 days to 6 months). 
This shouldn't be an issue since users can now filter down the results. 

### Test: 
Tested locally.


thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/AnalysisView.js
```diff
@@ -248,7 +248,7 @@ AnalysisView.prototype = {
       endDate: initialEnd,
       maxDate: maxTime,
       dateLimit: {
-        days: 60
+      months: 6
```
cyy0714: one minor: the indent of this line.

theseyi:
LG

PR #1549, opened 2017-06-12
[TE]Unify BaseAlertFilter Thirdeye and Anomaly Library

cecilynie:
There used to be two "BaseAlertFilter" classes in anomaly detection library and thirdeye. Refactoring this while deprecating wrapper classes for alert filter. Deprecate the based alert filter within AD library, let auto tuning to directly use "Performance Evaluation" in thirdeye to make it consistent.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
@@ -8,6 +8,19 @@ import org.slf4j.LoggerFactory;
 
 public abstract class BaseAlertFilter implements AlertFilter {
   private final static Logger LOG = LoggerFactory.getLogger(BaseAlertFilter.class);
+
+  // Check if the given string can be parsed as double value
+  public static boolean isNumeric(String str) {
+    try {
+      double d = Double.parseDouble(str);
+    }
+    catch(NumberFormatException nfe) {
+      return false;
+    }
+    return true;
+  }
+
+
   /**
    * Parses the parameter setting for this filter.
    *
@@ -23,9 +36,17 @@ public abstract class BaseAlertFilter implements AlertFilter {
     Class c = this.getClass();
     for (String fieldName : getPropertyNames()) {
       Double value = null;
+      String strVal = null;
+      boolean isNumeric = true;
       // Get user's value for the specified field
       if (parameterSetting.containsKey(fieldName)) {
-        value = Double.parseDouble(parameterSetting.get(fieldName));
+        String fieldVal = parameterSetting.get(fieldName);
+        if (isNumeric(fieldVal)) {
+          value = Double.parseDouble(parameterSetting.get(fieldName));
+        } else {
+          isNumeric = false;
+          strVal = fieldVal;
+        }
       } else {
         // If user's value does not exist, try to get the default value from Class definition
         try {
@@ -50,7 +71,12 @@ public abstract class BaseAlertFilter implements AlertFilter {
         Field field = c.getDeclaredField(fieldName);
         boolean accessible = field.isAccessible();
         field.setAccessible(true);
-        field.set(this, Double.valueOf(value));
+        if (isNumeric) {
+          field.set(this, Double.valueOf(value));
+        }
+        else {
+          field.set(this, strVal);
+        }
         field.setAccessible(accessible);
       } catch (NoSuchFieldException | IllegalAccessException e) {
         LOG.warn("Failed to set the field {} for class {} exception: {}", fieldName, c.getSimpleName(), e.toString());
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
@@ -2,23 +2,25 @@ package com.linkedin.thirdeye.detector.email.filter;
 
 import java.lang.reflect.Field;
 import java.util.Map;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import sun.swing.StringUIClientPropertyKey;
 
 
 public abstract class BaseAlertFilter implements AlertFilter {
   private final static Logger LOG = LoggerFactory.getLogger(BaseAlertFilter.class);
 
-  // Check if the given string can be parsed as double value
-  public static boolean isNumeric(String str) {
-    try {
-      double d = Double.parseDouble(str);
-    }
-    catch(NumberFormatException nfe) {
-      return false;
-    }
-    return true;
-  }
+//  // Check if the given string can be parsed as double value
+//  public static boolean isNumeric(String str) {
+//    try {
+//      double d = Double.parseDouble(str);
+//    }
+//    catch(NumberFormatException nfe) {
+//      return false;
+//    }
+//    return true;
+//  }
 
 
   /**
@@ -37,14 +39,12 @@ public abstract class BaseAlertFilter implements AlertFilter {
     for (String fieldName : getPropertyNames()) {
       Double value = null;
       String strVal = null;
-      boolean isNumeric = true;
       // Get user's value for the specified field
       if (parameterSetting.containsKey(fieldName)) {
         String fieldVal = parameterSetting.get(fieldName);
-        if (isNumeric(fieldVal)) {
+        if (StringUtils.isNumeric(fieldVal)) {
           value = Double.parseDouble(parameterSetting.get(fieldName));
         } else {
-          isNumeric = false;
           strVal = fieldVal;
         }
       } else {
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
```diff
@@ -50,7 +71,12 @@ public void setParameters(Map<String, String> parameterSetting) {
         Field field = c.getDeclaredField(fieldName);
         boolean accessible = field.isAccessible();
         field.setAccessible(true);
-        field.set(this, Double.valueOf(value));
+        if (isNumeric) {
```
apucher: This should be the other way around. Depend on the declared field type, not the input data - i.e. if the field has a numeric type, attempt conversion.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
```diff
@@ -8,6 +8,19 @@
 
 public abstract class BaseAlertFilter implements AlertFilter {
   private final static Logger LOG = LoggerFactory.getLogger(BaseAlertFilter.class);
+
+  // Check if the given string can be parsed as double value
+  public static boolean isNumeric(String str) {
```
apucher: Please use ApacheCommons instead. "StringUtils.isNumeric"

apucher: Probably "NumberUtils.isParsable" is a good choice too.
https://stackoverflow.com/questions/1102891/how-to-check-if-a-string-is-numeric-in-java

cecilynie: fixed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
cecilynie: Changed to use field type

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
```diff
@@ -2,12 +2,27 @@
...
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import sun.swing.StringUIClientPropertyKey;
 
 
 public abstract class BaseAlertFilter implements AlertFilter {
   private final static Logger LOG = LoggerFactory.getLogger(BaseAlertFilter.class);
+
+//  // Check if the given string can be parsed as double value
```
cyenjung: Remove the dead code.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
```diff
@@ -50,7 +66,13 @@ public void setParameters(Map<String, String> parameterSetting) {
         Field field = c.getDeclaredField(fieldName);
         boolean accessible = field.isAccessible();
         field.setAccessible(true);
-        field.set(this, Double.valueOf(value));
+        Object object = field.get(this);
```
apucher: Please check the declared field type, not the field value (which may be null)

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/detector/email/filter/TestBaseAlertFilter.java
```diff
@@ -0,0 +1,41 @@
...
+    Assert.assertEquals(alphaBetaAlertFilter.getThreshold(), Double.valueOf(alertfilter.get("threshold")));
+
+    // test missing field
+    alertfilter.remove("threshold");
+    AlphaBetaAlertFilter alphaBetaAlertFilter1 = new AlphaBetaAlertFilter();
+    alphaBetaAlertFilter1.setParameters(alertfilter);
+    Assert.assertEquals(alphaBetaAlertFilter1.getThreshold(), Double.valueOf(AlphaBetaAlertFilter.DEFAULT_THRESHOLD));
+  }
+
+}
```
apucher: Thanks for adding a unit test

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
```diff
@@ -23,23 +28,34 @@ public void setParameters(Map<String, String> parameterSetting) {
...
+        if (NumberUtils.isNumber(fieldVal)) {
+          value = Double.parseDouble(parameterSetting.get(fieldName));
+        } else {
+          strVal = fieldVal;
+        }
       } else {
         // If user's value does not exist, try to get the default value from Class definition
         try {
-          Field field = c.getDeclaredField("DEFAULT_" + fieldName.toUpperCase());
+          Field field = c.getDeclaredField(fieldName);
```
cyenjung: This change might break existing classes. This block is trying to get default value for variables when their value is not given by user. The naming convention DEFAULT_VARIABLE_NAME. If you remove the original line, then this block does the same thing at the block at lines 33-39.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
```diff
@@ -23,23 +28,34 @@ public void setParameters(Map<String, String> parameterSetting) {
...
           boolean accessible = field.isAccessible();
           field.setAccessible(true);
-          value = Double.parseDouble((String) field.get(this));
+          Object object = field.get(this);
```
cyenjung: Why the block at lines 35-39 and the blocks at lines 46-51 are different? It would be easier to maintain if we apply the same logic to populate the variables.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
```diff
@@ -50,7 +66,13 @@ public void setParameters(Map<String, String> parameterSetting) {
         Field field = c.getDeclaredField(fieldName);
         boolean accessible = field.isAccessible();
         field.setAccessible(true);
-        field.set(this, Double.valueOf(value));
+        Object object = field.get(this);
+        if (object instanceof Double) {
+          field.set(this, value);
+        }
```
cyenjung: Format lines 72 and 73.

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/detector/email/filter/TestBaseAlertFilter.java
```diff
@@ -0,0 +1,41 @@
...
+    Assert.assertEquals(alphaBetaAlertFilter.getType(), alertfilter.get("type"));
+    Assert.assertEquals(alphaBetaAlertFilter.getThreshold(), Double.valueOf(alertfilter.get("threshold")));
+
+    // test scientific decimal
+    double threshold = 1E-10;
+    alertfilter.put("threshold", String.valueOf(threshold));
+    alphaBetaAlertFilter.setParameters(alertfilter);
+    Assert.assertEquals(alphaBetaAlertFilter.getThreshold(), Double.valueOf(alertfilter.get("threshold")));
+
+    // test missing field
```
cyenjung: I am surprised that the default values could be read after you remove the parsing for DEFAULT_VARIABLES in BaseAlertFilter.java. Could you explain how it works after the change?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
cecilynie: All of the "DEFAULT_" variables are defined as String, thus cannot judge the field type to assign value. But within alert filter class, there are private variable with the variable name in their correct type

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/detector/email/filter/TestBaseAlertFilter.java
cecilynie: Since within alertfilter class, for example AlphaBetaAlertFilter, the private variables are initially being assigned as DEFAULT value, thus here it will be equal to default value when missing

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
cecilynie: This is to parse the input, but the next block is to grab default value from the original class, thus they are different

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
cecilynie: fixed.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/BaseAlertFilter.java
```diff
@@ -50,7 +64,12 @@ public void setParameters(Map<String, String> parameterSetting) {
         Field field = c.getDeclaredField(fieldName);
         boolean accessible = field.isAccessible();
         field.setAccessible(true);
-        field.set(this, Double.valueOf(value));
+        if (field.getType().equals(Double.class) || field.getType().equals(double.class)) {
+          field.set(this, value);
+        }
+        else {
```
apucher: Maybe check whether it's String-compatible here - and throw an Exception if not.

cecilynie: Add to throw Exception when field is none of these "Double.class", "double.class" or "String.class"

PR #1558, opened 2017-06-14
Adding partition aware routing table builder

kishoreg:
Partition aware routing table builder prunes the segment based on zkmetadata and picks the servers based on the partition assignment info in property store. This is done only for the offline tables for now. Will extend it to realtime in the another PR.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
@@ -560,7 +560,7 @@ public class BrokerRequestHandler {
     phaseTimes.addToRoutingTime(System.nanoTime() - routingStartTime);
     if (segmentServices == null || segmentServices.isEmpty()) {
       String tableNameWithType = brokerRequest.getQuerySource().getTableName();
-      LOGGER.info("No server found for table: {}", tableNameWithType);
+      LOGGER.info("No server found or all segments are pruned for table: {}", tableNameWithType);
       _brokerMetrics.addMeteredTableValue(tableNameWithType, BrokerMeter.NO_SERVER_FOUND_EXCEPTIONS, 1L);
       return null;
     }
@@ -594,7 +594,8 @@ public class BrokerRequestHandler {
       routingOptions =
           Splitter.on(",").omitEmptyStrings().trimResults().splitToList(debugOptions.get("routingOptions"));
     }
-    RoutingTableLookupRequest routingTableLookupRequest = new RoutingTableLookupRequest(tableName, routingOptions, brokerRequest);
+    RoutingTableLookupRequest routingTableLookupRequest = new RoutingTableLookupRequest(tableName, routingOptions,
+        brokerRequest);
     return _routingTable.findServers(routingTableLookupRequest);
   }
 
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
@@ -594,7 +594,7 @@ public class BrokerRequestHandler {
       routingOptions =
           Splitter.on(",").omitEmptyStrings().trimResults().splitToList(debugOptions.get("routingOptions"));
     }
-    RoutingTableLookupRequest routingTableLookupRequest = new RoutingTableLookupRequest(tableName, routingOptions);
+    RoutingTableLookupRequest routingTableLookupRequest = new RoutingTableLookupRequest(tableName, routingOptions, brokerRequest);
     return _routingTable.findServers(routingTableLookupRequest);
   }
 
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
@@ -18,6 +18,8 @@ package com.linkedin.pinot.broker.routing;
 import java.util.HashSet;
 import java.util.Set;
 import org.apache.commons.configuration.Configuration;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -36,12 +38,15 @@ public class RoutingTableBuilderFactory {
   private static final Set<Class<? extends RoutingTableBuilder>> _routingTableBuilders = new HashSet<>();
   private Configuration _configuration;
 
+  private ZkHelixPropertyStore<ZNRecord> _propertyStore;
+
   enum RoutingTableBuilderName {
     DefaultOffline, DefaultRealtime, BalancedRandom, KafkaLowLevel, KafkaHighLevel
   }
 
-  public RoutingTableBuilderFactory(Configuration configuration) {
+  public RoutingTableBuilderFactory(Configuration configuration, ZkHelixPropertyStore<ZNRecord> propertyStore) {
     _configuration = configuration;
+    _propertyStore = propertyStore;
   }
 
   public RoutingTableBuilder createRoutingTableBuilder(TableConfig tableConfig) {
@@ -87,7 +92,7 @@ public class RoutingTableBuilderFactory {
       builder = new KafkaLowLevelConsumerRoutingTableBuilder();
       break;
     }
-    builder.init(_configuration);
+    builder.init(_configuration, tableConfig, _propertyStore);
     return builder;
   }
 
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
@@ -15,6 +15,9 @@
  */
 package com.linkedin.pinot.broker.routing;
 
+import com.linkedin.pinot.broker.routing.builder.PartitionAwareOfflineRoutingTableBuilder;
+import com.linkedin.pinot.common.config.SegmentsValidationAndRetentionConfig;
+import com.linkedin.pinot.common.utils.CommonConstants;
 import java.util.HashSet;
 import java.util.Set;
 import org.apache.commons.configuration.Configuration;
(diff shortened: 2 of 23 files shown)


snleee:
Note: this PR is not ready for check-in. We are currently testing the code and running experiments. Feel free to take a look or review this though.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+      Map<String, ColumnPartitionMetadata> columnPartitionMap = partitionMetadata.getColumnPartitionMap();
+      ColumnPartitionMetadata columnPartitionMetadata;
+      if (columnPartitionMap.size() == 1) {
+        columnPartitionMetadata = columnPartitionMap.values().iterator().next();
+        int partitionIdStart = columnPartitionMetadata.getPartitionRanges().get(0).getMaximumInteger();
+        // int partitionIdEnd = columnPartitionMetadata.getPartitionRanges().get(0).getMaximumInteger();
+        return partitionIdStart;
+      }
+    }
+    return 0;
```
mayankshriv: May be use a static final so we know what 0 means?

pinot-common/src/main/java/com/linkedin/pinot/common/response/ServerInstance.java
```diff
@@ -204,4 +206,16 @@ public static ServerInstance forHostPort(String name, int port) {
       return newInstance;
     }
   }
+
+  public static ServerInstance forInstanceName(String instanceName) {
```
mayankshriv: Check if the optimization made in https://github.com/linkedin/pinot/pull/1487 is required here as well.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+import com.linkedin.pinot.common.metadata.ZKMetadataProvider;
+import com.linkedin.pinot.common.metadata.segment.ColumnPartitionMetadata;
+import com.linkedin.pinot.common.metadata.segment.PartitionToReplicaGroupMappingZKMetadata;
+import com.linkedin.pinot.common.metadata.segment.SegmentPartitionMetadata;
+import com.linkedin.pinot.common.metadata.segment.SegmentZKMetadata;
+import com.linkedin.pinot.common.response.ServerInstance;
+import com.linkedin.pinot.transport.common.SegmentId;
+import com.linkedin.pinot.transport.common.SegmentIdSet;
+
+public class PartitionAwareOfflineRoutingTableBuilder extends AbstractRoutingTableBuilder {
```
mayankshriv: We should add unit test for this, if not already added.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+  private int _numReplicas;
+  private boolean _isPartitionLevelReplicaGroupAssignment;
+
+  @Override
+  public void init(Configuration configuration, TableConfig tableConfig, ZkHelixPropertyStore<ZNRecord> propertyStore) {
+    _tableConfig = tableConfig;
+    _propertyStore = propertyStore;
+    _pruner = new SegmentZKMetadataPrunerService(new String[]{PARTITION_METADATA_PRUNER});
+
+    final String partitionColumn = _tableConfig.getValidationConfig().getReplicaGroupStrategyConfig().getPartitionColumn();
```
snleee: I think we should add the check in the routing table builder factory. If `ReplicaGroupStrategyConfig` does not exist, we should use the default routing builder.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+    _isPartitionLevelReplicaGroupAssignment = (partitionColumn == null) ? true : false;
+  }
+
+  @Override
+  public void computeRoutingTableFromExternalView(String tableName, ExternalView externalView, List<InstanceConfig> instanceConfigList) {
```
mayankshriv: This does not work for refresh use cases, because when segments are refreshed external view is not updated. We need to put a check somewhere so that this feature is not enabled for refresh usecases.

kishoreg: we can add the check in the factory that sets up the right routingtable for each table based on tableconfig.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
```diff
@@ -71,23 +78,33 @@ public RoutingTableBuilder createRoutingTableBuilder(TableConfig tableConfig) {
...
+        builder = new KafkaHighLevelConsumerBasedRoutingTableBuilder();
+        break;
+      case KafkaLowLevel:
+        builder = new KafkaLowLevelConsumerRoutingTableBuilder();
+        break;
+      case PartitionAwareOffline:
+        // We can use partition aware routing only if we use replica group aware segment assignment strategy.
+        String segmentAssignmentStrategy = tableConfig.getValidationConfig().getSegmentAssignmentStrategy();
+        if (CommonConstants.Helix.DataSource.SegmentAssignmentStrategyType.valueOf(segmentAssignmentStrategy)
+            != CommonConstants.Helix.DataSource.SegmentAssignmentStrategyType.ReplicaGroupSegmentAssignmentStrategy) {
```
kishoreg: its logged in markDataResourceOnline method.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+    Set<Integer> partitionIds = new HashSet<>();
+    for (String segment : segmentSet) {
+      SegmentId segmentId = new SegmentId(segment);
+      // retrieve the metadata for the segment and compute the partitionIds set
+      SegmentZKMetadata segmentZKMetadata = _segment2SegmentZkMetadataMap.get(segmentId);
+      if (segmentZKMetadata == null) {
+        segmentZKMetadata = ZKMetadataProvider.getOfflineSegmentZKMetadata(_propertyStore, tableName, segment);
+        _segment2SegmentZkMetadataMap.put(segmentId, segmentZKMetadata);
+      }
+      int partitionId = getPartitionId(segmentZKMetadata);
```
mayankshriv: What happens if segment zk metadata does not have partition info for a subset of segments? Should we drop those segments, or be smart enough to still pick the right servers for them?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
snleee: done

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -537,6 +537,8 @@ private BrokerResponse processOptimizedBrokerRequests(@Nonnull BrokerRequest ori
     long routingStartTime = System.nanoTime();
     Map<ServerInstance, SegmentIdSet> segmentServices = findCandidateServers(brokerRequest);
     phaseTimes.addToRoutingTime(System.nanoTime() - routingStartTime);
+
+    // TODO: need to differentiate the case of no server found because all segments are pruned.
```
snleee: After the discussion with @Jackie-Jiang , we can keep the current behavior for the case of no server found vs no server found due to pruning. I initially thought that `no server found` means the failure case where no server is alive for a segment (So, it makes sense to throw an exception). However, this covers more generic cases (i.e. empty table). In case of all servers down for a segment, all of broker builder implementations would throw an exception so it won't go through this code path.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  @Override
+  public void computeRoutingTableFromExternalView(String tableName, ExternalView externalView, List<InstanceConfig> instanceConfigList) {
+    RoutingTableInstancePruner pruner = new RoutingTableInstancePruner(instanceConfigList);
+    String[] segmentSet = externalView.getPartitionSet().toArray(new String[externalView.getPartitionSet().size()]);
+    _numReplicas = _tableConfig.getValidationConfig().getReplicationNumber();
+    _partitionToReplicaGroupMappingZKMedata = ZKMetadataProvider.getPartitionToReplicaGroupMappingZKMedata(_propertyStore, tableName);
+
+    // Compute partition id set
```
snleee: I will address this in separate pr for partition aware routing for realtime.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
snleee: I will address this in separate pr for partition aware routing for realtime.

snleee: done

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
snleee: I will handle this with separate pr with realtime

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
snleee: If there is no partition information, I sequentially assign replica id for each server and add it to look up table.

PR #1571, opened 2017-06-15
Reduce the table level broker metrics emitted

Jackie-Jiang:
1. Change RESOURCE_MISSING_EXCEPTIONS to a global meter so that random table name does not create a new metric
2. First check if any resource gets hit, so that random table name does not create a new metric
3. If multiple tables might be hit (hybrid case), use raw table name for the metrics
4. If only one table can be hit, use table name with type for the metrics

By doing this, metrics for a single table/resource will not split into multiple set of metrics
e.g. For offline use case, metircs for tableName and tableName_OFFLINE should be the same set of metrics


Jackie-Jiang:
@mayankshriv Currently went through some local manual tests. All queries work as expected, will check the metrics when deployed.

PR #1584, opened 2017-06-19
Adding feedback status and comments to anomaly details

npawar:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/AnomaliesResource.java
```diff
@@ -924,8 +924,13 @@ private AnomalyDetails constructAnomalyDetails(String metricName, String dataset
...
     }
-    if (mergedAnomaly.getFeedback() != null) {
-      anomalyDetails.setAnomalyFeedback(AnomalyDetails.getFeedbackStringFromFeedbackType(mergedAnomaly.getFeedback().getFeedbackType()));
+    AnomalyFeedback mergedAnomalyFeedback = mergedAnomaly.getFeedback();
+    if (mergedAnomalyFeedback != null) {
+      anomalyDetails.setAnomalyFeedback(AnomalyDetails.getFeedbackStringFromFeedbackType(mergedAnomalyFeedback.getFeedbackType()));
+      if (mergedAnomalyFeedback.getStatus() != null) {
+        anomalyDetails.setAnomalyFeedbackStatus(mergedAnomalyFeedback.getStatus().toString());
+      }
+      anomalyDetails.setAnomalyFeedbackComments(mergedAnomalyFeedback.getComment());
```
cyenjung: Should we check if the comment string is null?

npawar: it won't matter right? it wont cause any NPE, and the UI will just read null in that case

PR #1588, opened 2017-06-20
[TE] Rootcause events endpoint

apucher:
* unify rootcause search to into "query" endpoint

    now takes framework param, as defined in rca.yml

* generalize code for event ranking in RCA pipelines

* update frontend ajax URLs


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsPipeline.java
```diff
@@ -73,33 +73,30 @@ public HolidayEventsPipeline(String outputName, Set<String> inputNames, Map<Stri
   @Override
   public PipelineResult run(PipelineContext context) {
     TimeRangeEntity current = TimeRangeEntity.getContextCurrent(context);
-    TimeRangeEntity baseline = TimeRangeEntity.getContextBaseline(context);
+    //TimeRangeEntity baseline = TimeRangeEntity.getContextBaseline(context);
```
cyenjung: Remove the dead code?

PR #1591, opened 2017-06-20
[PINOT-4889] Implemented off-heap dictionary work for strings

mcvsubbu:
Off-heap dictionary for strings has been implemented using an MutableOffHeapByteStore class,
that supports three basic operations: add(byte[]), compareBytesAtIndex(byte[], index), and get(index)

The StringOffHeapDictionary class makes use of this store and the reverse mapping already implemented
in BaseOffHeapDictionary class.

Added tests.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
@@ -80,9 +80,13 @@ import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
  */
 public class MutableOffHeapByteArrayStore implements Closeable {
   private static final int START_SIZE = 32 * 1024;
+  private static final int INT_SIZE = V1Constants.Numbers.INTEGER_SIZE;
+
+  public static int getStartSize() {
+    return START_SIZE;
+  }
 
   private static class Buffer implements Closeable {
-    private static final int INT_SIZE = V1Constants.Numbers.INTEGER_SIZE;
 
     private final PinotDataBuffer _pinotDataBuffer;
     private final ByteBuffer _byteBuffer;
@@ -106,7 +110,7 @@ public class MutableOffHeapByteArrayStore implements Closeable {
 
     private int add(byte[] value) {
       int startOffset = _availEndOffset - value.length;
-      if (startOffset <= (_numValues + 1) * INT_SIZE) {
+      if (startOffset < (_numValues + 1) * INT_SIZE) {
         // full
         return -1;
       }
@@ -168,12 +172,19 @@ public class MutableOffHeapByteArrayStore implements Closeable {
   private volatile Buffer _currentBuffer;
 
   public MutableOffHeapByteArrayStore() {
-    expand(START_SIZE);
+    expand(START_SIZE, 0L);
   }
 
-  // Expand the buffer size, allocating a min of 32k
-  private Buffer expand(long size) {
-    Buffer buffer = new Buffer(size, _numElements);
+  /**
+   * Expand the buffer list to add a new buffer, allocating a buffer that can definitely fit
+   * the new value.
+   *
+   * @param suggestedSize is the size of the new buffer to be allocated
+   * @param minSize is the new value that must fit into the new buffer.
+   * @return
+   */
+  private Buffer expand(long suggestedSize, long minSize) {
+    Buffer buffer = new Buffer(Math.max(suggestedSize, minSize), _numElements);
     List<Buffer> newList = new LinkedList<>();
     for (Buffer b : _buffers) {
       newList.add(b);
@@ -184,8 +195,8 @@ public class MutableOffHeapByteArrayStore implements Closeable {
     return buffer;
   }
 
-  private Buffer expand() {
-    Buffer newBuffer = expand(_currentBuffer.getSize() * 2);
+  private Buffer expand(long sizeOfNewValue) {
+    Buffer newBuffer = expand(_currentBuffer.getSize() * 2, sizeOfNewValue + INT_SIZE);
     return newBuffer;
   }
 
@@ -207,7 +218,7 @@ public class MutableOffHeapByteArrayStore implements Closeable {
     Buffer buffer = _currentBuffer;
     int index = buffer.add(value);
     while (index < 0) {
-      buffer = expand();
+      buffer = expand(value.length);
       index = buffer.add(value);
     }
     _numElements++;
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
@@ -1,0 +1,238 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.linkedin.pinot.core.io.writer.impl;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import com.linkedin.pinot.core.segment.creator.impl.V1Constants;
+import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
+
+
+/**
+ * @class OffHeapMutableByteArrayStore
+ *
+ * An off-heap byte array store that provides APIs to add byte array (value), retrieve a value, and compare value at
+ * an index. No verification is made as to whether the value added already exists or not.
+ * Empty byte arrays are supported.
+ *
+ * @note The class is thread-safe for single writer and multiple readers.
+ *
+ * This class has a list of OffHeapMutableByteArrayStore.Buffer objects. As Buffer objects get filled, new Buffer objects
+ * are added to the list. New Buffers objects have twice the capacity of the previous Buffer
+ *
+ * Within a Buffer object byte arrays (values) are stored as below:
+ *
+ *                  __________________________________
+ *                  |  start offset of array  1      |
+ *                  |  start offset of array  2      |
+ *                  |        .....                   |
+ *                  |  start offset of array  N      |
+ *                  |                                |
+ *                  |         UNUSED                 |
+ *                  |                                |
+ *                  |  Array N .....                 |
+ *                  |          .....                 |
+ *                  |          .....                 |
+ *                  |  Array N-1                     |
+ *                  |          .....                 |
+ *                  |          .....                 |
+ *                  |  Array 0 .....                 |
+ *                  |          .....                 |
+ *                  |          .....                 |
+ *                  |________________________________|
+ *
+ *
+ * We fill the buffer as follows:
+ * - The values are added from the bottom, each new value appearing nearer to the top of the buffer, leaving no
+ *   room between them. Each value is stored as a sequence of bytes.
+ *
+ * - The start offsets of the byte arrays are added from the top. Each start offset is stored as an integer, taking 4 bytes.
+ *
+ * Each time we want to add a new value, we check if we have space to add the length of the value, and the value
+ * itself. If we do, then we compute the start offset of the new value as:
+ *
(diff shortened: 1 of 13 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/StringOffHeapMutableDictionary.java
```diff
@@ -0,0 +1,146 @@
...
+    if (_stringStore.equalsStringAt(value.toString(), dictId)) {
+      return true;
+    }
+    return false;
+  }
+
+  private void updateMinMax(String value) {
+    if (_min == null) {
+      _min = value;
+      _max = value;
```
jenniferdai: do we want to check if max is null before assigning it a value?

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,235 @@
...
+    final ByteBuffer _byteBuffer;
+    final int _startIndex;
+    final long _size;
+
+    int _numStrings = 0;
+    int _availEndOffset;  // Exclusive
+
+    private Buffer(long size, int startIndex) {
+      if (size > Integer.MAX_VALUE) {
+        size = Integer.MAX_VALUE;
```
jenniferdai: Does this conflict with the preconditions check you put in allocatedirect?
 Preconditions.checkArgument(size >= 0 && size < Integer.MAX_VALUE, "bad value for size " + size);

pinot-core/src/main/java/com/linkedin/pinot/core/segment/memory/PinotByteBuffer.java
```diff
@@ -117,7 +117,7 @@ static PinotDataBuffer loadFromFile(File file, long startPosition, long length,
   }
 
   public static PinotByteBuffer allocateDirect(long size, String context) {
-    Preconditions.checkArgument(size >= 0 && size < Integer.MAX_VALUE);
+    Preconditions.checkArgument(size >= 0 && size < Integer.MAX_VALUE, "bad value for size " + size);
```
jenniferdai: do we want to allow size to be 0?

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,235 @@
...
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.ArrayList;
+import java.util.List;
+import com.linkedin.pinot.core.segment.creator.impl.V1Constants;
+import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
+
+
+/**
+ * @class OffHeaStringStore
```
mcvsubbu: fixed

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
mcvsubbu: Good catch, it does.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,235 @@
...
+    Buffer lastBuffer = _buffers.get(_buffers.size()-1);
+    Buffer newBuffer = expand(lastBuffer.getSize());
+    return newBuffer;
+  }
+
+  // Returns a string, given an index
+  public String get(int index) {
+    for (int x = _buffers.size()-1; x >= 0; x--) {
+      Buffer buffer = _buffers.get(x);
+      // Assumed that we will never ask for a
```
mcvsubbu: fixed

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/dictionary/StringOffHeapMutableDictionary.java
mcvsubbu: no, checking one of them should be enough.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/memory/PinotByteBuffer.java
mcvsubbu: I always have a min of 32k. I have now made this a private static.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,235 @@
...
+      _size = size;
+    }
+
+    private int add(String string) {
+      int startOffset = _availEndOffset - string.length() * CHAR_SIZE;
+      if (startOffset <= (_numStrings + 1) * INT_SIZE) {
+        // full
+        return -1;
+      }
+      for (int i = 0, j = startOffset; i < string.length(); i++, j = j + CHAR_SIZE) {
```
jfim: I disagree. While using bytes would make it more memory efficient, on comparison you'll need to interpret the bytes as UTF-8 or (some other encoding) on reads, which entails some conditional branching (it's a 4-way branch) and some bit shifting for each character.

It's been a while since I did any Unicode stuff, but IIRC interpreting UTF-8 entails counting the number of leading ones, then branching based on that. If the code point decoded is outside of the BMP, then you'll need to interpret it as two chars (ie. surrogate pair). That's a fair bit of complexity.

The alternative is to convert the string that's being compared into UTF-8 on every comparison, but that would end up generating some garbage on every lookup.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+    newList.add(buffer);
+    _buffers = newList;
+    _currentBuffer = buffer;
+    return buffer;
+  }
+
+  private Buffer expand() {
+    Buffer lastBuffer = _buffers.get(_buffers.size()-1);
+    Buffer newBuffer = expand(lastBuffer.getSize());
```
jfim: Expand buffers with a multiplicative factor, as to avoid degradation due to large numbers of buffers. This also prevents catastrophically allocating all of the memory if someone tries to add a String that's larger than your buffer size.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
mcvsubbu: I did a second implementation with byte (i.e. this class became a OffHeapByteStore instead of a string store). I re-factored the base class so that I convert incoming string into UTF-8 before doing any operation. Initial results were not nice, since running the unit tests took forever. I suspected gc, but could not follow up on things yesterday. I decided to run a benchmark, kicked one off last night, but it did not complete for some reason. Perhaps because the buffers were not being doubled each invocation. I iwll try to run the benchmark today to get some answer on this.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
mcvsubbu: Yes, that was the intention, and was implemented so before. Somehow, this was missed. I will add it back in.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
mcvsubbu: I ran an insertion benchmark overnight between these two implementations (insertion includes lookup, so it is a reasonable representative). The benchmark was to insert 2.5M string values (1M cardinality). String lengths were under 32 bytes (randomized)

charstore:
  1234 times, 1967 ms/op (an op is doing the work defined as above)
  gc alloc rate was 10.5MB/s
  gc count was 943
  gc time was 5k ms (0.2%)

ByteStore:
  1125 times 2152 ms/op
  gc alloc rate: 10 MB/s
  gc count: 10175
  gc time: 18k ms (0.7%)

So, about 3 times as much time was spent in GC, but the relative time spent in GC seems to be small. Given that we are generally memory bound, a 50% savings in memory seems to be useful (similar trade-off that we made in the offline segments)

I will go with the byte store for now. I will also check-in the (unused) stringstore. If we have GC issues, we can move things around a bit.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+    // Assumed that we will never ask for an index that does not exist.
+    throw new RuntimeException("dictionary ID '" + index + "' too low");
+  }
+
+  // Adds a string and returns the index. No verification is made as to whether the string already exists or not
+  public int add(String string) {
+    Buffer buffer = _currentBuffer;
+    int index = buffer.add(string);
+    while (index < 0) {
```
antumbde: we don't need while here. it should be if statement

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,241 @@
...
+  private volatile Buffer _currentBuffer;
+
+  public OffHeapStringStore() {
+    expand(START_SIZE);
+  }
+
+  // Expand the buffer size, allocating a min of 32k for strings.
+  private Buffer expand(long size) {
+    Buffer buffer = new Buffer(size, _numElements);
+    List<Buffer> newList = new LinkedList<>();
```
antumbde: I still don't get copying of list. We will need atomicreference for concurrency issue. Also, it's assigned LinkedList here and default initialized as ArrayList.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+    _numElements++;
+    return index + buffer.getStartIndex();
+  }
+
+  public boolean equalsStringAt(String string, int index) {
+    List<Buffer> bufList = _buffers;
+    for (int x = bufList.size()-1; x >= 0; x--) {
+      Buffer buffer = bufList.get(x);
+      // Assumed that we will never ask for a
```
antumbde: fix comment

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
```diff
@@ -0,0 +1,241 @@
...
+      if (index >= buffer.getStartIndex()) {
```
antumbde: what if index > largest index in buffer. Buffer class will throw error

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
mcvsubbu: It is a volatile instead of atomic reference. The idea is that a reader can iterate through one list while the writer is in the process of adding an element to the list. Lists cannot be concurrent, otherwise a concurrent list would do.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
mcvsubbu: We double the size everytime. If an odd-ball case comes up that we have to double multiple times, then we need a while.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
mcvsubbu: Not sure how it came back. @jenniferdai also pointed this out., and I thought i had fixed it. Sorry.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/OffHeapStringStore.java
mcvsubbu: We assume that this never happens. The general approach has been (in Pinot) to not add defensive programming in lower layers for performance reasons. You can see that pattern in many places in pinot.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+    // Assumed that we will never ask for an index that does not exist.
+    throw new RuntimeException("dictionary ID '" + index + "' too low");
+  }
+
+  // Adds a byte array and returns the index. No verification is made as to whether the byte array already exists or not
+  public int add(byte[] value) {
+    Buffer buffer = _currentBuffer;
+    int index = buffer.add(value);
+    while (index < 0) {
```
antumbde: As discussed, pass min size to expand().

PR #1595, opened 2017-06-21
[TE] move anomalies view by function id to AnomalyResource and add option to apply alert filter when query anomalies

bl44:
move anomalies view by function id to AnomalyResource and add option to apply alert filter when query anomalies

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -56,14 +56,7 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import javax.validation.constraints.NotNull;
-import javax.ws.rs.DELETE;
-import javax.ws.rs.DefaultValue;
-import javax.ws.rs.GET;
-import javax.ws.rs.POST;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
-import javax.ws.rs.QueryParam;
+import javax.ws.rs.*;
 import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -553,6 +546,72 @@ private String getDimensions(String dataset, String exploreDimensions) throws Ex
...
+   * @param functionId id of the anomaly function
+   * @param startTimeIso The start time of the monitoring window
+   * @param endTimeIso The start time of the monitoring window
+   * @param anomalyType type of anomaly: "raw" or "merged"
+   * @param applyAlertFiler can choose apply alert filter when query merged anomaly results
+   * @return list of anomalyIds (Long)
+   */
+  @GET
+  @Path("anomaly-function/{id}/anomalies")
+  public List<Long> getAnomaliesByFunctionId(@PathParam("id") Long functionId,
```
cyenjung: 1. This resource file is written for version 1 dashboard. The latest resource file is AnomaliesResource.java. Could you move this method there? There is a block in which the methods are named as getAnomaliesBySomething.

2. Rename to getAnomalyIdsByFunctionId.

bl44: I put it here primarily because this class host all endpoints that interact with anomaly-funciton.
This particular one does not fit where in AnomaliesResources, where those endpoints are primarily used by anomalies tab in UI -- the return is a list of metadata of anomaly results.
I do not think a customer facing tab need an endpoints to query anomaly by function Id (we already have the one by function name, this is more intuitive and friendly).

Here I used it primarily for the (onboarding) replay process, we only care about the ids (we also can query raw vs merged right now).

So I think this endpoints belongs here.

PR #1605, opened 2017-06-23
Adding partition aware realtime routing table builder

snleee:
- Added the partition aware realtime routing table builder that can prune segments.
- Refactored the offline routing table builder to share the duplicate code with the
  realtime version.
- Added unit tests for both offline and realtime partition aware routing table builders.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
@@ -16,8 +16,12 @@
 package com.linkedin.pinot.broker.routing;
 
 import com.linkedin.pinot.broker.routing.builder.PartitionAwareOfflineRoutingTableBuilder;
+import com.linkedin.pinot.broker.routing.builder.PartitionAwareRealtimeRoutingTableBuilder;
 import com.linkedin.pinot.common.config.SegmentsValidationAndRetentionConfig;
+import com.linkedin.pinot.common.metadata.stream.KafkaStreamMetadata;
 import com.linkedin.pinot.common.utils.CommonConstants;
+import com.linkedin.pinot.core.realtime.StreamProviderConfig;
+import com.linkedin.pinot.core.realtime.impl.kafka.KafkaLowLevelStreamProviderConfig;
 import java.util.HashSet;
 import java.util.Set;
 import org.apache.commons.configuration.Configuration;
@@ -45,7 +49,13 @@ public class RoutingTableBuilderFactory {
   private ZkHelixPropertyStore<ZNRecord> _propertyStore;
 
   enum RoutingTableBuilderName {
-    DefaultOffline, DefaultRealtime, BalancedRandom, KafkaLowLevel, KafkaHighLevel, PartitionAwareOffline
+    DefaultOffline,
+    DefaultRealtime,
+    BalancedRandom,
+    KafkaLowLevel,
+    KafkaHighLevel,
+    PartitionAwareOffline,
+    PartitionAwareRealtime
   }
 
   public RoutingTableBuilderFactory(Configuration configuration, ZkHelixPropertyStore<ZNRecord> propertyStore) {
@@ -97,20 +107,36 @@ public class RoutingTableBuilderFactory {
         builder = new KafkaLowLevelConsumerRoutingTableBuilder();
         break;
       case PartitionAwareOffline:
-        // In order to use the partition aware routing for offline table, we assume the following conditions.
-        // 1. The replica group aware segment assignment strategy is used.
-        // 2. Table config contains the replica group segment assignment related configurations.
-        // 3. The table push type is not 'refresh'.
-        String segmentAssignmentStrategy = tableConfig.getValidationConfig().getSegmentAssignmentStrategy();
         SegmentsValidationAndRetentionConfig validationConfig = tableConfig.getValidationConfig();
-        if (CommonConstants.Helix.DataSource.SegmentAssignmentStrategyType.valueOf(segmentAssignmentStrategy)
-            != CommonConstants.Helix.DataSource.SegmentAssignmentStrategyType.ReplicaGroupSegmentAssignmentStrategy ||
-            validationConfig.getReplicaGroupStrategyConfig() == null ||
-            (validationConfig.getSegmentPushType() != null && validationConfig.getSegmentPushType()
-                .equalsIgnoreCase("REFRESH"))) {
-          builder = new DefaultOfflineRoutingTableBuilder();
-        } else {
+        String segmentAssignmentStrategy = validationConfig.getSegmentAssignmentStrategy();
+
+        // Check that the replica group aware segment assignment strategy is used.
+        boolean isSegmentAssignmentStrategyCorrect =
+            (CommonConstants.Helix.DataSource.SegmentAssignmentStrategyType.valueOf(segmentAssignmentStrategy)
+                == CommonConstants.Helix.DataSource.SegmentAssignmentStrategyType.ReplicaGroupSegmentAssignmentStrategy);
+
+        // Check that replica group strategy config is correctly set
+        boolean hasReplicaGroupStrategyConfig = (validationConfig != null);
+
+        // Check that the table push type is not 'refresh'.
+        boolean isNotRefreshPush = (validationConfig.getSegmentPushType() != null) &&
+            !validationConfig.getSegmentPushType().equalsIgnoreCase("REFRESH");
+
+        if (isSegmentAssignmentStrategyCorrect && hasReplicaGroupStrategyConfig && isNotRefreshPush) {
           builder = new PartitionAwareOfflineRoutingTableBuilder();
+        } else {
+          builder = new DefaultOfflineRoutingTableBuilder();
+        }
+        break;
+      case PartitionAwareRealtime:
+        // Check that the table uses LLC kafka consumer.
+        KafkaStreamMetadata streamMetadata = new KafkaStreamMetadata(tableConfig.getIndexingConfig().getStreamConfigs());
+
+        if (streamMetadata.getConsumerTypes().size() == 1 && streamMetadata.getConsumerTypes().get(0)
+            == CommonConstants.Helix.DataSource.Realtime.Kafka.ConsumerType.simple) {
+          builder = new PartitionAwareRealtimeRoutingTableBuilder();
+        } else {
+          builder = new DefaultRealtimeRoutingTableBuilder();
         }
         break;
     }
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
@@ -1,0 +1,126 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.broker.routing.builder;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.commons.configuration.Configuration;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.linkedin.pinot.broker.pruner.SegmentPrunerContext;
+import com.linkedin.pinot.broker.pruner.SegmentZKMetadataPrunerService;
+import com.linkedin.pinot.broker.routing.RoutingTableLookupRequest;
+import com.linkedin.pinot.common.config.TableConfig;
+import com.linkedin.pinot.common.metadata.segment.SegmentZKMetadata;
+import com.linkedin.pinot.common.response.ServerInstance;
+import com.linkedin.pinot.transport.common.SegmentId;
+import com.linkedin.pinot.transport.common.SegmentIdSet;
+
+
+/**
+ * Abstract partition aware routing table builder.
+ *
+ * For an external view change, a subclass is in change of updating the look up table that is used
+ * for routing. The look up table is in the format of < segment_name -> (replica_id -> server_instance) >.
+ *
+ * When the query comes in, the routing algorithm is as follows:
+ *   1. Randomly pick a replica id (or replica group id)
+ *   2. For each segment of the given table,
+ *      a. Check if the segment can be pruned. If pruned, go to the next segment.
+ *      b. If not pruned, assign the segment to a server with the replica id that is picked above.
+ *
+ */
+public abstract class AbstractPartitionAwareRoutingTableBuilder extends AbstractRoutingTableBuilder {
+  private static final Logger LOGGER = LoggerFactory.getLogger(AbstractPartitionAwareRoutingTableBuilder.class);
+
+  protected static final String PARTITION_METADATA_PRUNER = "PartitionZKMetadataPruner";
+  protected static final int NO_PARTITION_NUMBER = -1;
+
+  // Reference mapping table (segment id, (replica group id, server)) that is used for routing.
+  Map<SegmentId, Map<Integer, ServerInstance>> _segmentNameToServersMapping = new HashMap<>();
+  AtomicReference<Map<SegmentId, Map<Integer, ServerInstance>>> _mappingReference =
+      new AtomicReference<>(_segmentNameToServersMapping);
+
+  // Cache for segment zk metadata to reduce the lookup to ZK store.
+  Map<SegmentId, SegmentZKMetadata> _segmentToZkMetadataMapping = new ConcurrentHashMap<>();
+
+  protected ZkHelixPropertyStore<ZNRecord> _propertyStore;
+  protected SegmentZKMetadataPrunerService _pruner;
+  protected TableConfig _tableConfig;
+  protected Random _random = new Random();
+  protected int _numReplicas;
+
+  @Override
+  public void init(Configuration configuration, TableConfig tableConfig, ZkHelixPropertyStore<ZNRecord> propertyStore) {
+    _tableConfig = tableConfig;
+    _propertyStore = propertyStore;
+
+    // TODO: We need to specify the type of pruners via config instead of hardcoding.
+    _pruner = new SegmentZKMetadataPrunerService(new String[]{PARTITION_METADATA_PRUNER});
+  }
+
+  @Override
+  public Map<ServerInstance, SegmentIdSet> findServers(RoutingTableLookupRequest request) {
+    Map<ServerInstance, SegmentIdSet> result = new HashMap<>();
+    Map<SegmentId, Map<Integer, ServerInstance>> mappingReference = _mappingReference.get();
+    SegmentPrunerContext prunerContext = new SegmentPrunerContext(request.getBrokerRequest());
+
+    // 1. Randomly pick a replica id
+    int replicaGroupId = _random.nextInt(_numReplicas);
+    for (SegmentId segmentId : mappingReference.keySet()) {
+      SegmentZKMetadata segmentZKMetadata = _segmentToZkMetadataMapping.get(segmentId);
+
+      // 2a. Check if the segment can be pruned
+      boolean segmentPruned = (segmentZKMetadata != null) && _pruner.prune(segmentZKMetadata, prunerContext);
+
+      if (!segmentPruned) {
+        // 2b. Segment cannot be pruned. Assign the segment to a server with the replica id picked above.
+        Map<Integer, ServerInstance> replicaId2ServerMapping = mappingReference.get(segmentId);
+        ServerInstance serverInstance = replicaId2ServerMapping.get(replicaGroupId);
+
+        // When the server is not available with this replica id, we need to pick another available server.
+        if (serverInstance == null) {
+          if (!replicaId2ServerMapping.isEmpty()) {
+            serverInstance = replicaId2ServerMapping.values().iterator().next();
+          } else {
+            // No server is found for this segment.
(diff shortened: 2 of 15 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+            .put(ServerInstance.forInstanceName(instanceName), replicaId);
+      }
+    }
+
+    // TODO: This part of code is borrowed from `KafkaLowlevelConsumerRoutingTableBuilder`. This needs to be cleaned.
+
+    // Start of the copy code
+
+    // 1. Gather all segments and group them by Kafka partition, sorted by sequence number
+    Map<String, SortedSet<SegmentName>> sortedSegmentsByKafkaPartition = new HashMap<String, SortedSet<SegmentName>>();
```
mcvsubbu: Not sure I understand why you need a SortedSet and not just a Set here

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -555,6 +587,13 @@ public boolean commitSegmentMetadata(String rawTableName, final String committin
     ZNRecord newZnRecord =
         makeZnRecordForNewSegment(rawTableName, newInstances.size(), newStartOffset, newSegmentNameStr);
     final LLCRealtimeSegmentZKMetadata newSegmentZKMetadata = new LLCRealtimeSegmentZKMetadata(newZnRecord);
+
+    // Add the partition metadata if available.
+    SegmentPartitionMetadata partitionMetadata = getColumnMetadata(realtimeTableName, partitionId);
```
mcvsubbu: We should limit the calls to get table metadata multiple times during a segment commit. Can we cache this?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
snleee: This part of code is borrowed from `KafkaLowLevelConsumingRoutingTableBuilder` and I will clean up this part from another pr. Anyway, the reason for using SortedSet here is to be able to pick the first segment that has `CONSUMING` state.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+import com.linkedin.pinot.broker.routing.RoutingTableLookupRequest;
+import com.linkedin.pinot.common.config.TableConfig;
+import com.linkedin.pinot.common.metadata.segment.SegmentZKMetadata;
+import com.linkedin.pinot.common.response.ServerInstance;
+import com.linkedin.pinot.transport.common.SegmentId;
+import com.linkedin.pinot.transport.common.SegmentIdSet;
+
+
+/**
+ * Abstract partition aware routing table builder for the Kafka low level consumer.
```
mayankshriv: Incorrect documentation? The abstract base class is probably for both realtime as well as offline, right?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+  protected SegmentZKMetadataPrunerService _pruner;
+  protected TableConfig _tableConfig;
+  protected Random _random = new Random();
+  protected int _numReplicas;
+
+  @Override
+  public void init(Configuration configuration, TableConfig tableConfig, ZkHelixPropertyStore<ZNRecord> propertyStore) {
+    _tableConfig = tableConfig;
+    _propertyStore = propertyStore;
+    _pruner = new SegmentZKMetadataPrunerService(new String[]{PARTITION_METADATA_PRUNER});
```
mayankshriv: Not required for this PR, but what pruners to wire-in should come in via config, as opposed to being hard coded here.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+        Map<Integer, ServerInstance> replicaId2ServerMapping = mappingReference.get(segmentId);
+        ServerInstance serverInstance = replicaId2ServerMapping.get(replicaGroupId);
+
+        // When the server is not available with this replica id, we need to pick another available server.
+        if (serverInstance == null) {
+          if (!replicaId2ServerMapping.isEmpty()) {
+            serverInstance = replicaId2ServerMapping.values().iterator().next();
+          } else {
+            // No server is found for this segment.
+            LOGGER.warn("No server is found for the segment {}.", segmentId.getSegmentId());
```
mayankshriv: Should we log a metric for this? Also, should this be sent back as partial response?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
```diff
@@ -42,149 +36,143 @@
...
+ *
+ * We go through multiple steps to get the final look up table. The high level algorithm is as follows:
+ *   1. Compute the partition id set by looking at the segment zk metadata and cache metadata when possible.
+ *   2. Build a mapping table of (partition_id -> (server -> replica_id)) using partition set from step 1 and
+ *      the partition to replica group mapping table from property store.
+ *   3. Compute the final routing look up table from external view and the mapping from step 2.
+ *
+ */
+public class PartitionAwareOfflineRoutingTableBuilder extends AbstractPartitionAwareRoutingTableBuilder {
   private static final int TABLE_LEVEL_PARTITION_NUMBER = 0;
```
mayankshriv: Please add comments to explain the difference between table level vs partition level replica groups

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+    // only in either replica1 (server0, server2) or replica2 (server1, server3). Since the change in the partition
+    // mapping is rare, this additional optimization can benefit us by bringing the concept of replica group even if
+    // it's not explicitly applied. We can fall back to the regular behavior if the kafka partition mapping changes.
+    //
+    _numReplicas = Integer.valueOf(_tableConfig.getValidationConfig().getReplicasPerPartition());
+
+    // Update the cache for the segment ZK metadata
+    for (String segment : externalView.getPartitionSet()) {
+      SegmentId segmentId = new SegmentId(segment);
+      SegmentZKMetadata segmentZKMetadata = _segmentToZkMetadataMapping.get(segmentId);
```
mayankshriv: Seems this map is is to reduce lookups to ZK? If so, please document the same.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * In contrast to the offline partition aware routing builder where the replica group aware segment assignment can be
+ * assumed, we do not use the concept of replica group for a realtime table. The routing look up table is simply built
+ * from the external view.
+ *
+ */
+public class PartitionAwareRealtimeRoutingTableBuilder extends AbstractPartitionAwareRoutingTableBuilder {
+
+  @Override
+  public void computeRoutingTableFromExternalView(String tableName, ExternalView externalView,
```
mayankshriv: Please check if this need to be a 'synchronized' method?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+
+    // Ensure that for each Kafka partition, we have at most one Helix partition (Pinot segment) in consuming state
+    Map<String, SegmentName> allowedSegmentInConsumingStateByKafkaPartition =
+        KafkaLowLevelRoutingTableBuilderUtil.getAllowedConsumingStateSegments(externalView,
+            sortedSegmentsByKafkaPartition);
+
+    RoutingTableInstancePruner pruner = new RoutingTableInstancePruner(instanceConfigList);
+
+    // Compute segment id to replica id to server instance
+    Map<SegmentId, Map<Integer, ServerInstance>> segmentId2ServersMapping = new HashMap<>();
```
mayankshriv: Unrelated to this PR, but seems like SegmentId is just a wrapper around string. We should evaluate if it can be eliminated completely, to avoid unnecessary object creation.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -602,10 +648,26 @@ public boolean commitSegmentMetadata(String rawTableName, final String committin
...
   }
 
-  // TODO: Would be good to cache table config fields which are repeatedly checked
+  /**
+   * Helper function to return cached table config. It updates if the newer version of table config is observed.
+   *
+   * @param tableName name of the table
+   * @return table configuration that reflects the most recent version
+   */
+  private TableConfig getTableConfig(String tableName) {
```
mcvsubbu: please write a more generic cache class like we discussed, and let us have only this class use the cache for now. We can expand that to be a static cache class later.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
snleee: done.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
snleee: done

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+    // Kafka low level does not support for the concept of replica group. However, it assigns servers sequentially
+    // when building the kafka partition table. Let's assume that we have 4 servers, 2 replicas, 4 partitions.
+    //
+    // partition | replica1 | replica2
+    //         0 |  server0 |  server1
+    //         1 |  server2 |  server3
+    //         2 |  server0 |  server1
+    //         3 |  server2 |  server3
+    //
+    // When there is no partitioning on data, we can still reduce the number of servers to route if we pick servers
```
mcvsubbu: No assumption should be made about how the partitions will be assigned across servers. The only assumption that can be made is that the number of replicas will be honored. In fact, this serial assignment that has been pointed out in the comments is actually because helix code had a bug https://issues.apache.org/jira/browse/HELIX-631 Once this is resolved and we start using this call, the Kafka partition assignment will be totally opaque to us.

pinot-broker/src/test/java/com/linkedin/pinot/broker/routing/FakePropertyStore.java
```diff
@@ -0,0 +1,65 @@
...
+      return true;
+    } catch (Exception e) {
+      return false;
+    }
+  }
+
+  public void setContents(String path, ZNRecord contents) throws Exception {
+    _contents.put(path, contents);
+    if (_listener != null) {
+      _listener.handleDataChange(path, contents);
```
mcvsubbu: should really schedule a thread to do this, right? Otherwise, tests may behave incorrectly.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -602,10 +642,26 @@ public boolean commitSegmentMetadata(String rawTableName, final String committin
...
+   * @param tableName name of the table
+   * @return table configuration that reflects the most recent version
+   */
+  private TableConfig getRealtimeTableConfig(String tableName) {
+    TableConfig tableConfig;
+    try {
+      String tableNameWithType = TableNameBuilder.REALTIME.tableNameWithType(tableName);
+      tableConfig = _tableConfigCache.getTableConfig(tableNameWithType);
+    } catch (ExecutionException e) {
+      LOGGER.warn("Exception happened while loading the table config ({}) from the property store to the cache.", tableName);
```
mcvsubbu: add the exception to the log. Also, log tableNameWitType instead of tableName

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -442,6 +458,28 @@ private IdealState addLLCRealtimeSegmentsInIdealState(final IdealState idealStat
     return idealState;
   }
 
+  private SegmentPartitionMetadata getPartitionMetadataFromTableConfig(String tableName, int partitionNum, int partitionId) {
```
mcvsubbu: Confusing variable naming here. What is partitionNum? If it is number of partitions (I think that is it), then please rename it as numPartitions. Otherwise, it is unclear what the difference is between partitionNum and partitionId

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -552,9 +590,11 @@ public boolean commitSegmentMetadata(String rawTableName, final String committin
     LLCSegmentName newHolder = new LLCSegmentName(oldSegmentName.getTableName(), partitionId, newSeqNum, now);
     final String newSegmentNameStr = newHolder.getSegmentName();
 
-    ZNRecord newZnRecord =
-        makeZnRecordForNewSegment(rawTableName, newInstances.size(), newStartOffset, newSegmentNameStr);
+    ZNRecord newZnRecord = makeZnRecordForNewSegment(rawTableName, newInstances.size(), newStartOffset, newHolder,
```
mcvsubbu: better to assign an integer as
final int numPartitions = partitionAssignment.getListFields.size();

and then use it

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -940,15 +997,23 @@ private void createSegment(String realtimeTableName, int numReplicas, int partit
   }
 
   private ZNRecord makeZnRecordForNewSegment(String realtimeTableName, int numReplicas, long startOffset,
-      String newSegmentNameStr) {
+      LLCSegmentName newSegmentName, int partitionSize) {
```
mcvsubbu: partitionSize? numPartitions? Not sure what it means here.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
```diff
@@ -99,18 +109,34 @@ public RoutingTableBuilder createRoutingTableBuilder(TableConfig tableConfig) {
...
+            (validationConfig.getSegmentPushType() != null && !validationConfig.getSegmentPushType()
                 .equalsIgnoreCase("REFRESH"))) {
+          builder = new PartitionAwareOfflineRoutingTableBuilder();
+        } else {
           builder = new DefaultOfflineRoutingTableBuilder();
+        }
+        break;
+      case PartitionAwareRealtime:
+        // In order to use the partition aware routing for realtime table, we assume that the following.
+        // 1. The table push type is not 'refresh'
```
snleee: I was not aware that the realtime segments do not care about the push type. I removed this check.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
```diff
@@ -99,18 +109,34 @@ public RoutingTableBuilder createRoutingTableBuilder(TableConfig tableConfig) {
...
+        // 2. The table uses LLC kafka consumer.
+        validationConfig = tableConfig.getValidationConfig();
+        KafkaStreamMetadata streamMetadata = new KafkaStreamMetadata(tableConfig.getIndexingConfig().getStreamConfigs());
+
+        if ((validationConfig.getSegmentPushType() != null && !validationConfig.getSegmentPushType()
+            .equalsIgnoreCase("REFRESH")) &&
```
snleee: Same as above

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/RoutingTableBuilderFactory.java
```diff
@@ -99,18 +109,34 @@ public RoutingTableBuilder createRoutingTableBuilder(TableConfig tableConfig) {
       case PartitionAwareOffline:
         // In order to use the partition aware routing for offline table, we assume the following conditions.
         // 1. The replica group aware segment assignment strategy is used.
-        // 2. Table config contains the replica group segment assignment related configurations.
-        // 3. The table push type is not 'refresh'.
-        String segmentAssignmentStrategy = tableConfig.getValidationConfig().getSegmentAssignmentStrategy();
+        // 2. The table push type is not 'refresh'.
         SegmentsValidationAndRetentionConfig validationConfig = tableConfig.getValidationConfig();
+        String segmentAssignmentStrategy = validationConfig.getSegmentAssignmentStrategy();
```
snleee: done

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
snleee: done

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
snleee: This is the case when all replicas for the segment are somehow missing. (i.e. no server that holds this segment is found). `serverInstance = replicaId2ServerMapping.values().iterator().next()` will get the different replica.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
snleee: Thank you for explaining the context of serial assignment. If that's the case, we cannot enforce this optimization. I will remove this `TODO` comment.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
snleee: added `TODO` comment on `SegmentId` class

pinot-broker/src/test/java/com/linkedin/pinot/broker/routing/FakePropertyStore.java
snleee: I moved out this fake property store from `RoutingTableTest`. Some tests in `RoutingTableTest` requires the `IZkDataListener` and this lines.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
snleee: done

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
snleee: done

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
snleee: done

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/PartitionAwareRealtimeRoutingTableBuilder.java
snleee: This is getting called from `ClusterChangeMediator` which uses a single thread. So, I think that we don't need `synchronized` method.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/AbstractPartitionAwareRoutingTableBuilder.java
mcvsubbu: ok, in that case a WARN is warranted. But, are we also returning partial results?  IIRC that is handled elsewhere in the code, but just making sure.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/TableConfigCache.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
+
+
+/**
+ * Cache for table config.
+ */
+public class TableConfigCache {
+  private static final long DEFAULT_CACHE_SIZE = 50;
+  private static final long DEFAULT_CACHE_TIMEOUT_IN_MINUTE = 5;
```
mcvsubbu: should make these two configurable.

snleee: I will address this from the separate PR.

PR #1606, opened 2017-06-23
Fix for star tree to handle offset larger than int

kishoreg:


pinot-core/src/main/java/com/linkedin/pinot/core/startree/StarTreeDataSorter.java
```diff
@@ -207,15 +206,37 @@ public void remove() {
         byte[] dimBuff = new byte[dimensionSizeInBytes];
         byte[] metBuff = new byte[metricSizeInBytes];
 
-        mappedByteBuffer.copyTo(startOffset + (pointer * totalSizeInBytes), dimBuff, 0, dimensionSizeInBytes);
+        copyTo(startOffset + (pointer * totalSizeInBytes), dimBuff, 0, dimensionSizeInBytes);
```
mayankshriv: Can (pointer * totalSizeInBytes) overflow?

kishoreg: overflow long you mean? I dont think we can handle offsets that overflow long.

kishoreg:
The test case I used for testing failed before this change and passed after this change. It was a 64G file. That should have ensured we handle offsets larger than int.max

pinot-core/src/main/java/com/linkedin/pinot/core/startree/StarTreeDataSorter.java
mayankshriv: No, I meant that since 'pointer' and 'totalSizeInBytes' are both integer, their multiplication can cause integer overflow unless one of them is upcasted to long. Filed PR: https://github.com/linkedin/pinot/pull/1664 for the same.

PR #1609, opened 2017-06-23
Remove all segments from store when a table is deleted

mcvsubbu:


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1311,6 +1311,8 @@ public void deleteOfflineTable(String tableName) {
     if (!_helixAdmin.getResourcesInCluster(_helixClusterName).contains(offlineTableName)) {
```
mayankshriv: I understand this is not part of this PR, but do we need to flag an error if table name is not found?

mcvsubbu: u r right. I thought that is handled at a higher level. Or, perhaps each of these calls are idempotent and run fine. i am not sure. But then we can do that as a separate PR. Not deleting segments is biting us now, so I will check this in

mcvsubbu:
@Jackie-Jiang yes, we don't delete segments when we drop a table.

PR #1614, opened 2017-06-27
[TE] filter-select component

justYves:
### What's new: 
Added a `filter-select` component that wraps around `power-select-multiple`. 
The default `power-select-multiple` component currently doesn't support lazy loading rendering; In thirdeye, the filter search can potentially return more than 1000 items, and this can take more than 15 sec to render. 

The `filter-select` component uses ember-concurrency's task to debounce searches and display result by group if a group has less than 25 items or the total search has less than 100 items (customizable) 

### Usage: 

```
{{filter-select
    options=metricFilters  // (required)  
    selected=filters  // (required) JSON filters (from endpoint)
    maxNumFilters=25  // (optional {default: 25}) max number of filters per group
    maxTotalFilters=100  // (optional {default:100}) max number of filters for the search
}}
```

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/filter-select/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/filter-select/component.js
@@ -1,0 +1,181 @@
+import Ember from 'ember';
+import { task, timeout } from 'ember-concurrency';
+
+/**
+ * Parse the filter Object and return a Object of groupName
+ * containing Arrays of filter options
+ * @param {Object} filters
+ */
+const buildFilterOptions = (filters) => {
+  return Object.keys(filters).map((filterName) => {
+
+    const options = filters[filterName]
+      .filter(value => !!value)
+      .map((value) => {
+        return {
+          name: value,
+          id: `${filterName}::${value}`
+        }
+      })
+
+    return {
+      groupName: `${filterName}`,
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/pods/components/filter-select/component.js
```diff
@@ -0,0 +1,181 @@
...
+    ? `Too Many results found (${count})`
+    : 'No results found.'
+
+  return [filters, NoMatchMessage];
+};
+
+
+export default Ember.Component.extend({
+  // Maximum filters by filter group
+  maxNumFilters: 25,
```
newsummit: So, groups with more than 25 filters will be hidden, or will only display 25? Good performance compromise. We can figure out a way for user to further expand a filter group, perhaps.

justYves: groups with more than 25 will be shown as `pageKey (688)`.
Yes, ultimately, we would want to have the expand and lazy loading, but that would potentially require our own fork of the power-select, or working with the open source project.

justYves: you can also pass any number you want into the component 
`maxNumFilters=40`

PR #1618, opened 2017-06-27
Refactor and optimize all integration tests (part 1)

Jackie-Jiang:
1. Added ClusterIntegrationTestUtils for all commonly used utils for cluster integration tests
2. Modified BaseClusterIntegrationTest to make it the base for all cluster integration test, and make most settings configurable
3. Move all commonly used tests into BaseClusterIntegrationTestSet
4. Merged DefaultColumnsClusterIntegrationTest, DefaultColumnsTriggerClusterIntegrationTest, InstanceDecommissionIntegrationTest, InvertedIndexOfflineIntegrationTest, MultipleNodeOfflineClusterIntegrationTest, NoDictionaryIntegrationTest into the OfflineClusterIntegrationTest
5. Merge InvertedIndexRealtimeIntegrationTest into RealtimeIntegrationTest
6. Removed the following tests that test nothing: FileBasedSentineTest, FlightsHybridClusterScanComparisonIntegrationTest, LLCPartitioningIntegrationTest, RealtimeSchemaChangeIntegrationTest
7. Refactored and optimized all other tests to use the new base and util classes


pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTest.java
```diff
@@ -16,1402 +16,322 @@
...
+  /**
+   * Get the schema file.
+   *
+   * @return Schema file
+   */
+  @Nonnull
+  protected File getSchemaFile() {
+    URL resourceUrl = BaseClusterIntegrationTest.class.getClassLoader().getResource(getSchemaFileName());
+    Assert.assertNotNull(resourceUrl);
+    return new File(resourceUrl.getFile());
```
jfim: Does this actually work properly? I'm pretty sure that when the tests are JAR'ed, this call will fail due to using a JAR URL from which a File cannot be created.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTest.java
```diff
@@ -16,1402 +16,322 @@
...
+              return true;
+            }
           }
-
-          return true;
-        } catch (JSONException e) {
-          LOGGER.warn("Caught exception while reading the routing table, will retry", e);
           return false;
+        } catch (Exception e) {
+          return null;
```
jfim: Shouldn't we rethrow the exception here?

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTest.java
```diff
@@ -16,1402 +16,322 @@
...
+   * @param timeoutMs Timeout in milliseconds
+   * @throws Exception
+   */
+  protected void waitForAllDocsLoaded(long timeoutMs)
+      throws Exception {
+    final long countStarResult = getCountStarResult();
+    TestUtils.waitForCondition(new Function<Void, Boolean>() {
       @Nullable
       @Override
-      public Boolean apply(@Nul
```
jfim: Same.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTest.java
Jackie-Jiang: Lots of tests are written this way. Will we ever access JAR'ed tests?

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTest.java
Jackie-Jiang: Depends. Throw exception means we don't want to try again, while return null and false are the same. I kept the old behavior here.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTest.java
jfim: I'm guessing it works mostly because nothing depends on the integration tests, so that works, as it uses JARs for all test dependencies except the module you're testing.

You can ignore this comment for now, I don't foresee us depending on the test JARs from the integration test.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTest.java
jfim: Ah I see, this is part of the loop in `waitForCondition`, makes sense.

PR #1621, opened 2017-06-28
[PINOT-4926] Introduce a memory manager so that we can switch between Mmap and direct allocation

mcvsubbu:
Introduced the abstract class OffHeapMemoryManager, and the two concrete sub-classes DirectMemoryManager
and MmapMemoryManager.

Re-factored all the classes allocating memory for real-time segments to use the OffHeapMemoryManager abstract
class to allocate memory.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
@@ -1,0 +1,71 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.linkedin.pinot.core.io.readerwriter;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.List;
+import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
+
+
+/**
+ * @class OffHeapMemoryManager is an abstract class that implements base functionality to allocate and release
+ * memory that is acquired during realtime segment consumption.
+ *
+ * Realtime consuming segments use memory for dictionary, forward index, and inverted indices. For off-heap
+ * allocation of memory, we instantiate one OffHeapMemoryManager for each segment in the server,
+ *
+ * Closing the OffHeapMemoryManager also releases all the resources allocated by the OffHeapMemoryManager.
+ */
+public abstract class OffHeapMemoryManager implements Closeable {
+  protected final List<PinotDataBuffer> _buffers = new LinkedList<>();
+  protected final String _segmentName;
+
+  protected OffHeapMemoryManager(String segmentName) {
+    _segmentName = segmentName;
+  }
+
+  /**
+   * @return A string representing a context for all allocations using this policy
+   */
+  public String getSegmentName() {
+    return _segmentName;
+  }
+
+  /**
+   * Allocate memory
+   * @param size size of memory
+   * @param columnName Name of the column for which memory is being allocated
+   * @return PinotDataBuffer
+   */
+  public abstract  PinotDataBuffer allocate(long size, String columnName);
+
+  public abstract void doClose();
+
+  /**
+   * Close out this memory manager and release all memory and resources
+   * @throws IOException
+   */
+  public void close() throws IOException {
+    for (PinotDataBuffer buffer : _buffers) {
+      buffer.close();
+    }
+    doClose();
+    _buffers.clear();
+  }
+}
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.List;
 import com.linkedin.pinot.core.io.reader.impl.FixedByteSingleValueMultiColReader;
 import com.linkedin.pinot.core.io.readerwriter.BaseSingleColumnMultiValueReaderWriter;
+import com.linkedin.pinot.core.io.readerwriter.OffHeapMemoryManager;
 import com.linkedin.pinot.core.io.writer.impl.FixedByteSingleValueMultiColWriter;
 import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
 
@@ -108,13 +109,15 @@ public class FixedByteSingleColumnMultiValueReaderWriter extends BaseSingleColum
   private int columnSizeInBytes;
   private int maxNumberOfMultiValuesPerRow;
   private final int rowCountPerChunk;
-  private final String description;
+  private final OffHeapMemoryManager memoryManager;
+  private final String columnName;
   private int prevRowStartIndex = 0;  // Offset in the databuffer for the last row added.
   private int prevRowLength = 0;  // Number of values in the column for the last row added.
 
   public FixedByteSingleColumnMultiValueReaderWriter(int maxNumberOfMultiValuesPerRow, int avgMultiValueCount, int rowCountPerChunk,
-      int columnSizeInBytes, String description) {
-    this.description = description;
+      int columnSizeInBytes, OffHeapMemoryManager memoryManager, String columnName) {
+    this.memoryManager = memoryManager;
+    this.columnName = columnName;
     int initialCapacity = Math.max(maxNumberOfMultiValuesPerRow, rowCountPerChunk * avgMultiValueCount);
     int incrementalCapacity =
         Math.max(maxNumberOfMultiValuesPerRow, (int) (initialCapacity * 1.0f * INCREMENT_PERCENTAGE / 100));
@@ -130,7 +133,7 @@ public class FixedByteSingleColumnMultiValueReaderWriter extends BaseSingleColum
   }
 
   private void addHeaderBuffers() {
-    headerBuffer = PinotDataBuffer.allocateDirect(headerSize, description);
+    headerBuffer = memoryManager.allocate(headerSize, columnName);
     // We know that these bufffers will not be copied directly into a file (or mapped from a file).
     // So, we can use native byte order here.
     headerBuffer.order(ByteOrder.nativeOrder());
@@ -153,7 +156,7 @@ public class FixedByteSingleColumnMultiValueReaderWriter extends BaseSingleColum
   private void addDataBuffers(int rowCapacity) throws RuntimeException {
     PinotDataBuffer dataBuffer;
     try {
-      dataBuffer = PinotDataBuffer.allocateDirect(rowCapacity * columnSizeInBytes, description);
+      dataBuffer = memoryManager.allocate(rowCapacity * columnSizeInBytes, columnName);
       dataBuffer.order(ByteOrder.nativeOrder());
       dataBuffers.add(dataBuffer);
       currentDataWriter =
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.List;
 import com.linkedin.pinot.core.io.reader.impl.FixedByteSingleValueMultiColReader;
 import com.linkedin.pinot.core.io.readerwriter.BaseSingleColumnSingleValueReaderWriter;
+import com.linkedin.pinot.core.io.readerwriter.OffHeapMemoryManager;
 import com.linkedin.pinot.core.io.writer.impl.FixedByteSingleValueMultiColWriter;
 import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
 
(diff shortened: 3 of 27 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/DirectMemoryManager.java
```diff
@@ -0,0 +1,39 @@
...
+// Allocates memory using direct allocation
+public class DirectMemoryManager extends OffHeapMemoryManager {
+
+  public DirectMemoryManager(final String segmentName) {
+    super(segmentName);
+  }
+
+  @Override
+  public PinotDataBuffer allocate(long size, String columnName) {
+    return PinotDataBuffer.allocateDirect(size, getSegmentName() + "." + columnName);
```
antumbde: you need to add this buffer to _buffers list to correctly release memory

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
```diff
@@ -0,0 +1,71 @@
...
+    return _segmentName;
+  }
+
+  /**
+   * Allocate memory
+   * @param size size of memory
+   * @param columnName Name of the column for which memory is being allocated
+   * @return PinotDataBuffer
+   */
+  public abstract  PinotDataBuffer allocate(long size, String columnName);
```
antumbde: See comment on DirectBuffers not adding buffer to _buffers which can cause leak.

It may be better to provide abstract protected allocateInternal method (similar to doClose) that derived classes can implement to provide buffer. This method can add that to _buffers (since _buffers is a member of this class)

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public abstract void doClose();
```
antumbde: private

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Close out this memory manager and release all memory and resources
+   * @throws IOException
+   */
+  public void close() throws IOException {
```
antumbde: Consider logging and swallowing IOException here. Higher layers can hardly take action on that

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * @class OffHeapMemoryManager is an abstract class that implements base functionality to allocate and release
+ * memory that is acquired during realtime segment consumption.
+ *
+ * Realtime consuming segments use memory for dictionary, forward index, and inverted indices. For off-heap
+ * allocation of memory, we instantiate one OffHeapMemoryManager for each segment in the server,
+ *
+ * Closing the OffHeapMemoryManager also releases all the resources allocated by the OffHeapMemoryManager.
+ */
+public abstract class OffHeapMemoryManager implements Closeable {
```
antumbde: I prefer IndexOffHeapMemoryManager given limited scope of memory management of this class

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
```diff
@@ -0,0 +1,71 @@
...
+  protected final List<PinotDataBuffer> _buffers = new LinkedList<>();
+  protected final String _segmentName;
+
+  protected OffHeapMemoryManager(String segmentName) {
```
antumbde: (If I correctly understand that this is Index level memory management) constructor argument should be 'context' or segmentName + indexName. Our Current allocationContext uses more dscriptive strings that can sometimes aid debugging

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MmapMemoryManager.java
```diff
@@ -0,0 +1,121 @@
...
+      throw new RuntimeException("File " + filePath + " already exists");
+    }
+    file.deleteOnExit();
+    RandomAccessFile raf;
+    try {
+      raf = new RandomAccessFile(filePath, "rw");
+    } catch (FileNotFoundException e) {
+      throw new RuntimeException(e);
+    }
+    long fileLen = Math.max(FILE_LENGTH, len);
```
antumbde: Still reading through code...trying to understand how length is passed in.
We need upper bound on fileLen to 2G. PinotDataBuffer won't handle sizes larger than that.
Why is fileLen capped at 0.5G vs say 2G limit. It's better to have a large file than multiple small ones (truncate() on close() to limit size)

ALso, if it's 0.5G why not extend the file instead of mapping new file ?

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MmapMemoryManager.java
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * @param size size of memory to be mmaped
+   * @param columnName Name of the column for which memory is being allocated
+   * @return
+   */
+  @Override
+  public PinotDataBuffer allocate(long size, String columnName) {
+    addFileIfNecessary(size);
+    PinotDataBuffer buffer = _currentBuffer.view(_availableOffset, _availableOffset+size);
+    _availableOffset += size;
```
antumbde: I think this _availableOffset calculation if off
First allocation: 512M, _availableOffset = 512M
Second allocation: 250M, _availableOffset = 500 + 250 = 750M (but we've allocated 512M buffer, _curFileLen = 512M), 
Third alloc: 100M,  then in addFileIfNecessary: len (100M) + _availableOffset(750M)  < _curFileLen is false  and we will allocate extra buffer

You probably should just = here instead of +=

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -108,18 +110,21 @@ public FixedByteSingleValueMultiColReader getReader() {
...
+   * @param columnName
    */
-  public FixedByteSingleColumnSingleValueReaderWriter(int numRowsPerChunk, int columnSizesInBytes, String description) {
+  public FixedByteSingleColumnSingleValueReaderWriter(int numRowsPerChunk, int columnSizesInBytes,
+      OffHeapMemoryManager memoryManager, String columnName) {
     chunkSizeInBytes = numRowsPerChunk * columnSizesInBytes;
     this.numRowsPerChunk = numRowsPerChunk;
     this.columnSizesInBytes = columnSizesInBytes;
-    this.description = description;
+    this.memoryManager = memoryManager;
```
antumbde: Who has the responsibility to close() memoryManager  ? Please add to javadoc

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
mcvsubbu: done, changed to RealtimeIndexOffHeapMemoryManager.

Kept the derived classes the same, though.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
mcvsubbu: In the realtime indexing, this memory manager is created each time a segment is created. And then for each column, we may allocate memory multiple times. I will add these as comments in the base class.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
mcvsubbu: agreed. done.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
mcvsubbu: private does not work, so made it protected

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/OffHeapMemoryManager.java
mcvsubbu: the only higher layer here is realtime segment impl, and that logs (and swallows) the exception. IF we log there, we get the segment name  in the log, as it does for other things in the destroy() method.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
mcvsubbu: The creator of memory manager is expected to close it (in our case there is only one creator, RealtimeSegmentImpl)

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/DirectMemoryManager.java
mcvsubbu: yes, added.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MmapMemoryManager.java
mcvsubbu: length is decided by the caller who wants the buffer. All callers make sure that  length is less than 2g. they have their own algorithms for deciding the length (e.g. use prev len, or double prev len, etc.)

mcvsubbu: I want to pick a number that is high enough so that we can get multiple buffers, but low enoug that we don't run out of disk trying to ingest 100 partitions in a host. 2g will be an overkill, so I picked 0.5g. I think I should reduce it a bit, maybe.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MmapMemoryManager.java
mcvsubbu: It cannot be = because we need to add the current buf size to get the next available offset. 

I ran a unit test with your test case, and am seeing the right results.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MmapMemoryManager.java
```diff
@@ -0,0 +1,121 @@
...
+   * @param dirPathName directory under which all mmap files are created.
+   * @param segmentName Name of the segment for which this memory manager allocates memory
+   */
+  public MmapMemoryManager(String dirPathName, String segmentName) {
+    super(segmentName);
+    _dirPathName = dirPathName;
+  }
+
+  private void addFileIfNecessary(long len) {
+    if (len + _availableOffset < _curFileLen) {
```
mcvsubbu: This condition should be <=, so that we can accommodate a chunk 1 byte longer in the same buffer. 

e.g, if a bufer is 4 bytes long, the first alloc asks for 1 byte. avail offset will become 1. At that point, we should be able to afford an allocation of 3 bytes in the same buffer

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MmapMemoryManager.java
mcvsubbu: In fact, we are a little conservative in the super-class, so I changed that, added a unit test for corner cases.

antumbde: Actually, _availableOffset is = to latest allocated size because addFileIfNecessary sets it to zero. So, += is effectively = size. That's different from your comment. Is this expected ?

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/RealtimeIndexOffHeapMemoryManager.java
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * Method to be implemented by inheriting concrete classes
+   */
+  protected abstract void doClose();
+
+  /**
+   * Sub-classes are required to call this method after they successfully allocate a PinotDataBuffer.
+   * @param buffer
+   */
+  protected void onBufferAdded(PinotDataBuffer buffer) {
```
antumbde: I would have preferred allocate to be a wrapper for abstract method, say, allocateInternal instead of having all subclasses do right thing by calling onBufferAdded. Any extension can easily miss calling onBufferAdded and tests will rarely catch that

mcvsubbu: yes, agreed, will do

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/MmapMemoryManager.java
mcvsubbu: what if addIfNecessary does not create a new file?

pinot-core/src/test/java/com/linkedin/pinot/core/indexsegment/utils/OffHeapStringStoreTest.java
```diff
@@ -19,14 +19,15 @@
...
+import com.linkedin.pinot.core.io.writer.impl.DirectMemoryManager;
 import com.linkedin.pinot.core.io.writer.impl.OffHeapStringStore;
 
 
 public class OffHeapStringStoreTest {
   private static Random RANDOM = new Random();
   @Test
   public void maxValueTest() throws Exception {
-    OffHeapStringStore store = new OffHeapStringStore();
+    OffHeapStringStore store = new OffHeapStringStore(new DirectMemoryManager("test"), "stringColumn");
```
mcvsubbu: yes, agreed.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnMultiValueReaderWriter.java
```diff
@@ -20,6 +20,7 @@
 import java.util.List;
 import com.linkedin.pinot.core.io.reader.impl.FixedByteSingleValueMultiColReader;
 import com.linkedin.pinot.core.io.readerwriter.BaseSingleColumnMultiValueReaderWriter;
+import com.linkedin.pinot.core.io.readerwriter.RealtimeIndexOffHeapMemoryManager;
```
rven1: Should this class know about a realtime specific memory manager, or should we reference an interface? Wouldn't this class also be called for offline segments.

mcvsubbu: This class is not used by oflfine segments

PR #1622, opened 2017-06-28
Fixed the bug for replica group aware segment assignment strategy.

snleee:
The expected behavior is to randomly pick a server from the pool of servers
within a replica group; however, it has a bug that it picks the index using
random(0, numReplica).


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/sharding/ReplicaGroupSegmentAssignmentStrategy.java
```diff
@@ -85,15 +84,15 @@
     for (int groupId = 0; groupId < numReplicas; groupId++) {
       List<String> instancesInReplicaGroup =
           partitionToReplicaGroupMapping.getInstancesfromReplicaGroup(partitionNumber, groupId);
-
+      int instanceSize = instancesInReplicaGroup.size();
```
mayankshriv: s/instanceSize/numInstances

PR #1623, opened 2017-06-28
Implement an endpoint that surfaces the baseline calculated by anomaly functions

cyy0714:
1. Add an end-point that can surface the baseline from anomaly functions to the user interface, even though no anomalies are detected. Normally, the baseline from anomaly functions can be surfaced when there is an anomaly. But if user wants to know the baseline information, there is no way to surface that yet. This endpoint can help thirdeye surface the baseline from anomaly functions to users.

2. Unify the baseline calculation in the AnomaliesResource, so that both AnomalyDetail and the new end-point use the same function to calculate the baseline values.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/AnomalyDetectionInputContext.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/AnomalyDetectionInputContext.java
@@ -13,19 +13,19 @@ import java.util.Map;
 
 
 public class AnomalyDetectionInputContext {
-  Map<DimensionMap, MetricTimeSeries> dimensionKeyMetricTimeSeriesMap = Collections.emptyMap();
+  Map<DimensionMap, MetricTimeSeries> dimensionMapMetricTimeSeriesMap = Collections.emptyMap();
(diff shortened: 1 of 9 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/AnomalyDetectionInputContext.java
```diff
@@ -13,19 +13,19 @@
 
 
 public class AnomalyDetectionInputContext {
-  Map<DimensionMap, MetricTimeSeries> dimensionKeyMetricTimeSeriesMap = Collections.emptyMap();
+  Map<DimensionMap, MetricTimeSeries> dimensionMapMetricTimeSeriesMap = Collections.emptyMap();
```
bl44: Is it absolutely necessary to rename the DimensionKeyMetric...Map to DimensionMapMetric...Map? if not let's keep it.

cyy0714: I think it is necessary. Using dimensionKeyMetricTimeSeriesMap is misleading. We should remain the consistency with the key-value type with the name.

PR #1629, opened 2017-06-29
Avoid creation of byte[] during DataTable deserialization.

mayankshriv:
During deserialization of DataTable, we create new byte[] and copy out
bytes from a ByteBuffer, and then wrap byte[] into a ByteBuffer for
deserialization. This change eliminates the creation of temporary byte[]
and its wrapping into ByteBuffer by directly slicing out a new
ByteBuffer from the original ByteBuffer.

While the total number of objects created does not reduce, creation of
byte[] and data copy is replaced by creation of shallow ByteBuffer
slices.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/common/datatable/ObjectCustomSerDe.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/common/datatable/ObjectCustomSerDe.java
@@ -110,6 +110,54 @@ public class ObjectCustomSerDe {
   }
 
   /**
+   * Given a ByteBuffer and the object type, de-serialize the ByteBuffer into an object.
+   */
+  @SuppressWarnings("unchecked")
+  @Nonnull
+  public static <T> T deserialize(@Nonnull ByteBuffer byteBuffer, @Nonnull ObjectType objectType)
+      throws IOException {
+    switch (objectType) {
+      case String:
+        byte[] bytes = getBytesFromByteBuffer(byteBuffer);
+        return (T) new String(bytes, UTF_8);
+      case Long:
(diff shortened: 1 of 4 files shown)


Jackie-Jiang:
Are there perf numbers supporting this optimization?

pinot-core/src/main/java/com/linkedin/pinot/core/common/datatable/ObjectCustomSerDe.java
```diff
@@ -241,21 +289,42 @@ private static DoubleArrayList deserializeDoubleArray(byte[] bytes) {
...
   }
 
   /**
+   * Helper method to slice out a ByteBuffer of given size from the original ByteBuffer.
+   *
+   * @param byteBuffer Original ByteBuffer.
+   * @param keyNumBytes Number of bytes to slice out.
+   * @return New ByteBuffer containing a slice of the original ByteBuffer.
+   */
+  private static ByteBuffer getByteBufferSlice(ByteBuffer byteBuffer, int keyNumBytes) {
```
mcvsubbu: nit: s/keyNumBytes/numBytes

mayankshriv:
@Jackie-Jiang I was able to measure a clear reduction in overall size of byte[] garbage (about 10x, for the specific production workload I chose). The number of GC'd objects remains the same though because byte[] objects are replaced by shallow ByteBuffer objects.

pinot-core/src/main/java/com/linkedin/pinot/core/common/datatable/ObjectCustomSerDe.java
mayankshriv: Thanks for catching, this is the by-product of IDE refactoring.

PR #1641, opened 2017-07-10
Optimized TablesResourceTest

snleee:
- Removed unnecessary segment generation
- Improved the performance from 12s to 4s locally

### DIFF
--- a/pinot-server/src/test/java/com/linkedin/pinot/server/api/resources/TablesResourceTest.java
+++ b/pinot-server/src/test/java/com/linkedin/pinot/server/api/resources/TablesResourceTest.java
@@ -16,7 +16,6 @@
 
 package com.linkedin.pinot.server.api.resources;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linkedin.pinot.common.restlet.resources.TableSegments;
 import com.linkedin.pinot.common.restlet.resources.TablesList;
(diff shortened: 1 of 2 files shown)


pinot-server/src/test/java/com/linkedin/pinot/server/api/resources/TablesResourceTest.java
```diff
@@ -190,4 +185,13 @@ public void testSegmentCrcMetadata() throws Exception {
       assertEquals(crcMap.get(segmentName), crc);
     }
   }
+
+  private Map<String, String> fetchCrcInformation(String urlFormat) throws Exception {
+    // Trigger crc api to fetch crc information
+    String response = testHelper.target.path(String.format(urlFormat, ResourceTestHelper.DEFAULT_TABLE_NAME))
+        .request().get(String.class);
+    ObjectMapper mapper = new ObjectMapper();
```
Jackie-Jiang: Make ObjectMapper a private static final member variable. Also, change the import to use codehaus one.

snleee:
@Jackie-Jiang On travis, this test takes 16.48s (49s previously). If we want to further improve, we should avoid using `SegmentIndexCreationDriver`. How do you think?

Jackie-Jiang:
@snleee You mean avoid creating segment?

PR #1642, opened 2017-07-10
Optimize BitmapInvertedIndexTest

Jackie-Jiang:
1. Reduce the number of records checked to 1000
2. Pick 3 columns to build inverted index on
3. Refactor to use API in IndexSegment instead of IndexSegmentImpl

Locally 46sec to 1sec


Jackie-Jiang:
@mcvsubbu Added

PR #1645, opened 2017-07-11
Removed all SentinelTests, refactor fastHll tests

Jackie-Jiang:
All sentinal tests are basically simulate the query execution flow, which are all covered in integration tests.
Unit tests should focus on small parts but avoid simulate the whole flow.
Added FastHllQueriesTest to check the functionality of fastHll aggregation function.


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/AvroRecordToPinotRowGenerator.java
```diff
@@ -95,22 +95,4 @@ public GenericRow transform(GenericData.Record record, org.apache.avro.Schema sc
 
     return destination;
   }
-
-  public GenericRow transform(GenericRecord avroRecord, GenericRow destination) {
```
mcvsubbu: make sure that this is not used outside (e.g. hadoop jobs)

Jackie-Jiang: Checked. And it's not safe to use because it does not handle null multi-value column case.

pinot-core/src/test/java/com/linkedin/pinot/queries/PregeneratedHllTest.java
```diff
@@ -1,153 +0,0 @@
...
-import org.apache.commons.io.FileUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.testng.Assert;
-import org.testng.annotations.AfterClass;
-import org.testng.annotations.BeforeClass;
-import org.testng.annotations.Test;
-
-
-public class PregeneratedHllTest {
```
antumbde: This test needs to exist to test segment generation for pregenerated hll column. It's okay to use different (non-default) value to limit segment size

pinot-core/src/test/java/com/linkedin/pinot/queries/PregeneratedHllTest.java
```diff
@@ -1,153 +0,0 @@
...
-
-      Double val1 = (Double)ApproximateQueryTestUtil
-          .runQuery(testHelper.queryExecutor, Arrays.asList(helper.getSegmentName()), aggCalls.get(0),
-              testHelper.serverMetrics);
-      // 5000, because we know
-      Assert.assertTrue(val1 > 5000);
-      // add one more segment. Goal to verify that combine works correctly
-      // across multiple segments. Not interested in exact values
-      final String secondSegmentName = "segment2";
-      createLoadSegment(secondSegmentName, destinationDir);
```
antumbde: This part of the code also verifies that combining hll results across segments works correctly. We need to preserve that test functionality (wherever we think is appropriate location)

pinot-core/src/test/java/com/linkedin/pinot/queries/FastHllQueriesTest.java
```diff
@@ -0,0 +1,257 @@
...
+    Assert.assertEquals(executionStatistics.getNumEntriesScannedInFilter(), 0L);
+    Assert.assertEquals(executionStatistics.getNumEntriesScannedPostFilter(), 13839L);
+    Assert.assertEquals(executionStatistics.getNumTotalRawDocs(), 30000L);
+    AggregationGroupByResult aggregationGroupByResult = resultsBlock.getAggregationGroupByResult();
+    GroupKeyGenerator.GroupKey firstGroupKey = aggregationGroupByResult.getGroupKeyIterator().next();
+    Assert.assertEquals(firstGroupKey.getStringKey(), "");
+    Assert.assertEquals(((HyperLogLog) aggregationGroupByResult.getResultForKey(firstGroupKey, 0)).cardinality(), 23L);
+    Assert.assertEquals(((HyperLogLog) aggregationGroupByResult.getResultForKey(firstGroupKey, 1)).cardinality(), 596L);
+
+    // Test inter segment queries
```
antumbde: This comment is misleading. We are testing queries on a single segment only. See comment below for PregeneratedHllTest about preserving test functionality for combining results across segments.

pinot-core/src/test/java/com/linkedin/pinot/queries/FastHllQueriesTest.java
```diff
@@ -0,0 +1,257 @@
...
+    deleteSegment();
+  }
+
+  @Test
+  public void testFastHllWithLog2m8() throws Exception {
+    buildSegment(8);
+
+    // Test inner segment queries
+    // Test query without filter
+    AggregationOperator aggregationOperator = getOperatorForQuery(BASE_QUERY);
```
antumbde: [Suggestion] It's better to use TestNG's DataProvider to provide the segment(s), query and result validator for the test. That way, we can define additional entry in data provider if our segment format or set of queries changes. All the tests can be run for those new formats by defining data provider. We won't have to change actual tests.

This was a major handicap when we added v3 format - there was no easy way to verify that all tests work with v3 format.

pinot-core/src/test/java/com/linkedin/pinot/queries/FastHllQueriesTest.java
Jackie-Jiang: This is testing queries on 4 identical segments (including broker reduce step).

pinot-core/src/test/java/com/linkedin/pinot/queries/PregeneratedHllTest.java
Jackie-Jiang: Will add this into the new fastHll test.

Jackie-Jiang:
@mayankshriv The path covered by old QueriesSentinalTest are covered by the new QueriesTest sets.

mayankshriv:
@Jackie-Jiang Thanks for clarifying.

PR #1647, opened 2017-07-11
[TE] Shade Apache hadoop jar to solve runtime error

cyenjung:
Problem:
thirdeye.org.apache.hadoop.mapreduce.Job could not be found during runtime. The reason is that we assume org.apache.hadoop is given and hence it is not shaded in thirdeye-hadoop.jar.

Fix:
Add org.apache.hadoop to thirdeye-hadoop.jar and rename it to thirdeye.org.apache.hadoop.

Test:
Local.

### DIFF
--- a/thirdeye/thirdeye-hadoop/pom.xml
+++ b/thirdeye/thirdeye-hadoop/pom.xml
@@ -45,7 +45,6 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client</artifactId>
-            <scope>provided</scope>
             <exclusions>
                 <exclusion>
                     <groupId>org.slf4j</groupId>


thirdeye/thirdeye-hadoop/pom.xml
```diff
@@ -45,7 +45,6 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client</artifactId>
-            <scope>provided</scope>
```
erwa: I think a better approach would be to keep Hadoop a provided dependency, but to NOT shade org.apache.hadoop classes in the shaded jar. See https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html for details.

cyenjung: That's a good suggestion. Let me try it out!

PR #1650, opened 2017-07-11
[TE] Improve the shading for apache hadoop

cyenjung:
The improvement to the PR: #1647 as Anthony Hsu's suggestion.

### DIFF
--- a/thirdeye/thirdeye-hadoop/pom.xml
+++ b/thirdeye/thirdeye-hadoop/pom.xml
@@ -45,6 +45,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-client</artifactId>
+            <scope>provided</scope>
             <exclusions>
                 <exclusion>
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-hadoop/pom.xml
```diff
@@ -151,6 +152,10 @@
                                 <relocation>
                                     <pattern>org.</pattern>
                                     <shadedPattern>thirdeye.org.</shadedPattern>
+                                    <excludes>
+                                        <exclude>org.apache.hadoop.**</exclude>
+                                        <exclude>org.apache.httpcomponents.**</exclude>
```
erwa: Just curious, why was this exclusion needed?

cyenjung: The scope of that package is the same as Hadoop, which is "provided". So we will get runtime error when thirdeye-hadoop runs the code from httpcomponents if we don't exclude it.

PR #1651, opened 2017-07-11
[TE] Enable anomaly function to partially update its properties

cyy0714:
- Implement an endpoint to accept post commands from front-end

- Test done offline


cyenjung:
Shouldn't this endpoint be added to AnomalyFunctionResource?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -438,6 +439,36 @@ public Response updateAnomalyFunction(@NotNull @QueryParam("id") Long id,
     return Response.ok(id).build();
   }
 
+  // Partially update anomaly function
+  @POST
```
bl44: please use PUT as the method name instead of POST, then you do not need update/properties in your path. 
Also agree with yen, this should be put in AnomalyFunctionResource where the current 'UPDATE' endpoint located.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -438,6 +439,36 @@ public Response updateAnomalyFunction(@NotNull @QueryParam("id") Long id,
...
+  @Path("/anomaly-function/update/properties")
+  public Response updateAnomalyFunctionProperties (@QueryParam("id") @NotNull Long id,
+      @QueryParam("config") @NotNull String propertiesJson) {
```
bl44: can you align the query parameter naming and input naming here (may add java doc) to avoid the trouble of readability issue.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
cyy0714: Good idea. Done the changes

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
cyy0714: Done.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyFunctionResource.java
```diff
@@ -1,57 +1,145 @@
...
+    if(id == null || anomalyFunctionDAO.findById(id) == null) {
+      String msg = "Unable to update function properties. " + id + " doesn't exist";
+      LOG.warn(msg);
+      return Response.status(Response.Status.BAD_REQUEST).entity(msg).build();
+    }
+
+    if(StringUtils.isNotBlank(config)) {
+      Map<String, String> configs = Collections.emptyMap();
+      try {
+        configs = OBJECT_MAPPER.readValue(config, Map.class);
```
puneetjaiswal: use proper request object instead of passing string and then converting that to an object.

puneetjaiswal:
use the update function end point instead of writing a new end point for partial update.

PR #1653, opened 2017-07-11
Refactor AvroRecordToPinotRowGenerator for bug-fix and cleaner logic

Jackie-Jiang:
1. Fix the issue where incomingTimeFieldSpec is not correctly created
2. Do not rely on time column name because incoming and outgoing could have the same column name
3. Simplify the logic of transform


Jackie-Jiang:
@rven1 Added a unit test.

PR #1655, opened 2017-07-11
[TE] Enable speedup for backfill

cyy0714:
- Allow the backfill endpoint to absorb multiple functionId

- Allow speed acceleration when performing backfill (Hard-Coded to 7 days)

TODO: send out alert to user when backfill is done

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
@@ -218,17 +218,110 @@ public class DetectionJobResource {
    * Regenerates anomalies for each window separately
    *
    * As the anomaly result regeneration is a heavy job, we move the function from Dashboard to worker
-   * @param ids a string containing multiple anomaly function ids, separated by comma (e.g. f1,f2,f3)
+   * @param id an anomaly function id
    * @param startTimeIso The start time of the monitoring window (in ISO Format), ex: 2016-5-23T00:00:00Z
    * @param endTimeIso The start time of the monitoring window (in ISO Format)
    * @param isForceBackfill false to resume backfill from the latest left off
-   * @param speedup whether this backfill should speedup with 7-day window
+   * @param speedup
+   *      whether this backfill should speedup with 7-day window. The assumption is that the functions are using
+   *      WoW-based algorithm, or Seasonal Data Model.
    * @return HTTP response of this request
    * @throws Exception
    */
   @POST
-  @Path("/{ids}/replay")
-  public Response generateAnomaliesInRange(@PathParam("ids") @NotNull String ids,
+  @Path("/{id}/replay")
+  public Response generateAnomaliesInRange(@PathParam("id") @NotNull final long id,
+      @QueryParam("start") @NotNull String startTimeIso,
+      @QueryParam("end") @NotNull String endTimeIso,
+      @QueryParam("force") @DefaultValue("false") String isForceBackfill,
+      @QueryParam("speedup") @DefaultValue("false") final Boolean speedup) throws Exception {
+    final boolean forceBackfill = Boolean.valueOf(isForceBackfill);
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -217,73 +218,123 @@ public Response computeSeverity(@NotNull @QueryParam("collection") String collec
    * Regenerates anomalies for each window separately
    *
    * As the anomaly result regeneration is a heavy job, we move the function from Dashboard to worker
-   * @param id anomaly function id
+   * @param ids a string containing multiple anomaly function ids, separated by comma (e.g. f1,f2,f3)
```
bl44: It does not feel right to take comma separated string as path parameter. Is there a particular reason why we want to feed f1, f2 at the same time instead of firing two calls for f1 and f2 separately?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -217,73 +218,123 @@ public Response computeSeverity(@NotNull @QueryParam("collection") String collec
...
-    AnomalyFunctionDTO anomalyFunction = anomalyFunctionDAO.findById(functionId);
-    if (anomalyFunction == null) {
+    final List<Long> functionIdList = new ArrayList<>();
+    for (String functionId : ids.split(",")) {
+      AnomalyFunctionDTO anomalyFunction = anomalyFunctionDAO.findById(Long.valueOf(functionId));
+      if (anomalyFunction != null && anomalyFunction.getIsActive()) {
+        functionIdList.add(Long.valueOf(functionId));
+      } else {
+        // Check if the anomaly function is active
+        if (!anomalyFunction.getIsActive()) {
```
bl44: should you check anomalyFunction == null here first? Also this treatment of different function ids together feels very arbitrary.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
bl44: Maybe we can use function/{id}/replay for single function,
and create another one function/replay?ids="f1,f2,f3" for multiple functions replay at a batch if you really think it is truly necessary.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -217,73 +218,123 @@ public Response computeSeverity(@NotNull @QueryParam("collection") String collec
...
+    new Thread(new Runnable() {
+      @Override
+      public void run() {
+        Map<Long, Integer> originalWindowSize = new HashMap<>();
+        Map<Long, TimeUnit> originalWindowUnit = new HashMap<>();
+        Map<Long, String> originalCron = new HashMap<>();
+        saveFunctionWindow(functionIdList, originalWindowSize, originalWindowUnit, originalCron);
+
+        // Update speed-up window and cron
+        if (speedup) {
```
bl44: This is not generic for all functions, so it does not make sense to do this type of speedup for all. 

1. please add more inline comments to state why you choose these 'speedup' configurations and when it should be used.
2. when you do speedup this way for multiple functions, you have even stronger assumption here for your input, which should be explained in java doc.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
cyy0714: The reason why I need a bulk replay is that when the replay is done, we would want to send the replay result to user. If we send an alert for each function the user creates, it spams the user. Therefore, I need a bulk replay method.

PR #1660, opened 2017-07-12
Fixing invalid enum constant in segment get API

jenniferdai:
* Also throws an illegal parameter if client puts an invalid type
* Verified that the error is thrown before code change and after by manually testing with HybridClusterIntegrationTest and calling the API


jenniferdai:
good call, all other calls in this file don't have this problem

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -195,7 +195,11 @@ private Representation getSegmentsForTable(
       ret.put(formatSegments(tableName, CommonConstants.Helix.TableType.valueOf(offline)));
       ret.put(formatSegments(tableName, CommonConstants.Helix.TableType.valueOf(realtime)));
     } else {
-      ret.put(formatSegments(tableName, CommonConstants.Helix.TableType.valueOf(type)));
+      if (!isValidTableType(type)) {
+        setStatus(Status.CLIENT_ERROR_BAD_REQUEST);
```
snleee: Can you update this error status code to `@Responses`?

jenniferdai: done

PR #1662, opened 2017-07-13
Optimize redundant adds into fixed size priority queue for topN computation.

mayankshriv:
The current code to compute topN adds elements into priority queue
unconditionally, only to de-queue immediately if queue size exceeds N.
One optimization here is to compare the new element against top of the
queue, and add only if it is better than the top value. Doing so
provides about 10% latency improvement for cases with large number of
groups (100k groups or more, before pruning).

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
@@ -115,10 +115,8 @@ public class AggregationGroupByTrimmingService {
           if (priorityQueue == null) {
             trimmedResults.get(i).put(groupKey, intermediateResults[i]);
           } else {
-            priorityQueue.add(new GroupKeyResultPair(groupKey, (Comparable) intermediateResults[i]));
-            if (priorityQueue.size() > _trimSize) {
-              priorityQueue.poll();
-            }
+            GroupKeyResultPair newValue = new GroupKeyResultPair(groupKey, (Comparable) intermediateResults[i]);
+            addToPriorityQueue(priorityQueue, newValue, _trimSize);
           }
         }
       }
@@ -171,10 +169,9 @@ public class AggregationGroupByTrimmingService {
       for (Map.Entry<String, Comparable> entry : finalResultMap.entrySet()) {
         String groupKey = entry.getKey();
         Comparable finalResult = entry.getValue();
-        priorityQueue.add(new GroupKeyResultPair(groupKey, finalResult));
-        if (priorityQueue.size() > _groupByTopN) {
-          priorityQueue.poll();
-        }
+
+        GroupKeyResultPair newValue = new GroupKeyResultPair(groupKey, finalResult);
+        addToPriorityQueue(priorityQueue, newValue, _groupByTopN);
       }
 
       // Fill trimmed results into the list.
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
```diff
@@ -115,9 +115,15 @@ public AggregationGroupByTrimmingService(@Nonnull AggregationFunction[] aggregat
           if (priorityQueue == null) {
             trimmedResults.get(i).put(groupKey, intermediateResults[i]);
           } else {
-            priorityQueue.add(new GroupKeyResultPair(groupKey, (Comparable) intermediateResults[i]));
-            if (priorityQueue.size() > _trimSize) {
-              priorityQueue.poll();
+            GroupKeyResultPair newValue = new GroupKeyResultPair(groupKey, (Comparable) intermediateResults[i]);
+            if (priorityQueue.size() >= _trimSize) {
```
Jackie-Jiang: "=="?

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
```diff
@@ -171,9 +177,16 @@ public AggregationGroupByTrimmingService(@Nonnull AggregationFunction[] aggregat
       for (Map.Entry<String, Comparable> entry : finalResultMap.entrySet()) {
         String groupKey = entry.getKey();
         Comparable finalResult = entry.getValue();
-        priorityQueue.add(new GroupKeyResultPair(groupKey, finalResult));
-        if (priorityQueue.size() > _groupByTopN) {
-          priorityQueue.poll();
+
+        if (priorityQueue.size() >= _groupByTopN) {
```
Jackie-Jiang: Same here

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
```diff
@@ -171,9 +177,16 @@ public AggregationGroupByTrimmingService(@Nonnull AggregationFunction[] aggregat
...
+          GroupKeyResultPair topValue = priorityQueue.peek();
+          GroupKeyResultPair newValue = new GroupKeyResultPair(groupKey, finalResult);
+          if (priorityQueue.comparator().compare(topValue, newValue) < 0) {
+            priorityQueue.poll();
+            priorityQueue.add(new GroupKeyResultPair(groupKey, finalResult));
```
Jackie-Jiang: newValue

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
```diff
@@ -171,9 +177,16 @@ public AggregationGroupByTrimmingService(@Nonnull AggregationFunction[] aggregat
...
+          }
+        } else {
+          priorityQueue.add(new GroupKeyResultPair(groupKey, finalResult));
```
Jackie-Jiang: newValue

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
Jackie-Jiang: Can _trimSize be 0? If yes, then we need to first add then pool

mayankshriv: _trimSize is set to a min of 5000 in the current code. Nonetheless, I have added a null check against topValue (will happen when the queue is empty).

mayankshriv: I feel >= is more resilient to future changes.

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
mayankshriv: Yeah, same response :-)

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
mayankshriv: Addressed

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
mayankshriv: Addressed

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
```diff
@@ -171,9 +177,16 @@ public AggregationGroupByTrimmingService(@Nonnull AggregationFunction[] aggregat
       for (Map.Entry<String, Comparable> entry : finalResultMap.entrySet()) {
         String groupKey = entry.getKey();
         Comparable finalResult = entry.getValue();
-        priorityQueue.add(new GroupKeyResultPair(groupKey, finalResult));
-        if (priorityQueue.size() > _groupByTopN) {
-          priorityQueue.poll();
+
+        GroupKeyResultPair newValue = new GroupKeyResultPair(groupKey, finalResult);
+        if (priorityQueue.size() >= _groupByTopN) {
```
snleee: This code looks the same as the code above. We can probably take out these as a separate function to reuse the code. i.e. `addToPriorityQueue(priorityQueue, maxSize)`

PR #1667, opened 2017-07-14
Avoid expensive String.format for non-double values.

mayankshriv:
For group-by responses with > 20k groups it seems upto 25% of the query
latency is just spent formatting doubles while generating response.
This can be avoided in cases where we know that the result is a whole
number (especially when column data type was short/int/long).

A cleaner change would be for servers to send response in appropriate
data type, and broker to upcast to the most appropriat data type. But
that needs quite a bit of changes. This PR is a band-aid until we have a
clean fix.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/AggregationFunctionUtils.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/AggregationFunctionUtils.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.core.query.aggregation.function;
 
+import com.google.common.math.DoubleMath;
 import com.linkedin.pinot.common.request.AggregationInfo;
 import com.linkedin.pinot.common.segment.SegmentMetadata;
 import com.linkedin.pinot.core.plan.AggregationFunctionInitializer;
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/AggregationFunctionUtils.java
```diff
@@ -65,7 +65,15 @@ private AggregationFunctionUtils() {
   @Nonnull
   public static String formatValue(@Nonnull Object value) {
     if (value instanceof Double) {
-      return String.format(Locale.US, "%1.5f", (Double) value);
+      Double doubleValue = (Double) value;
+
+      // String.format is very expensive, so avoid it for whole numbers that can fit in Long.
+      // We simply append ".00000" to long, in order to keep the existing behavior.
+      if (doubleValue < Long.MAX_VALUE && !Double.isInfinite(doubleValue) && doubleValue == Math.floor(doubleValue)) {
```
Jackie-Jiang: Use DoubleMath.isMathematicalInteger(value) instead, which should be faster.
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/math/DoubleMath.html#isMathematicalInteger(double)

mayankshriv: Updated with change.

PR #1672, opened 2017-07-17
Fix the bug where deleting HLC realtime table throws ZK exception

Jackie-Jiang:
When we delete a HLC table, we first remove segments from property store, then drop the table
When removing segments, after we remove the IN_PROGRESS one, controller will automatically assign a new one because table is still alive, which cause ZK failed to remove the parent node.

The fix is that:
Whenever we drop a table (OFFLINE/REALTIME), we should first drop the table from Helix, then clean up the state in property store or disk


jfim:
Can you make sure that realtime segments are dropped from the realtime servers when the table is deleted?

Jackie-Jiang:
@jfim Tested locally, segments get dropped.

PR #1678, opened 2017-07-18
Change log level for transient LLC exceptions

jfim:
Change the log level for transient LLC exceptions to info, so that they
don't hide more important log messages.


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -279,7 +279,7 @@ private void handleTransientKafkaErrors(Exception e) throws  Exception {
       segmentLogger.warn("Kafka transient exception when fetching messages, stopping consumption after {} attempts", consecutiveErrorCount, e);
       throw e;
     } else {
-      segmentLogger.warn("Kafka transient exception when fetching messages, retrying (count={})", consecutiveErrorCount, e);
+      segmentLogger.info("Kafka transient exception when fetching messages, retrying (count={})", consecutiveErrorCount, e);
```
mcvsubbu: this should be a warning, no?

WARN when an problem happens, INFO when that problem is corrected, and ERROR if it is never corrected.

npawar:

PR #1684, opened 2017-07-19
On-Heap implementation of immutable StringDictionary.

mayankshriv:
The call to StringDictionary.get()/indexOf() is expensive because it has
to pad/unpad the string, create new String from byte[], perform binary
search etc. This is done in the order of numDocsScanned, per query.
It can take > 30% of server side latency if numDocsScanned is of the
order of several tens of thousands or more (99th percentile).

For String columns with very low cardinality, we can simply build an
on-heap dictionary using array + map at the time of loading segment.
This gets rid of the padding/unpadding, binary search, as well as
eliminates all temporary String creation.

Added field in IndexingConfig to specify (String) columns for which
on-heap dictionary should be loaded.

Added unit tests for the new code.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
@@ -39,6 +39,7 @@ public class IndexingConfig {
   private String _starTreeFormat;
   private String _columnMinMaxValueGeneratorMode;
   private List<String> _noDictionaryColumns;
+  private List<String> _onHeapDictionaryColumns;
   private StarTreeIndexSpec _starTreeIndexSpec;
   private SegmentPartitionConfig _segmentPartitionConfig;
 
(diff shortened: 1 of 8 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
```diff
@@ -39,6 +39,7 @@
   private String _starTreeFormat;
   private String _columnMinMaxValueGeneratorMode;
   private List<String> _noDictionaryColumns;
+  List<String> _onHeapDictionaryColumns;
```
snleee: private?

pinot-common/src/test/java/com/linkedin/pinot/common/config/IndexingConfigTest.java
```diff
@@ -35,11 +35,14 @@
 public class IndexingConfigTest {
 
   @Test
-  public void testIgnoreUnknown()
+  public void test()
```
snleee: I think it's better to put the name of the test to what it is actually testing.

mayankshriv: It was no longer testing 'unknown' fields, and it was the only test specifically for indexing config, so I named it 'test'. Have renamed it to 'testSerDe' as it is testing ser/de of indexing config.

mayankshriv:
@snlee Config changes are not automatically picked up by server, and require server restart.

PR #1697, opened 2017-07-21
Fix FileNotFoundException when uploading multiple segments with same name

Jackie-Jiang:
The reason for the exception is as followings:
1. Two POST requests trying to upload segment with the same name
2. Before one thread try to move the file, the other one delete the file

The fix is that, always use unique path for each uploaded segment.


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/restlet/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -17,6 +17,7 @@
 
 import com.alibaba.fastjson.JSONObject;
 import com.fasterxml.jackson.core.JsonProcessingException;
```
Jackie-Jiang: That requires a full clean up because fasterxml and codehaus jackson are not compatible with each other...

PR #1700, opened 2017-07-24
Fix the flakiness of OfflineClusterIntegrationTest

Jackie-Jiang:
With multiple servers, there is no way to check and guarantee that all servers get all segments reloaded, which cause the flakiness of the tests.
To fix that, in OfflineClusterIntegrationTest, only start one server.
Added a MultiNodesOfflineClusterIntegrationTest to keep the same coverage for multiple brokers and multiple servers.

### DIFF
--- a/pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTestSet.java
+++ b/pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTestSet.java
@@ -40,6 +40,7 @@ import org.testng.Assert;
  * Shared set of common tests for cluster integration tests.
  * <p>To enable the test, override it and add @Test annotation.
  */
+@SuppressWarnings("unused")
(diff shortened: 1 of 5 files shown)


pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/BaseClusterIntegrationTestSet.java
```diff
@@ -40,6 +40,7 @@
  * Shared set of common tests for cluster integration tests.
  * <p>To enable the test, override it and add @Test annotation.
  */
+@SuppressWarnings("unused")
```
mayankshriv: Why unused?

Jackie-Jiang: We keep some tests that covered by other tests, but if necessary, we can enable the unused tests for debugging purpose.

Jackie-Jiang:
@mayankshriv To fix the flakiness, we have to use different settings for cluster. The expensive tests are running only in multi-nodes one, but we have to pay the extra cost for setting up.

PR #1705, opened 2017-07-25
Improve logging to identify failures during LLC segment commit

mcvsubbu:
LLC segment commit goes through multiple stages, and it is useful to know which stage failed.
Added logs at each state. Also, the existing log was printing object address, so needed a toJsonString() there.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -634,13 +634,14 @@ public class LLRealtimeSegmentDataManager extends SegmentDataManager {
     }
 
     SegmentCompletionProtocol.Response segmentCommitUploadResponse = _protocolHandler.segmentCommitUpload(
-        _currentOffset, _segmentNameStr, segmentTarFile, prevResponse.getSegmentLocation(), prevResponse.getControllerVipUrl());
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -626,19 +626,26 @@ protected String buildSegmentInternal(boolean forCommit) {
...
+  protected SegmentCompletionProtocol.Response doSplitCommit(File segmentTarFile, SegmentCompletionProtocol.Response prevResponse) {
     SegmentCompletionProtocol.Response segmentCommitStartResponse = _protocolHandler.segmentCommitStart(_currentOffset, _segmentNameStr);
     if (!segmentCommitStartResponse.getStatus().equals(SegmentCompletionProtocol.ControllerResponseStatus.COMMIT_CONTINUE)) {
+      segmentLogger.warn("CommitStart failed  with response {}", segmentCommitStartResponse.toJsonString());
       return SegmentCompletionProtocol.RESP_FAILED;
     }
 
     SegmentCompletionProtocol.Response segmentCommitUploadResponse = _protocolHandler.segmentCommitUpload(
-        _currentOffset, _segmentNameStr, segmentTarFile, response.getSegmentLocation(), response.getControllerVipUrl());
+        _currentOffset, _segmentNameStr, segmentTarFile, prevResponse.getSegmentLocation(), prevResponse.getControllerVipUrl());
```
jenniferdai: I don't think we need to have segment location here as it is created in the upload portion.

PR #1708, opened 2017-07-26
Set segment version to v3

jenniferdai:


Jackie-Jiang:
Is PBNJ directly generate v3 format segment and push it to controller? That will cause a lot of problem because a lot of code assume that segments on controller are in v1 format.

PR #1710, opened 2017-07-27
[Minion] Add Minion use case ConvertToRawIndexTask

Jackie-Jiang:
1. This task converts indexes for configured columns / (by default) metric columns that meet the condition (uncompressed raw index size <= CONVERSION_THRESHOLD * dictionary-based index size) into raw indexes (no dictionary indexes)
2. Add a list field 'optimizations' into segmentMetadata and segmentZKMetadata to mark the optimizations done (can be set by Minion or segment creation job)
3. Allow custom configs for each task type in TableTaskConfig
4. Add a new ConvertToRawIndexMinionClusterIntegrationTest to test the new use case

NOTE: In Helix Task framework, TaskQueue resource will only show up in IS but not in EV

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/minion/RawIndexConverter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/minion/RawIndexConverter.java
@@ -1,0 +1,200 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.core.minion;
+
+import com.google.common.base.Preconditions;
+import com.linkedin.pinot.common.data.FieldSpec;
+import com.linkedin.pinot.common.data.MetricFieldSpec;
+import com.linkedin.pinot.common.data.Schema;
+import com.linkedin.pinot.common.segment.ReadMode;
+import com.linkedin.pinot.core.common.BlockSingleValIterator;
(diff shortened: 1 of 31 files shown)


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
```diff
@@ -214,20 +217,34 @@ public void addOrReplaceOfflineSegment(String tableName, String segmentId, boole
...
+      LOGGER.info("Existed segment metadata is null for segment: {} in table: {}", segmentName, offlineTableName);
       return true;
     }
-    LOGGER.info("segmentMetadataForCheck.crc={},segmentMetadataFromServer.crc={} for {} of table {}",
-        segmentMetadataForCheck.getCrc(), segmentMetadataFromServer.getCrc(), segmentName, tableName);
-    if ((!segmentMetadataFromServer.getCrc().equalsIgnoreCase("null"))
-        && (segmentMetadataFromServer.getCrc().equals(segmentMetadataForCheck.getCrc()))) {
-      return false;
+
+    long newCrc = newSegmentZKMetadata.getCrc();
```
snleee: I think it's good to leave `TODO` or a comment to indicate that you try to keep the same logic as before. I can see we are not checking creation time here while the controller side checks it. Eventually, this part can even be removed as we discussed offline.

pinot-core/src/main/java/com/linkedin/pinot/core/minion/RawIndexConverter.java
```diff
@@ -0,0 +1,194 @@
...
+        int dictId = iterator.nextIntVal();
+        rawIndexCreator.index(docId++, dictionary.get(dictId));
+      }
+    }
+
+    // Update the segment metadata
+    _convertedProperties.setProperty(
+        V1Constants.MetadataKeys.Column.getKeyFor(columnName, V1Constants.MetadataKeys.Column.HAS_DICTIONARY), false);
+    _convertedProperties.setProperty(
+        V1Constants.MetadataKeys.Column.getKeyFor(columnName, V1Constants.MetadataKeys.Column.BITS_PER_ELEMENT), -1);
```
Jackie-Jiang: Added

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
Jackie-Jiang: Added

PR #1722, opened 2017-07-31
TE Create Alert Bug Fix

newsummit:
Fixing Create Alert bug that recognizes blank space as input for new alert config group name and releases submit button. Fixed by using "isBlank" instead of "isNone" helper, and trimming the input before it is sent to API.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
@@ -385,16 +385,16 @@ export default Ember.Controller.extend({
     'newConfigGroupName',
     'alertGroupNewRecipient',
     function() {
-      const requiredFields = this.get('requiredFields');
       let isDisabled = false;
+      const requiredFields = this.get('requiredFields');
       // Any missing required field values?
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
```diff
@@ -416,7 +416,7 @@ export default Ember.Controller.extend({
       const granularity = this.get('graphConfig.granularity').toLowerCase();
       const settingsByGranularity = {
         common: {
-          functionName: this.get('alertFunctionName'),
+          functionName: this.get('alertFunctionName').trim(),
```
justYves: Can this ever be `null` or `undefined` ? This might error out if that's the case

justYves: perhaps use the same logic as below `this.get('alertFunctionName') || this.get('alertFunctionName').trim()`

newsummit: won't hurt to add this. thanks.

PR #1735, opened 2017-08-02
In integration tests, when stopping Server/Minion instance, delete temp dir

Jackie-Jiang:
Without cleaning up the temp dir, Server will directly load local files.

### DIFF
--- a/pinot-minion/src/main/java/com/linkedin/pinot/minion/MinionStarter.java
+++ b/pinot-minion/src/main/java/com/linkedin/pinot/minion/MinionStarter.java
@@ -55,8 +55,7 @@ public class MinionStarter {
 
   private HelixAdmin _helixAdmin;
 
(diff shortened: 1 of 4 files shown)


pinot-minion/src/main/java/com/linkedin/pinot/minion/MinionStarter.java
```diff
@@ -55,8 +55,7 @@
 
   private HelixAdmin _helixAdmin;
 
-  public MinionStarter(String zkAddress, String helixClusterName, Configuration config)
-      throws Exception {
+  public MinionStarter(String zkAddress, String helixClusterName, Configuration config) throws Exception {
```
mayankshriv: Not a blocker for this PR, but we need to address the code formatting change due to IDE upgrade.

PR #1738, opened 2017-08-02
[TE] ignore dimensions when relative contribution did not change

apucher:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/DimensionAnalysisPipeline.java
```diff
@@ -120,6 +120,9 @@ public PipelineResult run(PipelineContext context) {
 
         for(int i=0; i<dfScores.size(); i++) {
           double score = dfScores.getDouble(COL_SCORE, i);
+          if (score <= 0)
```
cyenjung: Please add braces to this if-statement and the one at line 128 to prevent this bug in the future: https://blog.codecentric.de/en/2014/02/curly-braces/

PR #1740, opened 2017-08-02
Filter query tree serialization bugfix

jfim:
Fix the serialization of the filter query tree. During filter query tree
serialization, we convert the tree into a map of ids to individual query
predicates. Since the id is generated automatically, if there is a
collision on the id on the leaf nodes, individual query predicates can
be overwritten, causing incorrect query processing.

This patch fixes this issue by removing the individually generated node
ids and generates them sequentially at serialization time, ensuring that
there is no possible collisions, assuming that there are fewer than
$2^31-1$ query filter predicates.


jfim:
I tested this with a slightly modified version of the code that would generate the same id numbers for leaf nodes and confirmed that this bugfix does pass the integration tests even if there are id collisions (and fails if there are id collisions without the bugfix).

PR #1748, opened 2017-08-03
[TE] change default window delay setting and frenquency setting for anomaly detection creation endpoint

bl44:
[TE] change default window delay setting and frenquency setting for anomaly detection creation endpoint

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -331,15 +331,17 @@ public class AnomalyResource {
       break;
     case HOURS:
     default:
-      windowDelayTime = 2;
+      windowDelayTime = 0;
       windowDelayTimeUnit = TimeUnit.HOURS;
     }
     anomalyFunctionSpec.setWindowDelayUnit(windowDelayTimeUnit);
     anomalyFunctionSpec.setWindowDelay(windowDelayTime);
 
     // setup detection frequency if it's minutely function
+    // the default frequency is determined by data granularity
+    // this frequency is a override to lower the system overload
     if (dataGranularity.getUnit().equals(TimeUnit.MINUTES)) {
-      TimeGranularity frequency = new TimeGranularity(30, TimeUnit.MINUTES);
+      TimeGranularity frequency = new TimeGranularity(15, TimeUnit.MINUTES);
       anomalyFunctionSpec.setFrequency(frequency);
     }
 
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -280,7 +280,7 @@ public class AnomalyResource {
       @QueryParam("type") String type,
       @NotNull @QueryParam("windowSize") String windowSize,
       @NotNull @QueryParam("windowUnit") String windowUnit,
-      @QueryParam("windowDelay") String windowDelay,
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -331,15 +331,17 @@ public Response createAnomalyFunction(@NotNull @QueryParam("dataset") String dat
...
       windowDelayTimeUnit = TimeUnit.HOURS;
     }
     anomalyFunctionSpec.setWindowDelayUnit(windowDelayTimeUnit);
     anomalyFunctionSpec.setWindowDelay(windowDelayTime);
 
     // setup detection frequency if it's minutely function
+    // the default frequency is determined by data granularity
+    // this frequency is a override to lower the system overload
     if (dataGranularity.getUnit().equals(TimeUnit.MINUTES)) {
-      TimeGranularity frequency = new TimeGranularity(30, TimeUnit.MINUTES);
```
cecilynie: Do we have a convention what is the shortest frequency for minutes? I recall that ingraph is being refreshed 30mins. For sitespeed @cyenjung manually adjusted to 15 mins for quicker alerting.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -331,15 +331,17 @@ public Response createAnomalyFunction(@NotNull @QueryParam("dataset") String dat
       break;
     case HOURS:
     default:
-      windowDelayTime = 2;
```
cecilynie: Please also adjust the comment on line 320 accordingly. (Probably why changing the default time from 10 hours to 0?)

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
bl44: The previous setting is artificial limitation posted by old ingraph adaptor used. Which always have d15 minutes delay and we do not want to have the most recent one delayed and set its to 30 minutes to keep it safe.

Sitespeed data is in pinot, not from ingraph. Ingraph metric can be treated as real time for now.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
bl44: Sure. Need to change it.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
cecilynie: Is inGraph itself being refreshed every 30mins? Though we do real time data retrieval, but inGraph is not real time?..This is an impression I got when talking to Ads POC, correct me if I'm wrong.

bl44: in graph is real time. All data is streaming to ingraph database through kafka. If there is delay (more than say 5 minutes, it will be a big production issue). 

On the other side, our sign test method do not sensitive to missing data. Considering we are doing 6 hour moving window. We can tolerate the delay. This frequency change from 30 minutes to 15 minutes is not significant: that 6 new data points vs 3 data points.  But if we change granularity to 1 minutes in future. 15 minutes means 15 new observations -- this can really move the needle. And we may consider further push this to 5 minutues, if we really targeting for MTTD for some special use cases.

bl44:
This frequency is only used by DetectionJobSechulerUtils for now.

PR #1752, opened 2017-08-04
Remove the default segment version, allow loading segment as is without segment version configured

Jackie-Jiang:
Also clean up the code so that after conversion we don't depend on segment version to load segment, i.e. segment load only depends on the segment itself
When loading a segment, all method should take the top level index directory as the argument, and we can get the segment version from it's metadata

This fixed the problem that PerfBenchMarkDriver accidentally delete local files


pinot-core/src/test/java/com/linkedin/pinot/core/startree/TestStarTreeConverter.java
```diff
@@ -15,7 +15,6 @@
  */
```
snleee: Rename test?

Jackie-Jiang: Done~

PR #1753, opened 2017-08-04
Optimize sorted index

jfim:
Optimize the roaring bitmap generation for sorted indexes by adding
ranges directly instead of adding docIds one by one.

Performance impact of adding 1000 values is about 133x faster and is likely to have a large impact :
```
Benchmark                                             Mode      Cnt         Score    Error  Units
BenchmarkRoaringBitmap.addFixed                     sample  3956823       128.876 ±  9.871  ns/op
BenchmarkRoaringBitmap.addInLoop                    sample  3520867     17109.083 ± 33.184  ns/op
```

For one million values, it's over 100000x faster (900 nanoseconds vs 12.5 milliseconds):
```
Benchmark                                             Mode      Cnt         Score       Error  Units
BenchmarkRoaringBitmap.addFixed                     sample  4475444       972.145 ±    16.770  ns/op
BenchmarkRoaringBitmap.addInLoop                    sample     9463  12688404.668 ± 55250.599  ns/op
```

For single value columns (eg. days since epoch), they count as a single value sorted column, so this should have a large impact for queries that have those.


jfim:
Actually, I think we don't even use the inverted index for sorted columns since we have a specialized operator for them. If we ever use the inverted index for it, then it should be faster. :grin:

PR #1756, opened 2017-08-04
[TE] Self-Serve Create Alert Pre-Launch Modifications (table layout)

newsummit:
Changes made to Crete Alert view:

- Changing up replay triggering sequence… not waiting for jobId since we’re not using it yet
- Enhancing readability of “Alerts Monitored by This Group” with a table layout
- Styles - overriding bootstrap and power-select for one class (making power-select fields match “fatness” of bootstrap inputs)

Changes in commit 5dd360a...
- Removed "clone" call and not waiting for response from "replay" call, since the flow currently ends there.
- Changed handling of error messages on replay call (using the same banner with different states)
- Handled a small bug that was introducing trailing `,` chars to recipients list.
- Added ID column to alerts table, and made sure newly created alerts float to the top of the list for visual confirmation.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
@@ -25,12 +25,14 @@ export default Ember.Controller.extend({
   isCreateGroupSuccess: false,
   isCreateAlertError: false,
   isSelectMetricError: false,
-  isReplayError: false,
+  isReplayStatusError: false,
   isMetricDataLoading: false,
   isReplayStatusPending: true,
-  isReplaySuccess: false,
+  isReplayStatusSuccess: false,
+  isReplayStarted: false,
   metricGranularityOptions: [],
   originalDimensions: [],
+  bsAlertBannerType: 'success',
   graphEmailLinkProps: '',
   replayStatusClass: 'te-form__banner--pending',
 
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
```diff
@@ -301,17 +319,28 @@ export default Ember.Controller.extend({
...
+      isReplaySuccess: true,
+      isReplayStatusPending: true
+    });
+
+    // Begin replay sequence. Simulate trigger response time since we don't yet need
+    // to wait for the actual response, which takes 30+ seconds
     this.callCloneAlert(newFuncId)
       .then((clonedId) => {
+        const that = this;
+        Ember.run.later((function() {
```
justYves: this will probably be run after `this.callReplayStart`. Just want to make sure that's what we want

thirdeye/thirdeye-frontend/app/styles/wrapper/styles.scss
```diff
@@ -188,6 +186,24 @@ body {
...
+  .panel-title {
+    font-size: 15px;
+    color: #369ece;
+    font-weight: 600;
+    letter-spacing: .1px;
+  }
+
+  // overriding power-select basic field style
+  .ember-power-select-trigger {
+    padding-top: 2px;
```
justYves: I believe I already made a change to make the power select input consistent with our existing `input`. 
This might not be needed.

thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
newsummit: Yes, that was intentional, but I'm modifying it after today's meeting (replay sequence is now being modified)

PR #1759, opened 2017-08-04
Improve logs during realtime segment commit in the controller and server.

mcvsubbu:
Changed the controller to log the entire exception while committing the segment file.

Changed the server to log the controller host and version during split-commit upload phase,
since the segment may go to any controller host.

Re-factored the http post during split commit so that the code path is different from that
of regular get/post to the leader controller.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
@@ -23,6 +23,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import org.apache.commons.httpclient.Header;
 import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.HttpMethodBase;
 import org.apache.commons.httpclient.URI;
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
```diff
@@ -131,26 +132,67 @@ public InputStream createInputStream() throws IOException {
...
   }
 
-  private SegmentCompletionProtocol.Response doHttp(SegmentCompletionProtocol.Request request, Part[] parts, String uploadHost) {
-    ControllerLeaderLocator leaderLocator = ControllerLeaderLocator.getInstance();
+  // Much as it is tempting to merge postSegmentToVip() and doHttp() into one method, it is probably better to
+  // keep them separate. One posts to the vip, whereas the other posts (or gets) from the current leader controller.
+
+  // Post a segment to the VIP (used during split commit).
+  // Note that the VipURL may either be http or https (depending on controller configuration).
+  private SegmentCompletionProtocol.Response postSegmentToVip(SegmentCompletionProtocol.Request request, Part[] parts,
```
jenniferdai: do you have to change the calling method to call this?

pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
```diff
@@ -131,26 +132,67 @@ public InputStream createInputStream() throws IOException {
...
+    } catch (IOException e) {
+      LOGGER.error("IOException {}", this.toString(), e);
+      return response;
     }
+  }
 
+  // Do GET or POST operations on the leader controller. Since we are communicating directly with the leader controller,
+  // we must handle NOT_LEADER errors.
+  private SegmentCompletionProtocol.Response doHttp(SegmentCompletionProtocol.Request request, Part[] parts) {
+    ControllerLeaderLocator leaderLocator = ControllerLeaderLocator.getInstance();
```
jenniferdai: do we want the previous error message for NOT_LEADER?

PR #1767, opened 2017-08-09
Enhance AggregationGroupByTrimmingService to support trimming on non-comparable intermediate result

Jackie-Jiang:
Currently the following aggregation functions have non-comparable intermediate result:
DistinctCount, DistinctCountHLL, FastHLL, Percentile, PercentileEst
Current trimming service does not work on them, which might cause huge amount of data serialized and transfered to brokers

Implemented a TreeMap based class to support trimming on non-comparable intermediate result
Added a helper Sorter interface and two implementations: ComparableSorter and NonComparableSorter
Also enhanced the test to cover the new code

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
@@ -15,66 +15,47 @@
  */
 package com.linkedin.pinot.core.query.aggregation.groupby;
 
+import com.google.common.base.Preconditions;
 import com.linkedin.pinot.common.response.broker.GroupByResult;
-import com.linkedin.pinot.core.query.aggregation.AggregationFunctionContext;
 import com.linkedin.pinot.core.query.aggregation.function.AggregationFunction;
-import com.linkedin.pinot.core.query.aggregation.function.AggregationFunctionFactory;
 import com.linkedin.pinot.core.query.aggregation.function.AggregationFunctionUtils;
+import com.linkedin.pinot.core.query.aggregation.function.MinAggregationFunction;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.PriorityQueue;
+import java.util.TreeMap;
 import javax.annotation.Nonnull;
+import org.apache.commons.collections.comparators.ComparableComparator;
+import org.apache.commons.lang3.tuple.ImmutablePair;
 
 
 /**
- * The <code>AggregationGroupByTrimmingService</code> class provides trimming service for aggregation group-by query.
+ * The <code>AggregationGroupByTrimmingService</code> class provides trimming service for aggregation group-by queries.
  */
-// TODO: revisit the trim algorithm, implement trim on all Object but not only for Comparable.
 public class AggregationGroupByTrimmingService {
   public static final String GROUP_KEY_DELIMITER = "\t";
 
-  private final int _numAggregationFunctions;
-  private final boolean[] _minOrders;
-
+  private final AggregationFunction[] _aggregationFunctions;
   private final int _groupByTopN;
-  // To keep the precision, _trimSize is the larger of (_groupByTopN * 5) or 5000.
   private final int _trimSize;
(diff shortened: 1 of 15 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
```diff
@@ -93,46 +85,75 @@ public AggregationGroupByTrimmingService(@Nonnull AggregationFunction[] aggregat
...
-      // Need to trim.
-
-      // Construct the priority queues.
-      @SuppressWarnings("unchecked")
-      PriorityQueue<GroupKeyResultPair>[] priorityQueues = new PriorityQueue[_numAggregationFunctions];
-      Object[] sampleResults = intermediateResultsMap.values().iterator().next();
+      // Construct the sorters
+      // For comparable intermediate result, use PriorityQueue as the sorter
+      // For incomparable intermediate result, use TreeMapBasedSorter as the sorter
+      Object[] sorters = new Object[_numAggregationFunctions];
```
mayankshriv: May be sorter should be an interface, with PQ based and TreeMap based as its implementations?

Jackie-Jiang: We use very different way to sort Comparable and non-Comparable, and it's just a private helper class. Adding an interface seems an overkill, and might hurt performance.

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
```diff
@@ -93,46 +85,75 @@ public AggregationGroupByTrimmingService(@Nonnull AggregationFunction[] aggregat
...
-          if (priorityQueue == null) {
-            trimmedResults.get(i).put(groupKey, intermediateResults[i]);
+          Object intermediateResult = intermediateResults[i];
+          if (_aggregationFunctions[i].isIntermediateResultComparable()) {
+            GroupKeyResultPair value = new GroupKeyResultPair(groupKey, (Comparable) intermediateResult);
+            addToPriorityQueue((PriorityQueue<GroupKeyResultPair>) sorters[i], value, _trimSize);
           } else {
-            GroupKeyResultPair newValue = new GroupKeyResultPair(groupKey, (Comparable) intermediateResults[i]);
-            addToPriorityQueue(priorityQueue, newValue, _trimSize);
+            Comparable finalResult = _aggregationFunctions[i].extractFinalResult(intermediateResult);
```
mayankshriv: Is this adding additional call to extractFinalResults? If so, that call seems to be expensive for some aggregation functions (eg percentile), where there is sorting involved.

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
mayankshriv: But the interface should remain the same. Are you saying that overhead of using interface here would be a non-ignorable impact on latency? If so, how?

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
Jackie-Jiang: You we make intermediate result comparable, then for each comparison (log2m times per insertion or removal from heap), we need to calculate final result once. Implementing this way to calculate final result only once and cache all of them.

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
Jackie-Jiang: It's not about overhead. I will try to make them an interface, but that basically means refactoring most of the code... Still, as a private helper class, I don't think that's necessary.

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/AggregationGroupByTrimmingService.java
Jackie-Jiang: Okay, maybe a simpler implementation is just cache all the final results and still use Heap.

PR #1768, opened 2017-08-09
[TE] Self-serve Create Alert Bug Fix

newsummit:
Bug fix for Create Alert call sequence : for new alert groups, the name property was not properly set. Also: 
- Now deleting the new alert function created if any error occurs with config group creation (tested by sending unexpected data format to create endpoint)
- Now catching errors more efficiently
- URL encoding part of the function creation payload

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
@@ -124,6 +124,19 @@ export default Ember.Controller.extend({
   },
 
   /**
+   * Sets error message to alert create error state and disables form
+   * @method setAlertCreateErrorState
+   * @returns {undefined}
+   */
+  setAlertCreateErrorState(error) {
+    // TODO: Log alert creation errors with Piwik
+    this.setProperties({
+      isCreateAlertError: true,
+      isFormDisabled: true
+    });
+  },
+
+  /**
    * Determines if a metric should be filtered out
    * @method isMetricGraphable
    * @param {Object} metric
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
```diff
@@ -863,7 +888,10 @@ export default Ember.Controller.extend({
       let functionArray = [];
       let that = this;
```
justYves: I know this is not new, just wondering if this is actually being used anywhere

thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
```diff
@@ -877,45 +905,43 @@ export default Ember.Controller.extend({
...
+          // Confirm group creation if not in group edit mode
+          if (!isEditGroupMode) {
+            this.set('isCreateGroupSuccess', true);
+          }
+          // Display function added to group confirmation
+          this.prepareFunctions(finalConfigObj, newFunctionId).then(functionData => {
+            this.set('selectedGroupFunctions', functionData);
+          });
+          // Trigger alert replay.
+          this.triggerReplay(newFunctionId);
```
newsummit: Note: at this point, we don't do anything with the replay response.

thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
newsummit: Nice catch - thanks. Not needed any more.

PR #1771, opened 2017-08-09
[TE] multi-keyword metric search/autocomplete

apucher:
Currently, autocomplete only supports a single, continuous keyword (including spaces). This change enables ThirdEye to search for metric names via multiple keywords - e.g. "metric my" now matches "my_dataset::metric_a"


cyenjung:
Very useful function! Thanks for implementing this.

PR #1772, opened 2017-08-09
Fix the exception from HelixHelper.getEnabledInstancesWithTag() when instance config does not exist

Jackie-Jiang:
When cannot find instance config, Helix throw an exception instead of returning null
Catch the exception and log an error, then continue to check the next instance

This exception might happen when instance are not correctly dropped.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
@@ -393,17 +393,18 @@ public class HelixHelper {
     updateIdealState(helixManager, tableName, updater, DEFAULT_RETRY_POLICY);
   }
 
-  public static List<String> getEnabledInstancesWithTag(HelixAdmin helixAdmin, String helixClusterName, String instanceTag) {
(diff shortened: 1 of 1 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
```diff
@@ -393,17 +393,18 @@ public IdealState apply(IdealState idealState) {
...
     for (String instance : instances) {
-      InstanceConfig instanceConfig = helixAdmin.getInstanceConfig(helixClusterName, instance);
-      if (instanceConfig == null) {
-        LOGGER.warn("InstanceConfig not found for instance: {}", instance);
-        continue;
-      }
-      if (instanceConfig.containsTag(instanceTag) && instanceConfig.getInstanceEnabled()) {
-        enabledInstances.add(instance);
+      try {
+        InstanceConfig instanceConfig = helixAdmin.getInstanceConfig(helixClusterName, instance);
```
mcvsubbu: the null pointer was logged as a warn before, will now be an error. Do we want that? I would add the null check back.

Jackie-Jiang: It can never be null. Helix will directly throw exception if instance does not exist.

PR #1774, opened 2017-08-09
Compatibility patch for older http clients, a bug fix in segment state toggle, logging

mcvsubbu:
Some http clients do not send a leading slash in the URI in their http request.
As in "GET tables HTTP/1.1" instead of "GET /tables HTTP/1.1". Jersey framework does not
parse this correctly, so added logic to rewrite the request in such cases. Previous version
of third-eye library will now work correctly with the jersey server.

The log message, when such an API comes in, shows up as follows:
2017/08/09 10:42:14.643 WARN [HeaderAdder] [grizzly-http-server-15] [pinot-controller] [] Rewriting new Request URI http://localhost:8862/tables (incomingBaseUri = http://localhost:8862/, incomingReqUri = http://localhost:8862tables)

Also, fixed a but in the API that changes state of a segment. restlet API (and therefore
the derived jersey implementation) threw exception when changing state of all segments of
a table. Fixed in the jersey part.

Fixed a bug in the jersey implementation where toggling the state of one segment would end up
toggling the state of all segments

Also added a log of each request and response code.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/ControllerResponseFilter.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/ControllerResponseFilter.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/ControllerResponseFilter.java
```diff
@@ -0,0 +1,42 @@
...
+public class ControllerResponseFilter implements ContainerResponseFilter {
+  public static final Logger LOGGER = LoggerFactory.getLogger(ControllerResponseFilter.class);
+  @Override
+  public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext)
+      throws IOException {
+    final String method = requestContext.getMethod();
+    final String uri = requestContext.getUriInfo().getRequestUri().toString();
+    final int respStatus = responseContext.getStatus();
+    final String reasonPhrase = responseContext.getStatusInfo().getReasonPhrase();
+    LOGGER.info("Handled request {} {} status code {} {}", method, uri, respStatus, reasonPhrase);
```
antumbde: log source (ip)

PR #1777, opened 2017-08-10
[TE] Update_time should automatically update to current timestamp on table update

npawar:
The update time field in our tables should have been set to automatically update to current timestamp on table updates. This is not done. The update code assumes that this field will be automatically updated, and hence skips updating it. As a result, update time is always equal to create time. 
We need to make some enhancememts to thirdeye, wherein knowing the last update time is necessary.


thirdeye/thirdeye-pinot/src/test/resources/schema/create-schema.sql
```diff
@@ -27,7 +27,7 @@ create table if not exists anomaly_merge_config_index (
     name varchar(200) not null,
     base_id bigint(20) not null,
     create_time timestamp,
-    update_time timestamp default current_timestamp,
+    update_time timestamp default current_timestamp on update current_timestamp,
```
kishoreg: Did this work? There was some issue with this. Not sure if local test db support them

npawar: you're right, looks like test db doesn't support on update. Will update it in the code then.

npawar: Fixed @kishoreg

PR #1782, opened 2017-08-10
Fix the thread-safety issue for realtime inverted index

Jackie-Jiang:
There are 2 thread-safety issues:
1. HashMap itself is not thread-safe when multiple threads read and write concurrently
2. MutableRoaringBitmap is not thread-safe

To fix this:
1. Use read-write lock to control access to the bitmaps (optimize performance by replacing map with list)
2. Add a wrapper for MutableRoaringBitmap so that When fetching bitmap, make a clone and return the clone

Also cleaned up the code so that all columns use the same RealtimeInvertedIndexReader


pinot-core/src/main/java/com/linkedin/pinot/core/operator/docidsets/BitmapDocIdSet.java
```diff
@@ -70,11 +70,10 @@ public BitmapDocIdSet(String datasourceName, BlockMetadata blockMetadata, int st
         orBitmap.flip(startDocId, endDocId + 1); // end is exclusive
       }
       answer = orBitmap;
-
-    } else if (bitmaps.length == 1){
+    } else if (bitmaps.length == 1) {
       answer = bitmaps[0];
     } else {
-      answer = new MutableRoaringBitmap().toMutableRoaringBitmap();
+      answer = new MutableRoaringBitmap();
```
snleee: good catch 👍

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/invertedindex/RealtimeInvertedIndexReader.java
```diff
@@ -0,0 +1,102 @@
...
+    ReentrantReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+    _readLock = readWriteLock.readLock();
+    _writeLock = readWriteLock.writeLock();
+  }
+
+  /**
+   * Add the document id to the bitmap for the given dictionary id.
+   */
+  public void add(int dictId, int docId) {
+    if (_bitmaps.size() == dictId) {
```
snleee: Add a comment that this check is for new entry to the list.

Jackie-Jiang: Next line is the documentation :-P

PR #1783, opened 2017-08-10
Surface function baseline to users, and amend the baseline with anomaly results

cyy0714:
When onboard a new detection function, we propose to show multiple anomalies in a single view. This requires backend works on enable these functionalities. This endpoint is to modify an existing endpoint and enhance its functionality. 

- Modify the baseline endpoint to switch online and offline mode, and amend the return baseline by information in anomalies

- Modify AnomaliesResource to enable offline analysis baseline

- Add frequency into function update, and add java doc for the create and update funciton

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -75,6 +75,7 @@ import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.text.StrSubstitutor;
 import org.joda.time.DateTime;
+import org.joda.time.Hours;
 import org.joda.time.Interval;
 import org.joda.time.format.ISODateTimeFormat;
 import org.json.JSONObject;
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -563,26 +594,175 @@ public Response applyAutotuneConfig(@PathParam("autotuneConfigId") @NotNull long
...
+   */
+  public AnomalyTimelinesView amendAnomalyTimelinesViewWithAnomalyResults(AnomalyFunctionDTO anomalyFunctionSpec,
+      AnomalyTimelinesView originalTimelinesView, DimensionMap dimensionMap) {
+    if (originalTimelinesView == null || anomalyFunctionSpec == null) {
+      LOG.error("AnomalyTimelinesView or AnomalyFunctionDTO is null");
+      return null;
+    }
+    long bucketMillis =
+        (new TimeGranularity(anomalyFunctionSpec.getBucketSize(), anomalyFunctionSpec.getBucketUnit())).toMillis();
+    Tuple2<TimeSeries, TimeSeries> timeSeriesTuple = TimeSeriesUtils.toTimeSeries(originalTimelinesView);
```
cyy0714: We discussed about using Pair offline. However, Pair takes type implements Serializable; however, TimeSeries is not Serializable class. We are not able to use Pair.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/util/TimeSeriesUtils.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Convert the give observed and expected time series to anomaly time lines view
+   * @param observed
+   * @param expected
+   * @param bucketMillis
+   * @return
+   */
+  public static AnomalyTimelinesView toAnomalyTimeLinesView(TimeSeries observed, TimeSeries expected, long bucketMillis) {
+    if (!observed.timestampSet().equals(expected.timestampSet())) {
```
cyy0714: Thanks for the comment

cyy0714: Yes. If this happens, we need to take care of the timestamp set inconsistency.

PR #1791, opened 2017-08-14
Hot fix for controller admin health check API

mcvsubbu:


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotControllerHealthCheck.java
```diff
@@ -35,7 +35,7 @@
   ControllerConf controllerConf;
 
   @GET
-  @Path("/pinot-controller/admin")
+  @Path("/")
```
antumbde: This will conflict with LandingPageHandler.

Why change this ? It seems compatible with existing

PR #1795, opened 2017-08-14
Added support for dimension columns to be no-dictinary columns while consuming realtime segments

mcvsubbu:
All the plumbing has already been done for no-dictionary columns, we were just supporting the
feature only for metric columns.

Added logic to support no-dictionary for dimension columns as well, and added a dimension column
in the test case for coverage

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
@@ -140,8 +140,9 @@ public class RealtimeSegmentImpl implements RealtimeSegment {
     }
     // dictionary assignment for dimensions and time column
     for (String column : dataSchema.getDimensionNames()) {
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -784,6 +773,32 @@ public GenericRow getRawValueRowAt(int docId, GenericRow row) {
...
+            docId);
+        row.putField(columnName, floatValue);
+        break;
+      case DOUBLE:
+        double doubleValue = ((FixedByteSingleColumnSingleValueReaderWriter)columnIndexReaderWriterMap.get(columnName)).getDouble(
+            docId);
+        row.putField(columnName, doubleValue);
+        break;
+      default:
+        throw new UnsupportedOperationException("unsopported data type " + dataType.toString() + " for column " + columnName);
```
mayankshriv: Update message to indicate this is for no dictionary columns.

PR #1797, opened 2017-08-15
When adding a table, first check if tenant exists

Jackie-Jiang:
If tenant does not exist, adding table could result in an unstable state where routing table is not set up correctly (no error or exception thrown out)
With this fix, we can prevent wrong table config push into cluster
Also enhance log to log the table name

Modified tests to pass the new checks

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -1037,38 +1037,40 @@ public class PinotHelixResourceManager {
    * @throws InvalidTableConfigException
    * @throws TableAlreadyExistsException for offline tables only if the table already exists
    */
-  public void addTable(TableConfig config)
-      throws IOException {
-    TableType type = config.getTableType();
-
+  public void addTable(@Nonnull TableConfig tableConfig) throws IOException {
+    String tableNameWithType = tableConfig.getTableName();
+    TenantConfig tenantConfig;
     if (isSingleTenantCluster()) {
-      TenantConfig tenantConfig = new TenantConfig();
-      tenantConfig.setBroker(
-          ControllerTenantNameBuilder.getBrokerTenantNameForTenant(ControllerTenantNameBuilder.DEFAULT_TENANT_NAME));
-      switch (type) {
-        case OFFLINE:
-          tenantConfig.setServer(ControllerTenantNameBuilder.getOfflineTenantNameForTenant(
-              ControllerTenantNameBuilder.DEFAULT_TENANT_NAME));
-          break;
-        case REALTIME:
-          tenantConfig.setServer(ControllerTenantNameBuilder.getRealtimeTenantNameForTenant(
-              ControllerTenantNameBuilder.DEFAULT_TENANT_NAME));
-          break;
-        default:
-          throw new InvalidTableConfigException("UnSupported table type: " + type);
-      }
-      config.setTenantConfig(tenantConfig);
+      tenantConfig = new TenantConfig();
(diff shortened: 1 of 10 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1037,75 +1037,77 @@ public Schema getTableSchema(@Nonnull String tableName) {
...
     }
 
-    SegmentsValidationAndRetentionConfig segmentsConfig = config.getValidationConfig();
-    switch (type) {
+    // Check if tenant exists before creating the table
+    TableType tableType = tableConfig.getTableType();
+    String brokerTenantName = ControllerTenantNameBuilder.getBrokerTenantNameForTenant(tenantConfig.getBroker());
+    List<String> brokersForTenant = _helixAdmin.getInstancesInClusterWithTag(_helixClusterName, brokerTenantName);
+    if (brokersForTenant.isEmpty()) {
+      throw new InvalidTableConfigException(
```
mcvsubbu: log info before throwing exception

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1037,75 +1037,77 @@ public Schema getTableSchema(@Nonnull String tableName) {
...
+          "Broker tenant: " + brokerTenantName + " does not exist for table: " + tableNameWithType);
+    }
+    String serverTenantName =
+        ControllerTenantNameBuilder.getTenantName(tenantConfig.getServer(), tableType.getServerType());
+    if (_helixAdmin.getInstancesInClusterWithTag(_helixClusterName, serverTenantName).isEmpty()) {
+      throw new InvalidTableConfigException(
```
mcvsubbu: log info

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1272,31 +1275,18 @@ public void updateIndexingConfigFor(String tableName, TableType type, IndexingCo
...
   }
 
-  private void handleBrokerResource(TableConfig tableConfig) {
-    try {
-      final String brokerTenant =
-          ControllerTenantNameBuilder.getBrokerTenantNameForTenant(tableConfig.getTenantConfig().getBroker());
-      if (_helixAdmin.getInstancesInClusterWithTag(_helixClusterName, brokerTenant).isEmpty()) {
-        throw new RuntimeException("broker tenant : " + tableConfig.getTenantConfig().getBroker() + " does not exist");
-      }
-      LOGGER.info("Trying to update BrokerDataResource IdealState!");
```
mcvsubbu: please add this info log back, but with the table name now. thanks.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: This exception will be handled on upper level and we will log info/warning there

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: Same as above

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: Added back.

PR #1800, opened 2017-08-15
TE: fixing percentage change discrepancy between ui and email, adding created/updated by fields in objects

puneetjaiswal:
- fixing percentage change discrepancy between ui and email
- adding created/updated by fields in objects

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dao/GenericPojoDao.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dao/GenericPojoDao.java
@@ -4,6 +4,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.Sets;
 import com.google.inject.Inject;
 import com.linkedin.thirdeye.anomaly.utils.ThirdeyeMetricsUtil;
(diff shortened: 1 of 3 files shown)


apucher:
That's really 2 PRs

puneetjaiswal:
Yeah, as the description says. Changes are pretty small, is not worth effort splitting it.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dao/GenericPojoDao.java
```diff
@@ -163,8 +166,23 @@ public Connection getConnection() throws SQLException {
     return dataSource.getConnection();
   }
 
+  private String getCurrentPrincipal() {
```
bl44: Should we call it getCurrentUser? or provide some java doc here.

puneetjaiswal: Principal is the authenticated actor. It could be a user, account or group.

puneetjaiswal: - we will need to add a mocked user session provider for unit testing. Something that we would have when implementing authorization.

PR #1802, opened 2017-08-15
Fix tables/{tableName}/instances API and other exceptions

mcvsubbu:
Modified the tables/{tableName}/instances api in jersey to be the same as the older one.
Also, we were seeing internal exceptions since controllerMetrics was not injected correctly.
Fixed, and removed unused injects in classes


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -135,7 +138,7 @@ public String listTableConfigs(
     }
   }
 
-  private String listTableConfigs(@Nonnull String tableName, @Nonnull String tableTypeStr) {
```
Jackie-Jiang: Annotate it as @Nullable

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableSegmentConfigs.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.controller.api.resources;
 
+import com.linkedin.pinot.controller.api.ControllerAdminApiApplication;
```
Jackie-Jiang: Remove

PR #1804, opened 2017-08-16
Plumbing for realtime consuming segment statistics

mcvsubbu:
In order to allocate memory efficiently for dictionary for a column, we can attempt to use
the statistics gathered from previously completed segments.

Added a class to keep track of statistics for N previously completed segments for a table on local disk
of each server. For now, we have code to store average size (in case of a string column) and cardinality
(for all columns). These can be used when initializing a dictionary for the next segment.

Some methods are synchronized since we expect that segments may be completing at the same time for all
partitions hosted in a machine.

Added a test class.

Added methods in off-heap dictionary classes to get the statistics we want

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
@@ -21,7 +21,6 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.OutputStream;
@@ -37,6 +36,7 @@ import javax.annotation.Nullable;
  */
 public class RealtimeSegmentStatsHistory implements Serializable {
   private static final long serialVersionUID = 1L;  // Change this if a new field is added to this class
+  // XXX MAX_NUM_ENTRIES should be a final variable, but we need to modify it for testing.
   private static int MAX_NUM_ENTRIES = 16;  // Max number of past segments for which stats are kept
   // Fields to be serialzied.
   private int _cursor = 0;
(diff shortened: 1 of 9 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
```diff
@@ -0,0 +1,279 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+
+/*
+ * Keeps history of statistics for segments consumed from kafka for a single table.
+ */
+public class RealtimeSegmentStatsHistory implements Serializable {
+  private static final long serialVersionUID = 1L;  // Change this if a new field is added to this class
+  private static int MAX_NUM_ENTRIES = 16;  // Max number of past segments for which stats are kept
```
antumbde: I thought only final variables should be all caps. Please check once

Or is there a way to keep this final for safety ?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
```diff
@@ -0,0 +1,279 @@
...
+    return _entries[index];
+  }
+
+  @Override
+  public String toString() {
+    return "cursor=" + getCursor() + ",numEntries=" + getArraySize() + ",isFull=" + isFull();
+  }
+
+  public synchronized void serializeInto(File outFile) throws IOException {
+    OutputStream os = new FileOutputStream(outFile);
```
antumbde: Both streams need to be closed. Prefer try-with-resource or close in finally

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
```diff
@@ -0,0 +1,279 @@
...
+    ObjectOutputStream obos = new ObjectOutputStream(os);
+    obos.writeObject(this);
+    obos.flush();
+    os.close();
+  }
+
+  public static RealtimeSegmentStatsHistory deserialzeFrom(File inFile) throws IOException, ClassNotFoundException {
+    InputStream is = new FileInputStream(inFile);
```
antumbde: close both and close in finally or try-with-resource

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
mcvsubbu: I had set it to be final, but wanted to change it for testing reasons. I will add a comment saying that the static should really be final. Not sure if there are other ways of testing this.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
mcvsubbu: ok

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
mcvsubbu: ok

PR #1807, opened 2017-08-16
Fixing API incomaptibilities

mcvsubbu:
Fix jersey API descrepancies in /tables/{tableName}/segments/{segmentName}  and /schemas/{schemaName}

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
@@ -16,19 +16,6 @@
 
 package com.linkedin.pinot.controller.api.resources;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URLDecoder;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import org.apache.helix.ZNRecord;
-import org.apache.helix.store.zk.ZkHelixPropertyStore;
-import org.json.JSONArray;
-import org.json.JSONException;
-import org.json.JSONObject;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linkedin.pinot.common.config.TableNameBuilder;
@@ -41,6 +28,12 @@ import com.linkedin.pinot.controller.helix.core.PinotResourceManagerResponse;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -186,26 +189,48 @@ public String listMetadataForOneSegment(
...
       @ApiParam (value = "Name of segment", required = true) @PathParam("segmentName") @Encoded String segmentName,
       @ApiParam(value = "realtime|offline", required = false) @QueryParam("type") String tableTypeStr
-  ) {
+  ) throws JSONException {
     segmentName = checkGetEncodedParam(segmentName);
     CommonConstants.Helix.TableType tableType = Constants.validateTableType(tableTypeStr);
+    return listSegmentMetadataInternal(tableName, segmentName, tableType);
+  }
+
+  private String listSegmentMetadataInternal(@Nonnull  String tableName, @Nonnull String segmentName,
```
Jackie-Jiang: I don't think this helper method is necessary. We just need a check as following:
if ((tableType == null || tableType == OFFLINE) && _pinotHelixResourceManager.hasOfflineTable(tableName)) {
      String offlineTableName = TableNameBuilder.OFFLINE.tableNameWithType(tableName);
      result.put(getSegmentMetaData(offlineTableName, segmentName, TableType.OFFLINE));
}
This can ensure the behavior is the same as before

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -186,26 +189,48 @@ public String listMetadataForOneSegment(
...
     // The segment should appear only in one
     JSONArray metata;
 
-    metata = getSegmentMetaData(tableName, segmentName, CommonConstants.Helix.TableType.OFFLINE);
-    result.put(metata);
-    metata = getSegmentMetaData(tableName, segmentName, CommonConstants.Helix.TableType.REALTIME);
-    result.put(metata);
-    return result.toString();
+    if (_pinotHelixResourceManager.hasOfflineTable(tableName)) {
+      metata = getSegmentMetaData(TableNameBuilder.OFFLINE.tableNameWithType(tableName), segmentName, CommonConstants.Helix.TableType.OFFLINE);
```
antumbde: (typo) metadata

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -186,26 +189,48 @@ public String listMetadataForOneSegment(
...
     segmentName = checkGetEncodedParam(segmentName);
     CommonConstants.Helix.TableType tableType = Constants.validateTableType(tableTypeStr);
+    return listSegmentMetadataInternal(tableName, segmentName, tableType);
+  }
+
+  private String listSegmentMetadataInternal(@Nonnull  String tableName, @Nonnull String segmentName,
+      @Nullable CommonConstants.Helix.TableType tableType) throws JSONException {
     // The code in restlet.resources seems to return an array of arrays, so we will do the same
     // to maintain backward compatibility
     JSONArray result = new JSONArray();
```
antumbde: double checking: this will return json array as top level json and not json object. I hope that's desired goal here

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
mcvsubbu: I went through this method first, and then decided we need a helper method.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
mcvsubbu: yes, the two APIs now function same, except one returns a json array of strings, and the other returns a json array of objects.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
mcvsubbu: wil fix.

PR #1808, opened 2017-08-17
add frequency as query parameter for endpoint that update anomaly detection function, make time series retrieve endpoints do not roll-up

bl44:
add frequency as query parameter for endpoint that update anomaly detection function, make time series retrieve endpoints do not roll-up

Anomaly detection can work on 'dimension exploration' - for a given dimension, run parallel anomaly detection for all categories in this dimension.
for example if we set anomaly detection for 'page views' and explore on 'country' dimension, and there are ten distinct countries, we are going to do
10 parallel detection, treat each country as a independent metric.

If the rollup flag is 'true', we will group those low contribution coutries (<1%) to a single category 'OTHER'. For example assume most of pageviews come from
US, CA and CN, all 7 rest have only contribute 2% together, 0.7% each. Then all the rest will be group to 'OTHER' and effectively we are doing anomaly detection for
'US', 'CA', 'CN' and 'OTHER'

Here we set the default behavior to rollup = flase, to do all possible categories in a perticular dimension.

For API calls we make it a query variable.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -415,7 +415,8 @@ public class AnomalyResource {
       @QueryParam("exploreDimension") String exploreDimensions,
       @QueryParam("filters") String filters,
       @QueryParam("properties") String properties,
-      @QueryParam("isActive") Boolean isActive) throws Exception {
+      @QueryParam("isActive") Boolean isActive,
+      @QueryParam("frequency") String frequency) throws Exception {
 
     AnomalyFunctionDTO anomalyFunctionSpec = anomalyFunctionDAO.findById(id);
     if (anomalyFunctionSpec == null) {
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -415,7 +415,8 @@ public Response updateAnomalyFunction(@NotNull @PathParam("id") Long id,
       @QueryParam("exploreDimension") String exploreDimensions,
       @QueryParam("filters") String filters,
       @QueryParam("properties") String properties,
-      @QueryParam("isActive") Boolean isActive) throws Exception {
+      @QueryParam("isActive") Boolean isActive,
+      @QueryParam("frequency") String frequency) throws Exception {
```
puneetjaiswal: is it frequency of anomaly job or granularity of data ?
if later then please rename it to something like aggGranularity (aggregation granularity)

bl44: its frequency of anomaly job. Like schedule a job very 15 MINUTES.

PR #1809, opened 2017-08-17
Refactor DefaultGroupKeyGenerator for performance concern (rename to DictionaryBasedGroupKeyGenerator)

Jackie-Jiang:
Specialize single multi-value group-by column case
Add an interface RawKeyHolder as an abstraction on different storage type
Add IntMapBasedHolder to use Int2IntOpenHashMap
Make group id in GroupKey wrapper primitive int

Update tests to test against all 4 RawKeyHolders

Tested one query with single multi-value group-by column, reduced latency from 1100ms to 750ms


pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/groupby/DictionaryBasedGroupKeyGenerator.java
```diff
@@ -0,0 +1,844 @@
...
+    }
+  }
+
+  /**
+   * Helper method to calculate raw keys that can fit into integer for the given index.
+   *
+   * @param index Index in block
+   * @return Array of integer raw keys
+   */
+  @SuppressWarnings("Duplicates")
```
mayankshriv: Nit: Any way to get rid of the duplicate? Or is it IDE's limitation for not being smart enough to realize that data types changed?

Jackie-Jiang: Haha, for optimizations using primitive type, this is unavoidable.

PR #1812, opened 2017-08-18
Optimize query parsing

jfim:
Optimize some of the hot paths in query parsing:
- Avoid doing replaceAll if it would do nothing (eg. for string
  literals, we use replaceAll to replace double quoted quotes, for
  things such as 'O''Reilly')
- Use a HashSet instead of a TreeSet for items in an IN clause, avoids
  having to rebalance the tree many times if the items in the IN clause
  are in sorted order


jfim:
This gives about 30% speedup on large queries with many strings in an `IN` clause. Not 100% sure about the `HashSet` instead of `TreeSet`, I think some tests depend on the ordering of the items in an `IN` clause.

PR #1816, opened 2017-08-19
[TE] using new endpoint

justYves:
Using the new endpoint for create alerts


thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
```diff
@@ -346,7 +346,9 @@ export default Ember.Controller.extend({
   callReplayStart(functionId, startTime, endTime) {
     const granularity = this.get('graphConfig.granularity').toLowerCase();
     const speedUp = granularity.includes('hour') || granularity.includes('day');
```
cecilynie: speedUp is "true" for minutely metrics, speedUp is "false" for hourly/daily metrics, is the logic correct here?

justYves: yes, if it's `hour` or `day`  it'll return `true`

cecilynie: Actually if it's hour/day speedup should be "false". Minutely should have speedup to be "true" since it has more data.

justYves: ah gotcha. Was this a recent change ?

cecilynie: It's actually not. The logic is the same in old flow. Probably a bug?

justYves:
closing in favor of https://github.com/linkedin/pinot/pull/1867

PR #1820, opened 2017-08-21
Bubbling up error during Avro -> JSON schema conversion

jenniferdai:
Currently, when we try to convert schemas, we will ignore the fact that neither dimensions nor metrics is presented as an argument by the user. This is confusing because it ends up printing out an error after it tries to convert the schema and doesn't mention the real cause.

Now we check that at least one of dimensions or metrics is passed in at the beginning and print out the relevant error message.

Tested by running pinot-admin.

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/AvroSchemaToPinotSchema.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/AvroSchemaToPinotSchema.java
@@ -69,6 +69,11 @@ public class AvroSchemaToPinotSchema extends AbstractBaseAdminCommand implements
   public boolean execute() throws Exception {
     Schema schema = null;
 
+    if (_dimensions == null && _metrics == null) {
+      LOGGER.error("Error: Missing required argument, please specify -dimensions, -metrics, or both");
+      return false;
+    }
+    
     if (_avroSchemaFileName != null) {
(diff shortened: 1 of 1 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/AvroSchemaToPinotSchema.java
```diff
@@ -69,6 +69,11 @@
   public boolean execute() throws Exception {
     Schema schema = null;
 
+    if (_dimensions == null && _metrics == null) {
```
mayankshriv: Use the required attribute from args4j instead.

jenniferdai: The reason I didn't use the required attribute is in case someone wanted to create a table using only dimensions or only metrics

mayankshriv: These are not mutually exclusive, right? If so, we cannot use forbid. If args4j does not have a feature to have at least one of two (or more) as required, then go ahead and use the 'if' check you had originally.

jenniferdai:
I got the forbids idea from here: https://stackoverflow.com/questions/28161838/args4j-how-to-make-an-argument-required-if-another-argument-is-isnt-given

PR #1823, opened 2017-08-22
Wire-in config and segment stats history for off-heap memory allocation for consuming segments

mcvsubbu:
Added a server config (pinot.server.instance.realtime.alloc.offheap to be set to "true") for off-heap
allocation of memory for dictionary. Currently, off-heap dictionary is always memory-mapped.

Mmap files for dictionary are in <dataDir>/<tableName>/consumers/<segmentName>.{0, 1, ...}. Multiple columns
use the same file (0.5g in size), until the entire file is mapped, and then a new file is created if any column
needs more memory.

Past history of column statistics are stored in <dataDir>/<tableName>/consumers/stats.ser. History is saved
in a circular buffer that has room for 16 segments. The file is persisted across restarts of the same server.
The handle for the stats file is created when the TableDataManager is created for that
table (i.e when the first segment for that table is assigned to a server).

The same stats object is kept around in memory until the server shuts down.
The file is modified by adding new segment statistics each time a consuming segment is destroyed.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
@@ -105,7 +105,7 @@ public class RealtimeTableDataManager extends AbstractTableDataManager {
   }
 
   public String getConsumerDir() {
-    return _tableDataDir + "/" + CONSUMERS_DIR;
+    return _tableDataDir + File.separator + CONSUMERS_DIR;
   }
 
   public void notifySegmentCommitted(RealtimeSegmentZKMetadata metadata, IndexSegment segment) {
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
@@ -35,11 +35,14 @@ import com.linkedin.pinot.core.data.manager.offline.AbstractTableDataManager;
 import com.linkedin.pinot.core.data.manager.offline.SegmentDataManager;
 import com.linkedin.pinot.core.indexsegment.IndexSegment;
 import com.linkedin.pinot.core.indexsegment.columnar.ColumnarSegmentLoader;
+import com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory;
 import com.linkedin.pinot.core.realtime.impl.kafka.KafkaConsumerManager;
 import com.linkedin.pinot.core.segment.index.loader.IndexLoadingConfig;
 import com.linkedin.pinot.core.segment.index.loader.LoaderUtils;
 import java.io.File;
+import java.io.IOException;
 import java.util.List;
+import java.util.UUID;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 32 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -519,6 +554,24 @@ public int getSuccessIndexedCount() {
...
+      for (Map.Entry<String, MutableDictionary> entry : dictionaryMap.entrySet()) {
+        RealtimeSegmentStatsHistory.ColumnStats columnStats = new RealtimeSegmentStatsHistory.ColumnStats();
+        columnStats.setCardinality(entry.getValue().length());
+        columnStats.setAvgColumnSize(entry.getValue().getAvgValueSize());
+        segmentStats.setColumnStats(entry.getKey(), columnStats);
+      }
+      segmentStats.setNumRowsConsumed(numDocsIndexed);
+      segmentStats.setMemUsedBytes(memoryManager.getTotalMemBytes());
+      final long now = System.currentTimeMillis();
+      segmentStats.setNumSeconds((int)(now - startTimeMillis/1000));
```
mayankshriv: (now - startTimeMillis) /1000 ?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -76,6 +82,30 @@ protected void doShutdown() {
   protected void doInit() {
     _leaseExtender = SegmentBuildTimeLeaseExtender.create(getServerInstance());
     LOGGER = LoggerFactory.getLogger(_tableName + "-RealtimeTableDataManager");
+    File consumersDir = new File(_tableDataDir, CONSUMERS_DIR);
+    consumersDir.mkdirs();
+    File statsFile = new File(consumersDir, STATS_FILE_NAME);
+    try {
+      _statsHistory = RealtimeSegmentStatsHistory.deserialzeFrom(statsFile);
+    } catch (IOException |ClassNotFoundException e) {
```
mayankshriv: Bitwise OR?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
```diff
@@ -253,23 +267,27 @@ public String toString() {
     return "cursor=" + getCursor() + ",numEntries=" + getArraySize() + ",isFull=" + isFull();
   }
 
-  public synchronized void serializeInto(File outFile) throws IOException {
-    try (OutputStream os = new FileOutputStream(outFile);
+  public synchronized void save() {
+    try (OutputStream os = new FileOutputStream(new File(_historyFilePath));
```
mayankshriv: Is this using java serialization or is writeObject overridden elsewhere? If former, may be better to avoid that.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
mcvsubbu: yes, that is the syntax

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -76,6 +82,30 @@ protected void doShutdown() {
...
+      LOGGER.warn("Saved unreadable {} into {}", statsFile.getAbsolutePath(), savedFile.getAbsolutePath());
+    }
+  }
+
+  public RealtimeSegmentStatsHistory getStatsHistory() {
+    return _statsHistory;
+  }
+
+  public String getConsumerDir() {
+    return _tableDataDir + "/" + CONSUMERS_DIR;
```
mcvsubbu: Done

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
mcvsubbu: good catch

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
mcvsubbu: It is using java serialization. Worst case here is that we don't read history and start afresh, and nothing else breaks.

PR #1825, opened 2017-08-22
Adding a configuration for the length of query logs.

snleee:
Query logging becomes the bottleneck when a use case sends long queries with
high QPS. To avoid this, this change makes the query logging length configurable.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
@@ -67,6 +67,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 import org.apache.commons.configuration.Configuration;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.thrift.protocol.TCompactProtocol;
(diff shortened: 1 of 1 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -252,7 +257,7 @@ public BrokerResponse handleRequest(@Nonnull JSONObject request)
             + "numEntriesScannedPostFilter: {}, totalDocs: {}, scatterGatherStats: {}, query: {}", requestId,
         brokerRequest.getQuerySource().getTableName(), totalTimeMs, brokerResponse.getNumDocsScanned(),
         brokerResponse.getNumEntriesScannedInFilter(), brokerResponse.getNumEntriesScannedPostFilter(),
-        brokerResponse.getTotalDocs(), scatterGatherStats, pql);
+        brokerResponse.getTotalDocs(), scatterGatherStats, StringUtils.substring(pql, 0, _queryLogLength));
```
mayankshriv: I take it that the substring method here is a no-op if max length > query length?

snleee: Yes. If max length > query length, `substring()` will not create a new string object.

PR #1827, opened 2017-08-23
Fix NPE and add a test case for older history files

mcvsubbu:
If we fail to read the stats history file for any reason, we would end up getting an NPE because _statsHistory
went uninitialized. Fixed to create an instance with a new file.

Also added a v1 serialized file to the repository so that we can ensure that future changes to the class
can still read v1 files.


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -96,7 +96,12 @@ protected void doInit() {
         LOGGER.error("Could not move {} to {}", statsFile.getAbsolutePath(), savedFile.getAbsolutePath(), e1);
         throw new RuntimeException(e);
       }
-      LOGGER.warn("Saved unreadable {} into {}", statsFile.getAbsolutePath(), savedFile.getAbsolutePath());
+      LOGGER.warn("Saved unreadable {} into {}. Creating a fresh instance", statsFile.getAbsolutePath(), savedFile.getAbsolutePath());
+      try {
+        _statsHistory = RealtimeSegmentStatsHistory.deserialzeFrom(statsFile);
+      } catch (Exception e2) {
+        throw new RuntimeException(e2);
```
mayankshriv: Could use Utils.rethrowException() to avoid throwing a RuntimeException.

pinot-core/src/test/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistoryTest.java
```diff
@@ -180,6 +181,30 @@ public void testMultiThreadedUse() throws Exception {
     FileUtils.deleteQuietly(serializedFile);
   }
 
+  // This test attempts to ensure that future modifications to RealtimeSegmentStatsHistory does not prevent the software
```
mayankshriv: May be use javadoc?

PR #1828, opened 2017-08-23
Fixed a bug with priority queue selection logic

antumbde:
Select a group with lower priority if the group is below soft thread limit
and current winner is above soft thread limit.


kishoreg:
I see that we create StringBuilder to log the winner group but it gets logged only in debug mode. Can we avoid creating the string buffer if debug is turned off?

antumbde:
Modified to invoke StringBuilder only if debug is enabled

PR #1829, opened 2017-08-23
Fix time granularity in anomalies resource

cyy0714:
The getTimelinesViewInMonitoringWindow() takes only the bucket time granularity from dataset configs. The function also has bucket time granularity from user. This design fails when the two config are not the same. This pull request is to resolve this issue. It takes both definitions of buckets, and apply the definition from function in default. But, it also compares the definition from dataset. If the function's is smaller than the dataset's, it applies the dataset's instead.

Another part is to allow users to update function bucket time granularity via the function update endpoint.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -471,8 +471,7 @@ public class AnomalyResource {
       @QueryParam("properties") String properties,
       @QueryParam("isActive") Boolean isActive,
       @QueryParam("frequency") String frequency,
-      @QueryParam("bucketSize") Integer bucketSize,
-      @QueryParam("bucketUnit") String bucketUnit) throws Exception {
+      @QueryParam("bucket") String bucketTimeGranularity) throws Exception {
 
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -470,7 +470,9 @@ public Response updateAnomalyFunction(@NotNull @PathParam("id") Long id,
       @QueryParam("filters") String filters,
       @QueryParam("properties") String properties,
       @QueryParam("isActive") Boolean isActive,
-      @QueryParam("frequency") String frequency) throws Exception {
+      @QueryParam("frequency") String frequency,
```
bl44: can you follow the pattern in 'post' method where you do '1_MINTUES' type of input.

bl44: something like this:

   // modify dataGranularity, the format of input should be like '1_MINUTES', '1_HOURS', '1_DAYS'
    if (StringUtils.isNotEmpty(userInputDataGranularity)) {
      TimeGranularity dataGranularity = TimeGranularity.fromString(userInputDataGranularity);
      anomalyFunctionSpec.setBucketSize(dataGranularity.getSize());
      anomalyFunctionSpec.setBucketUnit(dataGranularity.getUnit());
    }

PR #1841, opened 2017-08-25
Changing retry policy to increase timeout for kafka exceptions

jenniferdai:


jfim:
There needs to be a retry for data reads as well, not only metadata fetches. Can you make sure that the retry covers all calls to the Kafka consumer?

PR #1842, opened 2017-08-25
Add list of DateTimeFieldSpec to Schema

npawar:
Introduce DateTimeFieldSpec to pinot Schema for the following reasons:
1. Have an ability in the pinot schema to distinguish between time column format and data granularity. Currently there's no way to specify data granularity, and the timeSpec is being used stretched to do both in one shot.
2. Bring in a placeholder for creating a uniform time column across all pinot tables (will be useful in doing rollups)
3. Provision for supporting multiple time columns of different type (PRIMARY, DERIVED etc).

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
@@ -32,6 +32,12 @@ import com.linkedin.pinot.common.utils.EqualityUtils;
 @JsonIgnoreProperties(ignoreUnknown = true)
 public final class DateTimeFieldSpec extends FieldSpec {
 
+  private static final String FORMAT_TOKENS_ERROR_STR = "format must be of pattern size:timeunit:timeformat(:pattern)";
+  private static final String FORMAT_PATTERN_ERROR_STR = "format must be of format [0-9]+:<TimeUnit>:<TimeFormat>(:pattern)";
+  private static final String TIME_FORMAT_ERROR_STR =
+      "format must be of format [0-9]+:<TimeUnit>:EPOCH or [0-9]+:<TimeUnit>:SIMPLE_DATE_FORMAT:<format>";
+  private static final String GRANULARITY_TOKENS_ERROR_STR = "granularity must be of format size:timeunit";
+  private static final String GRANULARITY_PATTERN_ERROR_STR = "granularity must be of format [0-9]+:<TimeUnit>";
   private static final String NUMBER_REGEX = "[1-9][0-9]*";
   private static final String COLON_SEPARATOR = ":";
 
@@ -150,22 +156,22 @@ public final class DateTimeFieldSpec extends FieldSpec {
     Preconditions.checkNotNull(format);
     String[] formatTokens = format.split(COLON_SEPARATOR);
     Preconditions.checkArgument(formatTokens.length == 3 || formatTokens.length == 4,
-        "format must be of format size:timeunit:timeformat(:pattern)");
+        FORMAT_TOKENS_ERROR_STR);
     Preconditions.checkArgument(formatTokens[0].matches(NUMBER_REGEX)
-        && EnumUtils.isValidEnum(TimeUnit.class, formatTokens[1]), "format must be of format [0-9]+:<TimeUnit>:<TimeFormat>(:pattern)");
+        && EnumUtils.isValidEnum(TimeUnit.class, formatTokens[1]), FORMAT_PATTERN_ERROR_STR);
     if (formatTokens.length == 3) {
       Preconditions.checkArgument(formatTokens[2].equals(TimeFormat.EPOCH.toString()),
-          "format must be of format [0-9]+:<TimeUnit>:EPOCH or [0-9]+:<TimeUnit>:SIMPLE_DATE_FORMAT:<format>");
+          TIME_FORMAT_ERROR_STR);
     } else {
       Preconditions.checkArgument(formatTokens[2].equals(TimeFormat.SIMPLE_DATE_FORMAT.toString()),
-          "format must be of format [0-9]+:<TimeUnit>:EPOCH or [0-9]+:<TimeUnit>:SIMPLE_DATE_FORMAT:<format>");
+          TIME_FORMAT_ERROR_STR);
     }
 
     Preconditions.checkNotNull(granularity);
     String[] granularityTokens = granularity.split(COLON_SEPARATOR);
-    Preconditions.checkArgument(granularityTokens.length == 2, "granularity must be of format size:timeunit");
+    Preconditions.checkArgument(granularityTokens.length == 2, GRANULARITY_TOKENS_ERROR_STR);
     Preconditions.checkArgument(granularityTokens[0].matches(NUMBER_REGEX)
-        && EnumUtils.isValidEnum(TimeUnit.class, granularityTokens[1]), "granularity must be of format [0-9]+:<TimeUnit>");
+        && EnumUtils.isValidEnum(TimeUnit.class, granularityTokens[1]), GRANULARITY_PATTERN_ERROR_STR);
 
   }
 
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
@@ -95,57 +95,12 @@ public final class DateTimeFieldSpec extends FieldSpec {
    *          the granularity will be 1:HOURS
    */
   public DateTimeFieldSpec(@Nonnull String name, @Nonnull DataType dataType, @Nonnull String format,
-      @Nonnull String granularity, Object defaultNullValue) {
-    super(name, dataType, true, defaultNullValue);
-    check(name, dataType, format, granularity);
-
-    _format = format;
-    _granularity = granularity;
-  }
-
-
-  /**
-   *
-   * @param name
-   * @param dataType
-   * @param format - size:timeunit:timeformat eg: 1:MILLISECONDS:EPOCH, 5:MINUTES:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd
-   * @param granularity - size:timeunit eg: 5:MINUTES
-   * @param dateTimeType
-   * @param defaultNullValue
-   */
-  public DateTimeFieldSpec(@Nonnull String name, @Nonnull DataType dataType, @Nonnull String format,
-      @Nonnull String granularity, DateTimeType dateTimeType, Object defaultNullValue) {
-    this(name, dataType, format, granularity, defaultNullValue);
-    _dateTimeType = dateTimeType;
-  }
-
-
-  /**
-   * @param name
-   * @param dataType
-   * @param format - size:timeunit:timeformat eg: 1:MILLISECONDS:EPOCH, 5:MINUTES:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd
-   * @param granularity - size:timeunit eg: 5:MINUTES
-   */
-  public DateTimeFieldSpec(@Nonnull String name, @Nonnull DataType dataType, @Nonnull String format,
-      @Nonnull String granularity) {
+      @Nonnull String granularity, DateTimeType dateTimeType) {
     super(name, dataType, true);
     check(name, dataType, format, granularity);
 
     _format = format;
     _granularity = granularity;
-  }
-
-
-  /**
-   * @param name
-   * @param dataType
-   * @param format - size:timeunit:timeformat eg: 1:MILLISECONDS:EPOCH, 5:MINUTES:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd
-   * @param granularity - size:timeunit eg: 5:MINUTES
-   * @param dateTimeType
-   */
-  public DateTimeFieldSpec(@Nonnull String name, @Nonnull DataType dataType, @Nonnull String format,
-      @Nonnull String granularity, DateTimeType dateTimeType) {
-    this(name, dataType, format, granularity);
     _dateTimeType = dateTimeType;
   }
 
@@ -188,6 +143,10 @@ public final class DateTimeFieldSpec extends FieldSpec {
     return _format;
   }
 
+  /**
+   * Required by JSON deserializer. DO NOT USE. DO NOT REMOVE.
+   * @param format
+   */
   public void setFormat(String format) {
     _format = format;
   }
@@ -197,6 +156,10 @@ public final class DateTimeFieldSpec extends FieldSpec {
     return _granularity;
   }
 
+  /**
+   * Required by JSON deserializer. DO NOT USE. DO NOT REMOVE.
+   * @param granularity
+   */
   public void setGranularity(String granularity) {
     _granularity = granularity;
   }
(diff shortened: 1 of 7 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
```diff
@@ -0,0 +1,222 @@
...
+  private void check(String name, DataType dataType, String format, String granularity) {
+    Preconditions.checkNotNull(name);
+    Preconditions.checkNotNull(dataType);
+
+    Preconditions.checkNotNull(format);
+    String[] formatTokens = format.split(COLON_SEPARATOR);
+    Preconditions.checkArgument(formatTokens.length == 3 || formatTokens.length == 4,
+        "format must be of format size:timeunit:timeformat(:pattern)");
+    Preconditions.checkArgument(formatTokens[0].matches(NUMBER_REGEX)
+        && EnumUtils.isValidEnum(TimeUnit.class, formatTokens[1]), "format must be of format [0-9]+:<TimeUnit>:<TimeFormat>(:pattern)");
```
mcvsubbu: private static string ERROR_STR = "format must be ...."

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
```diff
@@ -0,0 +1,222 @@
...
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.linkedin.pinot.common.utils.EqualityUtils;
+
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public final class DateTimeFieldSpec extends FieldSpec {
+
+  private static final String NUMBER_REGEX = "[0-9]+";
```
mcvsubbu: is 0 allowed? what does 0 mean?

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
```diff
@@ -0,0 +1,222 @@
...
+  public void setGranularity(String granularity) {
+    _granularity = granularity;
+  }
+
+  @Nullable
+  public DateTimeType getDateTimeType() {
+    return _dateTimeType;
+  }
+
+  public void setDateTimeType(DateTimeType dateTimeType) {
```
mcvsubbu: check which of these setters we actually need. If we can avoid individual setters that is preferred, but I suspect the json parsers need the setters? At the minimum add an annottion & javadoc so that nobody uses the setters. Perhaps @VisibleForTesting will work?

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
```diff
@@ -121,6 +130,19 @@ public void setMetricFieldSpecs(@Nonnull List<MetricFieldSpec> metricFieldSpecs)
     }
   }
 
+  @Nonnull
+  public List<DateTimeFieldSpec> getDateTimeFieldSpecs() {
+    return _dateTimeFieldSpecs;
+  }
+
+  public void setDateTimeFieldSpecs(@Nonnull List<DateTimeFieldSpec> dateTimeFieldSpecs) {
```
mcvsubbu: re-checking whether this setter is needed. Same comment as before in terms of annotation and javadoc comments.

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
```diff
@@ -121,6 +130,19 @@ public void setMetricFieldSpecs(@Nonnull List<MetricFieldSpec> metricFieldSpecs)
...
+    Preconditions.checkState(_dateTimeFieldSpecs.isEmpty());
+
+    for (DateTimeFieldSpec dateTimeFieldSpec : dateTimeFieldSpecs) {
+      addField(dateTimeFieldSpec);
```
mcvsubbu: Do you need to add it to the set as well as to the list?

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
```diff
@@ -288,6 +334,7 @@ public boolean validate(Logger ctxLogger) {
         switch (fieldType) {
           case DIMENSION:
           case TIME:
+          case DATE_TIME:
```
mcvsubbu: can fieldType be a string/float/double?

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
```diff
@@ -475,6 +522,31 @@ public SchemaBuilder addTime(@Nonnull TimeGranularitySpec incomingTimeGranularit
       return this;
     }
 
+    public SchemaBuilder addDateTime(@Nonnull String name, @Nonnull DataType dataType, @Nonnull String format,
```
mcvsubbu: Not sure if we need so many different versions of addDateTime. Start with one, and add more as needed.

pinot-common/src/main/java/com/linkedin/pinot/common/data/TimeGranularitySpec.java
```diff
@@ -31,7 +30,7 @@
  * <p>- <code>DataType</code>: data type of the time column (e.g. INT, LONG).
  * <p>- <code>TimeType</code>: time unit of the time column (e.g. MINUTES, HOURS).
  * <p>- <code>TimeUnitSize</code>: size of the time buckets (e.g. 10 MINUTES, 2 HOURS). By default this is set to 1.
- * <p>- <code>TimeFormat</code>: Can be either EPOCH (default) or SIMPLE_DATE_FORMAT:pattern e.g SIMPLE_DATE_FORMAT:yyyyMMdd 
+ * <p>- <code>TimeFormat</code>: Can be either EPOCH (default) or SIMPLE_DATE_FORMAT:pattern e.g SIMPLE_DATE_FORMAT:yyyyMMdd
```
mcvsubbu: If u did not change this file at all (and it seems like that), then please restore the master version of the file

pinot-common/src/test/java/com/linkedin/pinot/common/data/FieldSpecTest.java
```diff
@@ -192,6 +195,69 @@ public void testTimeFieldSpecConstructor() {
   }
 
   /**
+   * Test {@link TimeFieldSpec} constructors.
+   */
+  @Test
+  public void testDateTimeFieldSpecConstructor() {
```
mcvsubbu: I would add negative test cases as well, with 0, -1, etc for numbers

pinot-common/src/test/java/com/linkedin/pinot/common/data/FieldSpecTest.java
```diff
@@ -192,6 +195,69 @@ public void testTimeFieldSpecConstructor() {
...
+    DateTimeFieldSpec dateTimeFieldSpec3 = new DateTimeFieldSpec(name, dataType, format, granularity, DateTimeType.PRIMARY);
+    DateTimeFieldSpec dateTimeFieldSpec4 = new DateTimeFieldSpec(name, dataType, format, granularity, DateTimeType.PRIMARY, defaultNullValue);
+
+    Assert.assertFalse(dateTimeFieldSpec3.equals(dateTimeFieldSpec4));
+    dateTimeFieldSpec3.setDefaultNullValue(defaultNullValue);
+    Assert.assertEquals(dateTimeFieldSpec3, dateTimeFieldSpec4);
+
+  }
+
+  @Test(expectedExceptions = {IllegalArgumentException.class})
```
mcvsubbu: If you get IllegalArgumentException in line 233, the other lines will not be executed, right? Isnt it better to actually test for the exception in the line in which you expect it to happen?

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
npawar: Fixed to not allow 0

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
```diff
@@ -0,0 +1,222 @@
...
+
+  // Default constructor required by JSON de-serializer. DO NOT REMOVE.
+  public DateTimeFieldSpec() {
+    super();
+  }
+
+  /**
+   * @param name
+   * @param dataType
+   * @param format - size:timeunit:timeformat eg: 1:MILLISECONDS:EPOCH, 5:MINUTES:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd
```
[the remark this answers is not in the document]
npawar: Done

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
npawar: Done

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
npawar: I tried to make it consistent with the already existing metricFieldSpecs and dimensionFieldSpecs and timeFieldSpec. Those have setters, with no annotations. Should I add the annotations to those as well then?

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
npawar: I tried to make it consistent with the already existing metricFieldSpecs and dimensionFieldSpecs. Those have sets and lists, and methods to add into both. Not sure why it was done for those, and not sure if they will be used for dateTimeFieldSpec, but wanted to have them there to be consistent with how it is designed.

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
npawar: No, it's an enum consisting of only these values which are in the cases (DIMENSION, TIME, DATE_TIME, METRIC)

pinot-common/src/test/java/com/linkedin/pinot/common/data/FieldSpecTest.java
npawar: Refactored the tests to address the comments related to tests

pinot-common/src/test/java/com/linkedin/pinot/common/data/FieldSpecTest.java
npawar: Fixed

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
```diff
@@ -0,0 +1,222 @@
...
+
+  private static final String NUMBER_REGEX = "[0-9]+";
+  private static final String COLON_SEPARATOR = ":";
+
+  private String _format;
+  private String _granularity;
+  private DateTimeType _dateTimeType;
+
+  public enum DateTimeType {
+    PRIMARY,
```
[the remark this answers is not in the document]
npawar: Done

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
npawar: Tried to follow what was done in the rest of the setters of the schema, which is either a comment saying "// Required by JSON de-serializer. DO NOT REMOVE." or no comment/annotation at all. Not sure if  @VisibleForTesting  makes sense here. Will add comments/javadocs

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
```diff
@@ -0,0 +1,222 @@
...
+   *
+   * @param name
+   * @param dataType
+   * @param format - size:timeunit:timeformat eg: 1:MILLISECONDS:EPOCH, 5:MINUTES:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd
+   * @param granularity - size:timeunit eg: 5:MINUTES
+   * @param dateTimeType
+   * @param defaultNullValue
+   */
+  public DateTimeFieldSpec(@Nonnull String name, @Nonnull DataType dataType, @Nonnull String format,
+      @Nonnull String granularity, DateTimeType dateTimeType, Object defaultNullValue) {
```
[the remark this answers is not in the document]
npawar: Cleaned up constructors clutter, keeping only a single one which will be used immediately.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
npawar: @VisibleForTesting does not seem to disallow usage, even from IDE. Added javadoc, stating these should not be used.

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
npawar: Cleaned up addDateTime methods clutter, keeping just 1

pinot-common/src/main/java/com/linkedin/pinot/common/data/TimeGranularitySpec.java
npawar: I removed an import, the IDE seems to have added all the other line formatting.

npawar:
@mcvsubbu I have added the design doc on the wiki page in the Design Docs section: https://github.com/linkedin/pinot/wiki/Schema-TimeSpec-Refactoring
Addressed all your comments as well, please review again!

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
mcvsubbu: I was referring to the switch below, that checks for different dataTypes.

npawar: Yes, it can be string in case of more advanced simple date formats. And since it is an n units since epoch calculation, better to not restrict to whole numbers

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
```diff
@@ -0,0 +1,222 @@
...
+  private static final String GRANULARITY_TOKENS_ERROR_STR = "granularity must be of format size:timeunit";
+  private static final String GRANULARITY_PATTERN_ERROR_STR = "granularity must be of format [0-9]+:<TimeUnit>";
+  private static final String NUMBER_REGEX = "[1-9][0-9]*";
+  private static final String COLON_SEPARATOR = ":";
+
+  private String _format;
+  private String _granularity;
+  private DateTimeType _dateTimeType;
+
+  public enum DateTimeType {
```
kishoreg: I would like to have a method in Pinot that takes DateTimeSpec and a value and returns a DateTime object. This can be util or part of this class

pinot-common/src/main/java/com/linkedin/pinot/common/data/Schema.java
kishoreg: can we take DateTimeFieldSpec instead of individual each value. One reason why we had soo many methods for addTime was every time we enhanced TimeFieldSpec we added methods for every possible combination.

kishoreg:
with this change the incomingTimeGranularity will be removed from Schema but this is needed by the segment generation/realtime ingestion code. We need to make sure we add this in tableConfig.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
npawar: Will add this in another PR

PR #1846, opened 2017-08-28
[TE] adding rca link

justYves:
### What's new:

Adding RCA links to the events tab anomalies that have an `ingraph` link.
Previously, no rca links were shown.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/InvestigateView.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/InvestigateView.js
@@ -33,8 +33,8 @@ InvestigateView.prototype = {
     const anomaly = this.investigateModel.getAnomaly();
     const wowData = this.investigateModel.getWowData();
     const currentValue = wowData.currentVal;
-    const wowResults = this.formatWowResults(wowData.compareResults, anomaly);
-    const externalUrls = anomaly.externalUrl ? JSON.parse(anomaly.externalUrl) : null;
+    const externalUrls = anomaly.externalUrl ? JSON.parse(anomaly.externalUrl) : {};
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/resources/assets/javascript/views/InvestigateView.js
```diff
@@ -89,7 +90,7 @@ InvestigateView.prototype = {
             `heatMapBaselineEnd=${heatMapBaselineEnd.valueOf()}&filters=${anomalyFunctionDimension}&heatMapFilters=${anomalyFunctionDimension}`;
 
 
-        wow.newUrl = `app#/rca/${metricId}/dimensions?analysisStart=${anomalyRegionStart.valueOf()}&analysisEnd=${anomalyRegionEnd.valueOf()}&` +
+        wow.newUrl = `app#/rca/${metricId}/${tab}?analysisStart=${anomalyRegionStart.valueOf()}&analysisEnd=${anomalyRegionEnd.valueOf()}&` +
```
theseyi: Why the need to directly invoke the `valueOf` method?
If there's a special reason, might be helpful to add a comment here, otherwise, if it's just to unbox the variable / get the primitive value, i think having it in a template literal as you have, is sufficient hint for the js engine to automatically do the conversion

justYves: getting the valueOf is needed since we want all date time query params to be display in unix ms. Omitting it, doesn't consistently yield the expected result. Will add a comment. Thanks!

theseyi: Ah, it's a Date object. Yea, the comment should help since it wasn't obv to me

PR #1848, opened 2017-08-29
Retry segment download only on 5xx error codes.

mcvsubbu:
When a segment download fails with 4xx (esp 404) no amount of retry is going to help us get
the segment. We waste significant time (30s per segment) if we reach such a condition, unable to
report startup complete because the state transition is still being attempted.

Not to mention that a thread is held up sleeping for a long time.

Introduced a new PermanentDownloadException that callers can examine before attempting a retry.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpSegmentFetcher.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpSegmentFetcher.java
@@ -15,6 +15,8 @@
  */
 package com.linkedin.pinot.common.segment.fetcher;
 
+import com.linkedin.pinot.common.Utils;
+import com.linkedin.pinot.common.exception.PermanentDownloadException;
 import java.io.File;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpSegmentFetcher.java
```diff
@@ -50,6 +52,9 @@ public void fetchSegmentToLocal(String uri, File tempFile) throws Exception {
             "Downloaded file from {} to {}; Length of httpGetResponseContent: {}; Length of downloaded file: {}", uri,
             tempFile, httpGetResponseContentLength, tempFile.length());
         return;
+      } catch (PermanentDownloadException e) {
+        LOGGER.error("Failed to download file from {}", uri, e);
+        Utils.rethrowException(e);
```
antumbde: Need a metric to alert on missing segments ?

mcvsubbu: we have the segment status checker bumping metrics for error segments, so we should be ok

PR #1849, opened 2017-08-29
[TE] Distinguish missing data and zero

cyenjung:
Problem:
Previously, we could not distinguish zero value and missing data in time series. The reason is ThirdEye fills zero value to missing data points in order to not to crash the UI. However, anomaly detection needs to distinguish the difference because zero values could be true anomaly.

Solution:
We add a flag In the data storage, MetricTimeSeries, to indicate whether a data point is available. If the flag is false, then the data point is missing and user could specify the return value.

To compatible with old UI code, MetricTimeSeries returns 0 for missing data if the default return value is not specified.

Tests:
Tested with new unit tests and integration test is performed on a local detector.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/context/TimeSeries.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/context/TimeSeries.java
@@ -5,6 +5,9 @@ import java.util.List;
 import java.util.NavigableMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.commons.lang.builder.ToStringBuilder;
 import org.apache.commons.lang.builder.ToStringStyle;
 import org.joda.time.Interval;
@@ -21,12 +24,11 @@ public class TimeSeries implements MetricTimeSeries {
   private NavigableMap<Long, Double> timeSeries = new TreeMap<>();
 
   /**
-   * The max and min timestamp of this time series; both are inclusive.
+   * The max (exclusive) and min (inclusive) timestamp of this time series.
    */
   private Interval timeSeriesInterval = new Interval(0L, 0L);
 
   public TimeSeries() {
-    timeSeries = new TreeMap<>();
   }
 
   /**
@@ -34,14 +36,19 @@ public class TimeSeries implements MetricTimeSeries {
    */
   public TimeSeries(List<Long> timeStamps, List<Double> values)
       throws Exception {
+    if (CollectionUtils.isNotEmpty(timeStamps) && CollectionUtils.isNotEmpty(values)) {
+      if (timeStamps.size() != values.size()) {
+        throw new IOException("time stamps list and value list need to match in size!!");
+      }
 
-    timeSeries = new TreeMap<>();
-    if (timeStamps.size() != values.size()) {
-      throw new IOException("time stamps list and value list need to match in size!!");
-    }
+      for (int i = 0; i < timeStamps.size(); i++) {
+        timeSeries.put(timeStamps.get(i), values.get(i));
+      }
 
-    for (int i = 0; i < timeStamps.size(); i++) {
-      timeSeries.put(timeStamps.get(i), values.get(i));
+      long minTimestamp = timeSeries.firstEntry().getKey();
+      long maxTimestamp = timeSeries.lastEntry().getKey() + 1; // end time exclusive
+
+      timeSeriesInterval = new Interval(minTimestamp, maxTimestamp);
     }
   }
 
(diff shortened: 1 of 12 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/api/MetricTimeSeries.java
```diff
@@ -41,110 +45,108 @@ public MetricSchema getSchema() {
...
-    } else {
-      newValue = delta;
     }
+
     set(timeWindow, name, newValue);
   }
 
   public void aggregate(MetricTimeSeries series) {
-    for (long timeWindow : series.timeseries.keySet()) {
+    for (long timeWindow : series.metricsValue.keySet()) {
```
bl44: Is these aggregate metric actually used? It is a cross metric aggregation. Wondering where it could be used.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/api/MetricTimeSeries.java
```diff
@@ -206,16 +217,21 @@ public static MetricTimeSeries fromBytes(byte[] buf, MetricSchema schema) throws
...
           delim = ",";
         }
-        sb.append(delim).append(NumberUtils.readFromBuffer(buffer, schema.getMetricType(i)));
+        sb.append(delim);
+        Number number = getOrDefault(timeWindow, schema.getMetricName(i), null);
+        if (number != null) {
+          sb.append(number);
+        } else {
+          // TODO: read user specified null value from schema
+          sb.append("null");
```
bl44: define 'null' as a constant, for example, THIRDYE_NULL

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/api/MetricTimeSeries.java
```diff
@@ -265,9 +281,28 @@ public String toString() {
...
+      }
+    }
+
+    return result;
+  }
+
   @Override
   public int hashCode() {
-    return timeseries.keySet().hashCode() + 13 * schema.hashCode();
+    return metricsValue.keySet().hashCode() + 13 * schema.hashCode();
```
bl44: Is 13 a real 'magic' number?

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/anomalydetection/function/BackwardAnomalyFunctionUtilsTest.java
```diff
@@ -0,0 +1,80 @@
...
+      add(3.0); add(5.0);
+    }};
+    final TimeSeries timeSeries1 = new TimeSeries(timestamps1, values1);
+    timeSeries1.setTimeSeriesInterval(new Interval(3L, 6L));
+
+    List<Long> timestamps2 = new ArrayList<Long>() {{
+      add(1L); add(2L); add(3L);
+    }};
+    List<Double> values2 = new ArrayList<Double>() {{
+      add(1.0); add(2.0); add(3.0);
```
bl44: Nice way to construct new ArrayList.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/api/MetricTimeSeries.java
cyenjung: Yeah, it is a magic number, which is not a good idea to use it. Replacing it with Java7's Objects.hash().

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/context/TimeSeries.java
```diff
@@ -21,27 +24,31 @@
...
 
-    timeSeries = new TreeMap<>();
-    if (timeStamps.size() != values.size()) {
-      throw new IOException("time stamps list and value list need to match in size!!");
-    }
+      for (int i = 0; i < timeStamps.size(); i++) {
+        timeSeries.put(timeStamps.get(i), values.get(i));
+      }
+
+      long minTimestamp = timeSeries.firstEntry().getKey();
```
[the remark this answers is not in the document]
cyenjung: Yes, it is an ordered map and it is ensured by declaring it as a Navigable Map.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/context/TimeSeries.java
```diff
@@ -121,4 +128,25 @@ public void setTimeSeriesInterval(Interval timeSeriesInterval) {
...
+
+    TimeSeries that = (TimeSeries) o;
+
+    return new EqualsBuilder().append(timeSeries, that.timeSeries)
+        .append(getTimeSeriesInterval(), that.getTimeSeriesInterval()).isEquals();
+  }
+
+  @Override
+  public int hashCode() {
+    return new HashCodeBuilder(17, 37).append(timeSeries).append(getTimeSeriesInterval()).toHashCode();
```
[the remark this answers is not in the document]
cyenjung: Replacing it with Java7's Objects.hash().

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/api/MetricTimeSeries.java
```diff
@@ -41,110 +45,108 @@ public MetricSchema getSchema() {
...
+   * @param timeWindow the timestamp.
+   * @param name the metric name.
+   *
+   * @return the metric value if exists; otherwise, 0 is returned.
+   */
   public Number get(long timeWindow, String name) {
-    ByteBuffer buffer = timeseries.get(timeWindow);
-    Number ret = 0;
+    return getOrDefault(timeWindow, name, 0);
+  }
```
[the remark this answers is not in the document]
cyenjung: UI is still using this method. I probably will migrate those methods to getOrDefault(window, name, 0) and change get(window, name) to return null if no such number exists.

PR #1850, opened 2017-08-29
Fix incorrect read of history on restart.

mcvsubbu:
On restart, we read the existing file in the disk, and end up setting _isFull to false, and the cursor to 0
effectively dropping the history so far.

Fixed and added a unit test


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
```diff
@@ -323,4 +321,14 @@ private int getNumntriesToScan() {
     }
     return getCursor();
   }
+
+  public static void main(String[] args) throws Exception {
```
antumbde: can be removed?

mcvsubbu: I think it is useful for debugging. We can copy a history file from a server and look at it.

PR #1851, opened 2017-08-30
Fix the int overflow issue for large index

Jackie-Jiang:
Fix the issue where calculating bit offset for bit-encoded int causing int overflow


Jackie-Jiang:
@kishoreg I'm afraid adding such huge buffer will kill the tester. We caught it in the test cluster (scan-based comparator).

PR #1853, opened 2017-08-31
Re-implement all offline dictionary based on the fixed byte value reader

Jackie-Jiang:
Also fixed the OnHeapStringDictionary.indexOf() to return correct value when value does not exist (Should be the same as Arrays.binarySearch() to make range filter work properly)

Most performance gain comes from StringDictionary for the following reasons:
  1. No need to duplicate ByteBuffer, index byte-by-byte instead
  2. Reuse the byte buffer if possible
  3. When reading unpadded string, no need to read all bytes

Benchmark Results:
Old Dictionaries:
Benchmark                                     Mode  Cnt      Score      Error  Units
BenchmarkOfflineIndexReader.doubleDictionary  avgt    3   1131.025 ±   54.210  us/op
BenchmarkOfflineIndexReader.floatDictionary   avgt    3   1080.044 ±   16.400  us/op
BenchmarkOfflineIndexReader.intDictionary     avgt    3   1085.413 ±   61.710  us/op
BenchmarkOfflineIndexReader.longDictionary    avgt    3    901.671 ±    9.518  us/op
BenchmarkOfflineIndexReader.stringDictionary  avgt    3  16020.946 ± 1452.813  us/op

New Dictionaries:
Benchmark                                     Mode  Cnt     Score     Error  Units
BenchmarkOfflineIndexReader.doubleDictionary  avgt    3   985.736 ±   5.780  us/op
BenchmarkOfflineIndexReader.floatDictionary   avgt    3   939.546 ±  10.791  us/op
BenchmarkOfflineIndexReader.intDictionary     avgt    3   958.593 ±  33.669  us/op
BenchmarkOfflineIndexReader.longDictionary    avgt    3   769.179 ±  38.159  us/op
BenchmarkOfflineIndexReader.stringDictionary  avgt    3  8990.013 ± 256.303  us/op


pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/readers/OnHeapStringDictionary.java
```diff
@@ -15,162 +15,80 @@
...
-    _idToRawStringMap = new String[cardinality];
-    _stringToIdMap = new HashMap<>(cardinality);
-
-    for (int id = 0; id < cardinality; id++) {
-      byte[] bytes = dataFileReader.getBytes(id, 0);
-      String value = getStringFromBytes(bytes, true);
-      _idToStringMap[id] = value;
-      _idToRawStringMap[id] = getStringFromBytes(bytes, false);
-      _stringToIdMap.put(value, id);
+  private final byte _paddingByte;
```
snleee: Isn't this unnecessary? `_paddingByte` is already included in `ImmutableDictionaryReader`.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/readers/ImmutableDictionaryReader.java
```diff
@@ -15,137 +15,183 @@
...
+  }
 
-  @Override
-  public int length() {
-    return rows;
+  protected int binarySearch(String value) {
+    byte[] buffer = getBuffer();
+    int low = 0;
+    int high = _length - 1;
+    if (_paddingByte == 0) {
```
snleee: It seems that we use `0` for checking whether to use unpadded or padded. Can we use `public static final`?

pinot-core/src/main/java/com/linkedin/pinot/core/data/readers/RecordReaderUtils.java
```diff
@@ -15,70 +15,71 @@
...
     if ((tokens == null) || (tokens.length == 0)) {
-      value = new Object[1];
-      value[0] = getDefaultNullValue(dataType);
-
+      value = new Object[]{fieldSpec.getDefaultNullValue()};
     } else {
-      value = new Object[tokens.length];
-      for (int i = 0; i < tokens.length; ++ i) {
-        value[i] = convertToDataType(tokens[i], dataType);
+      int length = tokens.length;
```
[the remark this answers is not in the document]
Jackie-Jiang: tokens might be null, and this will throw NPE if we move it up

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/readers/ImmutableDictionaryReader.java
Jackie-Jiang: The final goal is to use only 0 for padding, and remove all checks. Will keep it like this for now because cannot find a good static name for this 0 :P

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/readers/OnHeapStringDictionary.java
Jackie-Jiang: We need this to decide whether to binary search using unpadded or padded strings

PR #1856, opened 2017-09-01
Adding an API to validate table config

snleee:
A Table config that we use to create a table and the one we get from api are
different because we fill some default values. This api will convert the given
table config in the same format that we see from `GET /tables/{tableName}`.


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -289,6 +289,27 @@ public SuccessResponse updateTableConfig(
     return new SuccessResponse("Table config updated for " + tableName);
   }
 
+  @POST
+  @Path("/tables/preview")
+  @Produces(MediaType.APPLICATION_JSON)
+  @ApiOperation(value = "Preview table config for a table", notes = "Preview table config for a table")
```
mayankshriv: Should this be named something more generic as validate as opposed to preview? Also please add more documentation, as well as description on the usage/purpose of this api.

PR #1858, opened 2017-09-01
Remove repo directory from Pinot distribution

jfim:
Remove the repo directory from the Pinot distribution, which contains a
copy of the lib folder. This roughly halves the size of the Pinot
distribution and slightly reduces build time.


jfim:
Tested running the quick start and pinot-admin tools.

Jackie-Jiang:
LGTM

PR #1859, opened 2017-09-05
Fix SizeBasedDocIdSet

Jackie-Jiang:
Fix the issue where the argument passed into SizeBasedDocIdSet is totalDocs instead of maxDocId

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/SizeBasedDocIdIterator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/SizeBasedDocIdIterator.java
@@ -18,38 +18,36 @@ package com.linkedin.pinot.core.operator.dociditerators;
 import com.linkedin.pinot.core.common.BlockDocIdIterator;
 import com.linkedin.pinot.core.common.Constants;
 
+
 public final class SizeBasedDocIdIterator implements BlockDocIdIterator {
-  int counter = 0;
-  private int maxDocId;
+  private final int _maxDocId;
+  private int _currentDocId = -1;
 
   public SizeBasedDocIdIterator(int maxDocId) {
-    this(0, maxDocId);
-  }
-
-  public SizeBasedDocIdIterator(int minDocId, int maxDocId) {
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/SizeBasedDocIdIterator.java
```diff
@@ -18,38 +18,31 @@
 import com.linkedin.pinot.core.common.BlockDocIdIterator;
 import com.linkedin.pinot.core.common.Constants;
 
+
 public final class SizeBasedDocIdIterator implements BlockDocIdIterator {
```
kishoreg: add java docs explaining the behaviour, is it inclusive or exclusive

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docidsets/SizeBasedDocIdSet.java
```diff
@@ -39,12 +38,10 @@ public int getMaxDocId() {
 
   @Override
   public void setStartDocId(int startDocId) {
-    throw new UnsupportedOperationException();
```
kishoreg: why did you remove it, its ok for docIdSets to be immutable, i.e the min and max doc id cannot be changed later.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/SizeBasedDocIdIterator.java
Jackie-Jiang: Will add java doc into the BlockDocIdIterator for the behavior in another PR.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docidsets/SizeBasedDocIdSet.java
Jackie-Jiang: They should not be changed, but calling it should not throw exception. That's why I leave it blank (no-op)

PR #1861, opened 2017-09-05
Add connect and read timeouts for kafka schema registry fetch so that we fail sooner if kafka fails

mcvsubbu:


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaAvroMessageDecoder.java
```diff
@@ -142,15 +144,22 @@ private String hex(byte[] bytes) {
...
+        conn.setReadTimeout(15000);
+        LOGGER.info("Fetching schema using url {}", url.toString());
 
-        reader = new BufferedReader(new InputStreamReader(url.openStream(), "UTF-8"));
         StringBuilder queryResp = new StringBuilder();
-        for (String respLine; (respLine = reader.readLine()) != null; ) {
-          queryResp.append(respLine);
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(conn.getInputStream(), "UTF-8"))) {
+          String line;
+          while ((line = reader.readLine()) != null) {
```
jfim: Avoid inner assignments in loop clauses, they're pretty bad style. See https://stackoverflow.com/questions/11222121/is-it-bad-to-do-inner-assignments for example.

PR #1864, opened 2017-09-06
TE: fixes related to alert subscriber group

puneetjaiswal:
- Adding end point to get list of sub. group given and alertId (anomaly function id)
- Updating sender's email when missing in sub.group.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EntityManagerResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EntityManagerResource.java
@@ -2,6 +2,7 @@ package com.linkedin.thirdeye.dashboard.resources;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
+import com.linkedin.thirdeye.common.ThirdEyeConfiguration;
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EntityManagerResource.java
```diff
@@ -182,6 +185,9 @@ public Response updateEntity(@QueryParam("entityType") String entityTypeStr, Str
         break;
       case ALERT_CONFIG:
         AlertConfigDTO alertConfigDTO = OBJECT_MAPPER.readValue(jsonPayload, AlertConfigDTO.class);
+        if (Strings.isNullOrEmpty(alertConfigDTO.getFromAddress())) {
+          alertConfigDTO.setFromAddress(config.getFailureToAddress());
```
cecilynie: Do we want to use "getFromAddress()" to setFromAddress()? It can be a little bit confusing here we are using a "toAddr" to set "fromAddr". Same thing for EmailResource line 65.

puneetjaiswal: thats toAddr from thirdeyeConfig which is te-dev address. Not same as subsGroup's toAddr.

PR #1866, opened 2017-09-07
[TE] conditional heatmap formatting for "0" as "0.0"

apucher:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/views/heatmap/HeatMapCell.java
```diff
@@ -113,6 +113,8 @@ public HeatMapCell(String dimensionValue, double baselineValue, double currentVa
 
   // NOTE: hack for display of small numbers
   public static String format(Double number) {
+    if (number == 0.0d)
```
kishoreg: this may not work, you might want to compare with epsilon or use BigDecimal https://stackoverflow.com/questions/9992006/equals-operator-for-zeros-bigdecimal-double-in-java

apucher: Works as intended. Only programmatic 0d is formatted as such - everything else with exponent and mantissa.

PR #1868, opened 2017-09-07
Optimize OrDocIdIterator

Jackie-Jiang:
Currently OrDocIdIterator is based on PriorityQueue, and the overhead of using PriorityQueue make it very slow
Re-implemented the OrDocIdIterator based on array (loop over all iterators inside)
Fixed the bug in ArrayBasedDocIdIterator where mixing call to advance() and next() will return duplicate doc id.
Verified that even with 50 iterators, it's still way faster than the old implementation

Benchmark                              Mode  Cnt      Score      Error  Units
BenchmarkOrDocIdIterator.arrayBased2   avgt    3   1742.943 ±   17.665  us/op
BenchmarkOrDocIdIterator.arrayBased5   avgt    3   2845.901 ±  139.131  us/op
BenchmarkOrDocIdIterator.arrayBased10  avgt    3   4473.418 ±  124.593  us/op
BenchmarkOrDocIdIterator.arrayBased20  avgt    3   8050.464 ±  140.682  us/op
BenchmarkOrDocIdIterator.arrayBased50  avgt    3  14618.601 ±  147.558  us/op
BenchmarkOrDocIdIterator.heapBased2    avgt    3  14534.728 ±  532.522  us/op
BenchmarkOrDocIdIterator.heapBased5    avgt    3  21284.697 ±  692.490  us/op
BenchmarkOrDocIdIterator.heapBased10   avgt    3  34858.201 ± 2341.998  us/op
BenchmarkOrDocIdIterator.heapBased20   avgt    3  54647.907 ± 2132.005  us/op
BenchmarkOrDocIdIterator.heapBased50   avgt    3  82754.187 ± 3432.884  us/op

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/OrDocIdIterator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/OrDocIdIterator.java
@@ -15,139 +15,138 @@
  */
 package com.linkedin.pinot.core.operator.dociditerators;
 
-import java.util.Iterator;
-import java.util.PriorityQueue;
-import java.util.concurrent.atomic.AtomicLong;
-
-import com.linkedin.pinot.common.utils.Pairs;
-import com.linkedin.pinot.common.utils.Pairs.IntPair;
 import com.linkedin.pinot.core.common.BlockDocIdIterator;
 import com.linkedin.pinot.core.common.Constants;
 
-public final class OrDocIdIterator implements BlockDocIdIterator {
-  private BlockDocIdIterator[] docIdIterators;
 
-  final PriorityQueue<IntPair> queue;
-  final boolean[] iteratorIsInQueue;
-  int currentDocId = -1;
-  private int minDocId;
-  private int maxDocId;
-  private AtomicLong timeMeasure = new AtomicLong(0);
-  IntPair[] pointers;
+public final class OrDocIdIterator implements BlockDocIdIterator {
+  private final BlockDocIdIterator[] _docIdIterators;
+  private final int[] _nextDocIds;
+  private final int _minDocId;
+  private final int _maxDocId;
+
+  private int _numNotExhaustedIterators;
+  private int _currentDocId = -1;
+
+  public OrDocIdIterator(BlockDocIdIterator[] docIdIterators, int minDocId, int maxDocId) {
+    _docIdIterators = docIdIterators;
+    _numNotExhaustedIterators = docIdIterators.length;
+    _nextDocIds = new int[_numNotExhaustedIterators];
+    for (int i = 0; i < _numNotExhaustedIterators; i++) {
+      _nextDocIds[i] = docIdIterators[i].advance(minDocId);
+    }
+    _minDocId = minDocId;
+    _maxDocId = maxDocId;
+    removeExhaustedIterators();
+  }
 
   /**
-   * @param docIdIterators
+   * Loop over the document id array and pick the smallest one so that the document ids returned are in ascending order.
+   * <p>For each child iterator, fetch the next document id when its current document id is the same as the current
+   * document id of parent iterator so that each document id is only returned once.
+   * <p>{@inheritDoc}
    */
-  public OrDocIdIterator(BlockDocIdIterator[] docIdIterators) {
-    this.docIdIterators = docIdIterators;
-    queue =
-        new PriorityQueue<IntPair>(docIdIterators.length, new Pairs.AscendingIntPairComparator());
-    iteratorIsInQueue = new boolean[docIdIterators.length];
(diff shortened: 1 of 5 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/OrDocIdIterator.java
```diff
@@ -15,139 +15,119 @@
...
     }
+    _minDocId = minDocId;
+    _maxDocId = maxDocId;
+    removeExhaustedIterators();
   }
 
   @Override
-  public int advance(int targetDocId) {
-    if (currentDocId == Constants.EOF) {
+  public int next() {
```
kishoreg: I got the logic by reading the code, it will be good to document the logic on top of the method.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/OrDocIdIterator.java
```diff
@@ -15,139 +15,119 @@
...
+    return _currentDocId;
   }
 
   @Override
-  public int next() {
-    long start = System.currentTimeMillis();
-
-    if (currentDocId == Constants.EOF) {
-      return currentDocId;
+  public int advance(int targetDocId) {
```
kishoreg: what is the contract of advance, find matching element that is greater than or equal to targetDocId or greater than targetDocId

pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/OrDocIdIterator.java
```diff
@@ -15,139 +15,119 @@
...
-    // if (currentDocId == Constants.EOF) {
-    // LOGGER.debug("OR took:" + timeMeasure.get());
-    // }
+    return _currentDocId;
+  }
 
-    return currentDocId;
+  private void removeExhaustedIterators() {
+    int i = 0;
+    while (i < _numNotExhaustedIterators) {
```
kishoreg: this probably wont work if two iterators exhaust in one round right? lets say the first and last iterator exhaust, we will copy the last entry to the first position in the array but that would  be pointing to EOF.

Another case is if the last iterator exhausts, we copy the same value to the same position.

see if you can you simplify the logic by maintaining a boolean array for iteratorState.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docidsets/OrBlockDocIdSet.java
```diff
@@ -22,129 +22,102 @@
...
-  }
-
-  @Override
-  public long getNumEntriesScannedInFilter() {
-    long numEntriesScannedInFilter = 0L;
-    for (FilterBlockDocIdSet docIdSet : docIdSets) {
-      numEntriesScannedInFilter += docIdSet.getNumEntriesScannedInFilter();
-    }
-    return numEntriesScannedInFilter;
+    throw new UnsupportedOperationException();
```
kishoreg: why is this throwing exception. How will the parent operator do bitmap intersection if we dont return raw bitmap

kishoreg:
Nicely done! Few comments. Also, just for my clarity which part was slow in the previous code. Hard to believe that priority queue has so much overhead compared to iterating over the entire array. (probably the overhead of re-heapify on every addition is the root cause).

pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/OrDocIdIterator.java
Jackie-Jiang: Added

pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/OrDocIdIterator.java
Jackie-Jiang: Yes, will add it into the interface

pinot-core/src/main/java/com/linkedin/pinot/core/operator/dociditerators/OrDocIdIterator.java
Jackie-Jiang: It works. Will add comment to explain

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docidsets/OrBlockDocIdSet.java
Jackie-Jiang: Only BitmapDocIdSet and SortedDocIdSet should implement this method

Jackie-Jiang:
@kishoreg Addressed the comments.

PR #1874, opened 2017-09-11
Add metrics for keeping track of off-heap allocation for consuming segments

mcvsubbu:


pinot-core/src/test/java/com/linkedin/pinot/core/realtime/RealtimeSegmentTest.java
```diff
@@ -97,7 +97,8 @@ public static void before() throws Exception {
     when(segmentDataManager.getInvertedIndexColumns()).thenReturn(invertedIdxCols);
     when(segmentDataManager.getNoDictionaryColumns()).thenReturn(new ArrayList<String>());
     when(segmentDataManager.getSchema()).thenReturn(schema);
-    when(segmentDataManager.getMemoryManager()).thenReturn(new DirectMemoryManager("noSegment"));
+    when(segmentDataManager.getMemoryManager()).thenReturn(new DirectMemoryManager("noSegment",
```
mayankshriv: Do tests really need to send a ServerMetrics object? Would it be better to just pass null, or have another constructor without this parameter, so that clients that do need metrics (eg test) don't need to worry about it?

mcvsubbu: A toss-up. I could special-case it inside the code for tests, like I did for segment name.  agreed i could have been consistent with segment name and metrics (the metrics case came in automagically with IDE :-). I can undo the test changes and special-case it if metrics object is null.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/DirectMemoryManager.java
```diff
@@ -26,8 +29,13 @@
...
    * @see RealtimeIndexOffHeapMemoryManager
    */
+  public DirectMemoryManager(final String segmentName, ServerMetrics serverMetrics) {
+    super(serverMetrics, segmentName);
+  }
+
+  @VisibleForTesting
   public DirectMemoryManager(final String segmentName) {
-    super(segmentName);
+    this(segmentName, new ServerMetrics(new MetricsRegistry()));
```
mayankshriv: The server metrics could be null for clients that don't care about it, but that means you will have null check against it. Up to you.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/RealtimeIndexOffHeapMemoryManager.java
```diff
@@ -63,6 +81,7 @@ public PinotDataBuffer allocate(long size, String columnName) {
     PinotDataBuffer buffer = allocateInternal(size, columnName);
     _totalMemBytes += size;
     _buffers.add(buffer);
+    _serverMetrics.addValueToTableGauge(_tableName, ServerGauge.REALTIME_OFFHEAP_MEMORY_USED, size);
```
antumbde: For gauge, we should report total value at this instance which is totalMemBytes. size variable only represents incremental change.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/RealtimeIndexOffHeapMemoryManager.java
```diff
@@ -84,6 +103,7 @@ public PinotDataBuffer allocate(long size, String columnName) {
   public void close() throws IOException {
     for (PinotDataBuffer buffer : _buffers) {
       buffer.close();
+      _serverMetrics.addValueToTableGauge(_tableName, ServerGauge.REALTIME_OFFHEAP_MEMORY_USED, -buffer.size());
```
antumbde: we should report full value (totalMemBytes - buffer.size() ? )

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/RealtimeIndexOffHeapMemoryManager.java
mcvsubbu: we will get a per-table value at the current level. We are adding value to the table guage on each alloc. Alternatively, we can set total MBytes, yes. The result is the same.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/RealtimeIndexOffHeapMemoryManager.java
mcvsubbu: We are subtracting he buffer size here, but yes, alternatively we can subtract totalMemBytes from the guage.

PR #1878, opened 2017-09-11
[TE] setting default for sensitivity

justYves:
### What's new:
- Setting default for sensitivity per follow rules: 
  - For minutely metrics, the default sensitivity should be "MEDIUM"
  - For daily and hourly metrics, the default sensitivity should be "HIGH"

### Test: 
Tested locally with sandbox back end

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
@@ -103,6 +103,29 @@ export default Ember.Controller.extend({
   },
 
   /**
+   * Setting default sensitivity if selectedSensitivity is undefined
+   * @returns {String}
+   */
+  sensitivityWithDefault: Ember.computed(
+    'selectedSensitivity',
+    'selectedGranularity',
+    function() {
+      let {
+       selectedSensitivity,
+       selectedGranularity
+      } =  this.getProperties('selectedSensitivity', 'selectedGranularity');
+
+      if (!selectedSensitivity) {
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
```diff
@@ -103,6 +103,29 @@ export default Ember.Controller.extend({
...
+    'selectedSensitivity', 
+    'selectedGranularity',
+    function() {
+      let {
+       selectedSensitivity, 
+       selectedGranularity
+      } =  this.getProperties('selectedSensitivity', 'selectedGranularity');
+
+      if (!selectedSensitivity) { 
+        selectedSensitivity = (selectedGranularity === '5_MINUTES')
```
cecilynie: The selected granularity for minutely metrics can be '5_MINUTES' or 'MINUTES' or '1_MINUTES', is there a better way that we can represent minutely metrics' granularities? Or here we can use "HOURS" or "DAYS" to represent the other two cases and set default value to be "Sensitive"?

justYves: Good point. will do the reverse per your suggestion. 
Btw, any reason why this is not consistent ?

cecilynie: It depends on the data source. For ingraph it can be either 1_MINUTE or 5_MINUTE, for sitespeed metrics I think they from a different data source (harrier) and being imported to pinot. Their granularity are being imported as "MINUTE". Probably needs to unify that.

PR #1879, opened 2017-09-12
[TE] changed url and script

justYves:
### What's new: 
- changed url from `/thirdeye#investigate?anomalyId=` (old ui) to `app/#/screenshot` 
- added getGraphPnj for reference only

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/resources/scripts/getGraphPnj.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/scripts/getGraphPnj.js
@@ -1,8 +1,11 @@
+// screenshot script for reference only 
+// actual used script lives in /thirdeye-configs/prod-config
 "use strict";
 
 var page = require('webpage').create();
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/resources/scripts/getGraphPnj.js
```diff
@@ -63,29 +42,58 @@ page.open(options.anomalyRoute, function(status) {
   }
 });
 
+page.onCallback = function(data) {
+  console.log(data);
```
newsummit: Do you want these console logs to remain?

justYves: the script also has a bunch of `system.stderr.writeLine` which acts like console.log. This is pretty useful for debugging purposes as phantomJS is a headless browser. 
Let me know if you disagree!

PR #1884, opened 2017-09-13
Make number of segments that can be refreshed simultaneously

mcvsubbu:
We introduced a lock on the helix threads refreshing a loaded segment, so that we do not run out of memory
with too many segments loaded. However, if segments are loaded via MMAP and the server is in a multi-tenant
(low qps) cluster, it should be fine to overload via MMAP for a brief period.

This commit makes the number of simultaneous threads configurable. A value of 0 (or less) indicates unlimited
number. Any value above 0 configures the maximum allowed threads to refresh.

Default is 1, preserving the old behvaior of a lock.

NOTE: While setting this number to a higher value (or 0) may increase segment load time, there will be more
      threads occupied loading segments, so query latencies may actually increase during refresh time

### DIFF
--- a/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
+++ b/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
@@ -15,8 +15,10 @@
  */
 package com.linkedin.pinot.server.starter.helix;
 
+import com.linkedin.pinot.common.data.DataManager;
 import com.linkedin.pinot.common.messages.SegmentRefreshMessage;
 import com.linkedin.pinot.common.messages.SegmentReloadMessage;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import org.apache.helix.NotificationContext;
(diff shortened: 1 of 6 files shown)


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
```diff
@@ -34,11 +36,36 @@
   // We only allow one segment refresh/reload happens at the same time
   // The reason for that is segment refresh/reload will temporarily use double-sized memory
   private final Lock _lock = new ReentrantLock();
+  private final Semaphore _refreshThreadSemaphore;
 
   private final SegmentFetcherAndLoader _fetcherAndLoader;
 
-  public SegmentMessageHandlerFactory(SegmentFetcherAndLoader fetcherAndLoader) {
+  public SegmentMessageHandlerFactory(SegmentFetcherAndLoader fetcherAndLoader, DataManager dataManager) {
```
mayankshriv: Does this class need dataManager, or can we just pass maxParallelRefreshThreads to the constructor?

mcvsubbu: The class does not need DataManager. However, I prefer passing DataManager in, so other configs can be retrieved if we need to do so. Otherwise, the number of arguments keep increasing over time, a la RealtimeSegmentImpl (that I re-factored a few weeks back to reduce the number of args)

PR #1888, opened 2017-09-13
[TE] Te auth ui fix

justYves:
### What's new: 
- Fixing front end redirection (from anomalies page)

### TEST:
tested locally


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auth/ThirdeyeAuthenticationManager.java
```diff
@@ -17,7 +17,7 @@
 public class ThirdeyeAuthenticationManager implements IAuthManager, Authenticator<AuthRequest, PrincipalAuthContext> {
   private static final String LDAP_CONTEXT_FACTORY = "com.sun.jndi.ldap.LdapCtxFactory";
   private final AuthConfiguration authConfiguration;
-  private final ThreadLocal<PrincipalAuthContext> principalAuthContextThreadLocal;
+  private volatile ThreadLocal<PrincipalAuthContext> principalAuthContextThreadLocal;
```
apucher: why volatile? this seems like a bug

puneetjaiswal: It can be `final static`

PR #1890, opened 2017-09-13
Tool for backfilling datetime column in pinot segments

npawar:
Adding a tool in pinot admin, for the backfill of pinot segments with dateTime column. 
The tool expects table name, timeSpec and dateTimeSpec as inputs, and will go over all pinot segments to backfill them with dateTime column, with a value corresponding to the time column value. 
Updated segment metadata, to include information about date time columns.
If the segment consists of star tree, only raw docs are read, and star tree is generated by treating date time columns as dimensions and skipping them from default split order

Testing done:

Segments used:
1) Download a single segment, and backup and backfill it with date time
2) Backup and backfill of all segments of a table
3) Error generated if segment name provided does not exist

Segment configs:
1) Backfilled an existing segment - ensured that segment is queryable with new datetime
2) Backfilled an exsiting segment, with star tree - ensured that star tree was generated, and segment is queryable with new datetime
3) Backfilled an existing segment, with same dateTime column already present - ensured that segment is queryable with new datetime
4) Backfilled an existing segment, with same dateTime column already present, but differnet datetime column configs - ensured that segment is queryable with new datetime

Segment generation: 
1) Generated segments from AVRO, JSON, PINOT
2) Generated segments from AVRO, with star tree config
3) Generated segments from AVRO, with dateTime in schema
4) Generated segments from AVRO, with dateTime in schema, and star tree enabled
Verified that these segments upload correctly, and can provide new datetime whenever available. The older segments and backfilled segments are all queryable together.

Query:
Queried for sum of metrics on backfilled segments, vs non backfilled segments, and verified the totals match

End to end:
1) Downloaded segment, backfilled, uploaded. Queried on segments to check for new column.

After this change is pushed, segments can be generated with the new datetime column, by specifying a dateTimeFieldSpec in the segment generation schema
All older segments can be backfilled using the tool.


mcvsubbu:
I will review it, but I have a question. What happens if the segment is refreshed between the time it is downloaded and re-uploaded (after conversion)?
Perhaps you should introduce a refresh-if-crc-equals API in the controller?

kishoreg:
This tool is doing the following
1. Download the segment, 
2. <do something>
3. upload the segment again.
Conceptually this is very similar to what a Pinot Minion does.

It will be great if we can model this as an ad-hoc or standalone Minion tool. I haven't looked into the interface of a Task in minion but if we can use the same interface to implement this task of adding date time column then we can leverage this tool to run any minion task.

@Jackie-Jiang @npawar what do you think?

npawar:
@kishoreg As discussed offline, there are currently multiple constraints to model this as a minion task and run it via a tool that runs adhoc minion tasks. The task executor uses minionContext which is created using cluster configs on startup. We will need to duplicate minion setup steps to make this work. There's also some configs needed by pinotTaskConfigs which we will have to manually construct.
As a result, keeping it independent of the minion as a separate tool seems like the best option right now.

pinot-core/src/main/java/com/linkedin/pinot/core/minion/BackfillDateTimeColumn.java
```diff
@@ -0,0 +1,258 @@
...
+      // create schema with updated dateTimeFieldSpec
+      else if (!_dateTimeFieldSpec.equals(schema.getDateTimeSpec(_dateTimeFieldSpec.getName()))) {
+        Schema newSchema = new Schema();
+        for (DimensionFieldSpec dimensionFieldSpec : schema.getDimensionFieldSpecs()) {
+          newSchema.addField(dimensionFieldSpec);
+        }
+        for (MetricFieldSpec metricFieldSpec : schema.getMetricFieldSpecs()) {
+          newSchema.addField(metricFieldSpec);
+        }
+        newSchema.addField(schema.getTimeFieldSpec());
```
mcvsubbu: Dont u have to re-add the existing DateTimeFieldSpec (that do not match the name of the new datetime field), if any?

npawar: Yes looks like I made an incorrect assumption that there will be only 1. Fixed this.

pinot-core/src/test/java/com/linkedin/pinot/core/data/readers/BackfillDateTimeRecordReaderTest.java
```diff
@@ -0,0 +1,250 @@
...
+    inputSchema = createPinotSchemaWithTimeSpec(timeFieldSpec, dateTimeFieldSpec);
+    inputRecordReader = new TestRecordReader(inputData, inputSchema);
+    dateTimeFieldSpec = new DateTimeFieldSpec("timestampInEpoch", DataType.LONG, "1:MILLISECONDS:EPOCH", "1:HOURS", DateTimeType.PRIMARY);
+    wrapperSchema = createPinotSchemaWithTimeSpec(timeFieldSpec, dateTimeFieldSpec);
+    entries.add(new Object[] {
+        inputRecordReader, timeFieldSpec, dateTimeFieldSpec, wrapperSchema
+    });
+
+    return entries.toArray(new Object[entries.size()][]);
+  }
```
mcvsubbu: Do you want to add a test with multiple datetimefields?

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackfillDateTimeColumnCommand.java
```diff
@@ -0,0 +1,232 @@
...
+        if (allSegmentNames.contains(segmentName)) {
+          segmentNames.add(segmentName);
+        } else {
+          throw new RuntimeException("Segment with name " + segmentName + " does not exist.");
+        }
+      }
+    }
+
+    if (_backupDir == null) {
+      _backupDir = TMP_DIR;
```
mcvsubbu: Not sure where TMP_DIR is coming from, but if backupDir is not defined, maybe you should just not back things up? TMP_DIR may fill up /tmp (if that is the backup dir) and cause a system to become unusable.

Or, maybe set it to /var/tmp which tends to hold some more data

pinot-core/src/main/java/com/linkedin/pinot/core/minion/BackfillDateTimeColumn.java
```diff
@@ -0,0 +1,263 @@
...
+          .getSkipMaterializationCardinality());
+      starTreeIndexSpec.setSkipMaterializationForDimensions(Sets.newHashSet(starTreeMetadata
+          .getSkipMaterializationForDimensions()));
+      starTreeIndexSpec.setSkipStarNodeCreationForDimensions(Sets.newHashSet(starTreeMetadata
+          .getSkipStarNodeCreationForDimensions()));
+      config.setStarTreeIndexSpec(starTreeIndexSpec);
+    }
+
+    LOGGER.info("Creating segment for {} with config {}", segmentName, config.toString());
+    SegmentIndexCreationDriverImpl driver = new SegmentIndexCreationDriverImpl();
```
mcvsubbu: It may be useful to set up SegmentIndexCreationDriverImpl (via config) so that it can log a message every (say N) rows. I understand this is not something you did, but maybe it is useful to add that information. You decide.

pinot-core/src/main/java/com/linkedin/pinot/core/minion/BackfillDateTimeColumn.java
```diff
@@ -0,0 +1,263 @@
...
+
+import javax.annotation.Nonnull;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The <code>BackfillDateTimeColumn</code> class takes a segment, a timeSpec from the segment, and a
+ * dateTimeSpec.
+ * It creates a column corresponding to the dateTimeSpec configs, using the values from the timeSpec
```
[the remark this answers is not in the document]
npawar: Changed

pinot-core/src/main/java/com/linkedin/pinot/core/minion/BackfillDateTimeColumn.java
```diff
@@ -0,0 +1,263 @@
...
+  private final TimeFieldSpec _srcTimeFieldSpec;
+  private final DateTimeFieldSpec _destDateTimeFieldSpec;
+
+  public BackfillDateTimeColumn(@Nonnull File originalIndexDir, @Nonnull File backfilledIndexDir,
+      @Nonnull TimeFieldSpec srcTimeSpec, @Nonnull DateTimeFieldSpec destDateTimeSpec)
+      throws Exception {
+    _originalIndexDir = originalIndexDir;
+    _backfilledIndexDir = backfilledIndexDir;
+    _srcTimeFieldSpec = srcTimeSpec;
+    _destDateTimeFieldSpec = destDateTimeSpec;
```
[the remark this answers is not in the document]
npawar: Added a check for this

pinot-core/src/main/java/com/linkedin/pinot/core/minion/BackfillDateTimeColumn.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /**
+     * Converts the time column value from timeFieldSpec to dateTimeFieldSpec
+     * @param timeColumnValue - time column value from timeFieldSpec
+     * @return
+     */
+    private Object convertTimeFieldToDateTimeFieldSpec(Object timeColumnValue) {
+
+      TimeGranularitySpec timeGranularitySpec = _timeFieldSpec.getIncomingGranularitySpec();
```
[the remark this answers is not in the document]
npawar: Changed this to look for outgoing first, and use incoming if outgoing not found

pinot-core/src/test/java/com/linkedin/pinot/core/data/readers/BackfillDateTimeRecordReaderTest.java
npawar: Added, thanks for catching this

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackfillDateTimeColumnCommand.java
npawar: Made backup dir mandatory. I prefer having a compulsory backup, and yes better to let user decide where to backup

pinot-core/src/main/java/com/linkedin/pinot/core/minion/BackfillDateTimeColumn.java
```diff
@@ -240,7 +243,10 @@ public GenericRow next(GenericRow row) {
      */
     private Object convertTimeFieldToDateTimeFieldSpec(Object timeColumnValue) {
 
-      TimeGranularitySpec timeGranularitySpec = _timeFieldSpec.getIncomingGranularitySpec();
+      TimeGranularitySpec timeGranularitySpec = _timeFieldSpec.getOutgoingGranularitySpec();
+      if (timeGranularitySpec == null) {
```
mcvsubbu: getOutgoingGranularitySpec already does this null check and returns incoming granularity spec

npawar: Ah nice, removed my redundant check

PR #1902, opened 2017-09-16
[PINOT-5314] Fix leak in OffHeapBitmapInvertedIndexCreator

mcvsubbu:
This class makes a bunch of calls to allocate direct memory in its constructor. If second or subsequent
calls fail to allocate memory then previously allocated memory is not released.

Added a try/catch and released memory in the constructor before re-throwing the exception.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -197,9 +197,13 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
         continue;
       }
 
-      OffHeapBitmapInvertedIndexCreator invertedIndexCreator =
-          new OffHeapBitmapInvertedIndexCreator(file, cardinality, totalDocs, totalNumberOfEntries, fieldSpec);
-      invertedIndexCreatorMap.put(column, invertedIndexCreator);
+      try {
+        OffHeapBitmapInvertedIndexCreator invertedIndexCreator =
+            new OffHeapBitmapInvertedIndexCreator(file, cardinality, totalDocs, totalNumberOfEntries, fieldSpec);
+        invertedIndexCreatorMap.put(column, invertedIndexCreator);
+      } catch (Exception e) {
+        LOGGER.warn("Inverted index not created for column {}", column, e);
+      }
     }
   }
 
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.core.segment.creator.impl.inv;
 
+import com.linkedin.pinot.common.Utils;
 import java.io.BufferedOutputStream;
 import java.io.DataOutputStream;
 import java.io.File;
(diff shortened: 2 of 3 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -238,26 +244,29 @@ public void seal() throws IOException {
...
+    MmapUtils.unloadByteBuffer(origValueBuffer);
+    origValueBuffer = null; valueBuffer = null;
+    MmapUtils.unloadByteBuffer(origLengths);
+    origLengths = null; lengths = null;
+    MmapUtils.unloadByteBuffer(origPostingListBuffer);
+    origPostingListBuffer = null; postingListBuffer = null;
+    MmapUtils.unloadByteBuffer(origPostingListCurrentOffsets);
+    origPostingListCurrentOffsets = null; postingListCurrentOffsets = null;
+    MmapUtils.unloadByteBuffer(origPostingListLengths);
+    origPostingListLengths = null; postingListLengths = null;
```
mayankshriv: What happens to the unloadByteBuffer call if one of the previous allocations failed and threw exceptions?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -197,9 +197,13 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
         continue;
       }
 
-      OffHeapBitmapInvertedIndexCreator invertedIndexCreator =
-          new OffHeapBitmapInvertedIndexCreator(file, cardinality, totalDocs, totalNumberOfEntries, fieldSpec);
-      invertedIndexCreatorMap.put(column, invertedIndexCreator);
+      try {
```
Jackie-Jiang: Why do we need this try catch and swallow the exception?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -117,31 +118,36 @@ public OffHeapBitmapInvertedIndexCreator(File indexDir, int cardinality, int num
...
+          "posting list lengths buffer to bitmap index for " + columnName);
+      postingListLengths = origPostingListLengths.asIntBuffer();
+      origPostingListStartOffsets = MmapUtils.allocateDirectByteBuffer(cardinality * INT_SIZE, null,
+          "posting list start offsets buffer to bitmap index for " + columnName);
+      postingListStartOffsets = origPostingListStartOffsets.asIntBuffer();
+      origPostingListCurrentOffsets = MmapUtils.allocateDirectByteBuffer(cardinality * INT_SIZE, null,
+          "posting list current offsets buffer to bitmap index for " + columnName);
+      postingListCurrentOffsets = origPostingListCurrentOffsets.asIntBuffer();
+    } catch (Exception e) {
+      releaseResources();
```
Jackie-Jiang: Release resource only if encountered exception?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
```diff
@@ -68,7 +69,11 @@ public void createInvertedIndices()
     Set<String> invertedIndexColumns = getInvertedIndexColumns();
 
     for (String column : invertedIndexColumns) {
-      createInvertedIndexForColumn(segmentMetadata.getColumnMetadataFor(column));
+      try {
+        createInvertedIndexForColumn(segmentMetadata.getColumnMetadataFor(column));
+      } catch (Exception e) {
+        LOGGER.warn("Inverted index not created for column {}", column);
```
Jackie-Jiang: Do we want to swallow this exception?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
mcvsubbu: Because we want to attempt to build inverted index for other columns if possible. See line 169, it is a loop.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
mcvsubbu: Correct. If there is no exception, this object is created fine, and the resources are released in the seal() method.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
mcvsubbu: all the fields are null to begin with, and unload handles the call with null buffer.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
mcvsubbu: Yes, so that we can try to build inverted index for the next field. Not building inverted index is not an error, so no need to abort here.

PR #1910, opened 2017-09-19
Remove jetty dependency from broker

snleee:
Currently, we are using jetty server for broker rest api. The recent version
of jetty only supports jvm 1.8. In order to support jvm 1.7, we are replacing
from jetty to grizzly http server.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerServerBuilder.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerServerBuilder.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.broker.broker;
 
+import com.google.common.base.Preconditions;
 import com.linkedin.pinot.broker.broker.helix.LiveInstancesChangeListenerImpl;
 import com.linkedin.pinot.broker.pruner.SegmentZKMetadataPrunerService;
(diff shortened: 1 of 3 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerServerBuilder.java
```diff
@@ -198,23 +202,35 @@ public void buildHTTP() {
...
     if (clientConfig.enableConsole()) {
-      context.setResourceBase(clientConfig.getConsoleWebappPath());
+      // Use "" for context path to make it "/"
+      context = new WebappContext("brokerServerContext", "", clientConfig.getConsoleWebappPath());
+
     } else {
-      context.setResourceBase("");
+      context = new WebappContext("brokerServerContext");
     }
+    ServletRegistration servletRegistration;
```
mcvsubbu: dont use servlets. see controller example.

snleee: I will work on porting to Jersey for the next release.

mcvsubbu: ok

PR #1913, opened 2017-09-19
Enable autotune to tune alert filter for a group of functions

cyy0714:
When tuning alert filters, the function require some anomalies to run the autotune. However, for some functions, the amount of anomalies is less than the required amount. Also, the criteria of a true anomaly is the same for all alerts labeled by the same team/person. We should be able to tune alert filter for a group of anomaly functions. 

- Modify the endpoint, allow multiple function ids as input
- Modify the output, returning a list of autotune config id

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
@@ -631,8 +631,6 @@ public class DetectionJobResource {
     for (long functionId : anomalyFunctionIds) {
       autotuneConfig.setId(null);
       autotuneConfig.setFunctionId(functionId);
-      AnomalyFunctionDTO anomalyFunctionSpec = DAO_REGISTRY.getAnomalyFunctionDAO().findById(functionId);
-      autotuneConfig.setAlertFilter(alertFilterFactory.fromSpec(anomalyFunctionSpec.getAlertFilter()));
       long autotuneId = DAO_REGISTRY.getAutotuneConfigDAO().save(autotuneConfig);
       autotuneIds.add(autotuneId);
     }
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -615,9 +627,16 @@ public Response tuneAlertFilter(@PathParam("functionId") long id,
     }
 
     // write to DB
-    autotuneConfig.setFunctionId(id);
-    autotuneId = DAO_REGISTRY.getAutotuneConfigDAO().save(autotuneConfig).toString();
-    return Response.ok(autotuneId).build();
+    List<Long> autotuneIds = new ArrayList<>();
+    for (long functionId : anomalyFunctionIds) {
+      autotuneConfig.setId(null);
```
cecilynie: why setting Id to null?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -615,9 +627,16 @@ public Response tuneAlertFilter(@PathParam("functionId") long id,
...
+      autotuneConfig.setFunctionId(functionId);
+      AnomalyFunctionDTO anomalyFunctionSpec = DAO_REGISTRY.getAnomalyFunctionDAO().findById(functionId);
+      autotuneConfig.setAlertFilter(alertFilterFactory.fromSpec(anomalyFunctionSpec.getAlertFilter()));
```
cecilynie: setConfiguration or setAlertFilter? Only use setConfiguration can be put to database. If want to record to db, please use setConfiguration.

cecilynie: Please remove this line if we don't want to save current alert filter anymore. But better have it logged. Thanks!

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
cyy0714: It is for create a new AutotuneConfig entry for each anomaly function.

PR #1919, opened 2017-09-20
Hdfs segment fetcher support

jamesyfshao:
1. add hdfs segment fetcher support
2. make sure all segment fetcher is initialized during controller/server startup
3. fix bug in segmentFactory where it cannot grab config for fetcher
4. add unit test for segment fetcher factory

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
@@ -1,0 +1,78 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.segment.fetcher;
+
+import com.google.common.base.Strings;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.security.UserGroupInformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Created by jamesshao on 9/14/17.
+ */
+public class HdfsSegmentFetcher implements SegmentFetcher {
+
+  private static final String PRINCIPLE = "kerberos.principle";
+  private static final String KEYTAB = "kerberos.keytab";
+  private static final Logger LOGGER = LoggerFactory.getLogger(HdfsSegmentFetcher.class);
+  private FileSystem fs = null;
+
+  @Override
+  public void init(Map<String, String> configs) {
+    Configuration hadoopConf = new Configuration();
+    authenticate(hadoopConf, configs);
+    try {
+      fs = FileSystem.get(new Configuration());
+    } catch (IOException e) {
+      LOGGER.error("failed to initialized the hdfs", e);
+    }
+  }
+
+  private void authenticate(Configuration hadoopConf, Map<String, String> configs) {
+    String principal = configs.get(PRINCIPLE);
+    String keytab = configs.get(KEYTAB);
+    if (!Strings.isNullOrEmpty(principal) && !Strings.isNullOrEmpty(keytab)) {
+      UserGroupInformation.setConfiguration(hadoopConf);
+      if (UserGroupInformation.isSecurityEnabled()) {
+        try {
+          if (!UserGroupInformation.getCurrentUser().hasKerberosCredentials()
+              || !UserGroupInformation.getCurrentUser().getUserName().equals(principal)) {
+            LOGGER.info("Trying to authenticate user [%s] with keytab [%s]..", principal, keytab);
+            UserGroupInformation.loginUserFromKeytab(principal, keytab);
+          }
+        }
+        catch (IOException e) {
+          throw new RuntimeException(String.format("Failed to authenticate user principal [%s] with keytab [%s]",
+                  principal, keytab), e);
+        }
+      }
+    }
+  }
+
+  @Override
+  public void fetchSegmentToLocal(String uri, File tempFile) throws Exception {
+      fs.copyToLocalFile(new Path(uri), new Path(tempFile.toURI()));
+      LOGGER.info("copied {} from hdfs to {} in local for size %d", uri, tempFile.getAbsolutePath(), tempFile.length());
+  }
+
+}
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
@@ -53,8 +53,12 @@ public class HdfsSegmentFetcher implements SegmentFetcher {
 
   private Configuration getConf(String hadoopConfPath) {
     Configuration hadoopConf = new Configuration();
-    hadoopConf.addResource(new Path(hadoopConfPath,"core-site.xml"));
-    hadoopConf.addResource(new Path(hadoopConfPath,"hdfs-site.xml"));
+    if (Strings.isNullOrEmpty(hadoopConfPath)) {
+      LOGGER.warn("no hadoop conf path is provided, will rely on default config");
+    } else {
+      hadoopConf.addResource(new Path(hadoopConfPath, "core-site.xml"));
+      hadoopConf.addResource(new Path(hadoopConfPath, "hdfs-site.xml"));
+    }
     return hadoopConf;
   }
 
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
@@ -26,28 +26,37 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
-/**
- * Created by jamesshao on 9/14/17.
- */
+import static com.linkedin.pinot.common.utils.CommonConstants.SegmentFetcher.RETRY;
+import static com.linkedin.pinot.common.utils.CommonConstants.SegmentFetcher.RETRY_WAITIME_MS;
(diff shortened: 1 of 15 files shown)


jamesyfshao:
a few things might need to be further worked on, depends on how critical they are:
1. retry/wait for hdfs fetch failure
2. more testing on kerberos or other hadoop authentication mechanism

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -61,15 +70,27 @@ public static void initSegmentFetcherFactory(Configuration pinotHelixProperties)
     }
   }
 
-  public static void initSegmentFetcher(Map<String, String> configs) {
+  @VisibleForTesting
+  protected static Map<String, SegmentFetcher> getPreloadedSegments() {
```
fx19880617: getSegmentFetcher()?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.hadoop.security.UserGroupInformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Created by jamesshao on 9/14/17.
```
kishoreg: As per ASF guidelines, we prefer not to have author names in the code.

pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/testSegmentFetcher.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package com.linkedin.pinot.common.segment.fetcher;
+
+import java.io.File;
+import java.util.Map;
+
+/**
+ * Created by jamesshao on 9/20/17.
+ */
+public class testSegmentFetcher implements SegmentFetcher {
```
kishoreg: Is this a test class or a DummySegmentFetcher?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
```diff
@@ -0,0 +1,103 @@
...
+    LOGGER.debug("starting to fetch segment from hdfs");
+    try {
+      if (fs == null) {
+        LOGGER.error("uninitialized fs for fetching data from hdfs");
+        throw new RuntimeException("failed to get hdfs client");
+      }
+      
+      fs.copyToLocalFile(new Path(uri), new Path(tempFile.toURI()));
+      LOGGER.debug("copied {} from hdfs to {} in local for size {}", uri, tempFile.getAbsolutePath(), tempFile.length());
+    } catch(Exception ex) {
```
kishoreg: - check for existing of the file first
- after downloading, check for size/crc to ensure the data is correctly copied
Are there different types of exceptions? If there is a connection break, It will be good to retry.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -15,44 +15,53 @@
...
+    SEGMENT_FETCHER_MAP.put("hdfs", new HdfsSegmentFetcher());
   }
 
   public static void initSegmentFetcherFactory(Configuration pinotHelixProperties) {
     Configuration segmentFetcherFactoryConfig =
         pinotHelixProperties.subset(CommonConstants.Server.PREFIX_OF_CONFIG_OF_SEGMENT_FETCHER_FACTORY);
 
+    // initialize predefined fetcher
+    for (String protocol: SEGMENT_FETCHER_MAP.keySet()) {
+      LOGGER.debug("initializing segment fetcher protocol {}", protocol);
```
kishoreg: this can be info since its one time during start up

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -15,44 +15,53 @@
...
     Iterator segmentFetcherFactoryConfigIterator = segmentFetcherFactoryConfig.getKeys();
     while (segmentFetcherFactoryConfigIterator.hasNext()) {
       Object configKeyObject = segmentFetcherFactoryConfigIterator.next();
       try {
         String segmentFetcherConfigKey = configKeyObject.toString();
-        String protocol = segmentFetcherConfigKey.split(".", 2)[0];
+        String protocol = segmentFetcherConfigKey.split("\\.", 2)[0];
         if (!SegmentFetcherFactory.containsProtocol(protocol)) {
-          SegmentFetcherFactory.initSegmentFetcher(
+          LOGGER.debug("initializing segment fetcher protocol " + protocol);
```
kishoreg: can be info and log the configuration if its not too big

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
jamesyfshao: updated for all files with author names

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
jamesyfshao: I researched a bit on crc calculation for hadoop/local fs, it seems that according to this stackoverflow answer https://stackoverflow.com/a/14564532 it is tricky to implement the crc validation across different file systems. I think for the sake of simplicity we can skip this part as we are not doing that for other fetchers as well. 

For checking local file existence, I think this might be dangerous as they might be partial copy results. to verify that file content is correct we will need to do checksum verification or other things. This leads us back to my previous comments

pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/testSegmentFetcher.java
jamesyfshao: it is supposed to be a dummy fetcher that we used for unit test (see pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactoryTest.java)

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
jamesyfshao: updated all start up log to info

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
jamesyfshao: log both message and config

jamesyfshao:
will update wiki once the pull request is merged

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
```diff
@@ -0,0 +1,131 @@
...
+  private FileSystem hadoopFS = null;
+  private int retryCount = 3;
+  private int retryWaitMs = 100;
+
+  @Override
+  public void init(Map<String, String> configs) {
+    String hadoopConfPath = configs.get(HADOOP_CONF_PATH);
+    try {
+      if (configs.containsKey(RETRY)) {
+        retryCount = Integer.valueOf(configs.get(RETRY));
```
jfim: Can you add a try/catch block and a log if it fails to parse? (eg. `Failed to parse value RETRY, using default value of retryCount`)

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
```diff
@@ -0,0 +1,131 @@
...
+  @Override
+  public void fetchSegmentToLocal(String uri, File tempFile) throws Exception {
+    LOGGER.debug("starting to fetch segment from hdfs");
+    String tempFilePath = tempFile.getAbsolutePath();
+    try {
+      Path remoteFile = new Path(uri);
+      Path localFile = new Path(tempFile.toURI());
+
+      int attemptCount = 0;
+      while (attemptCount++ < this.retryCount) {
```
jfim: Would be better to use `RetryPolicy` to implement the retry logic.

pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/testSegmentFetcher.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.segment.fetcher;
+
+import java.io.File;
+import java.util.Map;
+
+public class testSegmentFetcher implements SegmentFetcher {
```
jfim: `TestSegmentFetcher` with an uppercase `t`

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
jamesyfshao: changed the whole segment fetcher initialization argument to configuration to facilitate default/exception handling. The old way of passing in Map<String, String> is unnecessary and cause a lot more redundant codes. Also refactor other fetcher and updated tests

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
jamesyfshao: Done and updated HttpSegmentFetcher with similar logics as well

pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/testSegmentFetcher.java
jamesyfshao: Done

jamesyfshao:
@jfim updated based your feedback and cleaned up a bit more on httpSegmentFetcher following the recommendation/best-practice you provided as well, thanks for the review!

jfim:
@jamesyfshao Can you make sure to add a page to the Wiki that explains how to configure it? Thanks!

jamesyfshao:
@jfim I have written a wiki doc for hdfs segment fetching under my fork: https://github.com/jamesyfshao/pinot/wiki/Pinot-Segment-Fetcher-Configuration. Please take a look if it looks fine and I will add it to linkedin/pinot wiki. Thanks for all the review and merging!

jfim:
Great, saw that you added the wiki doc at https://github.com/linkedin/pinot/wiki/Pinot-Segment-Fetcher-Configuration

Thanks a lot!

mcvsubbu:
@jamesyfshao Your commit seems to be explicitly blocking overriding of a fetcher class via configuration (both in tests and code). This feature was present before the commit. We now have a requirement to be able to override the classes defined for specific protocol.

Did you have a requirement to NOT be able to override the pre-defined classes, or was that an oversight? I wanted to check before making the change to fix the issue.

mcvsubbu:
I take that back. I think the test is fine with it, but it is disabled in code in 2 places, I will fix it.

PR #1921, opened 2017-09-21
In MCombineGroupByOperator, add early termination to release thread resources

Jackie-Jiang:
While combining results, if encountered exception or query timed out, should stop the query execution.
With this change, we can prevent resources being wasted for queries that already failed.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
@@ -15,6 +15,12 @@
  */
 package com.linkedin.pinot.core.plan;
 
+import com.linkedin.pinot.common.request.BrokerRequest;
+import com.linkedin.pinot.core.common.Operator;
+import com.linkedin.pinot.core.operator.MCombineGroupByOperator;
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
```diff
@@ -15,24 +15,21 @@
...
 import org.slf4j.LoggerFactory;
-import com.linkedin.pinot.common.request.BrokerRequest;
-import com.linkedin.pinot.core.common.Operator;
-import com.linkedin.pinot.core.operator.MCombineGroupByOperator;
-import com.linkedin.pinot.core.operator.MCombineOperator;
-import com.linkedin.pinot.core.query.exception.BadQueryRequestException;
-import com.linkedin.pinot.core.util.trace.TraceCallable;
 
 
-/**
```
snleee: Any reason you kept comments in `MCombineGroupByOperator ` and removed in this class?

Jackie-Jiang: Good catch! I don't intent to remove them, will add them back.

PR #1926, opened 2017-09-23
Add dateTimeConvert udf to convert a dateTime column from one format to another, and bucket to given granularity

npawar:
This udf should be invoked with arguments:
1. Column name to convert eg. Date
2. format(as defined in {@link DateTimeFieldSpec}) of the input column. eg: 1:HOURS:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd 
3. format(as defined in {@link DateTimeFieldSpec}) of the output expected. eg: 1:MILLISECONDS:EPOCH, 1:WEEKS:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd
4. granularity to bucket data into eg: 1:HOURS, 15:MINUTES

Tested queries:
select count(*) from tablename group by dateTimeConvert(timecolumn, '1:HOURS:EPOCH', '1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd', '1:DAYS')

select count(*) from tablename group by dateTimeConvert(timecolumn, '1:HOURS:EPOCH', '1:MILLISECONDS:EPOCH', '1:DAYS')

select count(*) from tablename group by dateTimeConvert(timecolumn, '1:HOURS:EPOCH', '1:MILLISECONDS:EPOCH', '1:HOURS')

select count(*) from tablename group by dateTimeConvert(timecolumn, '1:HOURS:EPOCH', '1:DAYS:EPOCH', '1:DAYS')


pinot-common/src/main/java/com/linkedin/pinot/common/utils/request/RequestUtils.java
```diff
@@ -201,8 +205,25 @@ public static boolean isFitForStarTreeIndex(SegmentMetadata segmentMetadata, Fil
...
+
       List<String> groupByColumns = groupBy.getColumns();
-      for (String groupByColumn : groupByColumns) {
+      if (groupByColumns != null) {
+        allGroupByColumns.addAll(groupByColumns);
+      }
+
+      List<String> groupByExpressions = groupBy.getExpressions();
+      if (groupByExpressions != null) {
+        Pql2Compiler compiler = new Pql2Compiler();
```
mayankshriv: We should avoid multiple compilations of the expression tree. Lets explore a way to achieve that.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
```diff
@@ -0,0 +1,393 @@
...
+
+  /* DateTimeFieldSpec granularity is of format size:timeUnit */
+  private static final int GRANULARITY_SIZE_POSITION = 0;
+  private static final int GRANULARITY_UNIT_POSITION = 1;
+  private static final int MAX_GRANULARITY_TOKENS = 2;
+
+
+  private static final DateTime EPOCH_START_DATE = new DateTime(0L, DateTimeZone.UTC);
+
+  public enum CustomTimeUnit {
```
mayankshriv: The original implementation had Java and Custom TimeUnits, as they were added a different times. We should have a single time converter (probably JODA based) that handles all time units.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
```diff
@@ -0,0 +1,393 @@
...
+   * dateTimeSpec.fromMillis(1498892400000) = 4996308 (i.e. timeColumnValueMS/(1000*60*5))
+   *
+   * 3) given dateTimeColumnValueMS = 1498892400000 and format=1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd,
+   * dateTimeSpec.fromMillis(1498892400000) = 20170701
+   * @param dateTimeColumnValueMS
+   * @param toFormat - the format in which to convert the millis value
+   * @return dateTime column value in dateTimeFieldSpec
+   */
+  @JsonIgnore
+  public static Object fromMillisToFormat(Long dateTimeColumnValueMS, String toFormat) {
```
mayankshriv: Why is the return type Object, and not a concrete type?

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
```diff
@@ -0,0 +1,393 @@
...
+    long bucketedDateTimeValueMS = (dateTimeColumnValueMS / granularityMillis) * granularityMillis;
+    return bucketedDateTimeValueMS;
+  }
+
+  /**
+   * Check correctness of format of {@link DateTimeFieldSpec}
+   * @param format
+   * @return
+   */
+  public static boolean validFormat(String format) {
```
mayankshriv: s/validFormat/isValidFormat ?

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
```diff
@@ -0,0 +1,393 @@
...
+    Preconditions.checkArgument(granularityTokens[GRANULARITY_SIZE_POSITION].matches(NUMBER_REGEX)
+        && EnumUtils.isValidEnum(TimeUnit.class, granularityTokens[GRANULARITY_UNIT_POSITION]), GRANULARITY_PATTERN_ERROR_STR);
+
+    return true;
+  }
+
+  /**
+   * Given a timestamp in millis, convert it to the format specified, while also handling time units such as WEEKS, MONTHS, YEARS
+   * This is a wrapper over the fromMillisToFormat method, which only accepts units in TimeUnits
+   * eg:
```
mayankshriv: Use javadoc syntax for bullets.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
```diff
@@ -0,0 +1,393 @@
...
+   * @return
+   */
+  public static Long convertMillisToCustomFormat(Long dateTimeValueMillis, String outputFormat) {
+    Preconditions.checkNotNull(dateTimeValueMillis);
+    Preconditions.checkNotNull(outputFormat);
+
+    Long convertedDateTime = null;
+    String timeUnit = outputFormat.split(COLON_SEPARATOR)[FORMAT_UNIT_POSITION];
+
+    if (EnumUtils.isValidEnum(TimeUnit.class, timeUnit)) {
```
mayankshriv: Having a single converter that handles all time units will avoid the if/else here.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
```diff
@@ -0,0 +1,103 @@
...
+      _output = new long[Math.max(length, DocIdSetPlanNode.MAX_DOC_PER_CALL)];
+    }
+
+    for (int i = 0; i < inputDateTimeValues.length; i++) {
+      Long dateTimeColumnValueMS = DateTimeFieldSpecUtils.fromFormatToMillis(inputDateTimeValues[i], inputFormat);
+      Long bucketedDateTimevalueMS = DateTimeFieldSpecUtils.bucketDateTimeValueMS(dateTimeColumnValueMS, outputGranularity);
+      Long dateTimeValueConverted = DateTimeFieldSpecUtils.convertMillisToCustomFormat(bucketedDateTimevalueMS, outputFormat);
+      _output[i] = (Long) dateTimeValueConverted;
+    }
+    System.out.println();
```
mayankshriv: Remove this.

pinot-core/src/test/java/com/linkedin/pinot/query/transform/DateTimeConversionTransformTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+
+/**
+ * Unit test for {@link DateTimeConversionTransform}.
+ * Checks if the bucketing is right, and the final format is right
+ */
+public class DateTimeConversionTransformTest {
```
mayankshriv: Could you also add tests for conversion to different time zones?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
```diff
@@ -0,0 +1,103 @@
...
+  private static final String TRANSFORM_NAME = "dateTimeConvert";
+  private long[] _output = null;
+
+
+  @Override
+  public <T> T transform(int length, BlockValSet... input) {
+    Preconditions.checkArgument(input.length == 4, TRANSFORM_NAME + " expects four arguments");
+
+    long[] inputDateTimeValues = input[0].getLongValuesSV();
+    String[] inputDateTimeFormat = input[1].getStringValuesSV();
```
mayankshriv: Probably outside the scope of this PR, but would be good to cleanup requiring a Blockvalset with same value repeated 'length' times. I was thinking to add a separate api to fetch single value on ConstantBlockValSet. That requires casting to ConstantBlockValSet here, but that should be OK, since the function expects certain arguments.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+@NotThreadSafe
+public class DateTimeConversionTransform implements TransformFunction {
+  private static final String TRANSFORM_NAME = "dateTimeConvert";
+  private long[] _output = null;
+
+
+  @Override
+  public <T> T transform(int length, BlockValSet... input) {
+    Preconditions.checkArgument(input.length == 4, TRANSFORM_NAME + " expects four arguments");
```
mayankshriv: The last argument (bucket?) could be optional?

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
npawar: The format can be simple date format, which is a string, or it can be epoch which is a long

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
npawar: No it shouldn't be. We're changing the schema because we have been overloading column format and bucket into one. It is better if they are explicitly expressed separately

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
mayankshriv: There's a few options:
1. Have different methods for different return types.
2. Create an interface.
3. If the caller knows the return type, you can have this return a generic type as opposed to Object.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
mayankshriv: I understand it is better to separate the two. My question is for cases where bucketing is not needed, do clients still need to pass the third argument, or can there be a default value?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
```diff
@@ -0,0 +1,103 @@
...
+
+    if (_output == null || _output.length < length) {
+      _output = new long[Math.max(length, DocIdSetPlanNode.MAX_DOC_PER_CALL)];
+    }
+
+    for (int i = 0; i < inputDateTimeValues.length; i++) {
+      Long dateTimeColumnValueMS = DateTimeFieldSpecUtils.fromFormatToMillis(inputDateTimeValues[i], inputFormat);
+      Long bucketedDateTimevalueMS = DateTimeFieldSpecUtils.bucketDateTimeValueMS(dateTimeColumnValueMS, outputGranularity);
+      Long dateTimeValueConverted = DateTimeFieldSpecUtils.convertMillisToCustomFormat(bucketedDateTimevalueMS, outputFormat);
+      _output[i] = (Long) dateTimeValueConverted;
```
mayankshriv: Unnecessary cast on RHS.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * This class implements the date time conversion transform.
+ *
+ * <ul>
+ *   <li> This udf should be invoked with arguments: </li>
+ *   <ul>
+ *     <li> 1. Column name to convert eg. Date </li>
+ *     <li> 2. format(as defined in {@link DateTimeFieldSpec}) of the input column. eg: 1:HOURS:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd </li>
+ *     <li> 3. format(as defined in {@link DateTimeFieldSpec}) of the output expected. eg: 1:MILLISECONDS:EPOCH, 1:WEEKS:EPOCH, 1:DAYS:SIMPLE_DATE_FORMAT:yyyyMMdd </li>
```
mayankshriv: This fails style check, ie long than 160 characters.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
npawar: Fixed

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
npawar: Fixed, with option 3

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
npawar: done

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
npawar: done

pinot-common/src/main/java/com/linkedin/pinot/common/utils/time/DateTimeFieldSpecUtils.java
npawar: fixed

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
npawar: With all usecases I've encountered, the bucketing (last argument) is more important than the output format (second last), as the second last argument could often be the same as the first. I prefer having all three mandatory, so that users don't try to overload the function of one into the other. Let me know what you think

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
npawar: doen

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
npawar: Done

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/DateTimeConversionTransform.java
npawar: removed

PR #1943, opened 2017-09-27
[TE] Self serve alert edit bug fix

newsummit:
Fixing graph bug in Self-Serve Alert Edit UI - the wrong granularity was being used.


newsummit:
wrong branch

PR #1946, opened 2017-09-28
Add support for star tree segments with udfs

npawar:
Adding support in RequestUtils to parse expressions and check those columns in star tree segments case. Eventually we will move the logic for compiling expressions one level up, into ServerQueryRequest, and resuse them, instead of compiling multiple times in the lifespan of the query

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/request/RequestUtils.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/request/RequestUtils.java
@@ -207,22 +207,7 @@ public class RequestUtils {
     // Ensure that none of the group-by columns are metric or skipped for materialization.
     GroupBy groupBy = brokerRequest.getGroupBy();
     if (groupBy != null) {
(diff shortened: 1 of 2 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/request/RequestUtils.java
```diff
@@ -201,8 +207,24 @@ public static boolean isFitForStarTreeIndex(SegmentMetadata segmentMetadata, Fil
     // Ensure that none of the group-by columns are metric or skipped for materialization.
     GroupBy groupBy = brokerRequest.getGroupBy();
     if (groupBy != null) {
+      Set<String> allGroupByColumns = new HashSet<>();
```
mayankshriv: This could be refactored out as a utility function that returns all related columns to be checked.

npawar: done, added tests for the function too

PR #1949, opened 2017-09-28
[TE] Enable HttpUtils to pass Authentication tokens

cyy0714:
With the enable of authentication on ThirdEye, httpUtils are blocked because of fail to authenticate. This pull request is to allow http utils to include authentication tokens in the http request cookies, and pass the authentication check on ThirdEye.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/utils/AbstractResourceHttpUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/utils/AbstractResourceHttpUtils.java
@@ -1,6 +1,5 @@
 package com.linkedin.thirdeye.anomaly.utils;
 
-import com.sun.xml.internal.ws.api.policy.PolicyResolver;
 import java.io.IOException;
 import java.io.InputStream;
 
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/utils/AbstractResourceHttpUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/utils/AbstractResourceHttpUtils.java
@@ -1,5 +1,6 @@
 package com.linkedin.thirdeye.anomaly.utils;
 
+import com.sun.xml.internal.ws.api.policy.PolicyResolver;
 import java.io.IOException;
 import java.io.InputStream;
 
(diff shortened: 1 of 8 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/DetectorHttpUtils.java
```diff
@@ -18,8 +20,12 @@
   private static String EMAIL_JOBS_ENDPOINT = "/api/email-jobs/";
   private static String ADHOC = "/ad-hoc";
 
-  public DetectorHttpUtils(String detectorHost, int detectorPort) {
+  public DetectorHttpUtils(String detectorHost, int detectorPort, String authToken) {
     super(new HttpHost(detectorHost, detectorPort));
+    BasicClientCookie cookie = new BasicClientCookie("te_auth", authToken);
```
apucher: why is this code duplicated four times?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/utils/AbstractResourceHttpUtils.java
```diff
@@ -1,22 +1,35 @@
 package com.linkedin.thirdeye.anomaly.utils;
 
+import com.sun.xml.internal.ws.api.policy.PolicyResolver;
```
apucher: this private dependency should not be here

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/DetectorHttpUtils.java
cyy0714: It's for generating the cookie with authentication tokens. I do this way so that user only need to worry about the authentication token, and the constructor makes the cookie and insert into the request. I leave more flexibility for adding more cookie in the future.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/utils/AbstractResourceHttpUtils.java
cyy0714: It's removed. Thanks for pointing out.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/DetectorHttpUtils.java
cyy0714: Thanks for the comment. The duplicate lines are removed.

PR #1951, opened 2017-09-29
Pass simple date format into segment generation code

jenniferdai:
* Allows us to use this config to generate segment names correctly


pinot-core/src/main/java/com/linkedin/pinot/core/segment/DefaultSegmentNameGenerator.java
```diff
@@ -77,7 +77,7 @@ public DefaultSegmentNameGenerator(String timeColumnName, String tableName, Stri
    * @throws Exception
    */
   @Override
-  public String generateSegmentName(ColumnStatistics statsCollector) throws Exception {
+  public String generateSegmentName(ColumnStatistics statsCollector, String simpleDateFormat) throws Exception {
```
mcvsubbu: Would be better to take the timeColumnFormat in the constructor along with timeColumnName.

PR #1955, opened 2017-09-30
Remove SegmentId, SegmentIdSet and clean up related classes

Jackie-Jiang:
SegmentId and SegmentIdSet are unnecessary wrappers on String and Set<String> which would create garbage and potentilly hurt performance
Use Map<String, List<String>> (server to list of segments) to represent routing table
Clean up RoutingTableBuilder and RoutingTable classes
In GeneratorBasedRoutingTableBuilder, use balanced way (assign new segment to server with least segments) to assign segments instead of using weight


pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/KafkaLowLevelConsumerRoutingTableBuilder.java
```diff
@@ -88,54 +74,13 @@ protected RoutingTableGenerator buildRoutingTableGenerator() {
...
-    // Of those three weighting alternatives, the third one has the smallest standard deviation of the number of
-    // segments assigned per replica, so it corresponds to the weighting strategy used for segment assignment. Empirical
-    // testing shows that for 20 segments and three replicas, the standard deviation of each alternative is respectively
-    // 2.112, 1.496 and 0.853.
-    //
-    // This algorithm works as follows:
-    // 1. Gather all segments and group them by Kafka partition, sorted by sequence number
-    // 2. Ensure that for each partition, we have at most one partition in consuming state
-    // 3. Sort all the segments to be used during assignment in ascending order of replicas
-    // 4. For each segment to be used during assignment, pick a random replica, weighted by the number of segments
```
snleee: I think it's better to keep the comments for step 1,2,3.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/DefaultOfflineRoutingTableBuilder.java
```diff
@@ -28,24 +29,19 @@
...
 /**
  * Create a given number of routing tables based on random selections from ExternalView.
  */
-public class DefaultOfflineRoutingTableBuilder extends AbstractRoutingTableBuilder {
+public class DefaultOfflineRoutingTableBuilder extends BaseRoutingTableBuilder {
   private static final Logger LOGGER = LoggerFactory.getLogger(DefaultOfflineRoutingTableBuilder.class);
+
   private RoutingTableBuilder _largeClusterRoutingTableBuilder;
   private RoutingTableBuilder _smallClusterRoutingTableBuilder;
+  private volatile RoutingTableBuilder _routingTableBuilder;
```
snleee: Can you explain the logic behind of putting `volatile` for this variable?

Jackie-Jiang: Added comment

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/CfgBasedRouting.java
```diff
@@ -52,18 +49,7 @@ public void init(RoutingTableConfig cfg) {
...
-
-  @Override
-  public void start() {
-    // Nothing to be done here
-  }
-
-  @Override
-  public void shutdown() {
-    // Nothing to be done here
+    return _cfg.getPerTableRoutingCfg().containsKey(tableName);
```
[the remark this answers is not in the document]
Jackie-Jiang: Yes, this is an intentional change. The reason for this change is that, all other RoutingTable implementations don't check whether the routing table is empty (it will never be null if tableName exists). I changed the javadoc for the API as well.

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/KafkaLowLevelConsumerRoutingTableBuilder.java
Jackie-Jiang: The comment is in init() method

PR #1968, opened 2017-10-04
Remove FileBasedInstanceDataManager

Jackie-Jiang:
FileBasedInstanceDataManager is only used in unit tests, and has different set of APIs compare to HelixInstanceDataManager
Removed FileBasedInstanceDataManager to simplify the APIs
Added mocked InstanceDataManager for tests


kishoreg:
LGTM. Demolition man!

PR #1971, opened 2017-10-05
[RE] prepare rootcause ranking for service matching

apucher:
* remove related entities from formatters (now added by pipelines)

* support input and output filtering by entity mapping pipeline

* new scoring hierarchy for matching vs non-matching event entities

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/EntityMappingPipeline.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/EntityMappingPipeline.java
@@ -236,37 +236,47 @@ public class EntityMappingPipeline extends Pipeline {
   private List<EntityToEntityMappingDTO> filterMappings(List<EntityToEntityMappingDTO> mappings) {
     List<EntityToEntityMappingDTO> output = new ArrayList<>();
     for (EntityToEntityMappingDTO mapping : mappings) {
-      if (passesFilters(mapping)) {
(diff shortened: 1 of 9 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/EntityMappingPipeline.java
```diff
@@ -226,6 +233,42 @@ public PipelineResult run(PipelineContext context) {
...
+    List<EntityToEntityMappingDTO> output = new ArrayList<>();
+    for (EntityToEntityMappingDTO mapping : mappings) {
+      if (passesFilters(mapping)) {
+        output.add(mapping);
+      }
+    }
+    return output;
+  }
+
+  private boolean passesFilters(EntityToEntityMappingDTO mapping) {
```
cyenjung: Are the filters applied with OR logic? Could you put down some comments?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/EntityMappingPipeline.java
```diff
@@ -226,6 +233,42 @@ public PipelineResult run(PipelineContext context) {
...
+    boolean passesInput = this.inputFilters.isEmpty();
+    if (!passesInput) {
```
cyenjung: Extract this block and reuse it for checking both input and output filters.

PR #1977, opened 2017-10-06
Pass property store and server metrics into InstanceDataManager and TableDataManager

Jackie-Jiang:
Property store and server metrics are both useful inside InstanceDataManager and TableDataManager
With property store, we don't need to fetch everything outside and pass them in
Fix the hack inside RealtimeTableDataManager to get property store
Server metrics is useful for tracking the exceptions while loading segments

Reduce the arguments for adding segment since the InstanceDataManager has access to property store


pinot-common/src/main/java/com/linkedin/pinot/common/data/DataManager.java
```diff
@@ -15,44 +15,54 @@
...
+import java.io.File;
 import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.apache.commons.configuration.Configuration;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
 
 
+// TODO: maybe better to merge APIs for adding segment to OFFLINE/REALTIME table
```
mcvsubbu: Since FileBasedDataManager is now removed, can we just pass in serverConf to the init method? We will need to get rid of DataManager, InstanceDataManager interfaces, but that should be OK since there are no other DataManagers (and none in sight)

Jackie-Jiang: We need to move huge number of files to solve this because a lot of classes/interfaces in pinot-common / pinot-core use these 2 interfaces.

mcvsubbu: ok, agreed.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/offline/AbstractTableDataManager.java
```diff
@@ -30,78 +33,79 @@
...
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 
-// TODO: pass a reference to Helix property store during initialization. Both OFFLINE and REALTIME use case need it.
 public abstract class AbstractTableDataManager implements TableDataManager {
-  protected final List<String> _activeSegments = new ArrayList<String>();
-  protected final List<String> _loadingSegments = new ArrayList<String>();
+  protected final List<String> _activeSegments = new ArrayList<>();
```
mcvsubbu: I believe _activeSegments and _loadingSegments are not used at all. Can be removed.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/offline/AbstractTableDataManager.java
```diff
@@ -30,78 +33,79 @@
...
   protected final ReadWriteLock _rwLock = new ReentrantReadWriteLock();
   @VisibleForTesting
-  protected final Map<String, SegmentDataManager> _segmentsMap = new HashMap<String, SegmentDataManager>();
+  protected final Map<String, SegmentDataManager> _segmentsMap = new HashMap<>();
 
-  protected Logger LOGGER = LoggerFactory.getLogger(AbstractTableDataManager.class);
-  protected volatile boolean _isStarted = false;
-  protected String _tableName;
+  // Logger should be set up in doInit()
+  protected Logger _logger;
```
mcvsubbu: You should set it to the default logger, to avoid NPE just in case someone adds code to log before calling doInit()

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -268,44 +245,10 @@ private void setupHelixSystemProperties(Configuration conf) {
...
     setShuttingDownStatus(true);
-    if (_pinotHelixProperties.getBoolean(CommonConstants.Server.CONFIG_OF_ENABLE_SHUTDOWN_DELAY, true)) {
+    if (_helixServerConfig.getBoolean(CommonConstants.Server.CONFIG_OF_ENABLE_SHUTDOWN_DELAY, true)) {
       Uninterruptibles.sleepUninterruptibly(_maxQueryTimeMs, TimeUnit.MILLISECONDS);
     }
     _helixManager.disconnect();
     _serverInstance.shutDown();
   }
-
-  public static HelixServerStarter startDefault() throws Exception {
```
mcvsubbu: This is useful code to keep around, while needing to debug in production.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/offline/AbstractTableDataManager.java
Jackie-Jiang: They are useless but used lol. Removed.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/offline/AbstractTableDataManager.java
Jackie-Jiang: Fixed by setting up the logger in init()

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: I don't quite understand why this can be used in production for debugging. We start a local zookeeper and set up a cluster called "quickstart" in prod?

PR #1979, opened 2017-10-06
Change default query executor timeout in CommonConstants to 15s (same as default in query executor)

Jackie-Jiang:


pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
```diff
@@ -233,7 +233,7 @@ public ServerType getServerType() {
         "com.linkedin.pinot.server.starter.helix.HelixInstanceDataManager";
     public static final String DEFAULT_QUERY_EXECUTOR_CLASS =
         "com.linkedin.pinot.core.query.executor.ServerQueryExecutorV1Impl";
-    public static final String DEFAULT_QUERY_EXECUTOR_TIMEOUT = "150000";
+    public static final long DEFAULT_QUERY_EXECUTOR_TIMEOUT = 15_000L;
```
mayankshriv: Can we add time unit in the name?

Jackie-Jiang: Added

PR #1981, opened 2017-10-06
Fix memory leak issue on broker side

snleee:
After bumping up the jetty version, we have observed memory leak
error on broker side because ByteBufs were not properly released.
Tested by running load testing locally.


snleee:
I looked at the following documentation to decide the place to release.
http://netty.io/wiki/reference-counted-objects.html

PR #1983, opened 2017-10-08
Add StarTreeIndexSpec and hllConfig to segmentValidationAndRetentionConfig

jenniferdai:
* For the PinotBuildAndPushJob to configure startree and hll in the table config.
* Moving hllConfig from pinot-core to pinot-common


pinot-common/src/main/java/com/linkedin/pinot/startree/hll/HllConfig.java
```diff
@@ -118,4 +122,10 @@ public int getHllFieldSize() {
     }
     return derivedHllFieldToOriginMap;
   }
+
```
mayankshriv: Please check with @antumbde if this can be moved here or if there's a better solution.

pinot-common/src/main/java/com/linkedin/pinot/startree/hll/HllConfig.java
```diff
@@ -30,8 +31,11 @@
  * segment builder will generate corresponding hll derived fields on the fly.
  */
 public class HllConfig {
```
antumbde: To move this to table config, this class should play nicely with Jackson. **Add tests** to verify conversion to json and back works correctly. 
Fields like hllFieldSize are derived fields. Those will not come from json but should be set correctly. Missing fields can throw errors.

pinot-common/src/main/java/com/linkedin/pinot/startree/hll/HllConfig.java
```diff
@@ -30,8 +31,11 @@
...
+  private static final ImmutableBiMap<Integer, Integer> LOG2M_TO_SIZE_IN_BYTES =
+      ImmutableBiMap.of(5, 32, 6, 52, 7, 96, 8, 180, 9, 352);
+
   private int hllLog2m = HllConstants.DEFAULT_LOG2M;
-  private int hllFieldSize = HllUtil.getHllFieldSizeFromLog2m(HllConstants.DEFAULT_LOG2M);
+  private int hllFieldSize = getHllFieldSizeFromLog2m(HllConstants.DEFAULT_LOG2M);
```
antumbde: is LOG2M_TO_SIZE_IN_BYTES duplicated ? Why not continue to use HllUtil ?

jenniferdai: Originally, I didn't want to use HllUtil mainly because it has a method where it pulls in GenericRow, currently in pinot-core also. I've changed it so that I pull out the logic having to do with hllfieldsize into a different HllSizeUtils and put that in pinot-common so that no logic is duplicated.

pinot-common/src/main/java/com/linkedin/pinot/startree/hll/HllConfig.java
jenniferdai: good catch, checked this

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/retention/RetentionManagerTest.java
```diff
@@ -34,7 +34,7 @@
 import com.linkedin.pinot.controller.helix.core.util.ZKMetadataUtils;
 import com.linkedin.pinot.core.indexsegment.generator.SegmentVersion;
 import com.linkedin.pinot.core.segment.creator.impl.V1Constants;
-import com.linkedin.pinot.core.startree.hll.HllConstants;
+import com.linkedin.pinot.startree.hll.HllConstants;
```
snleee: Let's choose empty line or no empty line between `com.linkedin.pinot...` and `java...` and make all classes consistent.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -40,7 +40,8 @@
 import com.linkedin.pinot.core.segment.creator.impl.fwd.SingleValueVarByteRawIndexCreator;
 import com.linkedin.pinot.core.segment.creator.impl.inv.OffHeapBitmapInvertedIndexCreator;
 import com.linkedin.pinot.core.segment.creator.impl.inv.OnHeapBitmapInvertedIndexCreator;
-import com.linkedin.pinot.core.startree.hll.HllConfig;
+import com.linkedin.pinot.startree.hll.HllConfig;
+
```
snleee: Same as above.

PR #1985, opened 2017-10-10
[TE] Self-Serve UI Fixes: dimension call error

newsummit:
1) Fixed bug in calculation of `baselineStart` time in call for graph data top dimensions.
2) Corrected variable inconsistency related to display of warning message for missing email address.
3) Refreshing model for alert create view on exit.


thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/route.js
```diff
@@ -31,5 +31,33 @@ export default Ember.Route.extend({
...
+  /**
+   * Model hook for the create alert route.
+   * @method setupController
+   * @param {Object} controller - active controller
+   * @param {Object} model - active model
+   * @return {undefined}
+   */
+  setupController(controller, model) {
+    this._super(controller, model);
+    controller.set('model', model);
```
justYves: this step `controller.set('model', model);` should already be handled by ember. Any reason why we need to specifically call `setupController` here?

newsummit: Yes, its redundant. I'll remove it.

PR #1989, opened 2017-10-10
Refactor on-heap/off-heap inverted index creator to prevent leaking resources

Jackie-Jiang:
1. When inverted index cannot be created, throw exception instead of not creating it (Even we do not throw the exception, it will fail when trying to load it)
2. Add a field "onHeap" (default false) in SegmentGeneratorConfig so we can choose to generate index using on-heap/off-heap generator (currently only apply to inverted index)
3. On-heap inverted index creator does not have the limit on total number of values, so if using heap memory won't cause problems (Hadoop job), should prefer the on-heap creator
4. Add Closeable to creator classes to prevent leaks

TODO: fix leaks for other type of index creators


pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OnHeapBitmapInvertedIndexCreator.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+public class OnHeapBitmapInvertedIndexCreator implements InvertedIndexCreator {
+  private static final int INT_SIZE_IN_BYTES = Integer.SIZE / Byte.SIZE;
+
+  private final String _columnName;
+  private final File _invertedIndexFile;
+  private final MutableRoaringBitmap[] _bitmaps;
+
+  public OnHeapBitmapInvertedIndexCreator(File indexDir, String columnName, int cardinality) {
+    _columnName = columnName;
```
mcvsubbu: Suggest an info log here will allow us to know which index creator is being used for what column

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -26,301 +30,214 @@
...
+
+  // Posting list related buffers
+  // Constructed in first pass
+  private ByteBuffer _postingListLengthByteBuffer;
+  private IntBuffer _postingListLengthIntBuffer;
+  // Constructed in second pass
+  private ByteBuffer _postingListValueByteBuffer;
+  private ByteBuffer _postingListCurrentOffsetByteBuffer;
+
+  public OffHeapBitmapInvertedIndexCreator(File indexDir, FieldSpec fieldSpec, int cardinality, int numDocs,
```
mcvsubbu: We should be using Mmap for temporary buffers, so that we can load/create multiple segments in parallel without worrying about MaxDirectMemory. Does not have to be in this PR, but please add a TODO.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
```diff
@@ -42,95 +42,47 @@
...
+    }
   }
 
-  private void createInvertedIndexForColumn(ColumnMetadata columnMetadata)
-      throws IOException {
-    // Validation check
+  private void createInvertedIndexForColumn(ColumnMetadata columnMetadata) throws IOException {
     String column = columnMetadata.getColumnName();
-    int totalDocs = columnMetadata.getTotalDocs();
-    if (totalDocs > OffHeapBitmapInvertedIndexCreator.MAX_NUM_ENTRIES) {
```
mcvsubbu: Are these conditions possible for segments created in hadoop prior to this change?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/InvertedIndexCreator.java
```diff
@@ -15,29 +15,51 @@
...
- * Implementation of this class is used to create inverted Indexes
- * Currently two implementations are
- * p4Delta and
- * RoaringBitmap
+ * Currently only support RoaringBitmap inverted index.
+ * <pre>
+ * Layout for RoaringBitmap inverted index:
+ * |-------------------------------------------------------------------------|
+ * |                    Start offset of 1st bitmap                           |
+ * |    End offset of 1rd bitmap (exclusive) / Start offset of 2nd bitmap    |
```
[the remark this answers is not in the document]
Jackie-Jiang: Good catch

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/InvertedIndexCreator.java
```diff
@@ -15,29 +15,51 @@
...
+   * @param dictId Dictionary id
+   */
+  void addSV(int docId, int dictId);
 
-  long totalTimeTakeSoFar();
+  /**
+   * Add an entry for multi-value column.
+   *
+   * @param docId Document id
+   * @param dictIdBuffer Buffer for dictionary ids
```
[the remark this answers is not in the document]
Jackie-Jiang: Fixed

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -116,97 +107,74 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
...
+                new OnHeapBitmapInvertedIndexCreator(_indexDir, columnName, cardinality));
+          } else {
+            _invertedIndexCreatorMap.put(columnName,
+                new OffHeapBitmapInvertedIndexCreator(_indexDir, fieldSpec, cardinality, totalDocs,
+                    indexCreationInfo.getTotalNumberOfEntries()));
+          }
+        }
+      } else {
+        // Create raw index
+        // TODO: add support to multi-value column and inverted index
```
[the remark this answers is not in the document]
Jackie-Jiang: Added

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OnHeapBitmapInvertedIndexCreator.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.IOException;
+import org.apache.commons.io.FileUtils;
+import org.roaringbitmap.buffer.MutableRoaringBitmap;
+
+
+/**
+ * Implementation of {@link InvertedIndexCreator} that uses on-heap memory.
+ */
+public class OnHeapBitmapInvertedIndexCreator implements InvertedIndexCreator {
+  private static final int INT_SIZE_IN_BYTES = Integer.SIZE / Byte.SIZE;
```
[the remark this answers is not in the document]
Jackie-Jiang: Fixed

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OnHeapBitmapInvertedIndexCreator.java
Jackie-Jiang: We will create inverted index for all columns using either onHeap or offHeap

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
Jackie-Jiang: Yes, and we should fail the job instead of skip the inverted index creation

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
Jackie-Jiang: Changing it to mmap buffer will have bad performance impact because of the disk IO (this is already causing problem for star-tree generation). We need to discuss this later when we actually need to generate huge size direct buffers.

Jackie-Jiang:
@mcvsubbu Addressed all the comments. The try-with-resource for auto-close is how we prevent leaks when encountered exception, which is not trivial.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
mcvsubbu: But those jobs have already run for past segments, right? Unless I am missing something.

PR #1993, opened 2017-10-11
[TE] Distinguish different anomaly result source

cyy0714:
We are working on re-design the alert pipeline, removing the notified flag from anomaly results. One problem is that we need to distinguish anomalies from scheduled detection from others. Users are not expect to receive any backfill anomalies in their alerts. To achieve that, a label should be applied to distinguish the difference.

- Add AnomalyResultSource to distinguish the default anomaly detection from others

- Update the anomaly result source in detection task runner

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/constant/AnomalyResultSource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/constant/AnomalyResultSource.java
@@ -1,0 +1,6 @@
+package com.linkedin.thirdeye.constant;
+
+public enum AnomalyResultSource {
+  DEFAULT_ANOMALY_DETECTION,
+  ANOMALY_REPLAY
+}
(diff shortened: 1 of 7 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/constant/AnomalyResultSource.java
```diff
@@ -0,0 +1,6 @@
+package com.linkedin.thirdeye.constant;
+
+public enum AnomalyResultSource {
```
apucher: We'll need a value for user-reported anomalies as well

cyy0714: This user-reported anomalies will be in a separate table. I would ask Yen about his design on this part.

cyy0714: Resolved

apucher:
please always run a *mvn clean test* before checking in code

PR #1998, opened 2017-10-13
[TE] Unify DB Dao Registry in prod and test

cyy0714:
When writing unit tests in TE, it is painful to access to DB. Extending an abstract class is not even a good design. This pull request is to unify the DB access in prod and test, providing a unified and simple testing environment in TE.


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/DaoProvider.java
```diff
@@ -0,0 +1,68 @@
...
+import com.linkedin.thirdeye.datalayer.bao.MetricConfigManager;
+import com.linkedin.thirdeye.datalayer.bao.OnboardDatasetMetricManager;
+import com.linkedin.thirdeye.datalayer.bao.OverrideConfigManager;
+import com.linkedin.thirdeye.datalayer.bao.RawAnomalyResultManager;
+import com.linkedin.thirdeye.datalayer.bao.TaskManager;
+
+
+public interface DaoProvider {
+
+  void restart();
```
cyenjung: Specify the expected behavior of this method.

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/config/ConfigNamespaceIntegrationTest.java
```diff
@@ -13,8 +14,9 @@
 import org.testng.annotations.Test;
 
 
-public class ConfigNamespaceIntegrationTest extends AbstractManagerTestBase {
+public class ConfigNamespaceIntegrationTest {
   private final static Map<String, String> MAP_EXPECTED = new HashMap<>();
+  private DaoProvider DAO_REGISTRY;
```
cyenjung: Rename to testDAOProvider.

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/datalayer/bao/DAOTestBase.java
```diff
@@ -0,0 +1,279 @@
...
+  protected void init() {
+    try {
+      URL url = TestDBResources.class.getResource("/persistence-local.yml");
+      File configFile = new File(url.toURI());
+      PersistenceConfig configuration = DaoProviderUtil.createConfiguration(configFile);
+      initializeDs(configuration);
+
+      DaoProviderUtil.init(ds);
+
+      daoRegistry = DAORegistry.getInstance();
```
cyenjung: Can we do lazy initialization? For instance, we don't set DAO during init() but only when the corresponding get method is invoked. Because most tests only uses one or two of the DAOs.

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/datalayer/bao/DAOTestBase.java
```diff
@@ -0,0 +1,279 @@
...
+
+  @Override
+  public ApplicationManager getApplicationDAO() {
+    return applicationDAO;
+  }
+
+  @Override
+  public void restart() {
+    cleanup();
+    init();
```
cyenjung: We need to fix the problem that init() is invoked multiple times in next PR. The init() initializes database and then DAOs. However, the DAOs will be re-used even they are reset. Therefore, those DAOs will still using the very first DB instead of new DBs.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/DaoProvider.java
cyy0714: Done.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datasource/DAORegistry.java
```diff
@@ -79,85 +80,108 @@ private DAORegistry() {}
...
     return DaoProviderUtil.getInstance(ConfigManagerImpl.class);
   }
 
+  @Override
   public ApplicationManager getApplicationDAO() {
     return DaoProviderUtil.getInstance(ApplicationManagerImpl.class);
   }
+
+  @Override
+  public void restart(){
```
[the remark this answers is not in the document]
cyy0714: Done

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/auto/onboard/AutoOnboardPinotMetricsServiceTest.java
```diff
@@ -22,23 +25,31 @@
...
 import org.testng.annotations.Test;
 
-public class AutoOnboardPinotMetricsServiceTest  extends AbstractManagerTestBase {
+public class AutoOnboardPinotMetricsServiceTest {
 
   private AutoOnboardPinotDataSource testAutoLoadPinotMetricsService;
   private String dataset = "test-collection";
   private Schema schema;
 
+  private DaoProvider DAO_REGISTRY;
```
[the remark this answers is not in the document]
cyy0714: Done

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/config/ConfigNamespaceIntegrationTest.java
cyy0714: Done

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/datalayer/bao/DAOTestBase.java
cyy0714: I am also thinking of this. Otherwise, it is expensive when a new test begins. The problem is the current function uses the create-schema.sql and drop-tables.sql to create and cleanup the whole DB. There is no information for us to construct a single table. We need to go through the sql file. I don't know if there is any other way to solve this.

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/datalayer/bao/DAOTestBase.java
cyy0714: Yes. This operation is expensive. The origin design follows the same process. The current pr doesn't lose the performance. But we can think about how to improve the overall performance. It can be something good to have.

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/datalayer/DAOProviderUtilTest.java
```diff
@@ -1,30 +1,29 @@
...
   @BeforeClass
   public void beforeClass() {
-    super.init();
+    testDAOProvider = DAOTestBase.getInstance();
   }
 
   @AfterClass(alwaysRun = true)
   public void afterClass() {
-    super.cleanup();
+    testDAOProvider.restart();
```
apucher: Sorry to spoil the party, but this will cause concurrency issues. We've been there, done that. TestNG executes @AfterClass and @BeforeClass methods (of different test classes) in parallel.

cyy0714: Hi @apucher, Thanks for the comment. Is there a proper solution for that?

apucher:
Ok, great this should avoid the stale reference problem - i.e. maintain correctness.

The second thing to do is to get rid of the "DaoProvider" interface again. DAORegistry already is a singleton that's referenced all over TE. DaoTestBase only needs to hold the datasource reference - call init on test start, cleanup on the same reference on teardown. DAOs can be accessed via DAORegistry as usual.

PR #1999, opened 2017-10-14
Fix for a segment having no statistics

mcvsubbu:
If a table is dropped when no rows have been consumed, the statistics will get filled with zeros
and that can cause an index-out-of bounds exception the next time table is created and
consumption starts.

This is because the RealtimeTableDataManager is not removed when a table is dropped.

There are also other conditions that can cause 0 stats to happen, so this PR ensures that
stats returned is never 0.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
@@ -416,22 +416,23 @@ public class RealtimeSegmentImpl implements RealtimeSegment {
 
     // If offHeapAllocation, then gather some statistics before  destroying the segment.
     if (isOffHeapAllocation) {
-      RealtimeSegmentStatsHistory.SegmentStats segmentStats = new RealtimeSegmentStatsHistory.SegmentStats();
-      for (Map.Entry<String, MutableDictionary> entry : dictionaryMap.entrySet()) {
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -416,22 +416,23 @@ public void destroy() {
 
     // If offHeapAllocation, then gather some statistics before  destroying the segment.
     if (isOffHeapAllocation) {
-      RealtimeSegmentStatsHistory.SegmentStats segmentStats = new RealtimeSegmentStatsHistory.SegmentStats();
-      for (Map.Entry<String, MutableDictionary> entry : dictionaryMap.entrySet()) {
-        RealtimeSegmentStatsHistory.ColumnStats columnStats = new RealtimeSegmentStatsHistory.ColumnStats();
-        columnStats.setCardinality(entry.getValue().length());
-        columnStats.setAvgColumnSize(entry.getValue().getAvgValueSize());
-        segmentStats.setColumnStats(entry.getKey(), columnStats);
+      if (numDocsIndexed > 0) {
```
mayankshriv: Can the two if's be combined into one, or is there logic inside the outer 'if' that is not inside the inner 'if'?

mcvsubbu: I would prefer to leave them as separate if statements instead of relying on order of execution of the two conditions separated by &&

mcvsubbu:
Merging without travis clearance. Travis is hung. We need this for a hotfix

PR #2000, opened 2017-10-14
Add support for virtual columns

jfim:
Add support for virtual columns, which are columns that have no on-disk
representation but instead contain values that are filled in at run
time.

Add support for three built-in virtual columns:
 - $docId: Returns the document id for the current row
 - $segmentName: Returns the current segment name
 - $hostName: Returns the hostname of the current server

These columns can be used for debugging purposes, such as debugging
query routing (which server answered which query) and data issues (what
are the data ranges in each segment).

Additional built-in virtual columns can be added to
VirtualColumnProviderFactory and customized virtual columns can be added
to any schema by adding a provider to the classpath and specifying the
fully qualified class name as the "virtualColumnProvider" property of
the field specification in the Pinot table schema.

### DIFF
--- a/pinot-common/src/main/antlr4/com/linkedin/pinot/pql/parsers/PQL2.g4
+++ b/pinot-common/src/main/antlr4/com/linkedin/pinot/pql/parsers/PQL2.g4
@@ -139,7 +139,7 @@ WHITESPACE: [ \t\n]+ -> skip;
 
 LINE_COMMENT: '--' ~[\r\n]* -> channel(HIDDEN);
 
-IDENTIFIER: [A-Za-z_][A-Za-z0-9_-]* | '`' (~'`')+ '`';
+IDENTIFIER: '$'?[A-Za-z_][A-Za-z0-9_-]* | '`' (~'`')+ '`';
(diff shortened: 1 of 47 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -374,17 +381,43 @@ public SegmentMetadata getSegmentMetadata() {
...
+    }
+
     // Because document id searchable offset is inclusive, number of documents is 1 larger than it
     return new ColumnDataSource(dataSchema.getFieldSpecFor(columnName), docIdSearchableOffset + 1,
         maxNumberOfMultivaluesMap.get(columnName), columnIndexReaderWriterMap.get(columnName),
         invertedIndexMap.get(columnName), dictionaryMap.get(columnName));
   }
 
+  private ColumnDataSource getVirtualDataSource(String column) {
+    VirtualColumnContext virtualColumnContext = new VirtualColumnContext(NetUtil.getHostnameOrAddress(), _segmentMetadata.getTableName(), getSegmentName(),
```
mayankshriv: Can NetUtil.getHostnameOrAddress() be static final?

PR #2001, opened 2017-10-15
[TE] Self-Serve Manage Alerts: UI Improvements

newsummit:
Fixes added to the display of alerts:

- Styling for readability
- Adding missing data for properties `group` and `application`
- Fixing display of group emails in alert edit view


thirdeye/thirdeye-frontend/app/pods/manage/alerts/index/controller.js
```diff
@@ -186,7 +207,7 @@ export default Ember.Controller.extend({
     this.set('isLoading', true);
     yield timeout(600);
 
-    this.set('selectedsuscriberGroupNames', groupName);
+    this.set('selectedsubscriberGroupNames', groupName);
     this.set('currentPage', 1);
```
justYves: Let's use `this.setProperties` here

PR #2002, opened 2017-10-16
[TE] AllowUpdateFunctionWithOverrideProps

cecilynie:
Create endpoint to allow update self-defined alert filter for existing anomaly function
Extend update anomaly function by using override anomaly properties

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -582,7 +582,7 @@ public class AnomalyResource {
       targetFunction.setAlertFilter(alertFilterConfig);
       anomalyFunctionDAO.update(targetFunction);
     }
-    return Response.ok().build();
+    return Response.ok(id).build();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -523,6 +526,9 @@ public Response updateAnomalyFunction(@NotNull @PathParam("id") Long id,
       anomalyFunctionSpec.setFilters(filterString);
     }
     if (StringUtils.isNotBlank(properties)) {
+      if (isOverrideProp) {
```
bl44: Move this down to before line 534, such that the 'update parts' are  adjacent to each other

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -566,6 +572,36 @@ public Response updateAnomalyFunction(@NotNull @PathParam("id") Long id,
   }
 
   /**
+   * Enable apply self-defined alert filter to anomaly function
+   * @param id functionId to be updated
+   * @param alertFilter alert filter in JSON format
+   * @return updated anomaly function
```
bl44: May let is return only the function Id or 'OK' for a 'PUT' function. You already have the get endpoints if you want to view what you have modified.

cecilynie: Agree! Thanks for your suggestion!

PR #2005, opened 2017-10-17
Modify hadoop dep

jamesyfshao:
In my previous diff for the hdfs fetcher support, I set the Hadoop dep as provided because usually people generally run different versions of Hadoop and better to modify the dependency as needed. However, I noticed this will cause issues in the sample workflow (which is partially fixed in by @jfim [here](https://github.com/linkedin/pinot/pull/1970)). In order to provide user better experience while running the example, I think it is better to make the Hadoop Dep included in pom. The change is tested on local against 2.7.3 Hadoop environment

### DIFF
--- a/pinot-common/pom.xml
+++ b/pinot-common/pom.xml
@@ -159,12 +159,10 @@
     <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-client</artifactId>
-      <scope>provided</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


snleee:
We have fixed the hadoop dependency issue for HDFS fetcher along with JF's commit that you mentioned above and  https://github.com/linkedin/pinot/pull/1973.

I recommend not to include Hadoop dependency for Pinot because Hadoop introduced a lot of library version conflicts and we spent 2-3 days to resolve conflict issues for our internal release.

PR #2009, opened 2017-10-17
Adding an API to validate a schema

snleee:
This api will convert the given schema to the same format
that we see from GET /schemas/{schemaName}. Similar to #1856.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
@@ -129,6 +129,22 @@ public class PinotSchemaRestletResource {
     return addOrUpdateSchema(null, multiPart);
   }
 
+  @POST
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("/schemas/validate")
+  @ApiOperation(value = "Validate schema", notes = "This API returns the schema that matches the one you get "
+      + "from 'GET /schema/{schemaName}'. This allows us to validate schema before apply.")
+  @ApiResponses(value = {@ApiResponse(code = 200, message = "Successfully validated schema"),
+      @ApiResponse(code = 400, message = "Missing or invalid request body"),
+      @ApiResponse(code = 500, message = "Internal error")})
+  public String validateSchema (FormDataMultiPart multiPart) throws Exception {
+    Schema schema = getSchemaFromMultiPart(multiPart);
+    if (!schema.validate(LOGGER)) {
+      throw new ControllerApplicationException(LOGGER, "Invalid schema. Check controller logs", Response.Status.BAD_REQUEST);
+    }
+    return schema.getJSONSchema();
+  }
+
   /**
    * Internal method to add or update schema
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -129,6 +129,26 @@ public SuccessResponse addSchema(FormDataMultiPart multiPart) {
     return addOrUpdateSchema(null, multiPart);
   }
 
+  @POST
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("/schemas/validate")
+  @ApiOperation(value = "Validate schema", notes = "This API returns the table config that matches the one you get "
```
mcvsubbu: fix the message here.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -129,6 +129,26 @@ public SuccessResponse addSchema(FormDataMultiPart multiPart) {
...
+      + "from 'GET /schema/{schemaName}'. This allows us to validate schema before apply.")
+  @ApiResponses(value = {@ApiResponse(code = 200, message = "Successfully validated schema"),
+      @ApiResponse(code = 400, message = "Missing or invalid request body"),
+      @ApiResponse(code = 500, message = "Internal error")})
+  public String validateSchema (FormDataMultiPart multiPart) throws Exception {
+    try {
```
mcvsubbu: you don't need the try/catch. If it throws exception, a 5xx code will be sent out automatically

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -129,6 +129,26 @@ public SuccessResponse addSchema(FormDataMultiPart multiPart) {
...
+      Schema schema = getSchemaFromMultiPart(multiPart);
+      if (!schema.validate(LOGGER)) {
+        throw new ControllerApplicationException(LOGGER, "Invalid schema", Response.Status.BAD_REQUEST);
```
mcvsubbu: I suggest changing the message to: "Invalid schema. Check controller logs"

The logs will identify which field had an error, if it is possible to do so.

snleee: fixed.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
snleee: fixed

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
snleee: removed

PR #2014, opened 2017-10-19
Add data table to query console

jfim:
Add a data table with the query results to the query console, to make
reading query results easier.


kishoreg:
Nice, do you have the screenshots on how it would look? Is it possible to keep history :)

jfim:
You mean keep the query history in the query console? It would be doable, though not in this PR.

PR #2015, opened 2017-10-19
Fix off heap inverted index creation on MacOS X

jfim:
Fix the off heap inverted index creation on MacOS X. It seems that on
MacOS X, the behavior for the position on a file channel is to be
absolute while on Linux it is relative to the current file position.

Rewrite the transferFrom call to use an absolute position on all
platforms.


jfim:
Merged, added comment in the code that explains it.

PR #2017, opened 2017-10-19
[TE] Frontend dataset mapping support

apucher:
Currently related metrics and entities are defined on a per-metric basis. While the entity mapping pipeline supports convex hull exploration, this is has side-effects that are hard understand for our users. Instead, they fall back to repetitive entry of similar entity relationships.

This PR introduces a DatasetEntity that logically groups together multiple metrics. Common relationships of a group of metrics can now explicitly be expressed on the dataset level, as well as on an individual per-metric level as before.

* introduced DatasetEntity

* refactored MetricDatasetPipeline

* adaptations to frontend and thirdeye-admin

### DIFF
--- a/thirdeye/thirdeye-frontend/app/actions/metrics.js
+++ b/thirdeye/thirdeye-frontend/app/actions/metrics.js
@@ -143,10 +143,13 @@ function fetchRelatedMetricIds() {
 
     const filters = JSON.parse(filterJson);
     const filterUrns = Object.keys(filters).map(key => 'thirdeye:dimension:' + key + ':' + filters[key] + ':provided');
-    const urns = [`thirdeye:metric:${metricId}`].concat(filterUrns).join(',');
+    const primaryUrn = `thirdeye:metric:${metricId}`;
+
+    const urns = [primaryUrn].concat(filterUrns).join(',');
 
     return fetch(`/rootcause/query?framework=relatedMetrics&anomalyStart=${anomalyStart}&anomalyEnd=${anomalyEnd}&baselineStart=${baselineStart}&baselineEnd=${baselineEnd}&analysisStart=${analysisStart}&analysisEnd=${analysisEnd}&urns=${urns}`)
       .then(res => res.json())
+      .then(res => res.filter(m => m.urn != primaryUrn))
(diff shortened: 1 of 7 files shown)


thirdeye/thirdeye-frontend/app/actions/metrics.js
```diff
@@ -143,10 +143,13 @@ function fetchRelatedMetricIds() {
...
     const filters = JSON.parse(filterJson);
     const filterUrns = Object.keys(filters).map(key => 'thirdeye:dimension:' + key + ':' + filters[key] + ':provided');
-    const urns = [`thirdeye:metric:${metricId}`].concat(filterUrns).join(',');
+    const primaryUrn = `thirdeye:metric:${metricId}`;
+
+    const urns = [primaryUrn].concat(filterUrns).join(',');
 
     return fetch(`/rootcause/query?framework=relatedMetrics&anomalyStart=${anomalyStart}&anomalyEnd=${anomalyEnd}&baselineStart=${baselineStart}&baselineEnd=${baselineEnd}&analysisStart=${analysisStart}&analysisEnd=${analysisEnd}&urns=${urns}`)
       .then(res => res.json())
+      .then(res => res.filter(m => m.urn != primaryUrn))
```
justYves: nitpick: can we have a more meaningful naming instead of `m`?

thirdeye/thirdeye-pinot/src/main/resources/assets/js/lib/self-service-mappings.js
```diff
@@ -1,4 +1,14 @@
+const MAPPING_TYPES = [
```
justYves: What's the purpose of adding this in the legacy ui?

thirdeye/thirdeye-frontend/app/actions/metrics.js
apucher: fixed

thirdeye/thirdeye-pinot/src/main/resources/assets/js/lib/self-service-mappings.js
apucher: shipping a feature before it's ready. unfortunately.

PR #2018, opened 2017-10-19
[TE] Decouple Pinot data source and cache loader for future extension

cyenjung:
This PR decouple the cache of Pinot response from Cache Registry and move into Pinot data source. The idea is that Cache Registry provides the cache for meta-data, e.g., dataset names, filters, etc., and DATA cache is hidden in the corresponding data source. Consequently, we could connect to the same data source via different type of connection (e.g., direct connection to controller or indirect connection through Rest.li).

This PR also removes the queries that are directly sent to the database instead of via the data source. Specifically, any explicit calls to the cache of a data source are refactored to invoke data source's method instead.

Tested on local dashboard to ensure the connection to Pinot is not broken due to this change.


apucher:
Looks good. Does caching still work for third-party connectors, e.g. autometrics?

cyenjung:
Yes, it works because it has its own cache. From now on, each connection takes care of its own data cache since they could take different queries. The central cache only caches metadata.

PR #2023, opened 2017-10-20
[TE] entity filter component - part 1

newsummit:
This is part 1 of the entity filter component we plan to re-use in the app. It is currently set up to filter results in the alerts search view. We need to: 

- continue developing the "receiving" function in the `manage/alerts/index` controller.
- add "show more"  functionality in the filter blocks
- integration test it

This component is currently disabled.

screen shot 2017-10-20 at 4 16 12 pm


thirdeye/thirdeye-frontend/app/pods/components/entity-filter/component.js
```diff
@@ -0,0 +1,48 @@
...
+
+  alertFilters: [],
+
+  actions: {
+    /**
+     * Handles selection of filter items. Each time a filter is selected, this component will
+     * pass an array of selected filters to its parent.
+     */
+    onFilterSelection(category, filterObj) {
+      Ember.set(filterObj, 'isActive', !filterObj.isActive);
```
justYves: Looks like a good candidate for `this.toggleProperty`

thirdeye/thirdeye-frontend/app/pods/components/entity-filter/component.js
```diff
@@ -0,0 +1,48 @@
...
+      } else {
+        this.set('alertFilters', _.reject(this.get('alertFilters'), function(filter) { return filter.filter === filterObj.name; }));
+      }
+
+      // Return the 'userDidSelectFilter' function passed from the parent as the value of onSelectFilter, and invoke it.
+      this.get('onSelectFilter')(this.get('alertFilters'));
+    },
+
+    toggleDisplay(block) {
+      Ember.set(block, 'isHidden', !block.isHidden);
```
justYves: Same, `this.toggleProperty` should be semantically more correct

thirdeye/thirdeye-frontend/app/styles/wrapper/_color.scss
```diff
@@ -25,4 +25,4 @@ $te-red: #FF2C33;
 $te-green: #469A1F;
 $te-pink: #FF1B90;
 
-$te-grey--border: rgba(0, 0, 0, 0.15);
\ No newline at end of file
+$te-grey--border: app-shade(black, 1);
```
justYves: missing EOF line

thirdeye/thirdeye-frontend/app/pods/components/entity-filter/component.js
newsummit: After some investigation, I found that `toggleProperty` will not be able to find 'filterObj', as it is not an observed property.

thirdeye/thirdeye-frontend/app/pods/components/entity-filter/component.js
newsummit: After some investigation, I found that `toggleProperty` will not be able to find 'block', as it is not an observed property.

PR #2027, opened 2017-10-24
Add request ID to broker connection logs.

mcvsubbu:
The broker connection logs are missing the request ID. Having these will help us trace a particular
request through conneciton phases


pinot-transport/src/main/java/com/linkedin/pinot/transport/pool/KeyedPoolImpl.java
```diff
@@ -117,7 +117,7 @@ public void start() {
   }
 
   @Override
-  public ServerResponseFuture<T> checkoutObject(ServerInstance key) {
+  public ServerResponseFuture<T> checkoutObject(ServerInstance key, String context) {
```
Jackie-Jiang: This context is never picked up right?

mcvsubbu: right. fixed.

PR #2031, opened 2017-10-24
[TE] Self-Serve-Edit - bug fix

newsummit:
Fix for alert edit edge case (no email recipients)

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/manage/alerts/edit/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/manage/alerts/edit/controller.js
@@ -196,8 +196,8 @@ export default Ember.Controller.extend({
     'selectedConfigGroup',
     'updatedRecipients',
     function() {
-      const newRecipients = this.get('updatedRecipients');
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/manage/alerts/edit/controller.js
```diff
@@ -196,8 +196,8 @@ export default Ember.Controller.extend({
     'selectedConfigGroup',
     'updatedRecipients',
     function() {
-      const newRecipients = this.get('updatedRecipients');
-      const originalRecipients = this.get('selectedConfigGroup.recipients');
+      const newRecipients = this.get('updatedRecipients') || [];
```
justYves: nit: If the values is undefined, Lets use `this.getWithDefault` here: 
https://dockyard.com/blog/2016/03/18/get-with-default

Caveat: this does not work when the value is `null`

newsummit: Yeah, `selectedConfigGroup` is null, but its prop is undefined. So I'll use it.

PR #2032, opened 2017-10-25
[TE] frontend - show contribution table for investigation period only

apucher:


thirdeye/thirdeye-frontend/app/pods/components/contribution-table/component.js
```diff
@@ -123,7 +123,7 @@ export default Ember.Component.extend({
         return index;
       }
     }
-    return dates.length - 1;
+    return dates.length;
```
justYves: this might result in an undefined behavior. since our array starts at index 0, `dates[dates.length]` will result in an error

apucher: good catch. I fixed the code.
usually the end index is the first excluded index. I assume this this is true for JS as well.

kishoreg:
we should add a buffer to the investigation period based on the granularity of the data. 
 change it to contribution to overall change in the case of dimension analysis and percentage change for metric correlation.
can we show only 1 decimal point in the value

PR #2038, opened 2017-10-27
Added an option to allocate Direct Offheap memory for realtime consuming segments

mcvsubbu:
So far, off-heap allocation for realtime consuming segments meant MMAP-based allocation.
This may generate excessive disk i/o depending on the configuaration. Added a setting
to allocate this memory via direct allocation.

Since direct memory translates to an internal memory allocation call, we may end up
allocating more than what we actually need. Fragmentation is also possible, since multiple
threads will be allocating memory of different sizes, and may release them in different order.

A more fancy algorithm may be to do a bit of memory management in this code (like how mmap
does, but at a smaller mem level, perhaps). For now, we leave it at this.

Changed some logs to be more informative, and added one metric on the server to have it
report the number of segments scanned in a query as a gauge.

Tested LLRealtimeClusterIntegrationTest with both direct and mmap allocations

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeSegmentDataManager.java
@@ -49,13 +49,14 @@ public abstract class RealtimeSegmentDataManager extends SegmentDataManager {
     return _statsHistory;
   }
 
-  protected void initMemoryManager(RealtimeTableDataManager realtimeTableDataManager, boolean isOffHeapAllocation, String segmentName) {
+  protected void initMemoryManager(RealtimeTableDataManager realtimeTableDataManager, boolean isOffHeapAllocation,
+      boolean isDirectAllocation, String segmentName) {
     ServerMetrics serverMetrics = realtimeTableDataManager.getServerMetrics();
-    if (isOffHeapAllocation) {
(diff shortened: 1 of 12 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeSegmentDataManager.java
```diff
@@ -49,13 +49,14 @@ public RealtimeSegmentStatsHistory getStatsHistory() {
     return _statsHistory;
   }
 
-  protected void initMemoryManager(RealtimeTableDataManager realtimeTableDataManager, boolean isOffHeapAllocation, String segmentName) {
+  protected void initMemoryManager(RealtimeTableDataManager realtimeTableDataManager, boolean isOffHeapAllocation,
+      boolean isDirectAllocation, String segmentName) {
     ServerMetrics serverMetrics = realtimeTableDataManager.getServerMetrics();
-    if (isOffHeapAllocation) {
+    if (isOffHeapAllocation && !isDirectAllocation) {
```
Jackie-Jiang: We should just rename the 'isOffHeapAllocation' to 'isMmapAllocation' because both of them are off-heap.

mcvsubbu: i will add a comment here. The variables are derived from the configs. If offheap is not enabled, then we still need a mem manager for fwd index, but dictionaries are on-heap.

PR #2045, opened 2017-10-30
Fix the problem of directly memory leaking in scatter-gather

Jackie-Jiang:
Whenever we get the response from the server, read it into a byte[] and directly release the ByteBuf
In normal case, this won't add overhead because when deserializing the response into a DataTable, we need to first read it into a byte[]

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
@@ -47,7 +47,6 @@ import com.linkedin.pinot.transport.common.CompositeFuture;
 import com.linkedin.pinot.transport.scattergather.ScatterGather;
 import com.linkedin.pinot.transport.scattergather.ScatterGatherRequest;
 import com.linkedin.pinot.transport.scattergather.ScatterGatherStats;
-import io.netty.buffer.ByteBuf;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


pinot-transport/src/test/java/com/linkedin/pinot/transport/perf/ScatterGatherPerfClient.java
```diff
@@ -297,39 +295,20 @@ public SimpleScatterGatherRequest getRequest() throws IOException, JSONException
...
-    if (null != bMap) {
-      for (Entry<ServerInstance, ByteBuf> bEntry : bMap.entrySet()) {
-        ByteBuf b = bEntry.getValue();
-        if (null != b) {
-          b.release();
-        }
-      }
+    CompositeFuture<byte[]> future = _scatterGather.scatterGather(request, scatterGatherStats, brokerMetrics);
+    byte[] bytes = future.getOne();
+    if (bytes == null) {
```
mcvsubbu: bytes == null || bytes.length == 0

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -581,15 +580,15 @@ private BrokerResponse processOptimizedBrokerRequests(@Nonnull BrokerRequest ori
...
     try {
-      Map<ServerInstance, ByteBuf> serverResponseMap = compositeFuture.get();
-      Iterator<Entry<ServerInstance, ByteBuf>> iterator = serverResponseMap.entrySet().iterator();
+      Map<ServerInstance, byte[]> serverResponseMap = compositeFuture.get();
+      Iterator<Entry<ServerInstance, byte[]>> iterator = serverResponseMap.entrySet().iterator();
       while (iterator.hasNext()) {
-        Entry<ServerInstance, ByteBuf> entry = iterator.next();
-        if (entry.getValue().readableBytes() == 0) {
+        Entry<ServerInstance, byte[]> entry = iterator.next();
+        if (entry.getValue().length == 0) {
```
mcvsubbu: not sure if you want entry.getValue() == null test as well here?

pinot-transport/src/test/java/com/linkedin/pinot/transport/perf/ScatterGatherPerfClient.java
Jackie-Jiang: empty byte array could be a valid response

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
Jackie-Jiang: It can never be null here. If it is null, then just let it throw NPE

PR #2048, opened 2017-10-30
Handle consumder dir not configured, and cleanup old consuming mmap files if any

mcvsubbu:
If the consumer directory is not configured, when we create the consumer dir in some
unknown place (depending on the path env of the application). Changed it to be under the datadir.

Also, it is possible that segments may be re-assigned after a server restart, causing old
mmaped segments to remain (if the last shutdown was a hard kill of the process).
It is best to clean out all mmap files under the consumer directory when the table data manager
is inited.


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -80,6 +82,25 @@ protected void doInit() {
...
     }
+
+    String consumerDirPath = getConsumerDir();
+    File consumerDir = new File(consumerDirPath);
+
+    if (consumerDir.exists()) {
+      File[] segmentFiles = consumerDir.listFiles(new FilenameFilter() {
+        @Override
+        public boolean accept(File dir, String name) {
+          return ! name.equals(STATS_FILE_NAME);
```
Jackie-Jiang: (minor) remove the space

PR #2052, opened 2017-11-01
Do not put trace info in DataTable when trace is not enabled

Jackie-Jiang:
Fix the issue where trace info is set in DataTable even when trace is not enabled
When encountered exception, still put request id into data table
Will clean up Trace in another PR for performance

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.core.query.executor;
 
+import com.google.common.base.Preconditions;
 import com.linkedin.pinot.common.data.DataManager;
 import com.linkedin.pinot.common.exception.QueryException;
 import com.linkedin.pinot.common.metrics.ServerMeter;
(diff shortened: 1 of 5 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -90,82 +87,73 @@ public void init(Configuration configuration, DataManager dataManager, ServerMet
...
+    if (enableTrace) {
       TraceContext.register(instanceRequest);
-      final BrokerRequest brokerRequest = instanceRequest.getQuery();
-      LOGGER.debug("Incoming query is : {}", brokerRequest);
+    }
+    BrokerRequest brokerRequest = instanceRequest.getQuery();
+    LOGGER.debug("Incoming request Id: {}, query: {}", requestId, brokerRequest);
+    String tableName = brokerRequest.getQuerySource().getTableName();
+    TableDataManager tableDataManager = _instanceDataManager.getTableDataManager(tableName);
+    Preconditions.checkState(tableDataManager != null, "Failed to find data manager for table: {}", tableName);
```
mcvsubbu: All logic and pre-conditions should be inside the try/catch block, so that we return a correct dataTable to the broker and not an exception trace.

Jackie-Jiang: We have another level of exception handling (unhandled exception) on the caller. On this level, this part should not throw exception

PR #2054, opened 2017-11-01
Simplify the anomaly function implementation

cyy0714:
When designing an anomaly detection function, there are some duplicate codes in the new function. These function are redundant and easy to forget, especially useHistoryAnomaly(). To simplify the anomaly function in TE and reduce the errors from faulty configuration, we add the common variables and functions in the AbstractModularizedAnomalyFunction.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/function/AbstractModularizedAnomalyFunction.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/function/AbstractModularizedAnomalyFunction.java
@@ -4,9 +4,15 @@ import com.linkedin.pinot.pql.parsers.utils.Pair;
 import com.linkedin.thirdeye.anomaly.views.AnomalyTimelinesView;
 import com.linkedin.thirdeye.anomalydetection.context.AnomalyDetectionContext;
 import com.linkedin.thirdeye.anomalydetection.context.TimeSeries;
+import com.linkedin.thirdeye.anomalydetection.model.data.DataModel;
+import com.linkedin.thirdeye.anomalydetection.model.data.NoopDataModel;
+import com.linkedin.thirdeye.anomalydetection.model.detection.DetectionModel;
+import com.linkedin.thirdeye.anomalydetection.model.detection.NoopDetectionModel;
 import com.linkedin.thirdeye.anomalydetection.model.merge.MergeModel;
 import com.linkedin.thirdeye.anomalydetection.model.merge.NoPredictionMergeModel;
+import com.linkedin.thirdeye.anomalydetection.model.merge.NoopMergeModel;
 import com.linkedin.thirdeye.anomalydetection.model.prediction.ExpectedTimeSeriesPredictionModel;
+import com.linkedin.thirdeye.anomalydetection.model.prediction.NoopPredictionModel;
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/function/AbstractModularizedAnomalyFunction.java
```diff
@@ -270,4 +284,39 @@ public AnomalyTimelinesView getTimeSeriesView(MetricTimeSeries timeSeries,
     return this.getTimeSeriesView(anomalyDetectionContext, bucketMillis, metric, viewWindowStartTime, viewWindowEndTime,
         knownAnomalies);
   }
+
+  @Override
```
kishoreg: whats the contract on these methods. Please add java docs for these methods. Are these methods called only once or multiple times. Can the implementations override these methods. Can they return a new instance every time getXModel is invoked?.

If this is one time, then I suggest renaming them to init or create or setup - something that indicates that this will be called only once when the Function is being setup.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/function/WeekOverWeekRuleFunction.java
```diff
@@ -157,27 +150,8 @@ public static String parseWowString(String wowString) {
...
-
-  @Override public DetectionModel getDetectionModel() {
-    return detectionModel;
-  }
-
-  @Override public MergeModel getMergeModel() {
-    return mergeModel;
+  @Override
+  public boolean useHistoryAnomaly() {
+    return false;
```
cecilynie: Is it a conflict that if one anomaly function defines "AnomalyRemoval" functionality in  Transformation chain, but useHistoryAnomaly() is false. Or if use the new default value of "useHistoryAnomaly()" to be true, but failed to define any "AnomalyRemoval" strategy. Shall we come up with a way to finally unify "useHistoryAnomaly()" and "AnomalyRemoval()" in transformation chain?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/function/AbstractModularizedAnomalyFunction.java
cyy0714: Thanks for the comments. I'll add java docs to these methods. This methods return a user-defined models. These methods are usually called only once. They return the same instance every time they are called. The implementation can override these methods, but in realistic, they are seldom override by the implementations. That's why I want to move it forward to the abstract class. As the methods only return the pre-setup models (the models are created and initialized in init()), using init may disguise these methods.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/function/WeekOverWeekRuleFunction.java
cyy0714: This is a good question. We can examine the transformation functions and figure out if useHitoryAnomaly() should be true. But it can lead to an ad-hoc design that if there is new model requiring history anomalies, we don't know that. We can discuss a more systematic solution for this.

PR #2056, opened 2017-11-01
Failing segment creation with invalid inverted indices

jenniferdai:
* Segment generation can fail on servers due to inverted indices being too large, and we want a way to catch this early. We can fail during segment generation when estimated inverted index size exceeds 2GB.


pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
```diff
@@ -204,6 +204,23 @@ public void build() throws Exception {
...
+      numBytes += cardinality * indexCreationInfoMap.get(columnName).getLegnthOfLongestEntry();
+      LOGGER.info("Estimated inverted index size for column {} is {}, should be < 2GB", columnName, numBytes);
+      if (numBytes >= 2147483648L) {
+        return true;
+      }
+    }
+    return false;
+  }
+
   private void buildStarTree() throws Exception {
```
kishoreg: why is this part of buildStarTree method. Its better to call validate method in seal method and keep all the validation in one place.

PR #2057, opened 2017-11-01
Refactor Trace for performance concern

Jackie-Jiang:
Add API traceEnabled() into TraceContext
When trace is not enabled, do not log time in BaseOperator
Remove redundant thread local objects


pinot-core/src/main/java/com/linkedin/pinot/core/util/trace/Trace.java
```diff
@@ -17,79 +17,52 @@
...
 
-  private String _traceId;
-  private List<String> _key = new ArrayList<String>();
-  private List<Object> _value = new ArrayList<Object>();
-  private Long _threadId;
-  protected final Trace _parent;
-  protected final List<Trace> _children = new ArrayList<Trace>();
+    @Override
+    public String toString() {
+      return _key + ": " + _value;
```
mcvsubbu: Need to quote it to get the right json, otherwise we cannot parse it using standard json parsers.

PR #2059, opened 2017-11-02
Add support for datetime column in realtime segment generation

npawar:
Tested on a test cluster with datetime column setup in the time column name of table config. Verified that the segments generated contain datetime column, and it is queryable on the pinot console

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/RealtimeSegmentConverter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/RealtimeSegmentConverter.java
@@ -66,7 +66,9 @@ public class RealtimeSegmentConverter {
     for (String metric : schema.getMetricNames()) {
       newSchema.addField(schema.getFieldSpecFor(metric));
     }
-
+    for (String dateTime : schema.getDateTimeNames()) {
+      newSchema.addField(schema.getFieldSpecFor(dateTime));
+    }
     newSchema.addField(newTimeSpec);
     this.realtimeSegmentImpl = realtimeSegment;
     this.outputPath = outputPath;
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/RealtimeSegmentConverter.java
```diff
@@ -66,7 +66,9 @@ public RealtimeSegmentConverter(RealtimeSegmentImpl realtimeSegment, String outp
     for (String metric : schema.getMetricNames()) {
       newSchema.addField(schema.getFieldSpecFor(metric));
     }
-
+    for (String dateTime : schema.getDateTimeNames()) {
```
kishoreg: will this include the old time column as well?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/RealtimeSegmentRecordReader.java
```diff
@@ -55,6 +55,7 @@ public void init() throws Exception {
     columns.addAll(dataSchema.getDimensionNames());
     columns.addAll(dataSchema.getMetricNames());
     columns.add(dataSchema.getTimeFieldSpec().getOutgoingTimeColumnName());
+    columns.addAll(dataSchema.getDateTimeNames());
```
kishoreg: we should deprecate usage of getTimeFieldSpec right, else we might be adding the same column twice since getDateTimeNames might return timefieldspec as well?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -347,6 +376,9 @@ private void storeIncomingColumnValue(GenericRow row, Map<String, Object> rawRow
         case DOUBLE:
           readerWriter.setDouble(docId, (double)row.getValue(columnName));
           break;
+        case STRING:
```
kishoreg: why is this part of this RB?

PR #2064, opened 2017-11-06
[TE] frontend - implement toggle filter bar

ttbach:
filter-block-toggle

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/rca-poc/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rca-poc/route.js
@@ -12,6 +12,12 @@ export default Ember.Route.extend({
     };
   },
 
+  /**
+   * sets the model and filterBlocks in the controller
+   * sets filterBlocks to be built based on filter bar config
+   * @param {Object} controller
+   * @param {Object} model
+   */
   setupController(controller, model) {
     const filterBlocks = model.filterBarConfig;
     controller.setProperties({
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-frontend/app/pods/rca-poc/route.js
```diff
@@ -10,5 +10,19 @@ export default Ember.Route.extend({
...
+  },
+
+  /**
+   * sets the model and filterBlocks in the controller
+   * sets filterBlocks to be built based on filter bar config
+   * @param {Object} controller
+   * @param {Object} model
+   */
+  setupController(controller, model) {
+    const filterBlocks = model.filterBarConfig;
```
justYves: Other than for semantic purpose, do we need to do this in `setupController` here? would a `Ember.alias('model.filterblocks')` in the controller work the same way ? 

also (nitpick): do we still need the filterBarConfig on the model? If not, we can delete it from filterBlocks.

ttbach: We don't need filterBarConfig in the model. I'll remove it. I have it in `setUpController`, so it's available in the component `init()` hook. I'm not sure if this is the best way to do it. Any suggestions?

justYves: Assuming we don't use `setupController`, wouldn't it still be available through `this.get('model.filterblocks')` in the init hook anyway?

PR #2067, opened 2017-11-07
[TE] Create tables and DAOs for alert email refactoring

cyy0714:
We are working on the redesign of alerts. The first pr is to build up the database environment to generate and enable customized logic on alerts. The following pr will be pushed in to 1) introduce anomaly re-alert logic, and 2) enable email customization.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/alert/commons/AnomalySource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/alert/commons/AnomalySource.java
@@ -1,0 +1,31 @@
+package com.linkedin.thirdeye.alert.commons;
+
+import com.linkedin.thirdeye.datalayer.util.Predicate;
+
+
+public enum AnomalySource {
+  DATASET {
+    @Override
+    public Predicate getPredicate(String predicateValue) {
+      return Predicate.IN("collection", predicateValue.split(","));
+    }
+  },
+  METRIC {
+    @Override
+    public Predicate getPredicate(String predicateValue) {
+      return Predicate.IN("metric", predicateValue.split(","));
+    }
+  },
+  ANOMALY_FUNCTION {
+    @Override
+    public Predicate getPredicate(String predicateValue) {
+      return Predicate.IN("functionId", predicateValue.split(","));
+    }
+  };
+
+  AnomalySource(){
+
+  }
+
+  public abstract Predicate getPredicate(String predicateValue);
+}
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/util/StringUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/util/StringUtils.java
@@ -15,8 +15,11 @@ public class StringUtils {
   private static Joiner SEMICOLON = Joiner.on(";");
   private static Joiner EQUALS = Joiner.on("=");
 
-  // the following two functions encode and decode properties are moved from class
-  // com.linkedin.thirdeye.controller.mp.function.ThirdEyeAnomalyFunctionUtil
+  /**
+   * Encode the properties into String, which is in the format of field1=value1;field2=value2 and so on
+   * @param props the properties instance
+   * @return a String representation of the given properties
+   */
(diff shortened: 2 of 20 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/util/StringUtils.java
```diff
@@ -0,0 +1,36 @@
...
+public class StringUtils {
+  private static Splitter SEMICOLON_SPLITTER = Splitter.on(";").omitEmptyStrings();
+  private static Splitter EQUALS_SPLITTER = Splitter.on("=").omitEmptyStrings();
+
+  private static Joiner SEMICOLON = Joiner.on(";");
+  private static Joiner EQUALS = Joiner.on("=");
+
+  // the following two functions encode and decode properties are moved from class
+  // com.linkedin.thirdeye.controller.mp.function.ThirdEyeAnomalyFunctionUtil
+  public static String encodeCompactedProperties(Properties props) {
```
apucher: shouldn't this use a standard serializer, like jaxson?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/alert/commons/AnomalySource.java
```diff
@@ -0,0 +1,31 @@
+package com.linkedin.thirdeye.alert.commons;
+
+import com.linkedin.thirdeye.datalayer.util.Predicate;
+
+
+public enum AnomalySource {
+  DATASET {
+    @Override
+    public Predicate getPredicate(String predicateValue) {
```
apucher: This seems domain specific to database fetching. you may want to keep this out of the enum and instead have an enum->predicate map where it actually applies.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/util/StringUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class StringUtils {
+  private static Splitter SEMICOLON_SPLITTER = Splitter.on(";").omitEmptyStrings();
+  private static Splitter EQUALS_SPLITTER = Splitter.on("=").omitEmptyStrings();
+
+  private static Joiner SEMICOLON = Joiner.on(";");
+  private static Joiner EQUALS = Joiner.on("=");
+
+  // the following two functions encode and decode properties are moved from class
+  // com.linkedin.thirdeye.controller.mp.function.ThirdEyeAnomalyFunctionUtil
```
cyenjung: Remove the class name. Moreover, give us an example of this method's input and output.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/alert/commons/AnomalySource.java
cyy0714: Do you suggest to have a pure enum, and determine the predicate on the other side? May I know  the reason behind the choice? Thanks for the comment. I am curious about the choice.

thirdeye/thirdeye-pinot/src/test/resources/schema/create-schema.sql
```diff
@@ -334,3 +334,9 @@ create table application_index (
   recipients VARCHAR(1000) NOT NULL,
   base_id bigint(20) not null
 );
+
+create table if not exists alert_snapshot_index (
+    base_id bigint(20) not null,
+    create_time timestamp,
+    update_time timestamp default current_timestamp
```
cyenjung: I think there is one more field: version int(10).

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/util/StringUtils.java
cyy0714: I can try using serializer.

cyy0714: As the format of the output string is like field1=value1;field2=value2. The jaxson cannot recognize this format. This format is shared with all the other properties in the current design.

PR #2077, opened 2017-11-08
[TE] Self-Serve Replay Wrapper Logic Improvment

cecilynie:
We occasionally see replayed notification email contains number of anomalies replayed but to generate link contains the list of anomalies. After investigation see email notifications only send out "notified=true" anomalies. Currently this is a fix to skip the check of "notified" flag for replayed anomalies. 
1. Fix hourly replay no anomalies sent out
2. Harden exception logic

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
@@ -253,7 +253,7 @@ public class DetectionJobResource {
       Map<Long, Long> entity = (Map<Long, Long>) response.getEntity();
       jobId = entity.get(id);
     } catch (Exception e) {
-      return Response.ok("Failed to start replay!").build();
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -285,13 +285,12 @@ public Response triggerReplayTuningAndNotification(@PathParam("id") @NotNull fin
...
     }
 
-
     // send out email
     String subject = new StringBuilder(
         "Replay results for " + anomalyFunctionDAO.findById(id).getFunctionName() + " is ready for review!").toString();
 
     emailResource.generateAndSendAlertForFunctions(startTime, endTime, String.valueOf(id), fromAddr, toAddr, subject,
-        true, true, teHost, smtpHost, smtpPort, phantomJsPath);
+        false, true, teHost, smtpHost, smtpPort, phantomJsPath);
```
bl44: Thanks for the fix.

PR #2084, opened 2017-11-09
[TE] rootcause poc - add legend component

apucher:


thirdeye/thirdeye-frontend/app/pods/components/rootcause-legend/component.js
```diff
@@ -0,0 +1,40 @@
...
+  actions: {
+    toggleVisibility(urn) {
+      const { onVisibility, invisibleUrns } = this.getProperties('onVisibility', 'invisibleUrns');
+      if (onVisibility) {
+        const state = invisibleUrns.has(urn);
+        onVisibility(urn, state);
+      }
+    },
+    removeUrn(urn) {
+      const { onSelection } = this.getProperties('onSelection');
```
ttbach: Non-blocking: since you're only getting 1 property, you can just use `this.get('onSelection')`

apucher: I'm inconsistently consistent :)

PR #2090, opened 2017-11-09
[TE] Implement Anomaly Fetcher

cyy0714:
The first part of alert redesign is to introduce anomaly fetchers into ThirdEye. Anomaly fetcher is to fetch anomaly candidates based on time criteria. The time criteria comes from AlertSnapshotDTO. Two types of anomaly fetchers are implemented: Unnotified and Continuum. The unnotified anomaly fetcher fetches anomaly not yet notified before; that is, its create time is after last notified time. The continuum anomaly fetcher fetches anomalies that are notified but still in progress; the fetcher examines the end time of the anomaly and check if it is after last notified time. With the anomaly fetchers, we can introduce anomaly re-alert logic into our alerts.


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/common/BaseFactory.java
```diff
@@ -0,0 +1,13 @@
+package com.linkedin.thirdeye.common;
+
+public class BaseFactory<T> {
```
cyenjung: 1. Please format the code.
2. Should BaseFactory considering package name when creating an instance? What about making base factory creates instances from any full class names. The factory that extends the base factory could override the method to inject package name if necessary.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/common/FactorySource.java
```diff
@@ -0,0 +1,15 @@
...
+
+public enum FactorySource {
+  ALERT_FETCHER("com.linkedin.thirdeye.alert.fetcher"),
+  ANOMALY_FEED("com.linkedin.thirdeye.alert.feed"),
+  EMAIL_CONTENT_FORMATTER("com.linkedin.thirdeye.alert.content");
+
+  private String packagePath;
+  FactorySource(String packagePath) {
+    this.packagePath = packagePath;
+  }
```
cyenjung: Need a space before each method.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/common/FactorySource.java
```diff
@@ -0,0 +1,15 @@
+package com.linkedin.thirdeye.common;
+
+public enum FactorySource {
```
cyenjung: Add a comment to indicate that these strings are package paths.

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/alert/commons/TestAnomalyFetcherFactory.java
```diff
@@ -0,0 +1,21 @@
...
+import com.linkedin.thirdeye.alert.fetcher.UnnotifiedAnomalyFetcher;
+import org.testng.annotations.Test;
+
+
+public class TestAnomalyFetcherFactory {
+
+  @Test
+  public void testCreateAlertFetcher() throws Exception{
+    AnomalyFetcher anomalyFetcher = AnomalyFetcherFactory.fromClassName("ContinuumAnomalyFetcher");
+    assert anomalyFetcher != null;
```
cyenjung: Use TestNG's Assert.notNull instead of Java's assert, which might stop the entire program if the thrown exception is not properly handled.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/common/FactorySource.java
```diff
@@ -0,0 +1,15 @@
...
+  ALERT_FETCHER("com.linkedin.thirdeye.alert.fetcher"),
+  ANOMALY_FEED("com.linkedin.thirdeye.alert.feed"),
+  EMAIL_CONTENT_FORMATTER("com.linkedin.thirdeye.alert.content");
```
cyenjung: I am not sure if hard-code the path in enum is a good idea. Considering that we might have extension jar (e.g., MP) which may put the additional classes in different packages.

We should inject this class when the ThirdEye application is initialized. In that way, we could control if we want to add extension packages.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/alert/fetcher/ContinuumAnomalyFetcher.java
```diff
@@ -0,0 +1,89 @@
...
+    super();
+  }
+
+  /**
+   * Fetch continuum anomalies whose end time is after last notify time; that is, the anomaly continues after last alert
+   * @param current current DateTime
+   * @param alertSnapShot the snapshot of the AnomalyFeed
+   * @return a list of continuing merged anomalies
+   */
+
```
cyenjung: Why this additional space?

PR #2091, opened 2017-11-09
[TE] Yyuen/rca poc summary settings logic

justYves:
### What's new: 
- passing query params into the settings component and updating the controller's property and query params on change


thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -1,21 +1,105 @@
...
 
-    // TODO get initial attributes from query params
     controller.setProperties({
       selectedUrns: new Set(['thirdeye:metric:194592', 'thirdeye:event:holiday:2712391']),
       invisibleUrns: new Set(),
       filteredUrns: new Set(),
       hoverUrns: new Set(),
-      context: { urns, anomalyRange, baselineRange, analysisRange }
+      context: { urns, anomalyRange, baselineRange, analysisRange },
+      ...model.queryParams
```
apucher: The controller should only handle the context object. All the complexity of converting between query-params and context, and display options and context should be hidden in the route and the components - i.e. "context" is the interface.

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -1,21 +1,105 @@
...
+  model(params) {
+    const { rootcauseId: id } = params;
+
+    // TODO handle error better
+    if (!id) { return; }
+
+    return RSVP.hash({
+      granularityOptions: fetch(`/data/agg/granularity/metric/${id}`).then(res => res.json()),
+      filterOptions: fetch(`/data/autocomplete/filters/metric/${id}`).then(res => res.json()),
+      maxTime: fetch(`/data/maxDataTime/metricId/${id}`).then(res => res.json()),
```
apucher: let's not worry about fetching this data yet. This will be handled by asynchronous services.

justYves: This data needs to be fetched as it prepopulates the data for the rootcause settings component. This is only called once when a route is visited. Let me know how else you'd like to handle this

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
justYves: In order to use queryparams, those needs to be defined in both the route and controller, because the queryparams are only stored in a one level object, we would need some sort of conversion. Since the `queryparams` only update the url but not the model, the model hooks and setupController are not called. Therefore the only place to have this conversion logic is in the setupController on init and the `onChange` handler passed into the component. 
Let me know if you want to handle this another way

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -50,10 +72,53 @@ export default Ember.Controller.extend({
...
+    const model = this.get('model');
+    return getProperties(model, ['granularityOptions', 'filterOptions', 'maxTime', 'compareModeOptions']);
+  }),
+
+  /**
+   * Convert a  query value into option vlaue
+   * @param {String} key          - query param key
+   * @param {String|Object} value - query param value
+   */
+  queryToOption(key, value) {
```
justYves: fyi this is not used in the code. Keeping it here in case we need to quickly implement a 2 way queryparams binding

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
justYves: @apucher  passing it down as 'configs' from the controller

apucher:
great. thanks for updating!

PR #2092, opened 2017-11-10
[TE] frontend - populate filter bar dropdown on /rootcause

ttbach:
- Created a new `filter-bar-input` component
filter-bar-options

- Multi-select filter by "country" now available
screen shot 2017-11-13 at 2 30 31 pm


thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -488,6 +490,70 @@ export default Ember.Controller.extend({
...
+   *    countryCode: ['US', 'IN', 'MX'],
+   *    region: ['North America', 'Antarctica']
+   *  },
+   *  deployment: {
+   *    fabric: ['prod', 'local']
+   *  }
+   * }
+   * @type {Object}
+   */
+  attributesMap: Ember.computed(
```
apucher: this should be encapsulated within filter-bar. entities go in, a set of urns comes out via onSelection action - i.e. "entities" and "onSelection" is the interface

PR #2095, opened 2017-11-11
[TE] user anomaly reporting enpoint

apucher:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/AnomaliesResource.java
```diff
@@ -464,6 +467,69 @@ public void updateAnomalyMergedResultFeedback(@PathParam("mergedAnomalyId") long
...
+
+    if (endTime == null) {
+      throw new IllegalArgumentException("Must provide endTime");
+    }
+
+    if (feedbackType == null) {
+      throw new IllegalArgumentException("Must provide feedbackType");
+    }
+
+    MergedAnomalyResultDTO anomaly = new MergedAnomalyResultDTO();
```
cecilynie: Hi Alex, thanks for implementing this endpoint to take user-report anomaly. If there is already one (or more) anomaly overlap with user provided start and end time, shall we recover those as "user-report"? It will help on correcting tuning's training samples. Thanks.

apucher: Yes, I remember that discussion. IMO we should decide recovery/merge/ignore when we prepare the data for auto-tuning. Otherwise, we risk losing the user input irrevocably by mingling it with algo outputs.

cecilynie: Let me correct the purpose of having the match. Users' input time range can be arbitrary and can be interpreted as a guideline. In most of the case the start and end time of anomaly that user input can go beyond the region that the algorithm specify. User can have a rough impression to report (since in fact they don't usually know the accurate start time as well), thus we don't necessarily to take it as an accurate input. To make sure algorithm's anomaly removal + baseline prediction running properly, we need to 1) give user space to input anomalous region; 2) match user input region to existing anomalies (these anomalies are detected by the algorithm but just filtered out by alert-filter), following up with anomaly detection running, anomaly removal will be correctly done using the matched anomalies; 3) If no matches found,  run algorithm model selection (currently under design) to fully recover the user report. 

Overall, with user report anomaly, this is the only place to get matches, otherwise algorithm cannot pick up correct anomalies for removal, as well as auto-adjust the performance of algorithm.

Thanks!

apucher: as per our discussion: https://github.com/linkedin/pinot/pull/2119

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/AnomaliesResource.java
```diff
@@ -464,6 +467,69 @@ public void updateAnomalyMergedResultFeedback(@PathParam("mergedAnomalyId") long
...
+    if (mergedAnomalyResultDAO.save(anomaly) == null) {
+      throw new IllegalArgumentException(String.format("Could not store user reported anomaly: '%s'", anomaly));
+    }
+
+    // TODO fix feedback not being saved on create by DAO
+    AnomalyFeedbackDTO feedback = new AnomalyFeedbackDTO();
+    feedback.setFeedbackType(feedbackType);
+    feedback.setComment(comment);
+
+    anomaly.setFeedback(feedback);
```
cecilynie: Hi Alex, what's the "notified" flag for "user-report"? Here are two use cases:
1. In evaluation module, it relies on "isNotified=false" to make classify user-report anomalies
2. In search page (alert page), when showing anomalies is depending on "notified = true", but user report anomalies should also be shown on the page. You probably need to edit the logic of showing searched anomalies. (including user report, replayed anomalies)

apucher: @cecilynie yes this is another thing to change. it's for a separate PR.

PR #2097, opened 2017-11-13
Add support for dimension datatypes other than string

npawar:
Currently we had the assumption that dimensions are necessarily string. This PR adds support for dimension to be of types other than string. In topk phase, we used to put "other" for non topk non whitelist dimension values. Now that we support dimension types other than string, the others value can be provided by the user via config.


thirdeye/thirdeye-hadoop/src/main/java/com/linkedin/thirdeye/hadoop/aggregation/AggregationPhaseMapOutputKey.java
```diff
@@ -56,22 +64,45 @@ public long getTime() {
...
+      case INT:
+        dos.writeInt((int) dimension);
+        break;
+      case SHORT:
+        dos.writeShort((short) dimension);
+        break;
+      case LONG:
+        dos.writeLong((long) dimension);
+        break;
+      case STRING:
```
kishoreg: have a default and throw exception if we cannot handle the data type

thirdeye/thirdeye-hadoop/src/main/java/com/linkedin/thirdeye/hadoop/aggregation/AggregationPhaseMapOutputKey.java
```diff
@@ -56,22 +64,45 @@ public long getTime() {
...
-      bytes = dimension.getBytes();
-      dos.writeInt(bytes.length);
-      dos.write(bytes);
+    for (int i = 0; i < dimensions.size(); i++) {
+      Object dimension = dimensions.get(i);
+      DimensionType dimensionType = dimensionTypes.get(i);
+
+      switch (dimensionType) {
+      case DOUBLE:
+        dos.writeDouble((double) dimension) ;
```
kishoreg: We should be able to support conversion here right. There must be some method in Pinot to convert something from one datatype to another. We have two cases its either a String or Number. If its Number then use the right method to convert the data to appropriate type https://docs.oracle.com/javase/7/docs/api/java/lang/Number.html. If it's String always use the Integer.parseInt or Long.parseLong based on the data type. This will handle cases where each avro file might have a different data type for the same input column

thirdeye/thirdeye-hadoop/src/main/java/com/linkedin/thirdeye/hadoop/topk/TopKPhaseMapOutputKey.java
```diff
@@ -51,11 +59,36 @@ public String getDimensionValue() {
...
-    // dimension value
-    bytes = dimensionValue.getBytes();
+
+    // dimension type
+    bytes = dimensionType.toString().getBytes();
     dos.writeInt(bytes.length);
     dos.write(bytes);
 
+    // dimension value
+    switch (dimensionType) {
```
kishoreg: avoid this code duplication if possible

npawar:
Handled default case, and removed duplicate code

npawar:
Tested on hadoop for flows aggregation, topk, derived_column_transformation, segment_creation and segment_push
Tested cases:
1. No threshold, topk, whitelist - with all string dimensions, with 1 long dimension
2. Threshold
3. Topk - with all string dimensions in topk, with a long dimension in topk
4. Whitelist - With default other values, with configured other values

PR #2099, opened 2017-11-14
Remove BlockId class and nextBlock(BlockId) API in Operator

Jackie-Jiang:
1. Remove BlockId class
2. Remove nextBlock(BlockId) API in Operator because we should never index on block
3. Add the type of block to Operator for clarity
4. Change getNextBlock() API in BaseOperator to be protected so that we always call nextBlock() for tracking purpose
NOTE: inside filter operator, nextBlock() should be called only once


pinot-core/src/main/java/com/linkedin/pinot/core/operator/BaseOperator.java
```diff
@@ -26,42 +25,25 @@
...
-      return ret;
-    } else {
-      return getNextBlock(blockId);
-    }
-  }
-
-  public abstract Block getNextBlock();
-
-  public abstract Block getNextBlock(BlockId blockId);
+  protected abstract T getNextBlock();
```
snleee: Can we add the documentation on the reason why you put this as `protected`? I see that you put the explanation on the pr message but it would be helpful to put a documentation here.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/CompositeOperator.java
```diff
@@ -15,13 +15,11 @@
...
-import java.util.List;
-
-import com.linkedin.pinot.core.common.BlockId;
 import com.linkedin.pinot.core.common.Operator;
 import com.linkedin.pinot.core.operator.blocks.BaseFilterBlock;
 import com.linkedin.pinot.core.operator.blocks.CompositeBaseFilterBlock;
+import java.util.ArrayList;
+import java.util.List;
 
 public class CompositeOperator extends BaseFilterOperator{
```
snleee: Space ? `BaseFilterOperator {`

Jackie-Jiang: This class will be removed in the star-tree PR, but good catch

pinot-core/src/main/java/com/linkedin/pinot/core/operator/BaseOperator.java
Jackie-Jiang: Added.

PR #2102, opened 2017-11-14
Add DictionaryBasedAggregationOperator to support for max, min, minmaxrange

npawar:
min, max, minmaxrange queries can be answered by looking at the dictionary, for those columns that have one


pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/MetadataBasedAggregationOperator.java
```diff
@@ -46,23 +49,17 @@
    *
    * @param aggregationFunctionContexts Aggregation function contexts.
    * @param segmentMetadata Segment metadata.
-   * @param dataSourceMap Map of column to its data source.
+   * @param dictionaryMap Map of column to its dictionary.
    */
   public MetadataBasedAggregationOperator(AggregationFunctionContext[] aggregationFunctionContexts,
-      SegmentMetadata segmentMetadata, Map<String, BaseOperator> dataSourceMap) {
+      SegmentMetadata segmentMetadata, Map<String, Dictionary> dictionaryMap) {
```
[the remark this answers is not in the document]
kishoreg: This operator is based on Dictionary and it probably makes sense to have DictionaryBasedAggregationOperator instead of overloading MetadataBasedAggregationOperator

pinot-core/src/main/java/com/linkedin/pinot/core/plan/maker/InstancePlanMakerImplV2.java
```diff
@@ -141,13 +141,17 @@ private boolean isFitForMetadataBasedPlan(BrokerRequest brokerRequest) {
     }
 
     List<AggregationInfo> aggregationsInfo = brokerRequest.getAggregationsInfo();
-    if (aggregationsInfo != null && aggregationsInfo.size() == 1 && !brokerRequest.isSetGroupBy()) {
-      if (aggregationsInfo.get(0)
-          .getAggregationType()
-          .equalsIgnoreCase(AggregationFunctionFactory.AggregationFunctionType.COUNT.getName())) {
-        return true;
+    if (aggregationsInfo != null && !brokerRequest.isSetGroupBy()) {
```
kishoreg: These are bunch of optimizers, we need to have a better framework, instead of calling isFitForStarTree, isFitForMetadata, isFitForDictionary. Not really part of this PR but please create an issue to track this later.

pinot-core/src/main/java/com/linkedin/pinot/core/plan/maker/InstancePlanMakerImplV2.java
```diff
@@ -141,13 +141,17 @@ private boolean isFitForMetadataBasedPlan(BrokerRequest brokerRequest) {
...
+      for (int i = 0; i < aggregationsInfo.size(); i++) {
+        String aggregationType = aggregationsInfo.get(i).getAggregationType();
```
kishoreg: just checking for function name is not sufficient. We also need to know if the column has dictionary. Also, we are not really using metadata to answer these queries, so we should have separate method to check if the query can dictionary.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/DictionaryBasedAggregationOperator.java
```diff
@@ -0,0 +1,115 @@
...
+  private ExecutionStatistics _executionStatistics;
+
+  /**
+   * Constructor for the class.
+   *
+   * @param aggregationFunctionContexts Aggregation function contexts.
+   * @param segmentMetadata Segment metadata.
+   * @param dictionaryMap Map of column to its dictionary.
+   */
+  public DictionaryBasedAggregationOperator(AggregationFunctionContext[] aggregationFunctionContexts,
```
mayankshriv: It seems the approach here was to not use MetadataBasedOperator, and if so, then we should not need metadata here. If the only need for that is to get totalRawDocs, may be we should just pass that?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/DictionaryBasedAggregationOperator.java
```diff
@@ -0,0 +1,115 @@
...
+
+    for (AggregationFunctionContext aggregationFunctionContext : _aggregationFunctionContexts) {
+      AggregationFunction function = aggregationFunctionContext.getAggregationFunction();
+      AggregationFunctionFactory.AggregationFunctionType functionType =
+          AggregationFunctionFactory.AggregationFunctionType.valueOf(function.getName().toUpperCase());
+      String column = aggregationFunctionContext.getAggregationColumns()[0];
+      Dictionary dictionary = _dictionaryMap.get(column);
+      AggregationResultHolder resultHolder;
+      switch (functionType) {
+
```
mayankshriv: Auto-format using IDE

pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/DictionaryBasedAggregationOperator.java
```diff
@@ -0,0 +1,115 @@
...
+        case MAX:
+          resultHolder = new DoubleAggregationResultHolder(dictionary.getDoubleValue(dictionary.length() - 1));
```
mayankshriv: Will this work with star-tree where dictionary is modified to include the ALL node? If not, then either we should not select this operator in the plan phase, or handle it here (prefer latter if it can be cleanly achieved).

pinot-core/src/main/java/com/linkedin/pinot/core/plan/maker/InstancePlanMakerImplV2.java
```diff
@@ -92,6 +95,8 @@ public PlanNode makeInnerSegmentPlan(IndexSegment indexSegment, BrokerRequest br
           // Aggregation only query.
         if (isFitForMetadataBasedPlan(brokerRequest)) {
           return new MetadataBasedAggregationPlanNode(indexSegment, brokerRequest.getAggregationsInfo());
+        } else if (isFitForDictionaryBasedPlan(brokerRequest, indexSegment)) {
```
mayankshriv: Not a blocker for this PR, but seems like the number of if's are increasing. May be better to have a factory that returns the correct plan node.

pinot-core/src/test/java/com/linkedin/pinot/core/plan/maker/InstancePlanMakerTest.java
```diff
@@ -0,0 +1,216 @@
...
+    entries.add(new Object[] {
+        "select count(*),min(column1) from testTable group by daysSinceEpoch", /*multiple aggregations with group by*/
+        AggregationGroupByPlanNode.class
+    });
+
+    return entries.toArray(new Object[entries.size()][]);
+  }
+
+  @Test(dataProvider = "isFitForPlanDataProvider")
+  public void testIsFitForMetadataBased(String query, boolean expectedIsFitForMetadata, boolean expectedIsFitForDictionary) {
```
mayankshriv: For the 'fit' guys, we should also check the query result.

pinot-core/src/test/java/com/linkedin/pinot/queries/InterSegmentAggregationSingleValueQueriesTest.java
```diff
@@ -48,7 +48,7 @@ public void testMax() {
     String query = "SELECT MAX(column1), MAX(column3) FROM testTable";
 
     BrokerResponseNative brokerResponse = getBrokerResponseForQuery(query);
-    QueriesTestUtils.testInterSegmentAggregationResult(brokerResponse, 120000L, 0L, 240000L, 120000L,
+    QueriesTestUtils.testInterSegmentAggregationResult(brokerResponse, 120000L, 0L, 0L, 120000L,
```
mayankshriv: May be add a comment on why num docs is expected to be zero.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/DictionaryBasedAggregationClusterIntegrationTest.java
```diff
@@ -0,0 +1,336 @@
...
+import com.linkedin.pinot.common.utils.CommonConstants;
+import com.linkedin.pinot.common.utils.ServiceStatus;
+import com.linkedin.pinot.core.indexsegment.generator.SegmentVersion;
+import com.linkedin.pinot.util.TestUtils;
+
+
+/**
+ * Integration test to check aggregation functions which use DictionaryBasedAggregationPlan
+ */
+public class DictionaryBasedAggregationClusterIntegrationTest extends BaseClusterIntegrationTestSet {
```
mayankshriv: Why does this need a separate cluster integration test? Could we not leverage the existing integration tests? Adding new integration tests that are setting up the cluster increase overall testing time.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/MetadataBasedAggregationClusterIntegrationTest.java
```diff
@@ -0,0 +1,173 @@
...
+import com.linkedin.pinot.common.utils.CommonConstants;
+import com.linkedin.pinot.common.utils.ServiceStatus;
+import com.linkedin.pinot.core.indexsegment.generator.SegmentVersion;
+import com.linkedin.pinot.util.TestUtils;
+
+
+/**
+ * Integration test to check aggregation functions which use MetadataBasedAggregationOperator
+ */
+public class MetadataBasedAggregationClusterIntegrationTest extends BaseClusterIntegrationTestSet {
```
mayankshriv: Same here. But if these two cannot be merged into existing one, may be they could be merged with each other.

Jackie-Jiang:
Haven't read the code, but this could potentially break for star-tree use case (max might return the aggregated number), please do check that.

kishoreg:
Good catch @Jackie-Jiang. Can we check for star tree existence and if it exists, let's use regular path. I remember we also have min and max in the metadata, but this might be available only in the newer segments. If it exists we can use the values from metadata. So the planning phase as the setup the right operator
- use metadata if it has min/max
- if no star tree and has a dictionary, use dictionary
- fall back to the default code path

Jackie-Jiang:
@kishoreg The min/max value in the metadata are extracted from dictionary, so maybe we should first check star-tree, then check metadata

npawar:
1) Addressed concerns about values in case of star tree:
-Use metadata for i) count* ii) min,max,minmaxrange if no star tree and present in metadata
-Use dictionary for i) min,max,minmaxrange if no startree
2) Merged integration tests
3) Added factory to select right plan node
4) Corrected formatting 
5) Comments in tests
6) Passing only totalRawDocs in DictionaryBasedOperator, and not the whole metadata object

pinot-core/src/main/java/com/linkedin/pinot/core/plan/maker/PlanNodeFactory.java
```diff
@@ -0,0 +1,161 @@
...
+import com.linkedin.pinot.core.plan.AggregationPlanNode;
+import com.linkedin.pinot.core.plan.DictionaryBasedAggregationPlanNode;
+import com.linkedin.pinot.core.plan.MetadataBasedAggregationPlanNode;
+import com.linkedin.pinot.core.plan.PlanNode;
+import com.linkedin.pinot.core.plan.SelectionPlanNode;
+import com.linkedin.pinot.core.query.aggregation.function.AggregationFunctionFactory;
+import com.linkedin.pinot.core.segment.index.ColumnMetadata;
+import com.linkedin.pinot.core.segment.index.SegmentMetadataImpl;
+
+public class PlanNodeFactory {
```
kishoreg: whats the idea behind introducing a plannodeFactory.

npawar: Mayank made a suggestion to introduce a factory to return the right plan node

pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/DictionaryBasedAggregationOperator.java
```diff
@@ -0,0 +1,109 @@
...
+import com.linkedin.pinot.core.query.aggregation.AggregationResultHolder;
+import com.linkedin.pinot.core.query.aggregation.DoubleAggregationResultHolder;
+import com.linkedin.pinot.core.query.aggregation.ObjectAggregationResultHolder;
+import com.linkedin.pinot.core.query.aggregation.function.AggregationFunction;
+import com.linkedin.pinot.core.query.aggregation.function.AggregationFunctionFactory;
+import com.linkedin.pinot.core.query.aggregation.function.customobject.MinMaxRangePair;
+import com.linkedin.pinot.core.segment.index.readers.Dictionary;
+
+/**
+ * Aggregation operator that utilizes metadata for serving aggregation queries.
```
kishoreg: typo utilizes dictionary, add some more comments on how this works at a high level and the scenarios where this is applicable

pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/MetadataBasedAggregationOperator.java
```diff
@@ -67,13 +71,28 @@ protected IntermediateResultsBlock getNextBlock() {
...
-      // TODO: Add support for more aggregation functions that can be served with metadata.
+      SegmentMetadataImpl segmentMetadata = (SegmentMetadataImpl) _segmentMetadata;
       AggregationResultHolder resultHolder;
       switch (functionType) {
         case COUNT:
           resultHolder = new DoubleAggregationResultHolder(totalRawDocs);
           break;
+        case MIN:
+          String minValue = segmentMetadata.getColumnMetadataFor(column).getMinValue().toString();
+          resultHolder = new DoubleAggregationResultHolder(Double.valueOf(minValue));
```
kishoreg: what happens if its time column and the format is SDF and cannot be parsed. Let us try to use right result holder based on the column data type.  If this was not supported in the current version, please file another issue to track this. Ideally, we want to MIN/MAX etc on any data type.

pinot-core/src/main/java/com/linkedin/pinot/core/plan/DictionaryBasedAggregationPlanNode.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @Override
+  public Operator run() {
+    return new DictionaryBasedAggregationOperator(_aggregationFunctionContexts,
+        _indexSegment.getSegmentMetadata().getTotalRawDocs(), _dictionaryMap);
+  }
+
+  @Override
+  public void showTree(String prefix) {
```
kishoreg: this does not seem right. Can you look at the other plan nodes and add the appropriate code.

pinot-core/src/main/java/com/linkedin/pinot/core/plan/maker/PlanNodeFactory.java
```diff
@@ -0,0 +1,161 @@
...
+        || aggregationType.equalsIgnoreCase(AggregationFunctionFactory.AggregationFunctionType.MIN.getName())
+        || aggregationType.equalsIgnoreCase(AggregationFunctionFactory.AggregationFunctionType.MINMAXRANGE.getName()))
+        && !indexSegment.getSegmentMetadata().hasStarTree()) {
+      String column = aggregationInfo.getAggregationParams().get("column");
+      SegmentMetadataImpl segmentMetadata = (SegmentMetadataImpl) indexSegment.getSegmentMetadata();
+      if (segmentMetadata.getColumnMetadataMap() == null) {
+        return false;
+      }
+      ColumnMetadata columnMetadata = segmentMetadata.getColumnMetadataFor(column);
+      if (columnMetadata.getMaxValue() != null && columnMetadata.getMinValue() != null) {
```
kishoreg: what happens if min & max are not numeric. Will maxValue/minValue returned be null. Did we test that this works fine with old segments where we don't have min and max in the metadata.

pinot-core/src/main/java/com/linkedin/pinot/core/plan/maker/PlanNodeFactory.java
```diff
@@ -0,0 +1,161 @@
...
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private static boolean isDictionaryBasedAggregationFunction(AggregationInfo aggregationInfo,
+      IndexSegment indexSegment) {
+    String aggregationType = aggregationInfo.getAggregationType();
+    if (aggregationType.equalsIgnoreCase(AggregationFunctionFactory.AggregationFunctionType.MAX.getName())
```
kishoreg: extract the functiona name check into a util function (Its also used in metadatabased operator) AggregationFunctionType.isOfType(MIN, MAX, MINMAX)

pinot-core/src/test/java/com/linkedin/pinot/core/plan/maker/PlanNodeFactoryTest.java
```diff
@@ -0,0 +1,322 @@
...
+import com.linkedin.pinot.core.plan.DictionaryBasedAggregationPlanNode;
+import com.linkedin.pinot.core.plan.MetadataBasedAggregationPlanNode;
+import com.linkedin.pinot.core.plan.PlanNode;
+import com.linkedin.pinot.core.plan.SelectionPlanNode;
+import com.linkedin.pinot.core.segment.creator.SegmentIndexCreationDriver;
+import com.linkedin.pinot.core.segment.creator.impl.SegmentIndexCreationDriverImpl;
+import com.linkedin.pinot.pql.parsers.Pql2Compiler;
+
+
+public class PlanNodeFactoryTest {
```
kishoreg: very good test case but this is not really PlanNodeFactoryTest, this is testing specifically for the min/max/minmax case where we can use metadata/dictionary to execute the the query. We should rename the test case to reflect that.

pinot-core/src/main/java/com/linkedin/pinot/core/plan/DictionaryBasedAggregationPlanNode.java
npawar: I wrote this by referencing the MetadataBasedAggregationOperator. Something in particular that's missing?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/MetadataBasedAggregationOperator.java
npawar: min max doesn't support non numeric columns right no. Created https://github.com/linkedin/pinot/issues/2208 to track that.

pinot-core/src/main/java/com/linkedin/pinot/core/plan/maker/PlanNodeFactory.java
mayankshriv: Yes, my suggestion was to factor out the nested inlined 'ifs', and perhaps use factory for the same. This implementation does not adhere to factory pattern per say. The 'isFitFor' functions don't necessarily belong here, they could be either in the original class, or some utility class. Factory can just take the type of object to return.

pinot-core/src/main/java/com/linkedin/pinot/core/plan/DictionaryBasedAggregationPlanNode.java
mayankshriv: I think other plan nodes call the showTree of their child, but for metadata and dictionary based, they are the leaf nodes, so there's no subsequent call.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/MetadataAndDictionaryAggregationPlanClusterIntegrationTest.java
```diff
@@ -0,0 +1,536 @@
...
+import org.json.JSONObject;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+/**
+ * /**
+ * Integration test to check aggregation functions which use DictionaryBasedAggregationPlan and MetadataBasedAggregationPlan
```
mayankshriv: Curious as to why we didn't reuse existing integration test?

pinot-core/src/main/java/com/linkedin/pinot/core/plan/maker/PlanNodeFactory.java
npawar: Works fine with older segments yes. Disabled  the path to min max from metadata for now.

npawar:
Performance numbers on a table with 11G data :
Before the change, 
-----------------------------------
select max(time) from table
    "numDocsScanned": 677226153,
    "numEntriesScannedInFilter": 677703562,
    "numEntriesScannedPostFilter": 677226153,
    "totalDocs": 677703562,
    "timeUsedMs": 1042,
-----------------------------------
select min(time) from table
    "numDocsScanned": 677226153,
    "numEntriesScannedInFilter": 677703562,
    "numEntriesScannedPostFilter": 677226153,
    "totalDocs": 677703562,
    "timeUsedMs": 1099,
-----------------------------------
select minmaxrange(time) from table
    "numDocsScanned": 677226153,
    "numEntriesScannedInFilter": 677703562,
    "numEntriesScannedPostFilter": 677226153,
    "totalDocs": 677703562,
    "timeUsedMs": 908,
-------------------------------------
select min(time),max(time),minmaxrange(time) from table
   "numDocsScanned": 677226153,
    "numEntriesScannedInFilter": 677703562,
    "numEntriesScannedPostFilter": 677226153,
    "totalDocs": 677703562,
    "timeUsedMs": 1117

After the change,
-----------------------------------
select max(time) from table
-----------------------------------
select min(time) from table
-----------------------------------
select minmaxrange(time) from table
-------------------------------------
select min(time),max(time),minmaxrange(time) from table

PR #2108, opened 2017-11-14
[TE] rootcause poc - basic metric tabs with sorting

apucher:
* RCA page tabs for metrics, events, dimensions

* metric tab now displays boxes with change and is sortable

* heatmap refreshes when metric is unselected

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
@@ -1,6 +1,10 @@
 import Ember from 'ember';
 import { toBaselineUrn, hasPrefix, filterPrefix } from '../../../helpers/utils';
 
+const ROOTCAUSE_METRICS_SORT_METRIC = 'metric';
+const ROOTCAUSE_METRICS_SORT_DATASET = 'dataset';
+const ROOTCAUSE_METRICS_SORT_CHANGE = 'change';
+
 export default Ember.Component.extend({
   entities: null, // {}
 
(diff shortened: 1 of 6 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
```diff
@@ -1,6 +1,13 @@
 import Ember from 'ember';
 import { toBaselineUrn, hasPrefix, filterPrefix } from '../../../helpers/utils';
 
+const ROOTCAUSE_METRICS_SORT_PROPERTY_METRIC = 'metric';
```
justYves: I feel that having this in one const PROP will be more readable: 
```
const ROOTCAUSE_METRICS_SORT_PROPERTY = {
  DATASET= 'dataset',
  CHANGE= 'change',
  METRIC='metric'
}
```

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
```diff
@@ -1,6 +1,13 @@
...
+const ROOTCAUSE_METRICS_SORT_PROPERTY_DATASET = 'dataset';
+const ROOTCAUSE_METRICS_SORT_PROPERTY_CHANGE = 'change';
+
+const ROOTCAUSE_METRICS_SORT_MODE_ASC = 'asc';
```
justYves: same
```
const ROOTCAUSE_METRICS_SORT_MODE = {
  ASC: 'asc',
  DESC: 'desc'
}
```

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
apucher: hmm. the definition yes. the comparison will look worse "ROOTCAUSE_METRICS_SORT_PROPERTY.DATASET". isn't there a better way to do enums in JS?

justYves: not with Vanilla JS. But typescript allows for it: 
https://www.typescriptlang.org/docs/handbook/enums.html

PR #2109, opened 2017-11-14
[TE] rootcause poc - add chart subchart and zoom

apucher:


thirdeye/thirdeye-frontend/app/pods/components/timeseries-chart/component.js
```diff
@@ -146,6 +154,8 @@ export default Ember.Component.extend({
     config.regions = diffConfig.regions;
     config.tooltip = diffConfig.tooltip;
     config.legend = diffConfig.legend;
+    config.subchart = this.get('subchart');
```
justYves: nit: use `getProperties`

PR #2115, opened 2017-11-16
[TE] Append Pinot tables' custom configs to ThirdEye's dataset config

cyenjung:
Pinot stores addition configurations in table configs. Those configuration could provide the information regarding LinkedIn's Restli endpoint. In this PR, we start to import those configuration to provide more features for ThirdEye's Pinot client.

Test on local setups and new unit tests.


cyenjung:
To ensure the method is generic, we grab the entire configs map from Pinot and don't store key names. Only the code, which retrieves the information from the map, needs to know which key name it is going to use.

PR #2118, opened 2017-11-17
[TE] initial styling for the new rootcause page

justYves:
### What's new: 
- Added styling for the new rootcause poc
- changed 'paper-container' to 'card-container' for better semantics
### Testing: 
Visual sanity check locally

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
@@ -19,6 +19,8 @@
 import Ember from 'ember';
 import _ from 'lodash';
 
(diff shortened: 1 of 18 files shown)


thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -81,7 +83,10 @@ export default Ember.Component.extend({
       });
 
       // Now add new initialized props to block item
-      Object.assign(block, { filtersArray, isHidden });
```
justYves: @ttbach this was generating an error, you'll need to use Ember.set instead of Object assignment

PR #2121, opened 2017-11-17
Ttbach/search filter component

ttbach:
- Created search filter bar component to filter results in the events table
- Moved caching to rootcause controller (filters cache and urns cache)
- Moved filtering logic from child component to rootcause controller. TLDR; This is because in order for search filter bar to work, it needed a reference to all the filters that were selected from both the `filter-bar` and the `filter-bar-input`. For more explanation, see below:
- Before, I had the logic for filtering entities by event type in `filter-bar` and filtering by subfilters (i.e. country) in the `filter-bar-input`, both of which compute the urns and send an action with urns up to the controller that resets the `filteredUrns` and changes how entities are displayed. However, since the `filter-search-bar` needs to know what filters were previously applied (for event type and subfilters), I thought a better way to do this is to have the controller keep track of all filters and process the filtering logic in the controller, so the sub-filtering-components (filter-bar, filter-bar-input, and search-filter-bar) just need to worry about calling an action that bubbles up to update the filters.
- Minor bug: switching between tabs don't clear the input text in the `search-filter-bar`. For example, if i type "abc" into the search bar when I'm on the `Holiday` tab, it filters the `Holiday` events' labels by "abc." However, when I switch to the `Anomalies` tab, the search bar still has "abc", but does not filter the results based on "abc". Otherwise, it functions.

screen shot 2017-11-16 at 11 05 05 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
@@ -165,6 +165,7 @@ export default Ember.Component.extend({
   actions: {
 
     /**
+     * Filters results in the events table depending on a selected event type in the filter block
      * Expands/collapses a filter block
      * @method filterByEvent
      * @param {Object} clickedBlock - selected filter block object
@@ -185,7 +186,7 @@ export default Ember.Component.extend({
 
       /*
       * If results were already previously filtered for this filter block (i.e. "Holiday", "Deployment"),
-      * call onSelect on the cached urns
+      * call onSelect on the cached urns.
       */
       if (cachedHeader) {
         onSelect(cachedHeader);
@@ -194,7 +195,9 @@ export default Ember.Component.extend({
       else {
         const urns = Object.keys(entities).filter(urn => entities[urn].type == 'event'
                                                   && entities[urn].eventType == clickedBlock.eventType);
-        this.urnsCache[clickedBlock.header] = urns;
+        if (urns.length) {
+          this.urnsCache[clickedBlock.header] = urns;
+        }
         onSelect(urns);
       }
     },
--- a/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
@@ -28,17 +28,6 @@ export default Ember.Component.extend({
   options: ['All', 'None'],
 
   /**
-   * Cache for urns, filtered by search criteria
-   * @type {Object}
-   * @example
-   * {
-   *  Holiday: {urns1, urns2},
-   *  Region: {urns3, urns4}
-   * }
-   */
-  urnsCache: {},
-
-  /**
    * Overwrite the init function
    * Initializes values of the filter blocks
    * Example of a filter block:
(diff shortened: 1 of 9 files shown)


thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -81,7 +77,8 @@ export default Ember.Component.extend({
       });
 
       // Now add new initialized props to block item
-      Object.assign(block, { filtersArray, isHidden });
+      Ember.set(block, 'filtersArray', filtersArray);
+      Ember.set(block, 'isHidden', isHidden);
```
justYves: I believe the following should work here
```
Ember.setProperties(block, { 
  filtersArray,
  isHidden
})
```

justYves: when we use `this.set` and `this.setProperties` the context is obvious (this), but when using Ember.set, Ember.setProperties, the first arguments becomes the context.

thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -165,13 +162,13 @@ export default Ember.Component.extend({
...
+     * Filters results in the events table depending on a selected event type in the filter block
      * Expands/collapses a filter block
      * @method filterByEvent
      * @param {Object} clickedBlock - selected filter block object
      */
     selectEventType(clickedBlock) {
-      const { entities, onSelect } = this.getProperties('entities', 'onSelect');
-      const cachedHeader = this.urnsCache[clickedBlock.header];
+      const filterUrns = this.get('filterUrns');
       let filterBlocks = this.get('config');
```
justYves: Any reason we're using `let`here? I don't see the variable being modified latter on.
Also this can be reduced to a one line: 
```
const {
filterUrns,
config: filterBlocks
} = this.getProperties('filterUrns', 'config')
```

justYves: see `Assigning to new variable names` in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

thirdeye/thirdeye-frontend/app/pods/components/rca-filter-search-bar/component.js
```diff
@@ -0,0 +1,23 @@
...
+
+  actions: {
+
+    /**
+     * Filters the results in the events table based on user input in the search bar
+     * @method filterResults
+     */
+    filterResults() {
+      const { activeEventType, filterUrns, value } = this.getProperties('activeEventType', 'filterUrns', 'value');
+      this.set('value', this.get('value'));
```
justYves: you can just use `value` instead of this.get()

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -283,6 +312,92 @@ export default Ember.Controller.extend({
...
+      this.urnsCache[eventType] = urns;
+    },
+
+    /**
+     * Filters urns based on event type, subfilters selected, and user text from filter search bar
+     * @param {String} eventType - type of event
+     * @param {Object} subFilter - filter within event (i.e. country, region for Holiday)
+     * @param {String} text - user's input from filter search bar
+     */
+    filterUrns(eventType, subFilter, text) {
```
justYves: This is ok for poc, but let's refactor it together once you're back.

justYves: It's not very clear what the whole function does. We can improve readability by breaking it down into smaller utils/helper functions

PR #2123, opened 2017-11-17
[TE] Add try-and-catch block to protect the initialization of cache loader.

cyenjung:
Add protections when initializing Pinot cache loader. It falls back to the default class: PinotControllerResponseCacheLoader, if the external cache loader is unable to be initialized.


cyenjung:
It should be fine because the default cache loader is not going to work if user didn't set it up. In that case, user will notice that the actual one is not started up properly. The default one is just an extra try.

PR #2125, opened 2017-11-18
[TE] Rootcause frontend heatmap scrolling

apucher:
* moving sizing logic to rootcause-heatmap

* support two-sided rollup

* left-right scrolling per row


apucher:
still some graphical glitches, but good enough for a prototype

PR #2126, opened 2017-11-20
[TE] add endpoint to change alert group config, active/deactivate holiday mode by add/rm holiday cron

bl44:
[TE] add endpoint to change alert group config, active/deactivate holiday mode by add/rm holiday cron

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EmailResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EmailResource.java
@@ -6,38 +6,26 @@ import com.linkedin.thirdeye.anomaly.ThirdEyeAnomalyConfiguration;
 import com.linkedin.thirdeye.anomaly.alert.util.AlertFilterHelper;
 import com.linkedin.thirdeye.anomaly.alert.util.AnomalyReportGenerator;
 import com.linkedin.thirdeye.anomaly.alert.util.EmailHelper;
+import com.linkedin.thirdeye.common.ThirdEyeConfiguration;
 import com.linkedin.thirdeye.datalayer.bao.AlertConfigManager;
 import com.linkedin.thirdeye.datalayer.bao.ApplicationManager;
 import com.linkedin.thirdeye.datalayer.dto.AlertConfigDTO;
 import com.linkedin.thirdeye.datalayer.dto.ApplicationDTO;
 import com.linkedin.thirdeye.datalayer.dto.MergedAnomalyResultDTO;
-import com.linkedin.thirdeye.detector.email.filter.AlertFilterFactory;
-
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-
-import java.util.Map;
-import java.util.Set;
-import javax.ws.rs.DELETE;
-import javax.ws.rs.GET;
-import javax.ws.rs.POST;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
-import javax.ws.rs.QueryParam;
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EmailResource.java
```diff
@@ -81,6 +69,74 @@ public Response deleteByAlertId(@PathParam("alertId") Long alertId) {
...
+   * @param id alert config id
+   * @param cron cron expression for alert
+   * @param isActive activate or not
+   * @return Response
+   * @throws Exception
+   */
+  @PUT
+  @Path("/alert/{id}")
+  public Response updateAlertConfig(@NotNull @PathParam("id") Long id,
+      @QueryParam("cron") String cron, @QueryParam("isActive") Boolean isActive) throws Exception {
```
apucher: there already is an enable/disable endpoint in AlertJobResource.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EmailResource.java
```diff
@@ -81,6 +69,74 @@ public Response deleteByAlertId(@PathParam("alertId") Long alertId) {
...
+   * if this cron is not null then holiday mode is activate
+   * @param id id of the config
+   * @param cron holiday cron expression
+   * @return Response
+   * @throws Exception
+   */
+  @PUT
+  @Path("/alert/{id}/holiday-mode")
+  public Response updateAlertConfigHolidayCron(@NotNull @PathParam("id") Long id,
+      @QueryParam("cron") String cron) throws Exception {
```
apucher: thanks for implementing. please move this to AlertJobResource.

bl44: AlertJobResource is being deprecated. and We should not add more functions into it.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EmailResource.java
bl44: AlertJobResource. Add this new class to EmailResources.  We can consider rename the file later. 
For now just keep it here in the repo.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/EmailResource.java
```diff
@@ -81,6 +69,74 @@ public Response deleteByAlertId(@PathParam("alertId") Long alertId) {
...
+    }
+
+    if (StringUtils.isNotEmpty(cron)) {
+      // validate cron
+      if (!CronExpression.isValidExpression(cron)) {
+        throw new IllegalArgumentException("Invalid cron expression for cron : " + cron);
+      }
+      // as long as there is an valid holiday cron expression within the class
+      // the holiday model is activate
+      alert.setHolidayCronExpression(cron);
```
cecilynie: Since this endpoint is already defining "holiday mode". Are we expecting user of this endpoint to input a specific "cron" to define the holiday mode, or within this endpoint it can figure our the default cron by the granularity of the metric already? Then what user's input on cron can be a customization on the alert specifically.

bl44: the holidayCronExpression is used in two ways:
1) if this is not null, then its holiday mode
2) For holiday mode, use the cron for alert distribution.

One need to setup customized cron expression for each 'alert config'. This will be done through this endpoint, and of cause we can use a generic format for all alert config: '0 0 0 8 * *' -> 8 am in the morning.

PR #2133, opened 2017-11-21
Adding json string serialization/deserialization for startree

jenniferdai:


pinot-common/src/main/java/com/linkedin/pinot/common/data/StarTreeIndexSpec.java
```diff
@@ -40,9 +45,25 @@
...
+  private static final String DIMENSIONS_SPLIT_ORDER_PARAM = "dimensionsSplitOrder";
+  private static final String MAX_LEAF_RECORDS_PARAM = "maxLeafRecords";
+  private static final String SKIP_MATERIALIZATION_CARDINALITY_THRESHOLD_PARAM = "skipMaterializationCardinalityThreshold";
+  private static final String SKIP_MATERIALIZATION_FOR_DIMENSIONS_PARAM = "skipMaterializationForDimensions";
+  private static final String SKIP_STAR_NODE_CREATION_FOR_DIMENSIONS_PARAM = "skipStarNodeCreationForDimensions";
+
   public StarTreeIndexSpec() {
   }
 
+  public StarTreeIndexSpec(String jsonString) {
```
kishoreg: why are we not using jackson library

pinot-common/src/main/java/com/linkedin/pinot/common/data/StarTreeIndexSpec.java
```diff
@@ -15,7 +15,12 @@
  */
 package com.linkedin.pinot.common.data;
 
+import com.alibaba.fastjson.JSON;
```
mayankshriv: We use jackson in other places, lets use that.

pinot-common/src/main/java/com/linkedin/pinot/common/data/StarTreeIndexSpec.java
mayankshriv: Could we use a static method fromJson() for this, instead of a separate constructor?

pinot-common/src/main/java/com/linkedin/pinot/common/data/StarTreeIndexSpec.java
```diff
@@ -96,4 +117,15 @@ public void setExcludeSkipMaterializationDimensionsForStarTreeIndex(
   public String toString() {
     return ToStringBuilder.reflectionToString(this, ToStringStyle.SHORT_PREFIX_STYLE);
   }
+
+  public String toJsonString() {
+    Map<String, Object> map = new HashMap<>();
```
mayankshriv: Use ObjectMapper to serialize (and de-serialize) to (from) JSON.

pinot-common/src/main/java/com/linkedin/pinot/common/data/StarTreeIndexSpec.java
jenniferdai: done

pinot-common/src/main/java/com/linkedin/pinot/common/data/StarTreeIndexSpec.java
jenniferdai: done

pinot-common/src/main/java/com/linkedin/pinot/common/data/StarTreeIndexSpec.java
```diff
@@ -40,9 +41,20 @@
 
   private boolean _excludeSkipMaterializationDimensionsForStarTreeIndex;
 
+  private static final String SEPARATOR = ",";
+  private static final String DIMENSIONS_SPLIT_ORDER_PARAM = "dimensionsSplitOrder";
```
kishoreg: do we need all these constants if we are using jackson?

jenniferdai: thanks, done

pinot-common/src/main/java/com/linkedin/pinot/common/data/StarTreeIndexSpec.java
```diff
@@ -96,4 +101,8 @@ public void setExcludeSkipMaterializationDimensionsForStarTreeIndex(
   public String toString() {
     return ToStringBuilder.reflectionToString(this, ToStringStyle.SHORT_PREFIX_STYLE);
   }
+
+  public String toJsonString() throws Exception {
```
mayankshriv: I know this is already committed, but a couple of suggestions:
1. Use static final ObjectMapper so no need to new each time.
2. I use objectMapper().writerWithDefaultPrettyPrinter() to make the json output pretty printed.

jenniferdai: thanks, https://github.com/linkedin/pinot/pull/2134

PR #2144, opened 2017-11-23
[TE] rootcause poc - dynamic primary metric

apucher:
This PR adds support for dynamic change of the primary metric. Also adds basic support for dynamic loading of primary metric filter options.


thirdeye/thirdeye-frontend/app/pods/components/rootcause-metric-select/component.js
```diff
@@ -20,6 +28,26 @@ export default Ember.Component.extend({
...
+      this.set('selectedUrnCache', selectedUrn);
+
+      if (selectedUrn) {
+        const url = `/data/metric/${selectedUrn.split(':')[2]}`;
+        fetch(url)
+          .then(res => res.json())
+          .then(res => this.set('selectedMetric', res));
+
+      } else {
+        this.set('selectedMetric', undefined);
```
justYves: Is there any way we could refactor this piece of code so that there's only one this.setProperties at the bottom. 
Having multiple consecutive `set` usually results in a slight decrease in performance (it's minor, but it adds up)

apucher: the else clause here is defensive programming, i.e. it should never actually be called during execution.

PR #2146, opened 2017-11-27
[TE] Add Serializer/Deserializer for TE ResultSet and Create Pinot query proxy 

cyenjung:
1. Add serializer and deserializer for TE ResultSet.
2. Add Pinot query proxy to ThirdEyeAnomalyApplication. Now a worker could be used as a central portal to Pinot. Thus, other workers or dashboard could leverage the cache on the proxy worker.

Tested with new Unit test and local setup, in which a proxy worker and a dashboard server is started up. The dashboard server could get the cached data from the proxy worker successfully.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyApplication.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyApplication.java
@@ -8,17 +8,15 @@ import com.linkedin.thirdeye.anomalydetection.alertFilterAutotune.AlertFilterAut
 import com.linkedin.thirdeye.dashboard.resources.AnomalyFunctionResource;
 import com.linkedin.thirdeye.dashboard.resources.EmailResource;
 import com.linkedin.thirdeye.datasource.ThirdEyeCacheRegistry;
+import com.linkedin.thirdeye.datasource.pinot.resources.PinotDataSourceResource;
 import com.linkedin.thirdeye.detector.email.filter.AlertFilterFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
 
 import com.linkedin.thirdeye.dashboard.resources.DetectionJobResource;
 import com.linkedin.thirdeye.anomaly.detection.DetectionJobScheduler;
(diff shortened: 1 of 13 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyApplication.java
```diff
@@ -169,7 +158,34 @@ public void stop() throws Exception {
...
         }
+        if (config.isPinotProxy()) {
+          // Do nothing
+        }
       }
     });
   }
+
+  private void initAnomalyFunctionFactory(String functoinConfigPath) {
+    if (anomalyFunctionFactory == null) {
```
apucher: why is this null-check needed? this code path should execute exactly once

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datasource/pinot/resources/PinotDataSourceResource.java
```diff
@@ -0,0 +1,84 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Path("/pinot-data-source")
+@Produces(MediaType.APPLICATION_JSON)
+public class PinotDataSourceResource {
```
apucher: this is very dangerous. please at least add documentation that this is for testing purposes only. It allows full access to pinot

cyenjung: No, it won't. The accessibility of tables is set on Pinot. If a Pinot table is not accessable to TE, this endpoint won't make TE have superpower and access the table.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyApplication.java
cyenjung: It could be invoked multiple times when a controller has multiple rules, e.g., worker and alerter.

We didn't have this check and the second initialization overwrites the previous value. This helper function ensures that a factory instance is reused.

PR #2152, opened 2017-11-28
Fixed filter-bar bug with reloading new entities

ttbach:
- Bug: when entities are changed by user changing fields like "Comparing By" or "Display Window", the table results don't reload based on these changes.
- Modified urnsCache to have the key be `eventType` instead of `header` (reference in `filterBarConfig` file)

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
@@ -34,8 +34,8 @@ export default Ember.Component.extend({
    * @type {Object}
    * @example
    * {
-   *  Holiday: {urns1, urns2},
-   *  Region: {urns3, urns4}
+   *  anomaly: {urns1, urns2},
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -97,7 +97,16 @@ export default Ember.Component.extend({
   entitiesObserver: Ember.observer(
     'entities',
     function() {
-      const filterBlocks = this.get('config');
+      const { config: filterBlocks, entities } = this.getProperties('config', 'entities');
+
+      // When entities are changed (i.e. from user interacting with the selection form), recompute urns to cache
+      filterBlocks.forEach(block => {
```
justYves: can `filterBlocks` be null or undefined here?

ttbach: No, since it's a guarantee that the `config` file that gets passed is not null.

PR #2153, opened 2017-11-28
Clean up Star Tree specific configs in SegmentGenerationConfig.

mayankshriv:
Consolidated the following three api's into one, as there were multiple
occassions where callers would need to call more than one of these api's
to get the same effect, which also caused bugs on a few occassions when
callers missed to call one of the api's. This change helps avoid these
issues.

- setEnableStarTree
- setStarTreeIndexSpec
- setStarTreeIndexSpecFile

The new api setEnableStarTree(StarTreeIndexSpec) takes StarTreeIndexSpec
as input, which can be null as well.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
@@ -54,6 +54,7 @@ import org.slf4j.LoggerFactory;
 /**
  * Configuration properties used in the creation of index segments.
  */
+@SuppressWarnings("unused")
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class SegmentGeneratorConfig {
   public enum TimeColumnType {
@@ -84,7 +85,6 @@ public class SegmentGeneratorConfig {
   private String _readerConfigFile = null;
   private RecordReaderConfig _readerConfig = null;
   private boolean _enableStarTreeIndex = false;
-  private String _starTreeIndexSpecFile = null;
   private StarTreeIndexSpec _starTreeIndexSpec = null;
   private String _creatorVersion = null;
   private char _paddingCharacter = V1Constants.Str.DEFAULT_STRING_PAD_CHAR;
@@ -130,7 +130,6 @@ public class SegmentGeneratorConfig {
     _readerConfigFile = config._readerConfigFile;
     _readerConfig = config._readerConfig;
     _enableStarTreeIndex = config._enableStarTreeIndex;
-    _starTreeIndexSpecFile = config._starTreeIndexSpecFile;
     _starTreeIndexSpec = config._starTreeIndexSpec;
     _creatorVersion = config._creatorVersion;
     _paddingCharacter = config._paddingCharacter;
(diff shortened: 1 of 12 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
```diff
@@ -433,30 +429,19 @@ public boolean isEnableStarTreeIndex() {
...
-  public void setStarTreeIndexSpecFile(String starTreeIndexSpecFile) {
-    _starTreeIndexSpecFile = starTreeIndexSpecFile;
-
-    // Setting the star tree index spec should automatically enable star tree generation, so that clients
-    // don't have to explicitly set both.
+  /**
+   * Enable star tree generation.
+   * @param starTreeIndexSpec Indexing spec for star tree. If null, then default values are used.
+   */
+  public void setEnableStarTreeIndex(StarTreeIndexSpec starTreeIndexSpec) {
```
jenniferdai: maybe change the name so that it doesn't sound like it should be a boolean field?

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -323,18 +325,16 @@ public boolean accept(File dir, String name) {
...
-            "Find starTreeIndexSpecFile conflict in command line and config file, use config in command line: {}",
-            _starTreeIndexSpecFile);
-      }
-      segmentGeneratorConfig.setStarTreeIndexSpecFile(_starTreeIndexSpecFile);
+      StarTreeIndexSpec starTreeIndexSpec = StarTreeIndexSpec.fromFile(new File(_starTreeIndexSpecFile));
+
+      // Specifying star-tree index file enables star tree generation, even if _enableStarTreeIndex is not specified.
+      segmentGeneratorConfig.setEnableStarTreeIndex(starTreeIndexSpec);
+    } else if (_enableStarTreeIndex) {
+      segmentGeneratorConfig.setEnableStarTreeIndex(null);
```
jenniferdai: what is this for?

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
mayankshriv: How about enableStarTreeIndex()?

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/CreateSegmentCommand.java
mayankshriv: So one can specify -enableStarTreeIndex without the config file.

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
jenniferdai: i like setStarTreeIndexSpec or something without enable more, but as long as you feel that it isn't a boolean, fine with me

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/CreateSegmentCommand.java
jenniferdai: sorry if i'm misunderstanding, but if you're enabling startree, i assume you need some sort of config?

pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
mayankshriv: I prefer to have 'enable' in the name because that is the primary function of this api. StarTreeIndexSpec can be null.

PR #2163, opened 2017-11-30
[TE] rootcause poc - unified coloring for metrics, events

apucher:


thirdeye/thirdeye-frontend/app/styles/pods/rca.scss
```diff
@@ -137,13 +137,31 @@
...
+  "orange" $te-orange,
+  "teal" $te-teal,
+  "purple" $te-purple,
+  "red" $te-red,
+  "green" $te-green,
+  "pink" $te-pink;
+
+  // creates box--"color" classes
+@each $i in $colors {
+  .rootcause-legend__box--#{nth($i, 1)} {
```
justYves: nice try at styling! 
fyi we're using the BEM naming convention 
### relevant links: 
- http://getbem.com/introduction/
- https://css-tricks.com/bem-101/

Best practice here is to have a base styling `.rootcause-legend__box`which would have 
```
+    display: inline-block;
+    height: 0.8em;
+    width: 0.8em;
```
Then the modifier `.rootcause-legend__box--colorName` would only need to apply a background. 
Yes, this means a more verbose template (`class="rootcause-legend__box rootcause-legend__box--colorName"`), but we do gain a lot in readability and maintainability imo.

PR #2167, opened 2017-12-01
[TE] frontend - Fixed bug with filter bar on init

ttbach:
- Moved creation of `filterBlocks` to rootcause's controller to pass to the `filter-bar` template, so the default block will be selected when different tabs are selected that doesn't affect changes to `entities` (`Events Correlation`, `Dimensions Analysis`).

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
@@ -4,13 +4,15 @@
  * @module components/filter-bar
  * @property {object[]} config            - [required] array of objects (config file) passed in from the route that sets
  *                                          up the filter bar sub-filters
- * @property {number} maxStrLen           - number of characters for filter name truncation
+ * @property {object[]} filterBlock       - [required] array of objects constructed from config to render filter bar
+ * @property {number} maxStrLen           - [optional] number of characters for filter name truncation
  * @property {array}  onFilterSelection   - [required] closure action to bubble to controller on filter selection change
  *
  * @example
  * {{filter-bar
  *   config=filterBarConfig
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -4,13 +4,15 @@
...
- * @property {number} maxStrLen           - number of characters for filter name truncation
+ * @property {object[]} filterBlock       - [required] array of objects constructed from config to render filter bar
+ * @property {number} maxStrLen           - [optional] number of characters for filter name truncation
  * @property {array}  onFilterSelection   - [required] closure action to bubble to controller on filter selection change
  *
  * @example
  * {{filter-bar
  *   config=filterBarConfig
  *   filterBlocks=filterBlocks
+ *   entities=entities
```
justYves: let's add this to the jsdoc above

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -3,6 +3,7 @@ import { filterObject, filterPrefix, toBaselineUrn, toCurrentUrn } from 'thirdey
 import EVENT_TABLE_COLUMNS from 'thirdeye-frontend/mocks/eventTableColumns';
 import config from 'thirdeye-frontend/mocks/filterBarConfig';
 import CryptoJS from 'cryptojs';
+const { setProperties } = Ember;
```
justYves: since Ember 2.16, the best practice is to import modules instead of the whole ember library.
```
  import { computed, get, set, setProperties } from '@ember/object'
```
#### Official documentation: 
https://www.emberjs.com/blog/2017/10/11/ember-2-16-released.html

PR #2170, opened 2017-12-03
[PINOT-5611] Remove FileUploadUtils dependency on hadoop

mcvsubbu:


mcvsubbu:
@mayankshriv no pom changes that added hadoop dependency

PR #2171, opened 2017-12-03
[PINOT-5458] Fix SegmentFetcherFactory to allow overriding fetcher classes via configuration

mcvsubbu:
An earlier commit disabled the feature to override pre-loaded classes with new ones for a protocol.
Added back that feature, and added a test for it.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
@@ -78,27 +78,33 @@ public class SegmentFetcherFactory {
       Configuration protocolConfig = segmentFetcherFactoryConfig.subset(protocol);
       String configuredClassName = protocolConfig.getString(SEGMENT_FETCHER_CLASS_KEY);
       if (configuredClassName == null || configuredClassName.isEmpty()) {
-        // If the class is not in the map, then it should already be instantiated, we can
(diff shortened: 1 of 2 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -61,33 +60,61 @@ private static void instantiateSegmentFetcher(String protocol, String className)
...
+      Configuration protocolConfig = segmentFetcherFactoryConfig.subset(protocol);
+      String configuredClassName = protocolConfig.getString(SEGMENT_FETCHER_CLASS_KEY);
+      if (configuredClassName == null || configuredClassName.isEmpty()) {
+        // If the class is not in the map, then it should already be instantiated, we can
+        // still init it with the config provided. Otherwise, the config is ignored, so
+        // flag a warning.
+        if (!SEGMENT_FETCHER_MAP.containsKey(protocol)) {
+          LOGGER.warn("No class name given for protocol {}. Configuration ignored", protocol);
+        }
+        continue;
```
Jackie-Jiang: Logic is a little bit unclear by first glance. Recommend to change it to:
```
if (configuredClassName == null || configuredClassName.isEmpty()) {
  if (contains) {
    log info (use builtinClassName)
  } else {
    log warning
  }
} else {
  if (contains) {
    if (builtinClassName.equals(configuredClassName)) {
      log info (config same as builtin)
    } else {
      init
    }
  } else {
    init
  }
}
```

PR #2175, opened 2017-12-04
[TE] frontend - ttbach/bug fix filter bar

ttbach:
- Updated filter bar config
- Added a cache for selected filters in the `filter-bar` component to fix bugs
- Ignore https://github.com/ttbach/pinot/tree/ttbach/filter-bar-bug, which is a less complete PR for bug fixes
- Updates to module import to comply by new Ember standards will be submitted in later PRs

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
@@ -4,14 +4,13 @@
  * @module components/filter-bar
  * @property {object[]} config            - [required] array of objects (config file) passed in from the route that sets
  *                                          up the filter bar sub-filters
- * @property {number} maxStrLen           - number of characters for filter name truncation
- * @property {array}  onFilterSelection   - [required] closure action to bubble to controller on filter selection change
+ * @property {object} entities            - [required] data used to filter
+ * @property {array}  onFilterSelection   - [required] closure action to bubble to controller on filter selection
  *
  * @example
  * {{filter-bar
  *   config=filterBarConfig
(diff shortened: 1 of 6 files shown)


thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -88,26 +99,37 @@ export default Ember.Component.extend({
         isHidden
       });
     });
+
+    const eventType = filterBlocks[0].eventType;
```
justYves: nit: can filterBlocks be `null` or `undefined` here?

thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -172,54 +194,84 @@ export default Ember.Component.extend({
...
+   * Determines whether to apply filters
+   * @param {Object} entity
+   * @param {Object} filters - subfilters selected (i.e. {countryCode: 'DE'})
+   */
+  applyFilters(e, filters) {
+    if (_.isEmpty(filters)) {
+      return true;
+    }
+
+    return Object.keys(filters).every(dimName => {
```
justYves: nit: let's use parenthesis around the argument names when the array function spans over multiple line.

thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
ttbach: Assuming the config file is passed and not empty, filterBlocks will never be null or undefined.

PR #2177, opened 2017-12-05
Change AccessControl API and only call it when necessary

Jackie-Jiang:
Add API hasDataAccess(httpHeaders, tableName) to AccessControl
Only call this API when trying to download a segment (access data)
This way the access control will not affect other controller APIs


Jackie-Jiang:
@mcvsubbu Removed the old filter class, and added the hasAccess to sql query server.

PR #2180, opened 2017-12-05
[TE] frontend - filter bar styling

ttbach:
screen shot 2017-12-05 at 11 26 46 am


thirdeye/thirdeye-frontend/app/mocks/filterBarConfig.js
```diff
@@ -12,6 +13,7 @@ export default [
   {
     header: "Anomalies",
     eventType: "anomaly",
+    color: "teal",
```
justYves: hard coded for now?

thirdeye/thirdeye-frontend/app/styles/components/button.scss
```diff
@@ -61,3 +61,10 @@
     }
   }
 }
+
+.btn span.glyphicon {
```
justYves: ### Blocking Comment: 
This selector is too general (it also affects the icons in the rca table). 
Please wrap this with a custom css class. Since we're overriding bootstrap, I'm okay with using !important (bs uses it everywhere anyway)

thirdeye/thirdeye-frontend/app/styles/components/filter-bar.scss
```diff
@@ -1,3 +1,62 @@
...
+    border-top: 1px solid $te-grey--border;
+    padding: 25px $te-default-element-spacing;
+    margin-bottom: 0px;
+    background-color: #F3F6F8;
+
+    &--hidden {
+      display: none;
+    }
+  }
+}
```
justYves: nit: new line at EOF

thirdeye/thirdeye-frontend/app/mocks/filterBarConfig.js
ttbach: Yes. In the future, when the config is more finalized, this is assumed to be in the config.

PR #2185, opened 2017-12-07
[TE] Revert "Use dataframe in thirdeye response"

cyenjung:
Reverts linkedin/pinot#2176

This migration doesn't take care rolling up data. For instance, roll up data point from 5-MINUTES to 1-DAYS granularity. Therefore, we revert the change until the roll up functionality is implemented using data frame.

After the migration to data frame, the timestamp shifting issue should be resolved.


cyenjung:
Yup. Will come back for it in another day.

PR #2186, opened 2017-12-07
[TE] tooltip template helper

justYves:
### What's new: 
- Created a build-tooltip helper so that we can inject dynamic html inside of ember-c3. 
- abstracted the 

### Todo (improvements): 
- When the tooltip with a large number of selected urns, the tooltip is positioned in front of the data. 
- Shows the selected dot's information on top rather than inline.
- % data is not uniform (variable decimals)
- missing data is sometimes NaN or 0.

### Tests:
Visual sanity check locally

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import d3 from 'd3';
 import moment from 'moment';
+import buildTooltip from 'thirdeye-frontend/helpers/build-tooltip';
 import { toBaselineUrn, toMetricUrn, filterPrefix, hasPrefix, stripTail, toBaselineRange, toFilters, toMetricLabel } from 'thirdeye-frontend/helpers/utils';
 
 const TIMESERIES_MODE_ABSOLUTE = 'absolute';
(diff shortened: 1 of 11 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
```diff
@@ -30,19 +31,30 @@ export default Ember.Component.extend({
     show: false
   },
 
+  buildTooltip: buildTooltip,
```
ttbach: JSDoc? It'd be great to add as much as we can.

thirdeye/thirdeye-frontend/app/styles/components/filter-bar.scss
```diff
@@ -60,3 +37,33 @@
...
+  display: inline-block;
+  margin-right: 10px;
+  width: 10px;
+  height: 10px;
+  border-radius: 50px;
+  @each $i in $colors {
+    &--#{nth($i, 1)} {
+      background: nth($i, 2);
+    }
+  };
```
ttbach: 👍 👍 👍

thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
justYves: :+1: This is basically importing the template helper defined earlier. Will add a note with my following pr.

PR #2189, opened 2017-12-07
Ttbach/rca header styling

ttbach:
screen shot 2017-12-07 at 12 08 05 pm


thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -69,7 +79,11 @@ export default Ember.Component.extend({
 
   comment: Ember.computed('anomaly', function () {
     const attr = this.get('anomaly').attributes;
-    return attr.comment[0];
```
justYves: wondering if doing `const attr = this.get('anomaly.attributes.comment.firstObject')` would work here, 
so we can just `return attr || [];`

ttbach:
@apucher I was thinking about that. I'll make the consolidation.

PR #2191, opened 2017-12-07
[TE] rootcause poc - get-safe helper to gracefully handle urns with . (dot)

apucher:


thirdeye/thirdeye-frontend/app/helpers/get-safe.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
+
+export function getSafe([dict, key]) {
```
justYves: Would love to see some sort of documentation here i.e. why this custom helper is needed

apucher: fixed

PR #2192, opened 2017-12-07
[TE] rootcause poc - remove rootcause-filter dummy component

apucher:


apucher:
Thank you, thank you.

PR #2200, opened 2017-12-08
Fix SegmentFetcherFactory configuration to be different for controller and server

mcvsubbu:


pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
```diff
@@ -252,6 +252,10 @@ public ServerType getServerType() {
     public static final String DEFAULT_COLUMN_MIN_MAX_VALUE_GENERATOR_MODE = "TIME";
   }
 
+  public static class Controller {
+    public static final String PREFIX_OF_CONFIG_OF_SEGMENT_FETCHER_FACTORY = "pinot.controller.segment.fetcher";
```
jamesyfshao: the segment fetcher code is actually used by both controller and server (usage in server code in [here](https://github.com/linkedin/pinot/blob/master/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java#L78)) I am not sure prefixing it with either controller/server is a good way to do. Maybe we should consider the naming something like pinot.common.segment.fetcher?

pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactoryTest.java
```diff
@@ -47,7 +48,7 @@ public void testInitSegmentFetcherFactory() throws Exception {
     SegmentFetcherFactory.getPreloadSegmentFetchers().put("https", mockHttpsFetcher);
     SegmentFetcherFactory.getPreloadSegmentFetchers().put(replacableProto, replacableFetcher);
 
-    SegmentFetcherFactory.initSegmentFetcherFactory(conf);
+    SegmentFetcherFactory.initSegmentFetcherFactory(conf.subset(CommonConstants.Server.PREFIX_OF_CONFIG_OF_SEGMENT_FETCHER_FACTORY));
```
jamesyfshao: We would be fine to change the config to the latest key. However, I am not sure if any of other users might be relying on this configuration key. If you or other contributors are aware of other users using this configuration we should think about supporting backward compatibility.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
mcvsubbu: The  config for segment fetcher in controller   should be prefixed by pinot.controller.

the config for segment fetcher in server should be prefixed by pinot.server.

All configs inside have to be duplicated for controller and server as needed depending on the fetchers that they are using. 

I don't see a problem with this. Maybe I don't understand your question. 

Changing things to pinot.common will certainly break anyone who has been using pinot.server. for a long time (and it has been there for years now).

pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactoryTest.java
mcvsubbu: Yes, there is a chance that other users may be using pinot.server config in the controller between Sep and now. We will take that chance since this is a relatively new feature and small time window. The worst that will happen is that the hdfs fetcher (or any other custom fetchers that they may have added) will fail to initialize, and they will need to re-configure. Keeping pinot.server config inside controller is bad for the long run. Unfortunate that it slipped through the code.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
jamesyfshao: got it, I misunderstood the changes you made in server starter section earlier but it makes sense now. Thanks!

mcvsubbu:
This commit changes the way controller is configured with fetchers. Earlier configurations looked like: "pinot.server.fetcher.foo.bar". These will no longer be honored. Instead, the configuration should be changed to "pinot.controller.fetcher.foo.bar".

Server's configurations can remain unchanged.

PR #2201, opened 2017-12-08
Adding unified download method for minion

jenniferdai:


pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadUtils.java
```diff
@@ -94,11 +95,12 @@ public EntityEnclosingMethod forUri(String uri) {
...
   public static int sendFile(final String host, final String port, final String path, final String fileName,
       final InputStream inputStream, final long lengthInBytes, SendFileMethod httpMethod) {
-    return sendFile("http://" + host + ":" + port + "/" + path, fileName, inputStream, lengthInBytes, httpMethod);
+    return sendFile("http://" + host + ":" + port + "/" + path, fileName, inputStream, lengthInBytes,
+        httpMethod, null);
   }
 
   public static int sendFile(String uri, final String fileName, final InputStream inputStream, final long lengthInBytes,
-      SendFileMethod httpMethod) {
+      SendFileMethod httpMethod, String crc) {
```
mcvsubbu: suggest rename 'crc' to ifMatchSegmentCrc to clarify what it signifies.

Also, please add some comments that this is useful only for segment push to controller. This class seems to be a generic FileUpload from anywhere to anywhere...

Lastly, the type of ifMatchSegmentCrc should be long, not string

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadUtils.java
```diff
@@ -119,6 +121,9 @@ public InputStream createInputStream() throws IOException {
         }
       })};
       method.setRequestEntity(new MultipartRequestEntity(parts, new HttpMethodParams()));
+      Header header = new Header("If-Match", crc);
```
mcvsubbu: Can u get the string "If-Match" from some apache class?

pinot-core/src/main/java/com/linkedin/pinot/core/common/MinionConstants.java
```diff
@@ -23,6 +23,7 @@ private MinionConstants() {
   public static final String SEGMENT_NAME_KEY = "segmentName";
   public static final String DOWNLOAD_URL_KEY = "downloadURL";
   public static final String UPLOAD_URL_KEY = "uploadURL";
+  public static final String CRC_KEY = "crc";
```
mcvsubbu: should we call it 'matchingCrc'? This is the etag sent in the header, I presume?

pinot-minion/src/main/java/com/linkedin/pinot/minion/executor/BaseTaskExecutor.java
```diff
@@ -32,4 +34,10 @@ public void setMinionContext(@Nonnull MinionContext minionContext) {
   public void cancel() {
     _cancelled = true;
   }
+
+  @Override
+  public int uploadSegment(String uri, final String fileName, final InputStream inputStream, final long lengthInBytes,
+      FileUploadUtils.SendFileMethod httpMethod, String crc) {
```
mcvsubbu: Please rename 'crc' to 'ifMatchSegmentCrc'
Also, crc should be a long?

pinot-minion/src/main/java/com/linkedin/pinot/minion/executor/ConvertToRawIndexTaskExecutor.java
```diff
@@ -42,6 +42,7 @@ public void executeTask(@Nonnull PinotTaskConfig pinotTaskConfig) {
     String segmentName = configs.get(MinionConstants.SEGMENT_NAME_KEY);
     String downloadURL = configs.get(MinionConstants.DOWNLOAD_URL_KEY);
     String uploadURL = configs.get(MinionConstants.UPLOAD_URL_KEY);
+    String crc = configs.get(MinionConstants.CRC_KEY);
```
mcvsubbu: rename 'crc' to originalSegmentCrc?

pinot-core/src/main/java/com/linkedin/pinot/core/common/MinionConstants.java
mcvsubbu: please add some comments on CRC_KEY. Also, suggest renaming it to ORIGINAL_SEGMENT_CRC_KEY

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadUtils.java
mcvsubbu: suggest adding a Header[] argument instead of crc

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadUtils.java
jenniferdai: None of the apache classes have the if-match header

pinot-minion/src/main/java/com/linkedin/pinot/minion/executor/BaseTaskExecutor.java
```diff
@@ -32,4 +37,20 @@ public void setMinionContext(@Nonnull MinionContext minionContext) {
...
+
+  @Override
+  public int uploadSegment(String uri, final String fileName, final InputStream inputStream, final long lengthInBytes,
+      FileUploadUtils.SendFileMethod httpMethod, String originalSegmentCrc) {
+    List<Header> headers = new ArrayList<>();
+    Header header = new Header("If-Match", originalSegmentCrc);
+
+    // A pragma header is an implementation specific field that may have various effects along the
+    // request response chain. We will tell the controller minion is sending the segment.
+    Header minionHeader = new Header("Pragma", "Minion");
```
mcvsubbu: Pragma is an old http 1.0 compat header, and seems to be not he right one to use here. 

I think you can use "User-Agent", and set the value to "pinot-minion-<version>". Otherwise, you can use a header like "X-Sender-Identity" (or, anything prefixed with an X). Let us chat about this and reach a conclusion.

jfim: As Subbu mentioned, the `User-Agent` header is the correct header to use. In this case, what you want to do is to grab the default user agent value (see `DefaultHttpParams.getDefaultParams()`) and append something that indicates the minion version and/or job version, so that it looks something like this: `Jakarta Commons-HttpClient/3.1 Pinot-Minion/1.0`.

This way, if for some reason in the future we need to do some special processing for different versions of the HttpClient library or the minion code, then we can check exactly which version is used as opposed to having a single version.

Also, is the code generic for all minion based pushes or only for certain job types? If so, it might make more sense to specify the job type instead (or even both, eg. `Pinot-Minion/1.0 Pinot-Job-Foobar/1.2`).

jenniferdai: That's a good idea. I can add specific job types. Thanks

pinot-minion/src/main/java/com/linkedin/pinot/minion/executor/BaseTaskExecutor.java
```diff
@@ -32,4 +40,22 @@ public void setMinionContext(@Nonnull MinionContext minionContext) {
...
     _cancelled = true;
   }
+
+  @Override
+  public int uploadSegment(String uri, final String fileName, final InputStream inputStream, final long lengthInBytes,
+      FileUploadUtils.SendFileMethod httpMethod, String originalSegmentCrc, String jobType) {
+    List<Header> headers = new ArrayList<>();
+    Header ifMatchHeader = new Header(HttpHeaders.IF_MATCH, originalSegmentCrc);
+    HttpParams httpParams = DefaultHttpParams.getDefaultParams();
+    String userAgentParameter = String.valueOf(httpParams.getParameter("http.useragent"));
```
mcvsubbu: please use a static final string  for "http.useragent", better yet, try to find the definition for this string in the HttpParams (or another) object.

pinot-minion/src/main/java/com/linkedin/pinot/minion/executor/BaseTaskExecutor.java
```diff
@@ -32,4 +40,22 @@ public void setMinionContext(@Nonnull MinionContext minionContext) {
...
+
+    // TODO: Add proper version instead of hardcoding
+    userAgentParameter += " Pinot-Minion-" + jobType + "/1.0";
```
mcvsubbu: please use static final strings for "Pinot-Minion" and the version.

We will fill in the version later.

PR #2202, opened 2017-12-08
Adding committing field to segment metadata

jenniferdai:


pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
```diff
@@ -289,6 +289,7 @@ public ServerType getServerType() {
       public static final String DOWNLOAD_URL = "segment.offline.download.url";
       public static final String PUSH_TIME = "segment.offline.push.time";
       public static final String REFRESH_TIME = "segment.offline.refresh.time";
+      public static final String SEGMENT_PUSH_STATUS = "segment.push.status";
```
mcvsubbu: please add some comments here as to what this is, and how it is used

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -345,6 +345,8 @@ private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, Strin
         File indexDir = files[0];
 
         SegmentMetadata segmentMetadata = new SegmentMetadataImpl(indexDir);
+        segmentMetadata.getSegmentPushStatus();
```
mcvsubbu: not used? you can remove this line

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
```diff
@@ -152,6 +153,7 @@ public SegmentMetadataImpl(OfflineSegmentZKMetadata offlineSegmentZKMetadata) {
     _creationTime = offlineSegmentZKMetadata.getCreationTime();
     _pushTime = offlineSegmentZKMetadata.getPushTime();
     _refreshTime = offlineSegmentZKMetadata.getRefreshTime();
+    _segmentPushStatus = offlineSegmentZKMetadata.getSegmentPushStatus();
```
mcvsubbu: why do you need push status in segment metadata impl?

pinot-common/src/main/java/com/linkedin/pinot/common/metadata/segment/OfflineSegmentZKMetadata.java
```diff
@@ -32,6 +32,9 @@
   private String _downloadUrl = null;
   private long _pushTime = Long.MIN_VALUE;
   private long _refreshTime = Long.MIN_VALUE;
+  private String _segmentPushStatus = NONE;
```
mcvsubbu: your comments say it will be populated only when segment is committing. Are you populating it with the string "NONE" even otherwise?

pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
```diff
@@ -289,6 +289,13 @@ public ServerType getServerType() {
       public static final String DOWNLOAD_URL = "segment.offline.download.url";
       public static final String PUSH_TIME = "segment.offline.push.time";
       public static final String REFRESH_TIME = "segment.offline.refresh.time";
+      /**
+       * This field will be populated only for use cases that are refreshing segments and
+       * have minion configured. It will contain COMMITTING + System.currentMillis(), time right
+       * before any permanent changes happen to the segment, aka, right before
```
mcvsubbu: will be nice to have an example here. Maybe "COMMITTING-1513013204080" where time is in msecs since epoch? Better yet, something like:  "COMMITTING-20171210T09:31:55.765" ?

pinot-common/src/main/java/com/linkedin/pinot/common/metadata/segment/OfflineSegmentZKMetadata.java
jenniferdai: from my understanding, in zk, it will always be populated with something, whether it is the string "null" or something else. By that logic, I thought it would be better to always populate it with NONE to make it less confusing, which I was planning on doing inside upload segment though I could also do it here. I don't know what the convention we want to follow for "empty" fields in zk is; for example, if you look at the field schema_name, I've seen 
1. empty string
2. the string "null" (which I feel is very confusing when doing comparisons)
3. the string NO_SCHEMA

Do you have a suggestion about the convention that I should be following?

mcvsubbu:
You should move this to all metadata, not just offline zk metadata

pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
jfim: The human-readable timestamp would be better. Make sure that it's in UTC time though, otherwise there are some edge cases around DST which you really don't want to deal with for this kind of feature.

jenniferdai:
Starting another one for clarity: https://github.com/linkedin/pinot/pull/2207

PR #2204, opened 2017-12-09
Enhanced raw index readers/writers to support uncompressed data.

mayankshriv:
Metric columns are good candidate for raw index (no-dictionary).
However, in the current implementation, random access of metric columns
is slow due to de-compression. This change allows us to support
uncompressed data in raw index form, so it can be fetched efficiently.

1. Added total docs, compression type and data start offset to the header. Adding
data start offset in the header allows for adding more header
information in future, without fixing where data starts.

2. Compression type is now configured by SegmentColumnarIndexCreator,
instead of being hardcoded. Also, make compression optional.

3. Bumped up the writer version to 2, and enhanced reader to be able to
read the current as well as previous version.

4. Added tests for un-compressed format, as well as for backward
compatibility.


pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
```diff
@@ -105,17 +140,21 @@ protected ByteBuffer getChunkForRow(int row, ChunkReaderContext context) {
...
-    uncompressedBuffer.clear();
+    ByteBuffer decompressedBuffer = context.getChunkBuffer();
+    decompressedBuffer.clear();
 
     try {
-      _chunkDecompressor.decompress(_dataBuffer.toDirectByteBuffer(chunkPosition, chunkSize), uncompressedBuffer);
+      if (_chunkDecompressor != null) {
+        _chunkDecompressor.decompress(_dataBuffer.toDirectByteBuffer(chunkPosition, chunkSize), decompressedBuffer);
+      } else {
+        decompressedBuffer.put(_dataBuffer.toDirectByteBuffer(chunkPosition, chunkSize));
```
mcvsubbu: Do we need the copy here? can we return a slice of the dataBuffer here?

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
```diff
@@ -51,13 +54,14 @@
    * Constructor for the class.
    *
    * @param pinotDataBuffer Data buffer
-   * @param decompressor Data decompressor
    */
-  public BaseChunkSingleValueReader(PinotDataBuffer pinotDataBuffer, ChunkDecompressor decompressor) {
-    _chunkDecompressor = decompressor;
+  public BaseChunkSingleValueReader(PinotDataBuffer pinotDataBuffer) {
```
mcvsubbu: will be super useful if you can add a picture here, with different fiends in version 1 and beyond.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/v1/FixedByteChunkSingleValueWriter.java
```diff
@@ -47,24 +47,24 @@
 @NotThreadSafe
 public class FixedByteChunkSingleValueWriter extends BaseChunkSingleValueWriter {
 
-  private static final int VERSION = 1;
+  private static final int VERSION = 2;
```
mcvsubbu: suggest rename VERSION to CURRENT_VERSION

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/v1/VarByteChunkSingleValueWriter.java
```diff
@@ -48,7 +48,7 @@
 
   private static final int INT_SIZE = Integer.SIZE / Byte.SIZE;
   private static final Charset UTF_8 = Charset.forName("UTF-8");
-  private static final int VERSION = 1;
+  private static final int VERSION = 2;
```
mcvsubbu: CURRENT_VERSION?

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteChunkSingleValueReaderWriteTest.java
```diff
@@ -181,15 +205,50 @@ public void testDouble()
...
+        actual = reader.getDouble(i);
+        Assert.assertEquals(actual, expected[i]);
+      }
     }
     reader.close();
     FileUtils.deleteQuietly(outFile);
   }
+
+  /**
+   * This test ensures that the reader can read in an data file from previous version.
```
mcvsubbu: s/previous version/version 1/

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteChunkSingleValueReaderWriteTest.java
```diff
@@ -181,15 +205,50 @@ public void testDouble()
...
+   * @throws IOException
+   */
+  @Test
+  public void testBackwardCompatibility()
+      throws IOException {
+    // Get v1 from resources folder
+    ClassLoader classLoader = getClass().getClassLoader();
+    String fileName = "data/fixedByteSVRDoubles.v1";
```
mcvsubbu: Nice. thanks for doing this.

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/VarByteChunkSingleValueReaderWriteTest.java
```diff
@@ -91,4 +91,37 @@ public void test()
     reader.close();
     FileUtils.deleteQuietly(outFile);
   }
+
+  /**
+   * This test ensures that the reader can read in an data file from previous version.
```
mcvsubbu: s/previous version/version 1/

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
mayankshriv: Yes, it does not work otherwise, because the decompressed buffer in context is not updated, the subsequent reads don't get the data. I had thought about of an option to overwrite the chunkBuffer in the context, but that would require making it mutable, and didn't like it. If a client sets chunkBuffer in the context, I couldn't convince myself to over-write it under the hood.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
mayankshriv: Updated header layout in FixedByteChunkSingleValueWriter to explicitly mention which fields are in version 2 onwards.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/v1/FixedByteChunkSingleValueWriter.java
mayankshriv: Done

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/v1/VarByteChunkSingleValueWriter.java
mayankshriv: Done

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/FixedByteChunkSingleValueReaderWriteTest.java
mayankshriv: Done

pinot-core/src/test/java/com/linkedin/pinot/index/readerwriter/VarByteChunkSingleValueReaderWriteTest.java
mayankshriv: Done

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
```diff
@@ -66,18 +70,49 @@ public BaseChunkSingleValueReader(PinotDataBuffer pinotDataBuffer, ChunkDecompre
...
     // Nothing to close here.
   }
 
+  /**
+   * Returns the total number of docs in the file. This is only support from v2 onwards.
+   * Returns -1 for v1.
+   *
+   * @return Total docs for v2 onwards, -1 for v1.
+   */
+  public int getTotalDocs() {
```
kishoreg: why do we need this method.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
kishoreg: why not have no-op implementation for NONE compression case. We can get rid of the if check here. Compiler would optimize it anyway at runtime.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/FixedByteChunkSingleValueReader.java
```diff
@@ -34,12 +33,20 @@
...
-  public FixedByteChunkSingleValueReader(PinotDataBuffer pinotDataBuffer, ChunkDecompressor uncompressor)
+  public FixedByteChunkSingleValueReader(PinotDataBuffer pinotDataBuffer)
       throws IOException {
-    super(pinotDataBuffer, uncompressor);
+    super(pinotDataBuffer);
+  }
+
+  @Override
+  public int getInt(int row) {
+    if (!isCompressed()) {
```
kishoreg: why do we need this check for every access? Better to separate them into multiple implementations and get rid of the check for every access.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
```diff
@@ -66,18 +70,49 @@ public BaseChunkSingleValueReader(PinotDataBuffer pinotDataBuffer, ChunkDecompre
 
     _lengthOfLongestEntry = _dataBuffer.getInt(headerOffset);
     headerOffset += INT_SIZE;
+
+    int dataHeaderStart = headerOffset;
+    if (version > 1) {
```
kishoreg: if this is not widely used. Get rid of the initial version and write a converter from v1 to v2. I prefer keeping the runtime code simpler and fix everything as part of loading the segment if possible.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
mayankshriv: Yes, I had thought about that. And convinced myself to not do so because no-op would have required to copy the data, which the initial implementation did not have (line 150). But that led to a bug and I ended up copying the data, so now I can go back to using no-op compressor. Will update.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/FixedByteChunkSingleValueReader.java
mayankshriv: Had thought about that too, it was leading to either lot of duplicate code, or adding additional level of class hierarchy. And I wasn't really strongly feeling about necessity of either.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
mayankshriv: Agree with that. Unfortunately, It is used in production already, and would require a few TB of data to be converted.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
mayankshriv: It was needed in intermediate state of the code, but ended up being unused. However, IDE did not show it as unused, so missed to clean it up. Will clean.

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/BaseChunkSingleValueReader.java
```diff
@@ -105,17 +131,17 @@ protected ByteBuffer getChunkForRow(int row, ChunkReaderContext context) {
...
-    ByteBuffer uncompressedBuffer = context.getChunkBuffer();
-    uncompressedBuffer.clear();
+    ByteBuffer decompressedBuffer = context.getChunkBuffer();
+    decompressedBuffer.clear();
 
     try {
-      _chunkDecompressor.decompress(_dataBuffer.toDirectByteBuffer(chunkPosition, chunkSize), uncompressedBuffer);
+      _chunkDecompressor.decompress(_dataBuffer.toDirectByteBuffer(chunkPosition, chunkSize), decompressedBuffer);
     } catch (IOException e) {
       LOGGER.error("Exception caught while decompressing data chunk", e);
```
kishoreg: having chunkPosition and some reference to PinotDataBuffer will help in debugging if there is a corruption

pinot-core/src/main/java/com/linkedin/pinot/core/io/reader/impl/v1/FixedByteChunkSingleValueReader.java
kishoreg: there will be a performance hit with this check. Create another issue to fix this later if not in this PR/

mayankshriv: Created issue #2223 to track this.

mayankshriv:
IMO, metrics can always use this indexing scheme. @mcvsubbu plans to implement the same.

PR #2209, opened 2017-12-11
[TE] fix rootcause chart color bug

justYves:
### What's new: 
- fix the color bug: this was caused by ember-c3 using the color string, instead of the mapping. Adding a mapping helper fixed the issue

### Test: 
Visual Sanity check

### DIFF
--- a/thirdeye/thirdeye-frontend/app/helpers/utils.js
+++ b/thirdeye/thirdeye-frontend/app/helpers/utils.js
@@ -15,6 +15,18 @@ const metricColors = [
 ];
 
 // TODO load from config
+// colors mapping for charts
+const colorMapping = {
+  blue: '#0091CA',
+  green: '#469A1F',
+  red: '#FF2C33',
+  purple: '#827BE9',
+  orange: '#E55800',
+  teal: '#0E95A0',
+  pink: '#FF1B90'
+};
+
+// TODO load from config
 // colors for events
 const eventColorMapping = {
   holiday: 'green',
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
```diff
@@ -354,7 +364,7 @@ export default Ember.Component.extend({
       return {
         timestamps: [entities[urn].start, entities[urn].end || entities[urn].start],
         values: [val, val],
-        color: entities[urn].color,
+        color: getColor(entities[urn].color),
```
apucher: please also delete the code in timeseries-chart that applies a CSS class from the color string (lines 73-75)

thirdeye/thirdeye-frontend/app/helpers/utils.js
```diff
@@ -228,4 +249,24 @@ export function findLabelMapping(label, config) {
...
+  appendTail,
+  hasPrefix,
+  filterPrefix,
+  toBaselineRange,
+  toFilters,
+  toFilterMap,
+  findLabelMapping,
+  toMetricLabel,
+  toColor,
+  getColor});
```
apucher: line break

thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
```diff
@@ -2,7 +2,17 @@ import Ember from 'ember';
...
+  getColor,
+  toBaselineUrn,
+  toMetricUrn,
+  filterPrefix,
+  hasPrefix,
+  stripTail,
+  toBaselineRange,
+  toFilters,
+  toMetricLabel
+} from 'thirdeye-frontend/helpers/utils';
```
apucher: i wonder if there's a better way of doing this throughout our project

justYves: We could also import the whole utils as 'utils' and use them in the code as utils.toFilters etc...

thirdeye/thirdeye-frontend/app/helpers/utils.js
```diff
@@ -51,6 +63,15 @@ export function checkStatus(response, mode = 'get', recoverBlank = false) {
   }
 }
 
+/**
+ * Converts the color string into its rgba cunterpart
```
apucher: nice typo. lets delete the method if its not used - we can keep the mapping const

justYves: Oops haha

PR #2210, opened 2017-12-11
[PINOT-5458] Adding Https fetcher

mcvsubbu:
See comments in the class regarding the current features/limitations of this class.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
@@ -17,7 +17,10 @@
 package com.linkedin.pinot.common.segment.fetcher;
 
 import com.linkedin.pinot.common.Utils;
+import com.linkedin.pinot.common.exception.PermanentDownloadException;
 import com.linkedin.pinot.common.utils.FileUploadUtils;
+import com.linkedin.pinot.common.utils.retry.RetryPolicies;
+import com.linkedin.pinot.common.utils.retry.RetryPolicy;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -29,6 +32,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
+import java.util.concurrent.Callable;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.KeyManagerFactory;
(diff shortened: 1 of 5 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
```diff
@@ -0,0 +1,169 @@
...
+      return kmf.getKeyManagers();
+    } catch (Exception e) {
+      Utils.rethrowException(e);
+    }
+    return null;
+  }
+
+  @Override
+  public void fetchSegmentToLocal(String uri, File tempFile) throws Exception {
+    URL url = new URL(uri);
```
jenniferdai: do we want to implement retries here like we do in httpsegmentfetcher?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
```diff
@@ -0,0 +1,169 @@
...
+    FileInputStream is = new FileInputStream(new File(serverCACertFile));
+    CertificateFactory certificateFactory = CertificateFactory.getInstance(CERTIFICATE_TYPE);
+    X509Certificate cert = (X509Certificate) certificateFactory.generateCertificate(is);
+
+    String serverKey = "https-server";
+    KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
+    trustStore.load(null);
+    trustStore.setCertificateEntry(serverKey, cert);
+
+    TrustManagerFactory tmf = TrustManagerFactory.getInstance(CERTIFICATE_TYPE);
```
jenniferdai: Could we leave some comments regarding this logic? Spoke offline and it made sense, but might be difficult in the future to decrypt

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/NoOpFetcher.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.linkedin.pinot.common.segment.fetcher;
+
+import java.io.File;
+import org.apache.commons.configuration.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// This class is provided as a means to override any standard fetchers that will not be used
+// in a given Pinot installation, because
```
jenniferdai: because?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
```diff
@@ -0,0 +1,169 @@
...
+      Utils.rethrowException(e);
+    }
+  }
+
+  private TrustManager[] setupTrustManagers(Configuration configs)
+      throws CertificateException, KeyStoreException, IOException, NoSuchAlgorithmException {
+    // This is the cert authority that validates server's cert, so we need to put it in our
+    // trustStore.
+    String serverCACertFile = configs.getString(CONFIG_OF_SERVER_CA_CERT);
+    if (serverCACertFile == null) {
```
cyenjung: If encryption is top priority, shouldn't we let users enable TrustAllCerts in an explicit flag? I.e., if serverCACertFile is null and useTrustAllCert is true. Otherwise, users might not notice trust manager is accpeting all certificates if s/he is not looking at the log when booting up the machine.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
mcvsubbu: I have added some links. Most SSL APIs don't even have visible names for the method arguments, so there is little that can be done to name the strings used. I have tried my best after reading some of the links that I have added in the class.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
mcvsubbu: No, this class will do the same as HttpSegmentFetcher, returning either a permanent or temporary exception, and let the layer above retry.

mcvsubbu: Yes, good point, I will add retries.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/NoOpFetcher.java
mcvsubbu: :)

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadUtils.java
```diff
@@ -319,6 +294,35 @@ public static long getFile(String url, File file) throws Exception {
     }
   }
 
+  public static long storeFile(String url, File file, int responseCode, long contentLength, InputStream is)
+      throws IOException {
+    if (responseCode >= 400) {
+      if (contentLength > 0) {
+        // don't read more than 1000 bytes
```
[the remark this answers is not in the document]
mcvsubbu: Not sure, but that is existing code. I have not looked into modifying that. My guess would be that first 1k bytes of a server should be good enough to indicate an error. There is no point in reading a ton of data, given that it is a 400 anyway.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
mcvsubbu: Hmm, so u are suggesting adding another config variable. Let me think some more, and maybe we can add one.

cyenjung: We could define this special string: "AcceptAll"; When "AcceptAll".equals(serverCACertFile) holds, we trust all certificate. If serverCACertFile is null, we throw an exception and hence users know HttpsSegmentFetcher is not working properly. After former two cases, we assume that serverCACertFile is the path to a file.

PR #2211, opened 2017-12-12
[TE] rootcause poc - message about unavailable data for heatmap

apucher:
image


apucher:
yeah styling is questionable

PR #2213, opened 2017-12-12
[TE] Enable Anomaly Feed and Email Content Formatter

cyy0714:
This is the last commit for alert refactoring. This pr involves introducing anomaly feed and content formatter to alert tasks. After this push, users are able to modify their preference through alert config.


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/alert/v2/AlertTaskRunnerV2.java
```diff
@@ -135,126 +172,144 @@ private void sendAnomalyReport() throws Exception {
...
                 lastNotifiedAnomaly, false);
         if (CollectionUtils.isNotEmpty(filteredAnomalies)) {
           mergedAnomaliesAllResults.addAll(filteredAnomalies);
         }
       }
+    }
+    // apply filtration rule
+    results = AlertFilterHelper.applyFiltrationRule(mergedAnomaliesAllResults, alertFilterFactory);
+
+    // only anomalies detected by default scheduler are sent
```
cyenjung: Please add "; the anomalies detected during replay or given by users will not be sent" to make it clearer.

cyy0714:
Yes. The old alert pipeline, using emailConfig, is still there and runnable. They will use the MultipleAnomalyEmailContentFormatter to format alerts. The new one will follow the new pipeline. They will co-exist in the system. 
The to-do will clean up the old pipeline and migrate all alerts to the new pipeline. Then we don't need the old pipeline anymore.

cyenjung:
I will approve it for now. We still need the merge plan because I don't see the new pipeline plays well with the grouping logic for Sitespeed application.

PR #2214, opened 2017-12-12
[TE] rootcause poc - collapse anomaly header on existing feedback

apucher:


thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -116,6 +113,18 @@ export default Ember.Component.extend({
     return this.get('status') === 'NO_FEEDBACK';
   }),
 
+  isHidden: Ember.computed(
+    'isNeedFeedback',
```
justYves: nit: `needsFeedback` might be more correct (grammatically)

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -116,6 +113,18 @@ export default Ember.Component.extend({
...
+    'isHiddenUser',
+    function () {
+      const { isNeedFeedback, isHiddenUser } = this.getProperties('isNeedFeedback', 'isHiddenUser');
+      if (isHiddenUser === null) {
```
justYves: do we need to explicitely check for null (i.e. undefined should not pass? ) 
Other wise we can just do `if (isHiddenUser)`

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -130,7 +139,7 @@ export default Ember.Component.extend({
 
     toggleHidden() {
       const { isHidden } = this.getProperties('isHidden');
-      this.setProperties({ isHidden: !isHidden });
+      this.setProperties({ isHiddenUser: !isHidden });
```
justYves: you can also do `this.toggleProperty('isHidden')`

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
apucher: yeah that's a habit of prefixing any boolean return type with 'is'. I'll rename it

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
apucher: it's a tri-state variable - "default", "true", "false". null is the default value. spelling all of this out as an enum seems overkill

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
apucher: the one-way interaction between isHidden and isHiddenUser is important here

PR #2216, opened 2017-12-12
Adding minion specific logic to segment upload

jenniferdai:


pinot-common/src/main/java/com/linkedin/pinot/common/metadata/segment/SegmentZKMetadata.java
```diff
@@ -53,6 +53,7 @@
   private List<String> _optimizations;
   private String _segmentPushStatus;
   private Map<String, String> _customFields;
+  private int _version;
```
mcvsubbu: why are we versioning segment zk metadata?

pinot-common/src/main/java/com/linkedin/pinot/common/metadata/ZKMetadataProvider.java
```diff
@@ -142,6 +142,13 @@ public static void setInstancePartitionAssignmentFromPropertyStore(ZkHelixProper
         partitionMappingZKMetadata.toZNRecord(), AccessOption.PERSISTENT);
   }
 
+  public static boolean setOfflineSegmentZKMetadataOnlyIfCorrectVersion(ZkHelixPropertyStore<ZNRecord> propertyStore,
```
mcvsubbu: Are you intending to use znode version here?

jenniferdai: yes, i pass in the version from offline segment zk metadata, where i set the znode version

pinot-common/src/main/java/com/linkedin/pinot/common/metadata/ZKMetadataProvider.java
```diff
@@ -168,6 +175,15 @@ public static OfflineSegmentZKMetadata getOfflineSegmentZKMetadata(
     return new OfflineSegmentZKMetadata(znRecord);
   }
 
+  @Nullable
+  public static ZNRecord getZnRecord(
```
mcvsubbu: suggest name getSegmentMetadataZnRecord

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,21 +416,63 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+    if (isMinionConfigured) {
+      // Minion configured use case so that we write to zk sparingly
+      String segmentPushStatus = offlineSegmentZKMetadata.getSegmentPushStatus();
+      SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd'T'HH:mm:ss.SSS");
+      Date dt = new Date();
+      String currentHumanReadableTime = sdf.format(dt);
+      offlineSegmentZKMetadata.setSegmentPushStatus("COMMITTING-" + currentHumanReadableTime);
+
+      if (segmentPushStatus != null) {
+        if (tooLate(currentHumanReadableTime, sdf)) {
```
mcvsubbu: why not pass 'dt' as the parameter instead of human readable time and the format and then extracgting the time inside tooLate()?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,21 +416,63 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+          LOGGER.warn("Segment was not properly committed, replacing segment {} for table {}", segmentName, tableName);
```
mcvsubbu: Bump a metric here. We can raise an alert if this condition hits. You should log it as an ERROR

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,21 +416,63 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+      }
+
+      if (!_pinotHelixResourceManager.updateZkMetadataAtomically(offlineSegmentZKMetadata, znRecord.getVersion())) {
+        throw new ControllerApplicationException(LOGGER, "Retry later. ZK update failed.", Response.Status.INTERNAL_SERVER_ERROR);
+      }
+
+      try {
+        minionExpectedCrc = Long.parseLong(headers.getHeaderString(HttpHeaders.IF_MATCH));
+      } catch (NumberFormatException e) {
+        LOGGER.warn("Could not parse crc {}", minionExpectedCrc, e);
```
mcvsubbu: Don't you need to return if you cannot parse CRC? Minion is always expected to send the IF_MATCH header, right?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,21 +416,63 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+        } else {
+          throw new ControllerApplicationException(LOGGER, "Please retry later, segment upload in progress", Response.Status.INTERNAL_SERVER_ERROR);
+        }
+      }
+
+      if (!_pinotHelixResourceManager.updateZkMetadataAtomically(offlineSegmentZKMetadata, znRecord.getVersion())) {
+        throw new ControllerApplicationException(LOGGER, "Retry later. ZK update failed.", Response.Status.INTERNAL_SERVER_ERROR);
+      }
+
+      try {
```
mcvsubbu: You should put the entire rest of the logic in a try/catch and unset segmentPushStatus() in the finally block.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,21 +416,63 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
     }
-    // TODO: this will read table configuration again from ZK. We should optimize that
-    response = _pinotHelixResourceManager.refreshSegment(segmentMetadata, downloadUrl);
+
+    if (ifRefreshAnExistedSegment(
+        segmentMetadata, offlineSegmentZKMetadata, offlineTableName, segmentName, minionExpectedCrc, minionHeaderPresent)) {
+      response = _pinotHelixResourceManager.refreshSegment(segmentMetadata, downloadUrl);
+    } else {
+      String msg =
+          "Not refreshing identical segment " + segmentName + "of table " + offlineTableName + " with creation time " + segmentMetadata.getIndexCreationTime() + " and crc "
```
mcvsubbu: we don't know that it is identical. The method ifRefreshAnExistedSegment() also refuses segments that are older, right? (And rename that method to 'shouldRefreshSegment()' please, and move the log to that method, so that it logs the right reason why a segment was not refreshed.)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: yes, but currently there is backfilldatetimeconverter, which is a minion job inside pinot-tools. There are some dependency issues to solve to get that to also use the same upload method, will add a todo to change after refactor

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: I use the timestamp to set offline zk metadata as well

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: i have added in the writes to segment push status manually, I like the finally idea, but I want to preserve the ControllerApplicationException format we have

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -51,6 +51,7 @@
   private static final String JERSEY_ADMIN_API_PORT = "jersey.admin.api.port";
   private static final String JERSEY_ADMIN_IS_PRIMARY = "jersey.admin.isprimary";
   private static final String ACCESS_CONTROL_FACTORY_CLASS = "controller.admin.access.control.factory.class";
+  private static final String SEGMENT_PUSH_TIMEOUT_MILLIS = "segment.push.timeout.millis";
```
mcvsubbu: will be good to add a comment here on what this config means. Also, suggest rename to "segment.commit.timeout.millis"

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,124 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd'T'HH:mm:ss.SSS");
+        Date dt = new Date();
+        String currentHumanReadableTime = sdf.format(dt);
+        offlineSegmentZKMetadataWithPushStatus.setSegmentPushStatus("COMMITTING-" + currentHumanReadableTime);
+
+        if (segmentPushStatus != null) {
+          if (tooLate(dt)) {
+            LOGGER.error("Segment was not properly committed, replacing segment {} for table {}", segmentName,
+                tableName);
+            _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_SEGMENT_UPLOAD_ERROR, 1L);
```
mcvsubbu: You should bump a different (new) metric. ControllerMeter.NUMBER_SEGMENT_COMMIT_TIMEOUT_EXCEEDED. We will have alerts on this one.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,124 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+        }
+
+        if (!_pinotHelixResourceManager
+            .updateZkMetadataAtomically(offlineSegmentZKMetadataWithPushStatus, znRecord.getVersion())) {
+          // Write committing status to zk
+          throw new ControllerApplicationException(LOGGER, "Retry later. ZK update failed.", Response.Status.INTERNAL_SERVER_ERROR);
+        }
+
+        try {
+          minionExpectedCrc = Long.parseLong(headers.getHeaderString(HttpHeaders.IF_MATCH));
```
mcvsubbu: Will hadoop job also send this header? If not, should you not check for miniontHeaderPresent?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,124 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+
+      if (!response.isSuccessful()) {
+        _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_SEGMENT_UPLOAD_ERROR, 1L);
+        throw new ControllerApplicationException(LOGGER, "Error uploading segment", Response.Status.INTERNAL_SERVER_ERROR);
+      }
+      return response;
+    } catch (ControllerApplicationException e) {
+      throw e;
+    } finally {
+      if (isMinionConfigured) {
```
mcvsubbu: You need to check against whether or not the metadata was updated. Not if (minionConfigured)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,124 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+
+  private boolean shouldRefreshSegment(@Nonnull SegmentMetadata newSegmentMetadata,
+      @Nonnull OfflineSegmentZKMetadata existedSegmentZKMetadata, @Nonnull String offlineTableName,
+      @Nonnull String segmentName, long minionExpectedCrc, boolean minionHeaderPresent) {
+    long newCreationTime = newSegmentMetadata.getIndexCreationTime();
+    long existedCreationTime = existedSegmentZKMetadata.getCreationTime();
+    long newCrc = Long.valueOf(newSegmentMetadata.getCrc());
+    long existedCrc = existedSegmentZKMetadata.getCrc();
+
+    if (!minionHeaderPresent) {
```
mcvsubbu: Add a comment here
"we are skipping this check for minion because...."

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,124 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+        response = PinotResourceManagerResponse.SUCCESS_RESPONSE;
+      }
+
+      if (!response.isSuccessful()) {
+        _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_SEGMENT_UPLOAD_ERROR, 1L);
+        throw new ControllerApplicationException(LOGGER, "Error uploading segment", Response.Status.INTERNAL_SERVER_ERROR);
+      }
+      return response;
+    } catch (ControllerApplicationException e) {
+      throw e;
```
mcvsubbu: Utils.retrhowEXception

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,124 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+        }
+      }
+
+      if (downloadUrl == null) {
+        downloadUrl = moveSegmentToPermanentDirectory(provider, tableName, segmentName, tempTarredSegmentFile);
+      }
+
+      if (shouldRefreshSegment(segmentMetadata, originalOfflineSegmentZkMetadata, offlineTableName, segmentName,
+          minionExpectedCrc, minionHeaderPresent)) {
+        response = _pinotHelixResourceManager.refreshSegment(segmentMetadata, originalOfflineSegmentZkMetadata, downloadUrl);
```
mcvsubbu: add a boolean here saying that metadata has been set.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,124 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
       downloadUrl = moveSegmentToPermanentDirectory(provider, tableName, segmentName, tempTarredSegmentFile);
       _pinotHelixResourceManager.addNewSegment(segmentMetadata, downloadUrl);
       response = PinotResourceManagerResponse.SUCCESS_RESPONSE;
       return response;
     }
-    if (downloadUrl == null) {
-      downloadUrl = moveSegmentToPermanentDirectory(provider, tableName, segmentName, tempTarredSegmentFile);
+
+    final OfflineSegmentZKMetadata originalOfflineSegmentZkMetadata = new OfflineSegmentZKMetadata(znRecord);
+    OfflineSegmentZKMetadata offlineSegmentZKMetadataWithPushStatus = new OfflineSegmentZKMetadata(znRecord);
```
mcvsubbu: move this line below 435

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -65,6 +68,7 @@
   private static final int DEFAULT_JERSEY_ADMIN_PORT = 21000;
   private static final String DEFAULT_ACCESS_CONTROL_FACTORY_CLASS =
       "com.linkedin.pinot.controller.api.access.AllowAllAccessFactory";
+  private static final long DEFAULT_SEGMENT_COMMIT_TIMEOUT_MILLIS = 10 * 60 * 1000; // 10 minutes
```
mcvsubbu: This is a very long time to store a segment, but I guess we are being conservative here. That is fine. In that case, perhaps the sender (hadoop job) should set the socket read timeout to this number, maybe. Not sure if we can find a way to send 100 continue response as we get the segment... something to think about.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +416,132 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+    try {
+      if (isMinionConfigured) {
+        OfflineSegmentZKMetadata offlineSegmentZKMetadataWithPushStatus = new OfflineSegmentZKMetadata(znRecord);
+
+        // Minion configured use case so that we write to zk sparingly
+        String segmentPushStatus = originalOfflineSegmentZkMetadata.getSegmentPushStatus();
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd'T'HH:mm:ss.SSS");
+        Date dt = new Date();
+        String currentHumanReadableTime = sdf.format(dt);
+        offlineSegmentZKMetadataWithPushStatus.setSegmentPushStatus("COMMITTING-" + currentHumanReadableTime);
```
mcvsubbu: can we move this line to below 454?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +416,132 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+
+    final OfflineSegmentZKMetadata originalOfflineSegmentZkMetadata = new OfflineSegmentZKMetadata(znRecord);
+
+    long minionExpectedCrc = -1L;
+    boolean isMinionConfigured = offlineTableConfig.getTaskConfig() != null;
+    boolean minionHeaderPresent = headers.getHeaderString(HttpHeaders.USER_AGENT).contains(CommonConstants.Minion.MINION_HEADER);
+    boolean metadataRewritten = false;
+    try {
+      if (isMinionConfigured) {
+        OfflineSegmentZKMetadata offlineSegmentZKMetadataWithPushStatus = new OfflineSegmentZKMetadata(znRecord);
```
mcvsubbu: Can we move this line to below 454?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +416,132 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+          }
+        }
+
+        if (!_pinotHelixResourceManager
+            .updateZkMetadataAtomically(offlineSegmentZKMetadataWithPushStatus, znRecord.getVersion())) {
+          // Write committing status to zk
+          throw new ControllerApplicationException(LOGGER, "Retry later. ZK update failed.", Response.Status.INTERNAL_SERVER_ERROR);
+        }
+
+        if (minionHeaderPresent) {
```
mcvsubbu: we can do this check before we update metadata right?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +416,132 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+      } else {
+        response = PinotResourceManagerResponse.SUCCESS_RESPONSE;
+      }
+
+      if (!response.isSuccessful()) {
+        _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_SEGMENT_UPLOAD_ERROR, 1L);
+        throw new ControllerApplicationException(LOGGER, "Error uploading segment", Response.Status.INTERNAL_SERVER_ERROR);
+      }
+      return response;
+    } catch (ControllerApplicationException e) {
```
mcvsubbu: Do we need a catch here at all?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +416,132 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+    // Will never take this codepath, but the compiler expects something here because Utils.rethrowException doesn't
+    // register properly as throwing an exception
+    return null;
+  }
 
-    if (!response.isSuccessful()) {
-      _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_SEGMENT_UPLOAD_ERROR, 1L);
-      throw new ControllerApplicationException(LOGGER, "Error uploading segment", Response.Status.INTERNAL_SERVER_ERROR);
+  private boolean tooLate(Date date) {
+    Long timeoutMillis = _controllerConf.getSegmentCommitTimeoutMillis();
```
mcvsubbu: Use long instead of 'Long'

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +416,132 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+    Long currentMillis = System.currentTimeMillis();
```
mcvsubbu: Use long instead of Long

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,127 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+
+    long minionExpectedCrc = -1L;
+    boolean isMinionConfigured = offlineTableConfig.getTaskConfig() != null;
+    boolean minionHeaderPresent = headers.getHeaderString(HttpHeaders.USER_AGENT).contains(CommonConstants.Minion.MINION_HEADER);
+    boolean metadataRewritten = false;
+    if (isMinionConfigured) {
+      // Minion configured use case so that we write to zk sparingly
+      String segmentPushStatus = originalOfflineSegmentZkMetadata.getSegmentPushStatus();
+      SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd'T'HH:mm:ss.SSS");
+      Date dt = new Date();
```
Jackie-Jiang: Make sure this is the same as system time.

jenniferdai:
Addressed comments

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,127 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+      }
+
+      // TODO: Optimize this to only change refresh time if crc is same after changing crc of other minion tasks on data update
+      if (newCrc == existedCrc) {
+        LOGGER.info("Reject segment: {} in table: {} because of same CRC: {}", segmentName, offlineTableName, newCrc);
+        return false;
+      }
+    }
+
+    if (minionExpectedCrc != -1L && minionExpectedCrc != existedCrc) {
```
Jackie-Jiang: Please put this into a else block

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -413,27 +415,127 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+    boolean metadataRewritten = false;
+    if (isMinionConfigured) {
+      // Minion configured use case so that we write to zk sparingly
+      String segmentPushStatus = originalOfflineSegmentZkMetadata.getSegmentPushStatus();
+      SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd'T'HH:mm:ss.SSS");
+      Date dt = new Date();
+      String currentHumanReadableTime = sdf.format(dt);
+
+      if (segmentPushStatus != null) {
+        if (tooLate(dt)) {
```
Jackie-Jiang: (Critical) Should use the timestamp in segmentPushStatus

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: Good call. From the docs: /**
     * Allocates a <code>Date</code> object and initializes it so that
     * it represents the time at which it was allocated, measured to the
     * nearest millisecond.
     *
     * @see     java.lang.System#currentTimeMillis()
     */

PR #2217, opened 2017-12-12
[TE] Fix anomaly detection resource return type

cyenjung:
Change the return type of Endpoints from plain strings to Json strings.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/PrecisionRecallEvaluator.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/PrecisionRecallEvaluator.java
@@ -202,18 +202,18 @@ public class PrecisionRecallEvaluator {
     return evals;
   }
 
-  public Map<String, Double> toDoubleMap() {
-    Map<String, Double> evals = new HashMap<>();
+  public Map<String, Number> toNumberMap() {
+    Map<String, Number> evals = new HashMap<>();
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/PrecisionRecallEvaluator.java
```diff
@@ -200,6 +202,21 @@ public Properties toProperties() {
...
+    Map<String, Double> evals = new HashMap<>();
+    evals.put(RESPONSE_RATE, getResponseRate());
+    evals.put(PRECISION, getPrecision());
+    evals.put(WEIGHTED_PRECISION, getWeightedPrecision());
+    evals.put(RECALL, getRecall());
+    evals.put(TOTALALERTS, (double) getTotalAlerts());
+    evals.put(TOTALRESPONSES, (double) getTotalResponses());
+    evals.put(TRUEANOMALIES, (double) getTrueAnomalies());
+    evals.put(FALSEALARM, (double) getFalseAlarm());
+    evals.put(NEWTREND, (double) getTrueAnomalyNewTrend());
```
cecilynie: Hi Yen, here the get methods are all return (int) rather than (double), if casting them to double, will these shown on UI as "number of anomalies = 10.0"?

cyenjung: Fixed!

PR #2226, opened 2017-12-13
[TE] rootcause poc - default alignment to days for anomaly analysis range

apucher:


thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -198,10 +198,18 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
       const anomalyRangeUrns = filterPrefix(contextUrns, 'thirdeye:timerange:anomaly:');
       const analysisRangeUrns = filterPrefix(contextUrns, 'thirdeye:timerange:analysis:');
 
+      // thirdeye:timerange:anomaly:{start}:{end}
+      const anomalyRange = _.slice(anomalyRangeUrns[0].split(':'), 3, 5).map(i => parseInt(i, 10));
+
+      // thirdeye:timerange:analysis:{start}:{end}
+      // align to local end of day
+      const rawAnalysisRange = _.slice(analysisRangeUrns[0].split(':'), 3, 5).map(i => parseInt(i, 10));
```
justYves: How about deconstructing it this way: 
```
const [ start, end ] =  _.slice(analysisRangeUrns[0].split(':'), 3, 5).map(i => parseInt(i, 10));
```

then use `start` and `end` in analysisRange

PR #2233, opened 2017-12-14
[TE] UserReport Anomaly: feedback update by dimension

cecilynie:
Since user report anomaly will possibly have dimensional information. Feedback update needs to be extended to do dimensional matching.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/AnomaliesResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/AnomaliesResource.java
@@ -21,6 +21,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
+import javax.validation.constraints.NotNull;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/AnomaliesResource.java
```diff
@@ -499,6 +501,7 @@ public void updateFeedbackForAnomalyFunctionAndTimeRange(
       @PathParam("startTime") Long startTime,
       @PathParam("endTime") Long endTime,
       @PathParam("functionId") Long functionId,
+      @QueryParam("dimensionMap") String dimensionMapJSONString,
       @QueryParam("feedbackType") String feedbackType) {
 
     if (functionId == null) {
```
cyy0714: These checks can be done on with @NotNull decorator on the PathParam.

apucher: I should have used isNotEmpty() in the first place. This is stricter than not null

cecilynie: In fact changed to assign a default value to "", in case frontend don't pass in dimension information

cyenjung: The constructor of dimension map doesn't take empty strings. Fixing that issue here: https://github.com/linkedin/pinot/pull/2237. Please review it and push this PR with that one.

PR #2238, opened 2017-12-15
[PINOT-5458] Remove static initialization of classes in SegmentFetcherFactory

mcvsubbu:
Static initialization of classes in SegmentFetcher factory can have unneeded exception stack trace
when the system starts up. The HttpsSegmentFetcher, for example, will throw (and log) initialization
exceptions, even if it is overridden with some other class. This poses a problem for deployments with
automated exception detection, not to mention unnecessary exception logs confusing the admin when
they have configured a different class.

Removed static initialization in the SegmentFetcherFactory class.

If initSegmentFetcherFactory is not called, then map of protocol to SegmentFetcher is not populated.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
@@ -16,6 +16,7 @@
 package com.linkedin.pinot.common.segment.fetcher;
 
 import com.google.common.annotations.VisibleForTesting;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactoryTest.java
```diff
@@ -57,26 +57,35 @@ public void testInitSegmentFetcherFactory() throws Exception {
     Assert.assertEquals(((TestSegmentFetcher)SegmentFetcherFactory.getPreloadSegmentFetchers().get("test")).init_called, 1);
     ReplacedFetcher fetcher = (ReplacedFetcher)SegmentFetcherFactory.getPreloadSegmentFetchers().get(replacableProto);
     Assert.assertEquals(fetcher.getInitCalled(), 1);
+    SegmentFetcherFactory.getPreloadSegmentFetchers().clear();
```
mayankshriv: Should the clear be at the beginning of test? That would avoid any new tests that forgets to clear it at the end?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -66,42 +69,33 @@ private static void instantiateSegmentFetcher(String protocol, String className)
   // A config for a protocol may not specify the class name, but specify other configs params (used in init())
   public static void initSegmentFetcherFactory(Configuration segmentFetcherFactoryConfig) {
 
-    // Iterate through the configs to fill the map with any new configured classes or Overridden classes.
+    // Iterate through the configs to instantiate any configured classes first.
     Iterator segmentFetcherFactoryConfigIterator = segmentFetcherFactoryConfig.getKeys();
     while (segmentFetcherFactoryConfigIterator.hasNext()) {
       Object configKeyObject = segmentFetcherFactoryConfigIterator.next();
+      if (!configKeyObject.toString().endsWith(SEGMENT_FETCHER_CLASS_KEY)) {
```
Jackie-Jiang: Why do we need this? End with a class key seems weird.

pinot-common/src/test/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactoryTest.java
mcvsubbu: OK

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
mcvsubbu: This is the class config for the protocol. If we don't have this, we instantiate the class multiple times, and see any instantiation logs multiple times.

The first look instantiates any classes configured

the second one instantiates any remaining ones in the default list

and last loop calls init on all classes instantiated.

PR #2247, opened 2017-12-19
[TE] frontend - Updated test with await/async

ttbach:
Updated alerts acceptance test with `async/await`


ttbach:
@justYves LOL 😆

PR #2257, opened 2017-12-20
Adding purge time to metadata

jenniferdai:


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -426,17 +428,25 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
 
     long minionExpectedCrc = -1L;
     boolean isMinionConfigured = offlineTableConfig.getTaskConfig() != null;
-    boolean minionHeaderPresent = headers.getHeaderString(HttpHeaders.USER_AGENT).contains(CommonConstants.Minion.MINION_HEADER);
+    String userAgentHeader = headers.getHeaderString(HttpHeaders.USER_AGENT);
+    boolean minionHeaderPresent = userAgentHeader.contains(CommonConstants.Minion.MINION_HEADER_PREFIX);
+    String taskType = UserAgentHeader.getTaskType(userAgentHeader);
     boolean metadataRewritten = false;
+
     if (isMinionConfigured) {
```
kishoreg: why have we introduced this block of code based on the existence of minion task?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -426,17 +428,25 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
       // Minion configured use case so that we write to zk sparingly
       String segmentPushStatus = originalOfflineSegmentZkMetadata.getSegmentPushStatus();
       SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd'T'HH:mm:ss.SSS");
 
+      if (taskType.equals(MinionConstants.PurgeTask.TASK_TYPE)) {
```
kishoreg: I am not happy that we need to add checks for minion and task type. Can we come up with a better design?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: If minion is configured in the table config, that means that a race condition between minion and minion uploads or a race condition between hadoop and minion uploads can happen. We want to write to zk as little as possible, and handling the race condition only in minion configured tasks allows us to cut down writes a little since part of handling the race condition is an extra write to zk to write segmentPushStatus.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: I think minion-specific logic can be good(reasons below), but maybe not task-specific logic. However, I couldn't think of a better way to do it. I also think that purge is a specific use case, and we won't necessarily have to do this for other tasks.

Minion has specific logic that keeps creation time the same. Currently we reject segments of creation time the same or less. Additionally, a future looking optimization is that we don't need to necessarily refresh the whole segment in the case of use cases where minion generates segments of the same crc ( we don't want to reject, we still want to change refresh time, but that can be an optimization, for example ).

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
kishoreg: why should that logic be handled or even be specific to minion. Should we not simply have a compare and set api? Minion can read the crc or some unique identifier (zk segment metadata node version) when it downloads and upload api can take this as input and do the comparison.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
kishoreg: I dont see a case where minion would create a segment with same CRC, even in the rare case that it does, its ok to refresh it. 

can you please provide more details on the creation time. why does minion keep the creation time the same. Do we introduce a concept of update time?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
mcvsubbu: We discussed this alternative, and arrived at the conclusion that we don't want to add extra zk writes for use  cases that do not need it. We can talk about this.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: This can happen in the case of purge. We want to change metadata, eg: purge time, when a segment is purged so that we can monitor and schedule appropriately. 

The way minion is designed we decided to keep the creation time the same for all minion tasks. In all the optimization tasks, that is the plan as well. The idea is that minion is working on the "same segment," and we use creation time in order to denote that.

jenniferdai: Subbu and I discussed this offline, and we discussed storing a map of minionHistory, which will contain the task - timestamp. Thoughts? This way, we don't need task specific logic.

mcvsubbu: minion history will be a hashmap of minion-task-name to timestamp. the timestamp will be the last time that particular task touched the segment.

This will mean a change to the existing sgement metadata (not named appropriately) that has been out for about a week, or add a new segment metadatra and render the old one unused.

pinot-minion/src/main/java/com/linkedin/pinot/minion/executor/BaseSegmentConversionExecutor.java
```diff
@@ -113,7 +113,7 @@ public void executeTask(@Nonnull PinotTaskConfig pinotTaskConfig) throws Excepti
       // segment will not be submitted again
       Header ifMatchHeader = new Header(HttpHeaders.IF_MATCH, originalSegmentCrc);
       String minionUserAgentParameter =
-          CommonConstants.Minion.HTTP_TASK_TYPE_HEADER_PREFIX + taskType + "/" + MINION_CONTEXT.getMinionVersion();
+          CommonConstants.Minion.HTTP_TASK_TYPE_HEADER_PREFIX + taskType + CommonConstants.Minion.MINION_HEADER_SEPARATOR + MINION_CONTEXT.getMinionVersion();
```
mcvsubbu: please add this as a method in PinotUserAgentHeader class. Suggest
String constructUserAgentHeader(String taskType) {
}
and you can replace lines 116 through 118 with this method call.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/PinotUserAgentHeader.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.utils;
+
+import org.apache.commons.lang.StringUtils;
+
+
+public class PinotUserAgentHeader {
```
mcvsubbu: Suggest rename to PinotMinionUserAgentHeader

PR #2266, opened 2017-12-21
[MINION] Add PurgeTaskExecutor

Jackie-Jiang:
Added SegmentPurger class to purge segment.
- SegmentPurger takes a RecordPurger and a RecordModifier.
- For each record, we first determine whether it should be purged (removed), if not, then determine whether it should be modified.
Added RecordPurgerFactory and RecordModifierFactory interface that can be injected through MinionContext singleton class.
Added SegmentPurgerTest and PurgeTaskExecutor


pinot-core/src/main/java/com/linkedin/pinot/core/minion/SegmentPurger.java
```diff
@@ -0,0 +1,241 @@
...
+    // Flag to mark whether all records have been iterated
+    boolean _finished = false;
+
+    PurgeRecordReader() throws Exception {
+      _recordReader = new PinotSegmentRecordReader(_originalIndexDir);
+    }
+
+    @Override
+    public boolean hasNext() {
+      if (_recordPurger == null) {
```
mcvsubbu: The iterator logic seems too complex. Will the following work?
hasNext() {
     if (_recordPurger == null) {
        if (_recordReader.hasNext()) {
          _nextRow = _recordReader.next(_nextRow);
          return true;
        } else {
          return false;
        }
      } else {
        while (_recordReader.hasNext()) {
          _nextRow = _recordReader.next(_nextRow);
          if (_recordPurger.shouldPurge(_nextRow)) {
            _numRecordsPurged++;
            continue;
          }
          return true;
        }
        return false;
      }
}

next() {
    if (_recordModifier != null) {
        if (_recordModifier.modifyRecord(_nextRow)) {
          _numRecordsModified++;
        }
      }
      return _nextRow;
}

Jackie-Jiang: For now it might work, but it will break the contract of iterator. If user call next() multiple times, it will return the same record without iterate to the next one. Also, calling next() should get a new object instead of reusing the same one.

Jackie-Jiang: BTW, _finished flag is not mandatory, removed for simpler logic

mcvsubbu: ok, without the finished flag, i am fine with this.

PR #2275, opened 2017-12-22
[TE] frontend - ttbach/minor ui fixes

ttbach:
Fixed error toast width and date columns width


thirdeye/thirdeye-frontend/app/styles/pods/rootcause.scss
```diff
@@ -42,3 +42,9 @@
     margin-left: 8px;
   }
 }
+
+.rootcause-alert {
+  margin-left: $te-spacing-lg;
```
justYves: are you applying the property separately (instead of the shorthand) because of bootstrap's stronger selector?

PR #2279, opened 2017-12-22
[TE] rootcause - extended metric analysis

apucher:
This PR extends the RCA backend to support MetricEntities with a filter tail in the URN. Adds a specialized pipeline for mapping between metric entities that also translates filters if possible. Furthermore extends the one-step dimension analysis pipeline into an iterative component/factor analysis. Also adds a complement of unit tests and helpers.

* long-format metric urns ("thirdeye:metric:12345:filter=value:otherFilter=otherValue")

* support long-format urns in metric-related RCA pipelines

* iterative metric component analysis pipeline (restricted PCA)

* exclusion filters in thirdeye-requests ("filter != value")


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/EntityType.java
```diff
@@ -20,7 +29,32 @@ public EntityType(String prefix) {
...
+   *
+   * @param values parameters, in order
+   * @return formatted urn
+   */
   public String formatURN(Object... values) {
+    if (values.length <= 0) {
+      return this.prefix;
+    }
+
+    if (values[values.length - 1] instanceof Map) {
```
cyenjung: Why we check only the last element? Is this a special use case?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/MetricBreakdownPipeline.java
```diff
@@ -0,0 +1,242 @@
...
+import org.apache.commons.collections.MapUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Pipeline for identifying relevant dimensions by performing
+ * contribution analysis. The pipeline first fetches the Current and Baseline entities and
+ * MetricEntities in the search context. It then maps the metrics to ThirdEye's internal database
+ * and performs contribution analysis.
```
cyenjung: Could you elaborate and JavaDoc the method that you perform in the analysis? I am not sure I am following this comment.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/EntityType.java
apucher: It's an internal utility function for writing tests and pipeline code

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/MetricBreakdownPipeline.java
apucher: thanks for looking at this. I've expanded the javadoc.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/EntityType.java
cyenjung: I see. Could you put that in the comment?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/MetricBreakdownPipeline.java
cyenjung: I see you have added the explanation for DimensionAnalysisPipeline, which is much clearer than before. Could you JavaDoc the difference between this class and DimensionAnalysisPipeline. Thanks.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/EntityType.java
apucher: see javadoc of the method

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/MetricBreakdownPipeline.java
apucher: ok fixed. even more doc

PR #2280, opened 2017-12-23
[TE] self-serve Alert page flow: Update on pull 2228

newsummit:
These are the changes in response to feedback to 2228


newsummit:
Please ignore.

PR #2284, opened 2018-01-02
Add ability for min and max aggregation functions to work on string datatypes

npawar:
In the current implementation, min and max fail on string datatypes. Adding support for this, by changing Max and Min aggregation functions to work on Comparable, instead of Double. 
Tested for :
1) Numeric datatypes and string datatype
2) Dictionary based aggregation operator route and aggregation operator route
3) Single value and multi value columns


pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/DictionaryBasedAggregationOperator.java
```diff
@@ -74,15 +76,27 @@ protected IntermediateResultsBlock getNextBlock() {
           AggregationFunctionFactory.AggregationFunctionType.valueOf(function.getName().toUpperCase());
       String column = aggregationFunctionContext.getAggregationColumns()[0];
       Dictionary dictionary = _dictionaryMap.get(column);
-      AggregationResultHolder resultHolder;
+      FieldSpec.DataType dataType = _dataTypeMap.get(column);
+      AggregationResultHolder resultHolder = AggregationResultHolderFactory.getAggregationResultHolder(functionType, dataType);
       switch (functionType) {
       case MAX:
-        resultHolder = new DoubleAggregationResultHolder(dictionary.getDoubleValue(dictionary.length() - 1));
+        if (dataType.equals(FieldSpec.DataType.STRING)) {
```
mayankshriv: Minor nit: I prefer having the 'if' around reading the value from dictionary, and having resultHolder.setValue() outside of the 'if'.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/DictionaryBasedAggregationOperator.java
```diff
@@ -74,15 +76,27 @@ protected IntermediateResultsBlock getNextBlock() {
...
+          resultHolder.setValue(dictionary.getStringValue(dictionary.length() - 1));
+        } else {
+          resultHolder.setValue(dictionary.getDoubleValue(dictionary.length() - 1));
+        }
         break;
       case MIN:
-        resultHolder = new DoubleAggregationResultHolder(dictionary.getDoubleValue(0));
+        if (dataType.equals(FieldSpec.DataType.STRING)) {
```
mayankshriv: Actually, seeing the pattern, you may want to factor this out as a reusable function.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/AggregationOperator.java
```diff
@@ -36,20 +38,23 @@
   private final TransformExpressionOperator _transformOperator;
   private final long _numTotalRawDocs;
   private ExecutionStatistics _executionStatistics;
+  private final Map<String, FieldSpec.DataType> _dataTypeMap;
 
   public AggregationOperator(@Nonnull AggregationFunctionContext[] aggregationFunctionContexts,
-      @Nonnull TransformExpressionOperator transformOperator, long numTotalRawDocs) {
+      @Nonnull TransformExpressionOperator transformOperator, long numTotalRawDocs,
```
mayankshriv: It might be cleaner (though slightly less efficient) to have the dataType embedded inside aggregationFunctionContexts.

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/AggregationResultHolder.java
```diff
@@ -44,4 +44,17 @@
    * @return
    */
   <T> T getResult();
+
+  /**
+   * Returns default double value
+   * @return
+   */
+  double getDefaultDoubleValue();
```
mayankshriv: Would it be inefficient to just have one the generic version of this api?

pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/MinAggregationFunction.java
```diff
@@ -48,72 +47,158 @@ public void accept(@Nonnull AggregationFunctionVisitorBase visitor) {
...
   public AggregationResultHolder createAggregationResultHolder() {
-    return new DoubleAggregationResultHolder(DEFAULT_VALUE);
+    return new ObjectAggregationResultHolder();
   }
 
   @Nonnull
   @Override
   public GroupByResultHolder createGroupByResultHolder(int initialCapacity, int maxCapacity, int trimSize) {
-    return new DoubleGroupByResultHolder(initialCapacity, maxCapacity, trimSize, DEFAULT_VALUE, true);
+    return new ObjectGroupByResultHolder(initialCapacity, maxCapacity, trimSize);
```
mayankshriv: Should we have a comparable result holder?
Also, for group-by we need to measure any performance issues due to moving from primitive double[] to Object[].

PR #2285, opened 2018-01-02
Adding json for hll

jenniferdai:


pinot-common/src/main/java/com/linkedin/pinot/startree/hll/HllConfig.java
```diff
@@ -126,4 +130,12 @@ public int getHllFieldSize() {
   public String toString() {
     return ToStringBuilder.reflectionToString(this, ToStringStyle.SHORT_PREFIX_STYLE);
   }
+
+  public String toJsonString() throws Exception {
```
mayankshriv: Curious where the callers for these are?

jenniferdai: Adding hll for pbnj

PR #2287, opened 2018-01-03
Remove dateTimeType from dateTimeFieldSpec

npawar:
DateTimeType was added to be able to distinguish between primary and derived datetime types. Knowing which is primary is necessary for retention, and time boundary service. However this information need not be in the schema, and can instead be the responsibility of table config. Hence, removing dateTimeType enum from dateTimeFieldSpec.
We will be extending derived flag to all field specs instead of just dateTimeField.


npawar:
@mcvsubbu Yes they will. I tested Schema.fromFile() and Schema.fromString() methods on schemas which have dateTimeType. They are read successfully, and the output doesn't have dateTimeType, as jsonIgnoreUnknown is true.

PR #2289, opened 2018-01-03
Fixing NPE with regards to user agent header

jenniferdai:


pinot-common/src/main/java/com/linkedin/pinot/common/utils/PinotMinionUserAgentHeader.java
```diff
@@ -18,12 +18,20 @@
...
+import org.slf4j.LoggerFactory;
 
 
 public class PinotMinionUserAgentHeader {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PinotMinionUserAgentHeader.class);
+
   public static String getTaskType(String userAgentHeader) {
-    return StringUtils.substringsBetween(
-        userAgentHeader, CommonConstants.Minion.MINION_HEADER_PREFIX, CommonConstants.Minion.MINION_HEADER_SEPARATOR)[0];
+    try {
```
mcvsubbu: Don't catch NPE. Just handle the case when UserAgentHeader is null.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/PinotMinionUserAgentHeader.java
```diff
@@ -15,15 +15,31 @@
...
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class PinotMinionUserAgentHeader {
-  public static String getTaskType(String userAgentHeader) {
-    return StringUtils.substringsBetween(
-        userAgentHeader, CommonConstants.Minion.MINION_HEADER_PREFIX, CommonConstants.Minion.MINION_HEADER_SEPARATOR)[0];
+  private static final Logger LOGGER = LoggerFactory.getLogger(PinotMinionUserAgentHeader.class);
```
kishoreg: how about having one pojo class (e.g. UploadRequest) that is constructed after parsing the header and query parameters. It should encapsulate all the variables we need in that pojo class. 

Why do have to refer to Minion here? logic here should be generic irrespective of the client. Ignore my comment if that will happen in future PR's

jenniferdai: Agreed. It will be fixed once we redesign. This is for a hotfix because user agent header is not always sent.

PR #2293, opened 2018-01-04
[TE] frontend - ttbach/rca metrics table

ttbach:
Added a table view in addition to a card view for the metrics tab

screen shot 2018-01-05 at 1 47 53 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
@@ -27,6 +27,10 @@ export default Ember.Component.extend({
    */
   isLoading: false,
 
+  /**
+   * Columns for metrics table
+   * @type Object[]
+   */
   metricsTableColumns: [
     {
       template: 'custom/table-checkbox'
@@ -51,6 +55,10 @@ export default Ember.Component.extend({
     this.setProperties({ sortProperty: ROOTCAUSE_METRICS_SORT_PROPERTY_CHANGE, sortMode: ROOTCAUSE_METRICS_SORT_MODE_ASC });
   },
 
+  /**
+   * Data for metrics table
+   * @type Object[] - array of objects, each corresponding to a row in the table
+   */
   metricsTableData: Ember.computed(
     'urns',
     'metrics',
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
@@ -63,7 +63,7 @@ export default Ember.Component.extend({
     'urns',
     'metrics',
     'scores',
-    'changes',
+    'changesFormatted',
     function() {
       let arr = [];
       const { urns, metrics, scores, changesFormatted } = this.getProperties('urns', 'metrics', 'scores', 'changesFormatted');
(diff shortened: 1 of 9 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
```diff
@@ -27,11 +27,57 @@ export default Ember.Component.extend({
...
     this._super(...arguments);
     this.setProperties({ sortProperty: ROOTCAUSE_METRICS_SORT_PROPERTY_CHANGE, sortMode: ROOTCAUSE_METRICS_SORT_MODE_ASC });
   },
 
+  /**
+   * Data for metrics table
+   * @type Object[] - array of objects, each corresponding to a row in the table
+   */
+  metricsTableData: Ember.computed(
+    'urns',
```
cptran777: Are we expecting the value of these properties to change together or can each one be changed individually while the others remain the same? Based on the function, it seems like all these values are updated at once together. If so, it would seem like making a computed property with observers for each of these properties creates unnecessary references. If not, ignore this comment :)

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
```diff
@@ -27,11 +27,57 @@ export default Ember.Component.extend({
...
+    'metrics',
+    'scores',
+    'changes',
+    function() {
+      let arr = [];
+      const { urns, metrics, scores, changesFormatted } = this.getProperties('urns', 'metrics', 'scores', 'changesFormatted');
```
cptran777: You have a listener for 'changes' above but you grab a 'changesFormatted' here. Is that on purpose?

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
ttbach: Yeah, I notice that too. The engineer who wrote the logic for computing those properties are out, so I'll have to ask him for more context. Until then, I figured observing for all these changes is a safer approach. Thanks for the comment.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
ttbach: Nice catch. I meant to listen for 'changesFormatted'. Will make the change.

kishoreg:
can we keep the card view and have a link on the top to switch between the two, not sure how easy hard it is. If its hard, don't worry - we can bring it back later.

Lets format the number.  We need search box only the metric name/dimension. Can we have custom cell component for sparkline. something similar to this http://opensource.addepar.com/ember-table/#/sparkline

ttbach:
@kishoreg 
1. Sure, I can keep both views and have the option to switch between the two. 
2. Which number are you referring to for formatting and in what format?
3. The search box comes with the addon, so I can't remove it for one column and not the others.
4. We can probably make a custom component for the sparkline. What would the sparkline be plotting?

ttbach:
@kishoreg I've updated the screenshots with the suggestions you made (i.e. number formatting + 2 views). I've had another look at addepar, and I realize it uses a different library for the table from the current one used here, and adding a sparkling is not as straight-forward as I initially thought. If you think it's absolutely important to have it, we'll need to have a conversation because it will take a while to implement.

kishoreg:
LGTM

PR #2295, opened 2018-01-05
Add timezone support in dateTimeConvert udf

npawar:
Extending dateTimeFormat to include timezone information along with the sdf pattern. The timezone for an sdf datetime can be defined by adding tz(timezone) at the end of the sdf pattern. The tiemzone can be defined either as long form (America/Los_Angeles), or short form (PST) or in terms of GMT (GMT-0700). If no timezone is provided, UTC will be used as default.

Updated https://github.com/linkedin/pinot/wiki/dateTimeConvert-UDF

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatPatternSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatPatternSpec.java
@@ -1,0 +1,70 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.data;
+
+import java.util.TimeZone;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.joda.time.DateTimeZone;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+
+import static com.linkedin.pinot.common.data.DateTimeFieldSpec.*;
+
+
+public class DateTimeFormatPatternSpec {
+
+  /** eg: yyyyMMdd tz(CST) or yyyyMMdd HH tz(GMT+0700) or yyyyMMddHH tz(America/Chicago) **/
+  private static final Pattern SDF_PATTERN_WITH_TIMEZONE = Pattern.compile("^(.+)( tz[ ]*\\((.+)\\))[ ]*");
+  private static final int SDF_PATTERN_GROUP = 1;
+  private static final int TIMEZONE_GROUP = 3;
+  public static final DateTimeZone DEFAULT_DATETIMEZONE = DateTimeZone.UTC;
+
+  private TimeFormat _timeFormat;
+  private String _sdfPattern = null;
+  private DateTimeZone _dateTimeZone = DEFAULT_DATETIMEZONE;
+  private DateTimeFormatter _dateTimeFormatter ;
+
+  public DateTimeFormatPatternSpec(String timeFormat, String sdfPatternWithTz) {
+    _timeFormat = TimeFormat.valueOf(timeFormat);
+    if (_timeFormat.equals(TimeFormat.SIMPLE_DATE_FORMAT)) {
+      Matcher m = SDF_PATTERN_WITH_TIMEZONE.matcher(sdfPatternWithTz);
+      _sdfPattern = sdfPatternWithTz;
+      if (m.find()) {
+        _sdfPattern = m.group(SDF_PATTERN_GROUP).trim();
+        String timezoneString = m.group(TIMEZONE_GROUP).trim();
+        _dateTimeZone = DateTimeZone.forTimeZone(TimeZone.getTimeZone(timezoneString));
+      }
+      _dateTimeFormatter = DateTimeFormat.forPattern(_sdfPattern).withZone(_dateTimeZone);
+    }
+  }
+
+  public TimeFormat getTimeFormat() {
+    return _timeFormat;
+  }
+
+  public String getSdfPattern() {
+    return _sdfPattern;
+  }
+
+  public DateTimeZone getDateTimeZone() {
+    return _dateTimeZone;
+  }
+
+  public DateTimeFormatter getDateTimeFormatter() {
+    return _dateTimeFormatter;
+  }
+}
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
@@ -16,13 +16,18 @@
 package com.linkedin.pinot.common.data;
 
 import java.util.Objects;
+import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 
+import java.util.regex.Matcher;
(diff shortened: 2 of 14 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
```diff
@@ -16,13 +16,18 @@
...
 import java.util.concurrent.TimeUnit;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.commons.lang3.EnumUtils;
 
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.linkedin.pinot.common.data.DateTimeFieldSpec.TimeFormat;
+import org.joda.time.DateTimeZone;
```
mayankshriv: Why do we need both joda and java TimeZones?

npawar: TimeZone understands long form timezone notation (like America/Los_Angeles, Asia/Kolkata etc), short form (like CST, IST) as well as GMT offset (GMT+0530, GMT-07:00).
DateTimeZone understands only long form timezone notation (like America/Los_Angeles, Asia/Kolkata etc). But DateTimeZone can understand TimeZone.
The DateTimeFormatter we use, requires a DateTimeZone input. Hence we parse timezone using TimeZone, and give that to DateTimeZone, which enables us to support all formats.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
```diff
@@ -218,8 +261,9 @@ public String getSDFPattern() {
       dateTimeColumnValue = unit.convert(dateTimeColumnValueMS, TimeUnit.MILLISECONDS) / size;
     } else {
       String pattern = getSDFPattern();
+      DateTimeZone dateTimeZone = getDateTimezone();
```
kishoreg: we should not invoke this method for every row as it will impact perf. time zone should be extracted once in the constructor and re-used.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
```diff
@@ -39,11 +44,18 @@
       "granularity must be of format size:timeunit";
   public static final String GRANULARITY_PATTERN_ERROR_STR =
       "granularity must be of format [0-9]+:<TimeUnit>";
+
   public static final String NUMBER_REGEX = "[1-9][0-9]*";
+  /** eg: yyyyMMdd tz(CST) or yyyyMMdd HH tz(GMT+0700) or yyyyMMddHH tz(America/Chicago) **/
+  private static final Pattern SDF_PATTERN_WITH_TIMEZONE = Pattern.compile("^(.+)( tz\\((.+)\\))");
```
kishoreg: this is sensitive to spaces right? yymmdd  tz ('UTC') will probably not be able to parse the tz.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
```diff
@@ -169,22 +181,53 @@ public TimeFormat getTimeFormat() {
...
+   * Extracts the timezone from the format of a dateTimeSpec
+   * (4th token in colon separated format in case of TimeFormat=SIMPLE_DATE_FORMAT. tz is added to the end)
+   * This method should not do validation of outputGranularity.
+   * The validation should be handled by caller using {@link #isValidFormat(String)}
+   * eg: if format=1:HOURS:EPOCH, will return UTC
+   * if format=1:HOURS:SIMPLE_DATE_FORMAT:yyyyMMddHH will return UTC
+   * if format=1:HOURS:SIMPLE_DATE_FORMAT:yyyyMMddHH tz(America/Chicago) will return America/Chicago
+   * @return
+   */
+  public DateTimeZone getDateTimezone() {
```
kishoreg: looks like code the duplicated. can we have another class to represent the format and encapsulate the logic in that class and have getters such as getType (SDF/EPOCH), getSDFFormat get TimeZone etc.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
kishoreg: tz[ ]+\\([.*]+\\)[ ]+$ )  might handle spaces gracefully. please validate the regex.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
```diff
@@ -169,22 +181,53 @@ public TimeFormat getTimeFormat() {
...
+    DateTimeZone dateTimeZone = DEFAULT_DATETIMEZONE;
     // split with limit, as pattern can have ':'
     String[] formatTokens = _format.split(COLON_SEPARATOR, MAX_FORMAT_TOKENS);
     if (formatTokens.length == MAX_FORMAT_TOKENS) {
-      pattern = formatTokens[FORMAT_PATTERN_POSITION];
+      Matcher m = SDF_PATTERN_WITH_TIMEZONE.matcher(formatTokens[FORMAT_PATTERN_POSITION]);
+      if (m.find()) {
+        String timezoneString = m.group(TIMEZONE_GROUP);
+        TimeZone timezone = TimeZone.getTimeZone(timezoneString);
```
kishoreg: how do we handle invalid time zones. will this throw an exception?

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
```diff
@@ -169,22 +181,53 @@ public TimeFormat getTimeFormat() {
...
+   */
+  public DateTimeZone getDateTimezone() {
+    DateTimeZone dateTimeZone = DEFAULT_DATETIMEZONE;
     // split with limit, as pattern can have ':'
     String[] formatTokens = _format.split(COLON_SEPARATOR, MAX_FORMAT_TOKENS);
     if (formatTokens.length == MAX_FORMAT_TOKENS) {
-      pattern = formatTokens[FORMAT_PATTERN_POSITION];
+      Matcher m = SDF_PATTERN_WITH_TIMEZONE.matcher(formatTokens[FORMAT_PATTERN_POSITION]);
+      if (m.find()) {
+        String timezoneString = m.group(TIMEZONE_GROUP);
```
kishoreg: might have to trim the time zone string.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
npawar: Added space handling in regex. Also trimming timezone and sdf strings.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatSpec.java
npawar: It will default to UTC by default. This is the behavior of TimeZone

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatPatternSpec.java
```diff
@@ -0,0 +1,70 @@
...
+      }
+      _dateTimeFormatter = DateTimeFormat.forPattern(_sdfPattern).withZone(_dateTimeZone);
+    }
+  }
+
+  public TimeFormat getTimeFormat() {
+    return _timeFormat;
+  }
+
+  public String getSdfPattern() {
```
kishoreg: SDF

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatUnitSpec.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.lang3.EnumUtils;
+
+
+public class DateTimeFormatUnitSpec {
+
+  /**
+   * Time unit enum with range from MILLISECONDS to YEARS
+   */
+  public enum DateTimeTransformUnit {
```
kishoreg: why is this specific to Transform? we can put this in some common place

PR #2304, opened 2018-01-10
[TE] frontend - Added number of events to filter bar

ttbach:
screen shot 2018-01-10 at 11 13 26 am

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
@@ -210,6 +210,28 @@ export default Ember.Component.extend({
   ),
 
   /**
+   * A mapping of event type and number of events of that type
+   * @type {Object}
+   */
+  eventTypeMap: Ember.computed(
+    'entities',
+    function() {
+      const holidayCount = this.countEvents('holiday');
+      const gcnCount = this.countEvents('gcn');
+      const lixCount = this.countEvents('lix');
+      const informedCount = this.countEvents('informed');
+      const anomalyCount = this.countEvents('anomaly');
+      return {
+        holiday: holidayCount,
+        gcn: gcnCount,
+        lix: lixCount,
+        informed: informedCount,
+        anomaly: anomalyCount
+      };
+    }
+  ),
+
+  /**
    * Determines whether to apply filters
    * @param {Object} entity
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -225,6 +247,20 @@ export default Ember.Component.extend({
     });
   },
 
+  /**
+   * Counts the number of events of a given event type
+   * @param eventType - string
+   * @return {Number}
+   */
+  countEvents(eventType) {
```
cptran777: It seems like all these events are updated at once whenever 'entities'[] is updated. If so, this way of counting the entities could be arguably more clear but seems really roundabout and inefficient since you're counting through the same 'entities'[] for each eventType multiple times. Wouldn't it be easier to just count all the events at once in a single function that loops through 'entities'[] once and keeps track of the count for each type in the object you return instead of having to define two functions where one of them just calls the other function repeatedly?

thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -209,6 +209,28 @@ export default Ember.Component.extend({
     }
   ),
 
+  /**
+   * A mapping of event type and number of events of that type
+   * @type {Object}
+   */
+  eventTypeMap: Ember.computed(
```
newsummit: You can use the new `computed` syntax here.

thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
ttbach: Ooh that's a good point. I'll make the change. Thanks!

thirdeye/thirdeye-frontend/app/pods/components/filter-bar/component.js
```diff
@@ -225,6 +247,20 @@ export default Ember.Component.extend({
...
+  /**
+   * Counts the number of events of a given event type
+   * @param eventType - string
+   * @return {Number}
+   */
+  countEvents(eventType) {
+    const entities = this.get('entities');
+    const filteredUrns = Object.keys(entities).filter(urn => {
+      const e = entities[urn];
+      return e.type == 'event' && e.eventType == eventType;
```
newsummit: Just check whether you want `==` or `===` here.

PR #2312, opened 2018-01-11
[TE] Performance Tuning Bug Fixing

cecilynie:
As the pending launch of Alert, Tuning and User Report Page. We observe bugs:
1. Definition of system performance should be based only response data
2. Minimum-Time-To-Detect should be able to handle two-pattern alerts
3. Auto-tuning endpoint needs to pick up current properties from alert filter as the next tuning information. 

Within this pr, we fixed above issues.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
@@ -4,7 +4,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linkedin.thirdeye.anomalydetection.alertFilterAutotune.BaseAlertFilterAutoTune;
 import com.linkedin.thirdeye.detector.email.filter.BaseAlertFilter;
-import com.linkedin.thirdeye.detector.function.AnomalyFunction;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -62,6 +62,8 @@ import com.linkedin.thirdeye.detector.email.filter.AlertFilterFactory;
 import com.linkedin.thirdeye.detector.email.filter.PrecisionRecallEvaluator;
 import com.linkedin.thirdeye.util.SeverityComputationUtil;
 
+import static com.linkedin.thirdeye.datalayer.dto.AnomalyFunctionDTO.*;
+
 
 @Path("/detection-job")
 @Produces(MediaType.APPLICATION_JSON)
@@ -82,6 +84,7 @@ public class DetectionJobResource {
   public static final String AUTOTUNE_PATTERN_KEY = "pattern";
   public static final String AUTOTUNE_MTTD_KEY = "mttd";
   private static final String COMMA_SEPARATOR = ",";
+  private static final String SIGN_TEST_WINDOW_SIZE = "signTestWindowSize";
 
   public DetectionJobResource(DetectionJobScheduler detectionJobScheduler, AlertFilterFactory alertFilterFactory,
       AlertFilterAutotuneFactory alertFilterAutotuneFactory) {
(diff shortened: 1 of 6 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AutotuneConfigDTO.java
```diff
@@ -38,6 +38,12 @@ public void setAlertFilter(AlertFilter alertFilter) {
     this.alertFilter = alertFilter;
   }
 
+  public void setAlertFilter(BaseAlertFilter baseAlertFilter) {
```
cyy0714: The convention for setXXX is to set the value of XXX field. Using setAlertFilter to set both alert filter and tuning properties is not a good idea. I would suggest to name it initAlertFilter or so.

cecilynie: fixed.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -1075,19 +1078,29 @@ public Response getAlertFilterMTTD(@PathParam("functionId") @NotNull long id,
...
     int detectionBucketSize = anomalyFunctionSpec.getBucketSize();
-    double functionMTTDInHour = TimeUnit.HOURS.convert(detectionBucketSize, detectionUnit);
+    int detectionWindowSize = 1;
+    try {
+      Properties functionProps = toProperties(anomalyFunctionSpec.getProperties());
+      detectionWindowSize = Integer.valueOf(functionProps.getProperty(SIGN_TEST_WINDOW_SIZE, "1"));
+    } catch (IOException e) {
+      LOG.warn("Failed to fetch function properties when evaluating mttd!");
+    }
+    double functionMTTDInHour = TimeUnit.HOURS.convert(detectionBucketSize * detectionWindowSize, detectionUnit);
```
cyy0714: For sign test, the detection window is minute, but the window size is 6 hours. The calculation is wrong here.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -1109,10 +1122,30 @@ public Response getProjectedMTTD(@PathParam("autotuneId") @NotNull long id,
...
```
cyy0714: Same errors as above.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
```diff
@@ -1075,19 +1078,29 @@ public Response getAlertFilterMTTD(@PathParam("functionId") @NotNull long id,
...
+      alertFilterMTTDInHour = Math.max(alertFilterMTTDInHour, alertFilterMTTDDOWN);
     }
+
     TimeUnit detectionUnit = anomalyFunctionSpec.getBucketUnit();
     int detectionBucketSize = anomalyFunctionSpec.getBucketSize();
-    double functionMTTDInHour = TimeUnit.HOURS.convert(detectionBucketSize, detectionUnit);
+    int detectionWindowSize = 1;
+    try {
+      Properties functionProps = toProperties(anomalyFunctionSpec.getProperties());
+      detectionWindowSize = Integer.valueOf(functionProps.getProperty(SIGN_TEST_WINDOW_SIZE, "1"));
```
cyy0714: Name it detectionMinimumBuckets may reduce the confusion.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
cecilynie: This is not using "window size" for Sign Test as the function. The mttd is based on minimum detection window, which is based on "signTestWindowSize" as the key from Sign Test.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/DetectionJobResource.java
cecilynie: Not an error as stated above.

PR #2314, opened 2018-01-12
Added a descriptive status on ServiceStatus to get an indication of what it is waiting for.Often happens that a service is not finished starting up yet because of a specific table or partition being not available yet. Since we don't log each time we return a STARTING status, it is good to keep track of what we are waiting for and provide an interface to retrieve it.

mcvsubbu:


mcvsubbu:
In favor of another one

PR #2321, opened 2018-01-14
Minimize window in which two controllers can update metadata for LLC segment commit

mcvsubbu:
It is possible that the leader controller undergoes heavy GC after accepging the
segment commit request from a server. In this case, the leader could lose zk connection
for a while and wait to regain connection in order to test leadership.  During this time
another controller could gain leadership and commit the metadata, with a slightly different
kafka offset, and create a new segment. Meanwhile, the former leader, without knowing that it
has lost leadership, could go ahead and commit another new realtime segment, which could
have a different name than the segment committed by the new leader!

We minimize the window of this happening by checking one more time right before we commit
metadata.

Also, we check for leadership as well as zk connectivity instead of just leadership check.

Added unit tests

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -561,8 +561,9 @@ public class PinotLLCRealtimeSegmentManager {
     File fileToMoveTo = new File(tableDir, segmentName);
 
     if (!isConnected() || !isLeader()) {
-      LOGGER.warn("Lost leadership while committing segment file {}, {} for table {}", segmentName, segmentLocation,
-          tableName);
+      // We can potentially log a different value than what we saw ....
+      LOGGER.warn("Lost leadership while committing segment file {}, {} for table {}: isLeader={}, isConnected={}",
+          segmentName, segmentLocation, tableName, isLeader(), isConnected());
       return false;
     }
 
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -556,6 +560,12 @@ public boolean commitSegmentFile(String tableName, String segmentLocation, Strin
     File tableDir = new File(baseDir, tableName);
     File fileToMoveTo = new File(tableDir, segmentName);
 
+    if (!isConnected() || !isLeader()) {
+      LOGGER.warn("Lost leadership while committing segment file {}, {} for table {}", segmentName, segmentLocation,
```
kishoreg: Log the isleader and isConnected status. Can be helpful while debugging edge cases

mcvsubbu: Will do

mcvsubbu:
@kishoreg , actually we don't need to make sure that the segment name is the same. Instead, @jfim had a proposal that we discussed this morning. The synopsis is as follows:
The update happens as two steps, write to propstore and write to idealstate. We can handle it find if the first one passes and second one fails.
The write to propstore is two znodes - the current completing segment and the new segment. We can write the current completing segment only if the version matches (after modifying the state). THis way, we can avoid creation of two different segments with potentially different offsets (which the naming solution will not solve). We can then add the new segment. I noticed that in helix these are done as two writes anyway, so we will not be adding extra zk writes if we go with this approach.
Also, when writing idealstate, it is already a read-modify-writeIfVersionMatch, so we can check for another segment of the same seq number aleady present, and prevent writes.

These are more extensive changes, I will add them in the near future.

PR #2322, opened 2018-01-16
Add support for long queries for Pinot query console

snleee:
Currently, Pinot query console can only process the query
that is shorter than the maximum length of a GET request.
Changed the query console to use PUT request to remove the
limitation for the query length.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PqlQueryResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PqlQueryResource.java
@@ -38,6 +38,7 @@ import java.util.Random;
 import java.util.Set;
 import javax.inject.Inject;
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Context;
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/resources/webapp/js/init.js
```diff
@@ -105,10 +105,15 @@ var HELPERS = {
   },
 
   executeQuery: function(query, traceEnabled, callback) {
-    var url = "/pql?pql=" + encodeURIComponent(query) +"&trace=" + traceEnabled;
+    var url = "/pql";
+    var params = JSON.stringify({
```
kishoreg: can you please add an option to enable/disable starTree

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PqlQueryResource.java
```diff
@@ -68,44 +88,47 @@ public String get(
...
+    // Retain only online brokers.
+    instanceIds.retainAll(_pinotHelixResourceManager.getOnlineInstanceList());
+    if (instanceIds.isEmpty()) {
+      return QueryException.BROKER_INSTANCE_MISSING_ERROR.toString();
     }
+
+    // Send query to a random broker.
+    String instanceId = instanceIds.get(RANDOM.nextInt(instanceIds.size()));
+    InstanceConfig instanceConfig = _pinotHelixResourceManager.getHelixInstanceConfig(instanceId);
+    String url = "http://" + instanceConfig.getHostName().split("_")[1] + ":" + instanceConfig.getPort() + "/query";
```
kishoreg: this might break if there is an _ in the hostName. Better to replace Broker_ with "" or get everything after first occurrence of _

pinot-controller/src/main/resources/webapp/js/init.js
snleee: I will address starTree with another pr.

PR #2323, opened 2018-01-16
Support HTTPS for file upload/download

Jackie-Jiang:
Merge FileUploadUtils and FileDownloadUtils into FileUploadDownloadClient
Instead of an util class, FileUploadDownloadClient can be initialized by passing in an SSLContext to support HTTPS use case


pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpSegmentFetcher.java
```diff
@@ -30,39 +31,41 @@
...
+  protected int _retryCount;
+  protected int _retryWaitMs;
 
   @Override
   public void init(Configuration configs) {
-    retryCount = configs.getInt(RETRY, RETRY_DEFAULT);
-    retryWaitMs = configs.getInt(RETRY_WAITIME_MS, RETRY_WAITIME_MS_DEFAULT);
+    _retryCount = configs.getInt(RETRY, RETRY_DEFAULT);
+    _retryWaitMs = configs.getInt(RETRY_WAITIME_MS, RETRY_WAITIME_MS_DEFAULT);
+    _httpClient = new FileUploadDownloadClient();
```
mcvsubbu: Can provide a separate method to initialize httpClient, and override it in HttpsSegmentFetcher? You can pass in sslcontext as a null from here, and the real one in https? then you would not have to duplicate the lines to get retrycount and retry wait time in both classes

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
```diff
@@ -272,83 +318,127 @@ public static int uploadSegment(String host, int port, String segmentName, File
...
+    HttpUriRequest request = getDownloadFileRequest(uri, socketTimeoutMs);
+    try (CloseableHttpResponse response = _httpClient.execute(request)) {
+      StatusLine statusLine = response.getStatusLine();
+      int statusCode = statusLine.getStatusCode();
+      if (statusCode >= 400) {
+        throw new HttpErrorStatusException(getErrorMessage(request, response), statusCode);
+      }
+
+      HttpEntity entity = response.getEntity();
+      long contentLength = entity.getContentLength();
```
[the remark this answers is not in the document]
Jackie-Jiang: Fixed.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpSegmentFetcher.java
Jackie-Jiang: Fixed.

PR #2324, opened 2018-01-16
Let RetryPolicy throw exception when number of attempts exhausted

Jackie-Jiang:
Remove the return value from attempt() so it either succeeds or throws exception
Add AttemptsExceededException to indicates number of attempts exhausted

The reason for this change is to prevent using RetryPolicy without checking the return value which might cause silent failure.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
@@ -70,67 +70,67 @@ public class HelixHelper {
    */
   public static void updateIdealState(final HelixManager helixManager, final String resourceName,
       final Function<IdealState, IdealState> updater, RetryPolicy policy) {
-    boolean successful = policy.attempt(new Callable<Boolean>() {
-      @Override
-      public Boolean call() {
-        HelixDataAccessor dataAccessor = helixManager.getHelixDataAccessor();
-        PropertyKey propertyKey = dataAccessor.keyBuilder().idealStates(resourceName);
+    try {
+      policy.attempt(new Callable<Boolean>() {
+        @Override
+        public Boolean call() {
+          HelixDataAccessor dataAccessor = helixManager.getHelixDataAccessor();
+          PropertyKey propertyKey = dataAccessor.keyBuilder().idealStates(resourceName);
 
-        // Create an updated version of the ideal state
-        IdealState idealState = dataAccessor.getProperty(propertyKey);
-        PropertyKey key = dataAccessor.keyBuilder().idealStates(resourceName);
-        String path = key.getPath();
-        // Make a copy of the the idealState above to pass it to the updater, instead of querying again,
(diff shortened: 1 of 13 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
```diff
@@ -70,67 +70,67 @@
...
-          return true;
         }
-      }
-    });
-
-    if (!successful) {
-      throw new RuntimeException("Failed to update ideal state for resource " + resourceName);
+      });
+    } catch (Exception e) {
+      throw new RuntimeException("Caught exception while updating ideal state for resource: " + resourceName, e);
```
mcvsubbu: Correct me if I am wrong, but the earlier piece of code logs the actual exception encountered during attempts, whereas this one will only throw AttemptsExceededException ?

Jackie-Jiang: This one will log the actual exception encountered during attempts, or AttemptsExceededException only if no exception encountered but run out of attempts

pinot-common/src/main/java/com/linkedin/pinot/common/utils/retry/RetryPolicy.java
```diff
@@ -22,12 +22,13 @@
...
+   * Attempts to do the operation until it succeeds, aborting if an exception is thrown by the operation or number of
+   * attempts exhausted.
    *
    * @param operation The operation to attempt, which returns true on success and false on failure.
-   * @return true if the operation succeeded or false if the operation did not succeed within the retries specified by
-   * this retry policy
+   * @throws Exception
    */
-  boolean attempt(Callable<Boolean> operation);
+  void attempt(Callable<Boolean> operation) throws Exception;
```
jfim: This should throw a more specific type of exception than `Exception` (eg. `AttemptsExceededException`, etc.).

pinot-common/src/main/java/com/linkedin/pinot/common/utils/retry/AttemptsExceededException.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.utils.retry;
+
+/**
+ * The <code>AttemptsExceededException</code> indicates that the operation does not succeed within maximum number of
+ * attempts.
+ */
+public class AttemptsExceededException extends Exception {
```
jfim: Should probably have a parent exception class like `AttempedOperationFailureException` or something similar.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/retry/RetryPolicy.java
Jackie-Jiang: This could also throw arbitrary exceptions from the Callable

PR #2331, opened 2018-01-17
Fixed the flaky test for StringPredicateEvaluator

snleee:
Because we join and split with double tab delimiter, the integration test
becomes flaky when the random string generator generates a tab in the beginning
or the end of the string or generates two tabs in a row.

### DIFF
--- a/pinot-core/src/test/java/com/linkedin/pinot/core/predicate/NoDictionaryInPredicateEvaluatorTest.java
+++ b/pinot-core/src/test/java/com/linkedin/pinot/core/predicate/NoDictionaryInPredicateEvaluatorTest.java
@@ -225,7 +225,7 @@ public class NoDictionaryInPredicateEvaluatorTest {
     Set<String> valueSet = new HashSet<>();
 
     for (int i = 0; i < 100; i++) {
-      stringValues[i] = RandomStringUtils.random(MAX_STRING_LENGTH);
+      stringValues[i] = RandomStringUtils.random(MAX_STRING_LENGTH).replace("\t", "");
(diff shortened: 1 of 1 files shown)


pinot-core/src/test/java/com/linkedin/pinot/core/predicate/NoDictionaryInPredicateEvaluatorTest.java
```diff
@@ -225,7 +225,7 @@ public void testStringPredicateEvaluators() {
     Set<String> valueSet = new HashSet<>();
 
     for (int i = 0; i < 100; i++) {
-      stringValues[i] = RandomStringUtils.random(MAX_STRING_LENGTH);
+      stringValues[i] = RandomStringUtils.random(MAX_STRING_LENGTH).replace("\t", "");
```
Jackie-Jiang: Recommend using char ('\t') instead of string

mayankshriv: Does  RandomStringUtils.random return UTF-8? If not, we may want to use a library that returns UTF-8 strings.

snleee: Discussed offline

snleee: Discussed offline

PR #2334, opened 2018-01-18
Ttbach/anomaly baseline info

ttbach:
- Added `anomalyUrns` to the set of `selectedUrns` and `context`
- Added baseline information to the anomaly overview header (i.e. WoW, Wo2W, etc.)

screen shot 2018-01-17 at 5 55 10 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/helpers/color-delta.js
+++ b/thirdeye/thirdeye-frontend/app/helpers/color-delta.js
@@ -8,7 +8,7 @@ import Ember from 'ember';
  */
 export function colorDelta([value = 'N/A']) {
   let sign = true;
-  if (value ===  'N/A') { return; }
+  if (value ===  'N/A') return ;
 
   if (Number.isFinite(value)) {
     sign = parseInt(value * 100) >= 0;
--- a/thirdeye/thirdeye-frontend/app/helpers/color-delta.js
+++ b/thirdeye/thirdeye-frontend/app/helpers/color-delta.js
@@ -9,7 +9,7 @@ import Ember from 'ember';
 export function colorDelta([value = 'N/A']) {
   let sign = true;
   if (value ===  'N/A') { return; }
-  
+
   if (Number.isFinite(value)) {
     sign = parseInt(value * 100) >= 0;
   } else {
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
@@ -1,5 +1,5 @@
 import Component from "@ember/component";
-import { computed, setProperties, getProperties, get } from '@ember/object';
+import { computed, setProperties, getProperties, get, set } from '@ember/object';
 import moment from 'moment';
 import { humanizeFloat, filterPrefix, toOffsetUrn } from 'thirdeye-frontend/helpers/utils';
 
@@ -63,7 +63,10 @@ export default Component.extend({
     }
   ),
 
-  values: computed(
+  /**
+   *
+   */
+  anomalyInfo: computed(
     'aggregates',
     'anomalyUrns',
     function () {
@@ -74,35 +77,22 @@ export default Component.extend({
 
       if (!metricUrns) { return; }
 
-      // NOTE: single metric only
-      const values = {};
-      [...offsets].forEach(offset => {
-        values[offset] = aggregates[toOffsetUrn(metricUrns[0], offset)];
-      });
+      let anomalyInfo = {};
 
-      return values;
-    }
-  ),
-
-  changesFormatted: computed(
-    'aggregates',
-    'anomalyUrns',
-    function () {
-      const { aggregates, anomalyUrns } = getProperties(this, 'aggregates', 'anomalyUrns');
-
-      const offsets = ['current', 'baseline', 'wo1w', 'wo2w', 'wo3w', 'wo4w'];
-      const metricUrns = filterPrefix(anomalyUrns, 'thirdeye:metric:');
-
-      if (!metricUrns) { return; }
-
-      // NOTE: single metric only
-      const changes = {};
       [...offsets].forEach(offset => {
         const change = aggregates[toOffsetUrn(metricUrns[0], 'current')] / aggregates[toOffsetUrn(metricUrns[0], offset)] - 1;
-        changes[offset] = (Math.round(change * 1000) / 10.0).toFixed(1);
+        const roundedChange = (Math.round(change * 1000) / 10.0).toFixed(1);
+        const sign = change > 0 ? '+' : '';
+
+        let offsetFormatted = offset === 'baseline' ? 'Predicted Average' : offset;
+
+        anomalyInfo[offsetFormatted] = {};
+        anomalyInfo[offsetFormatted].change = roundedChange;
+        anomalyInfo[offsetFormatted].value = aggregates[toOffsetUrn(metricUrns[0], offset)];
+        anomalyInfo[offsetFormatted].changeFormatted = `${sign}${roundedChange}`;
       });
 
-      return changes;
+      return anomalyInfo;
     }
   ),
 
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
@@ -1,5 +1,5 @@
 import Component from "@ember/component";
-import { computed, setProperties, getProperties, get, set } from '@ember/object';
+import { computed, setProperties, getProperties, get } from '@ember/object';
 import moment from 'moment';
 import { humanizeFloat, filterPrefix, toOffsetUrn } from 'thirdeye-frontend/helpers/utils';
 
@@ -39,6 +39,10 @@ export default Component.extend({
    */
   optionsMapping: ANOMALY_OPTIONS_MAPPING,
 
+  /**
+   * Urn of an anomaly
+   * @type {String}
+   */
   anomalyUrn: computed(
     'anomalyUrns',
     function () {
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
@@ -1,9 +1,10 @@
 import Component from "@ember/component";
 import { computed, setProperties, getProperties, get } from '@ember/object';
 import moment from 'moment';
-import { humanizeFloat, filterPrefix, toOffsetUrn } from 'thirdeye-frontend/helpers/utils';
+import { humanizeFloat, filterPrefix, toOffsetUrn, humanizeChange } from 'thirdeye-frontend/helpers/utils';
 
 const ROOTCAUSE_HIDDEN_DEFAULT = 'default';
+const OFFSETS = ['current', 'baseline', 'wo1w', 'wo2w', 'wo3w', 'wo4w'];
 
 /**
  * Maps the status from the db to something human readable to display on the form
@@ -27,6 +28,8 @@ export default Component.extend({
   onFeedback: null, // func (urn, feedback, comment)
 
   isHiddenUser: ROOTCAUSE_HIDDEN_DEFAULT,
+
+  offsets: OFFSETS,
 
   /**
    * Array of human readable anomaly options for users to select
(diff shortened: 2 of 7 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -49,6 +67,64 @@ export default Component.extend({
...
+      const offsets = ['current', 'baseline', 'wo1w', 'wo2w', 'wo3w', 'wo4w'];
+      const metricUrns = filterPrefix(anomalyUrns, 'thirdeye:metric:');
+
+      if (!metricUrns) { return; }
+
+      let anomalyInfo = {};
+
+      [...offsets].forEach(offset => {
+        const change = aggregates[toOffsetUrn(metricUrns[0], 'current')] / aggregates[toOffsetUrn(metricUrns[0], offset)] - 1;
+        const roundedChange = (Math.round(change * 1000) / 10.0).toFixed(1);
```
apucher: you could use the "humanizeChange()" util (if it's already merged)

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/template.hbs
```diff
@@ -66,22 +66,18 @@
...
-          </div>
-          <div class="rootcause-anomaly__stats">
-            <label class="rootcause-anomaly__stats-label">Baseline Avg</label>
-            <span class="rootcause-anomaly__stats-value">{{baseline}}</span>
-          </div>
-          <div class="rootcause-anomaly__stats">
-            <label class="rootcause-anomaly__stats-label">Change</label>
-            <span class="rootcause-anomaly__stats-value rootcause-anomaly__stats-value--{{color-delta change}}">{{changeFormatted}}</span>
-          </div>
+          {{#each-in anomalyInfo as |offset value|}}
```
apucher: afaik this does not necessarily iterate in order. you may get 'wo2w', 'current', 'wow4', ...

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -49,6 +67,64 @@ export default Component.extend({
...
+        const sign = change > 0 ? '+' : '';
+
+        let offsetFormatted = '';
+
+        switch(offset) {
+          case 'baseline':
+            offsetFormatted = 'Predicted Average';
```
apucher: this isn't the predicted value. imo it's better to omit than to mislead

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -49,6 +67,64 @@ export default Component.extend({
...
+          case 'current':
+            offsetFormatted = 'Current Average';
+            break;
+          default:
+            offsetFormatted = offset;
+        }
+
+        anomalyInfo[offsetFormatted] = {};
+        anomalyInfo[offsetFormatted].change = roundedChange;
+        anomalyInfo[offsetFormatted].value = aggregates[toOffsetUrn(metricUrns[0], offset)];
```
apucher: you could cache this aggregates[...] value since it's also used above

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -209,11 +211,12 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
...
 
         context = {
-          urns: new Set([...metricUrns, anomalyUrn]),
+          urns: new Set([...metricUrns]),
           anomalyRange,
           analysisRange,
           granularity,
-          compareMode
+          compareMode,
+          anomalyUrns: new Set([...metricUrns, anomalyUrn])
```
apucher: Please add a description to the PR that mentions this change (in addition to describing the header bar). It's got to be readable using the text-only console as well.

thirdeye/thirdeye-frontend/app/helpers/color-delta.js
```diff
@@ -9,7 +9,7 @@ import Ember from 'ember';
 export function colorDelta([value = 'N/A']) {
   let sign = true;
   if (value ===  'N/A') { return; }
```
cptran777: If you're gonna put it on one line anyway, i think `if (value === 'N/A') return` works and looks cleaner.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -1,7 +1,7 @@
 import Component from "@ember/component";
 import { computed, setProperties, getProperties, get } from '@ember/object';
 import moment from 'moment';
-import { humanizeFloat } from 'thirdeye-frontend/helpers/utils';
+import { humanizeFloat, filterPrefix, toOffsetUrn } from 'thirdeye-frontend/helpers/utils';
```
cptran777: is humanizeFloat an IT reference? we'll all float one day...

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -20,7 +20,9 @@ export default Component.extend({
   classNames: ['rootcause-anomaly'],
   entities: null, // {}
 
-  anomalyUrn: null, // ""
+  aggregates: null, // {}
```
cptran777: I'm confused as to what the comment beside this is. Is that a default value? If so, jsdocs has a good format for that
```
/**
 * Your description
 * @type {type}
 * @default value, ex: 'Han shot first' or {}
 */
yourProperty: null,
```

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -37,6 +39,22 @@ export default Component.extend({
...
   optionsMapping: ANOMALY_OPTIONS_MAPPING,
 
+  /**
+   * Urn of an anomaly
+   * @type {String}
+   */
+  anomalyUrn: computed(
+    'anomalyUrns',
+    function () {
+      const { anomalyUrns } = getProperties(this, 'anomalyUrns');
```
cptran777: Are you expecting more values in the future? Seems pointless to use `getProperties` on a single item and then have to parse the resulting object.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -37,6 +39,22 @@ export default Component.extend({
...
+      const anomalyEventUrn = filterPrefix(anomalyUrns, 'thirdeye:event:anomaly:');
+
+      if (!anomalyEventUrn) { return; }
```
cptran777: Same as the above comment about single line if statement

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -49,6 +67,64 @@ export default Component.extend({
     }
   ),
 
+  /**
+   * Information about an anomaly's baselines, changes, and values to be displayed in the anomaly overview
+   * @type {Object}
+   * @example
```
cptran777: I like the example :)

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -49,6 +67,64 @@ export default Component.extend({
...
+            offsetFormatted = 'Predicted Average';
+            break;
+          case 'current':
+            offsetFormatted = 'Current Average';
+            break;
+          default:
+            offsetFormatted = offset;
+        }
+
+        anomalyInfo[offsetFormatted] = {};
```
cptran777: Just curious why you're defining an empty object literal here and then spending the extra lines setting properties on that object as opposed to just seeing
```
anomalyInfo[offsetFormatted] = {
  change: etc...
  value: etc...
}
```
right off the bat. Looks kinda strange/waste of typing/space to me.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -49,6 +67,64 @@ export default Component.extend({
...
+        anomalyInfo[offsetFormatted].value = aggregates[toOffsetUrn(metricUrns[0], offset)];
+        anomalyInfo[offsetFormatted].changeFormatted = `${sign}${roundedChange}`;
+      });
+
+      return anomalyInfo;
+    }
+  ),
+
   functionName: computed('anomaly', function () {
     return get(this, 'anomaly').attributes.function[0];
```
cptran777: I saw above that the `anomaly` property is a computed property that I guess returns an object, but it can also return a boolean as `false`. Will that be an error issue here?

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
ttbach: Will do. I'll still need to calculate the raw change to compute the color of the number in addition to computing the human-readable value that gets returned from `humanizeChange`, just to be clear.

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -104,26 +104,35 @@ export default Ember.Controller.extend({
...
-        const currentUrns = [...metricUrns].map(toCurrentUrn);
-        const baselineUrns = [...metricUrns].map(toBaselineUrn);
+      // aggregates
+      const offsets = ['current', 'baseline', 'wo1w', 'wo2w', 'wo3w', 'wo4w'];
+      const aggregatesUrns = new Set();
 
-        const offsets = ['wo1w', 'wo2w', 'wo3w', 'wo4w'];
-        const offsetUrns = [...metricUrns].map(urn => [].concat(offsets.map(offset => toOffsetUrn(urn, offset)))).reduce((agg, l) => agg.concat(l), []);
+      if (activeTab === ROOTCAUSE_TAB_METRICS) {
+        filterPrefix(Object.keys(entities), 'thirdeye:metric:').forEach(urn => aggregatesUrns.add(urn));
```
justYves: nit:  passing `aggregatesUrns.add` as the callback for `forEach`should work and is more succinct.

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -104,26 +104,35 @@ export default Ember.Controller.extend({
...
+      }
 
-        aggregatesService.request(context, new Set([...currentUrns, ...baselineUrns, ...offsetUrns]));
+      if (!_.isEmpty(context.anomalyUrns)) {
+        filterPrefix(context.anomalyUrns, 'thirdeye:metric:').forEach(urn => aggregatesUrns.add(urn));
```
justYves: same

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -104,26 +104,35 @@ export default Ember.Controller.extend({
...
       }
+
+      const offsetUrns = [...aggregatesUrns].map(urn => [].concat(offsets.map(offset => toOffsetUrn(urn, offset)))).reduce((agg, l) => agg.concat(l), []);
```
justYves: It is not apparent what this is doing, let's have some comments above. Also, let's use more meaningful variable name (not just `l`)

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
ttbach: Thanks for the note. I'll make the change.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
ttbach: 👍

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/template.hbs
ttbach: Good point. I'll have this loop through an array of ordered keys to iterate over the `anomalyInfo` object.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
ttbach: Haha I didn't write this, so I can't claim the credit, but that's a pretty cool reference.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
ttbach: That's a default value for a variable that's expected its value to be passed by the parent component. I'll update the jsdocs for this.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
ttbach: Nope, I'll make the fix.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
ttbach: I tried setting properties on the object, but it threw an error, saying I can't `set property change of undefined`, which meant I had to initialize the object to an empty object before setting its properties.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
ttbach: Good catch. I changed it to `return ;`. So in the anomaly computed property:
`if (!anomalyUrn || !entities || !entities[anomalyUrn]) return ;`

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
justYves: per our slack conversation: 
```
seems like when passed as a callback it loses it's `this` context
`bind` will be needed to make it work, but then, the initial implementation is cleaner
```

PR #2338, opened 2018-01-18
[TE] rootcause poc - prevent newly loaded session from displaying as modified

apucher:


thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-metric-dimension/component.js
```diff
@@ -102,11 +102,13 @@ export default Ember.Component.extend({
...
 
     onSelect() {
-      const { baseUrn, filterMap, onSelection } =
-        this.getProperties('baseUrn', 'filterMap', 'onSelection');
+      const { baseUrn, filterMap, selectedUrn, onSelection } =
+        this.getProperties('baseUrn', 'filterMap', 'selectedUrn', 'onSelection');
 
       const metricUrn = appendFilters(baseUrn, fromFilterMap(filterMap));
 
+      if (_.isEqual(metricUrn, selectedUrn)) { return; }
```
ttbach: Can we include a quick comment on why we're doing this check, just so we remember in the future?

apucher: fixed

PR #2343, opened 2018-01-18
Use dictionaryBasedOperator only if dictionary is immutable, hence guaranteeing that it is sorted

npawar:
In case of realtime consuming segments, the dictionary is unsorted. This results in wrong results for functions which use DictionaryBasedOperator. Introducing an api in Dictionary interface, isSorted(), which will return true if the dictionary is Immutable, and false if it is Mutable


pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/RealtimeClusterIntegrationTest.java
```diff
@@ -109,6 +110,41 @@ public void testGeneratedQueriesWithMultiValues() throws Exception {
...
+   * Adding explicit queries to test dictionary based functions,
+   * to ensure the right result is computed, wherein dictionary is not read if it is mutable
+   * @throws Exception
+   */
+  @Test
+  public void testDictionaryBasedQueries() throws Exception {
+
+    // Dictionary columns
+    // int
+    testDictionaryBasedFunctions("NASDelay");
```
kishoreg: can we not go over the schema and test this for all columns instead of hardcoding the column names

PR #2344, opened 2018-01-19
[TE] frontend - ttbach/add session update

ttbach:
Added information about last user and time anomaly was updated

screen shot 2018-01-18 at 5 16 01 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
@@ -20,12 +20,6 @@ const ROOTCAUSE_SERVICE_BREAKDOWNS = "breakdowns";
 export default Ember.Controller.extend({
   session: Ember.inject.service(),
 
-  /**
-   * User ldap
-   * @type {String}
-   */
-  user: Ember.computed.reads('session.data.authenticated.name'),
-
   queryParams: [
     'metricId',
     'anomalyId',
@@ -384,12 +378,13 @@ export default Ember.Controller.extend({
      */
     onSessionSave() {
       const jsonString = JSON.stringify(this._makeSession());
+      const sessionUpdatedBy = this.get('session.data.authenticated.name'); // fetch ldap of current user
 
       return fetch(`/session/`, { method: 'POST', body: jsonString })
         .then(checkStatus)
         .then(res => this.setProperties({
           sessionId: res,
-          sessionOwner: this.get('user'),
+          sessionUpdatedBy,
           sessionUpdatedTime: moment().format('LLLL'),
           sessionModified: false
         }));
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -231,6 +233,7 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
     // session-initialized context
     if (sessionId) {
       if (!_.isEmpty(session)) {
+        const { name, text, updatedBy, updated } = model.session;
```
apucher: I wonder if this is more confusing than helpful up here

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -17,6 +18,14 @@ const ROOTCAUSE_SERVICE_BREAKDOWNS = "breakdowns";
 // TODO: Update module import to comply by new Ember standards
 
 export default Ember.Controller.extend({
+  session: Ember.inject.service(),
+
+  /**
+   * User ldap
+   * @type {String}
+   */
+  user: Ember.computed.reads('session.data.authenticated.name'),
```
apucher: I think you can safely read this directly in onSessionSave()

justYves: I believe using onSessionSave would only work for just saved report. I think she's trying to handle the case where it's already saved and we're reading the info.

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -254,6 +260,8 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
       sessionId,
       sessionName,
       sessionText,
+      sessionOwner,
```
apucher: maybe rename this to "sessionUpdatedBy". The session-owner is the user that originally saved it, which may or may not be the last person to update it.

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
ttbach: @justYves I think what @apucher means is `user` can be retrieved like `this.get('session.data.authenticated.name` inside the `onSessionSave` method, as opposed to making a CP. I'll make the change.

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
ttbach: Gotcha. That makes sense. Will make the change.

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
ttbach: I think it's more concise than having to write `model.session.name`, `model.session.text`, etc.

PR #2345, opened 2018-01-19
[TE] rootcause poc - support special characters in urns

apucher:
People are horrible, so we need support for special characters in urns (with filter tail). Depending on the dataset, filters may take values that are misinterpreted as URN tokens, e.g. "pageId=my:app:/route/subroute?q=text". This PR adds frontend and backend support for URI-encoding URN fragments so that urns can be tokenized safely.


thirdeye/thirdeye-frontend/app/helpers/utils.js
```diff
@@ -155,7 +155,7 @@ export function appendTail(urn, tail) {
 }
 
 export function appendFilters(urn, filters) {
-  const tail = filters.map(t => `${t[0]}=${t[1]}`);
+  const tail = filters.map(t => encodeURIComponent(`${t[0]}=${t[1]}`));
```
justYves: nit: Let's add some JS doc for those utils functions when we get a chance

apucher: yes, i'll push a separate PR for that

PR #2347, opened 2018-01-19
[TE] rootcause poc - prevent stale data load

apucher:
reduce database load by not loading aggregates for metrics that are about to be evicted from the frontend entities cache

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/rootcause-entities-cache/service.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rootcause-entities-cache/service.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-import { checkStatus, filterObject, filterPrefix, toBaselineRange, toFilters } from 'thirdeye-frontend/helpers/utils';
+import { checkStatus, filterObject, filterPrefix, toBaselineRange, toFilters, toColor } from 'thirdeye-frontend/helpers/utils';
 import fetch from 'fetch';
 import _ from 'lodash';
 
@@ -12,7 +12,7 @@ export default Ember.Service.extend({
 
   pending: null, // Set
 
-  errors: null, // Set({ frameowrk, error })
+  errors: null, // Set({ framework, error })
 
   init() {
     this._super(...arguments);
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/rootcause-entities-cache/service.js
```diff
@@ -12,7 +12,7 @@ export default Ember.Service.extend({
 
   pending: null, // Set
 
-  errors: null, // Set({ frameowrk, error })
+  errors: null, // Set({ framework, error })
```
justYves: nit: let's document using the JS doc format. http://usejsdoc.org/
This should be straightforward. @ttbach  and I are available to discuss this as well.

thirdeye/thirdeye-frontend/app/pods/rootcause-entities-cache/service.js
```diff
@@ -43,15 +43,16 @@ export default Ember.Service.extend({
 
     // rootcause search
     if (!_.isEqual(context, requestContext)) {
+      const newEntities = filterObject(entities, (e) => urns.has(e.urn));
```
justYves: nit: let's start using better variable names than single letters

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -140,36 +146,13 @@ export default Ember.Controller.extend({
   // Public properties (computed)
   //
 
-  entities: Ember.computed(
-    'entitiesService.entities',
-    function () {
-      const entities = _.cloneDeep(this.get('entitiesService.entities'));
+  entities: Ember.computed.reads('entitiesService.entities'),
```
justYves: 👍  Looks so much cleaner!

thirdeye/thirdeye-frontend/app/pods/rootcause-entities-cache/service.js
ttbach: 👍. I think as we do more work on this, we should start being more diligent about documentation, so it's easier to read and also to refactor later. It would be unideal if we got to a size that's unmanageable, and we have little clue what's going on just from glancing at it.

apucher: Yeah I agree it's time to start documenting this. The RCA core UI has become fairly stable.

PR #2355, opened 2018-01-20
[TE] rootcause poc - warn on non-additive dataset in heatmap

apucher:
* display warning for non-additive datasets on heatmap view

* prevent warnings from being displayed while still loading data

* fix potential daylight-savings-time (DST) bug in time range offset utils

image

### DIFF
--- a/thirdeye/thirdeye-frontend/app/helpers/utils.js
+++ b/thirdeye/thirdeye-frontend/app/helpers/utils.js
@@ -401,7 +401,7 @@ export function fromFilterMap(filterMap) {
   Object.keys(filterMap).forEach(key => {
     [...filterMap[key]].forEach(value => {
       filters.push([key, value]);
-    })
+    });
   });
   return filters;
 }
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/helpers/utils.js
```diff
@@ -488,6 +488,21 @@ export function isInverse(urn, entities) {
...
 
+/**
+ * Extracts information about whether a metric is tagged as additive or not.
+ *
+ * @param {string} urn metric urn
+ * @param {Object} entities entities cache
+ * @returns {boolean} if metric is tagged as additive
+ */
+export function isAdditive(urn, entities) {
+  if (!entities) { return false; }
```
justYves: perhaps a try catch would be more succinct here. I feel it's fine to use for deeply nested objects where a null can pop up at any level.
```
try {
  return (entities[urn].attributes.additive[0] === 'true')
} catch (error) {
  return false; 
}
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Control_flow_and_error_handling#try...catch_statement

apucher: thanks. fixed

PR #2361, opened 2018-01-21
Fix duplicate LLC segments possible when lead controller goes through a huge GC

jackjlli:
1. check whether the status of the old metadata has changed.
2. use zookeeper version number to prevent duplicate write.

Unit tests have been added.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -88,6 +88,7 @@ import org.apache.helix.NotificationContext;
 import org.apache.helix.ZNRecord;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.store.zk.ZkHelixPropertyStore;
+import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTest.java
```diff
@@ -749,6 +750,79 @@ public void testCompleteCommittingSegments() throws Exception {
...
+    FakePinotLLCRealtimeSegmentManagerII(boolean setupInitialSegments, List<String> existingLLCSegments, int scenario) {
+      super(setupInitialSegments, existingLLCSegments);
+      _scenario = scenario;
+    }
+
+    @Override
+    public LLCRealtimeSegmentZKMetadata getRealtimeSegmentZKMetadata(String realtimeTableName, String segmentName, Stat stat) {
+      LLCRealtimeSegmentZKMetadata metadata = super.getRealtimeSegmentZKMetadata(realtimeTableName, segmentName, stat);
+      switch (_scenario) {
+        case 1:
```
mcvsubbu: Use the final int that you have defined.

kishoreg:
can you check why the test coverage came down?

jackjlli:
@kishoreg My code doesn't change the code coverage a lot. 
It seems the reason why CC drops from 70% to 69.3% is this PR: https://github.com/linkedin/pinot/pull/2327. 
Once `having clause` implementation is complete, CC will be back.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -605,11 +622,19 @@ public boolean commitSegmentMetadata(String rawTableName, final String committin
...
 
+    Stat stat = new Stat();
     final LLCRealtimeSegmentZKMetadata oldSegMetadata = getRealtimeSegmentZKMetadata(realtimeTableName,
-        committingSegmentNameStr);
+        committingSegmentNameStr, stat);
     final LLCSegmentName oldSegmentName = new LLCSegmentName(committingSegmentNameStr);
     final int partitionId = oldSegmentName.getPartitionId();
     final int oldSeqNum = oldSegmentName.getSequenceNumber();
+
+    if (oldSegMetadata.getStatus() == CommonConstants.Segment.Realtime.Status.DONE) {
```
mcvsubbu: can u please change the comparison to be NOT IN_PROGRESS?  I feel more comfortable checking for the exact state we should be in, before we continue.

PR #2362, opened 2018-01-21
Read all certificates from certification authority file.

mcvsubbu:
We were reading only the first certificate from the certification authority file, which resulted in
some servers not getting recognized by the trust store. We need to read all certificates present
and populate the trust store.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
@@ -21,7 +21,6 @@ import com.linkedin.pinot.common.utils.FileUploadDownloadClient;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
-import java.math.BigInteger;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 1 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
```diff
@@ -21,6 +21,7 @@
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.math.BigInteger;
```
Jackie-Jiang: ？

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
```diff
@@ -103,13 +104,18 @@ protected void initHttpClient(Configuration configs) {
...
-      CertificateFactory certificateFactory = CertificateFactory.getInstance(CERTIFICATE_TYPE);
-      X509Certificate cert = (X509Certificate) certificateFactory.generateCertificate(is);
-
-      String serverKey = "https-server";
       KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
       trustStore.load(null);
-      trustStore.setCertificateEntry(serverKey, cert);
+      int i = 0;
+      while (is.available() > 0) {
+        CertificateFactory certificateFactory = CertificateFactory.getInstance(CERTIFICATE_TYPE);
```
Jackie-Jiang: Please move the factory creation out of the loop

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HttpsSegmentFetcher.java
mcvsubbu: Used to format the result in hex.

PR #2365, opened 2018-01-22
[TE] rootcause poc - expand anomaly header by default (unless it has feedback)

apucher:


thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -198,6 +199,10 @@ export default Component.extend({
     return moment(get(this, 'anomaly').end).format('MMM D YYYY, hh:mm a');
   }),
 
+  requiresFeedback: computed('status', function () {
```
justYves: There's a more succinct way to do this: 
```
import { equal } from '@ember/object/computed';
requiresFeedback: equal('status', 'NO_FEEDBACK')
```

apucher:
@ttbach the header used to expand automatically when no feedback is provided. that functionality was broken in that previous PR

PR #2366, opened 2018-01-22
[TE] rootcause poc - controller cleanup and jsdoc

apucher:


thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -10,6 +10,7 @@ const ROOTCAUSE_TAB_DIMENSIONS = "dimensions";
 const ROOTCAUSE_TAB_METRICS = "metrics";
 const ROOTCAUSE_TAB_EVENTS = "events";
 
+const ROOTCAUSE_SERVICE_ROUTE = "route";
```
justYves: Please use single quotes `'` for string in JS. (This is a convention we adopted) 
Applies to all prs.

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -34,6 +33,8 @@ export default Ember.Controller.extend({
   //
   // services
   //
+  authService: Ember.inject.service('session'),
```
justYves: any reason why we changed the name here? (curious)

apucher: Naming collision. 'session' currently refers to 'RCA session' in the route and controller - this has to be refactored in the future

apucher: fixed

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
apucher: fixed

PR #2368, opened 2018-01-23
[TE] frontend - Added links to rca metrics

ttbach:
- Added links to each metric in the metrics table
- Create a custom cell for the `metrics-table`


thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
```diff
@@ -192,6 +192,43 @@ export default Ember.Component.extend({
...
+   * {
+   *  thirdeye:metric:12345: [],
+   *  thirdeye:metric:23456: [
+   *    {urlLabel: url},
+   *    {urlLabel: url}
+   *  ]
+   * }
+   */
+  links: Ember.computed(
+    'urns',
```
apucher: add 'entities'

ttbach: `urns` already listens to `entities`, so I figured it would be a duplicate listener if I add `entities`. Thoughts?

thirdeye/thirdeye-frontend/app/pods/custom/rca-metric-links/template.hbs
```diff
@@ -0,0 +1,9 @@
+{{#if record.links.length}}
+  {{#each record.links as |link|}}
+    {{#each-in link as |linkLabel url|}}
+      <a href="{{url}}" target="_blank">{{linkLabel}}</a>
```
justYves: \+ `class="thirdeye-link"`  for consistency

thirdeye/thirdeye-frontend/app/pods/components/metrics-table/component.js
```diff
@@ -13,40 +13,46 @@ export default Component.extend({
     }, {
       propertyName: 'label',
       title: 'Metric',
-      className: 'rootcause-metric__table__column'
+      className: 'rootcause-metric__table__column rootcause-metric__table__links-column--large'
```
cptran777: So many column configs... so much writing. It almost sounds like you could use a util function to build out these columns! <Insert shameless self-plug here> :D
In all seriousness, not a big deal but if you end up having multiple tables and want to have more readable configs, callisto and cmon-web both use artdeco table config builders

PR #2375, opened 2018-01-23
[TE] rootcause poc - time-based session modification check

apucher:
Adding a basic time-based check for session modification in the background by another user. It polls in large intervals (5min) and displays a warning if modification is detected. This isn't bullet-proof yet and users can still override existing sessions with stale data.

image

### DIFF
--- a/thirdeye/thirdeye-frontend/app/helpers/utils.js
+++ b/thirdeye/thirdeye-frontend/app/helpers/utils.js
@@ -46,6 +46,8 @@ export const eventColorMapping = {
   anomaly: 'teal'
 };
 
+export const dateFormatFull = 'ddd, MMM D YYYY, h:mm a';
+
 /**
  * The Promise returned from fetch() won't reject on HTTP error status even if the response is an HTTP 404 or 500.
  * This helps us define a custom response handler.
@@ -594,5 +596,6 @@ export default Ember.Helper.helper({
   eventColorMapping,
   parseProps,
   postProps,
-  toIso
+  toIso,
+  dateFormatFull
 });
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-header/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-header/component.js
@@ -1,36 +1,104 @@
 import Component from '@ember/component';
-import { getProperties, computed } from '@ember/object';
+import { get, getProperties, computed } from '@ember/object';
+import { dateFormatFull } from 'thirdeye-frontend/helpers/utils';
 
 export default Component.extend({
   classNames: ['rootcause-header'],
-  sessionId: null, // 0
-
-  sessionName: null, // ""
-
-  sessionText: null, // ""
-
-  sessionModified: null, // true
-
-  onSave: null, // func(sessionName, sessionText)
-
-  onCopy: null, // func(sessionName, sessionText)
 
   /**
-   * Toggles the comment textarea
+   * Rootcause session identifier
+   * @type {Int}
    */
-  showComment: computed.bool('sessionText'),
+  sessionId: null,
+
+  /**
+   * Rootcause session title
+   * @type {string}
+   */
+  sessionName: null,
+
+  /**
+   * Rootcause session comments
+   * @Type {string}
+   */
+  sessionText: null,
+
+  /**
+   * Rootcause session dirty flag
+   * @type {boolean}
+   */
+  sessionModified: null,
+
+  /**
+   * Rootcause session last updated timestamp
+   * @type {Int}
+   */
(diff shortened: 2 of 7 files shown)


thirdeye/thirdeye-frontend/app/helpers/utils.js
```diff
@@ -46,6 +46,8 @@ export const eventColorMapping = {
   anomaly: 'teal'
 };
 
+export const dateFormatFull = 'ddd, MMM D YYYY, h:mm a';
```
justYves: let's create a `constant.js` file to store all constants

thirdeye/thirdeye-frontend/app/pods/components/rootcause-header/component.js
```diff
@@ -1,36 +1,104 @@
...
+  sessionUpdatedTimeFormatted: computed('sessionUpdatedTime', function () {
+    return moment(get(this, 'sessionUpdatedTime')).format(dateFormatFull);
+  }),
 
   actions: {
+    /**
+     * "Save" button action
+     */
     onSave() {
       const { onSave } = getProperties(this, 'onSave');
```
justYves: fyi, if `onSave` is a passed function. we can just to 
`this.attrs.onSave()`

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -363,6 +369,46 @@ export default Ember.Controller.extend({
...
+   * Fetches the current session state from the backend and issues a notification if it has been updated
+   *
+   * @private
+   */
+  _checkSession() {
+    const { sessionId, sessionUpdatedTime } = this.getProperties('sessionId', 'sessionUpdatedTime');
+
+    if (!sessionId) { return; }
+
+    fetch(`/session/${sessionId}`)
```
justYves: can we try keeping all fetching logic out of the controllers? (Thank you for your effort in abstracting them to services btw!)

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -363,6 +369,46 @@ export default Ember.Controller.extend({
...
+      .then(checkStatus)
+      .then(res => {
```
justYves: nit: let's use parens `(res)` when the arrow function spans over multiple lines (our convention)

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -3,7 +3,7 @@ import RSVP from 'rsvp';
 import fetch from 'fetch';
 import moment from 'moment';
 import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
-import { toCurrentUrn, toBaselineUrn, filterPrefix, appendFilters, toFilters, checkStatus } from 'thirdeye-frontend/helpers/utils';
+import { toCurrentUrn, toBaselineUrn, filterPrefix, appendFilters, toFilters, checkStatus, dateFormatFull } from 'thirdeye-frontend/helpers/utils';
```
justYves: (just a thought) : this utils file is getting really big, if feel that we should break it into smaller files.

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -165,7 +163,7 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
     };
 
     let selectedUrns = new Set();
-    let sessionName = 'New Investigation (' + moment().format(dateFormat) + ')';
+    let sessionName = 'New Investigation (' + moment().format(dateFormatFull) + ')';
```
justYves: this looks like a UI only logic (copy), perhaps this should be handled in the template?

apucher: I agree it's copy, but it also affects what's stored in the backend. If you come from an anomaly you get a different default than coming from a metric or a stored session.

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
apucher: yes, something for a separate PR

thirdeye/thirdeye-frontend/app/helpers/utils.js
apucher: yes, something for a separate PR

thirdeye/thirdeye-frontend/app/pods/components/rootcause-header/component.js
apucher: thx

PR #2386, opened 2018-01-24
[TE] Onboarding jobs

cyy0714:
During onboarding, exceptions are hidden in system logs. Although users are able to fetch job status through api call, it is not a good user experience to expose exceptions to FE. This pr is to send out email notification upon exceptions, along with some bugs fixed. 
Note that, the wrapper function for onboarding is resumed for a short period of time, until the whole onboarding job migrates to the new design.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/onboard/tasks/FunctionCreationOnboardingTask.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/onboard/tasks/FunctionCreationOnboardingTask.java
@@ -249,21 +249,21 @@ public class FunctionCreationOnboardingTask extends BaseDetectionOnboardTask {
         anomalyFunctionSpec.setCron("0 0 0 * * ?");
         anomalyFunctionSpec.setWindowSize(6);
         anomalyFunctionSpec.setWindowUnit(TimeUnit.HOURS);
-        anomalyFunctionSpec.setProperties("signTestWindowSize=24;anomalyRemovalThreshold=0.6;baseline=w/3wAvg;decayRate=0.5;signTestStepSize=1;slidingWindowWidth=8;confidenceLevel=0.99");
+        anomalyFunctionSpec.setProperties("");
         break;
       case HOURS:
         anomalyFunctionSpec.setType("REGRESSION_GAUSSIAN_SCAN");
         anomalyFunctionSpec.setCron("0 0 14 1/1 * ? *");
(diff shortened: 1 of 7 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/onboard/tasks/FunctionCreationOnboardingTask.java
```diff
@@ -238,24 +249,28 @@ protected AnomalyFunctionDTO getDefaultFunctionSpecByTimeGranularity(TimeGranula
         anomalyFunctionSpec.setCron("0 0 0 * * ?");
         anomalyFunctionSpec.setWindowSize(6);
         anomalyFunctionSpec.setWindowUnit(TimeUnit.HOURS);
+        anomalyFunctionSpec.setProperties("signTestWindowSize=24;anomalyRemovalThreshold=0.6;baseline=w/3wAvg;decayRate=0.5;signTestStepSize=1;slidingWindowWidth=8;confidenceLevel=0.99");
```
cecilynie: Hi Yung-Yu, after a second think, these properties should have their own default values when using function. e.g. anomalyRemovalThreshold has defined in the class "AnomalyRemovalByWeight" in AD library. I recall that you have fully tested to trigger these anomaly function without any properties and they can still properly run. Here can we utilize the default values in AD library without writing explicitly these property values?

cyy0714: Yes. you are right. I also checked my code, these configurations are inserted in the init(). We should look into what happened to the exceptions during replay.

cyy0714: I just found out the code is still lying there for review.

cecilynie: Thanks Yung-Yu. Since we are having the default setup in AD library, how about within this pr only merge the wrapper code for deploy? Thanks!

cyy0714: Thanks Kexin for the comment. The rb has been pushed. I removed the default properties here.

PR #2387, opened 2018-01-24
Refactor code to avoid duplicate retry on downloading segment to local

jackjlli:
This PR removes the outer retry wrapper code. That's because:

1. `downloadSegmentToLocal` method has already used retry policy to fetch segment to local M times (M = 3 by default). If we keep the outer retry wrapper, which retries at most N times(N = 5 by default), the total retry times would be M*N.

2. if an exception is thrown when untarring(right after the method `downloadSegmentToLocal` is called), that means the tar file is broken or not found AFTER the retry. There's no need to do it again. Simply throw the exception would be good.


pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
```diff
@@ -246,7 +246,7 @@ public ServerType getServerType() {
     public static final long DEFAULT_QUERY_EXECUTOR_TIMEOUT_MS = 15_000L;
     public static final String DEFAULT_REQUEST_HANDLER_FACTORY_CLASS =
         "com.linkedin.pinot.server.request.SimpleRequestHandlerFactory";
-    public static final String DEFAULT_SEGMENT_LOAD_MAX_RETRY_COUNT = "5";
+    public static final String DEFAULT_SEGMENT_LOAD_MAX_RETRY_COUNT = String.valueOf(SegmentFetcher.RETRY_DEFAULT);
```
Jackie-Jiang: Don't use the value from SegmentFetcher, it's separate config. Recommend change the type to int

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
```diff
@@ -61,7 +62,9 @@ public SegmentFetcherAndLoader(DataManager dataManager, ZkHelixPropertyStore<ZNR
     } catch (Exception e) {
       // Keep the default value
     }
-    _segmentLoadMaxRetryCount = maxRetries;
+    if (maxRetries != Integer.parseInt(CommonConstants.Server.DEFAULT_SEGMENT_LOAD_MAX_RETRY_COUNT)) {
```
Jackie-Jiang: int maxRetries = pinotHelixProperties.getInt(CommonConstants.Server.CONFIG_OF_SEGMENT_LOAD_MAX_RETRY_COUNT, CommonConstants.Server.DEFAULT_SEGMENT_LOAD_MAX_RETRY_COUNT);

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
```diff
@@ -71,15 +74,18 @@ public SegmentFetcherAndLoader(DataManager dataManager, ZkHelixPropertyStore<ZNR
     } catch (Exception e) {
       // Keep the default value
     }
-    _segmentLoadMinRetryDelayMs = minRetryDelayMillis;
+    if (minRetryDelayMillis != Integer.parseInt(CommonConstants.Server.DEFAULT_SEGMENT_LOAD_MIN_RETRY_DELAY_MILLIS)) {
```
Jackie-Jiang: Same here after changing the data type of CommonConstants.Server.DEFAULT_SEGMENT_LOAD_MIN_RETRY_DELAY_MILLIS to long

jackjlli:
@Jackie-Jiang RetryPolicy has already been used in `fetchSegmentToLocal` method. Thus, there's no need to reuse it in this part.

PR #2391, opened 2018-01-25
[TE] Self-Serve Onboarding Flow: handling transition to alert page after creation

newsummit:
This takes care of the transition from alert creation to alert page, after the replay has been triggered. Here we are still using the `/notifyreplaytuning` API to trigger replay. In our next iteration we will be replacing this with the new task handler API `/detection-onboard/create-job`. For now, we are transitioning to the `manage/alert/{id}` route after a 2 min interval.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
@@ -579,64 +579,6 @@ export default Controller.extend({
   },
 
   /**
-   * Replay Flow Step 2 - Replay cloned function
-   * @method callReplayStart
-   * @param {Number} clonedId - id of the cloned function
-   * @param {Object} startTime - replay start time stamp
-   * @param {Object} endTime - replay end time stamp
-   * @return {Ember.RSVP.Promise}
-   */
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
```diff
@@ -1284,44 +1272,26 @@ export default Controller.extend({
...
-          // Redirect to onboarding page to trigger wrapper
-          if (redirectToAlertPage) {
-            this.transitionToRoute('manage.alert', this.get('newFuncId'), { queryParams: { replayId }});
-          }
+            // Display function added to group confirmation
+            this.prepareFunctions(finalConfigObj, newFunctionId).then(functionData => {
+              this.set('selectedGroupFunctions', functionData);
+            });
+            // Now, disable form
+            this.setProperties({
```
justYves: should `this.setProperties` and the `this.send` below happen after the this.preparefunctions's chained promises? if yes, the current implementation doesn't guarantee that

newsummit: Please disregard - I've removed everything except `this.send`.

PR #2397, opened 2018-01-25
Ttbach/refactored utils file to extract rca-related utils

ttbach:
Spring Cleaning Commences:
- Created a new util file `rca-utils` inside `app/helpers` and moved all rca-related util functions there.
- Updated all import references of those functions
- Removed unnecessary imports


ttbach:
@justYves Great suggestion! This should be a quick fix.

PR #2398, opened 2018-01-25
[TE] rootcause - return NaN for aggregate of non-existing data (i.e. future)

apucher:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/aggregation/DefaultAggregationLoader.java
```diff
@@ -91,6 +91,11 @@ public double loadAggregate(MetricSlice slice) throws Exception {
     ThirdEyeResponse res = this.cache.getQueryResult(rc.getRequest());
     DataFrame df = DataFrameUtils.evaluateResponse(res, rc);
 
+    final long maxTime = this.cache.getDataSource(dataset.getDataSource()).getMaxDataTime(dataset.getDataset());
+    if (slice.getStart() > maxTime) {
```
cyy0714: Could you explain 1) why the start time of the slice greater than max time, and 2) why we fill in NaN here?

apucher: "if the the date range we want to fetch is past the latest time stamp available in the database (i.e. in the future), return NaN". NaN here is an equivalent for null, but still a valid Double value.

cyy0714: I see. Thanks for the explanation.

PR #2399, opened 2018-01-25
[TE] rootcause - prevent external null links

apucher:


apucher:
I don't like links that go nowhere. You don't either. :-)

PR #2400, opened 2018-01-25
Enabling a table also resets errored segments

jfim:
Reset segments in a table that are in error state whenever a table is
enabled. This makes it possible to reset errors without restarting
servers.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -82,6 +82,7 @@ import org.apache.helix.ClusterMessagingService;
 import org.apache.helix.Criteria;
 import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixDataAccessor;
+import org.apache.helix.HelixException;
 import org.apache.helix.HelixManager;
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1419,7 +1420,19 @@ public PinotResourceManagerResponse toggleTableState(String tableName, boolean s
...
-    return (status) ? new PinotResourceManagerResponse("Table " + tableName + " successfully enabled.", true)
+
+    // If enabling a resource, also reset segments in error state for that resource
+    boolean resetSuccessful = false;
+    if (status) {
+      try {
+        _helixAdmin.resetResource(_helixClusterName, Collections.singletonList(tableName));
+        resetSuccessful = true;
+      } catch (HelixException e) {
+        LOGGER.warn("Caught exception while resetting resource {}, ignoring.", e, tableName);
```
Jackie-Jiang: Recommend logging an error

jfim: Warn is typically for things that did not work right but otherwise did not prevent the completion of the operation. Error is for things that completely prevented an operation.

In this case, the enabling went through, although it might be in a partial state.

PR #2401, opened 2018-01-25
Enhance the exception handling for segment upload

Jackie-Jiang:
1. Add check for whether ZK update succeeded
2. Remove the return value for PinotHelixResourceManager.addNewSegment() and refreshSegment(), throw exception for failure
3. Fix the issue where ZK might not be updated when there are exceptions thrown out
4. Only count not handled exception as INTERNAL_SERVER_ERROR and update metrics only for that

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
@@ -38,7 +38,6 @@ import com.linkedin.pinot.controller.api.access.AccessControl;
 import com.linkedin.pinot.controller.api.access.AccessControlFactory;
 import com.linkedin.pinot.controller.helix.core.PinotHelixResourceManager;
 import com.linkedin.pinot.controller.helix.core.PinotHelixSegmentOnlineOfflineStateModelGenerator;
-import com.linkedin.pinot.controller.helix.core.PinotResourceManagerResponse;
 import com.linkedin.pinot.controller.util.TableSizeReader;
 import com.linkedin.pinot.controller.validation.StorageQuotaChecker;
 import com.linkedin.pinot.core.segment.index.SegmentMetadataImpl;
@@ -52,7 +51,6 @@ import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.InetAddress;
 import java.net.URLDecoder;
-import java.text.ParseException;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -271,9 +269,7 @@ public class PinotSegmentUploadRestletResource {
       FileUploadPathProvider provider = new FileUploadPathProvider(_controllerConf);
       String tempSegmentName = "tmp-" + System.nanoTime();
       tempTarredSegmentFile = new File(provider.getFileUploadTmpDir(), tempSegmentName);
-      tempTarredSegmentFile.deleteOnExit();
       tempSegmentDir = new File(provider.getTmpUntarredPath(), tempSegmentName);
-      tempSegmentDir.deleteOnExit();
 
       // Get upload type
       String uploadTypeStr = headers.getHeaderString(FileUploadDownloadClient.CustomHeaders.UPLOAD_TYPE);
@@ -285,7 +281,6 @@ public class PinotSegmentUploadRestletResource {
       }
 
       String downloadURI = null;
-      boolean found = false;
       switch (uploadType) {
         case JSON:
         case URI:
@@ -293,11 +288,8 @@ public class PinotSegmentUploadRestletResource {
           try {
             downloadURI = getDownloadUri(uploadType, headers, segmentJsonStr);
           } catch (Exception e) {
-            String errorMsg =
-                String.format("Failed to get download Uri for upload file type: %s, with error %s", uploadType,
-                    e.getMessage());
-            _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_SEGMENT_UPLOAD_ERROR, 1L);
-            throw new ControllerApplicationException(LOGGER, errorMsg, Response.Status.BAD_REQUEST);
+            throw new ControllerApplicationException(LOGGER, "Failed to get download URI", Response.Status.BAD_REQUEST,
+                e);
           }
 
           // Get segment fetcher based on the download URI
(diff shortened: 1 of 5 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -445,42 +421,47 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
         expectedCrc = Long.parseLong(expectedCrcStr);
       } catch (NumberFormatException e) {
         throw new ControllerApplicationException(LOGGER,
-            "Caught exception while parsing IF-MATCH CRC: " + expectedCrcStr, Response.Status.PRECONDITION_FAILED, e);
+            "Caught exception for segment: " + segmentName + " of table: " + offlineTableName
```
jenniferdai: do we also want to include e?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -498,31 +479,26 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
 
-      PinotResourceManagerResponse response;
+      // Update ZK metadata and refresh the segment if necessary
       long newCrc = Long.valueOf(segmentMetadata.getCrc());
-      if (newCrc != existingCrc) {
-        if (downloadUrl == null) {
-          downloadUrl = moveSegmentToPermanentDirectory(provider, tableName, segmentName, tempTarredSegmentFile);
+      if (newCrc == existingCrc) {
+        // New segment is the same as the existing one, only update ZK metadata without refresh the segment
+        if (!_pinotHelixResourceManager.updateZkMetadata(existingSegmentZKMetadata)) {
```
jenniferdai: we can return here if this succeeds

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -498,31 +479,26 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
...
+          downloadUrl = moveSegmentToPermanentDirectory(provider, rawTableName, segmentName, tempTarredSegmentFile);
+        }
+        _pinotHelixResourceManager.refreshSegment(segmentMetadata, existingSegmentZKMetadata, downloadUrl);
       }
-      return response;
-    } finally {
-      if (enableParallelPushProtection && !metadataRewritten) {
-        _pinotHelixResourceManager.updateZkMetadata(existingSegmentZKMetadata);
+    } catch (Exception e) {
+      if (!_pinotHelixResourceManager.updateZkMetadata(existingSegmentZKMetadata)) {
```
jenniferdai: i could be missing something, but why do we not need to reset segment upload time upon exceptions thrown?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: Not necessary because from the message we already know the exception type

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: This is already the last statement lol

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: We need to, and this statement is resetting the upload time (see line 464)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -445,42 +421,47 @@ private PinotResourceManagerResponse uploadSegment(File indexDir, SegmentMetadat
         expectedCrc = Long.parseLong(expectedCrcStr);
       } catch (NumberFormatException e) {
         throw new ControllerApplicationException(LOGGER,
-            "Caught exception while parsing IF-MATCH CRC: " + expectedCrcStr, Response.Status.PRECONDITION_FAILED, e);
+            "Caught exception for segment: " + segmentName + " of table: " + offlineTableName
+                + " while parsing IF-MATCH CRC: " + expectedCrcStr, Response.Status.PRECONDITION_FAILED);
```
mcvsubbu: Nit.
If expectedCrc is sent as a space character or somehow 0 length string, then we won't be able to see it in the log. Given that it has not parsed, it may be best to include the value of expectedCrcStr within single quotes

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1656,19 +1619,12 @@ private void sendSegmentRefreshMessage(OfflineSegmentZKMetadata segmentZKMetadat
...
-      throws JSONException, IOException {
-    final TableConfig offlineTableConfig =
-        ZKMetadataProvider.getOfflineTableConfig(_propertyStore, segmentMetadata.getTableName());
-
-    final String segmentName = segmentMetadata.getName();
+  private void addNewOfflineSegment(final SegmentMetadata segmentMetadata) {
     final String offlineTableName = TableNameBuilder.OFFLINE.tableNameWithType(segmentMetadata.getTableName());
+    final String segmentName = segmentMetadata.getName();
+    final TableConfig offlineTableConfig = ZKMetadataProvider.getOfflineTableConfig(_propertyStore, offlineTableName);
+    Preconditions.checkNotNull(offlineTableConfig);
```
mcvsubbu: Should this not be a 400 table not found? Will precondition here end up sending back a 500?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: Good point, added

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: We already checked that in PinotSegmentUploadRestletResource.uploadSegment()

PR #2402, opened 2018-01-25
Ttbach/create utils folder

ttbach:
- Create `app/utils` folder to move existing js utils functions there


ttbach:
@justYves Yes, I made sure the flows for rca and manage alerts work.

PR #2403, opened 2018-01-25
Add broker meter to track result's group by size

jackjlli:
This PR adds broker meter to keep track of group by size from the result.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/reduce/BrokerReduceService.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/reduce/BrokerReduceService.java
@@ -203,6 +203,12 @@ public class BrokerReduceService implements ReduceService<BrokerResponseNative> 
           setGroupByHavingResults(brokerResponseNative, aggregationFunctions, aggregationFunctionSelectStatus,
               brokerRequest.getGroupBy(), dataTableMap, brokerRequest.getHavingFilterQuery(),
               brokerRequest.getHavingFilterSubQueryMap());
+          if (brokerMetrics != null && (!brokerResponseNative.getAggregationResults().isEmpty())) {
+            // We emit the group by size when the result isn't empty. All the sizes among group by results should be the same.
+            // Thus, we can just choose the size of 1st group by result.
+            brokerMetrics.addMeteredTableValue(rawTableName, BrokerMeter.GROUP_BY_SIZE,
+                brokerResponseNative.getAggregationResults().get(0).getGroupByResult().size());
+          }
         }
       }
     }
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -203,6 +203,13 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
           setGroupByHavingResults(brokerResponseNative, aggregationFunctions, aggregationFunctionSelectStatus,
               brokerRequest.getGroupBy(), dataTableMap, brokerRequest.getHavingFilterQuery(),
               brokerRequest.getHavingFilterSubQueryMap());
+          if (brokerMetrics != null) {
+            long groupBySize = 0L;
+            for (AggregationResult aggregationResult : brokerResponseNative.getAggregationResults()) {
+              groupBySize = Math.max(aggregationResult.getGroupByResult().size(), groupBySize);
+            }
+            brokerMetrics.addMeteredTableValue(rawTableName, BrokerMeter.GROUP_BY_SIZE, groupBySize);
```
snleee: We should not emit the metric if the query does not perform `group by`. Current implementation seems to emit 0 for that case.

Jackie-Jiang: It's inside the branch of group-by query, so should be good

pinot-core/src/main/java/com/linkedin/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -203,6 +203,13 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
           setGroupByHavingResults(brokerResponseNative, aggregationFunctions, aggregationFunctionSelectStatus,
               brokerRequest.getGroupBy(), dataTableMap, brokerRequest.getHavingFilterQuery(),
               brokerRequest.getHavingFilterSubQueryMap());
+          if (brokerMetrics != null) {
+            long groupBySize = 0L;
```
Jackie-Jiang: Should be int right?

pinot-core/src/main/java/com/linkedin/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -203,6 +203,13 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
...
+            for (AggregationResult aggregationResult : brokerResponseNative.getAggregationResults()) {
+              groupBySize = Math.max(aggregationResult.getGroupByResult().size(), groupBySize);
```
Jackie-Jiang: Can you explain why using the maximum group-by size from servers as the metric value?

pinot-core/src/main/java/com/linkedin/pinot/core/query/reduce/BrokerReduceService.java
jackjlli: That's because the type of parameter `unitCount` is long according to the method signature, i.e. `public void addMeteredTableValue(final String tableName, final M meter, final long unitCount)`

PR #2404, opened 2018-01-26
Re-factor PinotLLCRealtimeSegmentManager methods to reduce the number of arguments.

mcvsubbu:
The argument count seems to increase each time we want something new. Passing around
tableConfig and kafkaStreamMetadata seems to obviate the need for most other arguments.
In another step, maybe we can also get rid of KafkaStreamMetadata, since that should also
be (technically) got from tableconfig.

There is still a dependency with the kafka consumer, so we still need to be careful in tests
ensuring that we don't start to initiate a connect to Kafka. At some point, we should
pass in a mock kafka consumer, so that we can can get the offsets we want from the
consumer. We are only interested in offsets for these tests.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -210,39 +210,26 @@ public class PinotLLCRealtimeSegmentManager {
    * The topic name is being used as a dummy helix resource name. We do not read or write to zk in this
    * method.
    */
-  public void setupHelixEntries(final String topicName, final String realtimeTableName, int nPartitions,
-      final List<String> instanceNames, int nReplicas, String initialOffset, String bootstrapHosts, IdealState
-      idealState, boolean create, int flushSize) {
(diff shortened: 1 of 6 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1249,19 +1232,42 @@ public void updateKafkaPartitionsIfNecessary(String realtimeTableName, TableConf
      }
    }
    */
-  private ZNRecord generatePartitionAssignment(String topicName, int nPartitions, List<String> instanceNames,
-      int nReplicas) {
+  private ZNRecord generatePartitionAssignment(TableConfig tableConfig, String topicName, int nPartitions, List<String> instanceNames,
+      int nReplicasNU) {
```
npawar: What is the difference between nReplicasNU parameter and nReplicas on line 1237? They look like they're the same? Do we need both?

mcvsubbu: No, we can remove the NU. The NU was meant for not used. Let me take a look

PR #2407, opened 2018-01-26
[TE] rootcause - forward to latest session for anomaly id

apucher:


thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -551,12 +555,16 @@ export default Ember.Controller.extend({
...
       return sessionService
         .saveAsync(session)
-        .then((res) => this.setProperties({
-          sessionId: res,
-          sessionUpdatedBy,
-          sessionUpdatedTime: moment().valueOf(),
-          sessionModified: false
-        }))
+        .then((res) => {
+          const sessionId = res;
```
justYves: Looks like duplicate code, let's make it DRY-er  by moving this into a function.

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -84,6 +41,7 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
 
     if (anomalyUrn) {
       anomalyContext = fetch(`/rootcause/raw?framework=anomalyContext&urns=${anomalyUrn}`).then(checkStatus).catch(res => undefined);
+      anomalySessions = fetch(`/session/query?anomalyId=${anomalyId}`).then(checkStatus).catch(res => undefined);
```
justYves: This will save a promise into `anomalySessions` is that what we want?

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -93,54 +51,57 @@ export default Ember.Route.extend(AuthenticatedRouteMixin, {
...
+
+      Object.assign(newModel, {
+        anomalyId: null,
+        anomalyUrn: null,
+        anomalyContext: null,
+        sessionId: mostRecent.id,
+        session: mostRecent
+      });
+
+      // NOTE: apparently this does not abort the ongoing transition
```
justYves: `transition.abort()` is what you need? 

https://guides.emberjs.com/v2.18.0/routing/preventing-and-retrying-transitions/

apucher: this seems to be an undocumented quirk. transitionTo() is supposed to abort to ongoing transition, however, it apparently doesn't when only the queryParams change.

thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
apucher: yes, they all become part of the RSVP hash

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
apucher: fixed

PR #2413, opened 2018-01-30
Pass memory used by segment to the controller on segment completion, for using it to tune end criteria in future

npawar:


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -176,7 +176,8 @@ public String segmentCommitEnd(
       @QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE_ID) String instanceId,
       @QueryParam(SegmentCompletionProtocol.PARAM_SEGMENT_NAME) String segmentName,
       @QueryParam(SegmentCompletionProtocol.PARAM_SEGMENT_LOCATION) String segmentLocation,
-      @QueryParam(SegmentCompletionProtocol.PARAM_OFFSET) long offset
+      @QueryParam(SegmentCompletionProtocol.PARAM_OFFSET) long offset,
+      @QueryParam(SegmentCompletionProtocol.PARAM_MEMORY_USED) long memoryUsed
```
mcvsubbu: I don't remember what this param is set to if the QueryParam is not present in the request. Have you verified that it does not throw exception? or set to some weird value?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -599,9 +599,11 @@ private static File convertURIToSegmentLocation(String segmentLocation) {
    * @param rawTableName Raw table name
    * @param committingSegmentNameStr Committing segment name
    * @param nextOffset The offset with which the next segment should start.
+   * @param memoryUsed The memory used by committing segment
```
mcvsubbu: pls add comments here as to what the value will be if the client did not set it at all.

pinot-common/src/main/java/com/linkedin/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -235,6 +238,11 @@ public Params withSegmentLocation(String segmentLocation) {
         return this;
       }
 
+      public Params withMemoryUsed(long memoryUsed) {
+        _memoryUsed = memoryUsed;
```
kishoreg: any validation needed here?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
kishoreg: Also, add the units. I am assuming this is in Bytes.

mcvsubbu: +1, very good point.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
npawar: It is set to 0. It does not throw an exception. 
Added a comment to document this for when we start using it

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
npawar: Added units. Also added comments about behaviour when this is not set by client

PR #2414, opened 2018-01-31
[TE] email - prototype for embedding png resource into email

apucher:
!DO NOT MERGE


justYves:
@apucher Looks like this can be closed?

PR #2415, opened 2018-01-31
[TE] rootcause - fix metric spinner stuck on non-existing metric

apucher:


thirdeye/thirdeye-frontend/app/pods/rootcause-aggregates-cache/service.js
```diff
@@ -76,6 +76,8 @@ export default Ember.Service.extend({
   _extractAggregates(incoming, urn) {
     // NOTE: only supports single time range
     const aggregates = {};
+    aggregates[urn] = parseFloat('NaN'); // default
```
justYves: can't we just set it to NaN ? 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN

justYves: actually Number.NaN seems more correct https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/NaN

apucher: fixed

PR #2417, opened 2018-01-31
[TE] working redirection from legacy to ember app

justYves:
### What's new: 
- Fix the redirection after login. Now, the user should be redirected to the original pre-login url.

### Tested Redirection: 
- rootcause
- root cause + anomalyId (no saved session)
- root cause + anomalyId (saved session)
- roocause + saved session Id
- root cause + metric Id 
- /rca
- /rca + query params
- manage


kishoreg:
Thank a lot for fixing this

PR #2421, opened 2018-02-01
Working realtime table config

jenniferdai:
* Editing sample realtime table config to use low level kafka consumer


jenniferdai:
@snleee yes, it is

PR #2423, opened 2018-02-01
[TE] frontend - Translate granularity dropdown options to human-readable format

ttbach:
- Granularity dropdown values are now grammatically correct
- Default compare by mode is changed to "split"

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart-toolbar/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart-toolbar/component.js
@@ -1,5 +1,18 @@
 import Component from '@ember/component';
 import moment from 'moment';
+import { computed, get, set, getProperties, setProperties } from '@ember/object';
+
+/**
+ * Mapping between translated human-readable granularity values and values that are named on the backend
+ * @type {Object}
+ */
+const granularityMapping = {
+  '5 Minutes': '5_MINUTES',
+  '15 Minutes': '15_MINUTES',
+  '1 Hour': '1_HOURS',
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart-toolbar/component.js
```diff
@@ -42,12 +65,31 @@ export default Component.extend({
...
+    let selected = '';
+
+    // Find the key in granularityMapping whose value matches the granularity
+    keys.some(key => {
+      const value = this.granularityMapping[key];
+      if (value === get(this, 'granularity')) {
+        selected = key;
+      }
+    });
+
```
ttbach: @justYves Yeah I was itching about that. Should I extract the translation logic here into a util?

apucher: @ttbach imo the "correct" way would be to make this a two-sided map

ttbach: I agree. The only thing is there's no bi-directional map in javascript 😞

igbopie: how about this?
`
const granularityValue = get(this, 'granularity');
const selected = keys.find(key => granularityValue === this.granularityMapping[key])
`

ttbach: Ooh i like that. Thanks for the suggestion! @igbopie.

PR #2425, opened 2018-02-01
Create a command to check the cluster state

jackjlli:
This PR creates a program to check the cluster state, i.e. check whether all the tables are stable in the cluster. Once it's pushed to the master branch, we can start our test, which is we randomly pick some nodes and restart them, and we use this command to check the cluster state.

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
@@ -1,0 +1,125 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.tools;
+
+import com.linkedin.pinot.common.config.TableNameBuilder;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class ClusterStateVerifier extends PinotZKChanger {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ClusterStateVerifier.class);
+
+  public ClusterStateVerifier(String zkAddress, String clusterName) {
+    super(zkAddress, clusterName);
+  }
+
+  /**
+   * return true if all the tables are stable
+   * @param tableNames list of table names which are about to scan
+   * @param timeoutSec maximum timeout in second
+   * @return
+   */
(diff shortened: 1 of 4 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class ClusterStateVerifier {
+  private static final Logger LOGGER = LoggerFactory.getLogger(ClusterStateVerifier.class);
+  private PinotSegmentRebalancer _rebalancer;
```
mcvsubbu: Let us not have this class depend on rebalancer. In fact, rebalance is now a controller command so you can remove the PinotSegmentRebalancer class altogether.

jackjlli: That's a good idea! But let's separate these two tasks. I'll open another ticket to remove `PinotSegmentRebalancer` class.

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -0,0 +1,146 @@
...
+      isStable = isStable(tableName);
+      if (isStable != 0) {
+        LOGGER.error("Table {} is not stable. The cluster is not stable.", tableName);
+        break;
+      }
+    }
+    return isStable;
+  }
+
+  private int isStable(String tableName) {
```
mcvsubbu: You can return a boolean from here. The idea was to exit the command with an exit status of 1 (non-zero) if the cluster is not stable, and 0 if the cluster is. Not sure if our Command hiearchy allows return codes, or do we have to thrown an exception in order to exit with a non-zero code (in Linux)

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -0,0 +1,163 @@
...
+
+  public ClusterStateVerifier(String zkAddress, String clusterName) {
+    _zkAddress = zkAddress;
+    _clusterName = clusterName;
+    _helixResourceManager =  new PinotHelixResourceManager(getControllerConf());
+    _helixResourceManager.start();
+  }
+
+  private ControllerConf getControllerConf() {
+    ControllerConf conf = new ControllerConf();
```
mcvsubbu: shouldnt need to start a controller. Can you inherit from PinotZkChanger?

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  /**
+   * return true if all the tables are stable
+   * @param tableNames list of table names which are about to scan
+   * @return
+   */
+  private boolean isClusterStable(final List<String> tableNames) {
+    LOGGER.info("Start scanning the stability of all the tables... Num of tables: {}", tableNames.size());
+    int isStable = -1;
```
mcvsubbu: initialized to 0, because the cluster is stable if there are no tables in it.
Also, rename isStable as unstablePartitionCount

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -0,0 +1,125 @@
...
+    for (String tableName : tableNames) {
+      isStable = super.isStable(tableName);
+      if (isStable != 0) {
+        LOGGER.error("Attention: Table {} is not stable. The cluster is not stable.", tableName);
```
mcvsubbu: also log the number of partitions that are not stable

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -0,0 +1,125 @@
...
+        break;
+      }
+    }
+    LOGGER.info("Finish scanning.");
```
mcvsubbu: good to log the number of tables scanned.

PR #2426, opened 2018-02-01
[TE] Add email recipients when an alert group exists

cyy0714:
In current pipeline, the new assigned email recipients are missed when updating an existing alerts. This pr is to fix this bug.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/onboard/tasks/FunctionCreationOnboardingTask.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/onboard/tasks/FunctionCreationOnboardingTask.java
@@ -220,6 +220,15 @@ public class FunctionCreationOnboardingTask extends BaseDetectionOnboardTask {
       alertConfig = alertConfigDAO.findById(configuration.getLong(ALERT_ID));
       EmailConfig emailConfig = alertConfig.getEmailConfig();
       emailConfig.getFunctionIds().add(anomalyFunction.getId());
+
+      // Add recipients to existing alert group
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/onboard/tasks/FunctionCreationOnboardingTask.java
```diff
@@ -220,6 +220,15 @@ public void run() {
       alertConfig = alertConfigDAO.findById(configuration.getLong(ALERT_ID));
       EmailConfig emailConfig = alertConfig.getEmailConfig();
       emailConfig.getFunctionIds().add(anomalyFunction.getId());
+
+      // Add recipients to existing alert group
+      String recipients = configuration.getString(ALERT_TO);
+      if (StringUtils.isNotBlank(recipients)) {
+        if (StringUtils.isNotBlank(alertConfig.getRecipients())) {
+          recipients = recipients + "," + alertConfig.getRecipients();
```
cecilynie: Hi Yung-Yu, thanks for fixing this. do we check if the input recipients already exists in current recipient list?

cyy0714: That's a good point. Currently, it is no harm to add duplicate emails. But to check if the email is redundant, should we also check if the email format is valid?

cyy0714: It is done. Please review the change.

PR #2427, opened 2018-02-01
Removing old jersey jars (1.9) that were pulled by hadoop-common tran…

kishoreg:
…sitively

Hadoop common and Hadoop client was pulling in jersey 1.9 versions, whereas rest of the code depended on 2.3 version. Controller console application would fail to handle requests with an error message - Grizzly request failed. 

Added explicit exclusions in the pom to avoid this error.


kishoreg:
will wait for the test to pass and then merge

PR #2428, opened 2018-02-01
Partition assignment support in multi tenant setup

npawar:
Changing the generate partition assignment logic, to account for multi tenant setup. Introduced a StreamConsumptionConfig, which will provide the partition assignment strategy for the table. The strategies available right now are:
1. UniformStreamPartitionAssignment - uniform spraying partitions of the table across available instances. 
2. BalancedStreamPartitionAssignment - using AutoRebalanceStrategy to evenly distribute partitions of all tables on the tenant across the available instances.


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
```diff
@@ -211,7 +217,7 @@ public static void buildLowLevelRealtimeIdealStateFor(String realtimeTableName,
     final PinotLLCRealtimeSegmentManager segmentManager = PinotLLCRealtimeSegmentManager.getInstance();
     final int nPartitions = getPartitionCount(kafkaMetadata);
     LOGGER.info("Assigning {} partitions to instances for simple consumer for table {}", nPartitions, realtimeTableName);
-    segmentManager.setupHelixEntries(realtimeTableConfig, kafkaMetadata, nPartitions, realtimeInstances, idealState, create);
+    segmentManager.setupHelixEntries(realtimeTableConfig, kafkaMetadata, realtimeInstances, idealState, create);
```
mcvsubbu: We get the number of partitions for this topic from Kafka and pass it down to setupHelixEntries. If setupHelixEntries() doesn't have the number of partitions for this topic, how will it assign partitions to hosts? :-)

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1199,20 +1232,31 @@ public void updateKafkaPartitionsIfNecessary(TableConfig tableConfig) {
...
+          numOfInstancesNeeded);
       return;
     } else {
       _controllerMetrics.setValueOfTableGauge(realtimeTableName, ControllerGauge.SHORT_OF_LIVE_INSTANCES, 0);
     }
-    ZNRecord newPartitionAssignment = generatePartitionAssignment(tableConfig, kafkaStreamMetadata.getKafkaTopicName(), currentPartitionCount, currentInstances);
-    writeKafkaPartitionAssignment(realtimeTableName, newPartitionAssignment);
+
+    Map<String, ZNRecord> newPartitionAssignment = generatePartitionAssignment(tableConfig, currentInstances);
+    writeKafkaPartitionAssignment(newPartitionAssignment);
```
mcvsubbu: What happens when one kafka partition change is detected in the master controller, and vlauidation manager is updating a bunch of znodes, and a table gets added in another controller, and it tries to update the same set of znodes? What happens if a controller fails after updating some znodes and not others?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.helix.ZNRecord;
+import org.apache.helix.controller.rebalancer.strategy.AutoRebalanceStrategy;
+
+
+/**
+ * Generates partition assignment for all tables provided, using AutoRebalanceStrategy
+ */
+public class AutoRebalancePartitionAssignmentGenerator extends PartitionAssignmentGenerator {
+
+  public AutoRebalancePartitionAssignmentGenerator(TableConfig tableConfig, int nKafkaPartitions,
```
kishoreg: nKafkaPartitions -> numPartitions. this can be applied to any streaming source that supports partitioning right?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+      List<String> instanceNames, Map<String, TableConfig> allTableConfigsInTenant,
+      Map<String, ZNRecord> currentPartitionAssignment) {
+
+    super(tableConfig, nKafkaPartitions, instanceNames, currentPartitionAssignment);
+
+    _tablesForPartitionAssignment = new ArrayList<>();
+    for (TableConfig config : allTableConfigsInTenant.values()) {
+      RoutingConfig routingConfig = config.getRoutingConfig();
+      if (routingConfig == null || routingConfig.getRoutingTableBuilderName() == null ||
+          !RoutingTableBuilderName.PartitionAwareRealtime.toString().equals(routingConfig.getRoutingTableBuilderName())) {
```
kishoreg: why are we hardcoding the routing builder name?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+
+      // construct current partitions map
+      ZNRecord partitionAssignmentZNode = _currentPartitionAssignment.get(realtimeTableName);
+      if (partitionAssignmentZNode != null) {
+
+        Map<String, List<String>> partitionToInstances = partitionAssignmentZNode.getListFields();
+        for (Map.Entry<String, List<String>> partitionEntry : partitionToInstances.entrySet()) {
+          String partitionNumber = partitionEntry.getKey();
+          List<String> partitionInstances = partitionEntry.getValue();
+          String key = realtimeTableName + partitionJoiner + partitionNumber;
```
kishoreg: create a class to represent RealtimePartition which will encapsulate the name generation logic. Current Assignment could be Map<RealtimePartition, List<RealtimeInstance>>

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+      // construct new partitions list
+      int p = _nPartitions.get(realtimeTableName);
+      for (int i = 0; i < p; i++) {
+        newPartitions.add(realtimeTableName + partitionJoiner + Integer.toString(i));
+      }
+    }
+
+    // get states
+    LinkedHashMap<String, Integer> states = new LinkedHashMap<>(2);
+    states.put("OFFLINE", 0);
```
kishoreg: constants for states, they are probably defined some where else

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+    AutoRebalanceStrategy autoRebalanceStrategy =
+        new AutoRebalanceStrategy(_tablesForPartitionAssignment.get(0), newPartitions, states);
+    ZNRecord partitionAssignmentAcrossAllTables =
+        autoRebalanceStrategy.computePartitionAssignment(_instanceNames, _instanceNames, currentPartitions, null);
+
+    // extract kafka partition assignment znodes from auto rebalance result
+    Map<String, List<String>> listFields = partitionAssignmentAcrossAllTables.getListFields();
+    for (Map.Entry<String, List<String>> entry : listFields.entrySet()) {
+      String partitionName = entry.getKey();
+      int lastIndex = partitionName.lastIndexOf(partitionJoiner);
```
kishoreg: we need concrete classes to represent realtimepartition.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.helix.ZNRecord;
+
+
+/**
+ * Base class for partition assignment generation across tables of a tenant
+ */
+public abstract class PartitionAssignmentGenerator {
```
kishoreg: Good job creating a base abstraction. Is this only for realtime or applicable for offline as well.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PartitionAwarePartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+
+
+/**
+ * Generates partition assignment for all partition aware routing tables
+ * The partition assignment is done by uniformly spraying the partitions across available instances,
+ * such that the assignment can be used for a partition aware routing
+ *
+ * NOTE: We do not support/expect partition aware tables with multi tenant setup, hence this strategy should suffice
+ */
+public class PartitionAwarePartitionAssignmentGenerator extends PartitionAssignmentGenerator {
```
kishoreg: this name is confusing

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1231,45 +1250,70 @@ public void updateKafkaPartitionsIfNecessary(TableConfig tableConfig) {
...
-    return znRecord;
+    return realtimeTablesWithServerTenant;
+  }
+
+  /**
+   * Checks if given table has partition aware routing config set
+   * @param tableConfig
+   * @return
+   */
+  public boolean isPartitionAwareRealtimeRouting(TableConfig tableConfig) {
```
kishoreg: this logic seems to be duplicated elsewhere. Instead of checking explicitly for the name, its better to have boolean on the RoutingConfig.isPartitionAware right? 
Also why do we care if its partitionAware or not

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
npawar: nKafkaPartitions is named this way so that we can distinguish between nPartitions from kafka vs the one from partition assignment in znrecord

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+
+
+/**
+ * Generates partition assignment for all tables provided, using AutoRebalanceStrategy
+ */
+public class AutoRebalancePartitionAssignmentGenerator extends PartitionAssignmentGenerator {
+
+  public AutoRebalancePartitionAssignmentGenerator(TableConfig tableConfig, int nKafkaPartitions,
+      List<String> instanceNames, Map<String, TableConfig> allTableConfigsInTenant,
+      Map<String, ZNRecord> currentPartitionAssignment) {
```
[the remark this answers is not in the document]
npawar: avoid using in the method or completely not use it ?
if in this method, yes, we can pass listFields instead of ZNRecord. But why do we need to avoid ZNRecord?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PartitionAssignmentGenerator.java
npawar: Partition assignments are applicable only to realtime tables. Will put that in the javadoc

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+
+  List<String> _tablesForPartitionAssignment;
+  List<String> _instanceNames;
+  int _nReplicas;
+  Map<String, Integer> _nPartitions;
```
mcvsubbu: I would name maps by the items they represent (in this case _tableToNumPartitions, or something like that)

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: @kishoreg , PartitionAwareRealtimeRoutingTableBuilder attempts to optimize the number of servers to which the request is sent in the case that the number of consuming  hosts is a multiple of number of replicas, AND we do a uniform (kafka/stream) partition allocation strategy (what is referred to as 'spraying' in this PR). The routing table builder will still work if these conditions don't match, except that the request fan-out will be sub-optimal for the high performance use cases where we want to minimize the number of servers participating in a query.

What we are attempting to do here is to continue to preserve that optimized behavior as much as possible. Using helix auto-rebalancer will  throw off partition-aware routing completely. 

We can discuss more offline.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
npawar: I see the constants defined only in CommonConstants.Helix.StateModel.SegmentOnlineOfflineStateModel.ONLINE, and did not think using that made sense here

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotTableIdealStateBuilder.java
npawar: Fixed

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/AutoRebalancePartitionAssignmentGenerator.java
npawar: I can create a RealtimePartition class, but would prefer to keep it private to just the AutoRebalancePartitionAssignment, as these partition names make sense only for this strategy.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PartitionAwarePartitionAssignmentGenerator.java
npawar: Renamed to PartitionAwareAssignmentGenerator

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
npawar: Added a comment to fix this.

npawar:
Introduced streamConsumptionConfig in the tableConfig. StreamConsumptionConfig has streamPartitionAssignmentStrategy, which will decide which partition assignment strategy is used for the table. If nothing is specified, we default to UniformStreamPartitionAssignment.
Will address these things in future PRs:
1. Race conditions in case of partition assignments which try to read and write all the tables. We could enter into situations where 2 controllers are trying to read and update the partition assignments (simultaneous partition update by validation manager and new table addition). 
2. Abstracting out the stream to be able to support any stream, and not just kafka
@mcvsubbu @kishoreg

npawar:
Fixed issues with comments and constants
@mcvsubbu

PR #2429, opened 2018-02-02
[TE] frontend - Ttbach/acceptance test rootcause empty state

ttbach:
Added an acceptance test for the empty state for rootcause page

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/application/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/application/route.js
@@ -33,6 +33,8 @@ export default Ember.Route.extend(ApplicationRouteMixin, {
 
   /**
    * if the app is in test mode, have the sessions automatically authenticated
+   * TODO: This is a hack. In future tests, this should be removed and use authenticateSession() instead
+   * https://github.com/simplabs/ember-simple-auth/issues/727
    * @method afterModel
    * @return {undefined}
    */
--- a/thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
@@ -1,9 +1,10 @@
 import Ember from 'ember';
 import { filterObject, filterPrefix, toBaselineUrn, toCurrentUrn, toOffsetUrn, toFilters, appendFilters, dateFormatFull } from 'thirdeye-frontend/utils/rca-utils';
 import EVENT_TABLE_COLUMNS from 'thirdeye-frontend/mocks/eventTableColumns';
-import config from 'thirdeye-frontend/mocks/filterBarConfig';
+import filterBarConfig from 'thirdeye-frontend/mocks/filterBarConfig';
 import fetch from 'fetch';
 import moment from 'moment';
+import config from 'thirdeye-frontend/config/environment';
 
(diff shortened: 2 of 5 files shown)


thirdeye/thirdeye-frontend/tests/acceptance/rootcause-test.js
```diff
@@ -0,0 +1,24 @@
...
+
+test('empty state of rootcause page should have a placeholder and no tabs', async (assert) => {
+  await visit('/rootcause');
+
+  assert.equal(
+    currentURL(),
+    '/rootcause',
+    'link is correct');
+  assert.ok(
+    find(PLACEHOLDER).get(0),
```
theseyi: I believe using the `ember-native-dom-helpers` should work now, given the upcoming change to testing in Ember

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -137,7 +138,10 @@ export default Ember.Controller.extend({
       timeseriesMode: 'split'
     });
 
-    Ember.run.later(this, this._onCheckSessionTimer, ROOTCAUSE_SESSION_TIMER_INTERVAL);
+    // This is a flag for the acceptance test for rootcause to prevent it from timing out because of this run loop
+    if (config.environment !== 'test') {
+      Ember.run.later(this, this._onCheckSessionTimer, this.get('rootcauseTimerInterval'));
```
theseyi: What's the intent of this run loop? Can the intended action be done in a more deterministic way rather than what is essentially a `setTimeout` operation?

ttbach: The run loop is intended to fetch the latest state of the session every 300s. I don't know if there is another way of doing this other than this. Do you have any suggestions on a better way?

thirdeye/thirdeye-frontend/tests/acceptance/rootcause-test.js
ttbach: Good to know. Thanks.

thirdeye/thirdeye-frontend/app/pods/application/route.js
```diff
@@ -33,6 +33,8 @@ export default Ember.Route.extend(ApplicationRouteMixin, {
 
   /**
    * if the app is in test mode, have the sessions automatically authenticated
+   * TODO: This is a hack. In future tests, this should be removed and use authenticateSession() instead
```
justYves: Thanks for the honesty 👍

thirdeye/thirdeye-frontend/app/pods/rootcause/template.hbs
```diff
@@ -138,7 +138,7 @@
       </div>
 
       {{#if context.urns.size}}
-        <div class="card-container__header">
+        <div class="card-container__header rootcause-tabs">
```
justYves: undocumented change? what does this do? (nit obvsly)

ttbach: I needed to check whether tabs are absent in the acceptance test, and I needed a unique class name to identify the tabs div. I figured the test will be more readable with `find('.rootcause-tabs').get(0)` as opposed to something like `find('.card-container__header').get(3)`.

PR #2431, opened 2018-02-02
Modify murmur function to always use UTF-8 bytes

snleee:
Murmur partition function picks up the default charset, which can be different 
based on the system environment. This forces murmur partition function to use
UTF-8 bytes.


pinot-core/src/main/java/com/linkedin/pinot/core/data/partition/MurmurPartitionFunction.java
```diff
@@ -31,6 +32,7 @@
  */
 public class MurmurPartitionFunction implements PartitionFunction {
   private static final String NAME = "Murmur";
+  private static final Charset UTF_CHARSET = Charset.forName("UTF-8");
```
mayankshriv: May be better name is UTF8 or UTF8_ChARSET

PR #2432, opened 2018-02-02
Clean up interface IndexSegment and SegmentMetadata

Jackie-Jiang:
1. Removed IndexType (some code assume it's offline/realtime which is not the case)
2. Removed IndexSegment.getAssociatedDirectory()
3. Removed SegmentMetadata.toMap()
4. Changed IndexSegment.getColumnNames to return Set<String>
5. Cleaned up all dummy SegmentMetadata, and replaced with MockSegmentMetadataUtils for all tests


pinot-core/src/main/java/com/linkedin/pinot/core/indexsegment/IndexSegment.java
```diff
@@ -20,57 +20,53 @@
...
+  SegmentMetadata getSegmentMetadata();
 
   /**
-   * @return SegmentMetadata
+   * Returns all the columns inside the segment.
+   *
+   * @return Array of column names
    */
-  public SegmentMetadata getSegmentMetadata();
+  String[] getColumnNames();
```
mayankshriv: Does this require implementations to convert their storage (eg List/Set) into a new String[], creating unnecessary garbage? If so, let's see if we can cleanup the interface to appropriate return type.

PR #2435, opened 2018-02-02
[TE] Self-Serve Onboarding 02: enhancements

newsummit:
- Allowing transition to Alert Page even if replay returns "timeout"
- Refreshing the model once replay is done (to fetch fresh anomaly data)
- Changing default time span for alert page to 3 months to sync with tuning range
- Adding alert deletion method in search page to be used once we have proper access to admin permissions
- For fetching dimension data in Create Alert, only reloading graph if we have new data
- Fix for onboarding: prevent sending null for alertRecipients

In second commit:
- Adding temporary URL query param for testing and alert clean-up

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/manage/alerts/index/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/manage/alerts/index/controller.js
@@ -4,12 +4,13 @@
  * @exports alerts controller
  */
 import Ember from 'ember';
-import { task, timeout } from 'ember-concurrency';
 import fetch from 'fetch';
 import _ from 'lodash';
+import { task, timeout } from 'ember-concurrency';
+import { checkStatus } from 'thirdeye-frontend/utils/utils';
 
 export default Ember.Controller.extend({
-  queryParams: ['selectedSearchMode', 'alertId'],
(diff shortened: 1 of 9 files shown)


thirdeye/thirdeye-frontend/app/pods/manage/alerts/index/controller.js
```diff
@@ -248,6 +248,23 @@ export default Ember.Controller.extend({
...
     },
 
+    /**
+     * Send a DELETE request to the function delete endpoint.
+     * @method removeThirdEyeFunction
+     * @param {Number} functionId - The id of the alert to remove
+     * @return {Promise}
+     */
+    removeThirdEyeFunction(functionId) {
+      const postProps = {
```
justYves: nit: Let's move this into a `/utils/constant.js` since this is needed for every post request with fetch.

newsummit: My bad. Its actually already there in utils. Forgot to use it.

newsummit: On second thought, I do have such a utility, but needs refactoring - its only for POST requests. I'll make a note to include it.

PR #2437, opened 2018-02-02
[TE] Implement Image Embed Classes

cyy0714:
Implement an email image embed interfaces and classes so that future image embed process can be simplified.


justYves:
@cyy0714 Let's close this since we went with another implementation?

PR #2439, opened 2018-02-02
Clean up RealtimeSegment interface and SegmentMetadataImpl class

Jackie-Jiang:
1. Remove MutableIndexSegment interface, move APIs into RealtimeSegment interface
2. Remove the constructor with OfflineSegmentZKMetadata because we should never construct SegmentMetadataImpl with ZKMetadata
3. Fix an issue where RealtimeSegmentImpl numDocsIndexed should be volatile, remove redundant counters
4. Let ValidationManager directly work on SegmentZKMetadata, reducing the overhead of converting it to SegmentMetadataImpl


pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -258,53 +249,44 @@ void validateLLCSegments(final String realtimeTableName, TableConfig tableConfig
...
-      List<Interval> missingIntervals = computeMissingIntervals(segmentIntervals, segmentMetadataList.get(0).getTimeGranularity());
-      missingSegmentCount = missingIntervals.size();
-
+      List<Interval> missingIntervals = computeMissingIntervals(segmentIntervals, timeGranularity);
       for (Interval missingInterval : missingIntervals) {
-        LOGGER.warn("Missing data in table {} for time interval {}", tableName, missingInterval);
+        LOGGER.warn("Missing data in table {} for time interval {}", offlineTableName, missingInterval);
       }
+      // Update the gauge that contains the number of missing segments
+      _validationMetrics.updateMissingSegmentCountGauge(offlineTableName, missingIntervals.size());
```
mcvsubbu: This needs to be outside the loop, set to 0 if there are no missing segments, right? Otherwise, the gauge for the table, once set, can never be reset to 0

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -71,13 +70,10 @@
...
   private final Map<String, MutableDictionary> dictionaryMap = new HashMap<>();
 
-  private AtomicInteger docIdGenerator;
   private String outgoingTimeColumnName;
   private TimeGranularitySpec outgoingGranularitySpec;
 
-  private int docIdSearchableOffset = -1;
-  private int numDocsIndexed = 0;
-  private int numSuccessIndexed = 0;
+  private transient int _numDocsIndexed = 0;
```
mcvsubbu: volatilie? Why transient?

Jackie-Jiang: Good catch! Should be volatile, forgive my poor English...

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
Jackie-Jiang: My logic here is that, if the number of segments is not enough to support this check, we should not update the count. What do you think?

Jackie-Jiang: Discussed offline, will move this outside the loop

PR #2446, opened 2018-02-05
[TE] frontend - Acceptance test for rootcause metric

ttbach:
- Added test for rootcause page when only a metric is provided
- Created an `app/mirage/endpoints` folder to extract mirage endpoints to group and shorten `mirage/config.js` (inspired by this https://github.com/samselikoff/ember-cli-mirage/issues/1027).
- Removed unnecessary params from mirage endpoints

### DIFF
--- a/thirdeye/thirdeye-frontend/app/mirage/endpoints/rootcause.js
+++ b/thirdeye/thirdeye-frontend/app/mirage/endpoints/rootcause.js
@@ -1,0 +1,38 @@
+export default function(server) {
+  /**
+   * Get request for rootcause page
+   */
+  server.get(`/rootcause/raw`, () => {
+    return [ {
+      "urn" : "thirdeye:metric:1",
+      "score" : 1.0,
+      "label" : "thirdeyeKbmi::pageViews",
+      "type" : "metric",
+      "link" : null,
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-frontend/app/mirage/endpoints/rootcause.js
```diff
@@ -0,0 +1,48 @@
...
+      "id": 1,
+      "alias": "pageViews"
+    };
+  });
+
+  /**
+   * Retrieves timestamp of last element in timeseries
+   */
+  server.get('/data/maxDataTime/metricId/1', () => {
+    return 1517846399998;
```
justYves: is it timeproof? i.e. we run this in 5 months, is it still going to work?

ttbach: For the test, the number mocked here is irrelevant for our test case right now. I might revisit this for when we test the chart. @apucher thoughts?

apucher: not relevant at this point. subject to change once we mock timeseries data.

PR #2449, opened 2018-02-06
[TE] Bug fix for user report to add function Id

cecilynie:
User report anomalies to be exposed to UI
Add functionId to user report anomalies to let API fetch anomalies under function Id.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -1012,8 +1012,8 @@ public class AnomalyResource {
 
       for (MergedAnomalyResultDTO mergedAnomaly : mergedResults) {
         // if use notified flag, only keep anomalies isNotified == true
-        if ( (useNotified && mergedAnomaly.isNotified()) || !useNotified || mergedAnomaly.getAnomalyResultSource().equals(
-            AnomalyResultSource.USER_LABELED_ANOMALY)) {
+        if ( (useNotified && mergedAnomaly.isNotified()) || !useNotified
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -1011,7 +1012,8 @@ private String getDimensions(String dataset, String exploreDimensions) throws Ex
 
       for (MergedAnomalyResultDTO mergedAnomaly : mergedResults) {
         // if use notified flag, only keep anomalies isNotified == true
-        if ( (useNotified && mergedAnomaly.isNotified()) || !useNotified) {
+        if ( (useNotified && mergedAnomaly.isNotified()) || !useNotified || mergedAnomaly.getAnomalyResultSource().equals(
```
cyy0714: Use AnomalyResultSource.USER_LABELED_ANOMALY).equals(...) instead.

cecilynie:
@apucher It's because FE is fetching anomalies using anomaly function Id. The create user report logic  is adding anomaly function but not adding the function Id. From integration test I see the created anomaly has anomaly functionid=0. This is a quick fix. A better fix should be when adding "anomaly function", the functionId field should also be populated.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
cecilynie: fixed. thanks for advising!

apucher:
I'm referring to the change in the if-statement on line 1014

cecilynie:
@apucher "User-Report" Anomaly is "not notified". Frontend is using "notified flag" to show the anomaly list, hence previously "User-Report" is never shown in the list.

PR #2450, opened 2018-02-06
Refactor RealtimeSegmentImpl

Jackie-Jiang:
1. Add class RealtimeSegmentConfig to help construct the RealtimeSegmentImpl
2. Remove unnecessary row validation because the indexed row has already been transfered and validated
3. Always loop over all field specs instead of checking dimensions, metrics and time separately to support new added DateTimeField
4. Simplify the base class RealtimeSegmentDataManager


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -50,253 +43,222 @@
 
 
 public class RealtimeSegmentImpl implements RealtimeSegment {
-  private final Logger LOGGER;
-  public static final int[] EMPTY_DICTIONARY_IDS_ARRAY = new int[0];
+  // For multi-valued column, forward-index.
+  // Maximum number of multi-values per row. We assert on this.
+  private static final int MAX_MULTI_VALUES_PER_ROW = 1000;
 
-  private static final String MMGR_DICTIONARY_SUFFIX = ".dict";
```
mcvsubbu: these suffixes help us look over the logs and see which part allocates memory. so please add these back.

PR #2452, opened 2018-02-06
Utilize latest version of Helix for Pinot

jackjlli:
This PR utilizes latest version of Helix for Pinot. 
We won't merge this PR until we fully test all the components, making sure they're all good.

### DIFF
--- a/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
+++ b/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
@@ -15,9 +15,11 @@
  */
 package com.linkedin.pinot.server.starter.helix;
 
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
```diff
@@ -89,6 +91,11 @@ public String getMessageType() {
     return Message.MessageType.USER_DEFINE_MSG.toString();
   }
 
+  @Override
+  public List<String> getMessageTypes() {
+    return ImmutableList.of(Message.MessageType.USER_DEFINE_MSG.name());
```
mayankshriv: Can the result change between invocations? If not, why not use static final to avoid new object creation?

mcvsubbu: makes sense.

PR #2454, opened 2018-02-07
[TE] fix build-tooltip utils test

justYves:
### What's new: 
- Remove the incorrect template helper test (probably created with an ember g) 
- Added the correct unit test file under utility and added a basic unit test


thirdeye/thirdeye-frontend/tests/unit/utils/build-tooltip-test.js
```diff
@@ -0,0 +1,20 @@
+import BuildTooltip from 'thirdeye-frontend/utils/build-tooltip';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | build-tooltip');
+
+test('it builds', function(assert) {
```
ttbach: nit: we can use arrow function for here

PR #2463, opened 2018-02-08
Re-implement FixedByteSingleColumnSingleValueReaderWriter class to support mutablilty and writes at random rows.

mayankshriv:
The existing implementation of this class did not mutability of written
values, and writes at random rows. Also, it assumed that callers
will call writes with strictly increment (by 1) row index.

1. Re-implementated to support mutability of already written rows and
random writes.

2. Also eliminated assumption of writers calling writes at strictly
incrementing (by 1) row index (eg write row 1,2,3,....). With this,
writes can be at rows something like 1,4,7...
Rows not written will default to zero value.

3. Added unit tests to cover new features.


pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -29,191 +29,108 @@
...
-
-  private int getBufferId(int row) {
-    return row / numRowsPerChunk;
-  }
-
   @Override
   public void setInt(int row, int i) {
     addBufferIfNeeded(row);
-    currentWriter.setInt(row, i);
+    getWriterForRow(row).setInt(row, i);
```
kishoreg: will this impact performance for append-only use case?

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -224,19 +141,126 @@ public void readValues(int[] rows, int rowStartPos, int rowSize, int[] values, i
...
+        new FixedByteSingleValueMultiColReader(buffer, _numRowsPerChunk, new int[]{_columnSizesInBytes});
+    FixedByteSingleValueMultiColWriter writer =
+        new FixedByteSingleValueMultiColWriter(buffer, _numRowsPerChunk, /*cols=*/1, new int[]{_columnSizesInBytes});
+
+    final int startRowId = _numRowsPerChunk * (_dataBuffers.size() - 1);
+    _writers.add(new WriterWithOffset(writer, startRowId));
+    _readers.add(new ReaderWithOffset(reader, startRowId));
+  }
+
+  private void addBufferIfNeeded(int row) {
```
kishoreg: why is this computation needed for every row? when we add a new buffer, we can just compute the max number of rows right and check for that?

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -29,191 +29,108 @@
...
-
-    public double getDouble(int row) {
-      return reader.getDouble(row - startRowId, 0);
-    }
-
-    public FixedByteSingleValueMultiColReader getReader() {
-      return reader;
-    }
-  }
+  private final RealtimeIndexOffHeapMemoryManager _memoryManager;
```
kishoreg: why are we hard coding realtimememory manager, this class should just take a memorymanager and not care about the implementation

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
mayankshriv: This should be a very small overhead in comparison of all the other operations happening during indexing. I ran a benchmark with old and new code to set 1M Long values, and both perform about the same (Total of ~50ms for 1M values).

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
mayankshriv: Yes, good point. Will replace _capacity with _capacityInRows, so we can first check the row number against that, and then do the more expensive computation if required.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
mcvsubbu: you can rename RealtimeIndexOffHeapMemoryManager to MemoryManager (or whatever generic name you like) and rename getSegmentName() method to getAllocationContext(), and you should be good to go.

The class is not correctly named, but that was intentionally so, since we were using it in the realtime context only.

mayankshriv:
@mcvsubbu Caller needs to manager how they are using this class. Added comment indicating that rows not written will default to zero.

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
mcvsubbu: thanks for verification with jmh

pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -29,191 +29,109 @@
 import org.slf4j.LoggerFactory;
 
 
-// TODO Check if MMAP allows us to pack more partitions on a single machine.
+/**
+ * This class implements reader as well as writer interfaces for fixed-byte single column and single value data.
+ * <ul>
+ *   <li> Auto expands memory allocation on-demand. </li>
+ *   <li> Supports random reads and writes. </li>
+ *   <li> Rows not written but allocated default to 0. </li>
```
mcvsubbu: not sure if it guaranteed to be 0, maybe dependent on the mem manager. may want to make that clear that it is user's responsibility to use it appropriately.

PR #2469, opened 2018-02-08
[TE] frontend - ttbach/remove edit button on rootcause page

ttbach:
- Removed edit pencil icon on rootcause page, so it behaves like google doc

edit-title


ttbach:
@justYves Nice catch. I can increase the width.

PR #2472, opened 2018-02-09
Use Java 8

jfim:
Use Java 8 for binary builds and Travis


jfim:
Waiting on some confirmations before merging this one, even if it passes Travis.

PR #2476, opened 2018-02-09
Re-factor: Code re-factor to separate out realtime specific references from fixed-byte readers/writers.

mayankshriv:
1. Created an interface PinotDataBufferMemoryManager, that is now passed
to fixed-byte readers/writers.

2. Eliminated segmentName being stored in memory manager. Instead, now
callers allocating memory pass an allocationContext
(segmentName:columnName), that is used for logging within memory
manager. RealtimeOffHeapMemoryManager still has dependency on segment
name, that is harder to remove, and will explode this PR, therefore not
addressed.

3. Fixed code-style for all touched files.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
@@ -17,7 +17,7 @@ package com.linkedin.pinot.core.io.readerwriter.impl;
 
 import com.linkedin.pinot.core.io.reader.impl.FixedByteSingleValueMultiColReader;
 import com.linkedin.pinot.core.io.readerwriter.BaseSingleColumnSingleValueReaderWriter;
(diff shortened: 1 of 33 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/io/readerwriter/impl/FixedByteSingleColumnSingleValueReaderWriter.java
```diff
@@ -163,9 +163,8 @@ private int getBufferId(int row) {
   }
 
   private void addBuffer() {
-    LOGGER.info("Allocating {} bytes for column {} for segment {}", _chunkSizeInBytes, _columnName,
-        _memoryManager.getSegmentName());
-    PinotDataBuffer buffer = _memoryManager.allocate(_chunkSizeInBytes, _columnName);
+    LOGGER.info("Allocating {} bytes for segment:column {}", _chunkSizeInBytes, _allocationContext);
```
mcvsubbu: Since this class is expected to be for some generic use, perhaps you can just remove the "segment:column" from the log message.

mayankshriv: Done.

PR #2480, opened 2018-02-10
[TE] frontend - Ttbach/add baseline scores

ttbach:
- Added scores to each baseline column in metrics table
- Extracted `metrics-table` stylings to its own scss file
- Fixed bug on `humanizeFloat` util
- Refactored 4 `metrics-table-changes` custom templates into 1 general template for `metrics-table`'s offset columns

screen shot 2018-02-09 at 4 21 42 pm


thirdeye/thirdeye-frontend/app/pods/custom/metrics-table-changes/wo1w/template.hbs
```diff
@@ -1,3 +1,6 @@
-<span class="rootcause-metric__table--{{record.wo1w.direction}}">
+<p class="metrics-table__baseline-score">
```
justYves: BLOCKER: Any reasons we built 4 different templates here? 
It feels like this could be accomplished with the same template and just pass 1. a score and 2. a value?

apucher: that's a legacy hack of mine. if you find a better solution please go ahead, but it wasn't immediately clear how pass separate key/value pairs for the same "record"

apucher:
semantic comment: use metric "value" instead of "score" (that's what's in the "Anomalous" column)

PR #2487, opened 2018-02-14
[TE] datasource - use dataset maxtime when fetching filters from pinot

apucher:
TE uses a fixed offset to the current timestamp to determine the time range for retrieving dimension names/values for a dataset. if the dataset does not contain data (or the server time is off) no filters are read. this PR changes the behavior to fetch data based on the latest available timestamp in the dataset instead.


apucher:
this is a very special kind of troll

apucher:
fixed

PR #2488, opened 2018-02-14
Add failure message in entity of response from uploadSegment

npawar:
The failure message was not being propagated from controller to the client. Client received only status code and status reason, which are insufficient indicators of the failure. Passing the failure message in the response entity

`
com.linkedin.pinot.common.exception.HttpErrorStatusException: Got error status code: 403 with reason: Forbidden while sending request: http://localhost:8100/segments
Message: Quota check failed for segment: test_segment_0 of table: test_table_OFFLINE, reason: Estimated size: 140649 bytes exceeds the configured quota of 1024 (bytes) for table test_table_OFFLINE. Incoming segment size: 140650 (bytes) to controller: xxxx, version: Unknown
`

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
@@ -236,11 +236,11 @@ public class FileUploadDownloadClient implements Closeable {
       controllerVersion = response.getFirstHeader(CommonConstants.Controller.VERSION_HTTP_HEADER).getValue();
     }
     StatusLine statusLine = response.getStatusLine();
-    String message = null;
+    String message;
     try {
       message = EntityUtils.toString(response.getEntity());
     } catch (Exception e) {
-      // ignore
+      message = "No message";
     }
     String errorMessage =
         String.format("Got error status code: %d with reason: %s while sending request: %s\nMessage: %s", statusLine.getStatusCode(),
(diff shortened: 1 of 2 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
```diff
@@ -235,9 +236,15 @@ private static String getErrorMessage(HttpUriRequest request, CloseableHttpRespo
       controllerVersion = response.getFirstHeader(CommonConstants.Controller.VERSION_HTTP_HEADER).getValue();
     }
     StatusLine statusLine = response.getStatusLine();
+    String message = null;
+    try {
+      message = EntityUtils.toString(response.getEntity());
+    } catch (Exception e) {
+      // ignore
```
raviarin: Add an empty message or a generic message stating that unable to extract message.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
```diff
@@ -235,9 +236,15 @@ private static String getErrorMessage(HttpUriRequest request, CloseableHttpRespo
...
+    }
     String errorMessage =
-        String.format("Got error status code: %d with reason: %s while sending request: %s", statusLine.getStatusCode(),
-            statusLine.getReasonPhrase(), request.getURI());
+        String.format("Got error status code: %d with reason: %s while sending request: %s\nMessage: %s", statusLine.getStatusCode(),
+            statusLine.getReasonPhrase(), request.getURI(), message);
```
raviarin: Is this the only instance that we missed adding the message or do we need to d a sweep and add it everywhere?

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
npawar: Fixed

PR #2489, opened 2018-02-14
In MetricFieldSpec, only serialize non-default field size

Jackie-Jiang:
Only serialize field size when the data type is STRING and field size is set


Jackie-Jiang:
@snleee Caught illegal test (long with field size 10) lol

PR #2495, opened 2018-02-15
Enforce padding character to be '\0'

Jackie-Jiang:
The current code will break if the padding character is not '\0'

1. Remove the config for padding character
2. Simplify the logic of creating string dictionary because the order won't change


mcvsubbu:
We found about 23k segments across some 30 or so tables. Give us some time as we figure out a way please. Thanks.

mcvsubbu:
> We found about 23k segments across some 30 or so tables. Give us some time as we figure out a way please. Thanks.

This is not a blocker anymore, you can merge the PR.

PR #2496, opened 2018-02-15
In SimpleMinionClusterIntegrationTest, manually control the task execution process instead of using sleep

Jackie-Jiang:
Make test result more deterministic, and reduce redundant sleep

### DIFF
--- a/pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/SimpleMinionClusterIntegrationTest.java
+++ b/pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/SimpleMinionClusterIntegrationTest.java
@@ -16,7 +16,6 @@
 package com.linkedin.pinot.integration.tests;
 
 import com.google.common.base.Function;
-import com.google.common.util.concurrent.Uninterruptibles;
 import com.linkedin.pinot.common.config.PinotTaskConfig;
 import com.linkedin.pinot.common.config.TableConfig;
 import com.linkedin.pinot.common.config.TableNameBuilder;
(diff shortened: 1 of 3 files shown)


pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/SimpleMinionClusterIntegrationTest.java
```diff
@@ -89,22 +88,35 @@ public void setUp() throws Exception {
...
+    TestTaskExecutor.HOLD.set(true);
+
     // Should create the task queues and generate a task
-    Assert.assertEquals(_taskManager.scheduleTasks().size(), 1);
-    Assert.assertEquals(_helixTaskResourceManager.getTaskQueues().size(), 2);
+    Assert.assertTrue(_taskManager.scheduleTasks().containsKey(TestTaskGenerator.TASK_TYPE));
+    Assert.assertTrue(_helixTaskResourceManager.getTaskQueues()
+        .contains(PinotHelixTaskResourceManager.getHelixJobQueueName(TestTaskGenerator.TASK_TYPE)));
 
     // Should generate one more task
```
mcvsubbu: can u also fix the comment if it is not applicable? We are only testing for presence of task type, maybe 1, maybe many, right?

Jackie-Jiang: The scheduleTasks() will generate at most one task under each task type, so we just need to check the key.

PR #2497, opened 2018-02-15
Fixed ClusterStateVerifier to consider timeout

mcvsubbu:
ClusterStateVerifier now pays attention to the timeout setting that the caller may provide.
This was needed while testing upgrade to helix-0.8.0

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
@@ -32,6 +32,9 @@ import org.slf4j.LoggerFactory;
 public class ClusterStateVerifier extends PinotZKChanger {
   private static final Logger LOGGER = LoggerFactory.getLogger(ClusterStateVerifier.class);
 
+  private static int MIN_SLEEP_BETWEEN_CHECKS_MILLIS = 100;
+  private static int MAX_SLEEP_BETWEEN_CHECKS_MILLIS = 30_000;
+
   public ClusterStateVerifier(String zkAddress, String clusterName) {
     super(zkAddress, clusterName);
   }
@@ -47,8 +50,8 @@ public class ClusterStateVerifier extends PinotZKChanger {
     final ExecutorService executor = Executors.newSingleThreadExecutor();
     final Future<Boolean> future = executor.submit(new Callable<Boolean>() {
       @Override
-      public Boolean call() {
-        return isClusterStable(tableNames);
+      public Boolean call() throws Exception {
+        return waitForClusterStable(tableNames, timeoutSec);
       }
     });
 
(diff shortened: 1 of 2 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -74,18 +77,42 @@ public Boolean call() {
    * @param tableNames list of table names which are about to scan
```
antumbde: update javadoc

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -74,18 +77,42 @@ public Boolean call() {
...
+    while (!stable) {
+      iteration++;
+      LOGGER.info("Start scanning the stability of {} tables, iteration {}",
+          tableNames.size(), iteration);
+      int numUnstablePartitions = 0;
+      for (String tableName : tableNames) {
+        numUnstablePartitions = super.isStable(tableName);
+        if (numUnstablePartitions != 0) {
+          LOGGER.error("Table {} is not stable. numUnstablePartitions: {} ", tableName, numUnstablePartitions);
+          break;
```
antumbde: I've limited context for this tool and concept of stability. This check is not atomic (a checked table may turn unstable by the time this processes last table). Given that, in case of some unstable tables, does it make sense to check all tables and truncate already validated tables before sleeping ?  That will depend on the total time taken to check for stability.

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
```diff
@@ -74,18 +77,42 @@ public Boolean call() {
...
+        }
+      }
+      stable = (numUnstablePartitions == 0);
+      if (!stable) {
+        if (System.currentTimeMillis() >= maxEndTimeMillis) {
```
antumbde: How important is it to account for the time to check stability ? I guess not much...but a simpler modification to the condition can be (current_time + sleepTime)  >= maxEndTime

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
mcvsubbu: Correct, this may happen.The intention here is that all tables reach stability, so we are OK. Worst case, we time out, and there is a message that says which tables are not stable, so it is ok.

pinot-tools/src/main/java/com/linkedin/pinot/tools/ClusterStateVerifier.java
mcvsubbu: It is ok in this case. The time taken to check should be very small for a reasonable number of tables in the cluster.

mcvsubbu:
I guess it makes sense to provide some context. In this case, the tool was used to run a 24-hour test with new helix package, taking turns to kill a random server/broker/controller and wait for tables to get stable. Realtime tables were also adding segments in the meantime on a couple of tables.

PR #2498, opened 2018-02-15
[TE] detection - abort long-running detection task on interrupt

apucher:
ThirdEye currently does not allow interruption of detection tasks running across multiple dimension combinations (potentially millions). This causes the controller to appear unresponsive to systems and monitoring tools outside the VM. This PR makes the detection task runner aware of interrupts to abort ongoing tasks and shut down gracefully.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
@@ -175,6 +175,10 @@ public class DetectionTaskRunner implements TaskRunner {
 
     DataFilter dataFilter = DataFilterFactory.fromSpec(anomalyFunctionSpec.getDataFilter());
     for (DimensionMap dimensionMap : anomalyDetectionInputContext.getDimensionMapMetricTimeSeriesMap().keySet()) {
+      if (Thread.interrupted()) {
+        throw new IllegalStateException("Thread interrupted. Aborting.");
+      }
+
       // Skip anomaly detection if the current time series does not pass data filter, which may check if the traffic
       // or total count of the data has enough volume for produce sufficient confidence anomaly results
       MetricTimeSeries metricTimeSeries =


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
```diff
@@ -175,6 +175,10 @@ private void storeData(AnomalyDetectionOutputContext anomalyDetectionOutputConte
 
     DataFilter dataFilter = DataFilterFactory.fromSpec(anomalyFunctionSpec.getDataFilter());
     for (DimensionMap dimensionMap : anomalyDetectionInputContext.getDimensionMapMetricTimeSeriesMap().keySet()) {
+      if (Thread.interrupted()) {
```
cyy0714: How the interrupt is triggered?

cyy0714: With interruption, what will be the status of the task and job? If it is ongoing or failed, what will be the plan to recover that after the system is up?

apucher: Interrupt happens at process level, e.g. via SIGINT, and is passed down to the threads via the ExecutorService's shutdownNow(). The exit state of the task should be "FAILED" and recover by the same means you recover from other errors during detection.

cyy0714: Is it possible to apply to interrupt a specific detection job? FYI, currently the only way to recover the failed jobs is triggering replay offline.

cyy0714: Also, this works for the detection runs normally. How do we handle when there is a zombie task? For example, the detection stops when running on one of the dimensions. It won't enter into the if statement.

cyy0714: Can you add a task and job status as INTERRUPTED, so that we know this task/job was interrupted but failed?

apucher: Detection jobs a currently a single-task monolith, so interruption is all or nothing. Zombie tasks stuck inside an algorithm is out of scope - this has to be addressed by the developer of the algorithm (but can be done in the exact same way)

apucher:
this is merely a quick fix. The wider issue has to be addressed holistically with a new execution pipeline, such as the one @cyenjung is working on

PR #2499, opened 2018-02-15
Fix LLC consumer to handle segment build fails gracefully

mcvsubbu:
If segment build fails then we used throw NPE while tyring to delete a null file name.
Fixed the condition, and added a test.
We now handle the condition gracefully, and also set LLC_PARTITION_CONSUMING gauge to 0
so that we can alert on the partition that stopped consumption.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -139,9 +139,8 @@ public class LLRealtimeSegmentDataManager extends RealtimeSegmentDataManager {
     }
 
     public void deleteSegmentFile() {
-      // If segment build fails with an exception (e.g. if the value of a string column contained
-      // a nul character followed by non-null characters), then we will not be able to create a segment file,
-      // so the file name will be null.
+      // If segment build fails with an exception then we will not be able to create a segment file and
+      // the file name will be null.
       if (_segmentFile != null) {
         FileUtils.deleteQuietly(new File(_segmentFile));
       }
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -139,7 +139,12 @@ public String getSegmentFile() {
     }
 
     public void deleteSegmentFile() {
-      FileUtils.deleteQuietly(new File(_segmentFile));
+      // If segment build fails with an exception (e.g. if the value of a string column contained
```
jfim: Not sure it's really worth mentioning this specific bug as an example of an exception that can happen during segment building. If we fix the issue, for example, we'll need to update this comment.

pinot-core/src/test/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
```diff
@@ -760,6 +780,9 @@ protected boolean buildSegmentAndReplace() {
     @Override
     protected String buildSegmentInternal(boolean forCommit) {
       _buildSegmentCalled = true;
+      if (_failSegmentBuild) {
```
jfim: Seems like this test is very heavily coupled to the actual implementation (white box) as evidenced by the numerous flags (eg. `_buildAndReplaceCalled`). Do we have plans to refactor this test into something cleaner at some point?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mcvsubbu: ok, i will remove the comment

PR #2507, opened 2018-02-20
Rebalance strategy interface and DefaultRebalanceStrategy implementation

npawar:
Introduce an interface for RebalanceSegmentsStrategy, which has rebalancePartitions and rebalanceSegments
Adding defaultRebalanceStrategy implementation, which rebalances stream partition assignment for rebalancing consuming segments, and also rebalances completed segments using auto rebalance strategy


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/RebalanceSegmentsFactory.java
```diff
@@ -0,0 +1,56 @@
...
+package com.linkedin.pinot.controller.helix.core.rebalance;
+
+import com.linkedin.pinot.common.config.TableConfig;
+import org.apache.helix.HelixManager;
+
+
+/**
+ * Singleton factory class, to fetch the right rebalance segments strategy based on table config
+ */
+public class RebalanceSegmentsFactory {
```
snleee: Can we rename to `RebalanceSegmentStrategyFactory`?

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerStarter.java
```diff
@@ -152,6 +153,9 @@ public void start() {
       LOGGER.info("Starting segment status manager");
       _segmentStatusChecker.start();
 
+      LOGGER.info("Creating rebalance segments factory");
+      RebalanceSegmentsFactory.createInstance(_helixResourceManager.getHelixZkManager());
```
snleee: This will require us to restart the controller in case we change the segment assignment strategy since `RebalanceSegmentsFactory` is a singleton class. Do you prefer to leave this way or do you think it's better to create rebalance strategy every time we need to use?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  /**
+   * Helper method to update idealstate with the new segment assignment
+   *
+   * @param tableNameWithType Table name with type
+   * @param numReplica The number of replica
+   * @param segmentAssignmentMapping Segment assignment mapping
+   */
+  private void updateIdealStateWithNewSegmentMapping(final String tableNameWithType, final int numReplica,
```
snleee: I think this function can be shared across the `RebalanceSegmentStrategy`. One way to achieve is to make an abstract class `BaseRebalanceSegmentStrategy` and make all the implementation extends that. Another way is to change the interface not to include the update to zk and pull this function up to `PinotHelixResourceManager`. I personally prefer to the latter.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/partition/UniformStreamPartitionAssignmentStrategy.java
```diff
@@ -38,21 +39,21 @@
...
-     ,"3":["Server_s4.company.com_8001","Server_s5.company.com_8001","Server_s6.company.com_8001"]
-     ,"4":["Server_s1.company.com_8001","Server_s2.company.com_8001","Server_s3.company.com_8001"]
-     ,"5":["Server_s4.company.com_8001","Server_s5.company.com_8001","Server_s6.company.com_8001"]
-     ,"6":["Server_s1.company.com_8001","Server_s2.company.com_8001","Server_s3.company.com_8001"]
-     ,"7":["Server_s4.company.com_8001","Server_s5.company.com_8001","Server_s6.company.com_8001"]
-   }
-   ,"mapFields":{
-   }
+ "id":"KafkaTopicName"
+ ,"simpleFields":{
```
snleee: Adding spaces back would look better?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/RebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Rebalances segments and writes ideal state of table
+   * @param idealState old ideal state
+   * @param tableConfig table config of table tor rebalance
+   * @param rebalanceUserConfig custom user configs for specific rebalance strategies
+   * @param newPartitionAssignment new rebalaned partition assignments as part of the resource rebalance
+   * @return rebalanced ideal state
+   */
+  IdealState rebalanceIdealState(IdealState idealState, TableConfig tableConfig, RebalanceUserConfig rebalanceUserConfig,
```
snleee: I would prefer to keep this interface for computing new idealstate mappin. I think we can put the write to zk in `HelixResourceManager`. How do you think?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/partition/StreamPartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,112 @@
...
+    for (String tableName : allTables) {
+      allTableConfigs.add(getRealtimeTableConfig(tableName));
+      Map<String, List<String>> partitionToInstancesMap = getPartitionsToInstances(tableName);
+      if (partitionToInstancesMap != null) {
+        PartitionAssignment partitionAssignment = new PartitionAssignment(tableName, partitionToInstancesMap);
+        tableNameToPartitionAssignment.put(tableName, partitionAssignment);
+      }
+    }
+
+    StreamPartitionAssignmentStrategy streamPartitionAssignmentStrategy =
```
snleee: This should return `UniformPartitionAssignmentStrategy` by default if we don't want to break the current tables. The factory should already take care of that, right?

npawar: The factory does take care of it. Unless explicitly specified in the table config to use something else, the factory will return UniformPartitionAssignmentStrategy.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/RebalanceSegmentsFactory.java
npawar: Fixed renaming and formatting

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/RebalanceUserConfig.java
```diff
@@ -0,0 +1,53 @@
...
+package com.linkedin.pinot.controller.helix.core.rebalance;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * Custom user config provided in rebalance, which will vary according to rebalance strategy
+ */
+public class RebalanceUserConfig {
```
mcvsubbu: Rename to RebalanceUserParams?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
npawar: I had thought of putting it in abstract base class too, but was confused between which approach to take. I like the approach of abstract class better, because that way all responsibility of writing (partition assignment and ideal state)  is with the implementation and not with the caller. If we move updateIdealState to HelixResourceManager, the writing of partition assignment would still happen in the implementation, whereas the writing of ideal state will happen outside. If we go with approach 2, should we also be pulling out writePartitionAssignment to PinotHelixManager? And in the future if another strategy wants to write something else, should that be moved too?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -388,10 +390,12 @@ public ZNRecord rebalance(
       @ApiParam(value = "true|false", required = true, defaultValue = "true") @QueryParam("dryrun") Boolean dryRun
   )
   {
+    RebalanceUserConfig rebalanceUserConfig = new RebalanceUserConfig();
+    rebalanceUserConfig.addConfig(RebalanceUserConfigProperties.DRYRUN, String.valueOf(dryRun));
```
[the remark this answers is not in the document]
npawar: Yes, added it

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/partition/StreamPartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,112 @@
...
+  protected ZNRecord getKafkaPartitionAssignment(final String realtimeTableName) {
+    final String path = ZKMetadataProvider.constructPropertyStorePathForKafkaPartitions(realtimeTableName);
+    return _propertyStore.get(path, null, AccessOption.PERSISTENT);
+  }
+
+  /**
+   * Given map of table name to stream partition assignment, construct and write znodes to property store one by one
+   * @param newPartitionAssignment
+   */
+  public void writeKafkaPartitionAssignment(Map<String, PartitionAssignment> newPartitionAssignment) {
```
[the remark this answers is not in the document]
npawar: Another PR with the abstraction will follow. But fixed these few spots for now

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -383,15 +386,22 @@ private void ensureMinReplicas(TableConfig config) {
   @Path("/tables/{tableName}/rebalance")
   @ApiOperation(value = "Rebalances segments of a table across servers", notes = "Rebalances segments of a table across servers")
   public ZNRecord rebalance(
```
snleee: Can we change this API to return both new idealstate mapping and partition assignment in a JSON format? This will be helpful for debugging since this API can potentially update ZK twice. We should be able to see both when we run with `dryrun` flag. For offline table, we can return the partition assignment as empty.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,302 @@
...
+      LOGGER.info("Rebalancing partition assignment for table {}", tableConfig.getTableName());
+
+      if (tableConfig.getTableType().equals(CommonConstants.Helix.TableType.REALTIME)) {
+
+        StreamPartitionAssignmentGenerator streamPartitionAssignmentGenerator = new StreamPartitionAssignmentGenerator(_propertyStore);
+        TagConfig tagConfig = new TagConfig(tableConfig, _helixManager);
+        List<String> consumingInstances = _helixAdmin.getInstancesInClusterWithTag(_helixClusterName, tagConfig.getConsumingServerTag());
+        String tableNameWithType = tableConfig.getTableName();
+        // FIXME: where should this come from? from kafka metadata stream or just read from znode how many are existing?
+        int numPartitions = getKafkaPartitionCount(tableConfig);
```
snleee: I think you are doing correctly. This should be the same as the logic from LLC segment manager, where it fetches the partition information from Kafka. `PinotTableIdealStateBuilder.getPartitionCount` is the correct wrapper for doing that.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,300 @@
...
+              idealState.getResourceName(), segmentId, instanceStateMap.size(), targetNumReplicas);
+        }
+      }
+    } else { // Number of replicas is either the same or higher, so invoke Helix rebalancer.
+
+      final Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+
+      List<Map.Entry<String, Map<String, String>>> removedEntries = new LinkedList<>();
+      if (tableConfig.getTableType().equals(CommonConstants.Helix.TableType.REALTIME)) {
+        // FIXME: what about those in OFFLINE states?
```
mcvsubbu: Once you get the names of the latest segments for each partition, you can mark them as CONSUMING state even if they are OFFLINE in the current IDEALSTATE record. The worst that will happen is that they go back to OFFLINE state because of some issues with consumption

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,300 @@
...
+        removeSegmentsNotBalanceable(mapFields, removedEntries);
+      }
+
+      if (mapFields.isEmpty()) {// if ideal state had only CONSUMING segments so far, mapFields will be empty
+        // we do not need to do anything in that case
+        LOGGER.info("No LLC segments in ONLINE state for table {}", tableConfig.getTableName());
```
mcvsubbu: the log may be misleading if we have already rebalanced consuming segments.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,300 @@
...
+      }
+      else {
+
+        String tableNameWithType = tableConfig.getTableName();
+
+        LinkedHashMap<String, Integer> states = new LinkedHashMap<>();
+        List<String> partitions = Lists.newArrayList(idealState.getPartitionSet());
```
mcvsubbu: s/partitions/segments/ to avoid confusion with stream partitions.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,300 @@
...
+    LOGGER.info("Rebalancing consuming segments for table {}", idealState.getResourceName());
+    Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
+    // FIXME: is this sufficient to get all consuming segments? what is someone went into error/offline?
+    for (Map.Entry<String, Map<String, String>> entry : mapFields.entrySet()) {
+      Collection<String> instanceStates = entry.getValue().values();
+      if (instanceStates.contains(RealtimeSegmentOnlineOfflineStateModel.CONSUMING)) {
+        consumingSegments.add(entry.getKey());
+      }
+    }
+    // TODO: check that targetNumReplicas and replicas in newPartitionAssignment match
```
mcvsubbu: not clear what the todo is?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,300 @@
...
+      Map<String, String> instanceStateMap = new HashMap<>();
+      for (String instance : partitionToInstanceMap.get(String.valueOf(partitionId))) {
+        instanceStateMap.put(instance, RealtimeSegmentOnlineOfflineStateModel.CONSUMING);
+      }
+      idealState.setInstanceStateMap(llcSegmentName.getSegmentName(), instanceStateMap);
+    }
+  }
+
+  // Keep only those segments that are LLC and in ONLINE state.
+  private void removeSegmentsNotBalanceable(Map<String, Map<String, String>> mapFields,
```
mcvsubbu: removeConsumingAndHistoricalSegments()?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+      }
+      if (!keep) {
+        removedEntries.add(entry);
+        it.remove();
+      }
+    }
+  }
+
+  protected int getKafkaPartitionCount(TableConfig tableConfig) {
```
mcvsubbu: Remove

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/RebalanceUserParams.java
```diff
@@ -0,0 +1,58 @@
...
+package com.linkedin.pinot.controller.helix.core.rebalance;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * Custom user config provided in rebalance, which will vary according to rebalance strategy
+ */
+public class RebalanceUserParams {
```
mcvsubbu: Can we use PropertiesConfiguration instead? Then we reduce one new class.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
npawar: Made this change

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/PartitionAssignment.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Class to represent a partition assignment
+ */
+public class PartitionAssignment {
```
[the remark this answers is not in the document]
npawar: Will do this in a future PR. Will introduce a PartitionKey class with partitionId and replicaId

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -2092,131 +2090,21 @@ public PinotResourceManagerResponse toggleInstance(String instanceName, boolean
...
-    }
+    TableConfig tableConfig = getTableConfig(rawTableName, tableType);
+    String tableNameWithType = tableConfig.getTableName();
+    IdealState idealState = _helixAdmin.getResourceIdealState(_helixClusterName, tableNameWithType);
+
+    RebalanceSegmentStrategy rebalanceSegmentsStrategy =
+        RebalanceSegmentStrategyFactory.getInstance().getRebalanceSegmentsStrategy(tableConfig);
+    PartitionAssignment newPartitionAssignment =
+        rebalanceSegmentsStrategy.rebalancePartitionAssignment(idealState, tableConfig, rebalanceUserParams);
+    IdealState newIdealState = rebalanceSegmentsStrategy.rebalanceIdealState(idealState, tableConfig,
```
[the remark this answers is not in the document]
npawar: TableConfig could be passed and used from the constructor. However, if someone were to create an object once, and then call rebalance multiple times, any change made to the table config in between will be ignored. So I think it is better if the TableConfig comes in with every call to rebalance, so that it will reflect the latest change requested

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
npawar: removed it

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/RebalanceUserParams.java
npawar: Changed.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/PartitionAssignment.java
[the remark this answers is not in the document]
snleee: I think I can extend `PartitionAssignment` class to represent `ReplicaGroupPartitionAssignment`. Or, we can also introduce `PartitionKey` and implement `ReplicaGroupPartitionKey` as Neha suggested.

pinot-common/src/main/java/com/linkedin/pinot/common/config/RealtimeTagConfig.java
```diff
@@ -20,49 +20,34 @@
...
-  public String getCompletedRealtimeServerTag() {
+  public String getCompletedServerTag() {
     return _completedRealtimeServerTag;
   }
 
-  public String getServerTenantName() {
-    return _serverTenant;
-  }
-
   public boolean isMoveCompletedSegments() {
```
mcvsubbu: suggest rename to isRelocateCompletedSegments, but your choice

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/partition/StreamPartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @VisibleForTesting
+  protected TableConfig getRealtimeTableConfig(String tableNameWithType) {
+    return ZKMetadataProvider.getTableConfig(_propertyStore, tableNameWithType);
+  }
+
+  /**
+   * Given map of table name to stream partition assignment, construct and write znodes to property store one by one
+   * @param newPartitionAssignment
```
mcvsubbu: pls add a comment that the key in the map should be raw table name, not tablename with type.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,333 @@
...
+    _helixAdmin = helixManager.getClusterManagmentTool();
+    _helixClusterName = helixManager.getClusterName();
+    _propertyStore = helixManager.getHelixPropertyStore();
+  }
+
+  /**
+   * Rebalance partition assignment only for realtime tables, and if includeConsuming=true
+   * @param idealState old ideal state
+   * @param tableConfig table config of table tor rebalance
+   * @param rebalanceUserConfig custom user configs for specific rebalance strategies
```
mcvsubbu: user params? user configs? your choice. I prefer params

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,333 @@
...
+            streamPartitionAssignmentGenerator.generatePartitionAssignment(tableConfig, numPartitions,
+                consumingInstances, Lists.newArrayList(tableNameWithType));
+        newPartitionAssignment = tableNameToStreamPartitionAssignmentMap.get(tableNameWithType);
+
+        boolean dryRun = rebalanceUserConfig.getBoolean(RebalanceUserConfigConstants.DRYRUN, DEFAULT_DRY_RUN);
+        if (!dryRun) {
+          LOGGER.info("Updating stream partition assignment for table {}", tableNameWithType);
+          streamPartitionAssignmentGenerator.writeStreamPartitionAssignment(tableNameToStreamPartitionAssignmentMap);
+        } else {
+          LOGGER.info("Dry run. Skip writing stream partition assignment to property store");
```
mcvsubbu: we should find a way to include new partition assignment in the output of dryrun

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -0,0 +1,333 @@
...
+      }
+    }
+
+    // update ideal state of consuming segments, based on new stream partition assignment
+    Map<String, List<String>> partitionToInstanceMap = newPartitionAssignment.getPartitionToInstances();
+    for (LLCSegmentName llcSegmentName : partitionIdToLatestSegment.values()) {
+      int partitionId = llcSegmentName.getPartitionId();
+      Map<String, String> instanceStateMap = new HashMap<>();
+      for (String instance : partitionToInstanceMap.get(String.valueOf(partitionId))) {
+        instanceStateMap.put(instance, RealtimeSegmentOnlineOfflineStateModel.CONSUMING);
```
mcvsubbu: you can add a comment here that it is possible that we turn an OFFLINE state into CONSUMING state in the process, but that should be OK since the only way a CONSUMING segment can become OFFLINE is when there are temporary or permanent issues consuming from the incoming event stream. If it turns back into CONSUMING, we will attempt to consume again (and may fail again if the fault is permanent)

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
npawar: The api has already been modified to send back partition assignment along with the ideal state

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/partition/StreamPartitionAssignmentGenerator.java
npawar: It is actually the other way. Everything here operates assuming tableNameWithType

PR #2513, opened 2018-02-20
Realtime servers missed controller leadership change

mcvsubbu:
In the recent re-factors of realtime server segment completion protocol with the server
the calls to refresh the leadership information was missed.

Adding it back.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
@@ -169,7 +169,12 @@ public class ServerSegmentCompletionProtocolHandler {
   private SegmentCompletionProtocol.Response sendRequest(String url) {
     try {
       String responseStr = _fileUploadDownloadClient.sendSegmentCompletionProtocolRequest(url, OTHER_REQUESTS_TIMEOUT);
-      return new SegmentCompletionProtocol.Response(responseStr);
+      SegmentCompletionProtocol.Response response = new SegmentCompletionProtocol.Response(responseStr);
+      LOGGER.info("Controller response {} for {}", response.toJsonString(), url);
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
```diff
@@ -169,7 +169,12 @@ private String createSegmentCompletionUrl(SegmentCompletionProtocol.Request requ
   private SegmentCompletionProtocol.Response sendRequest(String url) {
     try {
       String responseStr = _fileUploadDownloadClient.sendSegmentCompletionProtocolRequest(url, OTHER_REQUESTS_TIMEOUT);
-      return new SegmentCompletionProtocol.Response(responseStr);
+      SegmentCompletionProtocol.Response response = new SegmentCompletionProtocol.Response(responseStr);
+      LOGGER.info("Controller response {} for {}", response.toJsonString(), url);
+      if (response.getStatus().equals(SegmentCompletionProtocol.ControllerResponseStatus.NOT_LEADER)) {
+        ControllerLeaderLocator.getInstance().refreshControllerLeader();
```
jfim: Why is `sendRequest` responsible for invalidating the leader cache? Isn't the protocol handler responsible for handling network IO and the leader cache (which should be completely orthogonal to the network protocol itself) handled somewhere else?

mcvsubbu: The way it was done was that the protocol handler uses the leader locator to get the leader, and send the request. So, the protocol handler is the one that can recognize if the response is "not leader", so it informs the leader locator that its response was wrong, and it needs to refresh.

In the previous check-in, these lines got removed accidentally, so I had to put them back in as a hotfix.

We can discuss other implementations as applicable in this case.

PR #2514, opened 2018-02-20
[TE] frontend - Ttbach/bug fixes

ttbach:
Fixed 2 bugs:
- humanize number utils
- console erros in `rootcause-data-indicator`

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-data-indicator/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-data-indicator/component.js
@@ -39,7 +39,12 @@ export default Component.extend({
 
     fetch(`/rootcause/raw?framework=identity&urns=${selectedUrn}`)
       .then(checkStatus)
-      .then(res => setProperties(this, { maxTime: parseInt(res[0].attributes.maxTime[0], 10) }));
+      .then(res => {
+        let maxTime = res[0].attributes.maxTime;
+        if (maxTime) {
+          setProperties(this, { maxTime: parseInt(maxTime[0], 10) });
+        }
+      });
   },
 
   /**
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-data-indicator/component.js
```diff
@@ -39,7 +39,12 @@ export default Component.extend({
 
     fetch(`/rootcause/raw?framework=identity&urns=${selectedUrn}`)
       .then(checkStatus)
-      .then(res => setProperties(this, { maxTime: parseInt(res[0].attributes.maxTime[0], 10) }));
+      .then(res => {
```
cptran777: Not a fan of using super short variable names like `res` vs something clearer like `result` even if it is more concise. It's like coming in and being all "what is res? *stare* ooooh result"

thirdeye/thirdeye-frontend/app/pods/components/rootcause-data-indicator/component.js
```diff
@@ -39,7 +39,12 @@ export default Component.extend({
...
+        let maxTime = res[0].attributes.maxTime;
```
cptran777: Are we sure res will always be a defined array of objects with an attributes property? Because if not this may break.

thirdeye/thirdeye-frontend/app/utils/utils.js
```diff
@@ -32,7 +31,7 @@ export function checkStatus(response, mode = 'get', recoverBlank = false) {
  * Formatter for the human-readable floating point numbers numbers
  */
 export function humanizeFloat(f) {
-  if (isNone(f) || Number.isNaN(f)) { return '-'; }
+  if (typeof f !== 'number') { return '-'; }
```
cptran777: Instead of having a check for whether f is a number in every single util function, it might be more concise to just check for f's type first and if it's a number do these other functions or if not then you only have to write the check once. Not worth a refactor or anything, just a thought for future purposes where you find yourself writing a the same logic repeatedly

PR #2517, opened 2018-02-21
[TE] rootcause - clean up metrics view (remove cards)

apucher:
* remove card view from rootcause-metrics

* merge metrics-table and rootcause-metrics

* unified shortening of long numbers

* add/re-order columns


thirdeye/thirdeye-frontend/app/utils/utils.js
```diff
@@ -32,24 +32,54 @@ export function checkStatus(response, mode = 'get', recoverBlank = false) {
...
+    return `${f > 0 ? '+' : '-'}1000%+`;
+  }
 }
 
 /**
  * Formatter for human-readable entity scores with 2 decimals
  */
 export function humanizeScore(f) {
-  if (isNone(f) || Number.isNaN(f)) { return '-'; }
+  if (isNone(f) || typeof f !== 'number' || Number.isNaN(f)) { return '-'; }
```
ttbach: Since this check is repeated 3 times, let's extract this into another function, maybe call it `isNotANumber(f)` and reference it in these places.

thirdeye/thirdeye-frontend/app/utils/utils.js
```diff
@@ -32,24 +32,54 @@ export function checkStatus(response, mode = 'get', recoverBlank = false) {
  * Formatter for the human-readable floating point numbers numbers
  */
 export function humanizeFloat(f) {
```
ttbach: Can you include an `@example` here for clarity. It's not immediately understandable from the inline comments below.

thirdeye/thirdeye-frontend/app/utils/utils.js
```diff
@@ -32,24 +32,54 @@ export function checkStatus(response, mode = 'get', recoverBlank = false) {
...
+  return `${(f * Math.pow(10, shift)).toFixed(2)}${suffix}`;
 }
 
 /**
  * Formatter for the human-readable change values in percent with 1 decimal
  */
 export function humanizeChange(f) {
-  if (isNone(f) || Number.isNaN(f)) { return '-'; }
-  return `${f > 0 ? '+' : ''}${(Math.round(f * 1000) / 10.0).toFixed(1)}%`;
+  if (isNone(f) || typeof f !== 'number' || Number.isNaN(f)) { return '-'; }
```
ttbach: Since this check is repeated 3 times, let's extract this into another function, maybe call it `isNotANumber(f)` and reference it in these places.

thirdeye/thirdeye-frontend/app/pods/custom/metrics-table-offset/template.hbs
```diff
@@ -1,6 +1,7 @@
...
-</p>
-<p class="metrics-table--{{record.wo1w.direction}}">
+<span class="metrics-table__change--{{get record (concat column.propertyName ".direction")}}">
+  {{get record (concat column.propertyName ".change")}}
+</span>
+<br/>
+<span class="metrics-table__value">
   {{get record (concat column.propertyName ".value")}}
-</p>
+</span>
```
ttbach: Since these are meant to be displayed in 2 lines, it's more semantic to have `<p>` tags instead of `<span>` because `span` is for inline texts.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
```diff
@@ -341,117 +145,141 @@ export default Component.extend({
    * }
    */
   links: computed(
-    'urns',
+    'entities',
     function() {
-      const { urns, entities } = this.getProperties('urns', 'entities');
+      const { entities } = this.getProperties('entities');
```
ttbach: You know what I'm about to comment here 😄

thirdeye/thirdeye-frontend/app/pods/components/rootcause-data-indicator/component.js
```diff
@@ -39,7 +40,13 @@ export default Component.extend({
 
     fetch(`/rootcause/raw?framework=identity&urns=${selectedUrn}`)
       .then(checkStatus)
-      .then(res => setProperties(this, { maxTime: parseInt(res[0].attributes.maxTime[0], 10) }));
+      .then(res => {
+        if (_.isEmpty(res[0].attributes.maxTime)) {
```
ttbach: Are we sure `res` will always be an array of objects, and that it will always have an `attributes` property?

justYves: we could just use a try catch if response from the back end are inconsistent

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
```diff
@@ -341,117 +145,141 @@ export default Component.extend({
...
-          urlArr.push({
-            [urlLabel]: attributes[urlLabel][0] // each type should only have 1 url
+      filterPrefix(Object.keys(entities), 'thirdeye:metric:')
+        .forEach(urn => {
+          const attributes = entities[urn].attributes;
+          const { externalUrls = [] } = attributes;
+          let urlArr = [];
+
+          // Add the list of urls for each url type
+          externalUrls.forEach(urlLabel => {
```
justYves: looks like `reduce` might be a good candidate here since we have transforming n to 1

thirdeye/thirdeye-frontend/app/pods/components/rootcause-data-indicator/component.js
apucher: yes. it's generated by the RCA formatters by default

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
apucher: what? :D

thirdeye/thirdeye-frontend/app/styles/components/metrics-table.scss
```diff
@@ -3,18 +3,25 @@
...
+    }
+
+    &--negative {
+      color: $te-red;
+    }
   }
 
-  &--negative {
-    color: $te-red;
+  &__value {
```
justYves: empty css selector?

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
ttbach: `this.getProperties` for getting 1 property 😉

thirdeye/thirdeye-frontend/app/pods/custom/metrics-table-current/template.hbs
```diff
@@ -0,0 +1,6 @@
+<p class="metrics-table__change">
+  &nbsp;
+</p>
+<p class="metrics-table__value">
+  {{get record (concat column.propertyName ".value")}}
+</p>
```
ttbach: When does this custom template get used? It's not clear to me.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-metrics/component.js
apucher: yes. code moved only

thirdeye/thirdeye-frontend/app/styles/components/metrics-table.scss
apucher: fixed

thirdeye/thirdeye-frontend/app/pods/custom/metrics-table-current/template.hbs
apucher: for the 'current' offset

ttbach: Since this template doesn't differ much from the `metrics-table-offset` template, can we do a check for whether it's a `current` column via the `propertyName` of the table and use 1 custom template for both current and offset columns?

apucher: hmm. a state-check in a template? it's kind of a different thing - only displaying a value rather than value & change.

PR #2519, opened 2018-02-21
[TE] frontend/ ttbach/Decreased card spacing in rca

ttbach:
- Refactored `card-container` styling and extracted it into its own file and created `card-container--md` for smaller size for new rca page. Styles for default `card-container` is used for the rest of the use cases.
- Changed shared scss variable to comply with BEM (`$te-spacing-lg` ---> `$te-spacing--lg`)
- Decreased spacing between cards in rca page


ttbach:
@justYves It's not too much work, since right now, `card-container--medium` will only be used in the new rca page. My question is when do we use one over the other? Are we only creating the medium one temporarily until we make a global change, or are you saying both versions of `card-container` will be useful in the future?

ttbach:
@justYves As per our in person discussion, I'll refactor the `card-container` styling for only the rca page, so it doesn't have a global effect.

PR #2521, opened 2018-02-21
[TE] Allow user to specify dimension for baseline fetching

cyy0714:
Originally TE returns the baseline of dimensions underneath an anomaly function. This pr changes the endpoint, allowing users to specify dimensions to be fetching.


apucher:
what's the "mode" flag about?

cyy0714:
We have two ways to generate baseline: offline and online. Online is for generating the anomaly view, while offline is for use case like RCA. By default, we use offline.

PR #2523, opened 2018-02-22
Add 0 value for LLC_PARTITION_CONSUMING gauge whenever we stop consuming

npawar:


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -311,17 +311,21 @@ protected boolean consumeLoop() throws Exception {
             _kafkaStreamMetadata.getKafkaFetchTimeoutMillis());
         consecutiveErrorCount = 0;
       } catch (TimeoutException e) {
+        _serverMetrics.setValueOfTableGauge(_metricKeyName, ServerGauge.LLC_PARTITION_CONSUMING, 0);
```
snleee: If you make this line as a function, the code may look simpler since they are repeating a lot. `resetGuage(ServerGuage.LLC_PARTITION_CONSUMING)` This is just a suggestion.

PR #2525, opened 2018-02-23
[TE] Ignore completeness entries that does not need to be updated

cyenjung:
Reduce the entries of data completeness check to be updated by omitting the ones that have been updated previously. This change is to expect to reduce the runtime of cleanup of data completeness check and hence to alleviate the issue that worker threads are blocked by this task.


cyenjung:
It's not measure by how much time we could save. It's about whether our workers could still be working after days of execution. 

Current issue that we encountered is that every cleanup task of data completeness could take 3-4 hours because it needs to update 10s of thousands of entries. The task is scheduled every 15 mins. Eventually, most threads or all threads will be used to perform the cleanup task.

This code could alleviate the problem because if the entries that have been updated (i.e., isTimeOut=true), then it is omitted and hence reduce the overhead to communicate with the database.

cyy0714:
I see. Thanks for the explanation and the fix.

PR #2526, opened 2018-02-23
[TE] rootcause - advanced baselines

apucher:
Week-over-week baselines in ThirdEye are useful for ad-hoc data-exploration workflows in the RCA UI, but the process of correcting for past outliers and data errors is manual and cumbersome. This PR adds support for synthetic baselines computed on-demand from multiple week-over-week time series that are both, robust to outliers and user-intuitive in construction.

* synthetic baseline computation from multiple weekly offsets (including DST correction)

* centralized endpoint for the retrieval of metric representations (time series, aggregation, breakdown)

* integration with RCA UI

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/rootcause-timeseries-cache/service.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rootcause-timeseries-cache/service.js
@@ -1,8 +1,7 @@
 import Service from '@ember/service';
 import {
-  toAbsoluteRange,
-  toFilters,
-  toFilterMap
+  toMetricUrn,
+  toAbsoluteUrn
 } from 'thirdeye-frontend/utils/rca-utils';
 import { checkStatus } from 'thirdeye-frontend/utils/utils';
 import fetch from 'fetch';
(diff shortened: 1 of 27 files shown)


thirdeye/thirdeye-frontend/app/pods/rootcause-timeseries-cache/service.js
```diff
@@ -81,39 +77,23 @@ export default Service.extend({
...
 
-    const url = `/timeseries/query?metricIds=${metricId}&ranges=${range[0]}:${range[1]}&filters=${filterString}&granularity=${context.granularity}`;
+    const url = `/rootcause/metric/timeseries?urn=${metricUrn}&start=${range[0]}&end=${range[1]}&offset=${offset}&granularity=${granularity}&timezone=${timezone}`;
     return fetch(url)
       .then(checkStatus)
       .then(res => this._extractTimeseries(res, urn))
-      .then(res => offsetFunc(res))
       .then(res => this._complete(context, res))
-      .catch(error => this._handleError(urn, error));
+      // .catch(error => this._handleError(urn, error));
```
justYves: Any reason why we're not handling the error anymore?

apucher: that, sir, is an excellent catch

PR #2528, opened 2018-02-23
[TE] Self-Serve Onboard/Tuning flows: perf enhancements 02

newsummit:
This includes the following self-serve flow improvements:
- Added loading icons for transitions and table loads
- Standardizing data range picker action in manage.alert.explore
- Converting fetch chains to cancellable **concurrency tasks** to prevent transition lag for large datasets
- Resetting the time span or “duration” cache each time a new alert is selected
- Preventing display of erroneous performance numbers (long number edge cases)
- If we encounter an error during onboarding (resulting in an empty alert function), we now delete it.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/controller.js
@@ -238,44 +238,44 @@ export default Controller.extend({
       const mttdWeight = Number(extractSeverity(alertData, defaultSeverity));
       const convertedWeight = severityUnit === '%' ? mttdWeight * 100 : mttdWeight;
       const statsCards = [
-          {
-            title: 'Number of anomalies',
-            key: 'totalAlerts',
-            tooltip: false,
-            hideProjected: false,
-            text: 'Actual number of alerts received'
-          },
-          {
-            title: 'Response Rate',
-            key: 'responseRate',
-            units: '%',
-            tooltip: false,
-            hideProjected: true,
-            text: '% of anomalies that are reviewed.'
-          },
-          {
-            title: 'Precision',
-            key: 'precision',
-            units: '%',
-            tooltip: false,
-            text: 'Among all anomalies reviewed, the % of them that are true.'
-          },
-          {
-            title: 'Recall',
-            key: 'recall',
-            units: '%',
-            tooltip: false,
-            text: 'Among all anomalies that happened, the % of them detected by the system.'
-          },
-          {
-            title: `MTTD for > ${convertedWeight}${severityUnit} change`,
-            key: 'mttd',
-            units: 'hrs',
-            tooltip: false,
-            hideProjected: true,
-            text: `Minimum time to detect for anomalies with > ${convertedWeight}${severityUnit} change`
-          }
-        ];
+        {
+          title: 'Number of anomalies',
+          key: 'totalAlerts',
+          tooltip: false,
+          hideProjected: false,
+          text: 'Actual number of alerts received'
+        },
+        {
+          title: 'Response Rate',
+          key: 'responseRate',
+          units: '%',
+          tooltip: false,
+          hideProjected: true,
+          text: '% of anomalies that are reviewed.'
+        },
+        {
+          title: 'Precision',
+          key: 'precision',
+          units: '%',
+          tooltip: false,
+          text: 'Among all anomalies reviewed, the % of them that are true.'
+        },
+        {
+          title: 'Recall',
+          key: 'recall',
+          units: '%',
+          tooltip: false,
+          text: 'Among all anomalies that happened, the % of them detected by the system.'
+        },
+        {
+          title: `MTTD for > ${convertedWeight}${severityUnit} change`,
+          key: 'mttd',
+          units: 'hrs',
+          tooltip: false,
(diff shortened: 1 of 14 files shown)


thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/controller.js
```diff
@@ -696,15 +696,25 @@ export default Controller.extend({
      * @param {Object} rangeOption - the selected range object
      */
     onRangeOptionClick(rangeOption) {
```
newsummit: Making this action more consistent with the manage.alerts.tune action for the same component (date-range-picker)

thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/route.js
```diff
@@ -79,46 +77,6 @@ const newWowList = wowOptions.map((item) => {
   return { name: item, isActive: false };
 });
 
-/**
```
newsummit: Converting these fetching functions to concurrency tasks. (to be ported into services later on)

thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/route.js
```diff
@@ -379,27 +337,12 @@ export default Route.extend({
 
     // Cancel all pending concurrency tasks in controller
     if (isExiting) {
+      this.get('loadAnomalyData').cancelAll();
```
newsummit: I tested this locally and it seems to help with quicker transitions while large datasets are still loading.

thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/route.js
```diff
@@ -423,6 +366,77 @@ export default Route.extend({
     }
   },
 
+  /**
+   * Fetches all anomaly data for found anomalies - downloads all 'pages' of data from server
+   * in order to handle sorting/filtering on the entire set locally. Start/end date are not used here.
+   * @param {Array} anomalyIds - list of all found anomaly ids
+   * @returns {RSVP promise}
+   */
+  fetchCombinedAnomalies: task(function * (anomalyIds) {
```
newsummit: From what I have researched, I can't yet use async/await here because they have not been integrated with concurrency due to "cancellability". Feel free to correct me if I'm wrong.

thirdeye/thirdeye-frontend/app/utils/manage-alert-utils.js
```diff
@@ -16,7 +16,9 @@ export function formatEvalMetric(metric, isPercentage = false) {
   const multiplier = isPercentage ? 100 : 1;
   const convertedNum = metric * multiplier;
   const formattedNum = isWhole ? convertedNum : convertedNum.toFixed(1);
-  return isFinite(metric) ? formattedNum : shown;
+  let displayNum = isFinite(metric) ? formattedNum : shown;
+  // Prevent meaninglessly large numbers
+  return (Number(displayNum) > 10000) ? 'N/A' : displayNum;
```
newsummit: Arbitrary eliminator of excessively long performance numbers, which would be errors.

thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/route.js
```diff
@@ -423,6 +366,77 @@ export default Route.extend({
...
+   * @param {Array} anomalyData - array of processed anomalies
+   * @returns {RSVP promise}
+   */
+  fetchCombinedAnomalyChangeData: task(function * (anomalyData) {
+    yield timeout(300);
+    let promises = [];
+
+    anomalyData.forEach((anomaly) => {
+      let id = anomaly.anomalyId;
+      promises[id] = this.get('fetchAnomalyEntity').perform(`/anomalies/${id}`);
```
justYves: are we assigning at index id here? Seems like an object would be more correct. Any thought?

thirdeye/thirdeye-frontend/app/pods/manage/alert/tune/route.js
```diff
@@ -354,6 +355,55 @@ export default Route.extend({
...
 
+  /**
+   * This concurrency task fetches anomaly performance metrics and data for all anomalies which
+   * would not be included in the notification set under the user-selected tuning settings.
+   * @method triggerTuningSequence
+   * @param {Object} configObj - the user-selected type and value of tuning severity thresholds
+   * @return {undefined}
+   */
+  triggerTuningSequence: task(function * (configObj) {
+    try {
```
justYves: do we need the try around the whole function here? ideally, this can only error out at the fetching step

thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/route.js
newsummit: This promise resolves with the anomaly Id as associative array so that I can filter it if needed later.

thirdeye/thirdeye-frontend/app/pods/manage/alert/tune/route.js
newsummit: True - no need to wrap the entire function. I'll move it down to where the fetching begins.

PR #2530, opened 2018-02-23
[TE] Include number of user report anomalies to total response rates

cecilynie:
In previous response rate calculation, we are including number of user report anomalies. As communicate with UX design, user report should be part of total responses.


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detector/email/filter/PrecisionRecallEvaluator.java
```diff
@@ -109,7 +109,8 @@ public double getResponseRate() {
   }
 
   public int getTotalResponses() {
```
cyy0714: For future developer, it is good if we can leave a comment, indicating what they expect to receive from getTotalResponses.

cecilynie: added. thanks

PR #2537, opened 2018-02-26
Support on-the-fly metrics aggregation during realtime consumption.

mayankshriv:
One of the reasons that serving latency of realtime nodes may be worse
than the offline counterparts is that data can (and almost always is)
aggregated in the offline pipeline, reducing the number of rows to be
processed.

For cases with high Kafka event rate and with duplicate dimension
values, this performance delta can be huge, impacting the overall
latency.

This PR addresses this problem by allowing for updating of metrics for
rows with same dimension values. The implementation is as follows:

1. We maintain an off-heap HashMap with FixedIntArray (containing
dictIds of dimension columns) as key, and docId as the value.

2. When a new row comes in, we check if the dimension values are present
in the map:
   If yes, then we fetch the docId from the map, and update the metric
columns.
   - If not, then we index the new row.
   This features requires metric columns to use no-dictionary indexing.
The reason being updating of metrics may leave unsused values in
dictionary that need to be cleaned up before segment flush. Using
no-dictionary indexing eliminates the problem completely.
   Also, this feature is only supported for LLC consumer.

3. The feature is off-by-default and can be enabled with a config
   - TableConfig.IndexingConfg.updateMetrics

4. Added unit tests for the new feature.

5. Performance gains measured by this feature using real production
data:
   - An average of 4x reduction in numRows. Although, more common
     dimension combinations (that are also frequently queried) see upto
1000x reduction in number of rows.
   - Realtime nodes now perform almost at par with offline nodes for
     query serving.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
@@ -43,6 +43,7 @@ public class IndexingConfig {
   private List<String> _onHeapDictionaryColumns;
   private StarTreeIndexSpec _starTreeIndexSpec;
   private SegmentPartitionConfig _segmentPartitionConfig;
+  private boolean _aggregateMetrics;
 
   public List<String> getInvertedIndexColumns() {
     return _invertedIndexColumns;
@@ -136,8 +137,16 @@ public class IndexingConfig {
     _segmentPartitionConfig = config;
   }
 
+  public void setAggregateMetrics(boolean value) {
+    _aggregateMetrics = value;
+  }
+
   public SegmentPartitionConfig getSegmentPartitionConfig() {
     return _segmentPartitionConfig;
+  }
+
+  public boolean getAggregateMetrics() {
+    return _aggregateMetrics;
   }
 
   @Override
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -18,6 +18,7 @@ package com.linkedin.pinot.core.data.manager.realtime;
 
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.Uninterruptibles;
+import com.linkedin.pinot.common.Utils;
 import com.linkedin.pinot.common.config.IndexingConfig;
 import com.linkedin.pinot.common.config.SegmentPartitionConfig;
 import com.linkedin.pinot.common.config.TableConfig;
@@ -168,6 +169,7 @@ public class LLRealtimeSegmentDataManager extends RealtimeSegmentDataManager {
   private volatile long _currentOffset;
   private volatile State _state;
   private volatile int _numRowsConsumed = 0;
+  private volatile int _numRowsIndexed = 0; // Can be different from _numRowsConsumed when metrics update is enabled.
   private volatile int consecutiveErrorCount = 0;
   private long _startTimeMs = 0;
   private final String _segmentNameStr;
@@ -201,7 +203,7 @@ public class LLRealtimeSegmentDataManager extends RealtimeSegmentDataManager {
   private Logger segmentLogger = LOGGER;
   private final String _tableStreamName;
   private final PinotDataBufferMemoryManager _memoryManager;
-  private AtomicLong _lastUpdatedRawDocuments = new AtomicLong(0);
+  private AtomicLong _lastUpdatedRowsIndexed = new AtomicLong(0);
   private final String _instanceId;
   private final ServerSegmentCompletionProtocolHandler _protocolHandler;
   private final long _consumeStartTime;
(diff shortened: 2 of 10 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
```diff
@@ -43,6 +43,7 @@
   private List<String> _onHeapDictionaryColumns;
   private StarTreeIndexSpec _starTreeIndexSpec;
   private SegmentPartitionConfig _segmentPartitionConfig;
+  private boolean _updateMetrics;
```
mcvsubbu: aggregateMetrics?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -1002,6 +1008,7 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
       }
     }
 
+    realtimeSegmentConfigBuilder.setUpdateMetrics(indexingConfig.getUpdateMetrics());
```
mcvsubbu: can u pls move this up to line 989, so we have the builder in one place?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -1062,17 +1069,26 @@ private void makeConsumerWrapper(String reason) {
...
-    _serverMetrics.addValueToTableGauge(_tableName, ServerGauge.DOCUMENT_COUNT, (currentRawDocs - _lastUpdatedRawDocuments
-        .get()));
-    _lastUpdatedRawDocuments.set(currentRawDocs);
+
+    // When updating of metrics is enabled, numRowsIndexed can be <= numRowsConsumed. This is because when in this
+    // case when a new row with existing dimension combination comes in, we find the existing row and update metrics.
+
+    // Number of rows indexed should be used for DOCUMENT_COUNT metric, and also for segment flush. Whereas,
+    // Number of rows consumed should be used for consumption metric.
+    long rowsIndexed = _numRowsIndexed - _lastUpdatedRawDocuments.get();
```
mcvsubbu: So, lastUpdatedRawDocuments now has the number of rows indexed (instead of raw docs). You may want to rename it as such.  Perhaps this var can be a volatile int instead of AtomicLong ... not sure why it is that way.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -1062,17 +1069,26 @@ private void makeConsumerWrapper(String reason) {
...
     // Log every minute or 100k events
-    if (now - prevTime > TimeUnit.MINUTES.toMillis(TIME_THRESHOLD_FOR_LOG_MINUTES) || rowsConsumed >= MSG_COUNT_THRESHOLD_FOR_LOG) {
-      segmentLogger.info("Consumed {} events from (rate:{}/s), currentOffset={}, numRowsSoFar={}", rowsConsumed,
-            (float) (rowsConsumed) * 1000 / (now - prevTime), _currentOffset, _numRowsConsumed);
+    if (now - prevTime > TimeUnit.MINUTES.toMillis(TIME_THRESHOLD_FOR_LOG_MINUTES)
+        || rowsConsumed >= MSG_COUNT_THRESHOLD_FOR_LOG) {
+      segmentLogger.info(
+          "Consumed {} events from (rate:{}/s), currentOffset={}, numRowsConsumedSoFar={}, numRowsIndexedSoFar={}",
+          rowsConsumed, (float) (rowsConsumed) * 1000 / (now - prevTime), _currentOffset, _numRowsConsumed,
+          rowsIndexed);
```
mcvsubbu: s/rowsIndexed/_numRowsIndexed/ ? As I understand it, rowsIndexed is the number of rows indexed since the last update, but _numRowsIndexed is the number of rows indexed since start of segment.

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentImpl.java
```diff
@@ -260,9 +299,36 @@ public boolean index(GenericRow row) {
...
-    // Update number of document indexed at last to make the latest record queryable
-    return _numDocsIndexed++ < _capacity;
+  private boolean updateMetrics(GenericRow row, int docId) {
+    for (FieldSpec metricSpec : _schema.getMetricFieldSpecs()) {
+      String column = metricSpec.getName();
+      Object value = row.getValue(column);
+      Preconditions.checkState(metricSpec.isSingleValueField(), "Multivalued metrics cannot be updated.");
+      FixedByteSingleColumnSingleValueReaderWriter indexReaderWriter =
+          (FixedByteSingleColumnSingleValueReaderWriter) _indexReaderWriterMap.get(column);
+      Preconditions.checkState(_dictionaryMap.get(column) == null, "Updating metrics not support with dictionary.");
```
mcvsubbu: supported?

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mayankshriv: Renamed (will revisit the atomicLong vs volatile later).

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mayankshriv: Ran out of direct memory while trying to allocate 214721752 bytes

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
mayankshriv: Yes, done.

pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
mayankshriv: Renamed.

PR #2543, opened 2018-02-27
[TE] dashboard - anomalies summary endpoint

apucher:
Clean endpoint for retrieval of anomaly summaries for display in a user dashboard. Supports time range, owner, and application predicates. Several potential performance bottlenecks remain for future work.

* anomaly summary query endpoint (start, end, application, owner, limit)

* anomaly summary container

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/UserDashboardResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/UserDashboardResource.java
@@ -3,7 +3,7 @@ package com.linkedin.thirdeye.dashboard.resources.v2;
 import com.google.common.base.Preconditions;
 import com.linkedin.thirdeye.constant.AnomalyFeedbackType;
 import com.linkedin.thirdeye.constant.AnomalyResultSource;
-import com.linkedin.thirdeye.dashboard.resources.v2.userdashboard.AnomalySummary;
+import com.linkedin.thirdeye.dashboard.resources.v2.pojo.AnomalySummary;
 import com.linkedin.thirdeye.datalayer.bao.AlertConfigManager;
 import com.linkedin.thirdeye.datalayer.bao.AnomalyFunctionManager;
 import com.linkedin.thirdeye.datalayer.bao.MergedAnomalyResultManager;
@@ -60,7 +60,9 @@ public class UserDashboardResource {
    *     "current" : 1213.0,
    *     "baseline" : 550.0,
    *     "feedback" : "NO_FEEDBACK",
+   *     "metricId" : 12346
    *     "metric" : "page_views",
+   *     "functionId" : 12347
    *     "functionName" : "page_views_monitoring"
    *     },
    *     ...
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/UserDashboardResource.java
```diff
@@ -0,0 +1,217 @@
...
+
+    List<Predicate> predicates = new ArrayList<>();
+
+    // TODO support index select on user-reported anomalies
+//    predicates.add(Predicate.OR(
+//        Predicate.EQ("notified", true),
+//        Predicate.EQ("anomalyResultSource", AnomalyResultSource.USER_LABELED_ANOMALY)));
+
+
+    // application (indirect)
```
cyy0714: For the part of joining application with the anomaly function, can you move this part to an util or a static method, as we will use this method for many other endpoints.

apucher: This is a simple code fragment. Moving this to a separate method creates way more clutter than it saves

cyy0714: I can understand, but this code segment is shared by multiple methods, for the sake of code re-usability, or easier for us to improve the performance in the future. In my opinion, I think we should better utilize the code segment.

apucher: Where would you put this method, and what would be it's signature?

cyy0714: I'm thinking of having this method in ApplicationManager, where we can fetch the list of anomaly function ids from the DAO.

apucher: If this were to go into a DAO, then it would be the AlertConfig DAO. That wouldn't be the place where anyone would look for it though - i.e. this is a classic case of cross-entity business logic.

cecilynie: In this sense how do we come up with a generic way to fetch anomalies under an application. Noting this can be executed drastically in the dashboard that we are going to build. How can we make sure avoid code duplications and improve execution efficiency?

apucher:
Folks, we're blocked on dashboard development until this gets merged. If you feel this should be refactored, please do it in a follow-up PR.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/UserDashboardResource.java
cyy0714: Please move this code segment to an util method. I know this code segment may be simple. I still hope the implementation can be simplified.

apucher: I gladly invite your contribution to this PR. Please refactor the code as you see fit.

cyy0714: Please change the code followed by https://github.com/linkedin/pinot/pull/2650.

PR #2546, opened 2018-02-27
Add ControllerMeter NOT_LEADER in the controller

npawar:
Recording the value in the metrics registry, whenever a not leader response is detected by the controller

@mcvsubbu @jfim Please take a look, and let me know if this is sufficient/excessive (I'm not sure if I've gone overboard)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -156,6 +157,8 @@ private void stopThread() {
   public void runSegmentMetrics() {
     if (!_pinotHelixResourceManager.isLeader()) {
       LOGGER.info("Skipping Segment Status check, not leader!");
+      _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_NOT_LEADER, 1L);
```
mcvsubbu: No, we don't need to do this for retention manager, segment status checker, we only need it when we return NOT_LEADER as a part of LLC consumption. thanks,

PR #2549, opened 2018-02-28
Added ReplicaGroupPartitionAssignment

snleee:
1. Added ReplicaGroupPartitionAssignment and its generator
2. Replace all the usages of PartitionToReplicaGroupMappingZKMetadata
3. Moved PartitionAssignment to pinot-common because it needs to be
   used in pinot-broker package for routing


pinot-common/src/main/java/com/linkedin/pinot/common/partition/PartitionAssignment.java
```diff
@@ -26,9 +26,9 @@
  */
 public class PartitionAssignment {
 
-  private String _tableName;
+  protected String _tableName;
   // TODO: include replica in the key, by creating a class PartitionKey to encapsulate partitionId and replicaNum
```
npawar: Should we remove this TODO, since you chose the approach of extending PartitionAssignment instead of creating a PartitionKey class? Or do you think it might still be a good idea to create a PartitionKey?

pinot-common/src/main/java/com/linkedin/pinot/common/partition/ReplicaGroupPartitionAssignment.java
```diff
@@ -0,0 +1,134 @@
...
+    return partitions.size();
+  }
+
+  /**
+   * Compute total number of replica groups in the mapping.
+   *
+   * @return the number of replica groups
+   */
+  public int getNumReplicaGroups() {
+    Set<String> partitions = new HashSet<>();
```
npawar: should this be Set<String> replicaGroups instead of partitions?

pinot-common/src/main/java/com/linkedin/pinot/common/partition/ReplicaGroupPartitionAssignment.java
```diff
@@ -0,0 +1,134 @@
...
+    String key = createMappingKey(partition, replicaGroup);
+    addPartition(key, instances);
+  }
+
+  /**
+   * Get all instances.
+   *
+   * @return Set of all instances for this table
+   */
+  public List<String> getAllInstances() {
```
npawar: This might be useful to have in the super class instead of here

pinot-common/src/main/java/com/linkedin/pinot/common/partition/ReplicaGroupPartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+   *
+   * @param tableNameWithType: Name of table
+   * @param tableConfig: Configuration for table
+   * @return Replica group partition assignment from the given configuration
+   */
+  public ReplicaGroupPartitionAssignment buildReplicaGroupPartitionAssignment(String tableNameWithType,
+      TableConfig tableConfig, List<String> servers) {
+
+    // Fetch information required to build the mapping table from the table configuration.
+    ReplicaGroupStrategyConfig replicaGroupStrategyConfig = tableConfig.getValidationConfig().getReplicaGroupStrategyConfig();
```
npawar: Can replicaGroupStrategyConfig be null? I see null checking only for segmentConfig in the place where buildReplicaGroupPartitionAssignment is called

pinot-common/src/main/java/com/linkedin/pinot/common/partition/ReplicaGroupPartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,121 @@
...
+    // which is equivalent to have the same partition number for all segments (i.e. 1 partition).
+    int numPartitions = 1;
+    if (partitionColumn != null) {
+      numPartitions = tableConfig.getIndexingConfig().getSegmentPartitionConfig().getNumPartitions(partitionColumn);
+    }
+    int numReplicas = tableConfig.getValidationConfig().getReplicationNumber();
+    int numServers = servers.size();
+
+    // Enforcing disjoint server sets for each replica group.
+    if (numInstancesPerPartition * numReplicas > numServers) {
```
npawar: numPartitions and numInstancesPerPartition terminology is a little confusing. 
numPartitions is referring to the partitions on a column right? so 0_* 1_* means 0, 1 are the partitions i.e.numPartitions=2. 
numInstancesPerPartition seems to refer to 0_0 0_1 .. 0_n as separate partitions. So it is actually referring to numInstancesPerReplica

Can we change one of them? I see that numInstancesPerPartition is already in the tableConfig. So maybe we can rename numPartitions to something like numColumnPartitions

pinot-common/src/main/java/com/linkedin/pinot/common/partition/PartitionAssignment.java
snleee: I think this approach would be more simple. If we introduced `PartitionKey`, we will need to make the current existing realtime partition assignment compatible (e.g. 1 -> 0_1).

pinot-common/src/main/java/com/linkedin/pinot/common/partition/ReplicaGroupPartitionAssignment.java
snleee: fixed.

pinot-common/src/main/java/com/linkedin/pinot/common/partition/ReplicaGroupPartitionAssignmentGenerator.java
snleee: `numInstancesPerReplicaGroup` will be better name but as you said, it's already in table config. I changed the name for `numPartitions`

pinot-common/src/main/java/com/linkedin/pinot/common/partition/ReplicaGroupPartitionAssignment.java
snleee: moved

pinot-common/src/main/java/com/linkedin/pinot/common/partition/ReplicaGroupPartitionAssignment.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+    return getInstancesListForPartition(key);
+  }
+
+  /**
+   * Set instances of a replica group for a partition.
+   *
+   * @param partition Partition number
+   * @param replicaGroup Replica group number
+   * @param instances Instances that belongs to the gibn partition and replica group
```
npawar: gibn -> given

PR #2551, opened 2018-02-28
Adding response code for segments uploaded during minion task

jenniferdai:


pinot-minion/src/main/java/com/linkedin/pinot/minion/executor/BaseSegmentConversionExecutor.java
```diff
@@ -153,8 +153,9 @@ public void executeTask(@Nonnull PinotTaskConfig pinotTaskConfig) throws Excepti
           @Override
           public Boolean call() throws Exception {
             try {
-              fileUploadDownloadClient.uploadSegment(new URI(uploadURL), segmentName, convertedTarredSegmentFile,
+              int responseCode = fileUploadDownloadClient.uploadSegment(new URI(uploadURL), segmentName, convertedTarredSegmentFile,
                   httpHeaders, parameters, FileUploadDownloadClient.DEFAULT_SOCKET_TIMEOUT_MS);
+              LOGGER.info("Response code {} received for file {} from uri {}", responseCode, convertedTarredSegmentDir.getName(), uploadURL);
```
snleee: Why not use `segmentName`?

Jackie-Jiang: The name of "convertedTarredSegmentDir" will always be "convertedTarredSegmentDir", which is not useful. You might want to log the table name and segment name instead?

PR #2553, opened 2018-02-28
Fix PinotSegmentReatletResource APIs when raw table name is provided and table type is not specified

Jackie-Jiang:
Added a helper method getExistingTableNamesWithType to get all existing tables with the given table name and table type

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
@@ -32,8 +32,10 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -291,42 +281,73 @@ public String getCrcForAllSegments(
     return getAllCrcMetadataForTable(tableName);
   }
 
-  //////////////////////////////////////////////////////////////////////////////////////
+  private List<String> getExistingTableNamesWithType(@Nonnull String tableName,
```
snleee: Can you add the documentation for this helper function?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -481,26 +487,23 @@ private JSONArray getAllSegmentsMetadataForTable(String tableName, CommonConstan
    * XXX
```
snleee: I know that you didn't add this but I think we should remove this?

Jackie-Jiang: Done

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: Done

PR #2555, opened 2018-02-28
[TE] implemented toggle between legacy and new rootcause

justYves:
### What's new: 
- moved `te-toggle`'s css into its own file
- removed the `Root cause Analysis (beta)` link in favor of the current one
- added toggle between legacy / new rootcause 

### Screenshot: 
screen shot 2018-02-28 at 3 29 52 pm
screen shot 2018-02-28 at 3 21 18 pm
screen shot 2018-02-28 at 3 20 02 pm

### Tests: 
- basic unit test from rootcause -> rca (no tests for rca.details as the page will be deprecated and mocking all the endpoints will be too time consuming)
- Visual Sanity check locally

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/rca/details/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rca/details/controller.js
@@ -316,7 +316,7 @@ export default Controller.extend({
     },
 
     /**
-     * Resets anoamly and investigations regions
+     * Resets anomaly and investigations regions
      */
     resetAnalysisDates() {
       let offset = 1;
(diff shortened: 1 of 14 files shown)


thirdeye/thirdeye-frontend/app/pods/rca/details/controller.js
```diff
@@ -341,7 +341,14 @@ export default Controller.extend({
...
       }
-    }
+    },
 
+    /**
+     * Redirects the user to the new rootcause
+     * @param {Number} metricId
+     */
+    onLegacyToggle(metricId) {
+      this.send('transitionToNewRca', metricId);
```
cptran777: I think using bubbling through this.send is not very performant. Would it not be easier to just call that function directly?

justYves: It's a route action. Do we have a better alternative? @cptran777

PR #2556, opened 2018-02-28
Fix the bug of data table deserialization failure for empty string

Jackie-Jiang:
When hitting the end of the input stream, ByteArrayInputStream.read(b) returns -1 instead of 0 when length of b is 0 (not the same as the java doc).
Added decodeString() method to handle this behavior.
Added empty string test to DataTableSerDeTest and default column integration test

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/common/datatable/DataTableImplV2.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/common/datatable/DataTableImplV2.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
 
 
 public class DataTableImplV2 implements DataTable {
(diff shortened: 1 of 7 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/common/datatable/DataTableImplV2.java
```diff
@@ -166,61 +165,54 @@ public DataTableImplV2(@Nonnull ByteBuffer byteBuffer)
...
-      metadata.put(new String(keyBytes, UTF_8), new String(valueBytes, UTF_8));
+      return metadata;
     }
+  }
 
-    return metadata;
+  private static String decodeString(DataInputStream dataInputStream) throws IOException {
+    int length = dataInputStream.readInt();
+    if (length == 0) {
+      return "";
```
mayankshriv: could we use private static final?

Jackie-Jiang: Fixed with StringUtils.EMPTY

PR #2559, opened 2018-03-01
Add class SimpleHttpResponse to wrap status code and response

Jackie-Jiang:
Let all APIs in FileUploadDownloadClient to return SimpleHttpResponse for flexibility

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
@@ -54,6 +54,11 @@ import org.apache.http.util.EntityUtils;
 import org.json.JSONObject;
 
 
+/**
+ * The <code>FileUploadDownloadClient</code> class provides methods to upload schema/segment, download segment or send
+ * segment completion protocol request through HTTP/HTTPS.
+ */
+@SuppressWarnings("unused")
 public class FileUploadDownloadClient implements Closeable {
   public static class CustomHeaders {
     public static final String UPLOAD_TYPE = "UPLOAD_TYPE";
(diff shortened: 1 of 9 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
```diff
@@ -217,26 +221,30 @@ private static void setTimeout(RequestBuilder requestBuilder, int socketTimeoutM
...
   }
 
-  private int sendRequest(HttpUriRequest request) throws Exception {
+  private SimpleHttpResponse sendRequest(HttpUriRequest request) throws IOException, HttpErrorStatusException {
     try (CloseableHttpResponse response = _httpClient.execute(request)) {
-      StatusLine statusLine = response.getStatusLine();
-      int statusCode = statusLine.getStatusCode();
-      if (statusCode >= 400) {
+      int statusCode = response.getStatusLine().getStatusCode();
+      if (statusCode >= 300) {
```
snleee: 3xx is redirection. Can you add some documentation why we would want to regard this as error?

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
```diff
@@ -403,28 +403,43 @@ public int sendSegmentJson(URI uri, String jsonString, @Nullable List<Header> he
...
+  /**
+   * Send segment completion protocol request.
+   *
+   * @param uri URI
+   * @param socketTimeoutMs Socket timeout in milliseconds
+   * @return Response
+   * @throws IOException
+   * @throws HttpErrorStatusException
+   */
+  public SimpleHttpResponse sendSegmentCompletionProtocolRequest(URI uri, int socketTimeoutMs) throws Exception {
```
snleee: `IOException, HttpErrorStatusException` instead of `Exception`

Jackie-Jiang: Wow, good catch!

pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
Jackie-Jiang: Anything other than success response are count as error. (In practice, we will never get 3xx)

PR #2560, opened 2018-03-01
Handle incomplete commit scenarios in realtime segment metadata commit

npawar:
In commitSegmentMetadata in PinotLLCRealtimeSegmentManager, we update zookeeper in 3 steps.
     
Step 1: Update PROPERTYSTORE to change the old segment metadata status to DONE
Step 2: Update PROPERTYSTORE to create the new segment metadata with status IN_PROGRESS
Step 3: Update IDEALSTATES to include new segment in CONSUMING state, and change old segment to ONLINE state.
     
The controller may fail between these three steps.
We have 2 chances to correct the state created by these failures. One is when a new controller takes over as leader. Other is via ValidationManager. 
They needs to check the following:
     
If segment commit fails between step-1 and step-2:
Check whether there are any segments in the PROPERTYSTORE with status DONE, but no new segment in status IN_PROGRESS, and hence the status of the segment in the IDEALSTATE is still CONSUMING
     
If segment commit fails between step-2 and-3:
Check whether there are any segments in PROPERTYSTORE with status IN_PROGRESS, that are not accounted for in idealState. If so, it should create the new segments in idealState.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -519,18 +519,21 @@ public class PinotLLCRealtimeSegmentManager {
 
   }
 
-  protected boolean writeSegmentsToPropertyStore(String oldZnodePath, String newZnodePath, ZNRecord oldRecord, ZNRecord newRecord,
-      final String realtimeTableName, int expectedVersion) {
-
-    boolean success = _propertyStore.set(oldZnodePath, oldRecord, expectedVersion, AccessOption.PERSISTENT);
+  protected boolean writeSegmentToPropertyStore(String znodePath, ZNRecord znRecord, final String realtimeTableName,
+      int expectedVersion) {
+    boolean success = _propertyStore.set(znodePath, znRecord, expectedVersion, AccessOption.PERSISTENT);
     if (!success) {
-      LOGGER.error("Failed to write old segments to property store for table {}. Expected zookeeper version number: {}",
+      LOGGER.error("Failed to write segment to property store for table {}. Expected zookeeper version number: {}",
           realtimeTableName, expectedVersion);
       return false;
     }
-    success = _propertyStore.set(newZnodePath, newRecord, AccessOption.PERSISTENT);
+    return success;
+  }
+
+  protected boolean writeSegmentToPropertyStore(String znodePath, ZNRecord znRecord, final String realtimeTableName) {
+    boolean success = _propertyStore.set(znodePath, znRecord, AccessOption.PERSISTENT);
     if (!success) {
-      LOGGER.error("Failed to write new segments to property store for table {}.", realtimeTableName);
+      LOGGER.error("Failed to write segment to property store for table {}.", realtimeTableName);
     }
     return success;
   }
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -605,25 +608,109 @@ private static File convertURIToSegmentLocation(String segmentLocation) {
...
+    // When multiple segments of the same table complete around the same time it is possible that
+    // the idealstate udpate fails due to contention. We serialize the updates to the idealstate
+    // to reduce this contention. We may still contend with RetentionManager, or other updates
+    // to idealstate from other controllers, but then we have the retry mechanism to get around that.
+    // hash code can be negative, so make sure we are getting a positive lock index
+
+    String newSegmentNameStr = newLLCSegmentName.getSegmentName();
+    List<String> newInstances = partitionAssignment.getInstancesListForPartition(String.valueOf(partitionId));
+
+    int lockIndex = (realtimeTableName.hashCode() & Integer.MAX_VALUE) % NUM_LOCKS;
```
kishoreg: why do we need this lock?

npawar:
Not sure which lines you're referring to, there is no reformatting done as part of this PR

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
npawar: I haven't introduced the locking part. I have simply split the commitSegmentMetadata method into 3 parts 
1. update old segment metadata 
2. create new segment metadata 
3. update ideal state. 
I did that, so that it is easier to call just some of the parts from the code which tries to correct incomplete commits (completeCommittingSegment)

kishoreg:
I commented on acquired lock while updating idealstate but then realized that code already existed in the previous version but is shown as newly added code in the diff.  Will give it another try.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -519,18 +519,21 @@ private SegmentPartitionMetadata getPartitionMetadataFromSegmentMetadata(Segment
 
   }
 
-  protected boolean writeSegmentsToPropertyStore(String oldZnodePath, String newZnodePath, ZNRecord oldRecord, ZNRecord newRecord,
-      final String realtimeTableName, int expectedVersion) {
-
-    boolean success = _propertyStore.set(oldZnodePath, oldRecord, expectedVersion, AccessOption.PERSISTENT);
+  protected boolean writeSegmentToPropertyStore(String znodePath, ZNRecord znRecord, final String realtimeTableName,
+      int expectedVersion) {
+    boolean success = _propertyStore.set(znodePath, znRecord, expectedVersion, AccessOption.PERSISTENT);
```
kishoreg: can we verify that propertystore handles runtime exceptions as well

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -519,18 +519,21 @@ private SegmentPartitionMetadata getPartitionMetadataFromSegmentMetadata(Segment
...
-      LOGGER.error("Failed to write old segments to property store for table {}. Expected zookeeper version number: {}",
+      LOGGER.error("Failed to write segment to property store for table {}. Expected zookeeper version number: {}",
           realtimeTableName, expectedVersion);
       return false;
     }
-    success = _propertyStore.set(newZnodePath, newRecord, AccessOption.PERSISTENT);
+    return success;
+  }
+
+  protected boolean writeSegmentToPropertyStore(String znodePath, ZNRecord znRecord, final String realtimeTableName) {
```
kishoreg: can we call the other method with expectedVerion set to -1.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -605,25 +608,109 @@ private static File convertURIToSegmentLocation(String segmentLocation) {
...
+     * If the controller fails after step-3, we are fine because the idealState has the new segments.
+     * If the controller fails before step-1, the server will see this as an upload failure, and will re-try.
+     */
+
+    final LLCSegmentName committingLLCSegmentName = new LLCSegmentName(committingSegmentNameStr);
+
+    // Step-1
+    boolean success = updateOldSegmentMetadataZNRecord(realtimeTableName, committingLLCSegmentName, nextOffset);
+    if (!success) {
+      return false;
```
kishoreg: what happens on the server side, if this method returns false? Will the servers retry?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -862,44 +927,73 @@ private void completeCommittingSegments() {
     }
   }
 
-  protected void completeCommittingSegments(String realtimeTableName) {
+  public void completeCommittingSegments(String realtimeTableName) {
```
kishoreg: did we make this public for testing

kishoreg:
can you please add details on how it gets fixed when the new controller takes over and via validation manager. Can both of them try to fix it at the same time? An orthogonal question is what happens if there was a failure after let's say step2 and the server retries to commit.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -519,18 +519,21 @@ private SegmentPartitionMetadata getPartitionMetadataFromSegmentMetadata(Segment
...
-  protected boolean writeSegmentsToPropertyStore(String oldZnodePath, String newZnodePath, ZNRecord oldRecord, ZNRecord newRecord,
-      final String realtimeTableName, int expectedVersion) {
-
-    boolean success = _propertyStore.set(oldZnodePath, oldRecord, expectedVersion, AccessOption.PERSISTENT);
+  protected boolean writeSegmentToPropertyStore(String znodePath, ZNRecord znRecord, final String realtimeTableName,
+      int expectedVersion) {
+    boolean success = _propertyStore.set(znodePath, znRecord, expectedVersion, AccessOption.PERSISTENT);
     if (!success) {
-      LOGGER.error("Failed to write old segments to property store for table {}. Expected zookeeper version number: {}",
+      LOGGER.error("Failed to write segment to property store for table {}. Expected zookeeper version number: {}",
```
mcvsubbu: pls add the znode path in the log so we know what the segment is about, thanks

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: suspect this is for testing so that we can indepdently override each method. will leave it to neha

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -519,18 +519,21 @@ private SegmentPartitionMetadata getPartitionMetadataFromSegmentMetadata(Segment
...
     }
-    success = _propertyStore.set(newZnodePath, newRecord, AccessOption.PERSISTENT);
+    return success;
+  }
+
+  protected boolean writeSegmentToPropertyStore(String znodePath, ZNRecord znRecord, final String realtimeTableName) {
+    boolean success = _propertyStore.set(znodePath, znRecord, AccessOption.PERSISTENT);
     if (!success) {
-      LOGGER.error("Failed to write new segments to property store for table {}.", realtimeTableName);
+      LOGGER.error("Failed to write segment to property store for table {}.", realtimeTableName);
```
mcvsubbu: pls include znode path as well in the log, thanks

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: the servers handle receiving false just fine. they retry to the appropriate lead controller again.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -605,25 +608,109 @@ private static File convertURIToSegmentLocation(String segmentLocation) {
...
+     * Check whether there are any segments in PROPERTYSTORE with status IN_PROGRESS, that are not accounted for in idealState.
+     * If so, it should create the new segments in idealState.
+     *
+     * If the controller fails after step-3, we are fine because the idealState has the new segments.
+     * If the controller fails before step-1, the server will see this as an upload failure, and will re-try.
+     */
+
+    final LLCSegmentName committingLLCSegmentName = new LLCSegmentName(committingSegmentNameStr);
+
+    // Step-1
```
mcvsubbu: can u push step 1 after determining if the partition assignment is present? basically, check all pre-conditions before atempting to execute the transaction. thanks

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: there is a comment "when multiple segments of the same table..." . to justify the lock. It was added more than a year ago. without the lock, we were seeing each thread contending to update the idealstate, and backing off exponentially. Never saw the problem after the lock.

We can potentially batch the updates, but that is an optimization...

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: No, for calls from validation manager

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -862,44 +927,73 @@ private void completeCommittingSegments() {
...
+
     for (Map.Entry<Integer, MinMaxPriorityQueue<LLCSegmentName>> entry : partitionToLatestSegments.entrySet()) {
+      final int partitionId = entry.getKey();
+      List<String> newInstances = partitionAssignment.getInstancesListForPartition(Integer.toString(partitionId));
       final LLCSegmentName segmentName = entry.getValue().pollFirst();
       final String segmentId = segmentName.getSegmentName();
-      final int partitionId = entry.getKey();
+      LLCRealtimeSegmentZKMetadata segmentZKMetadata = segmentIdToMetadataMap.get(segmentId);
+
+      // controller failed between step-1 and step-2 of commitSegmentMetadata.
```
mcvsubbu: can u move these coments to be after line 970?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -862,44 +927,73 @@ private void completeCommittingSegments() {
...
+          final int newSeqNum = segmentName.getSequenceNumber() + 1;
+          final long now = System.currentTimeMillis();
+          LLCSegmentName newLLCSegmentName = new LLCSegmentName(segmentName.getTableName(), partitionId, newSeqNum, now);
+
+          LOGGER.info("{}: Creating new segment metadata for {}", realtimeTableName, newLLCSegmentName.getSegmentName());
+          createNewSegmentMetadataZNRecord(realtimeTableName, newLLCSegmentName, segmentZKMetadata.getEndOffset(),
+              partitionAssignment);
+          updateIdealState(realtimeTableName, newInstances, segmentId, newLLCSegmentName.getSegmentName());
+        }
+      }
```
mcvsubbu: else?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -862,44 +927,73 @@ private void completeCommittingSegments() {
...
+
+      // controller failed between step-2 and step-3 of commitSegmentMetadata.
```
mcvsubbu: pls move these comments to below line 989

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -241,7 +241,7 @@ void validateLLCSegments(final String realtimeTableName, TableConfig tableConfig
       if (_autoCreateOnError) {
         _llcRealtimeSegmentManager.createConsumingSegment(realtimeTableName, nonConsumingKafkaPartitions, llcSegments,
             tableConfig);
-        _llcRealtimeSegmentManager.completeCommittingSegments(realtimeTableName, llcSegments);
+        _llcRealtimeSegmentManager.completeCommittingSegments(realtimeTableName);
```
mcvsubbu: Nice

mcvsubbu:
Excellent comments in code, and thanks for adding the unit tests as well

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
kishoreg: I thought you mentioned that all retries will fail because metadata is already updated. In this case IdealState will not be updated until validation manager kicks in. Until then the server will keep receiving false and keep retrying? 
. @npawar I am trying to understand the flow. Feel free to merge

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
npawar: Yes, in the tests the write method checks for an increasing version number, and it will break if I send -1. Will try to see if I can easily modify test to handle this.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
npawar: fixed

PR #2561, opened 2018-03-01
[TE] Return no content with empty time line view upon exception

cyy0714:
When generating the anomaly baselines, we face data issues during calculation, where there is short of data on some dimensions. To proper handle the exception and notifying front-end, the pr return an empty time series view with NO_CONTENT status back to FE.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
@@ -762,23 +762,28 @@ public class AnomalyResource {
     }
 
     // Get the anomaly time lines view of the anomaly function on each dimension
-    HashMap<String, AnomalyTimelinesView> dimensionMapAnomalyTimelinesViewMap = new HashMap<>();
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/AnomalyResource.java
```diff
@@ -762,23 +762,28 @@ public Response getTimeSeriesAndBaselineData(@PathParam("id") long functionId,
...
+    try {
+      if (mode.equalsIgnoreCase("ONLINE")) {
+        // If online, use the monitoring time window to query time range and generate the baseline
+        anomalyTimelinesView =
+            anomaliesResource.getTimelinesViewInMonitoringWindow(anomalyFunctionSpec, datasetConfigDTO, startTime,
+                endTime, dimensionsToBeEvaluated);
+      } else {
+        // If offline, request baseline in user-defined data range
+        List<Pair<Long, Long>> dataRangeIntervals = new ArrayList<>();
+        dataRangeIntervals.add(new Pair<Long, Long>(endTime.getMillis(), endTime.getMillis()));
```
apucher: ist endTime/endTime intentional?

cyy0714: Yes. As we don't need the testing window for that.

PR #2564, opened 2018-03-01
Ttbach/create stats-cards component for alerts dashboard

ttbach:
- Created `stats-cards` component
- Added mock data for anomaly performance by application
- Created a util to convert a float to percent
- Fixed failing test for toggle to rca legacy page
screen shot 2018-03-01 at 5 38 46 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/styles/shared/_styles.scss
+++ b/thirdeye/thirdeye-frontend/app/styles/shared/_styles.scss
@@ -356,7 +356,6 @@ body {
 
   &__container {
     display: flex;
-    justify-content: flex-start;
     margin-top: 15px;
     clear: both;
   }
--- a/thirdeye/thirdeye-frontend/app/styles/shared/_styles.scss
+++ b/thirdeye/thirdeye-frontend/app/styles/shared/_styles.scss
@@ -356,7 +356,7 @@ body {
 
   &__container {
     display: flex;
-    justify-content: flex-end;
+    justify-content: flex-start;
     margin-top: 15px;
     clear: both;
   }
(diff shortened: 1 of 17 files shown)


thirdeye/thirdeye-frontend/app/styles/shared/_styles.scss
```diff
@@ -356,7 +356,7 @@ body {
 
   &__container {
     display: flex;
-    justify-content: flex-end;
+    justify-content: flex-start;
```
justYves: any reason we need to specifically set this? default value is `flex-start`

thirdeye/thirdeye-frontend/tests/integration/pods/components/stats-cards/component-test.js
```diff
@@ -0,0 +1,67 @@
...
+  this.setProperties({ stats });
+
+  this.render(
+    hbs`{{stats-cards
+          stats=stats}}`
+  );
+
+  // Testing titles of all cards
+  assert.equal(
+    this.$(TITLE).get(0).innerText,
```
justYves: nit: each time we do `$()`, we're essentially using jquery to grab a dom object (not performant). 
Let's refactor this so that we only do that once per selector
```
const $title = this.$(TITLE);
const $description = this.$(DESCRIPTION)
const $number = this.$(NUMBER)

// and in your assert :
assert.equal($title.get(1).innerText .....

```

thirdeye/thirdeye-frontend/app/styles/shared/_styles.scss
ttbach: I'll remove this line then. Didn't realize `flex-start` is the default. Checked the UI, and it looks fine after removal.

PR #2566, opened 2018-03-02
[TE] datasource - filter pre-aggregated dimensions by default

apucher:
ThirdEye already supports pre-computed data cubes for non-additive metrics in Pinot. However, ThirdEye still exposes the dummy values used for pre-aggregation to the user as valid dimension and filter values. This PR excludes dummy values from filter sets and result sets - making them fully transparent.


cyenjung:
They are valid dimensions actually. For instance, Sitespeed use this dimension 'all' as the pre-aggregated value. Since the metric is not additive, the only way we could see the aggregated values for all dimensions is to select the dimension value 'all'. In short, the value is not a dummy value.

apucher:
Yes. When the user does not select any filters, then Thirdeye already requests the data for the "all" dimension value for all dimensions. If a filter is applied for a specific dimension, then ThirdEye requests the specific dimension for the selected filter, and "all" for any remaining dimensions.

This PR simply removes the dimension value "all" from to available options for user selection and display, i.e. it does not change TE behavior. As before, the absence of a filter value for a given dimension leads to the use of the pre-cubed "all" for this dimension.

PR #2574, opened 2018-03-02
Ttbach/redirect link

ttbach:
- Added a redirect link to each anomaly table
- Added more anomalies to mock data
screen shot 2018-03-02 at 3 35 57 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/mirage/fixtures/applicationAnomalies.js
+++ b/thirdeye/thirdeye-frontend/app/mirage/fixtures/applicationAnomalies.js
@@ -29,6 +29,62 @@ export default [
     metric: "Metric Name 1",
     functionName: "Alert Name 1"
   }, {
+    id: 2,
+    start: 12345,
+    end: 23456,
+    dimensions: {
+      country: "US",
+      pageName: "some_page"
+    },
+    severity: 0.86,
+    current: 1444,
+    baseline: 1000,
+    feedback: "Yes, but New Trend",
+    metric: "Metric Name 1",
+    functionName: "Alert Name 1"
+  }, {
+    id: 2,
+    start: 12345,
+    end: 23456,
+    dimensions: {
+      country: "US",
+      pageName: "some_page"
+    },
+    severity: 0.86,
+    current: 1444,
+    baseline: 1000,
+    feedback: "Yes, but New Trend",
+    metric: "Metric Name 1",
+    functionName: "Alert Name 1"
+  }, {
+    id: 2,
+    start: 12345,
+    end: 23456,
+    dimensions: {
+      country: "US",
+      pageName: "some_page"
+    },
+    severity: 0.86,
+    current: 1444,
+    baseline: 1000,
+    feedback: "Yes, but New Trend",
+    metric: "Metric Name 1",
+    functionName: "Alert Name 1"
+  }, {
+    id: 2,
+    start: 12345,
+    end: 23456,
+    dimensions: {
+      country: "US",
+      pageName: "some_page"
+    },
+    severity: 0.86,
+    current: 1444,
+    baseline: 1000,
+    feedback: "Yes, but New Trend",
+    metric: "Metric Name 1",
+    functionName: "Alert Name 1"
+  }, {
     id: 3,
     start: 12345,
     end: 23456,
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-frontend/app/mirage/fixtures/applicationAnomalies.js
```diff
@@ -26,7 +27,68 @@ export default [
     current: 1444,
     baseline: 1000,
     feedback: "Yes, but New Trend",
-    metric: "Metric Name 1",
+    metricName: "Metric Name 1",
```
justYves: This is starting to become pretty big. not needed in this pr, but we should start considering using scenarios/factories

thirdeye/thirdeye-frontend/app/styles/components/te-anomaly-table.scss
```diff
@@ -145,4 +145,15 @@
   &--performance {
     margin-top: 30px;
   }
+
+  &__redirect-link {
+    display: flex;
+    justify-content: center;
+    border: solid 1px $te-light-grey;
+    margin-top: -21px;
```
justYves: I'd rather not use magic numbers with margin/padding for positioning. 
What is the intention here? Let's discuss this on Monday

thirdeye/thirdeye-frontend/app/mirage/fixtures/applicationAnomalies.js
ttbach: I'll add a `TODO` to refactor this in another PR.

thirdeye/thirdeye-frontend/app/styles/components/te-anomaly-table.scss
ttbach: The intention is for the `See More` row to be adjacent to the bottom of the table, as opposed to having a gap between the two. I tried setting the table's margin and padding and border bottom to be `0px`, but there was still some gap that I'm not sure what is causing it. We can talk about this more in person.

thirdeye/thirdeye-frontend/app/pods/home/controller.js
```diff
@@ -4,6 +4,13 @@ import { computed, get } from '@ember/object';
 
 export default Controller.extend({
 
+  /**
+   * Overrides ember-models-table's css classes
+   */
+  classes: {
+    table: 'table-bordered table-condensed te-anomaly-table--no-margin'
```
aaronucsd: Is there a reason why we can not do this in the CSS vs here in the Controller? Maybe in the `te-anomaly-table.scss` file? The reason it's not ideal to put css in JS because any updates done in CSS might not catch the ones in JS classes definitions. If we want specific styles for this component or page only, we can assign a specific class name and use that to style in the CSS. Also for searching it's hard to do in both CSS and JS files for any style maintenance. I could be wrong if this is best practices.

snleee:
@ttbach Can you make sure to pass travis-ci test before merge the commit from the next time? This commit pushed an empty file `thirdeye/thirdeye-frontend/app/pods/components/ts` and this caused all the travis tests to fail due to license check. I removed the file from https://github.com/linkedin/pinot/pull/2584

ttbach:
@snleee Oops, apologies from that. Will do next time

thirdeye/thirdeye-frontend/app/pods/home/controller.js
ttbach: Based on the documentation for `ember-models-table` (http://onechiporenko.github.io/ember-models-table/v.1/#/), in order to override some default css, I'd have to pass custom classes to the component. It is possible to do in the CSS through, but @justYves has done this in the controller previously here, so I figured it's acceptable: https://github.com/justYves/pinot/blob/4fc67f78c74fe12f9bce95f46609c0930b599aca/thirdeye/thirdeye-frontend/app/pods/components/modals/entity-mapping-modal/component.js#L270. 

Maybe @justYves can also weigh in on this also?

aaronucsd: Sorry it was late last night i meant not CSS file. But in the HBS. As it's best to keep the separation of concerns with the definitions in CSS and usage in HBS/Template files vs JS. It's not a big thing, I felt it's better if we apply so many classes to a component to do in the template vs the JS. Someone can do a clean up of a class and miss the JS files vs HBS files. If we do this in the JS, it's best to consolidate to a single BEM class perhap.

PR #2580, opened 2018-03-05
Introduce abstraction of stream, to be able to operate on any stream, not just kafka

npawar:
This is just the first pass. A complete abstraction will be completed in multiple iterations.


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/CombinedStreamProviderConfig.java
```diff
@@ -36,7 +35,7 @@
 import static com.linkedin.pinot.common.utils.EqualityUtils.isSameReference;
 
 
-public class KafkaHighLevelStreamProviderConfig implements StreamProviderConfig {
+public class CombinedStreamProviderConfig implements StreamProviderConfig {
```
mcvsubbu: This is a kafka specific class, right? Why call it a generic name?

npawar: Reverted the renaming

PR #2582, opened 2018-03-06
[TE] Authenticate users with multiple LDAP domain names

cyenjung:
The LDAP authenticator now try out multiple combinations of username and the given domain names of the system.

The authentication flow is improved as follows:
1. If user's name contains domain name or the system doesn't have any given domain names, then use the username as is.
2. Else, try out all combinations of username and the given domain names of the system.

Tested on local dashboard.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auth/ThirdEyeAuthenticatorLdap.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auth/ThirdEyeAuthenticatorLdap.java
@@ -4,9 +4,12 @@ import com.google.common.base.Optional;
 import io.dropwizard.auth.AuthenticationException;
 import io.dropwizard.auth.Authenticator;
 import java.util.Hashtable;
+import java.util.List;
 import javax.naming.Context;
 import javax.naming.NamingException;
 import javax.naming.directory.InitialDirContext;
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auth/ThirdEyeAuthenticatorLdap.java
```diff
@@ -29,48 +35,122 @@ public ThirdEyeAuthenticatorLdap(String domainSuffix, String ldapUrl) {
...
-        } catch (NamingException e) {
-          LOG.error("Could not authenticate {} with LDAP", principalName, e);
+        // Attempt ldap authentication with the following steps:
+        // 1. If user's name contains domain name or the system doesn't have any given domain names, then
+        //    use the username as is.
+        // 2. Else, try out all combinations of username and the given domain names of the system.
+        AuthenticationResult authenticationResult = new AuthenticationResult();
+        if (principalName.contains("@") || CollectionUtils.isEmpty(domainSuffix)) {
+          env.put(Context.SECURITY_PRINCIPAL, principalName);
+          authenticate(env, authenticationResult);
```
apucher: AuthenticationResult should be a return value

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auth/ThirdEyeAuthenticatorLdap.java
```diff
@@ -29,48 +36,155 @@ public ThirdEyeAuthenticatorLdap(String domainSuffix, String ldapUrl) {
...
+            if (authenticationResults.isAuthenticated()) {
+              break;
+            }
+          }
+        }
+
+        if (authenticationResults.isAuthenticated()) {
+          ThirdEyePrincipal principal = new ThirdEyePrincipal();
+          principal.setName(env.get(Context.SECURITY_PRINCIPAL));
+          LOG.info("Successfully authenticated {} with LDAP", principalName);
```
kishoreg: will be good to know which one was used to authenticate

cyenjung: See https://github.com/linkedin/pinot/pull/2591

PR #2584, opened 2018-03-06
Removed an empty file that is causing license check failure

snleee:


mayankshriv:
Curious how this got pushed in the first place. Was the commit merged even with building failing on branch? Or is there a gap in where the license check does not trigger in TE build? If latter, we should identify the gap and fix it.

snleee:
@mayankshriv Travis failed for https://github.com/linkedin/pinot/pull/2574. We will need to be extra  careful to make all travis test before we merge the code.

PR #2589, opened 2018-03-06
[TE] frontend - Ttbach/add application dropdown

ttbach:
- Added application title and dropdown 
screen shot 2018-03-06 at 1 45 48 pm


thirdeye/thirdeye-frontend/app/pods/home/controller.js
```diff
@@ -32,5 +43,17 @@ export default Controller.extend({
...
+
+  actions: {
+
+    /**
+     * Sets the selected application property based on user selection
+     * @param {Object} selectedApplication - object that represents selected application
+     * @return {undefined}
+     */
+    selectApplication(selectedApplication) {
+      setProperties(this, { selectedApplication });
```
aaronucsd: Should we use only `set` here if it's only for one prop?

thirdeye/thirdeye-frontend/app/pods/home/controller.js
```diff
@@ -11,6 +11,17 @@ export default Controller.extend({
...
   },
 
+  /**
+   * Default application for the application dropdown
+   * @type {Object} - the first application object from a list of applications
+   */
+  selectedApplication: computed(
+    'model.applications',
+    function() {
+      return this.get('model.applications')[0];
```
aaronucsd: I feel this name could be better named perhaps. Maybe something like `defaultApplication`, as it's just happens to be the selected as well. And i'm not sure we need to make this a computed if it's returning the first one anyway. Are we assuming here `model.applications` will update?

thirdeye/thirdeye-frontend/app/pods/home/controller.js
ttbach: I figured it would be shorter than: 
`set (this, 'selectedApplication', this.selectedApplication)`

What do you think?

thirdeye/thirdeye-frontend/app/pods/home/controller.js
ttbach: As per our in-person conversation, I have changed this to set this variable in the route's `setupController` method.

thirdeye/thirdeye-frontend/app/pods/home/route.js
```diff
@@ -96,13 +96,14 @@ export default Route.extend({
...
-  setupController(controller) {
+  setupController(controller, model) {
     this._super(...arguments);
 
     controller.setProperties({
       columns,
       metricList: this.getMetrics(),
-      alertList: this.getAlerts()
+      alertList: this.getAlerts(),
+      defaultApplication: model.applications[0]
```
aaronucsd: perfect! As controller itself being a singleton (it does not get called over and over like the route) is not ideal for setting anything dynamic and having a need of model itself. The `setupController` is primarily for setting up some controller context dynamically. And in this case, it is a good usage of   accessing the model and setting the controller's `defaultApplication` property.  Thanks for doing this!

PR #2592, opened 2018-03-07
[TE] Self-Serve Api import 1

newsummit:
Collecting self-serve endpoints into master import file.

In second commit, moved API file under /utils and removed 'swagger' placeholders in jsDocs.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/api/self-serve.js
+++ b/thirdeye/thirdeye-frontend/app/api/self-serve.js
@@ -1,0 +1,65 @@
+/**
+ * GET request for all alert subscription groups
+ * @see {@link namepathOrURL|link text}
+ */
+const allConfigGroups = '/thirdeye/entity/ALERT_CONFIG';
+
+/**
+ * GET request for all alert application groups
+ * @see {@link namepathOrURL|link text}
+ */
+const allApplications = '/thirdeye/entity/APPLICATION';
+
+/**
+ * GET request for job status (a sequence of events including create, replay, autotune)
+ * Requires 'jobId' query param
+ * @see {@link namepathOrURL|link text}
+ */
+const jobStatus = jobId => `/detection-onboard/get-status?jobId=${jobId}`;
+
+/**
+ * POST request to create alert container in DB
+ * Requires 'name' query param
+ * @param {String} newAlertName: Name of new alert function
+ * @see {@link namepathOrURL|link text}
+ */
+const createAlertFunction = newAlertName => `/function-onboard/create-function?name=${newAlertName}`;
+
+/**
+ * POST request to update an existing alert function with properties payload
+ * Requires 'payload' object in request body and 'jobName' in query param
+ * @param {String} jobName: unique name of new job for alert setup task
+ * @see {@link namepathOrURL|link text}
+ */
+const updateAlertFunction = jobName => `/detection-onboard/create-job?jobName=${jobName}`;
+
+/**
+ * DELETE request to remove an alert function from the DB
+ * Requires query param 'id'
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/api/self-serve.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * GET request for all alert subscription groups
+ * @see {@link namepathOrURL|link text}
```
newsummit: These links will be replaced once we know where the swagger doc will live.

thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/route.js
```diff
@@ -453,9 +453,8 @@ export default Route.extend({
    * @param {Array} anomalyIds - the IDs of anomalies that have been reported for this alert.
    * @return {undefined}
    */
-  loadAnomalyData: task(function * (anomalyIds) {
-    yield timeout(300);
-    const hasDimensions = this.currentModel.exploreDimensions && this.currentModel.exploreDimensions.length;
+  loadAnomalyData: task(function * (anomalyIds, exploreDimensions) {
```
justYves: Does this still need to be a task since you remove the `yield` keyword?

thirdeye/thirdeye-frontend/app/api/self-serve.js
aaronucsd: I like how we will link it to Swagger. That is very cool indeed!

aaronucsd: This is a great idea Steve. Thanks for doing it! However, it all `const` or static definitions. Can we find a more formal folder where we keep all global consts? like say `app/utils/api/self-serve.js`
i think i saw `utils/constant.js` there too. SO maybe good to group all kind of constant or static helper files there? Or where we keep `common` files of this type.

thirdeye/thirdeye-frontend/app/api/self-serve.js
```diff
@@ -0,0 +1,65 @@
...
+  jobStatus,
+  createAlertFunction,
+  updateAlertFunction,
+  deleteAlertFunction
+};
+
+export default {
+  general,
+  onboard
+};
```
aaronucsd: I see what you did here. So by default we get all the endpoints on import or selectively choose to import individual groups like`general` or `onboard` seperately`. Works for me.  However, I wonder we should be more explicit with the export names to tie with the `self-serve` api file name itself. Cause if i use `general` or `onboard` in the calling file, i might not have context if `general` ties to `self-serve` api or maybe it conflict with a later api with the same export name. Just a thought.

thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/route.js
aaronucsd: +1

thirdeye/thirdeye-frontend/app/api/self-serve.js
newsummit: Yes, I want a better naming convention for this. Agree.

thirdeye/thirdeye-frontend/app/pods/manage/alert/explore/route.js
newsummit: Yes - it does yield for child tasks and a data processing function to complete.

thirdeye/thirdeye-frontend/app/api/self-serve.js
newsummit: Yes, I like that as a location for it.

PR #2596, opened 2018-03-07
Change AvroUtils to not infer field type by field name

Jackie-Jiang:
1. When field type is not specified, always count it as dimension
2. Change AvroSchemaToPinotSchema tool to print simplified schema without redundant fields

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/util/AvroUtils.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/util/AvroUtils.java
@@ -16,14 +16,21 @@
 package com.linkedin.pinot.core.util;
 
 import com.google.common.base.Preconditions;
-
 import com.linkedin.pinot.common.data.DimensionFieldSpec;
 import com.linkedin.pinot.common.data.FieldSpec;
 import com.linkedin.pinot.common.data.MetricFieldSpec;
 import com.linkedin.pinot.common.data.Schema;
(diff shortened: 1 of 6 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/util/AvroUtils.java
```diff
@@ -16,138 +16,120 @@
...
-          pinotSchema.addField(new MetricFieldSpec(fieldName, dataType));
-          break;
-        case TIME:
-          Preconditions.checkState(isSingleValueField, "Unsupported multi-value for time field.");
-          pinotSchema.addField(new TimeFieldSpec(field.name(), dataType, timeUnit));
-          break;
-        default:
-          throw new UnsupportedOperationException("Unsupported field type: " + fieldType + " for field: " + fieldName);
+      if (fieldTypeMap == null) {
+        pinotSchema.addField(new DimensionFieldSpec(fieldName, dataType, isSingleValueField));
```
mayankshriv: Can we log a message here, so user knows we are treating these columns as dimensions?

Jackie-Jiang: This happens only when we call it with null fieldTypeMap (API: getPinotSchemaFromAvroDataFile(@Nonnull File avroDataFile)), and the java doc describes the behavior.

PR #2601, opened 2018-03-08
[TE] Compress anomaly timelines view when the size is too large

cyy0714:
As the anomaly timelines view is also stored along with the anomaly result, we occasionally see exception of json_value overflow. Main reason is because of the size of anomaly timelines view stored in the system. To tackle this problem, we compress the view and save more space .


cyenjung:
Could you give the overview of the method that is used to compress the timeline view?

PR #2604, opened 2018-03-08
Realtime Segment Relocation Manager

npawar:
As part of our ongoing efforts to support multitenant realtime, we have introduced separation of consuming servers and completed servers. The consuming servers will be responsible for hosting the CONSUMING segments, and the completed servers will be responsible for the ONLINE (completed) segments. When a segment transitions from CONSUMING to ONLINE, it however remains on the same server where it was consuming.
Introducing a service, which will periodically scan ideal state of all realtime tables, and move segments which are completed, from the consuming servers to the completed servers. The  tags to use for relocation will be provided by the tenant config (soon to be introduced), which will specify the realtime consuming tag and the realtime completed tag.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
@@ -41,6 +41,7 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final String RETENTION_MANAGER_FREQUENCY_IN_SECONDS = "controller.retention.frequencyInSeconds";
   private static final String VALIDATION_MANAGER_FREQUENCY_IN_SECONDS = "controller.validation.frequencyInSeconds";
   private static final String STATUS_CHECKER_FREQUENCY_IN_SECONDS = "controller.statuschecker.frequencyInSeconds";
+  private static final String RELOCATION_MANAGER_FREQUENCY_IN_SECONDS = "controller.relocation.frequencyInSeconds";
   private static final String STATUS_CHECKER_WAIT_FOR_PUSH_TIME_IN_SECONDS = "controller.statuschecker.waitForPushTimeInSeconds";
   private static final String SERVER_ADMIN_REQUEST_TIMEOUT_SECONDS = "server.request.timeoutSeconds";
   private static final String SEGMENT_COMMIT_TIMEOUT_SECONDS = "controller.realtime.segment.commit.timeoutSeconds";
@@ -58,6 +59,7 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final int DEFAULT_RETENTION_CONTROLLER_FREQUENCY_IN_SECONDS = 6 * 60 * 60; // 6 Hours.
   private static final int DEFAULT_VALIDATION_CONTROLLER_FREQUENCY_IN_SECONDS = 60 * 60; // 1 Hour.
   private static final int DEFAULT_STATUS_CONTROLLER_FREQUENCY_IN_SECONDS = 5 * 60; // 5 minutes
+  private static final int DEFAULT_RELOCATION_MANAGER_FREQUENCY_IN_SECONDS = 60 * 60; // 1 hour
   private static final int DEFAULT_STATUS_CONTROLLER_WAIT_FOR_PUSH_TIME_IN_SECONDS = 10 * 60; // 10 minutes
   private static final long DEFAULT_EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT_MILLIS = 120_000L; // 2 minutes
   private static final int DEFAULT_SERVER_ADMIN_REQUEST_TIMEOUT_SECONDS = 30;
@@ -263,6 +265,17 @@ public class ControllerConf extends PropertiesConfiguration {
 
   public void setStatusCheckerFrequencyInSeconds(int statusCheckerFrequencyInSeconds) {
     setProperty(STATUS_CHECKER_FREQUENCY_IN_SECONDS, Integer.toString(statusCheckerFrequencyInSeconds));
+  }
+
+  public int getRelocationManagerFrequencyInSeconds() {
+    if (containsKey(RELOCATION_MANAGER_FREQUENCY_IN_SECONDS)) {
+      return Integer.parseInt((String) getProperty(RELOCATION_MANAGER_FREQUENCY_IN_SECONDS));
+    }
+    return DEFAULT_RELOCATION_MANAGER_FREQUENCY_IN_SECONDS;
+  }
+
+  public void setRelocationManagerFrequencyInSeconds(int relocationManagerFrequencyInSeconds) {
+    setProperty(RELOCATION_MANAGER_FREQUENCY_IN_SECONDS, Integer.toString(relocationManagerFrequencyInSeconds));
   }
 
   public int getStatusCheckerWaitForPushTimeInSeconds() {
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
@@ -41,7 +41,7 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final String RETENTION_MANAGER_FREQUENCY_IN_SECONDS = "controller.retention.frequencyInSeconds";
   private static final String VALIDATION_MANAGER_FREQUENCY_IN_SECONDS = "controller.validation.frequencyInSeconds";
   private static final String STATUS_CHECKER_FREQUENCY_IN_SECONDS = "controller.statuschecker.frequencyInSeconds";
-  private static final String REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES = "controller.segment.relocation.frequencyInMinutes";
+  private static final String REALTIME_SEGMENT_RELOCATOR_FREQUENCY = "controller.realtime.segment.relocator.frequency";
   private static final String STATUS_CHECKER_WAIT_FOR_PUSH_TIME_IN_SECONDS = "controller.statuschecker.waitForPushTimeInSeconds";
   private static final String SERVER_ADMIN_REQUEST_TIMEOUT_SECONDS = "server.request.timeoutSeconds";
   private static final String SEGMENT_COMMIT_TIMEOUT_SECONDS = "controller.realtime.segment.commit.timeoutSeconds";
@@ -59,7 +59,7 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final int DEFAULT_RETENTION_CONTROLLER_FREQUENCY_IN_SECONDS = 6 * 60 * 60; // 6 Hours.
   private static final int DEFAULT_VALIDATION_CONTROLLER_FREQUENCY_IN_SECONDS = 60 * 60; // 1 Hour.
   private static final int DEFAULT_STATUS_CONTROLLER_FREQUENCY_IN_SECONDS = 5 * 60; // 5 minutes
-  private static final int DEFAULT_REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES = 60; // 1 hour
+  private static final String DEFAULT_REALTIME_SEGMENT_RELOCATOR_FREQUENCY = "1h"; // 1 hour
   private static final int DEFAULT_STATUS_CONTROLLER_WAIT_FOR_PUSH_TIME_IN_SECONDS = 10 * 60; // 10 minutes
   private static final long DEFAULT_EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT_MILLIS = 120_000L; // 2 minutes
   private static final int DEFAULT_SERVER_ADMIN_REQUEST_TIMEOUT_SECONDS = 30;
@@ -267,16 +267,15 @@ public class ControllerConf extends PropertiesConfiguration {
     setProperty(STATUS_CHECKER_FREQUENCY_IN_SECONDS, Integer.toString(statusCheckerFrequencyInSeconds));
   }
 
-  public int getRelocationManagerFrequencyInMinutes() {
-    if (containsKey(REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES)) {
-      return Integer.parseInt((String) getProperty(REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES));
+  public String getRealtimeSegmentRelocatorFrequency() {
+    if (containsKey(REALTIME_SEGMENT_RELOCATOR_FREQUENCY)) {
+      return (String) getProperty(REALTIME_SEGMENT_RELOCATOR_FREQUENCY);
     }
-    return DEFAULT_REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES;
+    return DEFAULT_REALTIME_SEGMENT_RELOCATOR_FREQUENCY;
   }
 
-  public void setRelocationManagerFrequencyInMinutes(int relocationManagerFrequencyInMinutes) {
-    setProperty(REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES,
-        Integer.toString(relocationManagerFrequencyInMinutes));
+  public void setRealtiemSegmentRelocatorFrequency(String relocatorFrequency) {
+    setProperty(REALTIME_SEGMENT_RELOCATOR_FREQUENCY, relocatorFrequency);
   }
 
   public int getStatusCheckerWaitForPushTimeInSeconds() {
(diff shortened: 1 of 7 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -265,6 +267,18 @@ public void setStatusCheckerFrequencyInSeconds(int statusCheckerFrequencyInSecon
     setProperty(STATUS_CHECKER_FREQUENCY_IN_SECONDS, Integer.toString(statusCheckerFrequencyInSeconds));
   }
 
+  public int getRelocationManagerFrequencyInMinutes() {
+    if (containsKey(REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES)) {
+      return Integer.parseInt((String) getProperty(REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES));
```
mcvsubbu: Can we have the config similar to that in KafkaHighLevelStreamProviderConfig (using PeriodFormatter)? thanks

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+
+/**
+ * Manager to relocate completed segments from consuming servers to completed servers
+ * Segment relocation will be done by this manager, instead of directly moving segments to completed servers on completion,
+ * so that we don't get segment downtime when a segment is in transition
+ *
+ * We only relocate segments for realtime tables, and only if tenant config indicates that relocation is required
+ * A segment will be relocated, one replica at a time, once all of its replicas are in ONLINE state and on consuming servers
+ */
+public class RealtimeSegmentRelocationManager {
```
mcvsubbu: we should base-class the logic for this one, ValidationManager, RetentionManager, and SegmentStatusChecker. You can add it as TODO or do it in this PR, either way is fine.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+   * Check all tables. Perform relocation of segments if table is realtime and relocation is required
+   */
+  public void runRelocation() {
+    if (!_pinotHelixResourceManager.isLeader()) {
+      LOGGER.info("Skipping realtime segment relocation, not leader!");
+      return;
+    }
+
+    LOGGER.info("Starting relocation of realtime segments");
+    List<String> allTableNames = _pinotHelixResourceManager.getAllTables();
```
mcvsubbu: getAllRealtimeTables()

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+
+    for (String tableNameWithType : allTableNames) {
+
+      TableType tableType = TableNameBuilder.getTableTypeFromTableName(tableNameWithType);
+      if (tableType.equals(TableType.REALTIME)) {
+        LOGGER.info("Starting relocation of segments for table: {}", tableNameWithType);
+
+        TableConfig tableConfig = _pinotHelixResourceManager.getRealtimeTableConfig(tableNameWithType);
```
mcvsubbu: Do the whole thing under a try/catch, moving on (after logging the table name that had a problem) if there is an issue with one table. Otherwise, we may never move past a table that had any exception for whatever reason.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+    LOGGER.info("Realtime segment relocation completed");
+  }
+
+  /**
+   * Given a realtime tag config and an ideal state, relocate the segments
+   * which are completed but still hanging around on consuming servers, one replica at a time
+   * @param  realtimeTagConfig
+   * @param idealState
+   */
+  protected Map<String, Map<String, String>> relocateSegments(RealtimeTagConfig realtimeTagConfig,
```
mcvsubbu: Maybe the return type can be IdealState? (but the caller can use the map fields for replacing segments). We may need to document that it is a partial idealstate (containing only the segments that have been moved)

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+          "Found no realtime consuming servers with tag " + realtimeTagConfig.getConsumingServerTag());
+    }
+    List<String> completedServers = _helixAdmin.getInstancesInClusterWithTag(_helixManager.getClusterName(),
+        realtimeTagConfig.getCompletedServerTag());
+    if (completedServers.isEmpty()) {
+      throw new IllegalStateException(
+          "Found no realtime completed servers with tag " + realtimeTagConfig.getCompletedServerTag());
+    }
+
+    Map<String, Map<String, String>> mapFields = idealState.getRecord().getMapFields();
```
mcvsubbu: Please add a TODO here to use the SegmentAssigmentStrategy to place newly completed segments

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+   */
+  private Map<String, Map<String, String>> createNewSegmentToInstanceStateMap(
+      Map<String, Map<String, String>> mapFields, List<String> consumingServers,
+      MinMaxPriorityQueue<Map.Entry<String, Integer>> completedServersQueue) {
+
+    Map<String, Map<String, String>> segmentNameToInstancesMap = new HashMap<>();
+    // TODO: we are scanning the entire segments list every time. This is unnecessary because only the latest segments will need relocation
+    // Can we do something to avoid this?
+    // 1. Time boundary: scan only last day whereas runFrequency = hourly
+    // 2. For each partition, scan in descending order, and stop when the first segment not needing relocation is found
```
mcvsubbu: (2) is good, but then we will need to sort the segments anyway. If we have a table that has a long retention for realtime, there may be a lot of segments in there....

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManagerTest.java
```diff
@@ -0,0 +1,243 @@
...
+      serverNames[i] = "Server_" + i;
+      consumingServerNames[i] = "ConsumingServer_" + i;
+    }
+  }
+
+  /**
+   * Tests the relocation of segments from consuming server to completed server, once all replicas become ONLINE
+   */
+  @Test
+  public void testRelocationOfConsumingSegments() {
```
mcvsubbu: Nice job on the test, thanks.

kishoreg:
Why is it called Relocation Manager?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+
+  public void stop() {
+    _executorService.shutdown();
+  }
+
+  /**
+   * Check all tables. Perform relocation of segments if table is realtime and relocation is required
+   */
+  public void runRelocation() {
+    if (!_pinotHelixResourceManager.isLeader()) {
```
kishoreg: every back ground service needs to check for this. Is it possible to have something that manages background services that starts and stops on leadership change?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+    return newInstanceStateMap;
+  }
+
+  /**
+   * Updates the ideal state of given tablename with the new segment to instance state map
+   * @param tableNameWithType
+   * @param segmentNameToInstanceMap
+   */
+  private void updateIdealState(final String tableNameWithType,
```
kishoreg: what happens when this fails? It probably fails because the version has changed. Retrying is not going to help without computing the new idealstate.

mcvsubbu: +1, good catch

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
```diff
@@ -0,0 +1,287 @@
...
+ */
+public class RealtimeSegmentRelocationManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(RealtimeSegmentRelocationManager.class);
+  private final ScheduledExecutorService _executorService;
+  private final PinotHelixResourceManager _pinotHelixResourceManager;
+  private final HelixManager _helixManager;
+  private final HelixAdmin _helixAdmin;
+  private final long _runFrequencyMinutes;
+
+  public RealtimeSegmentRelocationManager(PinotHelixResourceManager pinotHelixResourceManager, ControllerConf config) {
```
mcvsubbu: RealtimeSegmentRelocator?

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -41,6 +41,7 @@
   private static final String RETENTION_MANAGER_FREQUENCY_IN_SECONDS = "controller.retention.frequencyInSeconds";
   private static final String VALIDATION_MANAGER_FREQUENCY_IN_SECONDS = "controller.validation.frequencyInSeconds";
   private static final String STATUS_CHECKER_FREQUENCY_IN_SECONDS = "controller.statuschecker.frequencyInSeconds";
+  private static final String REALTIME_SEGMENTS_RELOCATION_MANAGER_FREQUENCY_IN_MINUTES = "controller.segment.relocation.frequencyInMinutes";
```
mcvsubbu: You can change this to controller.segment.relocation.frequency and accept strings using PeriodFormatter. Also, you may want to consider controller.realtime.segment.relocation.frequency

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
npawar: Changed it. I was trying to keep it consistent with all the other frequency values in the config

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
npawar: I can look into this as part of a future PR, will avoid trying to touch so many files in this PR as it is only for RelocatorService

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
npawar: Ah yes, good suggestion

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocationManager.java
npawar: Yes, thanks for catching that. Changed it to put the compute logic in retry along with the update

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -265,6 +267,17 @@ public void setStatusCheckerFrequencyInSeconds(int statusCheckerFrequencyInSecon
...
   }
 
+  public String getRealtimeSegmentRelocatorFrequency() {
+    if (containsKey(REALTIME_SEGMENT_RELOCATOR_FREQUENCY)) {
+      return (String) getProperty(REALTIME_SEGMENT_RELOCATOR_FREQUENCY);
+    }
+    return DEFAULT_REALTIME_SEGMENT_RELOCATOR_FREQUENCY;
+  }
+
+  public void setRealtiemSegmentRelocatorFrequency(String relocatorFrequency) {
```
kishoreg: realtime typo

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
```diff
@@ -0,0 +1,278 @@
...
+    if (completedServers.isEmpty()) {
+      throw new IllegalStateException(
+          "Found no realtime completed servers with tag " + realtimeTagConfig.getCompletedServerTag());
+    }
+
+    // TODO: use segment assignment strategy to decide where to place relocated segment
+
+    // create map of completed server name to num segments it holds.
+    // This will help us decide which completed server to choose for replacing a consuming server
+    Map<String, Integer> completedServerToNumSegments = new HashMap<>(completedServers.size());
```
kishoreg: will be good to have extract this logic into an interface, we might come up with different implementations and will abstract out the logic. The interface should basically take an idealstate, set of consuming nodes, set of completed nodes and return a new idealstate.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * Manager to relocate completed segments from consuming servers to completed servers
+ * Segment relocation will be done by this manager, instead of directly moving segments to completed servers on completion,
+ * so that we don't get segment downtime when a segment is in transition
+ *
+ * We only relocate segments for realtime tables, and only if tenant config indicates that relocation is required
+ * A segment will be relocated, one replica at a time, once all of its replicas are in ONLINE state and on consuming servers
+ */
+public class RealtimeSegmentRelocator {
```
kishoreg: Conceptually, this is another rebalancer that is getting triggered periodically instead of getting triggered manually right?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
```diff
@@ -0,0 +1,278 @@
...
+   * Given a realtime tag config and an ideal state, relocate the segments
+   * which are completed but still hanging around on consuming servers, one replica at a time
+   * @param  realtimeTagConfig
+   * @param idealState
+   */
+  protected void relocateSegments(RealtimeTagConfig realtimeTagConfig, IdealState idealState) {
+
+    List<String> consumingServers = _helixAdmin.getInstancesInClusterWithTag(_helixManager.getClusterName(),
+        realtimeTagConfig.getConsumingServerTag());
+    if (consumingServers.isEmpty()) {
```
kishoreg: do we want to retry in the previous method if there is config error?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
```diff
@@ -0,0 +1,278 @@
...
+
+        Function<IdealState, IdealState> updater = new Function<IdealState, IdealState>() {
+          @Nullable
+          @Override
+          public IdealState apply(@Nullable IdealState idealState) {
+            if (!idealState.isEnabled()) {
+              LOGGER.info("Skipping relocation of segments for {} since ideal state is disabled", tableNameWithType);
+              return null;
+            }
+            relocateSegments(realtimeTagConfig, idealState);
```
kishoreg: its probably better to clone the idealstate and then compare if there was a change before updating the idealstate. Currently, we might just keep updating the idealstate everytime this runs.

npawar: All that happens inside HelixHelper.updateIdealState. It compares updated ideal state and original ideal state, and updates only if there's a change

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
npawar: I don't plan to keep this logic for long (see TODO). The plan is to be use SegmentAssignmentStrategy for this assignment. Currently the interface and implementations of SegmentAssignmentStrategy aren't directly usable for for this purpose. Will be working on that in another PR.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
npawar: Hmm, no point in retrying for config error. Will see if i can restructure it a bit

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
npawar: Yes, as per our discussion offline, will model the logic in this service to use RebalanceSegmentsStrategy - https://github.com/linkedin/pinot/issues/2609

Issue #2583, opened 2018-03-06
Create an abstraction for stream in realtime


npawar:
Pinot should be able to support any stream in realtime, not just kafka. Our code currently is not designed to allow that. We have references to Kafka and all realtime code designed assuming the stream is always kafka. 
We will bring in the abstraction for stream. We will do this over multiple iterations. 
Things to be done:
1) Introduce interfaces for Stream, StreamMetadata, StreamProvider, StreamProvider config, factories and others to make realtime code stream agnostic
2) Rename all references to "kafka" appropriately

mcvsubbu:
The things we assume in the code are:

The stream is partitioned (if you want to use a partition-based consumer)
The stream provides APIs to consume all partitions at a time, or a single partition at a time
The stream has a topic name, some timeout settings, etc.
The stream has a (comparable) offset that is a long

Most of the code changes required are moving classes and renaming methods/members to not contain the word 'kafka'. We can do this in a few steps to get to the point where all Kafka related code is in one package, and then look to making that an class(or classes) that can be injected.

The segment completion protocol deals with offsets as being long, so some work may be needed if we wanted a generic Comparable as an offset. We can get there later.

npawar:
Tasks:
- In LLRealtimeDataManager, remove the use of Kafka specific errors
- Introduce abstraction for StreamMetadataProvider, which would answer fetchPartitionCount.

EronWright:
@npawar are you envisioning a plugin system here?  i.e. install a stream provider, configure it, provide a facility for storing state (e.g. offsets), some documentation, etc?

mcvsubbu:
@EronWright what we are trying to do is to unweave Kafka and Pinot.  The idea is that you should be able to inject classes that can parse the configuration that they need and provide functionality to read from an incoming stream. This is ongoing iterative work, but we are getting started.

EronWright:
Thanks @mcvsubbu, by formalizing the effort a bit more, the community could jump in and help out.

fx19880617:
Also mentioned in #3364, there is a backward-incompatible issue of config consumer.factory.class.name. This was not a required config when we create realtime tables.

Could we add a default value for this or infer it from somewhere?

Below are typical configs are are using:
For highlevel consumer:

      "streamConfigs": {
        "stream.kafka.decoder.class.name": "com.uber.uPinot.decoder.HeatpipeKafkaMessageDecoder",
        "streamType": "kafka",
[...]

PR #2606, opened 2018-03-09
Issue-2583: Step 1 to support generic streams

mcvsubbu:
- Create a new package com.linkedin.pinot.core.realtime.stream in pinot-core.
- Move KafkaStreamMetadata to com.linkedin.pinot.core.realtime.stream, renaming it to StreamMetadata
- Get rid of (unused) StreamMetadata
- Move (and rename) classes to be kafka-agnostic (which they are for most parts)
- Add TODO in a couple of places where classes are kafka-specific


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/StreamMessageDecoder.java
```diff
@@ -21,7 +21,7 @@
 import com.linkedin.pinot.core.data.GenericRow;
 
 
-public interface KafkaMessageDecoder<T> {
```
npawar: Do we want to move this to realtime/stream ? it is currently in realtime/impl/kafka

mcvsubbu: yes, good point.

EronWright:
Linking to #2583

PR #2611, opened 2018-03-12
[TE] rootcause - cap change indicator at 10x

apucher:


thirdeye/thirdeye-frontend/app/utils/utils.js
```diff
@@ -53,13 +53,24 @@ export function humanizeFloat(f) {
...
+ * @example
+ * 0.5177  => +51.8%
+ * -0.0001 => -0.0%
+ * 10.001  => +1000%+
+ * -10.01  => -1000%+
  */
 export function humanizeChange(f) {
   if (!isValidForDisplay(f)) { return '-'; }
-    return `${f > 0 ? '+' : ''}${Math.round(f * 100).toFixed(1)}%`;
+  if (f > 10) {
```
justYves: nit: let's use a more meaningful name instead of `f`

apucher: yeah. I'll do this across the entire utils in a separate PR

PR #2614, opened 2018-03-12
Cleanup: Enable splitting of in-clause strings by default.

mayankshriv:
PR's #1819 and #2188 enabled optimization that avoids unnecessary
split/join of strings in IN/NOT-IN clauses. This was done in a backward
compatible way where server side could handle both split as well as
unsplit strings.

This optimization was off-by default and was enabled with a config.
Now that the changes have been around for several months, this PR cleans
up the config and enables the change by default.

 The server-side is still backward compatible with old broker. This can
 be cleaned up future PR once broker changes have been around for a
while.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
@@ -30,6 +30,7 @@ import com.linkedin.pinot.pql.parsers.pql2.ast.RegexpLikePredicateAstNode;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Stack;
+import javax.annotation.concurrent.ThreadSafe;
 import org.antlr.v4.runtime.ANTLRInputStream;
 import org.antlr.v4.runtime.BailErrorStrategy;
 import org.antlr.v4.runtime.BaseErrorListener;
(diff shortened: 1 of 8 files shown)


pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
```diff
@@ -52,7 +52,6 @@
  * PQL 2 compiler.
  */
 public class Pql2Compiler implements AbstractCompiler {
```
Jackie-Jiang: Please add ThreadSafe annotation so that next time we don't add such changeable member variable.

mayankshriv:
@Jackie-Jiang is right.

mayankshriv:
Updated PR to keep the server-side backward compatible with old broker. Added TODO to clean it up.

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
mayankshriv: I am not sure if it is thread safe.

Jackie-Jiang: No member variable, ErrorListener is thread safe, so I think it is thread safe. Also, we already assume it is thread safe in the callers.

PR #2615, opened 2018-03-12
[TE] yyuen/ember upgrade 2.18

justYves:
### What's new: 
- `ember-cli` upgrade from `2.16` to `2.18`
- set the following rules to "warnings" until all fixed cc @aaronucsd 
  -    "ember/avoid-leaking-state-in-ember-objects": ["warn"],
  -   "ember/no-attrs-in-components": ["warn"]
- Fix tests
  - call `this._super(...arguments)` in all `init()`
  - use Snake Case for route queryParams
- `ember-cli-mirage` upgrade from `0.3` to `0.4`
- Start `ember-cli-mirage`'s server for all tests

### Tests:
- no failing tests
- This needs to be tested extensively on staging cc @apucher


thirdeye/thirdeye-frontend/app/pods/components/alert-report-modal/component.js
```diff
@@ -82,7 +82,7 @@ export default Component.extend({
    * @method bubbleModalInput
    */
   bubbleModalInput() {
-    this.sendAction('inputAction', this.get('reportAnomalyPayload'));
+    this.get('inputAction')(this.get('reportAnomalyPayload'));
```
igbopie: You can do this.inputActions(...) if you want.

thirdeye/thirdeye-frontend/app/pods/components/events-table/component.js
```diff
@@ -52,57 +52,62 @@ export default Component.extend({
...
     later(() => {
-      this._super(args);
+      this._super(...arguments);
     });
   },
 
-  /**
-   * Columns required by ember-models-table
-   */
-  columns: [
```
igbopie: You can use Object.freeze(...)

justYves: oh thanks! didn't know about this work around

thirdeye/thirdeye-frontend/app/pods/components/alert-report-modal/component.js
aaronucsd: I personally would keep it `get` as that's more proper way of getting and setting props for Ember.Object.

thirdeye/thirdeye-frontend/app/pods/components/anomaly-id/component.js
```diff
@@ -24,7 +24,7 @@ export default Component.extend({
    * @method anomalyChangeRate
    * @return {Number} - total % change from baseline
    */
-  anomalyChangeRate: computed('anomaly.current', 'anomaly.baseline', function() {
+  anomalyChangeRate: computed('anomaly.{current,baseline}', function() {
```
aaronucsd: This is interesting to combined props of the same parent object.

justYves: This is actually the new syntax, ember https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/use-brace-expansion.md

thirdeye/thirdeye-frontend/app/pods/components/events-table/component.js
```diff
@@ -52,57 +52,62 @@ export default Component.extend({
...
-      sortedBy: 'duration',
-      disableFiltering: true
-    }
+  init() {
+    this._super(...arguments);
+
+    /**
+     * Columns required by ember-models-table
+     */
+    const columns = [
```
aaronucsd: This init is growing too large. Can we move this outside of init? Like at the top? Or better yet, some const file?

thirdeye/thirdeye-frontend/app/pods/components/alert-report-modal/component.js
igbopie: That is changing ;)
https://github.com/emberjs/rfcs/blob/master/text/0281-es5-getters.md

aaronucsd: Man. Ember keeps changing like this. Soon we just do regular vanilla. :)

PR #2640, opened 2018-03-19
Printing headers for segment upload requests

jenniferdai:


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -279,6 +279,10 @@ private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, Strin
     File tempTarredSegmentFile = null;
     File tempSegmentDir = null;
 
+    if (headers != null) {
+      LOGGER.info("Got HTTP Request headers {}", headers.getRequestHeaders());
```
Jackie-Jiang: Could headers have confidential information?

jenniferdai: Not at this point. If there is confidential information, I'm not sure how to distinguish especially since we have internal-specific headers, and we want to keep this method generic.

Jackie-Jiang: Just log the info needed (e.g. headers.getHeaderString(FileUploadDownloadClient.CustomHeaders.UPLOAD_TYPE)), but not the whole headers

jenniferdai: I also have segment name and table name headers specific to pbnj, I'm not looking for upload type necessarily

Jackie-Jiang: The point is, the validation information will also be in headers, and cannot be logged

mcvsubbu: +1, just log the headers that we need to log.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -279,6 +279,11 @@ private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, Strin
     File tempTarredSegmentFile = null;
     File tempSegmentDir = null;
 
+    if (headers != null) {
+      LOGGER.info("HTTP Header {} is {}", CommonConstants.Controller.SEGMENT_NAME_HTTP_HEADER, headers.getRequestHeader(CommonConstants.Controller.SEGMENT_NAME_HTTP_HEADER));
```
Jackie-Jiang: Let's add a TODO here to add these 2 headers into segment upload

PR #2641, opened 2018-03-19
[TE] rootcause - endpoint cleanup and bug fixes

apucher:
This PR combines a number of fixes and refactoring in the rootcause data endpoints.

* remove ahead-of-time caching again as ineffective with non-sticky load balancing

* remove deprecated timeseries and aggregates endpoints

* externally inject daily data flag into BaselineAggregate

* remove DST heuristic from BaselineAggregate

* use client timezone for baseline range alignment


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/RootCauseMetricResource.java
```diff
@@ -588,13 +539,33 @@ private MetricSlice alignSlice(MetricSlice slice) throws Exception {
       granularity = slice.getGranularity();
     }
 
+    long offset = DateTimeZone.forID(timezone).getOffset(slice.getStart());
     long timeGranularity = granularity.toMillis();
-    long start = (slice.getStart() / timeGranularity) * timeGranularity;
-    long end = ((slice.getEnd() + timeGranularity - 1) / timeGranularity) * timeGranularity;
+    long start = ((slice.getStart() + offset) / timeGranularity) * timeGranularity - offset;
```
cyy0714: Could you explain why we need to add and reduce an offset here?

apucher: Time buckets are computed in days, but may need to be offset against UTC for clients in non-utc time zones, i.e. the bucket-boundaries should align to UTC + offset day boundaries instead of plain UTC day boundaries

cyy0714: Could you leave a comment on the code segment? It is easier for people to understand it later.

apucher: I added javadoc above

cyy0714: I mean if you can leave a comment for line 542 to line 545. It is easier for people to quickly understand.

PR #2643, opened 2018-03-20
Return the updated ideal state for api after rebalance

npawar:


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/rebalance/DefaultRebalanceSegmentStrategy.java
```diff
@@ -189,6 +191,10 @@ public IdealState apply(@Nullable IdealState idealState) {
         rebalanceIdealState(idealState, tableConfig, targetNumReplicas, rebalanceUserConfig,
             newPartitionAssignment);
         idealState.setReplicas(Integer.toString(targetNumReplicas));
+        for (Map.Entry<String, Map<String, String>> entry : idealState.getRecord().getMapFields().entrySet()) {
```
kishoreg: what if the update fails but the rebalanceIdealState is already updated. We should update the rebalanceIdealState only update to ZK was successful

npawar: HelixHelper.updateIdealState does not return any indication that it succeeded

mcvsubbu: Can't we do all the updates to 'idealState' in the method rebalanceIdealState(...) and just return the updated idealState object?

npawar: That is exactly what is happening. But how do I send it back from rebalanceAndUpdateIdealState? It needs to be sent back to be displayed

npawar: That is exactly what is happening. But how do I send it back from rebalanceAndUpdateIdealState? It needs to be sent back to be displayed

snleee: Maybe we can create a blocking retry policy?

PR #2645, opened 2018-03-20
[TE] Add holiday events loader

jihaozh:
- Load holiday events from Google Calendar API
- Run holiday events loader periodically

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
@@ -4,8 +4,11 @@ import com.linkedin.thirdeye.anomaly.monitor.MonitorConfiguration;
 import com.linkedin.thirdeye.anomaly.task.TaskDriverConfiguration;
 import com.linkedin.thirdeye.auto.onboard.AutoOnboardConfiguration;
 import com.linkedin.thirdeye.common.ThirdEyeConfiguration;
+import java.util.List;
+
 
 public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
+  private boolean holidayEventsLoader = false;
   private boolean scheduler = false;
   private boolean worker = false;
   private boolean monitor = false;
@@ -26,6 +29,32 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
   private TaskDriverConfiguration taskDriverConfiguration = new TaskDriverConfiguration();
   private String failureFromAddress;
   private String failureToAddress;
+  private String keyPath;
+  private List<String> calendars;
+
+  public String getKeyPath() {
+    return keyPath;
+  }
+
+  public void setKeyPath(String keyPath) {
+    this.keyPath = keyPath;
+  }
+
+  public List<String> getCalendars() {
+    return calendars;
+  }
+
+  public void setCalendars(List<String> calendars) {
+    this.calendars = calendars;
+  }
+
+  public boolean isHolidayEventsLoader() {
+    return holidayEventsLoader;
+  }
+
+  public void setHolidayEventsLoader(boolean holidayEventsLoader) {
+    this.holidayEventsLoader = holidayEventsLoader;
+  }
 
   public String getDashboardHost() {
     return dashboardHost;
@@ -74,8 +103,6 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
   public void setMonitorConfiguration(MonitorConfiguration monitorConfiguration) {
     this.monitorConfiguration = monitorConfiguration;
   }
-
-
 
   public AutoOnboardConfiguration getAutoOnboardConfiguration() {
     return autoOnboardConfiguration;
@@ -172,5 +199,4 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
   public void setFailureToAddress(String failureToAddress) {
     this.failureToAddress = failureToAddress;
   }
-
 }
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
@@ -22,6 +22,7 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
   private boolean worker = false;
 
   private long id;
+  private long holidayRange;
   private String dashboardHost;
   private SmtpConfiguration smtpConfiguration;
   private MonitorConfiguration monitorConfiguration = new MonitorConfiguration();
@@ -31,6 +32,14 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
   private String failureToAddress;
   private String keyPath;
   private List<String> calendars = Collections.emptyList();
+
+  public long getHolidayRange() {
+    return holidayRange;
+  }
+
+  public void setHolidayRange(long holidayRange) {
+    this.holidayRange = holidayRange;
+  }
 
   public String getKeyPath() {
     return keyPath;
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
@@ -4,8 +4,6 @@ import com.linkedin.thirdeye.anomaly.monitor.MonitorConfiguration;
 import com.linkedin.thirdeye.anomaly.task.TaskDriverConfiguration;
 import com.linkedin.thirdeye.auto.onboard.AutoOnboardConfiguration;
 import com.linkedin.thirdeye.common.ThirdEyeConfiguration;
-import java.util.Collections;
-import java.util.List;
 
 
 public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
@@ -22,39 +20,22 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
   private boolean worker = false;
 
   private long id;
-  private long holidayRange;
   private String dashboardHost;
   private SmtpConfiguration smtpConfiguration;
+  private HolidayEventsLoaderConfiguration holidayEventsLoaderConfiguration;
   private MonitorConfiguration monitorConfiguration = new MonitorConfiguration();
   private AutoOnboardConfiguration autoOnboardConfiguration = new AutoOnboardConfiguration();
   private TaskDriverConfiguration taskDriverConfiguration = new TaskDriverConfiguration();
   private String failureFromAddress;
   private String failureToAddress;
-  private String keyPath;
-  private List<String> calendars = Collections.emptyList();
 
-  public long getHolidayRange() {
-    return holidayRange;
+
+  public HolidayEventsLoaderConfiguration getHolidayEventsLoaderConfiguration() {
+    return holidayEventsLoaderConfiguration;
   }
 
-  public void setHolidayRange(long holidayRange) {
-    this.holidayRange = holidayRange;
-  }
-
-  public String getKeyPath() {
-    return keyPath;
-  }
-
-  public void setKeyPath(String keyPath) {
-    this.keyPath = keyPath;
-  }
-
-  public List<String> getCalendars() {
-    return calendars;
-  }
-
-  public void setCalendars(List<String> calendars) {
-    this.calendars = calendars;
+  public void setHolidayEventsLoaderConfiguration(HolidayEventsLoaderConfiguration holidayEventsLoaderConfiguration) {
+    this.holidayEventsLoaderConfiguration = holidayEventsLoaderConfiguration;
   }
 
   public boolean isHolidayEventsLoader() {
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
@@ -4,22 +4,22 @@ import com.linkedin.thirdeye.anomaly.monitor.MonitorConfiguration;
 import com.linkedin.thirdeye.anomaly.task.TaskDriverConfiguration;
 import com.linkedin.thirdeye.auto.onboard.AutoOnboardConfiguration;
 import com.linkedin.thirdeye.common.ThirdEyeConfiguration;
+import java.util.Collections;
 import java.util.List;
 
 
 public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
+
+  private boolean alert = false;
+  private boolean autoload = false;
+  private boolean classifier = false;
+  private boolean dataCompleteness = false;
+  private boolean detectionOnboard = false;
   private boolean holidayEventsLoader = false;
+  private boolean monitor = false;
+  private boolean pinotProxy = false;
   private boolean scheduler = false;
   private boolean worker = false;
-  private boolean monitor = false;
-  private boolean alert = false;
-  @Deprecated
-  private boolean merger = false;
-  private boolean autoload = false;
-  private boolean dataCompleteness = false;
-  private boolean classifier = false;
-  private boolean pinotProxy = false;
-  private boolean detectionOnboard = false;
 
   private long id;
   private String dashboardHost;
@@ -30,7 +30,7 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
   private String failureFromAddress;
   private String failureToAddress;
   private String keyPath;
-  private List<String> calendars;
+  private List<String> calendars = Collections.emptyList();
 
   public String getKeyPath() {
     return keyPath;
@@ -130,14 +130,6 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
 
   public SmtpConfiguration getSmtpConfiguration() {
     return smtpConfiguration;
-  }
-
-  public boolean isMerger() {
-    return merger;
-  }
-
-  public void setMerger(boolean merger) {
-    this.merger = merger;
   }
 
   public boolean isAutoload() {
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
@@ -22,7 +22,7 @@ public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
   private long id;
   private String dashboardHost;
   private SmtpConfiguration smtpConfiguration;
-  private HolidayEventsLoaderConfiguration holidayEventsLoaderConfiguration;
+  private HolidayEventsLoaderConfiguration holidayEventsLoaderConfiguration = new HolidayEventsLoaderConfiguration();
   private MonitorConfiguration monitorConfiguration = new MonitorConfiguration();
   private AutoOnboardConfiguration autoOnboardConfiguration = new AutoOnboardConfiguration();
   private TaskDriverConfiguration taskDriverConfiguration = new TaskDriverConfiguration();
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
@@ -9,11 +9,10 @@ import com.google.api.client.util.DateTime;
 import com.google.api.services.calendar.Calendar;
 import com.google.api.services.calendar.CalendarScopes;
 import com.google.api.services.calendar.model.Event;
-import com.linkedin.thirdeye.anomaly.ThirdEyeAnomalyConfiguration;
+import com.linkedin.thirdeye.anomaly.HolidayEventsLoaderConfiguration;
 import com.linkedin.thirdeye.api.TimeGranularity;
 import com.linkedin.thirdeye.datalayer.bao.EventManager;
 import com.linkedin.thirdeye.datalayer.dto.EventDTO;
-import com.linkedin.thirdeye.datasource.DAORegistry;
 import java.io.FileInputStream;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -36,7 +35,7 @@ import org.slf4j.LoggerFactory;
  */
 public class HolidayEventsLoader implements Runnable {
 
-  private class HolidayEvent {
+  static class HolidayEvent {
     /**
      * The Name.
      */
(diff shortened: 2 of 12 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
```diff
@@ -4,8 +4,11 @@
 import com.linkedin.thirdeye.anomaly.task.TaskDriverConfiguration;
 import com.linkedin.thirdeye.auto.onboard.AutoOnboardConfiguration;
 import com.linkedin.thirdeye.common.ThirdEyeConfiguration;
+import java.util.List;
+
 
 public class ThirdEyeAnomalyConfiguration extends ThirdEyeConfiguration {
+  private boolean holidayEventsLoader = false;
```
cyenjung: Since you are modifying this part of code, could you order the flags alphabetically? And remove the deprecated isMerger flag. Thanks.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
```diff
@@ -26,6 +29,32 @@
   private TaskDriverConfiguration taskDriverConfiguration = new TaskDriverConfiguration();
   private String failureFromAddress;
   private String failureToAddress;
+  private String keyPath;
+  private List<String> calendars;
```
cyenjung: Initialize this list with Collections.emptyList() to prevent nullPointException in the future.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/EventDTO.java
```diff
@@ -1,7 +1,24 @@
...
 
 import com.linkedin.thirdeye.datalayer.pojo.EventBean;
+import java.util.Objects;
+
 
 public class EventDTO extends EventBean {
 
+  @Override
+  public int hashCode() {
+    return com.google.common.base.Objects.hashCode(getName(), getStartTime(), getEndTime(), getEventType());
```
cyenjung: Use Java 7's Objects instead. In fact, you have imported java.util.Objects.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
```diff
@@ -0,0 +1,142 @@
...
+        events.addAll(this.getCalendarEvents(calendar));
+      } catch (Exception e) {
+        // left blank
+      }
+    }
+    return events;
+  }
+
+  private List<Event> getCalendarEvents(String Calendar_id) throws Exception {
+
```
cyenjung: Throw an exception if HTTP_TRANSPORT is not initialized successfully at line 67.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
```diff
@@ -0,0 +1,142 @@
...
+  private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+
+  /** Global instance of the scopes. */
+  private static final Set<String> SCOPES = Collections.singleton(CalendarScopes.CALENDAR_READONLY);
+
+  static {
+    try {
+      HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+    } catch (Exception e) {
+      LOG.error("Can't create http transport with google api.");
```
cyenjung: Append the exception at the end of this message:
    LOG.error("Can't create http transport with google api.", e);

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
jihaozh: Sure!

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * Fetch all holidays from Google Calendar API.
+   */
+  private List<Event> getAllHolidays() {
+    List<Event> events = new ArrayList<>();
+    for (String calendar : calendarList) {
+      try {
+        events.addAll(this.getCalendarEvents(calendar));
+      } catch (Exception e) {
+        // left blank
```
cyenjung: Leave a message to indicate from which calendar we have failed to retrieve the events.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+    return events;
+  }
+
+  private List<Event> getCalendarEvents(String Calendar_id) throws Exception {
+
+    GoogleCredential credential = GoogleCredential.fromStream(new FileInputStream(keyPath)).createScoped(SCOPES);
+    Calendar service =
+        new Calendar.Builder(HTTP_TRANSPORT, JSON_FACTORY, credential).setApplicationName("thirdeye").build();
+    Events events = service.events().list(Calendar_id).setTimeMin(new DateTime(System.currentTimeMillis())).execute();
```
cyenjung: Dose this list have bounds on start time? Otherwise, we would get a long list of events and most of them are duplicated to the ones that have stored in TE's DB.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
```diff
@@ -0,0 +1,142 @@
...
+      EventDTO eventDTO = new EventDTO();
+
+      eventDTO.setStartTime(holiday.getStart().getDate().getValue());
+      eventDTO.setEndTime(holiday.getEnd().getDate().getValue());
+      eventDTO.setName(holiday.getSummary());
+      eventDTO.setEventType(EventType.HOLIDAY.toString());
+
+      newEvents.add(eventDTO);
+    }
+    Set<EventDTO> existingEvents = new HashSet<>(eventDAO.findAll());
```
cyenjung: This method is going to become very slow after we collect a lot of holiday events. Please use the notion of watermark. There are two advantages: 1. we could fetch new holiday events that happens after the watermark. 2. watermark of each calendar is a timestamp instead of a list of events.

jihaozh: I could do that. But the problem is that there are a lot of existing holiday events in our DB right now. If I don't collect existing events there might be duplication.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
jihaozh: Yes. I've set up the minimum time to now, so this will only fetch the future holidays.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
apucher: There's various solutions to this, but I wonder if this would be premature optimization. the database has about 600 events per year. hardly a concern anytime soon

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
```diff
@@ -0,0 +1,142 @@
+package com.linkedin.thirdeye.rootcause.impl;
```
apucher: I'd move this to the existing events package "com.linkedin.thirdeye.anomaly.events"

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/rootcause/impl/HolidayEventsLoader.java
```diff
@@ -0,0 +1,142 @@
...
+      eventDTO.setEndTime(holiday.getEnd().getDate().getValue());
+      eventDTO.setName(holiday.getSummary());
+      eventDTO.setEventType(EventType.HOLIDAY.toString());
+
+      newEvents.add(eventDTO);
+    }
+    Set<EventDTO> existingEvents = new HashSet<>(eventDAO.findAll());
+
+    // remove existing holidays
+    newEvents.removeAll(existingEvents);
```
apucher: This may need to be a bit smarter. Events sometimes change start/end time - this logic would duplicate them. I'd suggest to write a custom, name-based logic here instead of relying on the equals() and hashCode() methods.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/EventDTO.java
```diff
@@ -1,7 +1,24 @@
...
 
 import com.linkedin.thirdeye.datalayer.pojo.EventBean;
+import java.util.Objects;
+
 
 public class EventDTO extends EventBean {
 
+  @Override
+  public int hashCode() {
+    return Objects.hash(getName(), getStartTime(), getEndTime(), getEventType());
```
apucher: I strongly suggest to include id in equals() and hashCode(). Otherwise, we might see some strage side-effects with the storage backend

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+    // Group by holidays and put the country codes in the corresponding set
+    for (Event holiday : newHolidays) {
+      HolidayEvent holidayEvent =
+          new HolidayEvent(holiday.getSummary(), EventType.HOLIDAY.toString(), holiday.getStart().getDate().getValue(),
+              holiday.getEnd().getDate().getValue());
+
+      if (!newHolidayEventToCountryCodes.containsKey(holidayEvent)) {
+        newHolidayEventToCountryCodes.put(holidayEvent, new HashSet<String>());
+      }
+      newHolidayEventToCountryCodes.get(holidayEvent).add(getCountryCode(holiday));
```
cyenjung: Take care of the case in which getCountryCode() returns null.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+      }
+      existingHolidayEventToEventDTO.put(holidayEvent, existingEventDTO);
+    }
+
+    for (Map.Entry<HolidayEvent, Set<String>> entry : newHolidayEventToCountryCodes.entrySet()) {
+      HolidayEvent newHolidayEvent = entry.getKey();
+      Set<String> newCountryCodes = entry.getValue();
+
+      if (existingHolidayEventToCountryCodes.containsKey(newHolidayEvent)) {
+        if (existingHolidayEventToCountryCodes.get(newHolidayEvent).addAll(newCountryCodes)) {
```
cyenjung: Take care of the duplicated country codes in existingHolidayEventToCountryCodes.get(newHolidayEvent) and newCountryCodes. You could insert all country codes to a set and output the unique code to the list.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+        Map<String, List<String>> targetDimensionMap = new HashMap<>();
+        targetDimensionMap.put("countryCode", new ArrayList<>(entry.getValue()));
+        eventDTO.setTargetDimensionMap(targetDimensionMap);
+        eventDAO.save(eventDTO);
+      }
+    }
+  }
+
+  private String getCountryCode(Event holiday) {
+    String countryName = holiday.getCreator().getDisplayName().substring(12);
```
cyenjung: Does country name always starts at 12th index?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+    for (Locale locale : Locale.getAvailableLocales()) {
+      if (locale.getDisplayCountry().equals(countryName)) {
+        return locale.getCountry();
+      }
+    }
+    return null;
```
cyenjung: We should return a string like "NO_COUNTRY_CODE" instead of a null string.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  /**
+   * Fetch all holidays from Google Calendar API.
+   */
+  private List<Event> getAllHolidays(Long start, Long end) {
```
cyenjung: Change Long to long because you don't want someone to pass in null longs.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+      try {
+        events.addAll(this.getCalendarEvents(calendar, start, end));
+      } catch (Exception e) {
+        LOG.error("{} Fetch holiday events failed.", calendar);
+      }
+    }
+    return events;
+  }
+
+  private List<Event> getCalendarEvents(String Calendar_id, Long start, Long end) throws Exception {
```
cyenjung: Long --> long

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+    for (Locale locale : Locale.getAvailableLocales()) {
+      if (locale.getDisplayCountry().equals(countryName)) {
+        return locale.getCountry();
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Fetch all holidays from Google Calendar API.
```
cyenjung: Add the condition of holidays that will be returned by this method: a description like holidays whose end time GE start and start time LT end (assuming you are using end time exclusive).

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  /**
+   * Fetch holidays and save to ThirdEye database.
+   */
+  public void run() {
+    EventManager eventDAO = DAO_REGISTRY.getEventDAO();
+
+    long start = System.currentTimeMillis();
+    long end = start + 2592000000L; // to get holidays within a month
```
cyenjung: Store the value 2592000000L in the configuration file; like what you did for the key path and calendar list.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
jihaozh: existingHolidayEventToCountryCodes maps the holiday in to a country codes hashset.  So this shouldn't be a problem

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,337 @@
...
+            .addAll(existingEventDTO.getTargetDimensionMap().get("countryCode"));
+      }
+      existingHolidayEventToEventDTO.put(holidayEvent, existingEventDTO);
+    }
+
+    for (Map.Entry<HolidayEvent, Set<String>> entry : newHolidayEventToCountryCodes.entrySet()) {
+      HolidayEvent newHolidayEvent = entry.getKey();
+      Set<String> newCountryCodes = entry.getValue();
+
+      if (existingHolidayEventToCountryCodes.containsKey(newHolidayEvent)) {
```
apucher: as before, when the holiday changes dates, it is not picked up here. Ultimately, there are three cases: (1) new event appears, (2) existing event changes, and (3) event disappears. The code handles (1) and country additions for (2). I would modify this section to simply override any pre-existing events (and their country codes) with the same name for (2). For 3 you may need to introduce some kind of time-based cutoff and then simply delete events that aren't returned from the API.

Probably makes sense to wrap this logic into a testable unit, too.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/EventDTO.java
```diff
@@ -2,6 +2,8 @@
 
 import com.linkedin.thirdeye.datalayer.pojo.EventBean;
 
+
```
apucher: you could revert these whitespace changes to make the PR smaller

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
jihaozh: Sure

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
jihaozh: If the holiday belongs to a specific country, holiday.getCreator().getDisplayName() will return strings like "Holidays in United States", "Holidays in Canada", etc, If the holiday doesn't belong to a specific country, this returns null.  I'll change this code to check the string length is greater than 12 and also return NO_COUNTRY_CODE if no country code found.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/EventDTO.java
jihaozh: Sure

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
jihaozh: Thanks! I will add this logic and tests.

jihaozh: Can we just remove all existing events within the specific time range and write the new ones?

apucher: yes, that could work. this assumes that there's strict time barrier for moving around events, i.e. events from the future cannot move to the past and vice versa.

there may be an issue with RCA however - saved investigations keep track of event by id, and these events could be in the future (just a few days). If we delete events and the re-create them, they get a new id.

If we put the time barrier for delete/re-create in the future, e.g. by a week, this should work. It may, however, have some non-obvious side-effects.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
```diff
@@ -105,14 +122,6 @@ public SmtpConfiguration getSmtpConfiguration() {
     return smtpConfiguration;
   }
 
-  public boolean isMerger() {
-    return merger;
-  }
-
-  public void setMerger(boolean merger) {
-    this.merger = merger;
-  }
```
apucher: did you rebase to the latest master? this looks like it doesn't belong here

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/common/ThirdEyeConfiguration.java
```diff
@@ -110,6 +110,10 @@ public String getAnomalyClassifierConfigPath() {
     return getRootDir() + "/detector-config/anomaly-functions/anomalyClassifier.properties";
   }
 
+  public String getCalendarApiKeyPath(){
+    return getRootDir() + "/holiday-loader-key.json";
+  }
```
apucher: Remember the CSV data source? Make the path a (partial) URI and part of the configuration for easy testing. You can probably reuse some of the methods

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,364 @@
...
+    scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+  }
+
+  /** List of google holiday calendar ids */
+  private List<String> calendarList;
+
+  /** Calendar Api private key path */
+  private String keyPath;
+  private ScheduledExecutorService scheduledExecutorService;
+  private TimeGranularity runFrequency = new TimeGranularity(7, TimeUnit.DAYS);
```
apucher: This should be part of the configuration

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
```diff
@@ -0,0 +1,364 @@
...
+   *
+   * @param holidayEventsLoaderConfiguration the configuration
+   * @param calendarApiKeyPath the calendar api key path
+   * @param eventDAO the event dao
+   */
+  public HolidayEventsLoader(HolidayEventsLoaderConfiguration holidayEventsLoaderConfiguration,
+      String calendarApiKeyPath, EventManager eventDAO) {
+    this.holidayRange = holidayEventsLoaderConfiguration.getHolidayRange();
+    this.calendarList = holidayEventsLoaderConfiguration.getCalendars();
+    this.keyPath = calendarApiKeyPath;
```
apucher: This should be part of the configuration

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/anomaly/events/MockEventsManager.java
```diff
@@ -0,0 +1,76 @@
...
+    entities.add(entity);
+    return entity.getId();
+  }
+
+  @Override
+  public int update(EventDTO entity) {
+    for (EventDTO eventDTO : entities) {
+      if (eventDTO.getId().equals(entity.getId())) {
+        eventDTO = entity;
+        return 0;
```
apucher: i think this should return 1 on delete, and 0 if not found. but not critical

thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoaderTest.java
```diff
@@ -0,0 +1,106 @@
...
+    List<EventDTO> festivalEvents = eventNameToEventDto.get("Some festival");
+
+    Assert.assertEquals(festivalEvents.size(), 2);
+    Assert.assertTrue(
+        festivalEvents.get(0).getTargetDimensionMap().get("countryCode").equals(Collections.singletonList("uk"))
+            && new HashSet<>(festivalEvents.get(1).getTargetDimensionMap().get("countryCode")).equals(
+            new HashSet<>(Arrays.asList("us", "cn")))
+            || festivalEvents.get(1).getTargetDimensionMap().get("countryCode").equals(Collections.singletonList("uk"))
+            && new HashSet<>(festivalEvents.get(0).getTargetDimensionMap().get("countryCode")).equals(
+            new HashSet<>(Arrays.asList("us", "cn"))));
```
apucher: You can simplify this by custom sorting *festivalEvents* before the assertions

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
jihaozh: Yen let me remove the deprecated isMerger flag since I'm modifying this part of this file.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
jihaozh: Sure. Will change.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/events/HolidayEventsLoader.java
jihaozh: Sure. Will change

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/common/ThirdEyeConfiguration.java
apucher: ok, we'll delay this a sdiscussed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/ThirdEyeAnomalyConfiguration.java
apucher: ok

PR #2647, opened 2018-03-20
[TE] frontend - aaronucsd/Added application loading/error start

aaronucsd:
screen shot 2018-03-19 at 10 39 40 pm

1. Added new error route for catching other routes errors.
2. Added an application level loading handler


thirdeye/thirdeye-frontend/app/pods/manage/alerts/index/route.js
```diff
@@ -1,6 +1,7 @@
 import { hash } from 'rsvp';
 import Route from '@ember/routing/route';
 import fetch from 'fetch';
+import RSVP from 'rsvp';
```
justYves: are we actually using rsvp here?

thirdeye/thirdeye-frontend/app/styles/pods/loading.scss
```diff
@@ -0,0 +1,5 @@
+.route-loader-container {
+  position: absolute;
+  left: 48%;
```
justYves: couldn't we achieve the same layout with flex box ? 
Personally not a big fan of magic numbers to position items. if you were trying to offset the spinner's own height and width. You could keep both `left` and `top` at 50%, and then use the transform() trick.
`transform: translateX(-50%) translateY(-50%);`

thirdeye/thirdeye-frontend/app/pods/manage/alerts/index/route.js
aaronucsd: Let me remove. I had it for testing a promise reject for the error. Thanks.

thirdeye/thirdeye-frontend/app/styles/pods/loading.scss
aaronucsd: i added the flexbox.

thirdeye/thirdeye-frontend/app/pods/manage/alerts/index/route.js
aaronucsd: DONE

thirdeye/thirdeye-frontend/app/styles/pods/loading.scss
aaronucsd: DONE.

PR #2652, opened 2018-03-22
Add VALUE_IN UDF to support multi-valued column group-by without extra values

Jackie-Jiang:
1. Refactor all layers above Projection
2. Transform layer treat all columns as expressions
3. Re-design the TransformFunction interface to support dictionary in UDF
4. Add UDF support to both aggrgation columns and group-by columns
5. VALUE_IN UDF can act on multi-valued group-by columns and returns only the values inside the UDF

TODO: refactor selection to work on top of transform so that we can add UDF to selection in the future


pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/GroupByAstNode.java
```diff
@@ -36,10 +37,9 @@ public void updateBrokerRequest(BrokerRequest brokerRequest) {
         // List of expression contains columns as well as expressions to maintain ordering of group by columns.
         groupBy.addToExpressions(groupByColumnName);
       } else {
-        FunctionCallAstNode functionCallAstNode = (FunctionCallAstNode) astNode;
-
-        // Remove all white-space until we start compiling expressions on broker side.
-        groupBy.addToExpressions(functionCallAstNode.getExpression());
+        // Compile to standard expression string
+        String expression = ((FunctionCallAstNode) astNode).getExpression();
+        groupBy.addToExpressions(TransformExpressionTree.compileToExpressionTree(expression).toString());
```
kishoreg: why are we parsing it here. Is this data structure serializable? How will this backward compatible?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/BaseTransformFunction.java
```diff
@@ -0,0 +1,511 @@
...
+          _stringValuesMV[i] = new String[numValues];
+          copyToString(doubleValues[i], _stringValuesMV[i], numValues);
+        }
+        return _stringValuesMV;
+      default:
+        throw new UnsupportedOperationException();
+    }
+  }
+
+  private static void copyToInt(long[] source, int[] dest, int length) {
```
kishoreg: move all this methods to some util class. Also add comments on why we cant use Arrays.copy methods

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/MultiplicationTransformFunction.java
```diff
@@ -0,0 +1,104 @@
...
+import com.linkedin.pinot.core.plan.DocIdSetPlanNode;
+import com.linkedin.pinot.core.query.exception.BadQueryRequestException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+
+public class MultiplicationTransformFunction extends BaseTransformFunction {
```
kishoreg: why are all renames showing up as deletion and addition.

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/GroupByAstNode.java
Jackie-Jiang: Here we just check whether the expression string can be compiled, and put the standard equivalent expression string into the group-by expressions. This way we can avoid invalid query and ensure the "groupByColumns" in the broker response is standard. Will update the comments

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/MultiplicationTransformFunction.java
Jackie-Jiang: This is not renaming, the interface is redesigned so too many changes, and git identifies it as deletion and addition.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/BaseTransformFunction.java
Jackie-Jiang: Done

pinot-server/src/main/java/com/linkedin/pinot/server/starter/ServerBuilder.java
```diff
@@ -69,9 +70,19 @@ private void initMetrics() {
...
+    for (String transformFunctionClassName : _serverConf.getTransformFunctions()) {
+      Class<TransformFunction> transformFunctionClass;
+      try {
+        transformFunctionClass = (Class<TransformFunction>) Class.forName(transformFunctionClassName);
+      } catch (ClassNotFoundException e) {
+        throw new RuntimeException("Failed to find transform function class of name: " + transformFunctionClassName);
+      }
+      transformFunctionClasses.add(transformFunctionClass);
+    }
+    TransformFunctionFactory.init(transformFunctionClasses);
```
kishoreg: check if the transform function already exists

Jackie-Jiang: Done

PR #2654, opened 2018-03-22
[TE] rootcause - improve concurrency in breakdown endpoint

apucher:
This PR improves data concurrency in the breakdowns endpoint to per-dimension granularity to handle high-dimensional data sets gracefully.

* per-dimension granularity in breakdowns endpoint

* deadline-based connection timeout

* track active connections in pinot connector


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datasource/pinot/PinotControllerResponseCacheLoader.java
```diff
@@ -39,6 +40,8 @@
 
   private static final String BROKER_PREFIX = "Broker_";
 
+  private final AtomicInteger activeConnections = new AtomicInteger();
```
cyenjung: Just use an int is sufficient because you are using it in a synchronized block; JVM will make sure every thread reads the latest update.

apucher: we're synchronizing on one of 25 connections. the int is not protected.

PR #2655, opened 2018-03-22
[PINOT-6005] Add broker metric of maximum response size

jackjlli:
This PR adds broker metric of maximum response size. 
This is for measuring the maximum amount of data that broker needs to process, which is different from the ultimate response size.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
@@ -63,6 +63,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 import org.apache.commons.configuration.Configuration;
+import org.apache.commons.lang.mutable.MutableLong;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.thrift.protocol.TCompactProtocol;
 import org.json.JSONObject;
@@ -511,13 +512,16 @@ public class BrokerRequestHandler {
     int numServersResponded = 0;
     long deserializationStartTime = System.nanoTime();
     Map<ServerInstance, DataTable> dataTableMap = new HashMap<>();
+    // Add a long variable to sum the total response sizes from both realtime and offline servers.
+    long totalServerResponseSize = 0L;
     if (offlineServerResponseMap != null) {
       numServersResponded += offlineServerResponseMap.size();
-      deserializeServerResponses(offlineServerResponseMap, true, dataTableMap, offlineTableName, processingExceptions);
+      totalServerResponseSize += deserializeServerResponses(offlineServerResponseMap, true, dataTableMap, offlineTableName,
+          processingExceptions);
     }
     if (realtimeServerResponseMap != null) {
       numServersResponded += realtimeServerResponseMap.size();
-      deserializeServerResponses(realtimeServerResponseMap, false, dataTableMap, realtimeTableName,
+      totalServerResponseSize += deserializeServerResponses(realtimeServerResponseMap, false, dataTableMap, realtimeTableName,
           processingExceptions);
     }
     phaseTimes.addToDeserializationTime(System.nanoTime() - deserializationStartTime);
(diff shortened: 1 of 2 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/metrics/BrokerMeter.java
```diff
@@ -80,7 +80,8 @@
 
   ROUTING_TABLE_REBUILD_FAILURES("failures", false),
 
-  GROUP_BY_SIZE("queries", false);
+  GROUP_BY_SIZE("queries", false),
+  MAXIMUM_RESPONSE_SIZE("queries", false);
```
mcvsubbu: RESPONSE_SIZE_UPPER_BOUND?

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -511,14 +511,16 @@ private BrokerResponse processOptimizedBrokerRequests(@Nonnull BrokerRequest ori
     int numServersResponded = 0;
     long deserializationStartTime = System.nanoTime();
     Map<ServerInstance, DataTable> dataTableMap = new HashMap<>();
+    long[] maxResponseSize = new long[1];
```
mcvsubbu: Use MutableLong instead, and rename to totalRespSize

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -511,14 +512,17 @@ private BrokerResponse processOptimizedBrokerRequests(@Nonnull BrokerRequest ori
...
       numServersResponded += offlineServerResponseMap.size();
-      deserializeServerResponses(offlineServerResponseMap, true, dataTableMap, offlineTableName, processingExceptions);
+      deserializeServerResponses(offlineServerResponseMap, true, dataTableMap, offlineTableName,
+          processingExceptions, totalResponseSize);
     }
     if (realtimeServerResponseMap != null) {
       numServersResponded += realtimeServerResponseMap.size();
       deserializeServerResponses(realtimeServerResponseMap, false, dataTableMap, realtimeTableName,
-          processingExceptions);
+          processingExceptions, totalResponseSize);
```
mayankshriv: Seems like you have to pass this argument because computation and emitting of metric is happening in different methods. Why can't deserializeServerResponse do the computation and directly emit the metric?

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -543,6 +547,7 @@ private BrokerResponse processOptimizedBrokerRequests(@Nonnull BrokerRequest ori
       _brokerMetrics.addMeteredTableValue(rawTableName, BrokerMeter.BROKER_RESPONSES_WITH_PARTIAL_SERVERS_RESPONDED,
           1L);
     }
+    _brokerMetrics.addMeteredQueryValue(originalBrokerRequest, BrokerMeter.RESPONSE_SIZE_UPPER_BOUND, totalResponseSize.longValue());
```
mayankshriv: Can we rename the metric to be totalServerResponseSize or something more explicit like this? UPPER_BOUND is not so intuitive to understand.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
jackjlli: That's because the response may come from both realtime and offline servers. If we make `deserializeServerResponses` do the computation and emit the metric, the metric only represents either realtime or offline total response size. Thus, I'm thinking whether we sum the totalResponseSize and then emit it in one time.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
jackjlli: Good idea 👍

PR #2659, opened 2018-03-23
Add an interface DataTransformer for general transform on values array

Jackie-Jiang:
1. API in DataTransformer: void transform(inputArray, outputArray, length)
2. Re-implement time unit transformer and date time transformer to use the new interface
3. Add tests for the new time unit transformer and date time transformer


npawar:
No bucketing in case of simple date format was intentional. In simple date format, the bucketing is implicit in the format. Please make sure the original tests pass if you've added it back

Jackie-Jiang:
@npawar Removed the bucketing for SDF. Moved all the existing tests into the new test, and add new cases for bucketing

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/converter/datetime/BaseDateTimeConverter.java
```diff
@@ -0,0 +1,73 @@
...
+    _outputDateTimeFormatter = outputFormat.getDateTimeFormatter();
+    _outputGranularityMillis = outputGranularity.granularityToMillis();
+  }
+
+  protected long convertEpochToMillis(long epochTime) {
+    return _inputTimeUnit.toMillis(epochTime * _inputTimeSize);
+  }
+
+  protected long convertSDFToMillis(@Nonnull String sdfTime) {
+    return _inputDateTimeFormatter.parseMillis(sdfTime);
```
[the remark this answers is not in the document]
Jackie-Jiang: We don't allow null value in executor, and we try to use primitive type as much as possible for performance.

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatUnitSpec.java
```diff
@@ -26,14 +30,66 @@
...
-    MONTHS,
-    WEEKS,
-    DAYS,
-    HOURS,
-    MINUTES,
-    SECONDS,
-    MILLISECONDS;
+    YEARS {
+      @Override
+      public long convert(long millisSinceEpoch, MutableInterval reuse) {
```
mayankshriv: Can we have a more intuitive name than 'reuse'?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/converter/TransformConverter.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.annotation.Nonnull;
+
+
+/**
+ * Interface for transform converter to batch convert array of input values and store the results into the output array.
+ *
+ * @param <I> Type of input array
+ * @param <O> Type of output array
+ */
+public interface TransformConverter<I, O> {
```
mayankshriv: Seems like this does not need to be tied to transform module? If so, can we take it out and rename it something like DataTransformer?

pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFormatUnitSpec.java
Jackie-Jiang: Re-implement the method and removed the parameter

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/function/converter/TransformConverter.java
Jackie-Jiang: Done

PR #2660, opened 2018-03-23
Fixed the issue with replica group mapping znode in the property store

snleee:
1. Make sure to delete when the table is dropped
2. Make sure to add the mapping on the table config update


snleee:
@mcvsubbu 
The test code that I added initially create the table without replica group, update the table config with replica group, and check if the mapping znode is created correctly.

PR #2663, opened 2018-03-23
[TE] root cause acceptance test fix

newsummit:
- Modifying root-cause acceptance test and mock data after template change.
- Adding new props for mock data in `/session/1` call

All 278 test passing


thirdeye/thirdeye-frontend/tests/acceptance/rootcause-test.js
```diff
@@ -1,29 +1,9 @@
 import { module, test } from 'qunit';
 import { setupApplicationTest } from 'ember-qunit';
 import { visit, fillIn, click, currentURL } from '@ember/test-helpers';
+import { rootCauseConst as rcEl } from 'thirdeye-frontend/tests/test-const';
```
aaronucsd: Thanks for moving these consts out Steve. They will be great for other acceptance tests. We might want to pick a different name for this folder `test-const`. Maybe something like `tests/utils/constants.js` to be consistent with the one we have under `apps/utils/constants.js`

newsummit: yes - I like that.

PR #2672, opened 2018-03-27
Compile expression when creating GroupByAstNode

Jackie-Jiang:
On broker side, when creating GroupByAstNode, compile the expression and serialize it back so that:
1. Avoid sending invalid queries to servers
2. The expression is with standard format (no whitespace etc.) in the broker response


Jackie-Jiang:
@mayankshriv It's already covered

PR #2676, opened 2018-03-28
Realtime Segment Assignment interface

npawar:
Add interface for RealtimeSegmentAssignmentStrategy and a PartitionAssignmentGenerator which does not depend on stream partition znode
In our attempts to remove the dependency on znode for stream partition assignment, we will replace StreamPartitionAssigmentGenerator with PartitionAssignmentGenerator. PartitionAssignmentGenerator does not depend on znode for partition assignment, and instead computes it everytime, based on numPartitions, numRpelicas and tagged instances. The tenant config can specify an override, if we need to limit the number of tagged instances that we are allowed to use. Currently we have implemented only the uniform partition assignment strategy. We will extend this to also support other partition assignment schemes like Balanced (using autorebalance), random, etc. We also plan to add an override to read from znode instead of calculating assignment on the fly.


pinot-common/src/main/java/com/linkedin/pinot/common/partition/PartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+
+    // pick a hashed start again, so that even in cases where we use all servers, we don't overload the servers at the start of the list
+    int hashedStartingServer = Math.abs(EqualityUtils.hashCodeOf(tableName)) % numInstancesToUse;
+    for (String partition : partitions) {
+      List<String> instances = new ArrayList<>(numReplicas);
+      for (int r = 0; r < numReplicas; r++) {
+        instances.add(instancesToUse.get(hashedStartingServer++));
+        if (hashedStartingServer == numInstancesToUse) {
```
mcvsubbu: index  = (index+1) % numInstancesToUse  : makes it explicit assignment in one place

pinot-common/src/test/java/com/linkedin/pinot/common/partition/PartitionAssignmentGeneratorTest.java
```diff
@@ -0,0 +1,352 @@
...
+import org.testng.Assert;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import static org.mockito.Mockito.*;
+
+
+public class PartitionAssignmentGeneratorTest {
+
+  private String tableName = "aTableName_REALTIME";
```
mcvsubbu: maybe randomize it so we get different starting points, and invoke the test a few different times?
if you are doing this, always a good idea to System.println() the seed value, so that we can reproduce a failing test case if we want to.
you can add this later if you choose to.

pinot-common/src/test/java/com/linkedin/pinot/common/partition/PartitionAssignmentGeneratorTest.java
```diff
@@ -0,0 +1,352 @@
...
+        idealStateBuilder.transitionToState(2, 0, "ONLINE").addConsumingSegment(2, 1, numReplicas, instances).build();
+    verifyPartitionAssignmentFromIdealState(tableConfig, idealState, numPartitions);
+
+    // 3 partitions 2 replicas, seq 0 has moved to COMPLETED servers, seq 1 on CONSUMING instances
+    instances = Lists.newArrayList("s1_completed", "s2_completed");
+    idealState = idealStateBuilder.moveToServers(0, 0, instances).build();
+    verifyPartitionAssignmentFromIdealState(tableConfig, idealState, numPartitions);
+
+    // status of latest segments OFFLINE/ERROR
+    idealState = idealStateBuilder.transitionToState(1, 1, "ERROR").transitionToState(2, 1, "OFFLINE").build();
```
mcvsubbu: idealstate will never have ERROR in it

pinot-common/src/test/java/com/linkedin/pinot/common/partition/PartitionAssignmentGeneratorTest.java
```diff
@@ -0,0 +1,352 @@
...
+
+  private void verifyPartitionAssignmentFromIdealState(TableConfig tableConfig, IdealState idealState,
+      int numPartitions) {
+    PartitionAssignment partitionAssignmentFromIdealState =
+        _partitionAssignmentGenerator.getPartitionAssignmentFromIdealState(tableConfig, idealState);
+    Assert.assertEquals(tableConfig.getTableName(), partitionAssignmentFromIdealState.getTableName());
+    Assert.assertEquals(partitionAssignmentFromIdealState.getNumPartitions(), numPartitions);
+    // check that latest segments are honoring partition assignment
+    Map<Integer, LLCSegmentName> partitionIdToLatestLLCSegment = new HashMap<>(numPartitions);
+    for (Map.Entry<String, Map<String, String>> entry : idealState.getRecord().getMapFields().entrySet()) {
```
mcvsubbu: makes sense to add a method getLatestSegmentsAndPartitions() in the code, and then use that in the test as well?

pinot-common/src/test/java/com/linkedin/pinot/common/partition/TestIdealStateBuilder.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Map;
+import org.apache.helix.model.IdealState;
+
+import static com.linkedin.pinot.common.utils.CommonConstants.Helix.StateModel.RealtimeSegmentOnlineOfflineStateModel.*;
+
+
+/**
+ * Builder for ideal state for testing
+ */
+public class TestIdealStateBuilder {
```
mcvsubbu: IdealStateBuilderUtil?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/segment/ConsumingSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,55 @@
...
+
+    Map<String, List<String>> segmentAssignment = new HashMap<>(newSegments.size());
+
+    for (String segmentName : newSegments) {
+      if (LLCSegmentName.isLowLevelConsumerSegmentName(segmentName)) {
+        LLCSegmentName llcSegmentName = new LLCSegmentName(segmentName);
+        int partitionId = llcSegmentName.getPartitionId();
+        List<String> instancesListForPartition = partitionAssignment.getInstancesListForPartition(String.valueOf(partitionId));
+        if (instancesListForPartition == null) {
+          throw new IllegalStateException("No partition assignment " + partitionId + " found for segment " + segmentName);
```
mcvsubbu: better to throw a compile exception so the caller is forced to check. If it is new table access or rebalance, we can throw back the exception, but if it is on a dynamic change, we need to update a metric.

pinot-common/src/main/java/com/linkedin/pinot/common/config/RealtimeTagConfig.java
```diff
@@ -49,7 +53,11 @@ public String getCompletedServerTag() {
   }
 
   public boolean isRelocateCompletedSegments() {
-    return _moveCompletedSegments;
+    return _relocateCompletedSegments;
+  }
+
+  public int getMaxConsumingServers() {
```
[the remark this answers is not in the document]
npawar: Removed all code related to num servers to be used

pinot-common/src/test/java/com/linkedin/pinot/common/partition/PartitionAssignmentGeneratorTest.java
npawar: Removed all code related to picking specific set of servers and number of servers to be picked. Will add randomization of tests when we add the logic to pick random servers

pinot-common/src/test/java/com/linkedin/pinot/common/partition/PartitionAssignmentGeneratorTest.java
npawar: added

pinot-common/src/main/java/com/linkedin/pinot/common/partition/PartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,169 @@
...
+      partitions.add(String.valueOf(i));
+    }
+
+    String tableNameWithType = tableConfig.getTableName();
+    int numReplicas = tableConfig.getValidationConfig().getReplicasPerPartitionNumber();
+    RealtimeTagConfig realtimeTagConfig = getRealtimeTagConfig(tableConfig);
+
+    List<String> consumingTaggedInstances = getConsumingTaggedInstances(realtimeTagConfig);
+    if (consumingTaggedInstances.size() < numReplicas) {
+      throw new IllegalStateException(
```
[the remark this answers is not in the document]
npawar: Not sure what you mean by checking for exception when number of stream partitions changes underneath us?

pinot-common/src/main/java/com/linkedin/pinot/common/partition/PartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+    return partitionAssignment;
+  }
+
+  /**
+   * Generates a map of partition id to latest llc segment
+   * @param idealState
+   * @return
+   */
+  public Map<String, LLCSegmentName> getPartitionToLatestSegments(IdealState idealState) {
```
mcvsubbu: pls annotate @VisibleForTesting

pinot-common/src/main/java/com/linkedin/pinot/common/partition/PartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+      }
+    }
+    return partitionIdToLatestSegment;
+  }
+
+  /**
+   * Generates partition assignment for given table, using tagged hosts and num partitions
+   */
+  public PartitionAssignment generatePartitionAssignment(TableConfig tableConfig, int numPartitions) throws Exception {
```
mcvsubbu: pls throw specific compile time exception  instead of a generic one.

pinot-common/src/main/java/com/linkedin/pinot/common/partition/PartitionAssignmentGenerator.java
```diff
@@ -0,0 +1,165 @@
...
+   */
+  private PartitionAssignment uniformAssignment(String tableName, List<String> partitions, int numReplicas,
+      List<String> allInstances) {
+
+    PartitionAssignment partitionAssignment = new PartitionAssignment(tableName);
+
+    Collections.sort(allInstances);
+
+    int numInstances = allInstances.size();
+    int serverId = 0;
```
mcvsubbu: you still want to start from a random point to accommodate multi-tenant tables now, right?

pinot-common/src/test/java/com/linkedin/pinot/common/partition/PartitionAssignmentGeneratorTest.java
mcvsubbu: See other comment, You still need to start from a random start point, in case number of servers exceed number of partitions times number of replicas

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -408,7 +408,7 @@ public String rebalance(
       JSONObject jsonObject = _pinotHelixResourceManager.rebalanceTable(tableName,
           CommonConstants.Helix.TableType.valueOf(tableType.toUpperCase()), rebalanceUserConfig);
       return jsonObject.toString();
-    } catch (JSONException e) {
+    } catch (Exception e) {
```
mcvsubbu: If this is InvalidConfigException, then it is not a 5xx return code right?

pinot-core/src/test/java/com/linkedin/pinot/core/realtime/segment/ConsumingSegmentAssignmentStrategyTest.java
```diff
@@ -0,0 +1,305 @@
...
+   * Tests a segment lifecycle
+   * 1) new CONSUMING segments in a newly created table, residing on consuming servers
+   * 2) consuming segments become ONLINE on the consuming servers and a new sequence of CONSUMING segments is added
+   * 3) ONLINE segments move to completed servers, cCONSUMING segments still on consuming servers
+   * 4) latest CONSUMING segments become OFFLINE
+   * In all these scenarios, we want to check that the final assignment received from the strategy, adheres strictly to what the latest segments say
+   *
+   */
+  @Test
+  public void testSegmentLifecycle() {
```
mcvsubbu: you can add throw Exception here to the test, and that way you don't have to catch all exceptions that you don't expect. Test will fail if exceptions are thrown.

PR #2682, opened 2018-03-29
Resolve maven dependency conflicts

jackjlli:
This PR resolves maven dependency conflicts.

### DIFF
--- a/pinot-api/pom.xml
+++ b/pinot-api/pom.xml
@@ -56,10 +56,26 @@
     <dependency>
       <groupId>com.ning</groupId>
       <artifactId>async-http-client</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>netty</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty</artifactId>
     </dependency>
     <dependency>
       <groupId>com.101tec</groupId>
       <artifactId>zkclient</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>io.netty</groupId>
+          <artifactId>netty</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/pinot-common/pom.xml
+++ b/pinot-common/pom.xml
@@ -76,6 +76,12 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>commons-logging</groupId>
+          <artifactId>commons-logging</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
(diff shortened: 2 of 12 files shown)


Jackie-Jiang:
Good job. But after excluding the dependencies, you need to explicitly pull the dependencies in right? Please always pick the latest version from the excluded dependencies.

jackjlli:
@Jackie-Jiang Ah, gotcha! Thanks for reminding! 👍 
Just addressed your comment. 😉

pinot-common/pom.xml
```diff
@@ -175,11 +191,32 @@
...
+        <exclusion>
+          <groupId>org.apache.commons</groupId>
+          <artifactId>commons-math3</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-math3</artifactId>
+      <version>${commons-math3.version}</version>
```
Jackie-Jiang: Don't include version in child level pom file. Only define them in the root level pom file. Same for other places

pinot-api/pom.xml
```diff
@@ -56,6 +56,12 @@
     <dependency>
       <groupId>com.ning</groupId>
       <artifactId>async-http-client</artifactId>
+      <exclusions>
```
Jackie-Jiang: I think after excluding this dependency, you have to explicitly pull in a dependency or it might not work right?

pom.xml
```diff
@@ -117,6 +117,8 @@
     <jersey.version>2.23</jersey.version>
     <swagger.version>1.5.10</swagger.version>
     <hadoop.version>2.7.0</hadoop.version>
+    <commons-lang.version>2.6</commons-lang.version>
```
Jackie-Jiang: Specify the dependency with version info here.

pinot-api/pom.xml
jackjlli: If we print out the dependency tree, we'll find out there's another same jar but with higher version in the same module.
```[INFO] com.linkedin.pinot:pinot-api:jar:0.016
[INFO] +- com.ning:async-http-client:jar:1.9.21:compile
[INFO] |  \- io.netty:netty:jar:3.10.1.Final:compile
[INFO] \- com.101tec:zkclient:jar:0.7:compile
[INFO]    \- org.apache.zookeeper:zookeeper:jar:3.4.9:compile (version managed from 3.4.6)
[INFO]       \- (io.netty:netty:jar:3.10.5.Final:compile - omitted for conflict with 3.10.1.Final)
```
But after excluding it from `com.ning:async-http-client`, the dependency tree now looks like:
```
[INFO] com.linkedin.pinot:pinot-api:jar:0.016
[INFO] \- com.101tec:zkclient:jar:0.7:compile
[INFO]    \- org.apache.zookeeper:zookeeper:jar:3.4.9:compile (version managed from 3.4.6)
[INFO]       \- io.netty:netty:jar:3.10.5.Final:compile
```
maven will traverse the tree and then find the  io.netty:netty:jar in deeper level and ultimately imports that `3.10.5.Final` jar, which is what we want.

But yeah, the strategy of explicitly putting dependency in higher level might be the better approach. I'll adjust the code. 👍

jackjlli:
This PR has been verified in local machine. Everything works fine.

PR #2685, opened 2018-03-30
[TE] frontend = aaronucsd/removed error default state

aaronucsd:


apucher:
I wonder if this is actually the root-cause :-) It seems like the RCA controller should catch error by itself and fix it.

aaronucsd:
@apucher Yeah, my error catch all for any  route that does not handle it itself. So the fix really is why it broke and gracefull add code to handle the error. But you said the `catch-all `return home` link was preventing so i can remove for now. But this is meant to catch any error that bubbled all the why up.

aaronucsd:
I don't have much context yet until i research into it more..Maybe leave the error substate for now?

apucher:
yes. i'd leave this in for now and instead add some "message-and-reset-to-blank" error-handling in the rootcause route

npawar:

PR #2686, opened 2018-03-30
Adding minion purge event emitter

jenniferdai:


pinot-minion/src/main/java/com/linkedin/pinot/minion/events/DefaultMinionEventNotifier.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package com.linkedin.pinot.minion.events;
+
+import com.linkedin.pinot.minion.executor.SegmentConversionInfo;
+
+
+public class DefaultMinionEventNotifier implements MinionEventNotifier {
+  public void notifyOnPurge(SegmentConversionInfo segmentConversionInfo) {
+
```
jfim: Add a comment that explicitly says that nothing is to be done, eg. `// Nothing to do`

pinot-core/src/main/java/com/linkedin/pinot/core/minion/SegmentPurger.java
```diff
@@ -29,6 +29,8 @@
 import com.linkedin.pinot.core.segment.index.SegmentMetadataImpl;
 import java.io.File;
 import java.io.IOException;
+import java.util.Collections;
```
jfim: Do we need these imports?

pinot-minion/src/main/java/com/linkedin/pinot/minion/events/MinionEventNotifier.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package com.linkedin.pinot.minion.events;
+
+import com.linkedin.pinot.common.config.TableConfig;
+import com.linkedin.pinot.common.data.Schema;
+import com.linkedin.pinot.minion.executor.SegmentConversionInfo;
+
+
+public interface MinionEventNotifier {
```
jfim: Add Javadoc that explains what this interface is for and when the `notifyOnPurge` method is invoked.

pinot-minion/src/main/java/com/linkedin/pinot/minion/events/MinionEventNotifier.java
```diff
@@ -0,0 +1,26 @@
...
+
+  void notifyOnPurge(SegmentConversionInfo segmentConversionInfo);
```
jfim: It would be more useful to have an interface that is something like `notifyMinionTaskStart/notifyMinionTaskEnd` as opposed to something very specific to the purge task.

pinot-minion/src/main/java/com/linkedin/pinot/minion/events/MinionEventNotifierFactory.java
```diff
@@ -0,0 +1,47 @@
...
+public abstract class MinionEventNotifierFactory {
+  public static final Logger LOGGER = LoggerFactory.getLogger(MinionEventNotifierFactory.class);
+  public static final String METADATA_EVENT_CLASS_CONFIG = "factory.class";
+
+  public abstract void init(Configuration configuration);
+
+  public abstract MinionEventNotifier create();
+
+  public static MinionEventNotifierFactory loadFactory(Configuration configuration) {
+    MinionEventNotifierFactory metadataEventNotifierFactory;
```
jfim: Would be cool if we could have more than one `MinionEventNotifier`, that way we could implement one that logs task executions in Kafka (and read them back in Pinot!), one that emails on task failures, and so on.

pinot-minion/src/main/java/com/linkedin/pinot/minion/events/MinionEventNotifier.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package com.linkedin.pinot.minion.events;
+
+import com.linkedin.pinot.minion.executor.SegmentConversionInfo;
+
+
+/**
+ * Emits events that provide information about the state of minion tasks.
+ */
+public interface MinionEventNotifier {
```
jfim: `MinionTaskObserver` would be a more appropriate name (see observer pattern)

pinot-minion/src/main/java/com/linkedin/pinot/minion/events/MinionEventObserver.java
```diff
@@ -21,16 +21,16 @@
 public interface MinionEventObserver {
 
   /**
-   * Invoked when a minion task begins
+   * Invoked when a minion job begins
    * @param taskInformation
    * @param taskType
    */
-  void notifyMinionTaskStart(Object taskInformation, String taskType);
+  void notifyMinionJobStart(Object taskInformation, String taskType);
```
jfim: I would swap the order of these arguments, so that it's type, then information.

PR #2695, opened 2018-04-02
Add TableQueryQuotaManager to enable rate limiter on broker

jackjlli:
This PR adds TableQueryQuotaManager enable rate limiter on broker.
Unit tests added.


pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+    Map<String, String> stateMap = brokerResource.getStateMap(tableName);
+    int onlineCount = 0;
+    for (String state : stateMap.values()) {
+      if (state.equals(CommonConstants.Helix.StateModel.SegmentOnlineOfflineStateModel.ONLINE)) {
+        onlineCount++;
+      }
+    }
+    // Get the dynamic rate
+    double rate = Double.parseDouble(quotaConfig.getQps()) / onlineCount;
```
jfim: Does `NumberFormatException` get caught anywhere? If not, it will prevent a table with an incorrect table configuration from being served.

Also, what happens if there are no servers online? Or if the table is not in the external view?

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,138 @@
...
+    RateLimiter rateLimiter = RateLimiter.create(rate);
+    _rateLimiterMap.put(rawTableName, rateLimiter);
+    LOGGER.info("Rate limiter with Rate: {} has been initialized for Table: {}", rate, rawTableName);
```
jfim: Should log the overall rate, per-broker rate, and number of online instances.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Acquire a token from rate limiter based on the table name.
+   * @param tableName raw table name
+   * @return true if there is no query quota or a token is acquired, otherwise return false
+   * */
+  public boolean acquire(String tableName) {
+    RateLimiter rateLimiter = _rateLimiterMap.get(tableName);
+    return rateLimiter == null || rateLimiter.tryAcquire();
```
jfim: Expand this to an `if`/`else` block.

pinot-common/src/main/java/com/linkedin/pinot/common/config/QuotaConfig.java
```diff
@@ -33,8 +33,10 @@
 public class QuotaConfig {
   private static final Logger LOGGER = LoggerFactory.getLogger(QuotaConfig.class);
   private static final String STORAGE_FIELD_NAME = "storage";
+  private static final String QPS_FIELD_NAME = "qps";
 
   private String _storage;
+  private String _qps;
```
jfim: Should be `maxQueriesPerSecond`

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -221,6 +225,15 @@ public BrokerResponse handleRequest(@Nonnull JSONObject request, RequesterIdenti
       return BrokerResponseFactory.getStaticNoTableHitBrokerResponse(ResponseType.BROKER_RESPONSE_TYPE_NATIVE);
     }
 
+    // Validate qps quota.
+    if (!_tableQueryQuotaManager.acquire(rawTableName)) {
```
kishoreg: No need to log the pql. Just log the requestId. If possible, lets try to have the return at the end of the method. Avoids a bunch of duplicate code and makes it easy to read.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
jackjlli: Good point! I'll adjust the code and add unit tests for it. 👍

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,138 @@
...
+      QuotaConfig quotaConfig = tableConfig.getQuotaConfig();
+      if (quotaConfig != null && quotaConfig.getQps() != null) {
+        createRateLimiter(tableName, rawTableName, brokerResource, quotaConfig);
+      } else {
+        String realtimeTableName = TableNameBuilder.REALTIME.tableNameWithType(rawTableName);
+        TableConfig realTimeTableConfig = ZKMetadataProvider.getTableConfig(_propertyStore, realtimeTableName);
+        if (realTimeTableConfig == null) {
+          return;
+        }
+        QuotaConfig realTimeTableQuotaConfig = realTimeTableConfig.getQuotaConfig();
```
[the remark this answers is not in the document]
jackjlli: That's because we want to reuse the `QuotaConfig` class by adding `qps` field in it, while `QuotaConfig` field is under `TableConfig`. And `TableConfig` belongs to the table with a specific type(real-time or offline).  We can discuss it to find a proper place to store this `qps` field.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,138 @@
...
+    Map<String, String> stateMap = brokerResource.getStateMap(tableName);
+    int onlineCount = 0;
+    for (String state : stateMap.values()) {
+      if (state.equals(CommonConstants.Helix.StateModel.SegmentOnlineOfflineStateModel.ONLINE)) {
+        onlineCount++;
+      }
+    }
+    // Get the dynamic rate
+    double rate = Double.parseDouble(quotaConfig.getQps()) / onlineCount;
+    RateLimiter rateLimiter = RateLimiter.create(rate);
```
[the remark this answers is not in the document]
jackjlli: Then we need to adjust the value of `qps` field to be sth like `10qps` or `10 query/minute`. Let me create a doc to discuss this and the like.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,138 @@
...
+import org.slf4j.LoggerFactory;
+
+
+public class TableQueryQuotaManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(TableQueryQuotaManager.class);
+
+  private final ZkHelixPropertyStore<ZNRecord> _propertyStore;
+  private final Map<String, RateLimiter> _rateLimiterMap;
+
+  public TableQueryQuotaManager(ZkHelixPropertyStore<ZNRecord> propertyStore) {
```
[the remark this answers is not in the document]
jackjlli: Just noticed that `PinotHelixResourceManager` belongs to pinot-controller. 
So I think for now it'd be ok to just pass the propertyStore as the param into the constructor, just like how `HelixExternalViewBasedRouting` does.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.broker.queryquota;
+
+public class HitCounter {
```
jfim: Can you add a description of what it is that you're doing in this class? From reading the code, it seems like you're averaging the number of queries over the last 100 time buckets.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,58 @@
...
+    _timeRangeInSecond = timeRangeInSecond;
+    _timeUnit = _timeRangeInSecond * 1000 / TIME_WINDOW;
+    _times = new long[TIME_WINDOW];
+    _hits = new long[TIME_WINDOW];
+  }
+
+  public void hit(long timestamp) {
+    long numTimeUnits = timestamp / _timeUnit;
+    int index = (int)(numTimeUnits % TIME_WINDOW);
+    if (_times[index] == numTimeUnits) {
```
jfim: Seems like you'd need a compare-and-swap op for multithreaded correctness here.

jackjlli: Good catch! 👍 I'll use `AtomicInteger` in this case. Thx!

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * This hit counter is for counting the number of hits within a range of time. Right now the granularity we use is second.
+ * In order to save the space and time, we store the number of hits over the last 100 time buckets. When the method hit
+ * gets called, we put the timestamp to the specified bucket. When the method getHits gets called, we sum all the number
+ * of hits within the last 100 time buckets.
+ * */
+public class HitCounter {
+
+  private static int TIME_WINDOW = 100;
```
jfim: This should be called `BUCKET_COUNT`

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,74 @@
...
+  private int _timeRangeInSecond;
+  private int _timeUnit;
```
jfim: `timeBucketWidthMs`

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,74 @@
...
+  private long[] _times;
```
jfim: `bucketStartTime`

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,74 @@
...
+  private AtomicInteger[] _hits;
```
jfim: `bucketHitCount`

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,74 @@
...
+
+  public HitCounter(int timeRangeInSecond) {
+    _timeRangeInSecond = timeRangeInSecond;
```
jfim: Remove this if it's not used, we can always add it later if needed.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,74 @@
...
+    int index = (int)(numTimeUnits % TIME_WINDOW);
+    if (_times[index] == numTimeUnits) {
+      _hits[index].incrementAndGet();
+    } else {
+      _times[index] = numTimeUnits;
+      _hits[index].set(1);
+    }
+  }
+
+  public int getHits(long timestamp) {
```
jfim:
```
public int getHitCount() { getHitCount(System.currentTimeMillis()); }

@VisibleForTesting
int getHitCount(long millis) { ... }
```

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
jfim: Your `if` statement is dependent on all threads doing no stale reads and no parallel execution of the `if`, which is not the case.

Quick test for multithreaded correctness:

```
public class TestHitCounter {
  public static void main(String[] args) {
    int expectedHitCount = 30 * 200000;
    HitCounter hitCounter = new HitCounter(10);

    List<Thread> threadList = new ArrayList<>();


    for (int i = 0; i < 30; i++) {
      Thread thread = new Thread() {
        @Override
        public void run() {
          for (int j = 0; j < 200000; j++) {
            hitCounter.hit(System.currentTimeMillis());
          }
        }
      };
      thread.start();
      threadList.add(thread);
    }

    for (Thread thread : threadList) {
      Uninterruptibles.joinUninterruptibly(thread);
    }

    System.out
        .println("hitCounter.getHits(System.currentTimeMillis()) = " + hitCounter.getHits(System.currentTimeMillis()));
  }
}
```

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/QueryQuotaConfig.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.google.common.util.concurrent.RateLimiter;
+import javax.annotation.Nonnull;
+
+
+public class QueryQuotaConfig {
+
+  private RateLimiter _rateLimiter;
+  private HitCounter _hitCounter;
+  private int _numTables;
```
jfim: `tableReferenceCount` and/or refcount

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  /**
+   * Get table name with type.
+   * @param rawTableName table name without table type.
+   * @param tableType table type: offline or real-time.
+   * @return table name with type.
+   * */
+  private String getTableNameWithType(String rawTableName, CommonConstants.Helix.TableType tableType) {
+    return tableType == CommonConstants.Helix.TableType.OFFLINE ?
```
jfim: `TableNameBuilder.forType(tableType).tableNameWithType(rawTableName)`

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/HitCounter.java
```diff
@@ -0,0 +1,87 @@
...
+ * of hits within the last 100 time buckets.
+ * */
+public class HitCounter {
+
+  private static int BUCKET_COUNT = 100;
+  private int _timeBucketWidthMs;
+  private final AtomicLongArray _bucketStartTime;
+  private AtomicIntegerArray _bucketHitCount;
+
+  public HitCounter(int timeRangeInSecond) {
```
jfim: `timeRangeInSeconds`

pinot-broker/src/test/java/com/linkedin/pinot/broker/queryquota/HitCounterTest.java
```diff
@@ -0,0 +1,117 @@
...
+      currentTimestamp += 1L;
+    }
+    Assert.assertNotNull(hitCounter);
+    Assert.assertEquals(hitCounter.getHitCount(currentTimestamp), 7);
+  }
+
+  @Test
+  public void testConcurrency() {
+    Random random = new Random();
+    // Run the test 5 times
```
jfim: Is it run 3 or 5 times? :)

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,267 @@
...
+    LOGGER.info("Rate limiter for table: {} has been initialized. Overall rate: {}. Per-broker rate: {}. Number of online broker instances: {}",
+        tableName, overallRate, perBrokerRate, onlineCount);
+  }
+
+  /**
+   * Acquire a token from rate limiter based on the table name.
+   * @param tableName original table name which could be raw.
+   * @return true if there is no query quota specified for the table or a token can be acquired, otherwise return false.
+   * */
+  public boolean acquire(String tableName, String offlineTableName, String realtimeTableName) {
```
mcvsubbu: we do we pass 3 table names in the argument? why not just pass the raw table name and construct the other two s needed?

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,267 @@
...
+    LOGGER.info("Trying to acquire token for table: {}", tableName);
```
mcvsubbu: I am assuming this will be logged for each request? If so, please remove the log or make it debug

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,267 @@
...
+    QueryQuotaConfig realtimeTableQueryQuotaConfig = null;
+    if (offlineTableName != null) {
+      offlineTableQueryQuotaConfig = _rateLimiterMap.get(offlineTableName);
+    }
+    if (realtimeTableName != null) {
+      realtimeTableQueryQuotaConfig = _rateLimiterMap.get(realtimeTableName);
+    }
+
+    if (offlineTableQueryQuotaConfig == null && realtimeTableQueryQuotaConfig == null) {
+      LOGGER.info("No qps quota is specified for table: {}", tableName);
```
mcvsubbu: same comment for logging, will not repeat

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,267 @@
...
+   * @return true if there's no qps quota for that table, or a token is acquired successfully.
+   * */
+  private boolean tryAcquireToken(String tableName, QueryQuotaConfig queryQuotaConfig) {
+    // Use hit counter to count the number of hits.
+    queryQuotaConfig.getHitCounter().hit();
+
+    RateLimiter rateLimiter = queryQuotaConfig.getRateLimiter();
+    double perBrokerRate = rateLimiter.getRate();
+    if (!rateLimiter.tryAcquire()) {
+      LOGGER.error("Quota is exceeded for table: {}. Per-broker rate: {}", tableName, perBrokerRate);
```
mcvsubbu: this is not an ERROR log. should be info.

Also, bump a metric when quota is exceeded

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,267 @@
...
+      return false;
+    }
+
+    // Emit the qps capacity utilization rate.
+    if (_brokerMetrics != null) {
+      int numHits = queryQuotaConfig.getHitCounter().getHitCount();
+      int percentageOfCapacityUtilization = (int)(numHits * 100 / perBrokerRate);
+      LOGGER.info("The percentage of rate limit capacity utilization is {}", percentageOfCapacityUtilization);
```
mcvsubbu: no need to log here

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
jackjlli: Yup, a metric called "QUERY_QUOTA_EXCEEDED" will be emitted in `BrokerRequestHandler`.

pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/BrokerResourceOnlineOfflineStateModelFactory.java
```diff
@@ -113,7 +127,12 @@ public void onBecomeDroppedFromOffline(Message message, NotificationContext cont
       try {
         LOGGER.info("BrokerResourceOnlineOfflineStateModel.onBecomeDroppedFromOffline() : " + message);
         String tableName = message.getPartitionName();
+        TableConfig tableConfig = ZKMetadataProvider.getTableConfig(_propertyStore, tableName);
+
         _helixExternalViewBasedRouting.markDataResourceOffline(tableName);
+        _tableQueryQuotaManager.dropTableQueryQuota(
+            tableConfig,
```
jfim: Dropping the table query quota should really only depend on the table name, not any other thing (external view, table configuration). If there is no quota configuration, by definition, dropping the table query quota does nothing. This also ensures that `dropTableQueryQuota` will always behave as expected, even if there are odd edge cases where the table quota configuration is changed while the broker is running.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,265 @@
...
+  public TableQueryQuotaManager(HelixManager helixManager) {
+    _helixManager = helixManager;
+    _rateLimiterMap  = new ConcurrentHashMap<>();
+  }
+
+  /**
+   * Initialize dynamic rate limiter with table query quota.
+   * @param tableConfig table config.
+   * @param brokerResource broker resource which stores all the broker states of each table.
+   * */
```
jfim: No `* *` on the last line

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,265 @@
...
+   * @param rawTableName table name without table type.
+   * @param tableType table type: offline or real-time.
+   * @return QuotaConfig, which could be null.
+   * */
+  private QuotaConfig getQuotaConfigFromPropertyStore(String rawTableName, CommonConstants.Helix.TableType tableType) {
+    ZkHelixPropertyStore<ZNRecord> propertyStore = _helixManager.getHelixPropertyStore();
+
+    String tableNameWithType = TableNameBuilder.forType(tableType).tableNameWithType(rawTableName);
+    TableConfig tableConfig = ZKMetadataProvider.getTableConfig(propertyStore, tableNameWithType);
+    if (tableConfig == null) {
```
jfim: Should log a warning if we can't fetch the table configuration, this really shouldn't happen.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,265 @@
...
+    if (brokerResource == null) {
+      LOGGER.warn("Failed to init qps quota for table {}. No broker resource connected!", tableName);
+      return;
+    }
+
+    Map<String, String> stateMap = brokerResource.getStateMap(tableName);
+    if (stateMap == null) {
+      LOGGER.warn("Failed to init qps quota: table {} is not in the broker resource!", tableName);
+      return;
+    }
```
jfim: It's possible that if there are no other brokers serving this table, the table won't be part of the external view. What you want to do here is to count the number of brokers currently serving this table (which does not include your instance, assuming no zk/helix latency issues).

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,265 @@
...
+    int onlineCount = 0;
+    for (String state : stateMap.values()) {
+      if (state.equals(CommonConstants.Helix.StateModel.SegmentOnlineOfflineStateModel.ONLINE)) {
```
jfim: Here you want to check that the state is `ONLINE` and that the instance name is not your instance name (which could happen if the external view is stale).

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,265 @@
...
+        onlineCount++;
+      }
+    }
+    LOGGER.info("The number of online brokers for table {} is {}", tableName, onlineCount);
+
+    // assert onlineCount > 0;
+    if (onlineCount == 0) {
```
jfim: Update with previous feedback, as per discussion.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+    int onlineCount = 0;
+    for (String state : stateMap.values()) {
+      if (state.equals(CommonConstants.Helix.StateModel.SegmentOnlineOfflineStateModel.ONLINE)) {
+        onlineCount++;
+      }
+    }
+    LOGGER.info("The number of online brokers for table {} is {}", tableName, onlineCount);
+
+    // assert onlineCount > 0;
```
jfim: Remove commented out code.

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,265 @@
...
+    QueryQuotaConfig offlineTableQueryQuotaConfig = _rateLimiterMap.get(offlineTableName);
+    QueryQuotaConfig realtimeTableQueryQuotaConfig = _rateLimiterMap.get(realtimeTableName);
+
+    if (offlineTableQueryQuotaConfig == null && realtimeTableQueryQuotaConfig == null) {
+      LOGGER.debug("No qps quota is specified for table: {}", tableName);
+      return true;
+    }
+
+    // Try to acquire token for offline table if there's a qps quota.
+    if (offlineTableQueryQuotaConfig != null && !tryAcquireToken(tableName, offlineTableQueryQuotaConfig)) {
```
jfim: Deduct tokens from both quotas, what you can do is:
```
bool offlineQuotaOk = offlineTableQueryQuotaConfig == null || tryAcquireToken(...);
...
return offlineQuotaOk && realtimeQuotaOk;
```

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -0,0 +1,265 @@
...
+      LOGGER.info("Quota is exceeded for table: {}. Per-broker rate: {}", tableName, perBrokerRate);
+      return false;
+    }
+
+    // Emit the qps capacity utilization rate.
+    if (_brokerMetrics != null) {
+      int numHits = queryQuotaConfig.getHitCounter().getHitCount();
+      int percentageOfCapacityUtilization = (int)(numHits * 100 / perBrokerRate);
+      LOGGER.debug("The percentage of rate limit capacity utilization is {}", percentageOfCapacityUtilization);
+      _brokerMetrics.addMeteredTableValue(tableName, BrokerMeter.QUERY_QUOTA_CAPACITY_UTILIZATION_RATE, percentageOfCapacityUtilization);
```
jfim: You want to use `setValueOfTableGauge` here, since adding percentages doesn't really make sense.

pinot-broker/src/test/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManagerTest.java
```diff
@@ -0,0 +1,351 @@
...
+  public void testOfflineTableWithNullQuotaAndNoRealtimeTableConfig() throws Exception {
+    ExternalView brokerResource = generateBrokerResource(OFFLINE_TABLE_NAME);
+    TableConfig tableConfig = generateDefaultTableConfig(OFFLINE_TABLE_NAME);
+    _tableQueryQuotaManager.initTableQueryQuota(tableConfig, brokerResource);
+    Assert.assertEquals(_tableQueryQuotaManager.getRateLimiterMapSize(), 0);
+  }
+
+  @Test
+  public void testOfflineTableWithNullQuotaButWithRealtimeTableConfigNullQpsConfig() throws Exception {
+    ZNRecord znRecord = new ZNRecord("broker_instance_1");
```
jfim: You probably want to build the table configuration here, check `TableConfig.Builder`

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
jackjlli: This method is used to fetch the table config with another table type. E.g. if the the table type is offline, then we try to fetch the real-time table config, and wise versa. And it's possible that there's only one table config in the property store.

PR #2696, opened 2018-04-03
Disable metric aggregation when metrics have dictionary or dimensions dont.

mayankshriv:
The current implementation of metrics aggregation requires:
1. Metrics are not dictionary encoded, aggregating metrics in presence
of dictionary encoding is a bit complex.
2. Dimensions are dictionary encoded metrics are not dictionary encode.
This is because dictionary id's are used key for mapping to a doc id.

Both these limitations can be eliminated in future. Until then, adding
check to disable metric aggregation under these cases.


Jackie-Jiang:
Please check the failed test and fix either the test or the code

PR #2697, opened 2018-04-03
Add LimitZeroSelectionOperator for cases with limit <= 0

Jackie-Jiang:
Do not filter docs when selection limit <= 0
Fix the exception from selection order-by with limit < -1


pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/LimitZeroSelectionOperator.java
```diff
@@ -0,0 +1,57 @@
...
+import com.linkedin.pinot.core.operator.BaseOperator;
+import com.linkedin.pinot.core.operator.ExecutionStatistics;
+import com.linkedin.pinot.core.operator.blocks.IntermediateResultsBlock;
+import com.linkedin.pinot.core.query.selection.SelectionOperatorUtils;
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.List;
+
+
+public class LimitZeroSelectionOperator extends BaseOperator<IntermediateResultsBlock> {
```
mayankshriv: Trying to understand, why do we need a separate operator for this? Seems too much work for a special case. Shouldn't the existing operators take care of this case?

mayankshriv: Also, name could be more generic like EmptySelectionOperator?

Jackie-Jiang: This operator is specific for selection with limit less or equal to 0. We don't need to fetch any data, so we don't need to create an underlying operator. The branch logic happens in the SelectionPlanNode, and with a separate operator, the logic is more clear.
I agree on the name EmptySelectionOperator.

PR #2703, opened 2018-04-05
Adding DateTime UDF (leverages existing datetimeconvert udf). This wi…

kishoreg:
…ll be used to do convert time during ingestion.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
@@ -28,7 +28,9 @@ import org.codehaus.jackson.annotate.JsonIgnoreProperties;
 public final class DateTimeFieldSpec extends FieldSpec {
   private String _format;
   private String _granularity;
-
+  private DateTimeFormatSpec _formatSpec;
+  private DateTimeGranularitySpec _granularitySpec;
+  
   public enum TimeFormat {
     EPOCH,
     SIMPLE_DATE_FORMAT
@@ -99,7 +101,9 @@ public final class DateTimeFieldSpec extends FieldSpec {
 
   // Required by JSON de-serializer. DO NOT REMOVE.
   public void setFormat(@Nonnull String format) {
+    Preconditions.checkArgument(DateTimeFormatSpec.isValidFormat(format));
     _format = format;
+    _formatSpec = new DateTimeFormatSpec(format);
   }
 
   @Nonnull
@@ -109,7 +113,9 @@ public final class DateTimeFieldSpec extends FieldSpec {
 
   // Required by JSON de-serializer. DO NOT REMOVE.
   public void setGranularity(@Nonnull String granularity) {
+    Preconditions.checkArgument(DateTimeGranularitySpec.isValidGranularity(granularity));
     _granularity = granularity;
+    _granularitySpec = new DateTimeGranularitySpec(granularity);
   }
 
   @Nonnull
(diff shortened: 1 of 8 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/data/DateTimeFieldSpec.java
```diff
@@ -109,7 +113,9 @@ public String getGranularity() {
 
   // Required by JSON de-serializer. DO NOT REMOVE.
   public void setGranularity(@Nonnull String granularity) {
+    Preconditions.checkArgument(DateTimeGranularitySpec.isValidGranularity(granularity));
     _granularity = granularity;
+    _granularitySpec = new DateTimeGranularitySpec(granularity);
```
mayankshriv: Seems like JSON serde should work with this change, but would be good to ensure via a unit test.

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
```diff
@@ -190,6 +203,41 @@ public String getTransformFunction() {
...
+    if(_columnsInTransformFunction != null) {
+      return _columnsInTransformFunction;
+    }
+    if (_columnsInTransformFunction == null && _transformFunction != null) {
+      TransformExpressionTree expressionTree = Pql2Compiler.get().compileToExpressionTree(_transformFunction);
+      _columnsInTransformFunction = new HashSet<>();
+      traverse(expressionTree, _columnsInTransformFunction);
+      return _columnsInTransformFunction;
+    } else {
+      _columnsInTransformFunction = Collections.emptySet();
```
mayankshriv: Is this for column with fixed value? If not, when would this happen?

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/Pql2Compiler.java
```diff
@@ -55,6 +55,19 @@
 @ThreadSafe
 public class Pql2Compiler implements AbstractCompiler {
 
+  /**
+   * PQL compilers are not thread safe. Creating one per thread
```
mayankshriv: I think PQL compiler is now threadsafe, which is why the @ThreadSafe annotation was added at line 55.

pinot-core/src/main/java/com/linkedin/pinot/core/data/udfs/DateTimeFunctions.java
```diff
@@ -0,0 +1,115 @@
...
+
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
+
+import com.linkedin.pinot.core.operator.transform.transformer.datetime.BaseDateTimeTransformer;
+import com.linkedin.pinot.core.operator.transform.transformer.datetime.DateTimeTransformerFactory;
+public class DateTimeFunctions {
+  private static ThreadLocal<long[]> longInputArrayCache = new ThreadLocal<long[]>(){
+    protected long[] initialValue() {
+      return new long[1];
```
mayankshriv: If the arrays are always of size 1, then why do we need array?

pinot-core/src/test/java/com/linkedin/pinot/core/data/function/FunctionExpressionEvaluatorTest.java
```diff
@@ -24,7 +24,9 @@
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+import com.linkedin.pinot.common.request.transform.TransformExpressionTree;
```
mayankshriv: Only import change in the test? Would be good to add so unit tests for the new code.

PR #2706, opened 2018-04-05
[TE] rootcause - trend analysis tab

apucher:
We're bringing back the trend analysis tab from the legacy RCA page, but without the extra-long scroll bar.

* trend analysis tab

* dynamic time granularity and paging

* incremental loading

* external links

image

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
@@ -219,7 +219,6 @@ export default Component.extend({
       const changes = {};
       const metricUrns = filterPrefix(Object.keys(entities), 'thirdeye:metric:');
 
-      console.log('changes', metricUrns);
       metricUrns.forEach(urn => {
         const currUrn = toCurrentUrn(urn);
         const baseUrn = toBaselineUrn(urn);
@@ -246,8 +245,6 @@ export default Component.extend({
         });
       });
 
-      console.log('changes done', changes);
-
       return changes;
     }
   ),
@@ -268,7 +265,6 @@ export default Component.extend({
 
       const metricUrns = filterPrefix(Object.keys(entities), 'thirdeye:metric:');
 
-      console.log('data', metricUrns);
       const rows = metricUrns.map(urn => {
         const row = {
           urn,
@@ -288,8 +284,6 @@ export default Component.extend({
 
         return row;
       });
-
-      console.log('data done', rows);
 
       return rows;
     }
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
@@ -1,10 +1,8 @@
-import { computed, observer } from '@ember/object';
-import { later } from '@ember/runloop';
+import { computed } from '@ember/object';
 import Component from '@ember/component';
 import {
   toCurrentUrn,
   toBaselineUrn,
-  toOffsetUrn,
   hasPrefix,
   filterPrefix,
   toMetricLabel,
@@ -68,20 +66,6 @@ export default Component.extend({
   //
   // internal properties
   //
-
-  /**
-   * Display flag for table. Hack for ember models table only re-rendering columns on init()
-   * @type {bool}
-   */
-  showTable: true,
-
-  /**
-   * Hack for ember models table only re-rendering columns on init()
-   */
-  _columnObserver: observer('columns', function () {
-    this.set('showTable', false);
-    later(this, () => this.set('showTable', true), 0);
-  }),
 
   /**
    * Start time for time table, overridden by drop down
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
@@ -1,0 +1,325 @@
+import { computed, observer } from '@ember/object';
+import { later } from '@ember/runloop';
+import Component from '@ember/component';
+import {
+  toCurrentUrn,
+  toBaselineUrn,
+  toOffsetUrn,
+  hasPrefix,
+  filterPrefix,
+  toMetricLabel,
+  isInverse,
+  toColorDirection,
+  makeSortable
+} from 'thirdeye-frontend/utils/rca-utils';
+import {
+  humanizeChange,
+  humanizeFloat,
+  humanizeScore
+} from 'thirdeye-frontend/utils/utils';
+import moment from 'moment';
+import _ from 'lodash';
+
+const ROOTCAUSE_TREND_MAX_COLUMNS = 12;
+
+export default Component.extend({
+  classNames: ['rootcause-metrics'],
+
+  //
+  // external properties
+  //
+
+  /**
+   * Entities cache
+   * @type {object}
+   */
+  entities: null,
+
+  /**
+   * Metric aggregates
+   * @type {object}
+   */
+  timeseries: null,
+
+  /**
+   * Investigation context
+   * @type {object}
+   */
+  context: null,
+
+  /**
+   * User-selected urns
+   * @type {Set}
+   */
+  selectedUrns: null,
+
+  /**
+   * Callback on metric selection
+   * @type {function}
+   */
+  onSelection: null, // function (Set, state)
+
+  /**
+   * loading status for component
+   * @type {boolean}
+   */
+  isLoading: false,
+
+  //
+  // internal properties
+  //
+
+  /**
+   * Display flag for table. Hack for ember models table not re-rendering columns on init()
+   * @type {bool}
+   */
+  showTable: true,
+
+  /**
+   * Observer hack for ember models table only re-rendering columns on init()
+   */
+  _columnObserver: observer(
+    'columns',
+    function () {
+      this.set('showTable', false);
+      later(this, () => this.set('showTable', true), 1000);
+    }
+  ),
+
+  /**
+   * Time buckets for columns
+   * @type {int[]}
+   */
+  buckets: computed(
+    'context',
+    function () {
+      const { context } = this.getProperties('context');
+
+      const buckets = [];
+      const [stepSize, stepUnit] = context.granularity.split('_').map(s => s.toLowerCase());
+      const limit = moment(context.analysisRange[1]);
+      let time = moment(context.analysisRange[0]);
+      let i = 0;
+      while ((time < limit) && (i < ROOTCAUSE_TREND_MAX_COLUMNS)) {
+        // do something
+        buckets.push(time.valueOf());
+        time = time.add(stepSize, stepUnit);
+        i++;
+      }
+
+      return buckets;
+    }
+  ),
+
+  /**
+   * Columns for trend table
+   * @type {object[]}
+   */
+  columns: computed(
+    'buckets',
+    function () {
+      const { buckets } = this.getProperties('buckets');
+
+      const columns = [
+        {
+          template: 'custom/table-checkbox',
+          className: 'metrics-table__column'
+        }, {
+          propertyName: 'label',
+          title: 'Metric',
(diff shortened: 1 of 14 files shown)


kishoreg:
Thanks Alex. That was super fast. Any reason why this is not within the Metric tab

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+
+          // TODO use O(logN) data structure (e.g. treemap)
+          const index = buckets.findIndex(b => b >= t);
+          if (index >= 0 && index < buckets.length) {
+            changes[urn][index] = change;
+          }
+        });
+      });
+
+      console.log('changes done', changes);
```
aaronucsd: console.log

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+    'changes',
+    'links',
+    'selectedUrns',
+    function () {
+      const { entities, buckets, changes, links, selectedUrns } =
+        this.getProperties('entities', 'buckets', 'changes', 'links', 'selectedUrns');
+
+      const metricUrns = filterPrefix(Object.keys(entities), 'thirdeye:metric:');
+
+      console.log('data', metricUrns);
```
aaronucsd: same here for console.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+            change: humanizeChange(change),
+            direction: toColorDirection(change, isInverse(urn, entities))
+          };
+          row[`${t}_raw`] = makeSortable(change);
+        });
+
+        return row;
+      });
+
+      console.log('data done', rows);
```
aaronucsd: here and all consoles.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+  /**
+   * Display flag for table. Hack for ember models table only re-rendering columns on init()
+   * @type {bool}
+   */
+  showTable: true,
+
+  /**
+   * Hack for ember models table only re-rendering columns on init()
+   */
+  _columnObserver: observer('columns', function () {
```
aaronucsd: Hi Alex. Not sure i understand the need for this observer. We should avoid using observer when possible due to it's synchronous nature. Can we not use an `init()` hook here?  I can't seem to find `_columnObserver` usage. 
https://guides.emberjs.com/v2.6.0/components/the-component-lifecycle/

Also why not have ` showTable: true,` be false by default vs setting to `this.set('showTable', false);`.

```
  init() {
    this._super(...arguments);
...
  },
```

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+
+    return options;
+  }),
+
+  /**
+   * Possible time buckets from analysis range
+   * @type {int[]}
+   */
+  availableBuckets: computed('context', function () {
+    const { context } = this.getProperties('context');
```
aaronucsd: Single retrieval of a prop can use `this.get()` instead.

`const context  = this.get('context');`

and all other places in this component code. Thanks.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+
+    const buckets = [];
```
aaronucsd: This should be a let if we decide to mutate the variable's value.

let buckets = [];

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+  ),
+
+  /**
+   * Columns for trend table
+   * @type {object[]}
+   */
+  columns: computed('buckets', function () {
+    const { buckets } = this.getProperties('buckets');
+
+    const columns = [
```
aaronucsd: please use `let` as it's mutating later.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+      return _.slice(availableBuckets, startIndex, startIndex + ROOTCAUSE_TREND_MAX_COLUMNS);
+    }
+  ),
+
+  /**
+   * Columns for trend table
+   * @type {object[]}
+   */
+  columns: computed('buckets', function () {
+    const { buckets } = this.getProperties('buckets');
```
aaronucsd: `const  buckets  = this.get('buckets');`

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+        const currUrn = toCurrentUrn(urn);
+        const baseUrn = toBaselineUrn(urn);
+        changes[urn] = Array(buckets.length).fill(Number.NaN);
+
+        if (!timeseries[currUrn] || !timeseries[baseUrn]) {
+          return;
+        }
+
+        // NOTE: lookup table to tolerate missing baseline values
+        const baseValueLookup = {};
```
aaronucsd: let baseValueLookup = {};

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
```diff
@@ -0,0 +1,389 @@
...
+      });
+
+    return metricUrlMapping;
+  }),
+
+  /**
+   * Keeps track of items that are selected in the table
+   * @type {Array}
+   */
+  preselectedItems: computed(
```
aaronucsd: you might be able to use `computed.filter` here.
https://www.emberjs.com/api/ember/2.15/namespaces/Ember.computed/methods/filter?anchor=filter

Something like,
preselectedItems: computed.filter( 'data', 'selectedUrns', function(urn, index, array) {
   return (urn => data[urn]).map(urn => data[urn]);
}),

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -274,44 +276,66 @@ export default Controller.extend({
...
+
+      if (context.anomalyUrns.size > 0) {
+        filterPrefix(context.anomalyUrns, 'thirdeye:metric:').forEach(urn => relatedMetricUrns.add(urn));
+      }
 
+      //
       // timeseries
-      timeseriesService.request(context, selectedUrns);
+      //
+      const timeseriesUrns = new Set(selectedUrns);
```
aaronucsd: `let` as it is not a constant.

thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -274,44 +276,66 @@ export default Controller.extend({
...
+
+      if (activeTab === ROOTCAUSE_TAB_TREND) {
+        [...relatedMetricUrns].forEach(urn => timeseriesUrns.add(toCurrentUrn(urn)));
```
aaronucsd: can we not combine the forEach here?

```
let currentUrns = new Set();
let baselineUrns = new Set();

 [...relatedMetricUrns].forEach((urn) => {
   currentUrns.add(toCurrentUrn(urn));
   baselineUrns.add(toBaselineUrn(urn));
}
timeseriesUrns.union(currentUrns);
timeseriesUrns.union(baselineUrns);
```

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
apucher: fixed

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
apucher: fixed

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
apucher: fixed

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
apucher: the table component does not re-render if we change the columns property after it has been initialized. My workaround for now is to destroy and re-create with this flag. If there is a better solution, I'd be glad to use it instead

apucher:
@kishoreg we already have tabs - adding sub views to tabs make it even more confusing. remember the metrics cards?

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
justYves: @apucher Yup that's fine, the same way we use both `set` and `setProperties`. The only difference is that `setProperties` is more performant (each `this.set()` would notify a property change, and cause `cp`s and `observer`s to recalculate), but `getProperties` and `get` are essentially doing the same thing.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
justYves: @aaronucsd I'm fine with `const` here since the reference doesn't change. Is it best practice to use a `let` to explicitly indicate that the array content might change?

kishoreg:
Good point!

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
justYves: food for thought: https://mathiasbynens.be/notes/es6-const

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
apucher: there is a better solution. :-)

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
apucher: yeah. this one pulls in data from two properties. not sure if computed.filter() was designed for that. at least I can't quite find examples

thirdeye/thirdeye-frontend/app/pods/components/rootcause-trend/component.js
aaronucsd: @justYves Yes. There are two factors for this. One is it is correct the reference default type (array) assignment isn't going to change (in this current case). Actually it's content/children is changing is the reason why this const is still valid with the linter. 

```
const foo = {};
foo.bar = 42;
console.log(foo.bar);
```

However, we can be blinded to see future reference changes if this were to be passed to a method later that does mutate the original reference. 

```
const foo = {};
someMethodChangesFoo(foo);//Uncaught TypeError: Assignment to constant variable.
console.log(foo);
```

apucher: The latter one does not mutate to original "reference" foo, but the object that is referenced by by "foo"

aaronucsd: HI @apucher.  True, as the reference of the param is to the object and not the original handler. My point is later in your code, when you decide to extract this out to a method (util/service etc), someone can accidentally modify it as such `foo = {a:42};` where it will just needs to update to `let`.  For now it's fine, just leave it as const. Thanks.

PR #2710, opened 2018-04-07
Identifier quote character support

subprotocol:
This PR adds the ability for identifiers to contain characters other than ```[A-Za-z_][A-Za-z0-9_-]*``` by way of an *identifier quote character* (ie- the backtick character).

Example-

```select avg(`attributes.age`) as `avg.age` from `person` group by `attributes.address_city` having avg(`attributes.age`)=20```

This mechansim is supported in MySQL for example: https://dev.mysql.com/doc/refman/5.7/en/identifiers.html

Further discussion in case it's helpful: https://github.com/linkedin/pinot/issues/2687

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/IdentifierAstNode.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/IdentifierAstNode.java
@@ -19,14 +19,27 @@ package com.linkedin.pinot.pql.parsers.pql2.ast;
  * AST node for identifiers (column names).
  */
 public class IdentifierAstNode extends BaseAstNode {
+  private String _expression;
   private String _name;
 
-  public IdentifierAstNode(String name) {
-    _name = name;
+  public IdentifierAstNode(String expression) {
+    _expression = expression;
+    if (
+      expression.charAt(0) == '`'
+      && expression.charAt(expression.length()-1) == '`'
+    ) {
+      _name = expression.substring(1, expression.length()-1);
+    } else {
+      _name = expression;
+    }
   }
 
   public String getName() {
     return _name;
+  }
+
+  public String getExpression() {
+    return _expression;
   }
 
   @Override
--- a/pinot-common/src/test/java/com/linkedin/pinot/pql/parsers/Pql2CompilerTest.java
+++ b/pinot-common/src/test/java/com/linkedin/pinot/pql/parsers/Pql2CompilerTest.java
@@ -15,12 +15,15 @@
  */
 package com.linkedin.pinot.pql.parsers;
 
+import com.linkedin.pinot.common.request.BrokerRequest;
 import com.linkedin.pinot.common.request.FilterOperator;
+import com.linkedin.pinot.common.request.GroupBy;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: pinot-common/src/test/java/com/linkedin/pinot/pql/parsers/Pql2CompilerTest.java [hunk 1]
```diff
@@ -186,4 +189,20 @@ public void testQueryOptions() {
...
+  @Test
+  public void testIdentifierQuoteCharacter() {
+    Pql2Compiler compiler = new Pql2Compiler();
+
+    TransformExpressionTree expTree = compiler.compileToExpressionTree("`a.b.c`");
+    Assert.assertEquals(expTree.getExpressionType(), TransformExpressionTree.ExpressionType.IDENTIFIER);
+    Assert.assertEquals(expTree.getValue(), "a.b.c");
+
+    BrokerRequest brokerRequest = compiler.compileToBrokerRequest(
+            "select avg(`attributes.age`) as `avg_age` from `person` group by `attributes.address_city` having avg(`attributes.age`)=20");
```
kishoreg: do we support aliasing?

subprotocol: Hmm, I am not certain.  I'm seeing references to aliasing in the codebase but I'm not 100% sure I follow.  Is aliasing allowing things like ```where test = :blah``` ?   And if so, are you wondering if we could also support things like ```where test = :`a.b.c` ``` ?

pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/IdentifierAstNode.java
```diff
@@ -19,16 +19,29 @@
...
+    } else {
+      _name = expression;
+    }
   }
 
   public String getName() {
     return _name;
   }
 
+  public String getExpression() {
```
jfim: Is there a particular reason why we're introducing the concept of an expression as opposed to just using the name? In which scenarios would we need the quoted identifier with backticks?

subprotocol: Great question, sorry I should have called that out.  The reason is that the expression is parsed and re-serialized back into a string.  Currently this happens here: https://github.com/linkedin/pinot/blob/a4b05037253ca5d8c1bfa7b21b767deaca6d105c/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/GroupByAstNode.java#L40-L47

If the ability is needed to go from expression tree back into string form, we would need to be able to distinguish between the name and expression of the identifier.

Without getExpression() this:
```
select count(*) from table group by `this.is.my.identifier`
```

would be parsed and toStringed with an invalid identifier:
```
select count(*) from table group by this.is.my.identifier
```

jfim: Ah good point, completely forgot about server-side reparse for expressions.

PR #2712, opened 2018-04-09
Remove failure modes from segment status checker

jfim:
Simplified the logic for the segment status checker, removing some
potential failure modes from the existing code:

- Exceptions in the segment status check loop now only cause metrics for
  that table to be unavailable, as opposed to bubbling out of the loop
- Emit Long.MIN_VALUE instead of 0 for unavailable data
- Remove flaky (and incorrect) logic that attempts to shut down and
  restart the thread based on leadership change, it is far too brittle
  for rather dubious performance gains

Minor improvements to the code:

- Emit a warning if the segment status checker is disabled, as opposed
  to silently not starting the segment status checker
- Rename segment status checker thread to "Segment status checker" from
  "SegStatChecker"
- Rename method "runSegmentMetrics" to "updateSegmentMetrics" to reflect
  its actual purpose
- Made "updateSegmentMetrics" package private as opposed to public
- Remove some pointless comments


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -174,144 +156,136 @@ public void runSegmentMetrics() {
...
+        }
+        if (nReplicasExternal < nReplicasIdealMax) {
+          LOGGER.warn("Table {} has {} replicas, below replication threshold :{}", tableName, nReplicasExternal,
+              nReplicasIdealMax);
+        }
+      } catch (Exception e) {
+        LOGGER.warn("Caught exception while updating segment status for table {}", e, tableName);
+
+        // Remove the metric for this table
+        _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.NUMBER_OF_REPLICAS, Long.MIN_VALUE);
```
mcvsubbu: Perhaps setting the value to a higher number than Long.MIN_VALUE may help? Otherwise it may be hard to eye-ball a graph showing the values and decipher whether the value was 0, 1, or 2 before the curve takes a plunge. May be -1 will be ok here. Your call.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -174,144 +156,136 @@ public void runSegmentMetrics() {
...
-    // Fetch the list of tables
     List<String> allTableNames = _pinotHelixResourceManager.getAllTables();
-    // Synchronization provided by Controller Gauge to make sure that only one thread updates the gauge
+
     for (String tableName : allTableNames) {
-      _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.NUMBER_OF_REPLICAS, 0);
-      _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.PERCENT_OF_REPLICAS, 0);
-      _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.SEGMENTS_IN_ERROR_STATE, 0);
-      _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.PERCENT_SEGMENTS_AVAILABLE, 0);
+      _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.NUMBER_OF_REPLICAS, Long.MIN_VALUE);
```
mcvsubbu: Same as above.

PR #2721, opened 2018-04-10
Refactor realtime code to eliminate use of znode for storing partition assignment

npawar:
Testing done:

I] Local testing using integration test setup.
Method: start LLCRealtimeClusterIntegrationTest for setup and then sleep. Made changes to ideal state/segment metadata and let validation manager correct them.
Tests:
1) Segment absent from ideal state 
  i) but metadata present (failed after step 2 of segment completion, check: new segment created in ideal state with CONSUMING, if prev exists make it ONLINE)
  ii) Special case of i. The missing segment was the very first segment in the ideal state for this partition
2) Segment present in ideal state
   i) metadata IN_PROGRESS, ideal state CONSUMING  (happy path)
   ii) metadata DONE, ideal state CONSUMING (failed after step 1 of segment completion, check: new metadata IN_PROGRESS, new segment CONSUMING, old segment ONLINE)
   iii) metadata IN_PROGRESS, ideal state OFFLINE (check: must generate new segment with prev segment’s start offset )
   iv) metadata DONE, ideal state OFFLINE (check: must generate new segment with prev segment’s end offset)
3) Too soon to correct case - waits till too soon, fixes after that
4) new partition

II] Testing on test cluster 
1) Check that partition assignment is unfazed for an existing table, after the new change is deployed
2) New table creation
3) Repairs:
i) Segment present in ideal state:
  a) metadata IN_PROGRESS, segment CONSUMING
  b) metadata IN_PROGRESS, segment OFFLINE
  c) metadata IN_PROGRESS, 1 replica CONSUMING 1 OFFLINE
  d) metadata DONE, segment CONSUMING
  e) metadata DONE, segment OFFLINE
  f) metadata DONE, 1 replica CONSUMING 1 OFFLINE
  g) Too soon to correct
  h) Instances changed
ii) Segment absent from ideal state
  a) Initial segment (seq 0)
  b) Mature segment

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -1586,7 +1586,7 @@ public class PinotLLCRealtimeSegmentManager {
    *   and so it is currently in back-off stage.
    * - ValidationManager starts to run, and finds the segment in improper state and manages to fix it in idealstate.
    * - The segment completion thread comes back to update the ideal state (should already find the same segment in there)
-   *
+   * TODO: split this method into multiple smaller methods
    */
   @VisibleForTesting
   protected IdealState validateLLCSegments(final TableConfig tableConfig, IdealState idealState,
@@ -1616,7 +1616,8 @@ public class PinotLLCRealtimeSegmentManager {
     try {
       partitionAssignment = _partitionAssignmentGenerator.generatePartitionAssignment(tableConfig, partitionCount);
     } catch (InvalidConfigException e) {
-      throw new RuntimeException(e.getMessage());
+      throw new IllegalStateException(
+          "Caught exception when generating partition assignment for table " + tableNameWithType, e);
     }
 
     Map<String, String> oldToNewSegmentsMap = new HashMap<>();
@@ -1747,7 +1748,8 @@ public class PinotLLCRealtimeSegmentManager {
       assignments =
           segmentAssignmentStrategy.assign(Lists.newArrayList(oldToNewSegmentsMap.values()), partitionAssignment);
     } catch (InvalidConfigException e) {
-      e.printStackTrace();
+      throw new IllegalStateException(
+          "Caught exception when assigning segments using partition assignment for table " + tableNameWithType);
     }
 
     for (Map.Entry<String, String> entry : oldToNewSegmentsMap.entrySet()) {
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -1518,6 +1518,8 @@ public class PinotLLCRealtimeSegmentManager {
    * If the controller fails after step-3, we are fine because the idealState has the new segments.
    * If the controller fails before step-1, the server will see this as an upload failure, and will re-try.
    * @param tableConfig
+   *
+   * TODO: We need to find a place to detect and update a gauge for nonConsumingPartitionsCount for a table, and reset it to 0 at the end of validateLLC
    */
   public void validateLLCSegments(final TableConfig tableConfig) {
     final String tableNameWithType = tableConfig.getTableName();
@@ -1874,7 +1876,7 @@ public class PinotLLCRealtimeSegmentManager {
 
       LLCSegmentName newLLCSegmentName = new LLCSegmentName(rawTableName, partition, nextSeqNum, now);
       createNewSegmentMetadataZNRecord(tableNameWithType, newLLCSegmentName, startOffset, partitionAssignment);
-      
+
       segmentMapper.addNewSegment(newLLCSegmentName.getSegmentName());
     }
   }
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -1613,11 +1613,17 @@ public class PinotLLCRealtimeSegmentManager {
     }
 
     PartitionAssignment partitionAssignment;
+    boolean skipNewPartitions = false;
     try {
       partitionAssignment = _partitionAssignmentGenerator.generatePartitionAssignment(tableConfig, partitionCount);
     } catch (InvalidConfigException e) {
-      throw new IllegalStateException(
-          "Caught exception when generating partition assignment for table " + tableNameWithType, e);
+      _controllerMetrics.addMeteredTableValue(tableNameWithType, ControllerMeter.PARTITION_ASSIGNMENT_GENERATION_ERROR,
+          1L);
+      LOGGER.warn(
+          "Could not generate partition assignment. Fetching partition assignment from ideal state for repair of table {}",
+          tableNameWithType);
+      partitionAssignment = _partitionAssignmentGenerator.getPartitionAssignmentFromIdealState(tableConfig, idealState);
+      skipNewPartitions = true;
     }
 
     OldToNewSegmentMapper segmentMapper = new OldToNewSegmentMapper();
@@ -1720,26 +1726,30 @@ public class PinotLLCRealtimeSegmentManager {
         if (prevMetadata != null) {
           prevSegmentId = prevMetadata.getSegmentName();
         }
+        if (skipNewPartitions && prevMetadata == null) {
+          continue;
+        }
 
         segmentMapper.addMapping(prevSegmentId, segmentId);
       }
     }
 
-    for (int partition : newPartitions) {
-      int nextSeqNum = STARTING_SEQUENCE_NUMBER;
-      LOGGER.info("Creating CONSUMING segment for {} partition {} with seq {}", tableNameWithType, partition,
-          nextSeqNum);
-      String consumerStartOffsetSpec = streamMetadata.getKafkaConsumerProperties()
-          .get(CommonConstants.Helix.DataSource.Realtime.Kafka.AUTO_OFFSET_RESET);
-      long startOffset = getKafkaPartitionOffset(streamMetadata, consumerStartOffsetSpec, partition);
-      LOGGER.info("Found kafka offset {} for table {} for partition {}", startOffset, tableNameWithType, partition);
-      String rawTableName = TableNameBuilder.extractRawTableName(tableNameWithType);
+    if (!skipNewPartitions) {
+      for (int partition : newPartitions) {
+        int nextSeqNum = STARTING_SEQUENCE_NUMBER;
+        LOGGER.info("Creating CONSUMING segment for {} partition {} with seq {}", tableNameWithType, partition,
+            nextSeqNum);
+        String consumerStartOffsetSpec = streamMetadata.getKafkaConsumerProperties().get(CommonConstants.Helix.DataSource.Realtime.Kafka.AUTO_OFFSET_RESET);
+        long startOffset = getKafkaPartitionOffset(streamMetadata, consumerStartOffsetSpec, partition);
+        LOGGER.info("Found kafka offset {} for table {} for partition {}", startOffset, tableNameWithType, partition);
+        String rawTableName = TableNameBuilder.extractRawTableName(tableNameWithType);
 
-      LLCSegmentName newLLCSegmentName = new LLCSegmentName(rawTableName, partition, nextSeqNum, now);
+        LLCSegmentName newLLCSegmentName = new LLCSegmentName(rawTableName, partition, nextSeqNum, now);
 
-      // TODO Need to create another method that we can use.
-      createNewSegmentMetadataZNRecord(tableNameWithType, newLLCSegmentName, startOffset, partitionAssignment);
-      segmentMapper.addMapping(segmentMapper.generateNewSegmentKey(partition), newLLCSegmentName.getSegmentName());
+        // TODO Need to create another method that we can use.
+        createNewSegmentMetadataZNRecord(tableNameWithType, newLLCSegmentName, startOffset, partitionAssignment);
+        segmentMapper.addMapping(segmentMapper.generateNewSegmentKey(partition), newLLCSegmentName.getSegmentName());
+      }
     }
 
     RealtimeSegmentAssignmentStrategy segmentAssignmentStrategy = new ConsumingSegmentAssignmentStrategy();
@@ -1789,7 +1799,9 @@ public class PinotLLCRealtimeSegmentManager {
     try {
       instanceAssignments = strategy.assign(Lists.newArrayList(newSegmentId), partitionAssignment);
     } catch (InvalidConfigException e) {
-      throw new RuntimeException(e.getMessage());
+      _controllerMetrics.addMeteredTableValue(idealState.getResourceName(),
+          ControllerMeter.CONTROLLER_REALTIME_TABLE_SEGMENT_ASSIGNMENT_ERROR, 1L);
+      throw new IllegalStateException("Caught exception when updating ideal state on segment completion", e);
     }
 
     List<String> newSegmentInstances = instanceAssignments.get(newSegmentId);
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -206,7 +206,7 @@ public class PinotLLCRealtimeSegmentManager {
         // Perhaps we should just let validationmanager do all the adjustments, and let the leadership transfer
         // be quick. If the leadership transfer happened due to zk connection issues, then writing to zk at this time
         // may be bad as well.
-        completeCommittingSegments();
+        //completeCommittingSegments(); // TODO: validateLLC() ?
       } else {
         // We already had leadership, nothing to do.
         LOGGER.info("Already leader. Duplicate notification");
@@ -1780,7 +1780,7 @@ public class PinotLLCRealtimeSegmentManager {
     try {
       instanceAssignments = strategy.assign(Lists.newArrayList(newSegmentId), partitionAssignment);
     } catch (InvalidConfigException e) {
-      e.printStackTrace();
+      throw new RuntimeException(e.getMessage());
     }
 
     List<String> newSegmentInstances = instanceAssignments.get(newSegmentId);
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -73,7 +73,6 @@ import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -115,6 +114,13 @@ public class PinotLLCRealtimeSegmentManager {
   private static final String METADATA_TEMP_DIR_SUFFIX = ".metadata.tmp";
   private static final String METADATA_EVENT_NOTIFIER_PREFIX = "metadata.event.notifier";
   // TODO: make this configurable with default set to 10
+  /**
+   * After step 1 of segment completion is done,
+   * this is the max time until which step 3 is allowed to complete.
+   * See {@link PinotLLCRealtimeSegmentManager#commitSegmentMetadata(String, String, long, long)} for explanation of steps 1 2 3
+   * This includes any backoffs and retries for the steps 2 and 3
+   * The segment will be eligible for repairs by the validation manager, if the time  exceeds this value
+   */
   private static int MAX_SEGMENT_COMPLETION_TIME_MINS = 10;
 
   private static PinotLLCRealtimeSegmentManager INSTANCE = null;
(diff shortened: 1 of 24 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -639,13 +681,23 @@ public boolean commitSegmentMetadata(String rawTableName, final String committin
     LLCSegmentName newLLCSegmentName = new LLCSegmentName(committingLLCSegmentName.getTableName(), partitionId, newSeqNum, now);
     String newSegmentNameStr = newLLCSegmentName.getSegmentName();
 
-    final PartitionAssignment partitionAssignment = getStreamPartitionAssignment(realtimeTableName);
+    IdealState idealState = getTableIdealState(realtimeTableName);
+    int numPartitions = _partitionAssignmentGenerator.getNumPartitionsFromIdealState(idealState);
```
mcvsubbu: you can add an assert here that the segment that is completing is in CONSUMING state.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1393,6 +1452,415 @@ public PartitionAssignment getStreamPartitionAssignment(String realtimeTableName
...
+        partitionToLatestSegments.put(partitionId, latestSegments);
+      }
+      latestSegments.offer(metadata);
+    }
+
+    return partitionToLatestSegments;
+  }
+
+  /**
+   * Validates llc segments in ideal state and repairs them if necessary
```
mcvsubbu: pls add a comment that this is also used for new table creations.
We need to check the logs and see if they are misleading in case of new table creation.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1393,6 +1452,415 @@ public PartitionAssignment getStreamPartitionAssignment(String realtimeTableName
...
+        newPartitions.add(partition);
+      }
+    }
+
+    PartitionAssignment partitionAssignment;
+    boolean skipNewPartitions = false;
+    try {
+      partitionAssignment = _partitionAssignmentGenerator.generatePartitionAssignment(tableConfig, partitionCount);
+    } catch (InvalidConfigException e) {
+      _controllerMetrics.addMeteredTableValue(tableNameWithType, ControllerMeter.PARTITION_ASSIGNMENT_GENERATION_ERROR,
```
mcvsubbu: for new tables, we should actually throw an error here (and not bump any metric).  Perhaps we should pull out the code to create new partition segments...

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1393,6 +1452,415 @@ public PartitionAssignment getStreamPartitionAssignment(String realtimeTableName
...
+    // Walk over all partitions that we have metadata for, and repair any partitions necessary.
+    // Possible things to repair:
+    // 1. The latest metadata is in DONE state, but the idealstate says segment is CONSUMING:
+    //    a. Create metadata for next segment and find hosts to assign it to.
+    //    b. update current segment in idealstate to ONLINE
+    //    c. add new segment in idealstate to CONSUMING on the hosts.
+    // 2. The latest metadata is IN_PROGRESS, but segment is not there in idealstate.
+    //    a. change prev segment to ONLINE in idealstate
+    //    b. add latest segment to CONSUMING in idealstate.
+    // 3.
```
mcvsubbu: incomplete comment. 
3. All instances of a segment are in OFFLINE state. Create a new segment (with the next seq number) and restart consumption from the same offset (if possible) or a newer offset (if realtime stream does not have the same offset). In latter case, report data loss.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1393,6 +1452,415 @@ public PartitionAssignment getStreamPartitionAssignment(String realtimeTableName
...
+              getBetterStartOffsetIfNeeded(tableNameWithType, partition, segmentName, startOffset, nextSeqNum);
+          LLCSegmentName newLLCSegmentName =
+              new LLCSegmentName(segmentName.getTableName(), partition, nextSeqNum, now);
+
+          createNewSegmentMetadataZNRecord(tableNameWithType, newLLCSegmentName, startOffset, partitionAssignment);
+          segmentMapper.addMapping(segmentMapper.generateNewSegmentKey(partition), newLLCSegmentName.getSegmentName());
+        }
+        // else It can be in ONLINE state, in which case there is no need to repair the segment.
+      } else {
+        if (isTooSoonToCorrect(tableNameWithType, segmentId, now)) {
```
mcvsubbu: can u move these lines below the comment please? thanks

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1393,6 +1452,415 @@ public PartitionAssignment getStreamPartitionAssignment(String realtimeTableName
...
+    for (String instance : newSegmentInstances) {
+      idealState.setPartitionState(newSegmentId, instance,
+          PinotHelixSegmentOnlineOfflineStateModelGenerator.CONSUMING_STATE);
+    }
+
+    return idealState;
+  }
+
+  /**
+   * Class to maintain the mapping from old segments to new segments
```
mcvsubbu: can u add more comments here?the class really keeps a mapping of  segment names that need to move from CONSUMING to ONLINE,   to new segment names (to be created in CONSUMING state). In some cases, we will not have old segment names (e.g. when a new partition is detected, or the old segment is OFFLINE).

Maybe we can keep two different sets instead? One that keeps the old segment that need to move to ONLINE and another that keeps new segments that need to be created in CONSUMING mode. Then the logic is more readable.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
```diff
@@ -942,6 +942,7 @@ private int numReplicasToLookFor() {
     private SegmentCompletionProtocol.Response processStoppedConsuming(String instanceId, long offset, String reason, boolean createNew) {
       LOGGER.info("Instance {} stopped consuming segment {} at offset {}, state {}, createNew: {}, reason:{}",
           instanceId, _segmentName, offset, _state, createNew, reason);
+      // TODO: what happens to this method? below method is deprecated
```
mcvsubbu: this method should not be deprecated :-)

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1212,6 +1268,7 @@ private ZNRecord makeZnRecordForNewSegment(String realtimeTableName, int numRepl
    * run, will auto-create a new segment with the appropriate offset.
    * See {@link #createConsumingSegment(String, Set, List, TableConfig)}
    */
+  @Deprecated
```
mcvsubbu: cant deprecate this method

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1393,6 +1452,415 @@ public PartitionAssignment getStreamPartitionAssignment(String realtimeTableName
...
+        Set<String> oldInstances = idealState.getInstanceSet(oldSeg);
+        Preconditions.checkArgument(CollectionUtils.isNotEmpty(oldInstances));
+        for (String instance : oldInstances) {
+          idealState.setPartitionState(oldSeg, instance,
+              PinotHelixSegmentOnlineOfflineStateModelGenerator.ONLINE_STATE);
+        }
+      }
+    }
+    return idealState;
+  }
```
mcvsubbu: we are missing the logic to update the nonconsumingPartitionCouintMetric (in validationmanager), but i think that metric was not updated correctly anyway. We need to create a new controller guage that reflects this. We can do that independently.

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1543 @@
...
+    segmentManager.createNewSegmentMetadataZNRecord(tableName, newLlcSegmentName, nextOffset, partitionAssignment);
+    segmentManager.updateIdealStateOnSegmentCompletion(idealState, segmentName, newLlcSegmentName.getSegmentName(),
+        partitionAssignment);
+  }
+
+  private void validateLLCPartitionsIncrease(FakePinotLLCRealtimeSegmentManager segmentManager, IdealState idealState,
+      TableConfig tableConfig, int nPartitions, int nReplicas, List<String> instances, boolean exceptionExpected,
+      boolean skipNewPartitions) {
+    try {
+      Map<String, Map<String, String>> oldMapFields = new HashMap<>(idealState.getRecord().getMapFields());
```
mcvsubbu: is this a deep copy? you can create a copy of the idealstate. i think there is a method to do that.

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1543 @@
...
+      Map<String, LLCRealtimeSegmentZKMetadata> oldMetadataMap = new HashMap<>(segmentManager._metadataMap);
```
mcvsubbu: we want a deep copy right? you may have to iterate and copy each metadata

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1543 @@
...
+    boolean exceptionExpected;
+    boolean skipNewPartitions = false;
+    List<String> instances = getInstanceList(5);
+
+    // empty to 4 partitions
+    tableConfig = makeTableConfig(tableName, nReplicas, KAFKA_OFFSET, DUMMY_HOST, DEFAULT_SERVER_TENANT,
+        DEFAULT_STREAM_ASSIGNMENT_STRATEGY);
+    idealState = idealStateBuilder.build();
+    nPartitions = 4;
+    exceptionExpected = false;
```
mcvsubbu: final

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1543 @@
...
+    segmentManager._partitionAssignmentGenerator.setConsumingInstances(instances);
+    PartitionAssignment expectedPartitionAssignment =
+        segmentManager._partitionAssignmentGenerator.generatePartitionAssignment(tableConfig, nPartitions);
+
+    // set up a happy path - segment is present in ideal state, is CONSUMING, metadata says IN_PROGRESS
+    idealState = clearAndSetupHappyPathIdealState(idealStateBuilder, segmentManager, tableConfig, nPartitions);
+
+    // randomly introduce an error condition and assert that we repair it
+    long seed = new Random().nextLong();
+    System.out.println("Random seed for validate llc repairs " + seed);
```
mcvsubbu: Move the seed to the top at the test level and add a comment why we are printing to system.out

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1543 @@
...
+
+          int randomlySelectedPartition = random.nextInt(nPartitions);
+          Map<String, LLCSegmentName> partitionToLatestSegments =
+              segmentManager._partitionAssignmentGenerator.getPartitionToLatestSegments(idealState);
+          LLCSegmentName llcSegmentName = partitionToLatestSegments.get(String.valueOf(randomlySelectedPartition));
+          idealStateBuilder.removeSegment(llcSegmentName.getSegmentName());
+
+          Assert.assertNull(idealState.getRecord().getMapFields().get(llcSegmentName.getSegmentName()));
+          nPartitions = expectedPartitionAssignment.getNumPartitions();
+          Map<String, Map<String, String>> oldMapFields = idealState.getRecord().getMapFields();
```
mcvsubbu: deep copy

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1543 @@
...
+
+    for (int run = 0; run < 200; run++) {
+
+      boolean consuming = true;
+      boolean inProgress = true;
+      if (random.nextBoolean()) {
+        segmentManager.tooSoonToCorrect = true;
+      }
+      if (!random.nextBoolean()) { // not present in ideal state
+        // 1. very first metadata IN_PROGRESS, but segment missing - repair: create segment CONSUMING
```
mcvsubbu: the comments are not clear. Would it help if you create a few final boolean variables for each case?
1. allSegmentsOffline
2. step3NotDone
3. tooSoon
4. step2NotDone
5. instancesChanged
6. nPartitionsIncreased

and then run the individual tests (perhaps in different methods if necessary)

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
npawar: Moved logic for new table creations outside this method. This addresses all concerns about new table logic and tests related to it

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
npawar: Added a TODO for this

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1543 @@
...
+  private IdealState validateLLCNewTable(TableConfig tableConfig, IdealState idealState, int nPartitions, int nReplicas,
+      List<String> instances, boolean exceptionExpected, boolean skipNewPartitions) {
+    IdealState updatedIdealState = null;
+    FakePinotLLCRealtimeSegmentManager segmentManager = new FakePinotLLCRealtimeSegmentManager(null);
+    segmentManager._partitionAssignmentGenerator.setConsumingInstances(instances);
+    try {
+      updatedIdealState = segmentManager.validateLLCSegments(tableConfig, idealState, nPartitions);
+      Assert.assertFalse(exceptionExpected);
+      Map<String, LLCSegmentName> partitionToLatestSegments =
+          segmentManager._partitionAssignmentGenerator.getPartitionToLatestSegments(updatedIdealState);
```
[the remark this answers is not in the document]
npawar: As of now the partitionAssignmentGenerator is a field inside the PinotLLCRealtimeSegmentManager, and hence accessed in this way. When we move to having a factory for partition assignment generators, this will be refactored accordingly

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1543 @@
...
+
+  private void validateLLCPartitionsIncrease(FakePinotLLCRealtimeSegmentManager segmentManager, IdealState idealState,
+      TableConfig tableConfig, int nPartitions, int nReplicas, List<String> instances, boolean exceptionExpected,
+      boolean skipNewPartitions) {
+    try {
+      Map<String, Map<String, String>> oldMapFields = new HashMap<>(idealState.getRecord().getMapFields());
+      Map<String, LLCRealtimeSegmentZKMetadata> oldMetadataMap = new HashMap<>(segmentManager._metadataMap);
+      segmentManager._partitionAssignmentGenerator.setConsumingInstances(instances);
+      IdealState updatedIdealState = segmentManager.validateLLCSegments(tableConfig, idealState, nPartitions);
+      Assert.assertFalse(exceptionExpected);
```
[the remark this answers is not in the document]
npawar: This is to make sure that we indeed catch an exception when we expected it, and did not just slide through the tests

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1617 @@
...
+            // get old state
+            nPartitions = expectedPartitionAssignment.getNumPartitions();
+            IdealState idealStateCopy = new IdealState(
+                (ZNRecord) znRecordSerializer.deserialize(znRecordSerializer.serialize(idealState.getRecord())));
+            Map<String, Map<String, String>> oldMapFields = idealStateCopy.getRecord().getMapFields();
+
+            if (tooSoonToCorrect) {
+              segmentManager.validateLLCSegments(tableConfig, idealState, nPartitions);
+              // validate nothing changed and try again with disabled
+              Assert.assertEquals(oldMapFields, idealState.getRecord().getMapFields());
```
mcvsubbu: I think this assert statement is not doing the comparison as we expect. It just asserts that oldMapFields is a subset of idealState.getRecord().getMapFields()

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1617 @@
...
+    segmentManager._partitionAssignmentGenerator.setConsumingInstances(instances);
+    PartitionAssignment expectedPartitionAssignment =
+        segmentManager._partitionAssignmentGenerator.generatePartitionAssignment(tableConfig, nPartitions);
+
+    // set up a happy path - segment is present in ideal state, is CONSUMING, metadata says IN_PROGRESS
+    idealState = clearAndSetupHappyPathIdealState(idealStateBuilder, segmentManager, tableConfig, nPartitions);
+
+    final int nInstancesChanged = 0;
+    final int nPartitionsIncreased = 1;
+    final int nInstancesChangedAndNPartitionsIncreased = 2;
```
mcvsubbu: can we make this an enum for readability?

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1617 @@
...
+
+            verifyRepairs(tableConfig, idealState, expectedPartitionAssignment, segmentManager, oldMapFields);
+          }
+        }
+      }
+
+      // randomly change instances used, or increase partition count
+      int randomExternalChanges = random.nextInt(20);
+
+      if (randomExternalChanges == nInstancesChanged) {
```
mcvsubbu: so, if the random is not 1, 2, or 3,  you want none of these changes applied, right? i.e. we want that 80% of the time there should be no changes, otherwise have one of these 3 changes. Can you code it up that way? and have enums as well like requested above? thanks,

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1617 @@
...
+          nPartitions = expectedPartitionAssignment.getNumPartitions();
+          IdealState idealStateCopy = new IdealState(
+              (ZNRecord) znRecordSerializer.deserialize(znRecordSerializer.serialize(idealState.getRecord())));
+          Map<String, Map<String, String>> oldMapFields = idealStateCopy.getRecord().getMapFields();
+
+          if (tooSoonToCorrect) {
+            segmentManager.validateLLCSegments(tableConfig, idealState, nPartitions);
+            // validate nothing changed and try again with disabled
+            Assert.assertEquals(oldMapFields, idealState.getRecord().getMapFields());
+            segmentManager.tooSoonToCorrect = false;
```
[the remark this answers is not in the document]
npawar: I set it back to false immediately, because I wanted it to try again with not tooSoonToCorrect, before going back up and introducing more error conditions. I did that because I wanted to avoid having to keep track of which partitions already have the error condition

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
```diff
@@ -0,0 +1,1617 @@
...
+          LLCSegmentName latestSegment = partitionToLatestSegments.get(String.valueOf(randomlySelectedPartition));
+          idealState = idealStateBuilder.setSegmentState(latestSegment.getSegmentName(), "OFFLINE").build();
+
+          // get old state
+          nPartitions = expectedPartitionAssignment.getNumPartitions();
+          IdealState idealStateCopy = new IdealState(
+              (ZNRecord) znRecordSerializer.deserialize(znRecordSerializer.serialize(idealState.getRecord())));
+          Map<String, Map<String, String>> oldMapFields = idealStateCopy.getRecord().getMapFields();
+
+          if (tooSoonToCorrect) {
```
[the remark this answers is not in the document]
npawar: Good catch! Fixed it. Is was undetected due to the ineffective map comparison

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTestNew.java
npawar: I had consciously kept it as final integers and not enum, because I wanted to keep it local. 
Refactored it to use enums, and restructured the random calculation to pick random 20%, instead of numbered 1-20

PR #2734, opened 2018-04-18
Make time column name available from table/user config in places where primary time column is needed

npawar:
Make time column name available from table/user config in places where primary time column is needed, in order to be able to support migration of single time column to multiple datetime columns


pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -272,6 +280,19 @@ public boolean execute()
       }
     }
 
+    String configDateTimeColumnName = segmentGeneratorConfig.getTimeColumnName();
+    if (_timeColumnName == null) {
```
kishoreg: should we make it mandatory? Until now we haven't mandated having time column in the data.

npawar: Fixed. Time column is not a mandatory input anymore. Added some TODOs to check that if this is not provided, use DATETIME field, and if multiple of them are found, throw an exception.

PR #2735, opened 2018-04-18
[TE] rootcause - workaround for function filters not stored in anomaly

apucher:
Anomaly functions store certain filter values only conditionally in anomalies. This Pr includes a workaround to re-insert the omitted values when necessary in the RCA UI


cyy0714:
There is a static method in AnomalyResource.generateFilterSetForTimeSeriesQuery(), which doing the similar thing. Would it be possible to merge the two functionality? It would be better to keep one version in the codebase.

Issue #2729, opened 2018-04-16
SIMPLE_DATE_FORMAT is not honoring output granularities


subprotocol:
SIMPLE_DATE_FORMAT does not take into account output granularities.  This causes group by queries to incorrectly bucket timeseries data.

For example-

inputFmt: `1:MILLISECONDS:EPOCH`
outputFmt: `1:MILLISECONDS:SIMPLE_DATE_FORMAT:yyyy-MM-dd HH:mm:ss.SSS`
outputGranularity: `15:MINUTES`

given a time of `1523560598233L`  currently returns `2018-04-12 19:16:38.233` instead of the expected output: `2018-04-12 19:15:00.000`.

The fix appears to be very simple.  BaseDateTimeTransformer.transformMillisToSDF just needs call transformToOutputGranularity before being passed to the outputDateTimeFormatter.

Unless anyone has an objection I would be happy to put together a PR to fix this.

Jackie-Jiang:
No bucketing in case of simple date format was intentional. Also, there are notes in BaseDataTimeTransformer.java line 30: "NOTE: time size, time unit and output granularity does not apply to simple date format."
Please refer to the comments in #2659

subprotocol:
@Jackie-Jiang thank you for pointing me to that. The specific use-case I am after is allowing customers to query their data at different granularities and timezones as determined at query time.  For example Druid has support for querying in this mannor: http://druid.io/docs/latest/querying/granularities.html

@kishoreg sure thing thanks.

Jackie-Jiang:
@kishoreg My initial PR has bucketing for SDF, but @npawar pointed out that we should not bucket for SDF, so I don't know if this would affect the current use case. @npawar Can you please verify that?

subprotocol:
@Jackie-Jiang @kishoreg Here is my take on this- https://github.com/linkedin/pinot/pull/2738

npawar:
Bucketing was left out, as it doesn't work correctly when timezone comes into the picture. 
inputFmt: 1:MILLISECONDS:EPOCH
outputFmt: 1:MILLISECONDS:SIMPLE_DATE_FORMAT:yyyy-MM-dd HH:mm:ss.SSS tz(America/Los_Angeles)
outputGranularity: 1:DAYS

In this case, convert inputFmt to millis and then millis to outputFmt works
But convert inputFmt to millis, bucket to days, convert millis to outputFmt will not work for some hours of the day
[...]

PR #2738, opened 2018-04-18
Add output granularity support to BaseDateTimeTransformer

subprotocol:
Allows date-time bucketing to be arbitrarily determined at query time.

For background discussion see: https://github.com/linkedin/pinot/issues/2729


pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/transformer/datetime/BaseDateTimeTransformer.java
```diff
@@ -62,7 +62,7 @@ protected long transformMillisToEpoch(long millisSinceEpoch) {
   }
 
   protected String transformMillisToSDF(long millisSinceEpoch) {
-    return _outputDateTimeFormatter.print(millisSinceEpoch);
+    return _outputDateTimeFormatter.print(transformToOutputGranularity(millisSinceEpoch));
```
Jackie-Jiang: Should not embed **transformToOutputGranularity** into **transformMillisToSDF**, but add it into **Epoch/SDFToSDFTransformer** (refer to **SDFToEpochTransformer**)

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/transformer/datetime/BaseDateTimeTransformer.java
```diff
@@ -27,7 +27,7 @@
 /**
  * Base date time transformer to transform and bucket date time values from epoch/simple date format to epoch/simple
  * date format.
- * <p>NOTE: time size, time unit and output granularity does not apply to simple date format.
+ * <p>NOTE: time size and time unit do not apply to simple date format.
```
Jackie-Jiang: Thanks for updating the javadoc

pinot-core/src/main/java/com/linkedin/pinot/core/operator/transform/transformer/datetime/BaseDateTimeTransformer.java
subprotocol: Ahh I see now thank you, I will get that changed.

kishoreg:
@npawar raised a good point about handling timezones in #2729. Lets make sure we handle that.

subprotocol:
any chance this could get a quick review?  I have some cycles coming up that if needed I can make additional improvements if required.

npawar:
LGTM as well. 
There were a lot of test cases at one point which we lost sometime back (https://github.com/linkedin/pinot/pull/2659/files#diff-7b9f15468a245a32ab97a2f6bd1f0af9)
Jackie is adding them back. If you could hold on until those are in, and verify that your change passes them all, that would be great

subprotocol:
Sure no problem, that sounds like a good a good plan. Thanks!

npawar:
Tests have been added back @subprotocol

subprotocol:
Thank you!  I will take a look at this today

subprotocol:
@npawar ok all merged up.  I updated the tests I had written to follow the flow of the tests that were put back into DateTimeConverterTest.

npawar:
Since your change is in millisToSDF, please also add some tests for SDF to SDF conversion, since SDFToSDFTransform is also affected by the change

subprotocol:
It's a slight tangent from the current path, but one thing I wanted to raise- should timeunits even be allowed when using SIMPLE_DATE_FORMAT?

Currently a time unit is required on the inputFmt and outputFmt. ie-
```1:MILLISECONDS:SIMPLE_DATE_FORMAT:yyyy-MM-dd HH:mm:ss.SSS tz(America/Los_Angeles)```

Should we update the DateTimeFormatSpec to not require the timeunit when using SIMPLE_DATE_FORMAT?

subprotocol:
:thumbsup:

subprotocol:
thanks!

PR #2742, opened 2018-04-19
 [TE] frontend - aaronucsd/Add ember data to TE on Dashboard (Home)

aaronucsd:
screen shot 2018-04-17 at 2 07 14 pm
1. Added the models, adapters, and serilizers for ember data for the anomaly  apis.
2. Added new anomaly service to replace the need of `utils/api/anomaly.js` and `utils/anomaly.js` as the adapters will do those internally.
   This anomaly servie has a built in humanized caching to cache the ember objects meant for display purposes only and not mutate the actual proxy models (records).
3. Added a new query cache service. This will add a new caching layer serivce to save the `union wrapper` data that ember-data do not cache in the store.
Ember data will only cache the entities (resources) only. This service can be used for other client persistant libs ($ajax, redux etc).
4. Updated Dashboard to use the new `api/anomalies/services.js` methods.
5. Made fix for the styles for the metric tables.
6. Added the pagination for the metric tables vs `Show more` link as table will not support that.
7. Fixed the time picker codes for dashboard to work now with the cachhing as well. Reload of the same url query will pull from the cache now and draw instantly.
8. TODO: tests for the new services
9. TODO: Write guideline to use ember data conversion for ThirdEye.
10. TODO: Update the anomaly's resolution option to also use ember data.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/adapters/base.js
+++ b/thirdeye/thirdeye-frontend/app/adapters/base.js
@@ -1,0 +1,51 @@
+import DS from 'ember-data';
+
+/*
+ * @summary This base supports the standard advanced configurations needed for making the requests.
+ */
+export default DS.RESTAdapter.extend({
+  headers: {
+    'Accept': '*/*'
+  },
+  ajaxOptions() {
+    let ajaxOptions = this._super(...arguments);
+    // when server returns an empty response, we'll make it valid by replacing with {}
+    ajaxOptions.converters = {
+      'text json': function(data) {
+        return data === '' ? {} : JSON.parse(data);
+      }
+    };
+
+    return ajaxOptions;
+  },
+
+  handleResponse(status, headers, payload, requestData) {
+    // For creates we look into the header
+    if (status === 201 || status === 204) {
+      headers['content-type'] = 'application/json';
+      const id = headers['x-restli-id'] || headers['X-RestLi-Id'];
+      if (id) {
+        payload = { id };
+      }
+    }
+    let response = this._super(status, headers, payload, requestData);
+    return response;
+  },
+
+  // TODO: Will keep for reference - lohuynh
+  // query: function query(store, type, query) {
+  //   const url = `${this.get('namespace')}/${query.appName}`;
+  //   delete query.appName;//remove from query as not needed for actual request
+  //   return this.ajax(url, 'GET', { data: query });
+  // },
+  // The urlForQuery works like the `query` hook above. Both allow mutating the request url.
+  urlForQuery (query, modelName) {
+    /* The switch allows for adding more model names that needs custom request url */
+    switch(modelName) {
+      case 'performance':
+        return `${this.get('namespace')}/${query.appName}`;
+      default:
+        return this._super(...arguments);
+    }
+  }
+});
--- a/thirdeye/thirdeye-frontend/app/pods/components/stats-cards/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/stats-cards/component.js
@@ -57,7 +57,7 @@ export default Component.extend({
 
       card.forEach((stat, index) => {
         const property = props[index];
-        obj[property] = (property === 'value' && Number.isNaN(stat)) ? 0 : stat;//let's return a number vs NaN
+        obj[property] = stat;
       });
       cards.push(obj);
     });
--- a/thirdeye/thirdeye-frontend/app/pods/custom/anomalies-table/current-wow/template.hbs
+++ b/thirdeye/thirdeye-frontend/app/pods/custom/anomalies-table/current-wow/template.hbs
@@ -1,4 +1,4 @@
 <p>{{record.current}}/{{record.baseline}}</p>
 <p class="te-anomaly-table__value-label te-anomaly-table__value-label--{{calculate-direction record.change}}">
-  ({{record.humanizedChange}})
+  ({{record.humanizedChangeDisplay}})
 </p>
--- a/thirdeye/thirdeye-frontend/app/pods/home/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/home/route.js
@@ -1,20 +1,14 @@
 import Route from '@ember/routing/route';
-import { humanizeFloat, humanizeChange, checkStatus } from 'thirdeye-frontend/utils/utils';
-import floatToPercent from 'thirdeye-frontend/utils/float-to-percent';
 import columns from 'thirdeye-frontend/shared/anomaliesTableColumns';
 import fetch from 'fetch';
 import { hash } from 'rsvp';
-import {
-  getFormatedDuration,
-  getAnomaliesByAppName,
-  anomalyResponseObj,
-  getPerformanceByAppNameUrl
-} from 'thirdeye-frontend/utils/anomaly';
 import { selfServeApiCommon } from 'thirdeye-frontend/utils/api/self-serve';
 import { setProperties } from '@ember/object';
 import { task } from 'ember-concurrency';
 import RSVP from "rsvp";
 import moment from 'moment';
+import { inject as service } from '@ember/service';
+import _ from 'lodash';
 
 const queryParamsConfig = {
   refreshModel: true,
@@ -22,6 +16,9 @@ const queryParamsConfig = {
 };
 
 export default Route.extend({
+  store: service('store'),
+  anomaliesApiService: service('services/api/anomalies'),
+
   queryParams: {
     appName: queryParamsConfig,
     startDate: queryParamsConfig,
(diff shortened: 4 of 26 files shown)


thirdeye/thirdeye-frontend/app/serializers/performance.js
```diff
@@ -0,0 +1,14 @@
+import DS from 'ember-data';
+
+export default DS.JSONSerializer.extend({
+  /*
+   * @summary normalizing the payload from api response simpler or legacy json that may not support the http://jsonapi.org/ spec
+   */
+  normalizeQueryResponse(store, primaryModelClass, payload) {
```
justYves: Can we also document the params in the jsdoc?

justYves: Can we also document the params in the jsdoc?

thirdeye/thirdeye-frontend/app/serializers/base.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    return {
+      data
+    };
+  }
+  /*
+   * serializing the data to send to the api server
+   */
+  // serialize(snapshot, options) {
```
justYves: do we still need this?

thirdeye/thirdeye-frontend/app/pods/services/query-cache/service.js
```diff
@@ -0,0 +1,49 @@
+import Service from '@ember/service';
+import { inject as service } from '@ember/service';
```
justYves: Do we need to import both Service and inject ?

thirdeye/thirdeye-frontend/app/pods/services/query-cache/service.js
```diff
@@ -0,0 +1,49 @@
...
+       cached = this._cache[adapterOptions.cacheKey] = this.get('store').query(type, query, adapterOptions);//add to our dictionary
+    }
+
+    return cached;
+  },
+  /*
+  We can use the `url` as the `cacheKey` in the `query` hook.
+  which would be different with each change in query params.
+  */
+  urlForQueryKey(modelName, query) {
```
justYves: let's document the params in the doc above, also, let's stick to the JS doc comment syntax: 
```
/**
 * This is a JS doc comment
 * @type/@method
 * @param
 * @example
 * @return
 */
```

thirdeye/thirdeye-frontend/app/pods/services/api/anomalies/service.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @type {Ember.Service}
+ * @summary This service provides all the api calls for anomalies related data. An `Ember.Service`
+   is a long-lived Ember object that can be made available in different parts of your application.
+ * @example anomaliesApiService: service('services/api/anomalies');
+ */
+export default Service.extend({
+  store: service('store'),
+  queryCache: service('services/query-cache'),
+  init() {
```
justYves: Do we need this `init()` if we're only calling `this_super()`?

thirdeye/thirdeye-frontend/app/pods/components/stats-cards/component.js
```diff
@@ -12,6 +12,7 @@
  */
 import Component from '@ember/component';
 import { set, get } from '@ember/object';
+import { humanizeChange } from 'thirdeye-frontend/utils/utils';
```
justYves: Is this import being used here?

thirdeye/thirdeye-frontend/app/pods/custom/anomalies-table/current-wow/template.hbs
```diff
@@ -1,4 +1,4 @@
 <p>{{record.current}}/{{record.baseline}}</p>
 <p class="te-anomaly-table__value-label te-anomaly-table__value-label--{{calculate-direction record.change}}">
-  ({{record.humanizedChange}})
+  ({{record.humanizedChangeDisplay}})
```
justYves: nit: Let's make this a template helper if this is a UI only display property

thirdeye/thirdeye-frontend/app/serializers/performance.js
aaronucsd: Yes! We should. Will do that. Thanks!

thirdeye/thirdeye-frontend/app/pods/services/query-cache/service.js
aaronucsd: Ah my oversight. Good catch!

thirdeye/thirdeye-frontend/app/pods/components/stats-cards/component.js
aaronucsd: looks like not. Good catch.

thirdeye/thirdeye-frontend/app/pods/services/api/anomalies/service.js
aaronucsd: Nope. You're right. I had something there before but removed it. Let me remove this init now. Thanks!

thirdeye/thirdeye-frontend/app/pods/components/stats-cards/component.js
aaronucsd: removed

thirdeye/thirdeye-frontend/app/pods/services/query-cache/service.js
aaronucsd: Actually yes. One is for ` Service.extend(` and the other is to inject `store: service('store'),`.

thirdeye/thirdeye-frontend/app/serializers/performance.js
aaronucsd: Actually this is a built in hook. Like `model`, `preModel` hooks in route. But i can explicitly document it as well.

aaronucsd: /**
    * @summary A serializer hook. Normalizing the payload from api response simpler or legacy json that may not support the http://jsonapi.org/ spec
    * @method normalizeQueryResponse
    * @param {DS.Store} store
    * @param {DS.Model} primaryModelClass
    * @param {Object} payload - the data object
    * @return {Obect}
    */

aaronucsd:
Updates with @justYves feedbacks. Thanks!

thirdeye/thirdeye-frontend/app/pods/home/route.js
```diff
@@ -48,38 +51,50 @@ export default Route.extend({
...
+      duration,
+      applications
     });
   },
 
   afterModel(model) {
-    const appName = model.appName || model.applications[0].application;
-    const startDate = Number(model.startDate) || this.get('startDate');
+    // Overrides with params if exists
+    const appName = model.appName || null;//model.applications.get('firstObject.application');
```
newsummit: Is this comment a TODO?

thirdeye/thirdeye-frontend/app/adapters/base.js
```diff
@@ -0,0 +1,51 @@
...
+  // query: function query(store, type, query) {
+  //   const url = `${this.get('namespace')}/${query.appName}`;
+  //   delete query.appName;//remove from query as not needed for actual request
+  //   return this.ajax(url, 'GET', { data: query });
+  // },
+  // The urlForQuery works like the `query` hook above. Both allow mutating the request url.
+  urlForQuery (query, modelName) {
+    /* The switch allows for adding more model names that needs custom request url */
+    switch(modelName) {
+      case 'performance':
```
newsummit: Thanks for setting this up: store/models/adapters. Much simpler than what we had before.

thirdeye/thirdeye-frontend/app/pods/services/api/anomalies/service.js
```diff
@@ -0,0 +1,78 @@
...
+  async queryAnomaliesByAppName(appName, start) {
+    assert('you must pass appName param as an required argument.', appName);
+    assert('you must pass start param as an required argument.', start);
+
+    const queryCache = this.get('queryCache');
+    const modelName = 'anomalies';
+    const query = { application: appName, start };
+    const anomalies = await queryCache.query(modelName, query, { reload: false, cacheKey: queryCache.urlForQueryKey(modelName, query) });
+    return anomalies;
+    //const url = anomalyApiUrls.getAnomaliesByAppNameUrl(appName, startTime);//TODO: remove from the utils/api/anomaly.js - lohuynh
```
newsummit: I see. So this new architecture renders our API utils obsolete. One of the reasons for having those was to have a centralized inventory of all APIs used. What do you recommend for that?

aaronucsd:
Updated with humanized caching to avoid mutating the actual ember proxy models (records from store).

PR #2744, opened 2018-04-19
[TE] dataframe - lazy timestamps

apucher:
The existing dataframe util eagerly generates timestamp in dataset-granularity resolution. This is not scalable to second or millisecond resolution data. This PR eschews eager timestamp generation for a lazy approach.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dataframe/util/DataFrameUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dataframe/util/DataFrameUtils.java
@@ -136,9 +136,7 @@ public class DataFrameUtils {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dataframe/util/DataFrameUtils.java
```diff
@@ -147,17 +147,12 @@ public static DataFrame evaluateExpressions(DataFrame df, Collection<MetricExpre
    * @return dataframe with modified timestamps
```
jihaozh: Fix comment? Seems no longer does filling NULL values for missing rows.

apucher: good catch. fixed

PR #2747, opened 2018-04-19
[TE] Add rule-based detection in addition to algorithm-based detection

jihaozh:
This PR allows rule-based anomaly detection and algorithm-based detection to run for the same anomaly function at the same time and merge the results. Also, enable the potential to add multiple secondary detection functions.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
@@ -65,7 +65,7 @@ public class DetectionTaskRunner implements TaskRunner {
   private List<String> collectionDimensions;
   private AnomalyFunctionFactory anomalyFunctionFactory;
   private BaseAnomalyFunction anomalyFunction;
-  private BaseAnomalyFunction ruleBasedAnomalyFunction;
+  private List<BaseAnomalyFunction> secondaryAnomalyFunctions;
 
   public List<TaskResult> execute(TaskInfo taskInfo, TaskContext taskContext) throws Exception {
     detectionTaskCounter.inc();
@@ -90,7 +90,7 @@ public class DetectionTaskRunner implements TaskRunner {
     jobExecutionId = detectionTaskInfo.getJobExecutionId();
     anomalyFunctionFactory = taskContext.getAnomalyFunctionFactory();
     anomalyFunction = anomalyFunctionFactory.fromSpec(anomalyFunctionSpec);
-    ruleBasedAnomalyFunction = anomalyFunctionFactory.getRuleBasedAnomalyFunction(anomalyFunctionSpec);
+    secondaryAnomalyFunctions = anomalyFunctionFactory.getSecondaryAnomalyFunctions(anomalyFunctionSpec);
     detectionJobType = detectionTaskInfo.getDetectionJobType();
 
     String dataset = anomalyFunctionSpec.getCollection();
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
```diff
@@ -264,6 +266,11 @@ private void storeData(AnomalyDetectionOutputContext anomalyDetectionOutputConte
       resultsOfAnEntry =
           anomalyFunction.analyze(dimensionMap, metricTimeSeries, windowStart, windowEnd, historyMergedAnomalies);
     }
+    if (ruleBasedAnomalyFunction != null) {
+      List<AnomalyResult> rulebasedAnomalies= ruleBasedAnomalyFunction.analyze(dimensionMap, metricTimeSeries, windowStart, windowEnd,
```
apucher: space

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
```diff
@@ -65,6 +65,7 @@
   private List<String> collectionDimensions;
   private AnomalyFunctionFactory anomalyFunctionFactory;
   private BaseAnomalyFunction anomalyFunction;
+  private BaseAnomalyFunction ruleBasedAnomalyFunction;
```
apucher: I wonder if it's better to call this "secondary" instead of "ruleBased". After all, any anomaly function should work (minus collisions in the properties)

kishoreg: why not a list of anomalyFunctions?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/pojo/AnomalyFunctionBean.java
```diff
@@ -30,6 +30,7 @@
   private MetricAggFunction metricFunction;
 
   private String type;
+  private String ruleBasedAnomalyFunctionType;
```
kishoreg: what if we have multiple anomaly functions?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/detection/DetectionTaskRunner.java
```diff
@@ -89,6 +90,7 @@ private void setupTask(TaskInfo taskInfo, TaskContext taskContext) throws Except
     jobExecutionId = detectionTaskInfo.getJobExecutionId();
     anomalyFunctionFactory = taskContext.getAnomalyFunctionFactory();
     anomalyFunction = anomalyFunctionFactory.fromSpec(anomalyFunctionSpec);
+    ruleBasedAnomalyFunction = anomalyFunctionFactory.getRuleBasedAnomalyFunction(anomalyFunctionSpec);
```
kishoreg: instead of this, we can have anomalyFunctionFactory return a list of anomalyFunctions and we call the analyze method for each anomalyFunction and the only code change required would be in anomalyFunctionFactory

jihaozh: Sure. Will change to a list of anomaly functions.

PR #2751, opened 2018-04-23
Fix status code for rebuildBrokerResourceFromHelixTags

jackjlli:
This PR fixes the status code for rebuildBrokerResourceFromHelixTags method.
Previously no matter what the response was, the response code was always 200.
Now it strictly follows the instruction shown on the method above(200/400/404/500).

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableTenantConfigs.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableTenantConfigs.java
@@ -15,6 +15,8 @@
  */
 package com.linkedin.pinot.controller.api.resources;
 
+import com.linkedin.pinot.common.config.TableNameBuilder;
+import com.linkedin.pinot.common.exception.InvalidConfigException;
 import com.linkedin.pinot.controller.helix.core.PinotHelixResourceManager;
 import com.linkedin.pinot.controller.helix.core.PinotResourceManagerResponse;
 import io.swagger.annotations.Api;
@@ -28,7 +30,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
-import org.json.JSONException;
+import javax.ws.rs.core.Response;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -46,15 +48,21 @@ public class PinotTableTenantConfigs {
   @Produces(MediaType.APPLICATION_JSON)
   @Path("/tables/{tableName}/rebuildBrokerResourceFromHelixTags")
   @ApiOperation(value = "Rebuild broker resource for table", notes = "when new brokers are added")
-  @ApiResponses(value = {@ApiResponse(code=200, message = "Success"),
-      @ApiResponse(code = 404, message = "Table not found"),
+  @ApiResponses(value = {
+      @ApiResponse(code = 200, message = "Success"),
+      @ApiResponse(code = 400, message = "Bad request: table name has to be with table type"),
       @ApiResponse(code = 500, message = "Internal error rebuilding broker resource or serializing response")})
-  public String rebuildBrokerResource(
+  public SuccessResponse rebuildBrokerResource(
       @ApiParam(value = "Table name (with type)", required = true) @PathParam("tableName") String tableNameWithType
-  ) throws JSONException {
-    final PinotResourceManagerResponse pinotResourceManagerResponse =
-        _helixResourceManager.rebuildBrokerResourceFromHelixTags(tableNameWithType);
-
-    return pinotResourceManagerResponse.toJSON().toString();
+  ) {
+    try {
+      final PinotResourceManagerResponse pinotResourceManagerResponse =
+          _helixResourceManager.rebuildBrokerResourceFromHelixTags(tableNameWithType);
+      return new SuccessResponse(pinotResourceManagerResponse.getMessage());
+    } catch (InvalidConfigException e) {
+      throw new ControllerApplicationException(LOGGER, e.getMessage(), Response.Status.BAD_REQUEST);
+    } catch (Exception e) {
+      throw new ControllerApplicationException(LOGGER, e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR);
+    }
   }
 }
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -20,6 +20,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.HashBiMap;
 import com.google.common.util.concurrent.Uninterruptibles;
+import com.linkedin.pinot.common.Utils;
 import com.linkedin.pinot.common.config.IndexingConfig;
 import com.linkedin.pinot.common.config.SegmentsValidationAndRetentionConfig;
 import com.linkedin.pinot.common.config.TableConfig;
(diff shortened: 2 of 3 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableTenantConfigs.java
```diff
@@ -49,12 +50,18 @@
...
       @ApiParam(value = "Table name (with type)", required = true) @PathParam("tableName") String tableNameWithType
-  ) throws JSONException {
+  ) {
     final PinotResourceManagerResponse pinotResourceManagerResponse =
         _helixResourceManager.rebuildBrokerResourceFromHelixTags(tableNameWithType);
 
-    return pinotResourceManagerResponse.toJSON().toString();
+    if (pinotResourceManagerResponse.isSuccessful()) {
+      return new SuccessResponse(pinotResourceManagerResponse.getMessage());
+    } else if (pinotResourceManagerResponse.getMessage().contains("does not exist")) {
```
snleee: Either throw the exception from `rebuildBrokerResourceFromHelixTags()` and catch here, or you can throw one kind of exception for all errors. Or, you can move table check here (Many ways to achieve this). Let's not catch it by looking at the message.

mcvsubbu: +1
throw a controllerApplicationException with the right message and status code.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -550,20 +553,18 @@ private PinotResourceManagerResponse scaleUpBroker(Tenant tenant, PinotResourceM
...
-        return new PinotResourceManagerResponse("Table " + tableNameWithType + " does not exist", false);
-      }
-      tenantConfig = tableConfig.getTenantConfig();
+      tableConfig = ZKMetadataProvider.getTableConfig(_propertyStore, tableNameWithType);
     } catch (Exception e) {
-      LOGGER.warn("Caught exception while getting tenant config for table {}", tableNameWithType, e);
-      return new PinotResourceManagerResponse(
-          "Failed to fetch broker tag for table " + tableNameWithType + " due to exception: " + e.getMessage(), false);
+      LOGGER.warn("Caught exception while getting table config for table {}", tableNameWithType, e);
+      throw new InternalServerErrorException("Failed to fetch broker tag for table " + tableNameWithType + " due to exception: " + e.getMessage());
```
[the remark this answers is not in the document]
mcvsubbu: The exception message is that we cannot fetch table config. The error message should be "Invalid table configuration for table " + tableName.  It is not an InternalServerException, rather InvalidConfigException.
Same if tableconfig is null, so you can fold it here.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -577,8 +578,7 @@ public PinotResourceManagerResponse rebuildBrokerResourceFromHelixTags(@Nonnull
     Set<String> idealStateBrokerInstances = brokerIdealState.getInstanceSet(tableNameWithType);
 
     if (idealStateBrokerInstances.equals(brokerInstances)) {
-      return new PinotResourceManagerResponse(
-          "Broker resource is not rebuilt because ideal state is the same for table {} " + tableNameWithType, false);
+      throw new BadRequestException("Broker resource is not rebuilt because ideal state is the same for table: " + tableNameWithType);
```
mcvsubbu: this is not a bad request. It is an idempotent operation, so why throw exception ? why not return success?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableTenantConfigs.java
```diff
@@ -46,15 +46,20 @@
...
-    final PinotResourceManagerResponse pinotResourceManagerResponse =
-        _helixResourceManager.rebuildBrokerResourceFromHelixTags(tableNameWithType);
-
-    return pinotResourceManagerResponse.toJSON().toString();
+  ) {
+    try {
+      final PinotResourceManagerResponse pinotResourceManagerResponse =
+          _helixResourceManager.rebuildBrokerResourceFromHelixTags(tableNameWithType);
+      return new SuccessResponse(pinotResourceManagerResponse.getMessage());
+    } catch (WebApplicationException e) {
```
mcvsubbu: You should catch the exceptions that are being thrown in PinotHelixResourceManager for this call sequence. In this case, it should be InvalidConfig, in which case you return 4xx, and any other in which case you return 5xx

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -77,6 +77,9 @@
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import javax.ws.rs.BadRequestException;
```
snleee: Let's clean up unused exceptions

pinot-controller/src/test/java/com/linkedin/pinot/controller/validation/ValidationManagerTest.java
```diff
@@ -112,7 +114,12 @@ public void testRebuildBrokerResourceWhenBrokerAdded() throws Exception {
     Assert.assertTrue(idealState.getInstanceSet(partitionName)
         .equals(_pinotHelixResourceManager.getAllInstancesForBrokerTenant(
             ControllerTenantNameBuilder.DEFAULT_TENANT_NAME)));
-    _pinotHelixResourceManager.rebuildBrokerResourceFromHelixTags(partitionName);
+    try {
+      _pinotHelixResourceManager.rebuildBrokerResourceFromHelixTags(partitionName);
+    } catch (BadRequestException e) {
```
snleee: this is not used

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableTenantConfigs.java
```diff
@@ -46,15 +47,22 @@
...
-    return pinotResourceManagerResponse.toJSON().toString();
+  ) {
+    if (TableNameBuilder.extractRawTableName(tableNameWithType).equals(tableNameWithType)) {
+      throw new ControllerApplicationException(LOGGER, "Table name has to be with table type.", Response.Status.BAD_REQUEST);
+    }
+    try {
+      final PinotResourceManagerResponse pinotResourceManagerResponse =
+          _helixResourceManager.rebuildBrokerResourceFromHelixTags(tableNameWithType);
+      return new SuccessResponse(pinotResourceManagerResponse.getMessage());
+    } catch (PinotHelixResourceManager.InvalidTableConfigException e) {
```
snleee: Subbu meant `InvalidConfigException`

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableTenantConfigs.java
```diff
@@ -46,15 +47,22 @@
...
-  public String rebuildBrokerResource(
+  public SuccessResponse rebuildBrokerResource(
       @ApiParam(value = "Table name (with type)", required = true) @PathParam("tableName") String tableNameWithType
-  ) throws JSONException {
-    final PinotResourceManagerResponse pinotResourceManagerResponse =
-        _helixResourceManager.rebuildBrokerResourceFromHelixTags(tableNameWithType);
-
-    return pinotResourceManagerResponse.toJSON().toString();
+  ) {
+    if (TableNameBuilder.extractRawTableName(tableNameWithType).equals(tableNameWithType)) {
```
snleee: I don't think we need this check since you are already checking `tableConfig == null` within `rebuildBrokerResourceFromHelixTags()`

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -45,6 +45,7 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
+import javax.ws.rs.WebApplicationException;
```
snleee: remove this

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableTenantConfigs.java
jackjlli: But think it this way: whatever mistakes client makes we'll return 400, whatever the mistakes server has we'll return 500. Obviously if client gives a wrong table name, we shouldn't return a 500 since it's not a mistake from the server.

snleee: You can catch `InvalidConfigException` and throw 400 error since invalid config exception indicates there's something wrong with the user input.

PR #2758, opened 2018-04-25
Adding multiple rollup pinot segment reader

snleee:
1. Added the record reader that can rollup the multiple pinot segments
2. Added ValueAggregator interface and its implementation

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/readers/MultipleRollupPinotSegmentRecordReader.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/readers/MultipleRollupPinotSegmentRecordReader.java
@@ -1,0 +1,233 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/readers/MultipleRollupPinotSegmentRecordReader.java
```diff
@@ -0,0 +1,233 @@
...
+        row.putField(timeColumn, (int) _timeValueOutput[0]);
+        break;
+      case LONG:
+        _timeValueInput[0] = (Long) row.getValue(_schema.getTimeColumnName());
+        _dateTimeTransformer.transform(_timeValueInput, _timeValueOutput, 1);
+        row.putField(timeColumn, _timeValueOutput[0]);
+        break;
+      default:
+        throw new IllegalStateException(
+            "Time column granularity change is only supported for a time column with int or long data type");
```
npawar: Why are we skipping out String? For simple date format, the time column might be in string right?

npawar: I see that the config itself has only EpochToEpochTransformer. Why are we sticking to just epoch input format?

kishoreg:
Is there a high-level design on how we would structure the code? It will be great if we can split rollup/conversion into two units - a high level driver which is generic and another set of specific interface where we can plug in the real logic. This should make it easy/trivial to add additional logic.

Let's try to list out the type of roll-ups we would have. Some things that come to my mind
- Input: Set of segments, some configuration for roll-up.
- Operation:
-- A single row in the input is simply transformed to another row in the output.  This can potentially remove the row as well
-- Multiple rows in the input result in 1 row in the output. This is your implementation of time aggregation. But that can be generalized by saying the driver will simply group by X columns and collect all the rows for remaining columns and pass it to another implementation. 
-- Sorting: Sort the data and here is the sort order
-- Add index: Adds star tree index or inverted index

Most of the things we would do on a segment will fall into one of the above categories. It will be great to come up with common drivers for each category.

PR #2761, opened 2018-04-26
Fix setting of time column name in create segment command

npawar:
Set time column name only if explicitly provided. Don't try to get it from segmentGeneratorConfig and set it back to segmentGeneratorConfig. If segmentGeneratorConfig has it, it will be used anyway.

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/CreateSegmentCommand.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/CreateSegmentCommand.java
@@ -280,16 +280,6 @@ public class CreateSegmentCommand extends AbstractBaseAdminCommand implements Co
       }
     }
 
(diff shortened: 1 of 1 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -327,7 +317,9 @@ public boolean accept(File dir, String name) {
     segmentGeneratorConfig.setOverwrite(_overwrite);
     segmentGeneratorConfig.setTableName(_tableName);
     segmentGeneratorConfig.setSegmentName(_segmentName);
-    segmentGeneratorConfig.setTimeColumnName(_timeColumnName);
+    if (_timeColumnName != null) {
```
mcvsubbu: we don't need this, do we? we can call the setter whether or not _timeColumnName is null?

PR #2767, opened 2018-04-30
[TE] Remove dimensions which also have top k dimensions in heatmap breakdown

jihaozh:
Remove group by dimensions which also have top k dimensions heatmap breakdown.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/RootCauseMetricResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/RootCauseMetricResource.java
@@ -21,10 +21,10 @@ import com.wordnik.swagger.annotations.Api;
 import com.wordnik.swagger.annotations.ApiOperation;
 import com.wordnik.swagger.annotations.ApiParam;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/RootCauseMetricResource.java
```diff
@@ -356,11 +358,15 @@ public double getAggregate(@ApiParam(value = "metric urn", required = true) @Que
...
     for (int i = 0; i < data.size(); i++) {
       final String dimName = data.getString(COL_DIMENSION_NAME, i);
       final String dimValue = data.getString(COL_DIMENSION_VALUE, i);
       final double value = data.getDouble(COL_VALUE, i);
-
+      // remove group by dimensions which also have topk
+      if (dimNames.contains(dimName + TOP_K_POSTFIX)) {
+        continue;
+      }
       if (!output.containsKey(dimName)) {
```
kishoreg: can we use treemap to store output, so that the order of dimensions is alphabetical

jihaozh: yes, I can add that.

PR #2772, opened 2018-05-02
Prevent Realtime table creation if schema is missing

sunithabeeram:
Currently, a realtime table can be added without a schema. The missing schema is detected on the server when a offline->consuming transition is attempted. This change ensures the controller returns a 400 error when a realtime table is created without schema.

The change was tested manually, where a schema was deleted via zooinspector just before the add call is made. Verified that 400 is returned in this case.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -1031,7 +1031,8 @@ public class PinotHelixResourceManager {
    * @throws TableAlreadyExistsException for offline tables only if the table already exists
    */
   public void addTable(@Nonnull TableConfig tableConfig) throws IOException {
-    String tableNameWithType = tableConfig.getTableName();
+    final String tableName = tableConfig.getTableName();
+
     TenantConfig tenantConfig;
     if (isSingleTenantCluster()) {
       tenantConfig = new TenantConfig();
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1088,8 +1089,14 @@ public void addTable(@Nonnull TableConfig tableConfig) throws IOException {
         break;
       case REALTIME:
         final String realtimeTableName = tableConfig.getTableName();
```
snleee: `realtimeTableName` should be the same as the `tableNameWithType`. Let's clean up to use one variable for the same information. You are using `realtimeTableName` for fetching schema while using `tableNameWithType` for exception log.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1031,7 +1031,8 @@ public Schema getTableSchema(@Nonnull String tableName) {
    * @throws TableAlreadyExistsException for offline tables only if the table already exists
    */
   public void addTable(@Nonnull TableConfig tableConfig) throws IOException {
-    String tableNameWithType = tableConfig.getTableName();
+    final String tableName = tableConfig.getTableName();
```
snleee: We prefer `tableNameWithType` when the table name has `_OFFLINE` or `_REALTIME` suffix. In this way, we can directly know that the variable contains the suffix.

snleee:
Can you make sure to pass all unit/integration tests before merging?
```
  PinotTableRestletResourceTest.testCreateTable:130->ControllerTest.sendPostRequest:157 » IO
  PinotTableRestletResourceTest.testTableMinReplication:139->testTableMinReplicationInternal:154->ControllerTest.sendPostRequest:157 » IO
  PinotTableRestletResourceTest.testUpdateTableConfig:194->ControllerTest.sendPostRequest:157 » IO
  TableViewsTest.setUp:78 » InvalidTableConfig No schema defined for realtime ta...
```

sunithabeeram:
@snleee thanks for the review! Sorry I missed the test failures. They've been fixed in https://github.com/linkedin/pinot/pull/2777. (I had issues pushing my local changes to the branch, hence closed PR/deleted the branch)

PR #2777, opened 2018-05-03
Prevent Realtime table creation if schema is missing

sunithabeeram:
Currently, a realtime table can be added without a schema. The missing schema is detected on the server when a offline->consuming transition is attempted. This change ensures the controller returns a 400 error when a realtime table is created without schema.

Changes include:
- unit tests to test the realtime table creation fails if there is no schema
- new method in base ControllerTest to add a default schema to use with realtime tables.

### DIFF
--- a/pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/ControllerTest.java
+++ b/pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/ControllerTest.java
@@ -15,6 +15,11 @@
  */
 package com.linkedin.pinot.controller.helix;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.linkedin.pinot.common.data.DimensionFieldSpec;
+import com.linkedin.pinot.common.data.FieldSpec;
+import com.linkedin.pinot.common.data.MetricFieldSpec;
+import com.linkedin.pinot.common.data.Schema;
 import com.linkedin.pinot.common.utils.ZkStarter;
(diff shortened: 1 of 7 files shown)


snleee:
Can you try to clone this repo directly instead of using fork? Maybe that's why you cannot merge.

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/ControllerTest.java
```diff
@@ -137,6 +144,30 @@ protected void stopController() {
     _zkClient.close();
   }
 
+  protected void addDefaultSchema(String tableName) throws JSONException, IOException {
+    JSONObject schemaJson = new JSONObject();
```
snleee: You can build the schema object and call `schema.getJSONSchema()` to get json string.

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/ControllerTest.java
```diff
@@ -49,6 +54,7 @@
  */
 public abstract class ControllerTest {
   public static final String LOCAL_HOST = "localhost";
```
snleee: Is this used outside of this class? Otherwise, let's change it to `private`

sunithabeeram: Its from previous version of the code; will fix it in a followup PR; will merge the current changes.

PR #2778, opened 2018-05-03
Fix PinotHelixTaskResourceManager to allow concurrent jobs

Jackie-Jiang:
1. Allow concurrent jobs for Minion
2. Make jobs independent with each other
3. Add support to control number of concurrent tasks per instance


pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerStarter.java
```diff
@@ -119,176 +121,149 @@ public void start() {
     MetricsHelper.registerMetricsRegistry(_metricsRegistry);
 
     // Start all components
+    LOGGER.info("initializing segment fetchers for all protocols");
```
mayankshriv: (OCD) Proper capitalization

PR #2782, opened 2018-05-04
Fix updating realtime table config

mcvsubbu:
Although the metadata was updated correctly, the PUT call returned an exception
because it found the kafka partition directory. We should not be checkig for the
kafka partition directory anymore

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/metadata/ZKMetadataProvider.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/metadata/ZKMetadataProvider.java
@@ -26,7 +26,6 @@ import com.linkedin.pinot.common.utils.CommonConstants;
 import com.linkedin.pinot.common.utils.SchemaUtils;
 import com.linkedin.pinot.common.utils.SegmentName;
 import com.linkedin.pinot.common.utils.StringUtil;
-import com.sun.istack.internal.NotNull;
 import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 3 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/metadata/ZKMetadataProvider.java
```diff
@@ -26,6 +26,7 @@
 import com.linkedin.pinot.common.utils.SchemaUtils;
 import com.linkedin.pinot.common.utils.SegmentName;
 import com.linkedin.pinot.common.utils.StringUtil;
+import com.sun.istack.internal.NotNull;
```
Jackie-Jiang: Wrong import

pinot-common/src/main/java/com/linkedin/pinot/common/metadata/ZKMetadataProvider.java
```diff
@@ -323,7 +325,7 @@ public static Schema getTableSchema(@Nonnull ZkHelixPropertyStore<ZNRecord> prop
     return resultList;
   }
 
-  public static List<LLCRealtimeSegmentZKMetadata> getLLCRealtimeSegmentZKMetadataListForTable(
+  public static @NotNull List<LLCRealtimeSegmentZKMetadata> getLLCRealtimeSegmentZKMetadataListForTable(
```
Jackie-Jiang: We usually put @NonNull in front of the method declaration as a separate line (refer to https://www.programcreek.com/java-api-examples/?api=javax.annotation.Nonnull)

mcvsubbu: Done

mcvsubbu:
Right now, the only way to test this is by adding integration test. I would prefer that we make pinotHelixResourceManager unit-testable instead of adding an integration for each of these cases, since it increases travis time.

PR #2783, opened 2018-05-04
Segment size based flush threshold update

npawar:
Adding a segment size based flush threshold updater which updates the flush threshold rows of the new segment metadata, based on the segment size and number of rows of previous segment
The formula used to compute new number of rows is:
  `targetNumRows = ideal_segment_size * ( a * current_rows_to_size_ratio + b * previous_rows_to_size_ratio)`
  where a = 0.25, b = 0.75
  This ensures that we take into account the history of the segment size and number rows

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -560,7 +560,8 @@ public class PinotLLCRealtimeSegmentManager {
         partitionAssignment.getNumPartitions());
     final LLCRealtimeSegmentZKMetadata newSegmentZKMetadata = new LLCRealtimeSegmentZKMetadata(newZnRecord);
 
-    FlushThresholdUpdater flushThresholdUpdater = getFlushThresholdUpdater(realtimeTableConfig);
+    FlushThresholdUpdater flushThresholdUpdater =
+        _flushThresholdUpdateManager.getFlushThresholdUpdater(realtimeTableConfig);
     FlushThresholdUpdaterParams params = new FlushThresholdUpdaterParams();
     params.setCommittingSegmentSizeBytes(committingSegmentSizeBytes);
     params.setPartitionAssignment(partitionAssignment);
@@ -587,10 +588,6 @@ public class PinotLLCRealtimeSegmentManager {
           newSegmentNameStr, rawTableName, isLeader(), isConnected());
     }
     return success;
-  }
-
-  protected FlushThresholdUpdater getFlushThresholdUpdater(TableConfig realtimeTableConfig) {
-    return _flushThresholdUpdateManager.getFlushThresholdUpdater(realtimeTableConfig);
   }
 
   /**
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -50,7 +50,7 @@ import com.linkedin.pinot.controller.helix.core.PinotHelixSegmentOnlineOfflineSt
 import com.linkedin.pinot.controller.helix.core.PinotTableIdealStateBuilder;
 import com.linkedin.pinot.controller.helix.core.realtime.segment.FlushThresholdUpdateManager;
 import com.linkedin.pinot.controller.helix.core.realtime.segment.FlushThresholdUpdater;
-import com.linkedin.pinot.controller.helix.core.realtime.segment.FlushThresholdUpdaterParams;
+import com.linkedin.pinot.controller.helix.core.realtime.segment.CommittingSegmentDescriptor;
 import com.linkedin.pinot.controller.util.SegmentCompletionUtils;
 import com.linkedin.pinot.core.realtime.impl.kafka.SimpleConsumerWrapper;
 import com.linkedin.pinot.core.realtime.segment.ConsumingSegmentAssignmentStrategy;
@@ -112,7 +112,7 @@ public class PinotLLCRealtimeSegmentManager {
   /**
    * After step 1 of segment completion is done,
    * this is the max time until which step 3 is allowed to complete.
-   * See {@link PinotLLCRealtimeSegmentManager#commitSegmentMetadata(String, String, long, SegmentCompletionProtocol.Request.Params)} for explanation of steps 1 2 3
+   * See {@link PinotLLCRealtimeSegmentManager#commitSegmentMetadata(String, SegmentCompletionProtocol.Request.Params)} for explanation of steps 1 2 3
    * This includes any backoffs and retries for the steps 2 and 3
    * The segment will be eligible for repairs by the validation manager, if the time  exceeds this value
    */
@@ -561,11 +561,10 @@ public class PinotLLCRealtimeSegmentManager {
 
     FlushThresholdUpdater flushThresholdUpdater =
         _flushThresholdUpdateManager.getFlushThresholdUpdater(realtimeTableConfig);
-    FlushThresholdUpdaterParams params = new FlushThresholdUpdaterParams();
-    params.setCommittingSegmentSizeBytes(committingSegmentSizeBytes);
-    params.setPartitionAssignment(partitionAssignment);
-    params.setCommittingSegmentZkMetadata(committingSegmentMetadata);
-    flushThresholdUpdater.updateFlushThreshold(newSegmentZKMetadata, params);
+    CommittingSegmentDescriptor committingSegmentDescriptor = new CommittingSegmentDescriptor();
+    committingSegmentDescriptor.setCommittingSegmentSizeBytes(committingSegmentSizeBytes);
+    committingSegmentDescriptor.setCommittingSegmentZkMetadata(committingSegmentMetadata);
+    flushThresholdUpdater.updateFlushThreshold(newSegmentZKMetadata, committingSegmentDescriptor, partitionAssignment);
 
     newZnRecord = newSegmentZKMetadata.toZNRecord();
 
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -48,8 +48,9 @@ import com.linkedin.pinot.controller.api.events.MetadataEventNotifierFactory;
 import com.linkedin.pinot.controller.helix.core.PinotHelixResourceManager;
 import com.linkedin.pinot.controller.helix.core.PinotHelixSegmentOnlineOfflineStateModelGenerator;
 import com.linkedin.pinot.controller.helix.core.PinotTableIdealStateBuilder;
+import com.linkedin.pinot.controller.helix.core.realtime.segment.FlushThresholdUpdateManager;
 import com.linkedin.pinot.controller.helix.core.realtime.segment.FlushThresholdUpdater;
-import com.linkedin.pinot.controller.helix.core.realtime.segment.FlushThresholdUpdaterFactory;
+import com.linkedin.pinot.controller.helix.core.realtime.segment.FlushThresholdUpdaterParams;
 import com.linkedin.pinot.controller.util.SegmentCompletionUtils;
 import com.linkedin.pinot.core.realtime.impl.kafka.SimpleConsumerWrapper;
 import com.linkedin.pinot.core.realtime.segment.ConsumingSegmentAssignmentStrategy;
@@ -113,7 +114,7 @@ public class PinotLLCRealtimeSegmentManager {
   /**
    * After step 1 of segment completion is done,
    * this is the max time until which step 3 is allowed to complete.
-   * See {@link PinotLLCRealtimeSegmentManager#commitSegmentMetadata(String, String, long, long)} for explanation of steps 1 2 3
+   * See {@link PinotLLCRealtimeSegmentManager#commitSegmentMetadata(String, String, long, long, long)} for explanation of steps 1 2 3
    * This includes any backoffs and retries for the steps 2 and 3
    * The segment will be eligible for repairs by the validation manager, if the time  exceeds this value
    */
@@ -132,6 +133,7 @@ public class PinotLLCRealtimeSegmentManager {
   private final Lock[] _idealstateUpdateLocks;
   private final TableConfigCache _tableConfigCache;
   private final PartitionAssignmentGenerator _partitionAssignmentGenerator;
+  private final FlushThresholdUpdateManager _flushThresholdUpdateManager;
 
   public boolean getIsSplitCommitEnabled() {
     return _controllerConf.getAcceptSplitCommit();
@@ -184,6 +186,7 @@ public class PinotLLCRealtimeSegmentManager {
     }
     _tableConfigCache = new TableConfigCache(_propertyStore);
     _partitionAssignmentGenerator = new PartitionAssignmentGenerator(_helixManager);
+    _flushThresholdUpdateManager = new FlushThresholdUpdateManager();
   }
 
   public static PinotLLCRealtimeSegmentManager getInstance() {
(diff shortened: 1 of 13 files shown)


npawar:
> How is the initial targetNumRows calculated when current_rows_to_size_ratio and previous_rows_to_size_ratio are not available.

We begin with INITIAL_ROWS_THRESHOLD = 100_000

> whats the idea behind this formula
targetNumRows = ideal_segment_size * ( a * current_rows_to_size_ratio + b * previous_rows_to_size_ratio)

this aims to use more than 1 ratio data point, so that we do not sway a lot because of some segment which was too large or too small

>is rows_to_size_ratio calculated on per partition basis or averaged across all partitions?

rows_to_size_ratio is calculated from the previous 2 committing segments, not specific to partitions

> will this result in every segment having different number of rows. I think we can retain the same targetNumRows as long as the segment size is within ideal_segment_size +/- x percent. If it goes out of bounds for X consecutive segments, we can apply the formula to compute new targetNumRows.

yes each segment will have different number of rows. Why is that a problem? That was happening anyway in the current design when the segment completed on time threshold (which it mostly did)

kishoreg:
Thanks. On the last point, I was referring to the fact that we change the threshold (rows, time, memory used) frequently. Currently, we have this threshold in some config and it's static. Now we are changing it to be dynamic for every segment. I am not saying its bad and it's actually good to have that ability but a bit worried that we might be reacting too fast and one or two bad segments might get us into a bad state. Let's make sure we have some checks in place to avoid extreme edge cases. 

Follow up question, how are we communicating the new flush thresholds to individual servers?

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/segment/FlushThresholdUpdaterTest.java
```diff
@@ -114,18 +120,355 @@ private String makeFakeSegmentName(int id) {
...
+        SegmentSizeBasedFlushThresholdUpdater.INITIAL_ROWS_THRESHOLD);
+    Assert.assertEquals(segmentMetadata_p0_0.getTimeThresholdToFlushSegment(),
+        SegmentSizeBasedFlushThresholdUpdater.MAX_TIME_THRESHOLD);
 
-    public FakeFlushThresholdUpdater(TableConfig realtimeTableConfig, int flushSize) {
-      super(realtimeTableConfig);
-      _flushSize = flushSize;
-    }
+    // partition:0 segment:0 commits with 200M size
+    segmentSizeBytes = 200 * 1024 * 1024;
```
mcvsubbu: use a visible for testing getter to get the constants from FlushThrsholdUpdateManager, and avoid the tests knowing what those values actally are.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -554,21 +538,35 @@ protected boolean updateOldSegmentMetadataZNRecord(String realtimeTableName, LLC
...
    * @param nextOffset - start offset for new segment
    * @param partitionAssignment - stream partition assignment for this table
+   * @param committingSegmentSizeBytes - size of the committing segment
    * @return
    */
-  protected boolean createNewSegmentMetadataZNRecord(TableConfig realtimeTableConfig, LLCSegmentName newLLCSegmentName,
-      long nextOffset, PartitionAssignment partitionAssignment) {
+  protected boolean createNewSegmentMetadataZNRecord(TableConfig realtimeTableConfig,
+      LLCSegmentName committingLLCSegmentName, LLCSegmentName newLLCSegmentName, long nextOffset,
+      PartitionAssignment partitionAssignment, long committingSegmentSizeBytes) {
```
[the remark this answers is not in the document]
npawar: If committingSegmentSizeBytes <=0 we use threshold from previous segment. Is that not enough?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+public class SegmentSizeBasedFlushThresholdUpdater implements FlushThresholdUpdater {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(SegmentSizeBasedFlushThresholdUpdater.class);
+
+  static final long IDEAL_SEGMENT_SIZE_BYTES = 500 * 1024 * 1024;
+  private static final double MIN_ALLOWED_SEGMENT_SIZE_BYTES = 250 * 1024 * 1024;
+  private static final double MAX_ALLOWED_SEGMENT_SIZE_BYTES = 750 * 1024 * 1024;
+  static final int INITIAL_ROWS_THRESHOLD = 100_000;
+  static final String MAX_TIME_THRESHOLD = "24h";
```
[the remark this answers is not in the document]
npawar: Removed the check for time threshold from segment in LLRealtimeDataManager, and removed setting of time threshold in segment metadata

npawar:
Each time a segment commits, it computes the new flush threshold for the next segment, and it is recorded in the next segment's metadata. The servers check the segment metadata of the segment that is consuming

mcvsubbu:
@kishoreg the formula for computing the segment size will favor history more than the most recent value. Currently, history is favored at 75% and most recent value at 25%. So, while there is a variation of number of rows for each segment, it should be small. @npawar will also be adding a unit test that iterates through the algorithm 100s of times for some random variation of segment sizes.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
```diff
@@ -75,25 +74,14 @@ public void updateFlushThreshold(@Nonnull LLCRealtimeSegmentZKMetadata newSegmen
...
-
-    if (timeConsumed >= TimeUtils.convertPeriodToMillis(MAX_TIME_THRESHOLD)) {
-      LOGGER.info("Time threshold {} reached for committing segment {}", MAX_TIME_THRESHOLD,
-          committingSegmentZkMetadata.getSegmentName());
-    } else if (numRowsConsumed >= numRowsThreshold) {
-      LOGGER.info("Number of rows threshold {} reached for segment {}", numRowsThreshold,
-          committingSegmentZkMetadata.getSegmentName());
-    }
+    LOGGER.info("Time consumed:{}  Num rows consumed:{} Num rows threshold:{} Committing segment size bytes:{}",
+        TimeUtils.convertMillisToPeriod(timeConsumed), numRowsConsumed, numRowsThreshold, committingSegmentSizeBytes);
```
mcvsubbu: The prevRatio should be the last calculated/applied ratio. You need to save it in the hash for that table so that you can use that next time a segment completes.

npawar: Done.

pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/segment/FlushThresholdUpdaterTest.java
```diff
@@ -21,19 +21,46 @@
...
+  private Map<String, double[][]> datasetGraph;
+
+  @BeforeClass
+  public void setup() {
+    long seed = new Random().nextLong();
+    System.out.println("Random seed for " + FlushThresholdUpdater.class.getSimpleName() + " is " + seed);
+    _random = new Random(seed);
+
+    datasetGraph = new HashMap<>(3);
+    double[][] exponentialGrowth = {{100000, 50}, {200000, 60}, {300000, 70}, {400000, 83}, {500000, 98}, {600000, 120},
```
mcvsubbu: gj on testing

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/FlushThresholdUpdateManager.java
```diff
@@ -0,0 +1,101 @@
...
+   * If flush size == 0, create new SegmentSizeBasedFlushThresholdUpdater if not already created. Create only 1 per table, because we want to maintain tuning information for the table in the updater
+   * @param realtimeTableConfig
+   * @return
+   */
+  public FlushThresholdUpdater getFlushThresholdUpdater(TableConfig realtimeTableConfig) {
+    String tableName = realtimeTableConfig.getTableName();
+
+    int tableFlushSize = getLLCRealtimeTableFlushSize(realtimeTableConfig);
+    if (tableFlushSize < 0) {
+      tableFlushSize = KafkaHighLevelStreamProviderConfig.getDefaultMaxRealtimeRowsCount();
```
mcvsubbu: Shouldn't you get it from KafkaLowLevelStreamProvviderConfig? I think it goes to high level config if low level isnot configured.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/FlushThresholdUpdaterParams.java
```diff
@@ -0,0 +1,54 @@
...
+package com.linkedin.pinot.controller.helix.core.realtime.segment;
+
+import com.linkedin.pinot.common.metadata.segment.LLCRealtimeSegmentZKMetadata;
+import com.linkedin.pinot.common.partition.PartitionAssignment;
+
+
+/**
+ * Class to hold params required by flush threshold updater strategies
+ */
+public class FlushThresholdUpdaterParams {
```
mcvsubbu: A better way to split this may be to introduce a class CommitingSegmentDescriptor, which has most of the fields in RequestParams. PartitionAssingment need not be in this class, but can be a separate argument to the flush threshold updater

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
```diff
@@ -996,12 +995,12 @@ private int numReplicasToLookFor() {
       // In case of splitCommit, the segment is uploaded to a unique file name indicated by segmentLocation,
       // so we need to move the segment file to its permanent location first before committing the metadata.
       if (isSplitCommit) {
-        if (!_segmentManager.commitSegmentFile(_segmentName.getTableName(), segmentLocation, _segmentName.getSegmentName())) {
+        if (!_segmentManager.commitSegmentFile(_segmentName.getTableName(), reqParams.getSegmentLocation(),
```
mcvsubbu: Use CommittingSegmentDescriptor when calling this API. Keep the usage of ReqParams local to this file

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
```diff
@@ -996,12 +995,12 @@ private int numReplicasToLookFor() {
...
+            _segmentName.getSegmentName())) {
           return SegmentCompletionProtocol.RESP_FAILED;
         }
       }
-      success = _segmentManager.commitSegmentMetadata(_segmentName.getTableName(), _segmentName.getSegmentName(),
-          _winningOffset, memoryUsedBytes);
+      success = _segmentManager.commitSegmentMetadata(_segmentName.getTableName(), reqParams);
```
mcvsubbu: Use CommittingSegmentDescriptor here

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/FlushThresholdUpdateManager.java
mcvsubbu: sorry, my bad

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
```diff
@@ -0,0 +1,147 @@
...
+      }
+      return;
+    }
+
+    long committingSegmentSizeBytes = committingSegmentDescriptor.getCommittingSegmentSizeBytes();
+    if (committingSegmentSizeBytes <= 0) { // repair segment case
+      LOGGER.info(
+          "Committing segment size is not available, setting thresholds for segment {} from previous segment {}",
+          newSegmentZKMetadata.getSegmentName(), committingSegmentZkMetadata.getSegmentName());
+      newSegmentZKMetadata.setSizeThresholdToFlushSegment(committingSegmentZkMetadata.getSizeThresholdToFlushSegment());
```
mcvsubbu: good to add a comment here that we have an option to use numRowsConsumed here, but we may be bouncing back & forth between that value and twice that value (if it is much lower). 
Perhaps we can set it to numRowsConsumed * 1.1?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
```diff
@@ -107,11 +113,16 @@ public void updateFlushThreshold(@Nonnull LLCRealtimeSegmentZKMetadata newSegmen
...
-      LOGGER.info("Segment {} reached time threshold, setting thresholds for segment {} from previous segment",
-          committingSegmentZKMetadata.getSegmentName(), newSegmentZKMetadata.getSegmentName());
       // TODO: add feature to adjust time threshold as well
-      // TODO: setting sizeThresholdToFlushSegment instead of numRowsConsumed, otherwise we can end up oscillating back and forth
-      newSegmentZKMetadata.setSizeThresholdToFlushSegment(committingSegmentZKMetadata.getSizeThresholdToFlushSegment());
+      // If we set new threshold to be numRowsConsumed, we might keep oscillating back and forth between doubling limit and time threshold being hit
+      // If we set new threshold to be committingSegmentZKMetadata.getSizeThresholdToFlushSegment(), we might end up using a lot more memory than required for the segment
+      // Using a minor bump strategy, until we add feature to adjust time
+      // We will only slightly bump the threshold based on numRowsConsumed
+      long targetSegmentNumRows = (long) (numRowsConsumed * 1.1);
```
mcvsubbu: Please make a static int for 1.1 and use it for the arithmetic as well as the log message

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/CommittingSegmentDescriptor.java
```diff
@@ -16,29 +16,67 @@
...
-  private LLCRealtimeSegmentZKMetadata _committingSegmentZkMetadata;
-  private long _committingSegmentSizeBytes;
+  private String _segmentName;
+  private long _segmentSizeBytes;
+  private String _segmentLocation;
+  private long _nextOffset;
 
-  public LLCRealtimeSegmentZKMetadata getCommittingSegmentZkMetadata() {
-    return _committingSegmentZkMetadata;
+  public static CommittingSegmentDescriptor fromSegmentCompletionReqParams(
```
mcvsubbu: Or, you can have a constructor that takes RequestParams in the argument and copies over. That may be the only constructor we will need?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/CommittingSegmentDescriptor.java
```diff
@@ -16,29 +16,67 @@
...
+    CommittingSegmentDescriptor committingSegmentDescriptor =
+        new CommittingSegmentDescriptor(reqParams.getSegmentName(), reqParams.getOffset(),
+            reqParams.getSegmentSizeBytes());
+    committingSegmentDescriptor.setSegmentLocation(reqParams.getSegmentLocation());
+    return committingSegmentDescriptor;
   }
 
-  public void setCommittingSegmentZkMetadata(LLCRealtimeSegmentZKMetadata committingSegmentZkMetadata) {
-    _committingSegmentZkMetadata = committingSegmentZkMetadata;
+  public CommittingSegmentDescriptor(String segmentName, long nextOffset, long segmentSizeBytes) {
```
mcvsubbu: if we dont need these constructrs we can delete them

PR #2786, opened 2018-05-08
Avoid exception for BYTES size() call when not generating Star Tree.

mayankshriv:
Only StartTree generation requires to know size of a metric. However,
for BYTES, we cannot define a size upfront.

1. Return UNDEFINED_METRIC_SIZE for getSize() on BYTES data type.

2. Modify test to define TDigest as metric to exercise the code.

Star Tree generation will continue to not work, but at least no Star
Tree code will not throw exception.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
@@ -383,6 +383,9 @@ public abstract class FieldSpec {
           return Float.BYTES;
         case DOUBLE:
           return Double.BYTES;
+        case BYTES:
+          // TODO: Metric size is only used for Star-tree generation, which is not supported yet.
+          return MetricFieldSpec.UNDEFINED_METRIC_SIZE;
         default:
           throw new IllegalStateException("Cannot get number of bytes for: " + this);
       }
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/MetricFieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/MetricFieldSpec.java
@@ -33,10 +33,10 @@ import org.codehaus.jackson.annotate.JsonIgnoreProperties;
 @SuppressWarnings("unused")
 @JsonIgnoreProperties(ignoreUnknown = true)
 public final class MetricFieldSpec extends FieldSpec {
-  private static final int UNDEFINED_FIELD_SIZE = -1;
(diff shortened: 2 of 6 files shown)


kishoreg:
this is ok for now, but we  need a better way to handle these type of metrics. We should be able to skip metrics similar to skipping dimensions with high cardinality.

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
```diff
@@ -383,6 +383,9 @@ public int size() {
           return Float.BYTES;
         case DOUBLE:
           return Double.BYTES;
+        case BYTES:
```
Jackie-Jiang: Don't add it here. This method should never be called for STRING and BYTES

pinot-common/src/main/java/com/linkedin/pinot/common/data/MetricFieldSpec.java
```diff
@@ -33,10 +33,10 @@
 @SuppressWarnings("unused")
 @JsonIgnoreProperties(ignoreUnknown = true)
 public final class MetricFieldSpec extends FieldSpec {
-  private static final int UNDEFINED_FIELD_SIZE = -1;
+  protected static final int UNDEFINED_METRIC_SIZE = -1;
```
Jackie-Jiang: Don't export this variable

pinot-common/src/main/java/com/linkedin/pinot/common/data/MetricFieldSpec.java
```diff
@@ -104,7 +105,7 @@ public FieldType getFieldType() {
   @Override
   public void setDataType(@Nonnull DataType dataType) {
     super.setDataType(dataType);
-    if (_dataType != DataType.STRING) {
+    if (_dataType != DataType.STRING || _dataType != DataType.BYTES) {
```
Jackie-Jiang: Please add a TODO

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
mayankshriv: That requires this to be a derivedMetric, something we want to get rid of soon.

pinot-common/src/main/java/com/linkedin/pinot/common/data/MetricFieldSpec.java
mayankshriv: It is protected, and being used only in FieldSpec

PR #2787, opened 2018-05-08
Purge task optimization

jenniferdai:
* Don't purge segment unless there are records to be purged or modified

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/minion/SegmentPurger.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/minion/SegmentPurger.java
@@ -16,7 +16,6 @@
 package com.linkedin.pinot.core.minion;
 
 import com.google.common.base.Preconditions;
-import com.linkedin.pinot.common.Utils;
 import com.linkedin.pinot.common.data.Schema;
 import com.linkedin.pinot.common.data.StarTreeIndexSpec;
 import com.linkedin.pinot.common.segment.SegmentMetadata;
@@ -47,7 +46,6 @@ public class SegmentPurger {
   private final File _workingDir;
   private final RecordPurger _recordPurger;
   private final RecordModifier _recordModifier;
-  private final PurgeRecordReader _purgeRecordReader;
 
   private int _numRecordsPurged;
   private int _numRecordsModified;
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/minion/SegmentPurger.java
```diff
@@ -66,29 +89,29 @@ public File purgeSegment() throws Exception {
...
-      // Keep index creation time the same as original segment because both segments use the same raw data.
-      // This way, for REFRESH case, when new segment gets pushed to controller, we can use index creation time to
-      // identify if the new pushed segment has newer data than the existing one.
-      config.setCreationTime(String.valueOf(segmentMetadata.getIndexCreationTime()));
-
-      StarTreeMetadata starTreeMetadata = segmentMetadata.getStarTreeMetadata();
-      if (starTreeMetadata != null) {
-        config.enableStarTreeIndex(StarTreeIndexSpec.fromStarTreeMetadata(starTreeMetadata));
-      }
+    SegmentGeneratorConfig config = new SegmentGeneratorConfig(_purgeRecordReader.getSchema());
```
Jackie-Jiang: Don't remove this try-with-resource block because it will close the record reader

jenniferdai: My concern with returning null was making sure is intentional - for instance, if we somehow return null for a different reason, I would want to know/decouple that. Thoughts?

Jackie-Jiang: IMO, we should only return purged file, nulll, or throw exception. So using null to mark nothing need to be purged should be fine.

jenniferdai: makes sense, changed

pinot-core/src/main/java/com/linkedin/pinot/core/minion/SegmentPurger.java
```diff
@@ -67,6 +67,16 @@ public File purgeSegment() throws Exception {
...
 
     try (PurgeRecordReader purgeRecordReader = new PurgeRecordReader()) {
+      // Make a first pass through the data to see if records need to be purged or modified
+      while (purgeRecordReader.hasNext()) {
+        purgeRecordReader.next();
+      }
+
+      if(_numRecordsModified == 0 && _numRecordsPurged == 0) {
+        // Returns null if no records to be modified or purged
+        return null;
```
sunithabeeram: Do we have a metric that tracks this condition?

Jackie-Jiang: We don't. Actually this might happen pretty frequently because we will purge same segment per several days.

jenniferdai: I definitely agree that we should have a metric. We don't have any open source minion metrics right now, so it might be a little more involved. I filed a ticket for this change.

PR #2792, opened 2018-05-10
Handle AVRO files with ByteBuffer instead of byte[] in input.

mayankshriv:
For columns with data type BYTES, avro may contain ByteBuffer or byte[].
This changes enables ingestion of both data types, by performing the
required conversion.

Added new unit test for segment generation from avro with BYTES type.


pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/stats/BytesColumnPredIndexStatsCollector.java
```diff
@@ -59,11 +59,11 @@ public BytesColumnPredIndexStatsCollector(String column, StatsCollectorConfig st
   private void collectEntry(Object entry, ObjectSet<ByteArray> set) {
     if (entry instanceof Object[]) {
       for (final Object e : (Object[]) entry) {
-        byte[] value = (byte[]) e;
-        set.add(new ByteArray(value));
+        ByteArray value = ((ByteArray) e).clone(); // Underlying byte[] may be re-cycled.
```
Jackie-Jiang: After PlainFieldExtractor, this value should always be byte[]？

mayankshriv: No, If you look at the stats collector code, it expects comparable, hence the ByteArray wrapper.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/stats/BytesColumnPredIndexStatsCollector.java
```diff
@@ -59,11 +59,11 @@ public BytesColumnPredIndexStatsCollector(String column, StatsCollectorConfig st
   private void collectEntry(Object entry, ObjectSet<ByteArray> set) {
     if (entry instanceof Object[]) {
       for (final Object e : (Object[]) entry) {
-        byte[] value = (byte[]) e;
-        set.add(new ByteArray(value));
+        ByteArray value = ((ByteArray) e);
```
Jackie-Jiang: We don't support MV for BYTES. So I suggest throw UnsupportedOperationException here or add a TODO.
(OCD) Remove redundant parentheses.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/stats/BytesColumnPredIndexStatsCollector.java
```diff
@@ -73,7 +73,7 @@ private void collectEntry(Object entry, ObjectSet<ByteArray> set) {
     } else {
       ByteArray value;
       if (entry != null) {
-        value = (ByteArray) entry;
+        value = ((ByteArray) entry);
```
Jackie-Jiang: (OCD) Remove redundant parentheses.

pinot-core/src/main/java/com/linkedin/pinot/core/data/extractors/PlainFieldExtractor.java
```diff
@@ -209,6 +210,15 @@ public GenericRow transform(GenericRow row, GenericRow destinationRow) {
             hasError = true;
             _errorCount.put(column, _errorCount.get(column) + 1);
           }
+        } else if (value instanceof ByteBuffer) { // TODO: Need better handle for ByteBuffers.
+          // ByteBuffer implementations are package private and cannot be put into TYPE_MAP.
+          ByteBuffer byteBuffer = (ByteBuffer) value;
+
+          // Assumes byte-buffer is ready to read. Also, avoid getting underlying array, as it may be over-sized.
```
kishoreg: you can check if hasArray and use the bytebuffer.array to avoid creating byte[].

PR #2800, opened 2018-05-14
Replace catch Throwable by catch Exception

jfim:
Fix a couple of instances where Throwable is caught instead of
Exception.


npawar:

PR #2803, opened 2018-05-14
Use current time millis in task name for easier debugging

Jackie-Jiang:
Use current time millis instead of nano time so that we know when the task get scheduled for easier debugging


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -290,7 +290,7 @@ private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, Strin
 
     try {
       FileUploadPathProvider provider = new FileUploadPathProvider(_controllerConf);
-      String tempSegmentName = "tmp-" + System.nanoTime();
+      String tempSegmentName = "tmp-" + System.currentTimeMillis();
```
mayankshriv: I guess our assumption (which seems fair to me) is that we won't have collision at millisecond level?

Jackie-Jiang: To further ensure that, I will append nano time after current time

PR #2805, opened 2018-05-15
Add code to handle missing admin port 

jackjlli:
This PR:
- adds logs to find out which host connection is lost;
- uses default adminPort.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/ServerTableSizeReader.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/ServerTableSizeReader.java
@@ -84,19 +84,19 @@ public class ServerTableSizeReader {
         TableSizeInfo tableSizeInfo = new ObjectMapper().readValue(getMethod.getResponseBodyAsString(), TableSizeInfo.class);
         serverSegmentSizes.put(instance, tableSizeInfo.segments);
       } catch (InterruptedException e) {
-        LOGGER.warn("Interrupted exception while reading segment size for table: {}. Host endpoint: {}", table, url, e);
+        LOGGER.warn("Interrupted exception while reading segment size for table: {}. Server: {}", table, url, e);
       } catch (ExecutionException e) {
         if (Throwables.getRootCause(e) instanceof SocketTimeoutException) {
-          LOGGER.warn("Server request to read table size was timed out for table: {}. Host endpoint: {}", table, url, e);
+          LOGGER.warn("Server request to read table size was timed out for table: {}. Server: {}", table, url, e);
         } else if (Throwables.getRootCause(e) instanceof ConnectTimeoutException) {
-          LOGGER.warn("Server request to read table size timed out waiting for connection. table: {}. Host endpoint: {}", table, url, e);
+          LOGGER.warn("Server request to read table size timed out waiting for connection. table: {}. Server: {}", table, url, e);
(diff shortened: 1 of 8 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -79,6 +78,7 @@
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import joptsimple.internal.Strings;
```
snleee: I guess the this is not needed?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/ServerTableSizeReader.java
```diff
@@ -83,19 +84,19 @@ public ServerTableSizeReader(Executor executor, HttpConnectionManager connection
         TableSizeInfo tableSizeInfo = new ObjectMapper().readValue(getMethod.getResponseBodyAsString(), TableSizeInfo.class);
         serverSegmentSizes.put(instance, tableSizeInfo.segments);
       } catch (InterruptedException e) {
-        LOGGER.warn("Interrupted exception while reading segment size for table: {}", table, e);
+        LOGGER.warn("Interrupted exception while reading segment size for table: {}. Host endpoint: {}", table, url, e);
```
mcvsubbu: Instead of "Host endpoint", can you please change it to "Server"? thanks.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -2154,8 +2154,13 @@ public boolean isSingleTenantCluster() {
       ZNRecord record = helixInstanceConfig.getRecord();
       String[] hostnameSplit = helixInstanceConfig.getHostName().split("_");
       Preconditions.checkState(hostnameSplit.length >= 2);
-      String port = record.getSimpleField(CommonConstants.Helix.Instance.ADMIN_PORT_KEY);
-      endpointToInstance.put(instance, hostnameSplit[1] + ":" + port);
+      String adminPort = record.getSimpleField(CommonConstants.Helix.Instance.ADMIN_PORT_KEY);
+      if (Strings.isNullOrEmpty(adminPort)) {
+        LOGGER.warn("Admin port is missing for host: {}. Using the default port: {}",
+            hostnameSplit[1], CommonConstants.Server.DEFAULT_ADMIN_API_PORT);
+        adminPort = Integer.toString(CommonConstants.Server.DEFAULT_ADMIN_API_PORT);
```
mcvsubbu: You should not be importing Server constants into controller. Please remove the code to pick up default admin port, because the server updates its own instance config with the admin port information. If somehow that update is wrong, then we should not try connecting to that server at all. Please be sure to print helixInstanceConfig.getHostName() whenever the admin port is not present or not parsable.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: It's for checking whether the admin port is missing or not.

PR #2806, opened 2018-05-15
Fixed NPE bug for PartitionAwareOfflineRoutingTableBuilder

snleee:
PartitionAwareOfflineRoutingTableBuilder was caching the number of replicas
from the table config during the initialization and never updates the value
even if there was a change. This caused NPE when the broker attempts to compute
routing table after servers were added to table and rebalance was invoked.
This PR addresses this bug and also added the unit test to make sure the fix
resolves the issue.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/BasePartitionAwareRoutingTableBuilder.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/BasePartitionAwareRoutingTableBuilder.java
@@ -61,9 +61,8 @@ public abstract class BasePartitionAwareRoutingTableBuilder implements RoutingTa
 
   protected ZkHelixPropertyStore<ZNRecord> _propertyStore;
   protected SegmentZKMetadataPrunerService _pruner;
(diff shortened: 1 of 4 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/BasePartitionAwareRoutingTableBuilder.java
```diff
@@ -61,17 +61,15 @@
 
   protected ZkHelixPropertyStore<ZNRecord> _propertyStore;
   protected SegmentZKMetadataPrunerService _pruner;
-  protected TableConfig _tableConfig;
   protected Random _random = new Random();
-  protected int _numReplicas;
+  protected volatile int _numReplicas;
```
mayankshriv: Curious, why does this need to be volatile?

snleee: As discussed offline, write to volatile int is atomic.

PR #2819, opened 2018-05-16
Change tenant to tag in tenant configuration

jfim:


pinot-common/src/main/java/com/linkedin/pinot/common/config/TenantConfig.java
```diff
@@ -26,12 +26,12 @@
...
-  @ConfigKey("brokerTenant")
-  @ConfigDoc("Broker tenant used by this table")
+  @ConfigKey("brokerTag")
+  @ConfigDoc("Broker tag used by this table")
   private String broker;
 
-  @ConfigKey("serverTenant")
-  @ConfigDoc("Server tenant used by this table")
+  @ConfigKey("serverTag")
+  @ConfigDoc("Server tag used by this table")
```
npawar: Are we going to start keeping the whole tag name here (with suffix)? My understanding was that tag=<tenant_name>_<suffix>

PR #2821, opened 2018-05-17
Remove 'queryOptions' from PinotClientRequest

Jackie-Jiang:
Query options should be parsed from query instead of from HTTP parameters
E.g. SELECT COUNT(*) FROM table OPTION(key='value')


sunithabeeram:
Few questions:
1. What are some examples of query options passed via HTTP parameters currently?
2. How widely is this used? Does it break compatiblity with existing usage? 
3. Can you add tests to cover this?

PR #2822, opened 2018-05-17
Comment the test assertion

jackjlli:
This PR comments the test assertion.
We'll add it back once we figure out the reason why it fails.


jackjlli:
Problem is found in https://github.com/linkedin/pinot/pull/2823. No need to have this RP.

PR #2831, opened 2018-05-21
Use smallest available kafka offset when a new partition is detected

sunithabeeram:
When a new partition is detected, the code currently uses the table configuration to decide the offset to seek to. This change updates it to seek to the smallest offset - the rationale is that we would like the partitions to have as much data as possible to serve requests effectively. (This would mostly impact tables that have the offset set to 'largest').

Changes include tests to ensure that the right offset is sought when new tables are created and when new partitions are added.


pinot-controller/src/test/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTest.java
```diff
@@ -154,21 +157,23 @@ public void testSetupNewTable() {
...
     // noop path - ideal state disabled - this can happen in the code path only if there is already an idealstate with HLC segments in it, and it has been disabled.
-    tableConfig = makeTableConfig(tableName, nReplicas, KAFKA_OFFSET, DUMMY_HOST, DEFAULT_SERVER_TENANT);
+    tableConfig = makeTableConfig(tableName, nReplicas, KAFKA_LARGEST_OFFSET, DUMMY_HOST, DEFAULT_SERVER_TENANT);
     idealState = idealStateBuilder.disableIdealState().build();
     nPartitions = 4;
     invalidConfig = false;
     testSetupNewTable(tableConfig, idealState, nPartitions, nReplicas, instances, invalidConfig, badStream);
 
+    // create a new builder with idealstate enabled
+    idealStateBuilder = new IdealStateBuilderUtil(tableName);
```
mcvsubbu: this is fine, alternatively you can do idealStateBuilder.clear() as well I suppose.

kishoreg:
largest/smallest -> latest/earliest to be in sync with the terminology used by Kafka.

sunithabeeram:
Changing smallest to earliest will require a few more changes for it to be consistent - for example, the configuration expects string "smallest" to pick the earliest offset. It would be better if that cleanup is rolled into effort @npawar is handling for making a generic streaming interface.

PR #2835, opened 2018-05-22
Check if the untarred file will be put outside of the target output directory

jackjlli:
This PR adds code to check whether the untarred file will be put outside of the target output dir.


kishoreg:
do we have a test case to check for the happy path and the failure path?

PR #2837, opened 2018-05-22
Add unit test

jackjlli:
This PR adds a unit test for testing whether our code can handle a bad file name from a tar file if we untar the file.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/TarGzCompressionUtils.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/TarGzCompressionUtils.java
@@ -50,6 +50,8 @@ import org.slf4j.LoggerFactory;
 public class TarGzCompressionUtils {
   private static final Logger LOGGER = LoggerFactory.getLogger(TarGzCompressionUtils.class);
   public static final String TAR_GZ_FILE_EXTENTION = ".tar.gz";
+  public static final String EXTRACT_FILE_OUTSIDE_OF_TARGET_DIR =
+      "Tar file must not be untarred outside of the target output directory!";
 
   /**
    * Creates a tar.gz file at the specified path with the contents of the
@@ -64,12 +66,12 @@ public class TarGzCompressionUtils {
    * @throws IOException
    *           If anything goes wrong
    */
-  public static String createTarGzOfDirectory(String directoryPath, String tarGzPath)
-      throws IOException {
+  public static String createTarGzOfDirectory(String directoryPath, String tarGzPath) throws IOException {
     return createTarGzOfDirectory(directoryPath, tarGzPath, "");
   }
 
-  public static String createTarGzOfDirectory(String directoryPath, String tarGzPath, String entryPrefix) throws IOException {
+  public static String createTarGzOfDirectory(String directoryPath, String tarGzPath, String entryPrefix)
+      throws IOException {
     FileOutputStream fOut = null;
     BufferedOutputStream bOut = null;
     GzipCompressorOutputStream gzOut = null;
@@ -165,11 +167,11 @@ public class TarGzCompressionUtils {
    * @throws IOException
    * @throws FileNotFoundException
    *
-   * @return  The {@link List} of {@link File}s with the untared content.
+   * @return The {@link List} of {@link File}s with the untared content.
    * @throws ArchiveException
    */
-  public static List<File> unTar(final File inputFile, final File outputDir) throws FileNotFoundException, IOException,
-      ArchiveException {
+  public static List<File> unTar(final File inputFile, final File outputDir)
+      throws FileNotFoundException, IOException, ArchiveException {
 
     String outputDirectoryPath = outputDir.getCanonicalPath();
     LOGGER.debug(String.format("Untaring %s to dir %s.", inputFile.getAbsolutePath(), outputDirectoryPath));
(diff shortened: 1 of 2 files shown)


pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
```diff
@@ -122,4 +128,61 @@ public void testEmptyDirectory()
     Assert.assertEquals(segmentFiles.length, 0);
 
   }
+
+  @Test
+  public void testFailurePath() throws IOException, ArchiveException {
```
mcvsubbu: you can change it to just throw Exception

pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
```diff
@@ -122,4 +128,61 @@ public void testEmptyDirectory()
...
+    Assert.assertFalse(metaFile.isDirectory());
+
+    File tarGzPath = new File(tarDir, SEGMENT_NAME + ".tar.gz");
+    mockIncorrectData(metaFile, tarGzPath);
+
+    try {
+      TarGzCompressionUtils.unTar(tarGzPath, untarDir);
+      Assert.fail("Did not get exception!!");
+    } catch (IOException e) {
+      Assert.assertTrue(e.getMessage().contains("Tar file must not be untarred outside of the target output directory!"));
```
mcvsubbu: If you are checking for exact msg in the exception, then please define a public static string in the code and re-use it here.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/TarGzCompressionUtils.java
```diff
@@ -50,6 +50,8 @@
 public class TarGzCompressionUtils {
   private static final Logger LOGGER = LoggerFactory.getLogger(TarGzCompressionUtils.class);
   public static final String TAR_GZ_FILE_EXTENTION = ".tar.gz";
+  public static final String EXTRACT_FILE_OUTSIDE_OF_TARGET_DIR =
```
kishoreg: don't have to make this public just for testing purpose. just test for the exception type.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/TarGzCompressionUtils.java
```diff
@@ -165,11 +167,11 @@ private static void addFileToTarGz(TarArchiveOutputStream tOut, String path, Str
...
+   * @return The {@link List} of {@link File}s with the untared content.
    * @throws ArchiveException
    */
-  public static List<File> unTar(final File inputFile, final File outputDir) throws FileNotFoundException, IOException,
-      ArchiveException {
+  public static List<File> unTar(final File inputFile, final File outputDir)
+      throws FileNotFoundException, IOException, ArchiveException {
 
     String outputDirectoryPath = outputDir.getCanonicalPath();
     LOGGER.debug(String.format("Untaring %s to dir %s.", inputFile.getAbsolutePath(), outputDirectoryPath));
```
kishoreg: we have excessive usage of String.format, can we use slf4j api and remove the usage of String.format. Should avoid unnecessary string creation

pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
```diff
@@ -122,4 +128,61 @@ public void testEmptyDirectory()
...
+
+    try {
+      TarGzCompressionUtils.unTar(tarGzPath, untarDir);
+      Assert.fail("Did not get exception!!");
+    } catch (IOException e) {
+      Assert.assertTrue(e.getMessage().contains(TarGzCompressionUtils.EXTRACT_FILE_OUTSIDE_OF_TARGET_DIR));
+    }
+  }
+
+  private void mockIncorrectData(File nonDirFile, File tarGzPath) throws IOException {
```
kishoreg: why are we naming this mockIncorrectData? can we call it something like createInvalidTarFile

pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
```diff
@@ -122,4 +128,61 @@ public void testEmptyDirectory()
...
+      String badEntryName =  "../foo/bar";
+      TarArchiveEntry tarEntry = new TarArchiveEntry(nonDirFile, badEntryName);
+      tOut.putArchiveEntry(tarEntry);
+      IOUtils.copy(new FileInputStream(nonDirFile), tOut);
+      tOut.closeArchiveEntry();
+
+    }  catch (IOException e) {
+      Assert.fail("Unexpected Exception!!");
+    } finally {
+      if (tOut != null) {
```
kishoreg: use the try with resources to eliminate all the null checks and boiler plate code. https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
```diff
@@ -122,4 +128,61 @@ public void testEmptyDirectory()
     Assert.assertEquals(segmentFiles.length, 0);
 
   }
+
+  @Test
+  public void testBadFilePath() throws Exception {
```
kishoreg: are we cleaning the files created in tearDown

pinot-common/src/main/java/com/linkedin/pinot/common/utils/TarGzCompressionUtils.java
jackjlli: Sure, I can clean this up.

pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
jackjlli: We have a `teardownTest` method that deletes TEST_DIR, which contains all the test files. It will be called when each unit test is finished.

PR #2838, opened 2018-05-22
Fix realtime table schema name reference

subprotocol:
https://github.com/linkedin/pinot/pull/2777 introduced a small issue when creating realtime tables whose schema name does not match the table name.

The integration unit tests didn't catch this because as luck would have it schemaName = tableName for those tests.  This PR resolves the issue and updates the schema names in those unit tests.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -1087,7 +1087,7 @@ public class PinotHelixResourceManager {
         break;
       case REALTIME:
         // Ensure that realtime table is not created for the realtime table
-        Schema schema = ZKMetadataProvider.getTableSchema(_propertyStore, tableNameWithType);
+        Schema schema = ZKMetadataProvider.getTableSchema(_propertyStore, segmentsConfig.getSchemaName());
         if (schema == null) {
           throw new InvalidTableConfigException("No schema defined for realtime table: " + tableNameWithType);
         }
--- a/pinot-integration-tests/src/test/resources/On_Time_On_Time_Performance_2014_100k_subset_nonulls.schema
+++ b/pinot-integration-tests/src/test/resources/On_Time_On_Time_Performance_2014_100k_subset_nonulls.schema
@@ -1,5 +1,5 @@
 {
-  "schemaName": "mytable",
+  "schemaName": "myschema",
   "dimensionFieldSpecs": [
     {
       "name": "AirlineID",
--- a/pinot-integration-tests/src/test/resources/On_Time_On_Time_Performance_2014_100k_subset_nonulls_default_column_test_extra_columns.schema
+++ b/pinot-integration-tests/src/test/resources/On_Time_On_Time_Performance_2014_100k_subset_nonulls_default_column_test_extra_columns.schema
@@ -1,5 +1,5 @@
 {
-  "schemaName": "mytable",
+  "schemaName": "myschema",
   "dimensionFieldSpecs": [
     {
       "name": "AirlineID",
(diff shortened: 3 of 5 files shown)


kishoreg:
LGTM. @Jackie-Jiang, do we violate this constraint in any of the existing use cases?

Jackie-Jiang:
@subprotocol ZKMetadataProvider.getTableSchema() handles the situation where schema name is not the same as table name (schema name should always be the same as raw table name, but we allow that for backward-compatibility).

subprotocol:
@Jackie-Jiang Thank you for the clarification.  I ran into case 3 and it wasn't clear to me what the issue was from the error that was thrown, as I was defining an existing schema.  So if I follow correctly defining schemaName inside segmentsConfig JSON should not be used, or should always = the table name correct?


1. Valid because schemaName = tableName:
```
{
  "tableName": "events",
  "tableType": "REALTIME",
  "segmentsConfig": {
    "timeColumnName": "time",
    "segmentAssignmentStrategy": "BalanceNumSegmentAssignmentStrategy",
    "schemaName": "events",
	...
  },
  ...
}
```

2. Valid because schemaName inferred from tableName:
```
{
  "tableName": "events",
  "tableType": "REALTIME",
  "segmentsConfig": {
    "timeColumnName": "time",
    "segmentAssignmentStrategy": "BalanceNumSegmentAssignmentStrategy",
	...
  },
  ...
}
```

3. Invalid because tableName (events) != schemaName (events_schema).
```
{
  "tableName": "events",
  "tableType": "REALTIME",
  "segmentsConfig": {
    "timeColumnName": "time",
    "segmentAssignmentStrategy": "BalanceNumSegmentAssignmentStrategy",
    "schemaName": "events_schema",
	...
  },
  ...
}
```

The following exception is thrown when they do not match: `Invalid table config for table: events_REALTIME exception: No schema defined for realtime table: events_REALTIME`

sunithabeeram:
@subprotocol thanks for the details. Are you sure "events_schema" is actually available? 

I added a test case to cover the case you are mentioning and it seems to be working. Let me know if you see #2840 doing something differently.

Jackie-Jiang:
@subprotocol You can check the code inside ZKMetadataProvider.getTableSchema() for the actual logic. For case 3, have you already uploaded a schema named "events" or "events_schema"? It will first try to use table name "events", then schema name "events_schema" to search for the schema.

subprotocol:
@sunithabeeram thank you!  I am not certain why the issue doesn't trigger in your test case.  I just threw together an example app that demonstrates the issue I ran into in case it is helpful:  https://github.com/subprotocol/pinot-schema-test

It creates a single schema, and tries to create two tables from it (one with the same name as the schema, and the other with a table name different than the schema).  Feel free to run it, but I also uploaded the controller logs so you can see the error (https://github.com/subprotocol/pinot-schema-test/blob/master/pinotController.log#L864)

@Jackie-Jiang Thanks I will double check what is going on in there.  In the example app I uploaded it seems ZKMetadataProvider.getTableSchema() is returning null for me in case 3.

image

sunithabeeram:
I see an error in the test that I added in #2840. Will fix it and check the behavior.

subprotocol:
It looks like in case 3 ZKMetadataProvider.getTableSchema is trying to get the schema for the table name 'events2' out of the property store, but the ZNode for the schema is 'events'.  

image


zkCli-
```
[zk: localhost:2181(CONNECTED) 6] ls /test-cluster/PROPERTYSTORE/SCHEMAS
[events]
```

sunithabeeram:
@subprotocol I fixed a sequencing issue in #2841. Thanks for reporting the issue - let us know if the fix works for you.

subprotocol:
@sunithabeeram Awesome!  Yes I believe that will solve it and make that method safe to use during table creation.  I will verify

subprotocol:
@sunithabeeram I confirmed that it does fix my issue,  it works great now.  Thank you!

PR #2841, opened 2018-05-23
Fix RT table creation schema check: the backwards compatibility path was broken due to a sequencing issue.

sunithabeeram:
The code for RT table creation checked for schema before the table config was available on ZK. This broke the path where if we can't find the schema that matches the table name, the fallback that fetched name from validation-config returned nothing.

Added test case to cover this as well.


subprotocol:
LGTM thanks again 👍

PR #2844, opened 2018-05-24
Clean up TarGZCompressionUtils and its test

jackjlli:
This PR cleans up the utils class and its unit test per request https://github.com/linkedin/pinot/pull/2837.

### DIFF
--- a/pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
+++ b/pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
@@ -33,8 +33,6 @@ import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 
-
-
 public class TarGzCompressionUtilsTest {
   private static final String SEGMENT_NAME = "mysegment";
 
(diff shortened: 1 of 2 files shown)


pinot-common/src/test/java/com/linkedin/pinot/common/utils/TarGzCompressionUtilsTest.java
```diff
@@ -138,26 +136,24 @@ public void testBadFilePath() throws Exception {
...
+    createInvalidTarFile(metaFile, tarGzPath);
 
     try {
       TarGzCompressionUtils.unTar(tarGzPath, untarDir);
       Assert.fail("Did not get exception!!");
-    } catch (IOException e) {
-      Assert.assertTrue(e.getMessage().contains(TarGzCompressionUtils.EXTRACT_FILE_OUTSIDE_OF_TARGET_DIR));
+    } catch (Exception e) {
+      Assert.assertTrue(e instanceof IOException);
+      Assert.assertTrue(e.getMessage().startsWith("Tar file must not"));
```
mcvsubbu: thanks for adding this string check.  As there are so many exceptions possible, it is best to check that we are getting the exception that we expect, from the place we expect it.
An alternative is to use a specific exception (in this case, a new pinot exception perhaps), but I am fine with this.

PR #2849, opened 2018-05-25
Fix the issue where start/end time defined in SegmentGeneratorConfig is not picked up by IndexCreator

Jackie-Jiang:
When generating segment, if start/end time is defined in SegmentGeneratorConfig, use the one in the config
Move the SIMPLE_DATE time value conversion logic into IndexCreator from SegmentIndexCreationDriver
New logic is covered by SegmentGenerationWithTimeColumnTest


sunithabeeram:
How are we catching these issues? Can we add appropriate unit tests with the fix?

PR #2850, opened 2018-05-25
Add log to track field name when exception thrown; Hot fix for calculating buffer length

jackjlli:
This PR adds log to track the field name when exception happens. 
Currently if some exceptions like `IndexOutOfBoundsException` are thrown, we don't know which field it is, which makes us hard to debug.
This PR also hot fixes the calculation of buffer length. Basically if two factors in the multiplication are of type integer, then the result is of type integer as well. Casting one of the factor to long will fix the problem.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -139,9 +139,9 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
         try {
           dictionaryCreator.build();
         } catch (Exception e) {
-          LOGGER.error("Error building dictionary for field: {}, cardinality: {}, max length in bytes: {}",
+          LOGGER.error("Error building dictionary for field: {}, cardinality: {}, number of bytes per entry: {}",
               fieldSpec.getName(), indexCreationInfo.getDistinctValueCount(), dictionaryCreator.getNumBytesPerEntry());
-          Utils.rethrowException(e);
+          throw e;
         }
 
         // Initialize forward index creator
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -16,6 +16,7 @@
 package com.linkedin.pinot.core.segment.creator.impl;
 
 import com.google.common.base.Preconditions;
+import com.linkedin.pinot.common.Utils;
 import com.linkedin.pinot.common.config.ColumnPartitionConfig;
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -133,7 +137,12 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
         _dictionaryCreatorMap.put(columnName, dictionaryCreator);
 
         // Create dictionary
-        dictionaryCreator.build();
+        try {
+          dictionaryCreator.build();
+        } catch (IOException e) {
```
sunithabeeram: The PR description says you want to track IndexOutOfBoundsException whereas you are catching IOE and these are thrown from PinotDataBuffer in the build path. You will want to catch Exception in general and rethrow (as currently done).

sunithabeeram:
LGTM. 

I am not sure about the usage of rethrowException in general. Its not clear to my why we can't use throw(e) again after catching it. Will check with @jfim about it.

The changes also included a few formatting only ones - we should keep formatting changes separate from actual code changes. If not already set, we should have the formatter only reformat updated lines - I think you can specify this in intellij as well.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -133,7 +136,13 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
         _dictionaryCreatorMap.put(columnName, dictionaryCreator);
 
         // Create dictionary
-        dictionaryCreator.build();
+        try {
+          dictionaryCreator.build();
+        } catch (Exception e) {
+          LOGGER.error("Error building dictionary for field: {}, cardinality: {}, max length in bytes: {}",
+              fieldSpec.getName(), indexCreationInfo.getDistinctValueCount(), dictionaryCreator.getNumBytesPerEntry());
+          Utils.rethrowException(e);
```
Jackie-Jiang: Change this to "throw e;"

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -133,7 +136,13 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
         _dictionaryCreatorMap.put(columnName, dictionaryCreator);
 
         // Create dictionary
-        dictionaryCreator.build();
+        try {
+          dictionaryCreator.build();
+        } catch (Exception e) {
+          LOGGER.error("Error building dictionary for field: {}, cardinality: {}, max length in bytes: {}",
```
Jackie-Jiang: Prefer "number bytes per entry" over "max length in bytes"

PR #2851, opened 2018-05-25
Config to move completed realtime segments

mcvsubbu:
Add tagOverrideConfig to define tag overrides for servers.
Add javadoc to tagOverrideConfig class Use tagOverrideConfig inside RealtimeTagConfig to return the right consuming and completed tags.
Rename TenantNameBuilder to TagNameBuilder, because that's what it really is. 
Iterate over all places using TagNameBuilder and use tag overrides where relevant


pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -176,7 +176,7 @@ private void addInstanceTagIfNeeded(String clusterName, String instanceName) {
     if (instanceTags == null || instanceTags.isEmpty()) {
       if (ZKMetadataProvider.getClusterTenantIsolationEnabled(_propertyStore)) {
         _helixAdmin.addInstanceTag(clusterName, instanceName,
-            TenantNameBuilder.getBrokerTenantNameForTenant(TenantNameBuilder.DEFAULT_TENANT_NAME));
+            TagNameBuilder.getBrokerTagForTenant(TagNameBuilder.DEFAULT_TENANT_NAME));
```
kishoreg: Builder having a get method?

pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -176,7 +176,7 @@ private void addInstanceTagIfNeeded(String clusterName, String instanceName) {
     if (instanceTags == null || instanceTags.isEmpty()) {
       if (ZKMetadataProvider.getClusterTenantIsolationEnabled(_propertyStore)) {
```
kishoreg: unrelated to this PR? what is getClusterTenantIsolationEnabled? We should not have utility methods like this in ZkMetadataProvider

pinot-common/src/main/java/com/linkedin/pinot/common/config/RealtimeTagConfig.java
```diff
@@ -31,9 +29,17 @@
   public RealtimeTagConfig(TableConfig tableConfig) {
     super(tableConfig);
 
-    // TODO: after we introduce config to override tags, pick the right ones from config for consuming and completed
-    _consumingRealtimeServerTag = TenantNameBuilder.getRealtimeTenantNameForTenant(_serverTenant);
-    _completedRealtimeServerTag = TenantNameBuilder.getRealtimeTenantNameForTenant(_serverTenant);
+    TagOverrideConfig tagOverrideConfig = tableConfig.getTenantConfig().getTagOverrideConfig();
+    if (tagOverrideConfig != null && tagOverrideConfig.getRealtimeConsuming() != null) {
```
kishoreg: shouldn't this be something like get(value, defaultValue) which returns defaultValue if the value is null?

pinot-common/src/main/java/com/linkedin/pinot/common/config/TagNameBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.config;
+
+import com.linkedin.pinot.common.utils.ServerType;
+import com.linkedin.pinot.common.utils.TenantRole;
+
+
+public class TagNameBuilder {
```
kishoreg: Is this supposed to be a builder? it has all kinds of methods get, build, has. It resembles a Util class

pinot-common/src/main/java/com/linkedin/pinot/common/config/TagNameBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+  public final static String DEFAULT_TENANT_NAME = "DefaultTenant";
```
kishoreg: i thought default tenant name was just "default", how is this different?

pinot-common/src/test/java/com/linkedin/pinot/common/config/TagConfigTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.json.JSONException;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+
+public class TagConfigTest {
```
kishoreg: TagOverideConfigTest?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/ControllerRequestBuilderUtil.java
```diff
@@ -59,8 +59,8 @@ public static void addFakeBrokerInstancesToAutoJoinHelixCluster(String helixClus
       helixZkManager.connect();
       if (isSingleTenant) {
         helixZkManager.getClusterManagmentTool()
-            .addInstanceTag(helixClusterName, brokerId, TenantNameBuilder.getBrokerTenantNameForTenant(
-                TenantNameBuilder.DEFAULT_TENANT_NAME));
+            .addInstanceTag(helixClusterName, brokerId,
+                TagNameBuilder.getBrokerTagForTenant(TagNameBuilder.DEFAULT_TENANT_NAME));
```
kishoreg: why did we move the constant from tenantNameBuilder to tagNameBuilder.

pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/HelixBrokerStarter.java
mcvsubbu: PROPERTYSTORE/CONFIGS/CLUSTER/tenantIsolationEnabled is a boolean in the cluster config. This is a helper method to get the value of that boolean. As are other methods in the same class to get stuff from propertystore.

pinot-common/src/main/java/com/linkedin/pinot/common/config/RealtimeTagConfig.java
mcvsubbu: The tag override config is the config holder. This class is the one that converts it to how we want it to behave if the config is not present. I would prefer that config classes return what exactly is in the config.

pinot-common/src/main/java/com/linkedin/pinot/common/config/TagNameBuilder.java
mcvsubbu: Agreed, Builder is not quite the right name, given there is a pattern around that name.

TagNameUtils?

pinot-common/src/test/java/com/linkedin/pinot/common/config/TagConfigTest.java
mcvsubbu: Sure

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/ControllerRequestBuilderUtil.java
mcvsubbu: The comment in the commit indicates why. It is not really a Tenant name builder, it is a Tag name builder. So the class was changed. @npawar forgot to delete the old class, I will delete it in the next commit to this branch.

PR #2857, opened 2018-05-30
Use PASS_THROUGH compressor for Metric columns, by default.

mayankshriv:
1. Unless explicitly specified, no-dictionary Metric columns will now use
   PASS_THROUGH compressor. This is because after applying predicates
   rows of metric columns are more likely to be spread across multiple
   chunk buffers, and in the worst cases would require one chunk
   decompression per row selected.

2. Bug fix in PASS_THROUGH decompressor where the output buffer was not
   flipped to be ready to read.

3. Enhanced unit tests to cover the new functionality.


pinot-core/src/main/java/com/linkedin/pinot/core/io/compression/PassThroughDecompressor.java
```diff
@@ -26,9 +25,11 @@
  */
 public class PassThroughDecompressor implements ChunkDecompressor {
   @Override
-  public int decompress(ByteBuffer inCompressed, ByteBuffer outDecompressed)
-      throws IOException {
+  public int decompress(ByteBuffer inCompressed, ByteBuffer outDecompressed) {
```
kishoreg: rename variables -> compressedInput and decompressedOutput.

is this a bug fix?

pinot-core/src/main/java/com/linkedin/pinot/core/io/compression/PassThroughDecompressor.java
```diff
@@ -26,9 +25,11 @@
...
     outDecompressed.put(inCompressed);
-    return inCompressed.limit();
+
+    // Flip the output ByteBuffer for reading.
```
kishoreg: any test case that would fail without this fix?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -211,7 +237,7 @@ private boolean createDictionaryForColumn(ColumnIndexCreationInfo info, SegmentG
       FieldSpec spec) {
     String column = spec.getName();
 
-    if (config.getRawIndexCreationColumns().contains(column)) {
+    if (config.getRawIndexCreationColumns().contains(column) || config.getRawIndexCompressionType().containsKey(column)) {
```
snleee: Is there any case where we have columns not specified in `RawIndexCreationColumns` but specified in`RawIndexCompressionType`? I think that it makes more sense to explicitly check `RawIndexCreationColumns` only.

pinot-core/src/main/java/com/linkedin/pinot/core/io/compression/PassThroughDecompressor.java
mayankshriv: Yes, this is a bug fix. The bug would have only impacted no-dictionary column with variable width and pass-through compression (we don't have any live use cases for that).

pinot-core/src/main/java/com/linkedin/pinot/core/io/compression/PassThroughDecompressor.java
mayankshriv: Yes, the testcase modified in the PR will fail if we remove the fix.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
mayankshriv: Ideally, we should only have one config, but seems like  the RawIndexCompressionTypeMap was introduced a while back. And we cannot simply merge the two due to backward compatibility. So for now, I made the code relaxed enough to handle both.

PR #2861, opened 2018-05-30
[TE] frontend - aaronucsd/added auth to dashboard

aaronucsd:
1. Made dashboard the home page
2. Updated fix to login's route willTransition to clean out old `/rca` path
3. Clean up router.js


aaronucsd:
fixed linting per Travis log.

PR #2867, opened 2018-06-01
[bug] : deleting temp files before loading the segment

RahulGuptaIIITA:
deleting temp files before loading the segment

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/SegmentPreProcessor.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/SegmentPreProcessor.java
@@ -17,6 +17,7 @@ package com.linkedin.pinot.core.segment.index.loader;
 
 import com.linkedin.pinot.common.data.Schema;
 import com.linkedin.pinot.common.segment.ReadMode;
+import com.linkedin.pinot.core.segment.creator.impl.V1Constants;
 import com.linkedin.pinot.core.segment.index.SegmentMetadataImpl;
 import com.linkedin.pinot.core.segment.index.loader.columnminmaxvalue.ColumnMinMaxValueGenerator;
 import com.linkedin.pinot.core.segment.index.loader.columnminmaxvalue.ColumnMinMaxValueGeneratorMode;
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/SegmentPreProcessor.java
```diff
@@ -62,6 +63,16 @@ public void process() throws Exception {
       return;
     }
 
+    File[] directoryListing = _indexDir.listFiles();
+    String tempFileExtension = V1Constants.Indexes.BITMAP_INVERTED_INDEX_FILE_EXTENSION + ".tmp";
```
mcvsubbu: I suppose this is coming from OffheapBitmapInvertedIndexCreator, and we are protecting against the server getting killed while the temp files are in place?
Would it make sense to make a call to OffheapBitmapInvertedIndexCreator to clear its tmp files?  Or, would it make sense to simply delete *.tmp (assuming every other class names its droppings with that extension)?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/SegmentPreProcessor.java
```diff
@@ -62,6 +63,16 @@ public void process() throws Exception {
...
+    if (directoryListing != null) {
+      for (File child : directoryListing) {
+        if (child.getName().endsWith(tempFileExtension)) {
+          org.apache.commons.io.FileUtils.deleteQuietly(child);
```
Jackie-Jiang: You can import this class instead of putting the full path (seems no class name collision)

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/loader/SegmentPreProcessor.java
```diff
@@ -62,6 +63,16 @@ public void process() throws Exception {
       return;
     }
 
+    File[] directoryListing = _indexDir.listFiles();
```
Jackie-Jiang: Please add comments here to explain why we are doing this. In the future, if other people look at this piece of code, they won't get confused.

PR #2874, opened 2018-06-05
Auto reload segments when schema or table config updated

jackjlli:
This PR adds code to enable auto reload segments when schema/table config is updated.
When schema is changed, since schema name and raw table name should be the same, we fetch existing table names by schema name.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
@@ -124,8 +124,8 @@ public class PinotSchemaRestletResource {
   public SuccessResponse updateSchema(
       @ApiParam(value = "Name of the schema", required = true) @PathParam("schemaName") String schemaName,
       FormDataMultiPart multiPart,
-      @ApiParam(value = "Whether to force to reload segments") @DefaultValue("false") @QueryParam("forceReload") Boolean forceReload) {
-    return addOrUpdateSchema(schemaName, multiPart, forceReload);
+      @ApiParam(value = "Whether to reload segments right after schema updated") @DefaultValue("false") @QueryParam("reload") boolean reload) {
+    return addOrUpdateSchema(schemaName, multiPart, reload);
   }
 
   // TODO: This should not update if the schema already exists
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -115,8 +123,9 @@ public SuccessResponse deleteSchema(
   @ApiResponses(value = {@ApiResponse(code = 200, message = "Successfully updated schema"), @ApiResponse(code = 404, message = "Schema not found"), @ApiResponse(code = 400, message = "Missing or invalid request body"), @ApiResponse(code = 500, message = "Internal error")})
   public SuccessResponse updateSchema(
       @ApiParam(value = "Name of the schema", required = true) @PathParam("schemaName") String schemaName,
-      FormDataMultiPart multiPart) {
-    return addOrUpdateSchema(schemaName, multiPart);
+      FormDataMultiPart multiPart,
+      @ApiParam(value = "Whether to force to reload segments") @DefaultValue("false") @QueryParam("forceReload") Boolean forceReload) {
```
Jackie-Jiang: Can we make this parameter **boolean** instead **Boolean**? Also I would recommend name it **"reload"**.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -178,14 +187,54 @@ private SuccessResponse addOrUpdateSchema(@Nullable String schemaName, FormDataM
...
+    }
+    successMessage.append(". Sent ").append(numReloadMessagesSent).append(" reload messages");
+  }
+
+  private List<String> getExistingTables(String schemaName) {
+    List<String> tableNamesWithType = new ArrayList<>(2);
+    ZkHelixPropertyStore<ZNRecord> propertyStore = _pinotHelixResourceManager.getPropertyStore();
+
+    // for real-time table
+    String realTimeTableName = TableNameBuilder.forType(CommonConstants.Helix.TableType.REALTIME).tableNameWithType(schemaName);
```
Jackie-Jiang: TableNameBuilder.REALTIME.tableNameWithType(schemaName);

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -178,14 +187,54 @@ private SuccessResponse addOrUpdateSchema(@Nullable String schemaName, FormDataM
...
+    TableConfig tableConfig = ZKMetadataProvider.getTableConfig(propertyStore, realTimeTableName);
+    if (tableConfig != null) {
+      tableNamesWithType.add(realTimeTableName);
+    }
+
+    // for offline table
+    String offlineTableName = TableNameBuilder.forType(CommonConstants.Helix.TableType.REALTIME).tableNameWithType(schemaName);
```
Jackie-Jiang: (CRITICAL) **OFFLINE**

jackjlli:
@Jackie-Jiang It's called `auto-reload` after schema/table config updated. 
One less API call if we set reload param to true. So you can say an easy way to reload. ;)
By default it's set to false in case of backward compatibility.

npawar:

PR #2875, opened 2018-06-05
Addressed comments from design review for segment size based row threshold updater (and more).

mcvsubbu:
Review comment: Added logic to set a minimum of 10 k rows if the target number of rows computes to be too small.
This avoids corner cases where incoming stream has issues and we consume too small number of rows.

Changed updateFlushThreshold method to be synchronized since multiple partitions could be calling invoking the same method
in the same instance of the object, and we need to serialize them in order for the ratio to be correct.

Changed the threshold updater to accept input from partitions other than patition 0 if there the current
ratio is 0. This helps us jump start all consuming partitions sooner during table start up, or controller
switch of mastership.

Added logic to start a fresh threshold updater when a table is created. If a table is removed and recreated
(e.g. during experimentation), we get the old threshold values (not desirable)

Updated log messages to contain the correct value of target num rows.

Updated log messages to contain the new segment name for which target num rows is being computed.

Added test cases

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
@@ -47,6 +47,7 @@ public class SegmentSizeBasedFlushThresholdUpdater implements FlushThresholdUpda
   private static final double CURRENT_SEGMENT_RATIO_WEIGHT = 0.1;
   private static final double PREVIOUS_SEGMENT_RATIO_WEIGHT = 0.9;
   private static final double ROWS_MULTIPLIER_WHEN_TIME_THRESHOLD_HIT = 1.1;
+  private static final int MINIMUM_NUM_ROWS_THRESHOLD = 10_000;
 
   @VisibleForTesting
   long getIdealSegmentSizeBytes() {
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
```diff
@@ -153,6 +169,6 @@ private long capNumRowsIfOverflow(long targetSegmentNumRows) {
     if (targetSegmentNumRows > Integer.MAX_VALUE) {
       targetSegmentNumRows = Integer.MAX_VALUE;
```
mayankshriv: Could this potentially cause segment generation failure because the size of index could become > 2GB? Should we ceil the value to be a much smaller one?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
```diff
@@ -63,47 +64,67 @@ int getInitialRowsThreshold() {
...
     int numRowsThreshold = committingSegmentZKMetadata.getSizeThresholdToFlushSegment();
-    LOGGER.info("Time consumed:{}  Num rows consumed:{} Num rows threshold:{} Committing segment size bytes:{}",
-        TimeUtils.convertMillisToPeriod(timeConsumed), numRowsConsumed, numRowsThreshold, committingSegmentSizeBytes);
+    LOGGER.info("{}: Data from committing segment: Time {}  numRows {} threshold {} segmentSize(bytes) {}",
+        newSegmentName, TimeUtils.convertMillisToPeriod(timeConsumed), numRowsConsumed, numRowsThreshold,
+        committingSegmentSizeBytes);
 
     double currentRatio = (double) numRowsConsumed / committingSegmentSizeBytes;
-    if (new LLCSegmentName(committingSegmentZKMetadata.getSegmentName()).getPartitionId() == 0) {
+    // Compute segment size to rows ratio only from partition 0.
```
mayankshriv: What if partition 0 has an issue? Should we explicitly pick a working partition, or a random partition to minimize probability of reading a partition with issue?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
```diff
@@ -63,47 +64,67 @@ int getInitialRowsThreshold() {
...
+    // If we consider all partitions then it is likely that we will assign a much higher weight to the most
```
mayankshriv: Would taking average of all partitions avoid the skewed weight issue?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
mcvsubbu: We pick partition 0 for the reason that it always exists in every topic of the stream. If partition 0 has an issue consuming, we will have alerts that should lead to fixing the issue. The worst that will happen when partition 0 is not consuming is that the learning part will stop at what was learned before, so the ratio will not change. Not a catastrophe (compared to partition 0 not consuming at all)

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
mcvsubbu: No, but why complicate the logic taking averages of all partitions?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/segment/SegmentSizeBasedFlushThresholdUpdater.java
mcvsubbu: This is not a part of this PR, but I agree we can choose a smaller number. Not sure how to pick a number, though. I will add a TODO for now.

PR #2876, opened 2018-06-05
Limit updates to segment stats history so as to get a better learning curve

mcvsubbu:
The segment stats history is a circular buffer of last N segments characteristics. We average values from the
previous segments to provide a value for the new one. Since all partitions within a topic tend to have
similar characteristics within the same period of time, it is best that we take only one partition from
each time period for computing statistics. Then, average over time will actually capture trends rather than
being thrown off by one set of segments during a time period. (as can happen with samza jobs, or daily
variations, for example).

We attempt to get to "one update per bunch of completing partitions" by introducing a min time interval
between consecutive additions to the stats history. This will lead us to take stats from the first partition
that completes, and then hold off taking stats from other partitions that will complete around the same
time.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
@@ -57,6 +57,8 @@ public class RealtimeSegmentStatsHistory implements Serializable {
   // Not to be serialized
   transient int _arraySize;
   transient File _historyFile;
+  transient long _minIntervalBetweenUpdatesMillis = 0;
+  transient long _lastUpdateTimeMillis = 0;
 
   @VisibleForTesting
   public static int getMaxNumEntries() {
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/RealtimeSegmentStatsHistory.java
```diff
@@ -233,9 +235,21 @@ public synchronized boolean isEmpty() {
     return getNumntriesToScan() == 0;
   }
 
+  public void setMinIntervalBetweenUpdatesMillis(long millis) {
+    _minIntervalBetweenUpdatesMillis = millis;
+  }
+
   public synchronized void addSegmentStats(SegmentStats segmentStats) {
```
sunithabeeram: Many methods in this class are 'synchronized' - I am not sure if all private members need careful access; the newly added setMinIntervalBetweenUpdatesMillis is updating a long value without a lock. Is that safe? 

It might be good if we have some class level comments on thread-safety related aspects of the class so future modifiers have the right context.

sunithabeeram: Since the new method is public, there is no guarantee that other parts of the code won't call it any other time. Should the interval be really decided outside this class? If we move it inside this class, then we risk fewer issues with setters getting the way. This is mainly from a pure correctness point of view.

sunithabeeram: Also, I noticed that MutableSemgentImpl is the one that calls addSegmentStats followed by a save(). Should save also not happen if we haven't updated due to the interval check? I am wondering if the interface can be simplified to have addSegmentStats call save internally. That way its well contained and we don't have to rely on users calling or not calling save().

PR #2886, opened 2018-06-13
Re-implement OffHeapBitmapInvertedIndexCreator to support large number of values (> 500M)

Jackie-Jiang:
1. Use LBuffer or MMapBuffer instead of ByteBuffer to allocate buffer larger than 2G
2. Based on the number of values we need to store, choose LBuffer or MMapBuffer
3. Re-use buffer if possible to reduce memory footprint
4. Do not generate extra temporary file during index generation

NOTE: the temporary buffers used are not tracked by MMapUtils. After adding trackable big size buffer, will replace the LBufferAPI with the new one

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/InvertedIndexCreator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/InvertedIndexCreator.java
@@ -36,38 +36,35 @@ import java.io.IOException;
  * |                           Data for last bitmap                          |
  * |-------------------------------------------------------------------------|
  * </pre>
+ *
+ * <p>To create an inverted index:
+ * <ul>
+ *   <li>
+ *     Construct an instance of <code>InvertedIndexCreator</code>
+ *   </li>
+ *   <li>
+ *     Call add() for each docId in sequence starting with 0 to add dictId (dictIds for multi-valued column) into the
+ *     creator
+ *   </li>
+ *   <li>
+ *     Call seal() after all dictIds have been added
+ *   </li>
+ * </ul>
  */
 public interface InvertedIndexCreator extends Closeable {
 
   /**
-   * Add an entry for single-value column.
-   *
-   * @param docId Document id
-   * @param dictId Dictionary id
+   * For single-valued column, adds the dictionary Id for the next document.
    */
-  void addSV(int docId, int dictId);
+  void add(int dictId);
 
   /**
-   * Add an entry for multi-value column.
-   *
-   * @param docId Document id
-   * @param dictIds Array of dictionary ids
+   * For multi-valued column, adds the dictionary Ids for the next document.
    */
-  void addMV(int docId, int[] dictIds);
+  void add(int[] dictIds, int length);
 
   /**
-   * Add an entry for multi-value column.
-   *
-   * @param docId Document id
-   * @param dictIds Array of dictionary ids
-   * @param numDictIds Number of dictionary ids
-   */
-  void addMV(int docId, int[] dictIds, int numDictIds);
-
-  /**
-   * Seal the results into the file.
-   *
-   * @throws IOException
+   * Seals the index and flushes it to disk.
    */
   void seal() throws IOException;
 }
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
@@ -17,239 +17,232 @@ package com.linkedin.pinot.core.segment.creator.impl.inv;
 
 import com.google.common.base.Preconditions;
 import com.linkedin.pinot.common.data.FieldSpec;
-import com.linkedin.pinot.common.utils.MmapUtils;
 import com.linkedin.pinot.core.segment.creator.InvertedIndexCreator;
 import com.linkedin.pinot.core.segment.creator.impl.V1Constants;
 import java.io.BufferedOutputStream;
 import java.io.DataOutputStream;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.nio.IntBuffer;
-import java.nio.channels.FileChannel;
 import org.apache.commons.io.FileUtils;
 import org.roaringbitmap.buffer.MutableRoaringBitmap;
+import xerial.larray.buffer.LBuffer;
+import xerial.larray.buffer.LBufferAPI;
+import xerial.larray.mmap.MMapBuffer;
+import xerial.larray.mmap.MMapMode;
 
 
 /**
  * Implementation of {@link InvertedIndexCreator} that uses off-heap memory.
+ * <p>We use 2 passes to create the inverted index.
  * <ul>
- *   High level idea:
  *   <li>
- *     We use 2 passes to create the inverted index.
+ *     In the first pass (adding values phase), when add() method is called, store the dictIds into the forward index
+ *     value buffer (for multi-valued column also store number of values for each docId into forward index length
+ *     buffer). We also compute the inverted index length for each dictId while adding values.
  *   </li>
  *   <li>
- *     In the first pass (adding values phase), when addSV/addMV method is called, store the values (dictIds) into
- *     value buffer (for multi-value column also store number of values for each docId into numValues buffer). We also
- *     compute the posting list length for each value while processing values.
- *   </li>
- *   <li>
- *     We create inverted index in the second pass when seal method is called. By this time, all the values should
- *     already been added. We first construct the posting list by going over the values in value buffer (for multi-value
- *     column we also need numValues buffer to get the docId for each value).
- *     <p>Once we have the posting list for each value (dictId), we simply go over the posting list and create the
- *     bitmap for each value and serialize them into a file.
- *   </li>
- *   <li>
- *     When serializing bitmaps, we first create two files, one for bitmap offsets and one for serialized data. After
- *     serializing all bitmaps, append the serialized data file to the offsets file to get the final inverted index
- *     file.
+ *     In the second pass (processing values phase), when seal() method is called, all the dictIds should already been
+ *     added. We first reorder the values into the inverted index buffers by going over the dictIds in forward index
+ *     value buffer (for multi-valued column we also need forward index length buffer to get the docId for each dictId).
+ *     <p>Once we have the inverted index buffers, we simply go over them and create the bitmap for each dictId and
+ *     serialize them into a file.
  *   </li>
  * </ul>
+ * <p>Based on the number of values we need to store, we use direct memory or MMap file to allocate the buffer.
  */
 public final class OffHeapBitmapInvertedIndexCreator implements InvertedIndexCreator {
-  // Maximum number of values due to 2GB file size limit
-  public static final int MAX_NUM_VALUES = Integer.MAX_VALUE / V1Constants.Numbers.INTEGER_SIZE;
+  // Use MMapBuffer if the buffer size is larger than 100MB
+  private static final int NUM_VALUES_THRESHOLD_FOR_MMAP_BUFFER = 25_000_000;
 
-  private final String _columnName;
+  private static final String FORWARD_INDEX_VALUE_BUFFER_SUFFIX = ".fwd.idx.val.buf";
+  private static final String FORWARD_INDEX_LENGTH_BUFFER_SUFFIX = ".fwd.idx.len.buf";
+  private static final String INVERTED_INDEX_VALUE_BUFFER_SUFFIX = ".inv.idx.val.buf";
+  private static final String INVERTED_INDEX_LENGTH_BUFFER_SUFFIX = ".inv.idx.len.buf";
+
+  private final File _invertedIndexFile;
+  private final File _forwardIndexValueBufferFile;
+  private final File _forwardIndexLengthBufferFile;
+  private final File _invertedIndexValueBufferFile;
+  private final File _invertedIndexLengthBufferFile;
   private final boolean _singleValue;
   private final int _cardinality;
   private final int _numDocs;
   private final int _numValues;
-  private final File _indexDir;
-  private final File _invertedIndexFile;
+  private final boolean _useMMapBuffer;
 
-  private ByteBuffer _valueByteBuffer;
-  private IntBuffer _valueIntBuffer;
+  // TODO: After supporting trackable big size buffer, replace the LBufferAPI with the new buffer
+  // Right now the temporary buffers used are not tracked by MMapUtils
+  // Forward index buffers (from docId to dictId)
+  private int _nextDocId;
+  private LBufferAPI _forwardIndexValueBuffer;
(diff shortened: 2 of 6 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -60,196 +56,151 @@
...
-  // Constructed in first pass
-  private ByteBuffer _postingListLengthByteBuffer;
-  private IntBuffer _postingListLengthIntBuffer;
-  // Constructed in second pass
-  private ByteBuffer _postingListValueByteBuffer;
-  private ByteBuffer _postingListCurrentOffsetByteBuffer;
+  // For single-valued and multi-valued column
+  private int _nextDocId;
+  private LBuffer _valueBuffer;
+  private LBuffer _postingListIndexBuffer;
```
sunithabeeram: Its not clear what "postingList" refers to. Could you add clarifying comments?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -60,196 +56,151 @@
...
+      int valueId = 0;
       for (int docId = 0; docId < _numDocs; docId++) {
-        int numValues = _numValuesIntBuffer.get(docId);
+        int numValues = getInt(_numValuesBuffer, docId);
         for (int i = 0; i < numValues; i++) {
-          int dictId = _valueIntBuffer.get(valueOffset++);
-          int offset = postingListCurrentOffsetIntBuffer.get(dictId);
-          postingListValueIntBuffer.put(offset++, docId);
-          postingListCurrentOffsetIntBuffer.put(dictId, offset);
+          int dictId = getInt(_valueBuffer, valueId++);
```
sunithabeeram: Lines 128-131 and 138-141 look similar. Anychance they can be a single method call?

pinot-core/src/test/java/com/linkedin/pinot/segments/v1/creator/BitmapInvertedIndexCreatorTest.java
```diff
@@ -78,7 +78,7 @@ public void testSingleValue() throws IOException {
     try (OnHeapBitmapInvertedIndexCreator onHeapCreator = new OnHeapBitmapInvertedIndexCreator(ON_HEAP_INDEX_DIR,
         COLUMN_NAME, CARDINALITY)) {
       for (int docId = 0; docId < NUM_DOCS; docId++) {
-        onHeapCreator.addSV(docId, dictIds[docId]);
```
sunithabeeram: Can we add a test to cover the case where we add more than the earlier MAX values?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -60,196 +56,151 @@
...
     try {
-      // Create value buffer
-      _valueByteBuffer = MmapUtils.allocateDirectByteBuffer(_numValues * V1Constants.Numbers.INTEGER_SIZE, null,
-          "bitmap value buffer for: " + _columnName);
-      _valueIntBuffer = _valueByteBuffer.asIntBuffer();
-
-      // Create numValues buffer for multi-value column
+      _valueBuffer = new LBuffer((long) _numValues * Integer.BYTES);
+      _postingListIndexBuffer = new LBuffer((long) _cardinality * Integer.BYTES);
+      _postingListIndexBuffer.clear();
```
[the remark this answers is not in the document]
Jackie-Jiang: Because for _postingListIndexBuffer we rely on the initial value 0, and we keep adding to that value

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -17,38 +17,34 @@
...
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.nio.IntBuffer;
 import java.nio.channels.FileChannel;
 import org.apache.commons.io.FileUtils;
 import org.roaringbitmap.buffer.MutableRoaringBitmap;
+import xerial.larray.buffer.LBuffer;
```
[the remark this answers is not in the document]
Jackie-Jiang: The biggest issue before is endianess. Star-tree has been using this library for a long time, so it should be safe.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
Jackie-Jiang: Will add more comments

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
Jackie-Jiang: the dictId are fetched from different buffer, so I don't think we need a single method call for 3 lines of same code

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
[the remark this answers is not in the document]
kishoreg: LBuffer is bit flaky. Its ok to use it during generation (star-tree index generation uses it). I haven't looked at this PR in detail but are we going to use lbuffer at runtime?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/InvertedIndexCreator.java
```diff
@@ -40,34 +40,17 @@
...
 
   /**
-   * Add an entry for single-value column.
-   *
-   * @param docId Document id
-   * @param dictId Dictionary id
+   * For single-valued column, adds the dictionary Id for the next document.
    */
-  void addSV(int docId, int dictId);
+  void add(int dictId);
```
mcvsubbu: Can you please add a comment that all implementations of this interface must be  stateful, and you can only add items in sequence of docId? thanks

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -60,196 +56,151 @@
...
+    _numValues = _singleValue ? numDocs : numValues;
 
     try {
-      // Create value buffer
-      _valueByteBuffer = MmapUtils.allocateDirectByteBuffer(_numValues * V1Constants.Numbers.INTEGER_SIZE, null,
-          "bitmap value buffer for: " + _columnName);
-      _valueIntBuffer = _valueByteBuffer.asIntBuffer();
-
-      // Create numValues buffer for multi-value column
+      _valueBuffer = new LBuffer((long) _numValues * Integer.BYTES);
```
mcvsubbu: Can you try using larray? It supports mmapping the files. With mmap, we don't have to limit the number of inv index created, or somehow make sure that the jvm is started with the max amount of memory for all inv indices to be created in parallel.

Alternativey (safer, tested) can you create a list of PinotDataBuffer objects? It may be a bit more complex in implementation, but will save a lot in operational cost. Speed is not a concern here, so we should be good.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/InvertedIndexCreator.java
Jackie-Jiang: Added more javaDoc for the interface

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
Jackie-Jiang: Actually we do care about performance on server side, or segment load could take forever. On non-SSD machine, using MMapBuffer could be very slow because we need to write the whole buffer to the disk. We have been using direct memory buffer for inverted index generation, so I'll keep that behavior for now.

Jackie-Jiang: Add a threshold (25M values) so that the creator can choose to use LBuffer or MMapBuffer automatically.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OnHeapBitmapInvertedIndexCreator.java
```diff
@@ -45,35 +44,32 @@ public OnHeapBitmapInvertedIndexCreator(File indexDir, String columnName, int ca
...
+      // Write bitmap offsets
+      int bitmapOffset = (_bitmaps.length + 1) * Integer.BYTES;
+      out.writeInt(bitmapOffset);
       for (MutableRoaringBitmap bitmap : _bitmaps) {
-        offset += bitmap.serializedSizeInBytes();
+        bitmapOffset += bitmap.serializedSizeInBytes();
         // Check for int overflow
-        Preconditions.checkState(offset > 0, "Inverted index file exceeds 2GB limit for column: %s", _columnName);
-        out.writeInt(offset);
+        Preconditions.checkState(bitmapOffset > 0, "Inverted index file: {} exceeds 2GB limit", _invertedIndexFile);
```
sunithabeeram: This check is a bit confusing, esp given that its missing in the off-heap version. It appears that either its applicable in both places or neither. Let me know if I am missing something.

Jackie-Jiang: Seems I accidentally removed it in the offHeap one. Thanks for catching this

sunithabeeram: We should consider moving the part where we write the indices to file (and other common parts) into an abstract class - that way we can have common enforcement for such limits.

Jackie-Jiang: They are written in different ways. For the on-heap one, we already get all bitmaps, but for the off-heap one, we build bitmaps on the fly, so we keep two file descriptors at the same time.It's hard to make them a common method.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/InvertedIndexCreator.java
```diff
@@ -36,38 +36,28 @@
  * |                           Data for last bitmap                          |
  * |-------------------------------------------------------------------------|
  * </pre>
+ *
+ * <p>To create an inverted index:
+ * <ul>
+ *   <li>1. Construct an instance of <code>InvertedIndexCreator</code></li>
+ *   <li>2. Call add() for each docId to add dictId (dictIds for multi-valued column) into the creator</li>
```
mcvsubbu: nit: call add() in sequece starting with docid 0

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -17,239 +17,230 @@
...
+  // Inverted index buffers (from dictId to docId)
+  private LBufferAPI _invertedIndexValueBuffer;
+  private LBufferAPI _invertedIndexLengthBuffer;
 
   public OffHeapBitmapInvertedIndexCreator(File indexDir, FieldSpec fieldSpec, int cardinality, int numDocs,
-      int numValues) {
-    _columnName = fieldSpec.getName();
+      int numValues) throws IOException {
+    String columnName = fieldSpec.getName();
+    _invertedIndexFile = new File(indexDir, columnName + V1Constants.Indexes.BITMAP_INVERTED_INDEX_FILE_EXTENSION);
```
mcvsubbu: I think indexDir is unique, just confiirming that if multiple segments of the same table are being loaded we won't collide on mmap files

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
```diff
@@ -17,239 +17,230 @@
...
+    buffer.putInt(index << 2, value);
+  }
+
+  private static int getInt(LBufferAPI buffer, long index) {
+    return buffer.getInt(index << 2);
+  }
+
+  private LBufferAPI createBuffer(long size, File mmapFile) throws IOException {
+    if (_useMMapBuffer) {
+      return new MMapBuffer(mmapFile, 0L, size, MMapMode.READ_WRITE);
```
mcvsubbu: this is nice, the only problem is that our metrics on memory allocated will now be off. I am not sure how to fix the problem other than using list of pinot data buffers

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/InvertedIndexCreator.java
Jackie-Jiang: Fixed.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
Jackie-Jiang: **indexDir** is per segment based

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/inv/OffHeapBitmapInvertedIndexCreator.java
Jackie-Jiang: Added a TODO to add support to trackable big size buffer

mcvsubbu: i would feel more comfortable if you add a TODO here as well, since these APIs will also change when we make large bufferes trackable? thanks. We need to get that in fairly quickly, since we don't want to be allocating buffers without tracking them via metrics.

PR #2890, opened 2018-06-14
Add a time buffer in retention manager when picking segments which are missing from ideal state for deletion

npawar:
Retention could end up prematurely selecting a segment for deletion because of the absense of any time buffer. If a new segment metadata is created (step 2) and the retention manager kicks in before the ideal state entry for the new segment is created (step 3), then the retention manager would mark the brand new segment for deletion due to absence in ideal state. Adding a time buffer of 5 days, so that we do not pick a segment for deletion unless it is old enough. That gives the segment enough time to complete its lifecycle in committing, or to get repaired by validation manager

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/retention/RetentionManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/retention/RetentionManager.java
@@ -179,17 +179,23 @@ public class RetentionManager {
     if (idealState == null) {
       return false;
     }
+    // delete a segment only if it is old enough (5 days) or else,
+    // 1. latest segment could get deleted in the middle of repair by ValidationManager
+    // 2. for a brand new segment, if this code kicks in after new metadata is created but ideal state entry is not yet created (between step 2 and 3),
+    // the latest segment metadata could get marked for deletion
+    if (System.currentTimeMillis() - realtimeSegmentZKMetadata.getCreationTime()
+        <= OLD_LLC_SEGMENTS_RETENTION_IN_MILLIS) {
+      return false;
+    }
     Map<String, String> stateMap = idealState.getInstanceStateMap(segmentName);
     if (stateMap == null) {
       // Segment is in property store but not in ideal state, delete it
       return true;
     } else {
-      // Delete segment if all of its replicas are OFFLINE and it is old enough
+      // Delete segment if all of its replicas are OFFLINE
       Set<String> states = new HashSet<>(stateMap.values());
       return states.size() == 1 && states.contains(
-          CommonConstants.Helix.StateModel.SegmentOnlineOfflineStateModel.OFFLINE)
-          && System.currentTimeMillis() - realtimeSegmentZKMetadata.getCreationTime()
-          > OLD_LLC_SEGMENTS_RETENTION_IN_MILLIS;
+          CommonConstants.Helix.StateModel.SegmentOnlineOfflineStateModel.OFFLINE);
     }
   }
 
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/retention/RetentionManager.java
```diff
@@ -179,17 +179,23 @@ private boolean shouldDeleteInProgressLLCSegment(String segmentName, IdealState
     if (idealState == null) {
       return false;
     }
+    // delete a segment only if it is old enough (5 days) or else,
+    // 1. latest segment could get deleted in the middle of repair by ValidationManager
+    // 2. for a brand new segment, if this code kicks in after new metadata is created but ideal state entry is not yet created (between step 2 and 3),
+    // the latest segment metadata could get marked for deletion
+    if (System.currentTimeMillis() - realtimeSegmentZKMetadata.getCreationTime()
```
kishoreg: so we will never delete a segment in realtime for 5 days?

npawar: We will never delete an IN_PROGRESS segment in realtime for 5 days. This doesn't apply to completed segments. 
Currently the retention manager deletes a segment IN_PROGRESS if it doesn't find it in the ideal state, with no check for creation time.

kishoreg: my bad, I did not see that this is applied to the segment that's in InProgressLLC state.

mcvsubbu: Just so we are all on the same page:

A segment can have all its replicas in OFFLINE state and metadata as IN_PROGRESS if there is some problem with the stream consumption (say, a hitch in the stream source). In this case, after several retries, the server stops consumption and marks itself OFFLINE. If all replicas have trouble consuming that partition, they will all mark themselves as OFFLINE. If this happens, then the validation manager will triggger a method in PinotLLCRealtimeSegmentManager to fix the partition by creating new segments for the partition that stopped consumption (using the same start offset as in the segment that stopped consumption). Note that it will leave the previous segment in place, for debugging purpose. This logic in retention manager assumes that debugging is not needed for these segments after OLD_LLC_SEGMENT_RETENTION_IN_MILLIS, and removes those idealstate entries (and segment metadata). It uses the same segment delete logic that already handles removal in proper sequence, waiting for external state to be removed, etc. Didn't want to invent new logic to remove segment metadata and idealstate entries here.

mcvsubbu: Since the sequence of segment deletion is to remove segment from Idealstate first, wait for it to go away from externalview, and then remove metadata (and any copies in storage), we initiate the first step in retention manager and move on. The second step may happen much later, depending on Helix performance. A controller could get restarted before the second step happens. In that case, the new controller does not retain any memory of segments to be deleted. So, we also handle the case where the segment metadata is present and idealstate does not have the segment. The metadata, in this case, should be at least several days old (debugging days, as I mention before)

It happens that this condition (of segment metadata being present and idealsgtate not being there) can also exist during a brand new segment creation. It is that window that we avoid by checking time here.

mcvsubbu: @npawar  can we have one  time check up front and all the other checks inside that?

npawar: The time check is only once and we return false if segment is too young to be deleted
All other checks happen only after we come forward from the time check

PR #2891, opened 2018-06-14
Handle a state created due to race condition between RetentionManager and segment commit, resulting in no consuming segments for the partition

npawar:
Segment commit takes place in 3 steps:
1. mark committing segment metadata to DONE
2. create new segment metadata to IN_PROGRESS
3. update ideal state with segment state to ONLINE and new segment state to CONSUMING

The RetentionManager checks for segments which have metadata present but no entry in ideal state. If the RetentionManager happens to check between steps 2 and 3 above, it will mark the newly created segment for deletion. 
The fix for this is to make retrntion manager not pick segments that are new (provided 5 days buffer: https://github.com/linkedin/pinot/pull/2890)
Additionally, the ValidationManager should be able to detect and recover from sucha a scenario

In this PR, a check is put in place to look for all segments ONLINE in ideal state for latest segment, and metadata being DONE. In such a scenario, a new metadata is created IN_PROGRESS and new segment is created in ideal state CONSUMING. We make sure that we don't correct segments that are too soon to correct

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -980,11 +980,6 @@ public class PinotLLCRealtimeSegmentManager {
     return false;
   }
 
-
-  private boolean isAllInstancesInState(Map<String, String> instanceStateMap, String state) {
-    return instanceStateMap.values().stream().allMatch(value -> value.equals(state));
-  }
-
   /*
    * Validate LLC segments of a table.
    *
@@ -1085,7 +1080,7 @@ public class PinotLLCRealtimeSegmentManager {
                     + "Old segment metadata {} has status DONE, but segments are still in CONSUMING state in ideal STATE",
                 tableNameWithType, partition, segmentId);
 
-            LLCSegmentName newLLCSegmentName = getNextLLCSegment(segmentName, partition, now);
+            LLCSegmentName newLLCSegmentName = makeNextLLCSegmentName(segmentName, partition, now);
             LOGGER.info("{}: Creating new segment metadata for {}", tableNameWithType,
                 newLLCSegmentName.getSegmentName());
 
@@ -1098,9 +1093,19 @@ public class PinotLLCRealtimeSegmentManager {
             consumingSegments.add(newLLCSegmentName.getSegmentName());
           }
           // else, the metadata should be IN_PROGRESS, which is the right state for a consuming segment.
-        } else if (isAllInstancesInState(instanceStateMap,
-            PinotHelixSegmentOnlineOfflineStateModelGenerator.OFFLINE_STATE)) {
-          // An in-progress segment marked itself offline for some reason the server could not consume.
+        } else { // no replica in CONSUMING state
+
+          // Possible scenarios: for any of these scenarios, we need to create new metadata IN_PROGRESS and new CONSUMING segment
+          // 1. all replicas OFFLINE and metadata IN_PROGRESS/DONE - a segment marked itself OFFLINE during consumption for some reason
+          // 2. all replicas ONLINE or mixture of ONLINE/OFFLINE and metadata DONE
+          // This could happen due to a race condition with the retention manager where we lost the new segment
+          // The retention manager saw the new segment metadata before it was updated in ideal state, and hence marked it for deletion
+          // later the new segment was updated in the ideal state, but the retention manager would clean it out from the metadata and ideal state, having marked it earlier
+          // So, in commmitSegmentMetadata:
+          // step-1 was done (i.e. marking old segment metadata as DONE)
+          // step-2 was undone by retention manager (i.e. new IN_PROGRESS metadata got deleted)
+          // step 3 was partially undone by retention manager (i.e. marking old segment ONLINE was done but new CONSUMING segment was deleted)
+
           // No instances are consuming, so create a new consuming segment.
           int nextSeqNum = segmentName.getSequenceNumber() + 1;
           LOGGER.info("Creating CONSUMING segment for {} partition {} with seq {}", tableNameWithType, partition,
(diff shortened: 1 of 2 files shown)


kishoreg:
should we revisit the purpose of validateLLCSegments and the scenarios it is trying to handle. Do we know the reason a segment gets into an invalid state? validateLLCSegments has become too complicated.
Another orthogonal question. This bug seems to have existed for a long time, what is the change that surfaced this issue.

npawar:
Yes the bug existed for a long time. No change in particular surfaced the issue. It's just a very rare scenario I think, where the commitSegment steps and retentionManager steps exactly align

mcvsubbu:
The following sequence of events can cause the bug:

1. server A sent a request to complete Segment S1 (realtime) to Controller.
2. Controller marked metadata for S1 as Done
3. Controller created new metadata for S2 and marked IN-PROGRESS
4. Retntion manager came along and looked at IS and decided S2 needs to be deleted since it has metadata but not in idealstate.
5. Controller changed Idealstate to mark S1 ONLINE and S2 CONSUMING (single change)
6. Retention manager deleted S2.

steps 1 thorugh 5 has to have happened in the same ms in that order
or probably a few ms

The correction logic in the controller worked for this case until PR #2721 , when the bug was introduced to NOT handle the case when there are no consuming segments. 

Until then the race condition probably existed, but things auto-corrected, so we did not see it.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1179,6 +1195,13 @@ protected IdealState validateLLCSegments(final TableConfig tableConfig, IdealSta
     return idealState;
   }
 
+  private LLCSegmentName getNextLLCSegment(LLCSegmentName segmentName, int partition, long now) {
```
mcvsubbu: makeNextSegmentName()?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1129,8 +1119,34 @@ protected IdealState validateLLCSegments(final TableConfig tableConfig, IdealSta
               committingSegmentDescriptor);
 
           consumingSegments.add(newLLCSegmentName.getSegmentName());
+        } else if (isAllInstancesInState(instanceStateMap,
+            PinotHelixSegmentOnlineOfflineStateModelGenerator.ONLINE_STATE)) { // else the segment is in ONLINE state
```
mcvsubbu: checking for all states ONLINE will not work all the time (what if one of the realtime hosts is OFFLINE)?

We should be checking for absence of CONSUMING segments

npawar: Combined the else condition to handle absence of CONSUMING in general. The steps to repair are same no matter what we are left with in the absence of CONSUMING segments

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1093,24 +1080,32 @@ protected IdealState validateLLCSegments(final TableConfig tableConfig, IdealSta
...
-          // An in-progress segment marked itself offline for some reason the server could not consume.
+        } else { // no replica in CONSUMING state
+
+          // Possible scenarios: for any of these scenarios, we need to create new metadata IN_PROGRESS and new CONSUMING segment
+          // 1. all replicas OFFLINE and metadata IN_PROGRESS/DONE - a segment marked itself OFFLINE during consumption for some reason
+          // 2. all replicas ONLINE or mixture of ONLINE/OFFLINE and metadata DONE
+          // This could happen due to a race condition with the retention manager where we lost the new segment
+          // The retention manager saw the new segment metadata before it was updated in ideal state, and hence marked it for deletion
+          // later the new segment was updated in the ideal state, but the retention manager would clean it out from the metadata and ideal state, having marked it earlier
+          // So, in commmitSegmentMetadata:
```
mcvsubbu: Is it better to do a check like this before applying the correction?

if (isallInstancesOffline() || !isTooSoonToCorrect(latestMetadata)) {
  // apply correction
}

this will protect us against other possible races that we don't know about yet. I can't think of any right now, but that is not to say that it does not exist.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1093,44 +1084,55 @@ protected IdealState validateLLCSegments(final TableConfig tableConfig, IdealSta
...
-              new CommittingSegmentDescriptor(segmentId, startOffset, 0);
-          createNewSegmentMetadataZNRecord(tableConfig, segmentName, newLLCSegmentName, partitionAssignment,
-              committingSegmentDescriptor);
-
-          consumingSegments.add(newLLCSegmentName.getSegmentName());
+        } else { // no replica in CONSUMING state
+
+          // Possible scenarios: for any of these scenarios, we need to create new metadata IN_PROGRESS and new CONSUMING segment
+          // 1. all replicas OFFLINE and metadata IN_PROGRESS/DONE - a segment marked itself OFFLINE during consumption for some reason
+          // 2. all replicas ONLINE or mixture of ONLINE/OFFLINE and metadata DONE - Resolved in https://github.com/linkedin/pinot/pull/2890
```
mcvsubbu: can we remove these comments about condition 2, and just reference the PR?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1093,44 +1084,55 @@ protected IdealState validateLLCSegments(final TableConfig tableConfig, IdealSta
...
+          // later the new segment was updated in the ideal state, but the retention manager would clean it out from the metadata and ideal state, having marked it earlier
+          // So, in commmitSegmentMetadata:
+          // step-1 was done (i.e. marking old segment metadata as DONE)
+          // step-2 was undone by retention manager (i.e. new IN_PROGRESS metadata got deleted)
+          // step 3 was partially undone by retention manager (i.e. marking old segment ONLINE was done but new CONSUMING segment was deleted)
+
+          if (isAllInstancesInState(instanceStateMap, PinotHelixSegmentOnlineOfflineStateModelGenerator.OFFLINE_STATE)
+              || !(isTooSoonToCorrect(tableNameWithType, segmentId, now))) {
+            // No instances are consuming, so create a new consuming segment.
+            int nextSeqNum = segmentName.getSequenceNumber() + 1;
```
mcvsubbu: can we use the new method on makeNextSegmentName here?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1093,44 +1084,55 @@ protected IdealState validateLLCSegments(final TableConfig tableConfig, IdealSta
...
+          // This could happen due to a race condition with the retention manager where we lost the new segment
+          // The retention manager saw the new segment metadata before it was updated in ideal state, and hence marked it for deletion
+          // later the new segment was updated in the ideal state, but the retention manager would clean it out from the metadata and ideal state, having marked it earlier
+          // So, in commmitSegmentMetadata:
+          // step-1 was done (i.e. marking old segment metadata as DONE)
+          // step-2 was undone by retention manager (i.e. new IN_PROGRESS metadata got deleted)
+          // step 3 was partially undone by retention manager (i.e. marking old segment ONLINE was done but new CONSUMING segment was deleted)
+
+          if (isAllInstancesInState(instanceStateMap, PinotHelixSegmentOnlineOfflineStateModelGenerator.OFFLINE_STATE)
+              || !(isTooSoonToCorrect(tableNameWithType, segmentId, now))) {
```
mcvsubbu: Can you please add a log in isTooSoonToCorrect method (log only if it returns true). It should log the segment name and the times it is comparing. Thanks.

npawar: addressed all the comments @mcvsubbu

PR #2892, opened 2018-06-15
Fix the percentile aggregation functions

Jackie-Jiang:
After the fix, for all percentile aggregation functions (PERCENTILE, PERCENTILEEST, PERCENTILETDIGEST, PERCENTILEMV, PERCENTILEESTMV), support the following percentiles:
5, 10, 20, 25, 30, 40, 50, 60, 70, 75, 80, 90, 95, 99


Jackie-Jiang:
@kishoreg Good point. But that requires more changes. Will submit a separate PR for that.

PR #2893, opened 2018-06-15
[TE] Advanced Dimension Analysis Table (3): static table improvements

newsummit:
Improving on the stability of the static table, responsive to changes in metric, analysis range, graph legend selections. This is after initial feedback from users:
- scalable spacing in change contribution visual column (bar width relative to max values)
- displaying first aggregate row (previously hidden)
- tooltips for excluded dimensions where available
- correction for `overall change` value

Sample data shown in screenshot:
screen shot 2018-06-14 at 5 48 54 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/custom/dimensions-table/dimension/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/custom/dimensions-table/dimension/component.js
@@ -1,6 +1,7 @@
 /**
  * Custom model table component
  * Constructs the dynamic dimension columns for the Advanced Dimension Analysis table
+ * This component will be rendered once for each dynamic dimension column
  * @module custom/dimensions-table/dimension
  *
  * @example for usage in models table columns definitions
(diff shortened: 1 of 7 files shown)


thirdeye/thirdeye-frontend/app/pods/custom/dimensions-table/dimension/component.js
```diff
@@ -15,44 +16,101 @@
...
+   * @type {Array}
+   */
+  classNames: [CUSTOM_CLASS, `${CUSTOM_CLASS}--dynamic`],
+
+
+  init() {
     this._super(...arguments);
-    this._formatDimensionLabel();
+
+    once(() => {
```
aaronucsd: is it running more than once without this?

newsummit: It was running more than once, but not for this reason. I've corrected it by adding refresh logic in `init`.

PR #2894, opened 2018-06-15
Support arbitrary percentile for percentile aggregation functions

Jackie-Jiang:
1. Use regex matching to automatically figure out the percentile
2. Remove redundant aggregation function types
3. Change the AggregationFunctionType.isOfType() so that it won't throw exception when function name is not a valid aggregation function type. When matching function name with specific types, it should not throw any exception

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/MetadataBasedAggregationOperator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/MetadataBasedAggregationOperator.java
@@ -26,6 +26,7 @@ import com.linkedin.pinot.core.query.aggregation.DoubleAggregationResultHolder;
 import com.linkedin.pinot.core.query.aggregation.ObjectAggregationResultHolder;
 import com.linkedin.pinot.core.query.aggregation.function.AggregationFunction;
 import com.linkedin.pinot.core.query.aggregation.function.AggregationFunctionFactory;
+import com.linkedin.pinot.core.query.aggregation.function.AggregationFunctionType;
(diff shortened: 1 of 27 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/operator/query/MetadataBasedAggregationOperator.java
```diff
@@ -69,34 +69,26 @@ protected IntermediateResultsBlock getNextBlock() {
...
-          String minValueRange = segmentMetadata.getColumnMetadataFor(column).getMinValue().toString();
-          String maxValueRange = segmentMetadata.getColumnMetadataFor(column).getMaxValue().toString();
-          resultHolder = new ObjectAggregationResultHolder();
-          resultHolder.setValue(new MinMaxRangePair(Double.valueOf(minValueRange), Double.valueOf(maxValueRange)));
-          break;
-
-        default:
-          throw new UnsupportedOperationException(
-              "Metadata based aggregation operator does not support function " + function.getName());
+      if (functionName.equals(AggregationFunctionType.COUNT.getName())) {
```
kishoreg: you can still use the switch, just add a new method in the enum to return the functionType in of relying on valueOf.

Jackie-Jiang: Basically that's the current **boolean matchFunctionName(String functionName)**. I'll change it to **getAggregationFunctionType** and return the type or **null**

Jackie-Jiang:
@kishoreg Addressed the comments

PR #2902, opened 2018-06-21
[TE] rootcause - handle anomalies with multiple values for same dimension gracefully

apucher:


thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -283,7 +283,12 @@ export default Route.extend(AuthenticatedRouteMixin, {
         const analysisRange = [analysisRangeStart, analysisRangeEnd];
 
         const anomalyDimNames = anomalyEntity.attributes['dimensions'] || [];
-        const anomalyFilters = anomalyDimNames.map(dimName => [dimName, anomalyEntity.attributes[dimName]]);
+        const anomalyFilters = [];
+        anomalyDimNames.forEach(dimName => {
+          anomalyEntity.attributes[dimName].forEach(dimValue => {
+            anomalyFilters.pushObject([dimName, dimValue]);
```
aaronucsd: if you not wanting any duplicates. Consider `addObject`.

apucher: thanks for the info. key-value pairs should be unique from backend

PR #2904, opened 2018-06-22
[TE] Fix empty email screenshot

jihaozh:
This PR fixes empty email screenshot caused by PhantomJS script.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/resources/scripts/getGraphPnj.js
+++ b/thirdeye/thirdeye-pinot/src/main/resources/scripts/getGraphPnj.js
@@ -1,4 +1,4 @@
-// screenshot script for reference only 
+// screenshot script for reference only
 // actual used script lives in /thirdeye-configs/prod-config
 "use strict";
 
@@ -59,7 +59,7 @@ page.onLoadFinished = function(status) {
 
     if (requestsArray.length === 0 && isReady) {
       system.stderr.writeLine('taking screenshot');
-      
+
       clearInterval(interval);
         var clipRect = page.evaluate(function () {
           return document.querySelector('#anomaly-screenshot').getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/resources/scripts/getGraphPnj.js
```diff
@@ -73,7 +73,7 @@ page.onLoadFinished = function(status) {
         page.render(options.output);
         phantom.exit();
     }
-  }, 500);
+  }, 5000);
 
   interval();
```
aaronucsd: Why we doing this here? `interval` is a number that representing the ID value of the timer that is set, it's not a function. Can we remove this? It should already self execute with `setInterval()`. If this still an issue, we will revisit with await/async.

jihaozh: This script is used for PhantomJS to open an anomaly page and take a screenshot. Then it will be put in an alert email. I increased this delay here so that the script will take the screenshot after the page is fully loaded, otherwise we get an empty picture in the email.

aaronucsd: Yes. I understand. I'm just curious to why we adding this line `interval();` as it's not a function. The `setInterval()` should already does that.

jihaozh: I guess it's because `interval` is a function pointer to `setInterval`.  Adding `interval();` actually calls the function.

aaronucsd: screen shot 2018-06-22 at 10 47 38 am

jihaozh: Okay. It's redundant then. This was written a year ago. Maybe it's time to clean it up.

aaronucsd: Thanks JiHao!

PR #2906, opened 2018-06-22
Fix the race condition when modifying the same segment file on pinot server

Jackie-Jiang:
Add/Delete/Refresh/Reload operations might modify the segment file. Prevent them running on the same segment.
Use a global lock array to lock the segment based on hash value of table name and segment name.

NOTE: This is a quick fix. The clean fix would be managing everything in TableDataManager

### DIFF
--- a/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentLocks.java
+++ b/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentLocks.java
@@ -1,0 +1,35 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentLocks.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+package com.linkedin.pinot.server.starter.helix;
+
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+
+public class SegmentLocks {
+  private static final int NUM_LOCKS = 10000;
```
mcvsubbu: this should be made configurable, since we allocate all the locks all the time. Or, maybe allocate locks on demand?

Jackie-Jiang: There will be only 10K locks globally, which consumes less than 1MB memory.

PR #2908, opened 2018-06-22
[TE] anomalies - unified anomaly status classification

apucher:
* unified anomaly classification in dashboard and RCA. Allows on-the-fly computation of precision, recall, and other performance measures

* none, true positive, false positive, true negative, false negative (precision = TP / (TP + FP), recall = TN / (TN + FN))

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/ResourceUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/ResourceUtils.java
@@ -3,6 +3,9 @@ package com.linkedin.thirdeye.dashboard.resources.v2;
 import com.google.common.collect.SetMultimap;
 import com.google.common.collect.TreeMultimap;
 import com.linkedin.thirdeye.api.TimeGranularity;
+import com.linkedin.thirdeye.constant.AnomalyFeedbackType;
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/dashboard/resources/v2/ResourceUtils.java
```diff
@@ -283,5 +287,41 @@ private static void putExternalLinkTimeContext(long startTime, long endTime, Str
...
 
+  /**
+   * Resolves the (convoluted) anomaly feedback properties into a clear classification of
+   * {@code NONE, TRUE_POSITIVE, FALSE_POSITIVE, TRUE_NEGATIVE, FALSE_NEGATIVE}.
+   *
+   * @param anomaly anomaly dto
+   * @return feedback classification
+   */
+  public static AnomalyClassificationType getStatusClassification(MergedAnomalyResultDTO anomaly) {
+    if (anomaly.getAnomalyResultSource() != null) {
```
aaronucsd: As part of moving the anomaly report button to the dashboard. We will need a new resolution type of `User Reported + True anomaly`. Not sure it will need change here.

apucher: Yes, that's a false negative. (TE prediction negative, User condition positive)

PR #2910, opened 2018-06-22
[TE] rootcause - adjust slider margins to line up with chart

apucher:
proudly making my OCD happy since 2018

image

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-comparison-range/template.hbs
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-comparison-range/template.hbs
@@ -43,7 +43,7 @@
     applyAction=(action "onRange")
   }} --}}
   <!-- Investigation range slider -->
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-comparison-range/template.hbs
```diff
@@ -43,7 +43,7 @@
     applyAction=(action "onRange")
   }} --}}
   <!-- Investigation range slider -->
-  <div class="date-slider">
+  <div class="date-slider date-slider--offset">
```
aaronucsd: Looks good. Is there a reason not to fix this on the `.date-slider ` class itself vs creating a new modifier class?

apucher: not really, just figured the slider may be re-used somewhere else

PR #2917, opened 2018-06-25
[TE] Request statistic logger run daily

jihaozh:
This PR makes request statistics logger run daily instead of hourly.


jihaozh:
This PR makes request statistics logger run daily instead of hourly.

PR #2921, opened 2018-06-26
Print error log message when schema is null

jackjlli:
This PR prints error log message when realtime segment is about to upload while schema is null. Previously it only throws NPE.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
@@ -227,7 +227,11 @@ public class RealtimeTableDataManager extends BaseTableDataManager {
         return;
       }
       Schema schema = ZKMetadataProvider.getTableSchema(_propertyStore, _tableNameWithType);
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -227,7 +227,11 @@ public void addSegment(@Nonnull String segmentName, @Nonnull TableConfig tableCo
         return;
       }
       Schema schema = ZKMetadataProvider.getTableSchema(_propertyStore, _tableNameWithType);
-      Preconditions.checkNotNull(schema);
+      if (schema == null) {
```
mcvsubbu: We should be failing table add if schema is null. I think there was a PR that was merged which fixes the table add API. In that case, this null condition should arise only if someone deletes a schema, right? Precondition check seems fine for that? Maybe I am missing something?

jackjlli:
@mcvsubbu You're right. Schema is proved to be not null when adding table(https://github.com/linkedin/pinot/pull/2841). 
We don't need this PR any more. I'm closing this PR.

PR #2922, opened 2018-06-27
Dynamic rate limit for pinot brokers

jackjlli:
This PR uses dynamic rate limit for pinot brokers.
Basically dynamic per broker rate  = overall rate / number of online brokers.
If # of online brokers is changed, BrokerResourceOnlineOfflineStateModelFactory will update rate for a particular table.
If overall rate is modified, ClusterChangeMediator will capture the broker resourceEV change, and then update all the per broker rates with the latest qps quotas.
If overall rate is newly added for an existing table, since TableQueryQuotaManager only keeps track of the existing quotas, pinot brokers need to be restarted in order to fetch the newly added quota from table config in property store.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
@@ -25,8 +25,10 @@ import com.linkedin.pinot.common.metadata.ZKMetadataProvider;
 import com.linkedin.pinot.common.metrics.BrokerGauge;
 import com.linkedin.pinot.common.metrics.BrokerMetrics;
 import com.linkedin.pinot.common.utils.CommonConstants;
+import com.linkedin.pinot.common.utils.helix.HelixHelper;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.helix.HelixManager;
 import org.apache.helix.ZNRecord;
 import org.apache.helix.model.ExternalView;
@@ -34,18 +36,22 @@ import org.apache.helix.store.zk.ZkHelixPropertyStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.linkedin.pinot.common.utils.CommonConstants.Helix.*;
+
 
 public class TableQueryQuotaManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(TableQueryQuotaManager.class);
 
   private BrokerMetrics _brokerMetrics;
   private final HelixManager _helixManager;
+  private final AtomicInteger _lastKnownBrokerResourceVersion;
   private final Map<String, QueryQuotaConfig> _rateLimiterMap;
   private static int TIME_RANGE_IN_SECOND = 1;
 
   public TableQueryQuotaManager(HelixManager helixManager) {
     _helixManager = helixManager;
     _rateLimiterMap  = new ConcurrentHashMap<>();
+    _lastKnownBrokerResourceVersion = new AtomicInteger();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


kishoreg:
whats the motivation behind this change? If we simply divide total qps/number of online brokers, we will probably bring down the healthy brokers as well and potentially impact other use cases. Can we compute some kind of max possible qps per broker automatically and set that as the max qps and start throttling after that?.

jackjlli:
@kishoreg The motivation is that users don't know how many brokers are serving their tables behind the scene. If one broker goes down, overall qps quota can be shared across the rest of the brokers. So that from users' view, pinot cluster can provide fix overall rate.
If we use fixed rate for each broker, when one broker goes down, overall rate goes down as well. Since requests get throttled, users will notice the decrease of qps, which's not what they want.
I think the strategy you mention would be total qps / number of replicas(correct me if I'm wrong), which is a fixed rate. So it falls into the 2nd case.
And yeah, it'd be great to have more discussions on this.

kishoreg:
max_qps/replicas will be the minimum peak qps a broker will serve when all brokers are up. My suggestion is we should also find a max. With your PR max will be max_qps/1 and this might bring down the server or the queries will be timing out. Conceptually, max_qps a broker can handle is the highest qps at which a broker can still meet the SLA.

jackjlli:
@kishoreg Gotcha! We can bring in a factor. The max_qps can be calculated from min_qps multipled by this factor. The ideal situation is that all replicas are up and each broker'd use the min_qps quota. If some replicas go down, the qps rate should not exceed max_qps.

mcvsubbu:
@kishoreg I think what you are saying is that a single broker must have some upper limit of qps on a table. That is fine, except that we should ensure that when the client exceeds qps we return 429, but when the broker is overloaded, we return 503. Also, to implement means that we need one more config for each table. Maybe this can be a broker config as a percentage? (so, it will translate to some percentage of overall qps as the max handled by a single broker)

jackjlli:
I found an article from Stripe. Seems they have more strategies on rate limit. 
https://stripe.com/blog/rate-limiters

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -146,11 +151,10 @@ private void createRateLimiter(String tableName, ExternalView brokerResource, Qu
         }
       }
     }
-    LOGGER.info("The number of online brokers for table {} is {}", tableName, otherOnlineBrokerCount + 1);
-    //int onlineCount = otherOnlineBrokerCount + 1;
 
-    // FIXME We use fixed rate for the 1st version.
-    int onlineCount = 1;
+    int onlineCount = otherOnlineBrokerCount + 1;
+    System.out.println("The number of online brokers for table " + tableName + " is " + onlineCount);
```
jfim: Remove this System.out.println

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -165,6 +169,7 @@ private void createRateLimiter(String tableName, ExternalView brokerResource, Qu
     double perBrokerRate =  overallRate / onlineCount;
     QueryQuotaConfig queryQuotaConfig = new QueryQuotaConfig(RateLimiter.create(perBrokerRate), new HitCounter(TIME_RANGE_IN_SECOND));
     _rateLimiterMap.put(tableName, queryQuotaConfig);
+    System.out.println("Overall rate: " + overallRate + ". Per-broker rate: " + perBrokerRate);
```
jfim: Same

pinot-broker/src/main/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManager.java
```diff
@@ -234,7 +239,58 @@ public void cleanUpRateLimiterMap() {
...
 
+  /**
+   * Process query quota change when number of online brokers has changed.
+   */
   public void processQueryQuotaChange() {
-    // TODO: update rate
+    long startTime = System.currentTimeMillis();
+
+    ExternalView currentBrokerResource = HelixHelper.getExternalViewForResource(_helixManager.getClusterManagmentTool(),
+        _helixManager.getClusterName(), BROKER_RESOURCE_INSTANCE);
```
jfim: This can be `null` in certain cases, should log a warning and exit cleanly here.

PR #2924, opened 2018-06-27
[TE] detection - enable window function change points from anomalies

apucher:


apucher:
Let's see about that ...

PR #2925, opened 2018-06-28
[TE] detection - tutorial documentation

apucher:


aaronucsd:
A long time ago when i was coding java. We used to call them javadocs.

PR #2926, opened 2018-06-28
Re-implement TimerContext to avoid mixing of millis and nanos

Jackie-Jiang:
Because we don't need nanos resolution for latency, change all of then using wall-clock time
Also fix the bug in the scheduler where we compare CPU time with wall-clock time

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/query/ServerQueryRequest.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/query/ServerQueryRequest.java
@@ -41,12 +41,13 @@ public class ServerQueryRequest {
 
   private int _segmentCountAfterPruning = -1;
 
-  public ServerQueryRequest(@Nonnull InstanceRequest instanceRequest, @Nonnull ServerMetrics serverMetrics) {
+  public ServerQueryRequest(@Nonnull InstanceRequest instanceRequest, @Nonnull ServerMetrics serverMetrics,
+      long queryArrivalTimeMs) {
     _instanceRequest = instanceRequest;
     _serverMetrics = serverMetrics;
     _brokerRequest = instanceRequest.getQuery();
     _filterQueryTree = RequestUtils.generateFilterQueryTree(_brokerRequest);
-    _timerContext = new TimerContext(_brokerRequest, serverMetrics);
+    _timerContext = new TimerContext(_brokerRequest, serverMetrics, queryArrivalTimeMs);
   }
 
   /**
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/query/context/TimerContext.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/query/context/TimerContext.java
@@ -19,127 +19,81 @@ package com.linkedin.pinot.common.query.context;
 import com.linkedin.pinot.common.metrics.ServerMetrics;
 import com.linkedin.pinot.common.metrics.ServerQueryPhase;
 import com.linkedin.pinot.common.request.BrokerRequest;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.concurrent.TimeUnit;
-import javax.annotation.Nullable;
+
 
 public class TimerContext {
+  private final BrokerRequest _brokerRequest;
+  private final ServerMetrics _serverMetrics;
+  private final long _queryArrivalTimeMs;
+  private final Timer[] _phaseTimers = new Timer[ServerQueryPhase.values().length];
 
-  private final ServerMetrics serverMetrics;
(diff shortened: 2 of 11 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/query/context/TimerContext.java
```diff
@@ -16,130 +16,85 @@
...
-        endTimeNs = System.nanoTime();
-
-        recordPhaseTime(this);
-        isRunning = false;
-      }
-    }
-
-    public long getDurationNs() {
-      return endTimeNs - startTimeNs;
+      Preconditions.checkState(!_ended, "Timer has already stopped for query phase: " + _queryPhase);
```
mcvsubbu: I think it is OK to not do a Preconditions check here. We should not fail a query because of some bug in our code that calls stop() twice. I would suggest to replace this with a warning and return

pinot-common/src/main/java/com/linkedin/pinot/common/query/context/TimerContext.java
```diff
@@ -16,130 +16,85 @@
...
+      _durationNs = System.nanoTime() - _startTimeNs;
```
mcvsubbu: should you account for the fact that this may be negative (if stop is invoked from a different thread)?

pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/MultiLevelPriorityQueue.java
```diff
@@ -189,20 +186,20 @@ private SchedulerQueryContext takeNextInternal() {
...
 
-    SchedulerQueryContext query = null;
+    SchedulerQueryContext queryContext = null;
     if (currentWinnerGroup != null) {
-      ServerQueryRequest queryRequest = currentWinnerGroup.peekFirst().getQueryRequest();
+      queryContext = currentWinnerGroup.removeFirst();
+      ServerQueryRequest queryRequest = queryContext.getQueryRequest();
       if (LOGGER.isDebugEnabled()) {
-        sb.append(String.format(" Winner: %s: [%d,%d,%d,%d]", currentWinnerGroup.name(), queryRequest.getTimerContext().getQueryArrivalTimeMs(),
+        sb.append(String.format(" Winner: %s: [%d,%d,%d,%d]", currentWinnerGroup.name(), queryContext.getArrivalTimeNs(),
```
mcvsubbu: I think we are looking to get absolute time in this log, ns time is not going to help us. I am not sure how to get that unless we record both System.currentMillis and System.nanos on query arrival. On the other hand, this is a debug log, so maybe we are ok ...

pinot-common/src/main/java/com/linkedin/pinot/common/query/context/TimerContext.java
Jackie-Jiang: Fixed.

pinot-common/src/main/java/com/linkedin/pinot/common/query/context/TimerContext.java
Jackie-Jiang: System.nanoTime() is per JVM based, so will not be negative. Also, I'm planning to use System.currentTimeMillis() because we don't need such high resolution.

pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/MultiLevelPriorityQueue.java
Jackie-Jiang: Fixed.

pinot-common/src/main/java/com/linkedin/pinot/common/query/ServerQueryRequest.java
```diff
@@ -41,12 +41,17 @@
...
+      long queryArrivalTimeMs) {
     _instanceRequest = instanceRequest;
     _serverMetrics = serverMetrics;
     _brokerRequest = instanceRequest.getQuery();
     _filterQueryTree = RequestUtils.generateFilterQueryTree(_brokerRequest);
-    _timerContext = new TimerContext(_brokerRequest, serverMetrics);
+    _timerContext = new TimerContext(_brokerRequest, serverMetrics, queryArrivalTimeMs);
+  }
+
+  public ServerQueryRequest(@Nonnull InstanceRequest instanceRequest, @Nonnull ServerMetrics serverMetrics) {
```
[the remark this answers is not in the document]
Jackie-Jiang: This is used by test only. Modified the test and removed the constructor.

PR #2929, opened 2018-06-28
Adding fat jar for pinot-hadoop

jenniferdai:
Please let me know if anyone has more maven expertise or knows of a more correct way to do this. This will create multiple jars, one of which becomes a fat jar.


kishoreg:
can you add the instructions to ReadMe and update the docs in the wiki.

PR #2933, opened 2018-07-03
[PINOT-6609] Ignore tables for segment status check if they are disabled in ideal state

sunithabeeram:
This PR has 2 main changes:
- Segment status checking is skipped for tables that are disabled in helix. Instead we now emit a disabled table count.
- The run() loop for the status check is modified to catch/log all throwables as any error in the run method will result in further checks silently suppressed (and typically, there will be no associated log for it to help debugging either)

Added unit test to cover the new code.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
@@ -64,6 +64,10 @@ public class SegmentStatusChecker {
   private final int _waitForPushTimeSeconds;
   private final TableSizeReader _tableSizeReader;
 
+  // log messages about disabled tables atmost once a day
+  private static final long DISABLED_TABLE_LOG_INTERVAL_MS = TimeUnit.DAYS.toMillis(1);
+  private long _lastDisabledTableLogTimestamp = 0;
+
   /**
    * Constructs the segment status checker.
    * @param pinotHelixResourceManager The resource checker used to interact with Helix
@@ -105,7 +109,9 @@ public class SegmentStatusChecker {
       public void run() {
         try {
           updateSegmentMetrics();
-        } catch (Exception e) {
+        } catch (Throwable e) {
+          // catch all errors to prevent subsequent exeuctions from being silently suppressed
+          // Ref: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay-java.lang.Runnable-long-long-java.util.concurrent.TimeUnit-
           LOGGER.warn("Caught exception while running segment status checker", e);
         }
       }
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -162,6 +165,8 @@ void updateSegmentMetrics() {
         } else {
           realTimeTableCount++;
         }
+
+
```
jfim: Formatting

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -177,6 +182,13 @@ void updateSegmentMetrics() {
           _metricsRegistry.setValueOfTableGauge(tableName, ControllerGauge.PERCENT_SEGMENTS_AVAILABLE, 100);
           continue;
         }
+
+        if (!idealState.isEnabled()) {
+          LOGGER.warn("Table {} is disabled. Skipping segment status checks", tableName);
```
jfim: If the table is disabled, the metrics for that table should be removed as well (see https://github.com/linkedin/pinot/pull/2933/files#diff-ec0bdaf9277528f6977db3cd163c518aR278)

sunithabeeram: Thanks for pointing that out. Fixed.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -177,6 +180,14 @@ void updateSegmentMetrics() {
...
```
mcvsubbu: Not sure if it is good to put out a warning here every 5 minutes.

sunithabeeram: Good point. I've updated the code to not log it more than once a day. Let me know if this works better.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -153,8 +159,19 @@ void updateSegmentMetrics() {
...
     int realTimeTableCount = 0;
     int offlineTableCount = 0;
+    int disabledTableCount = 0;
     ZkHelixPropertyStore<ZNRecord> propertyStore= _pinotHelixResourceManager.getPropertyStore();
 
+    // check if we need to log disabled tables log messages
+    boolean logDisabledTables = false;
+    long now = System.currentTimeMillis();
+    if (now -_lastDisabledTableLogTimestamp >= DISABLED_TABLE_LOG_INTERVAL_MS) {
+      logDisabledTables = true;
```
mcvsubbu: would it make sense to name this variable as 'noisy' and use it for other logs/warnings as well? Not sure what else is there in this class.

sunithabeeram: On a quick review, I didn't find others that might be candidates and didn't want to change any existing logs that we might be relying on. If we have more such instances then we could update the variable and logic accordingly at that time.

PR #2935, opened 2018-07-05
[PINOT-6614] Log more messages in the error path for table size estimations and handle estimate of -1 better

sunithabeeram:
- The code currently doesn't seem to handle the case where estimated size is -1. The PR makes it explicit that we intentionally allow the quota check to go through in this case.
- Add more log messages when we take the error paths to make it easier to understand how the estimates are arrived at.


kishoreg:
Unrelated to this PR but have a question - Are we calling each server for every segment upload. That seems to be overkill for use cases where we have approx million segments.

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/StorageQuotaChecker.java
```diff
@@ -110,6 +110,12 @@ public QuotaCheckerResponse isSegmentStorageWithinQuota(@Nonnull File segmentFil
       throw e;
     }
 
+    if (tableSubtypeSize.estimatedSizeInBytes == -1) {
```
kishoreg: will this result in rejecting the segment?

sunithabeeram:
>Are we calling each server for every segment upload?
Yes. In the (longish) email summarizing the issue, I have mentioned the possibility of using a cache that's periodically refreshed. 

We could perhaps even do with a controller local in-memory cache; if the table size is close to the limit and we are uploading a large chunk, the controller can kick off an ondemand refresh and/or fire off an alert requiring the quota to be reviewed. Anytime we are around 80% of quota usage, its probably good to review and adjust table quota.

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/StorageQuotaChecker.java
sunithabeeram: Yes, thats true. I haven't finalized this change yet (pending todos to say the least). The main point though is that estimate can be -1 and we shouldn't directly use it in the code that follows. If we go with the approach of caching, we could fall back to the last cached value etc; or accept the segment, but log a message/fire a metric that says we are flying blind a bit.

I can pull this out from this PR and leave only the logging in; we can review table-size estimations and address this issue along with others we would like to improve on.

sunithabeeram: Based on offline discussion, we'll not reject segment in this case. I've made the handling of this case clearer.

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableSizeReader.java
```diff
@@ -194,26 +194,33 @@ public TableSubTypeSizeDetails getTableSubtypeSize(String tableNameWithType, int
       // after iterating over all servers update summary reported and estimated size of the segment
       if (errors != segmentSizes.serverInfo.size()) {
         // atleast one server reported size for this segment
+        LOGGER.info("Could not get size for segment {} from {} servers. Using segmentLevelMax {} to estimate the size",
```
mcvsubbu: What happens if we segmentLevelMax is -1 (because we got errors from all replicas hosting a specific segment)? I know you mentioned there is an error in the math, if you are planning to fix it later, can u pls add a TODO?

sunithabeeram: In that case, the code uses table-level max segment size. If table level max is -1, then we return -1 as the estimate.

PR #2936, opened 2018-07-09
[PINOT-6513] Initial changes for segment refresh time prefetch

sunithabeeram:
Initial set of changes to load "hot" segment pages into memory. This change is to address increased latencies we see when such segments are refreshed/reloaded. The code determines if a segment is hot or not based on its reference count. This is just a heuristic - its possible that once the current queries drain, no new ones hit the loaded segment. 

Additionally, it appears that the earlier code would reload pages upto 66G (and headers only till 100G limit) - this seems a bit excessive given the amount of available physical memory on servers.

I plan to run to tests to validate the changes - posting this mainly for early feedback.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/PrefetchMode.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/PrefetchMode.java
@@ -1,0 +1,40 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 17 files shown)


kishoreg:
Like the idea of coming up with some sort of list. But looking at the previous/current query is probably to short. We can instead maintain an access count for the segments in the last X minutes and load the top ones until we hit the physical memory limit * some X percent.

Another solution is to pre-fetch only part of the segments such as dictionary and inverted index, star tree since they are accessed entirely for every query. Forward index can be skipped since the part of forward index accessed depends on the actual values in the query.

Also, what are the scenarios where this will be helpful. If the servers have to go over all the segments any way, we may not really benefit from pre-fetching

sunithabeeram:
Valid points about prefetching only the inverted/startree indices - I think thats a good optimization to make. 
Regarding all segments being loaded - I guess we need to make some considerations based on whether the data set fits in memory or not. If dataset > memory, then the routing tables generated should probably weigh segment locality more.

I think prefetching still helps even if all segments are hot - the reason being we warm up the page cache before switching requests to the new segment.

antumbde:
Do we see higher major page faults around the same time? If not, this will not solve the problem. Server unzips the segment during load which will cause pages to be in buffer. Prefetching should not be required.

npawar:

PR #2937, opened 2018-07-09
Add table APIs for the new config format

jfim:


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableConfigRestletResource.java
```diff
@@ -0,0 +1,188 @@
...
+      LOGGER.warn("Failed to deserialize the table configuration: {}", tableConfiguration);
+      return Response
+          .serverError()
+          .entity("Failed to deserialize the table configuration")
+          .type(MediaType.TEXT_PLAIN_TYPE)
+          .build();
+    }
+
+    if (config.getSchema() != null) {
+      _resourceManager.addOrUpdateSchema(config.getSchema());
```
sunithabeeram: It looks like resourceManager APIs can throw runtime exceptions some of which indicate client side errors (ie, broker tenant does not exist etc). Shouldn't we catch those and throw appropriate 4xxs?

(There are more occurences of this in the rest of the file)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableConfigRestletResource.java
```diff
@@ -0,0 +1,188 @@
...
+        } catch (IOException e) {
+          LOGGER.warn("Failed to update the offline table configuration for table {}", e, config.getOfflineTableConfig().getTableName());
+          return Response
+              .serverError()
+              .entity("Failed to update the offline table configuration")
+              .type(MediaType.TEXT_PLAIN_TYPE)
+              .build();
+        }
+      } else {
+        _resourceManager.addTable(config.getOfflineTableConfig());
```
sunithabeeram: This seems unexpected as this update API. Can we log a message saying we are adding a table?

jfim: `addTable` in `PinotHelixResourceManager` already logs the table addition (see https://github.com/linkedin/pinot/blob/6fbb79e734a9078190d3f4ee29c62e487c9100dd/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/PinotHelixResourceManager.java#L1120).

The fact that a table is added if a new table counterpart is added to the table configuration is an implementation detail, from the users' perspective, the table types just changed from say `[OFFLINE]` to `[OFFLINE, REALTIME]`.

jfim:
Is there anything that's blocking this review?

sunithabeeram:
It looks like not all comments were addressed. The code also needs to be rebased.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableConfigRestletResource.java
jfim: None of the 400 series of error codes are appropriate for this purpose. If there is an exception that is thrown by the restlet, this results in 500 (just like other error handlers), which is the expected code.

sunithabeeram: I am not sure if I follow. If someone tries to add a new table with a broker tenant that does not exist, is that not a client error? In which case, isn't 400 the right error code?

jfim: The request is fully formed and syntactically correct, so HTTP status code 400 does not apply in this case.

jfim: Since this has been blocked for over two months, what is the expected code change here. It seems that this pull request is blocked over a minor squabble over which HTTP status code should be returned instead of having a blocker on the design or anything major.

Considering this API isn't in active use and can be changed later, that this API not being available blocks adding the appropriate client tools for the new configuration format so that we can actually exercise the feature in order to make sure that it does work as expected, it seems like this level of scrutiny misses the forest for the trees at this point.

PR #2938, opened 2018-07-10
Realtime hosts provisioning tool

npawar:
Given the num replicas, num partitions and a sample completed segment for a realtime table to be setup, this tool will provide memory used by each host and an optimal segment size for various combinations of hours to consume and hosts to provision

Options:
- numReplicas: number of replicas for the table (required=true)
- numPartitions: number of stream partitions for the table (required=true)
- retentionHours: number of hours we would require this segment to be in memory. 
- numHosts: number of hosts as comma separated values (default 2,4,6,8,10,12,14,16) (required=false)
- numHours: number of hours to consume as comma separated values (default 2,3,4,5,6,7,8,9,10,11,12) (required=false)
- sampleCompletedSegmentDir: Consume from the topic for n hours and provide the path of the segment dir after it completes (required=true)
- periodSampleSegmentConsumed: Period for which the sample segment was consuming in format 4h, 5h30m, 40m etc (required=true)
- 

Example:
RealtimeHostsProvisioningCommand -numReplicas 3 -numPartitions 20 -sampleCompletedSegmentDir /path/to/segment/dir/segmentDir -periodSampleSegmentConsumed 6h

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/RealtimeHostsProvisioningCommand.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/RealtimeHostsProvisioningCommand.java
@@ -52,8 +52,8 @@ public class RealtimeHostsProvisioningCommand extends AbstractBaseAdminCommand i
   private static final String TMP_DIR = System.getProperty("java.io.tmpdir") + File.separator;
   private static final String STATS_FILE_NAME = "stats.ser";
 
-  private final int[] NUM_HOSTS = {2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12};
-  private final int[] NUM_HOURS_TO_CONSUME = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12};
+  private final int[] NUM_HOSTS = {2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15};
+  private final int[] NUM_HOURS_TO_CONSUME = {2, 4, 6, 8, 9, 10, 11, 12};
 
   @Option(name = "-numReplicas", required = true, metaVar = "<int>", usage = "number of replicas for the table")
   private int _numReplicas;
@@ -344,10 +344,12 @@ public class RealtimeHostsProvisioningCommand extends AbstractBaseAdminCommand i
    *****************************************************************/
   private void display(String[][] totalMemoryPerHost, String[][] optimalSegmentSize) {
 
+    System.out.println("\nMemory used per host");
     printNumHostsHeader();
     printNumHoursHeader();
     printOutputValues(totalMemoryPerHost);
 
+    System.out.println("\nOptimal segment size");
     printNumHostsHeader();
     printNumHoursHeader();
     printOutputValues(optimalSegmentSize);
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/RealtimeHostsProvisioningCommand.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/RealtimeHostsProvisioningCommand.java
@@ -33,6 +33,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import org.apache.commons.io.FileUtils;
 import org.kohsuke.args4j.Option;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -121,8 +122,8 @@ public class RealtimeHostsProvisioningCommand extends AbstractBaseAdminCommand i
   public boolean execute() throws Exception {
     LOGGER.info("Executing command: {}", toString());
 
-    long[][] totalMemory = new long[NUM_HOSTS.length][NUM_HOURS_TO_CONSUME.length];
-    long[][] optimalSegmentSize = new long[NUM_HOSTS.length][NUM_HOURS_TO_CONSUME.length];
+    long[][] totalMemoryPerHostMB = new long[NUM_HOSTS.length][NUM_HOURS_TO_CONSUME.length];
+    long[][] optimalSegmentSizeMB = new long[NUM_HOSTS.length][NUM_HOURS_TO_CONSUME.length];
     File sampleCompletedSegmentFile = new File(_sampleCompletedSegmentDir);
     SegmentMetadataImpl segmentMetadata;
     try {
@@ -146,7 +147,7 @@ public class RealtimeHostsProvisioningCommand extends AbstractBaseAdminCommand i
         int totalConsumingPartitionsPerHost = (totalConsumingPartitions + numHosts - 1) / numHosts;
 
         // calculate completed component
-        long sampleCompletedSegmentSizeBytes = sampleCompletedSegmentFile.length();
+        long sampleCompletedSegmentSizeBytes = FileUtils.sizeOfDirectory(sampleCompletedSegmentFile);
         // consuming for _numHoursSampleSegmentConsumed, and got size sampleCompletedSegmentSizeBytes
         // hence, consuming for numHoursToConsume would give:
         long completedSegmentSizeBytes =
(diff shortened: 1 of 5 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,342 @@
...
+   * Initialize the stats file using the sample segment provided.
+   * This involves indexing each row of the sample segment using MutableSegmentImpl.
+   * Invoking a destroy on the MutableSegmentImpl at the end will dump the collected stats into the stats.ser file provided in the statsHistory.
+   */
+  public File initializeStatsHistory() {
+
+    File statsFile = new File(_tableDataDir, STATS_FILE_NAME);
+    RealtimeSegmentStatsHistory sampleStatsHistory;
+    try {
+      sampleStatsHistory = RealtimeSegmentStatsHistory.deserialzeFrom(statsFile);
```
mcvsubbu: There will be no stats file to start with, so we don't need to deserialize anything here, right?

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,342 @@
...
+
+      long totalMemoryForCompletedSegmentsPerPartition =
+          calculateMemoryForCompletedSegmentsPerPartition(completedSegmentSizeBytes, numHoursToConsume, retentionHours);
+
+      int totalDocsInSampleSegment = _segmentMetadata.getTotalDocs();
+      // numHoursSampleSegmentConsumed created totalDocsInSampleSegment num rows
+      // numHoursToConsume will create ? rows
+      int totalDocs = (int) (((double) minutesToConsume / _minutesSampleSegmentConsumed) * totalDocsInSampleSegment);
+
+      File statsFileCopy = new File(_tableDataDir, STATS_FILE_COPY_NAME);
```
mcvsubbu: will be good to add a comment here saying why you are creating a copy before using the stats file

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,342 @@
...
+      FileUtils.copyFile(statsFile, statsFileCopy);
+      RealtimeSegmentStatsHistory statsHistory;
+      try {
+        statsHistory = RealtimeSegmentStatsHistory.deserialzeFrom(statsFileCopy);
```
mcvsubbu: so, this is just to make sure that the stats file is readable? I assume that the copy call above will fail if the origin file does not exist, or if the copy is not successful.

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,342 @@
...
+   * @param retentionHours - what is the amount of retention in memory expected for completed segments
+   * @throws IOException
+   */
+  public void estimateMemoryUsed(File statsFile, int[] numHosts, int[] numHours, int totalConsumingPartitions,
+      int retentionHours) throws IOException {
+    _totalMemoryPerHost = new String[numHours.length][numHosts.length];
+    _optimalSegmentSize = new String[numHours.length][numHosts.length];
+    _consumingMemoryPerHost = new String[numHours.length][numHosts.length];
+
+    int i = 0;
```
mcvsubbu: A block comment here describing your algorithm will be nice.

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,342 @@
...
+    for (int numHoursToConsume : numHours) {
+      long minutesToConsume = numHoursToConsume * 60;
+      // consuming for _numHoursSampleSegmentConsumed, gives size sampleCompletedSegmentSizeBytes
+      // hence, consuming for numHoursToConsume would give:
+      long completedSegmentSizeBytes =
```
mcvsubbu: Do you want to do this math in seconds instead of minutes? For high ingestion topics, it may make significant difference?

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,342 @@
...
+      // create mutable segment impl
+      MutableSegmentImpl mutableSegmentImpl = new MutableSegmentImpl(realtimeSegmentConfigBuilder.build());
+      long memoryForConsumingSegmentPerPartition = memoryManager.getTotalAllocatedBytes();
+      mutableSegmentImpl.destroy();
+      FileUtils.deleteQuietly(statsFileCopy);
+
+      memoryForConsumingSegmentPerPartition += getMemoryForInvertedIndex(memoryForConsumingSegmentPerPartition);
+
+      int j = 0;
+      for (int numHostsToProvision : numHosts) {
```
mcvsubbu: may be better to include the increment of j in the for() statement, so that statements like 'continue' are handled correctly. Something like this:
for (int j = 0; j < numHosts.length; j++) {
  numHostsToProvision = numHosts[j]

Same for the outer for loop

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
npawar: I need to initialize it before I can use it right? I thought that is the way I can map the object to that file

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/RealtimeHostsProvisioningCommand.java
```diff
@@ -0,0 +1,202 @@
...
+
+  @Option(name = "-numPartitions", required = true, metaVar = "<int>",
+      usage = "number of stream partitions for the table")
+  private int _numPartitions;
+
+  @Option(name = "-retentionHours", metaVar = "<int>",
+      usage = "number of hours we would require this segment to be in memory. "
+      + "\nThis would depend on the retention set in the table config for the table, as well as the frequency of offline flows. "
+      + "\nWhen the offline segments become available, those will be used for querying instead of the realtime segments."
+      + "\neg. If daily flows exist, we would need realtime segments in memory only for 2 days.")
```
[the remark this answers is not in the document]
npawar: Fixed, defaulting to 3 days

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
npawar: Sure, changed it to seconds. Though this would depend on whether the user of the tool is precise enough in the inputs

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
npawar: Changed

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/RealtimeProvisioningHelperCommand.java
```diff
@@ -0,0 +1,202 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
```
mcvsubbu: 2014-2018

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,364 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
```
mcvsubbu: 2014-2018

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
mcvsubbu: Ah yes, the other ctor is hidden. Sorry, my bad.

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,364 @@
...
+      FileUtils.deleteDirectory(_tableDataDir);
+    } catch (IOException e) {
+      throw new RuntimeException("Exception in deleting directory " + _tableDataDir.getAbsolutePath(), e);
+    }
+    _tableDataDir.mkdir();
+  }
+
+  /**
+   * Initialize the stats file using the sample segment provided.
+   * This involves indexing each row of the sample segment using MutableSegmentImpl.
```
mcvsubbu: you can add a comment here saying this is equivalent to consuming the rows of the segment, albeit in a different order than that consumed by the host. But the stats should be the same

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,364 @@
...
+
+  /**
+   * Algorithm:
+   * <br>Given numReplicas and numPartitions, we can find out total consuming partitions per host, for various numHosts
+   * <br><b>totalConsumingPartitionsPerHost = (numReplicas * numPartitions)/numHosts</b>
+   * <br>
+   * <br>Given a sample realtime completed segment (with size s), and how long it consumed for (t),
+   * <br>we can estimate how much memory the table would require for various combinations of num hosts and num hours
+   * <br>
+   * <br>For completed segments-
```
mcvsubbu: "For estimating the memory occupied by completed segments"

pinot-tools/src/main/java/com/linkedin/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -0,0 +1,364 @@
...
+   * <br>For each numHoursToConsume we compute:
+   * <br>If a segment with size s takes time t to complete, then consuming for time numHoursToConsume would create segment with size <b>estimatedSize = (numHoursToConsume/t)*s</b>
+   * <br>If retention for completed segments in memory is rt hours, then the segment would be in memory for <b>(rt-numHoursToConsume) hours</b>
+   * <br>A segment would complete every numHoursToConsume hours, so we would have at a time <b>numCompletedSegmentsAtATime = (rt-numHoursToConsume)/numHoursToConsume</b> to hold in memory
+   * <br>As a result, <b>totalCompletedSegmentsMemory per ConsumingPartition = estimatedSize * numCompletedSegmentsAtATime</b>
+   * <br>
+   * <br>For consuming segments-
```
mcvsubbu: "For estimating the memory occupied by consuming segments"

npawar:
Addressed all the comments. Thanks @mcvsubbu

PR #2939, opened 2018-07-10
Fixing maven shade plugin for SegmentTarPush so it doesn't include ht…

jenniferdai:
…tpclient

### DIFF
--- a/pinot-common/pom.xml
+++ b/pinot-common/pom.xml
@@ -295,6 +295,7 @@
                   <goal>shade</goal>
                 </goals>
                 <configuration>
+                  <shadedArtifactAttached>true</shadedArtifactAttached>
                   <relocations>
                     <relocation>
                       <pattern>org.apache.http</pattern>
--- a/pinot-common/pom.xml
+++ b/pinot-common/pom.xml
@@ -295,57 +295,6 @@
                   <goal>shade</goal>
                 </goals>
                 <configuration>
-                  <artifactSet>
-                    <excludes>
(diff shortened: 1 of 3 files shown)


pinot-hadoop/pom.xml
```diff
@@ -72,8 +72,18 @@
...
+        <exclusion>
+          <groupId>com.linkedin.pinot</groupId>
+          <artifactId>pinot-common</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>com.linkedin.pinot</groupId>
+      <artifactId>pinot-common</artifactId>
+      <version>0.016</version>
```
sunithabeeram: I don't think we should hard code the version.

jenniferdai: good call, removed

pinot-common/pom.xml
```diff
@@ -295,6 +295,7 @@
                   <goal>shade</goal>
                 </goals>
                 <configuration>
+                  <shadedArtifactAttached>true</shadedArtifactAttached>
```
sunithabeeram: Is this still required if we run with -Pbuild-shaded-jar above?

pinot-hadoop/pom.xml
```diff
@@ -49,6 +49,7 @@
                   <goal>shade</goal>
                 </goals>
                 <configuration>
+                  <shadedArtifactAttached>true</shadedArtifactAttached>
```
sunithabeeram: Same question as above.

PR #2943, opened 2018-07-11
Lower the probability of intermittent test failure for LargeClusterRoutingTableBuilderTest

mcvsubbu:
This test fails occasionally when the variance between least loaded and highest loaded server
is more than 50% across all routing tables. Since we have a random generation of routing tables
it is not guaranteed that this is always the case.

Lowered the probability of failure further by testing the criteria 5 consecutive times before
asserting a failure.

### DIFF
--- a/pinot-broker/src/test/java/com/linkedin/pinot/broker/routing/builder/LargeClusterRoutingTableBuilderTest.java
+++ b/pinot-broker/src/test/java/com/linkedin/pinot/broker/routing/builder/LargeClusterRoutingTableBuilderTest.java
@@ -26,7 +26,6 @@ import java.util.Set;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.InstanceConfig;
 import org.testng.Assert;
-import org.testng.annotations.BeforeSuite;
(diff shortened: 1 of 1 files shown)


pinot-broker/src/test/java/com/linkedin/pinot/broker/routing/builder/LargeClusterRoutingTableBuilderTest.java
```diff
@@ -35,17 +37,15 @@
...
-  private static final long RANDOM_SEED = System.currentTimeMillis();
   private LargeClusterRoutingTableBuilder _largeClusterRoutingTableBuilder = new LargeClusterRoutingTableBuilder();
 
   private interface RoutingTableValidator {
     boolean isRoutingTableValid(Map<String, List<String>> routingTable, ExternalView externalView,
         List<InstanceConfig> instanceConfigs);
   }
 
-  @Test
+  @BeforeSuite
```
Jackie-Jiang: Remove the setUp() method

mcvsubbu:
@jfim I have seen it break a few times before.

PR #2946, opened 2018-07-11
Add some messages to trace zk behaviors

jackjlli:
This PR won't be merged. 
It's used for testing zk in Travis.


sunithabeeram:
Since the issue hasn't reproduced recently, I think we should keep the sys.out.printlns to aid debugging when it happens next and merge the change.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ZkStarter.java
```diff
@@ -137,31 +139,39 @@ public static ZookeeperInstance startLocalZkServer(final int port) {
    * @param dataDirPath The path for the Zk data directory
    */
   public synchronized static ZookeeperInstance startLocalZkServer(final int port, final String dataDirPath) {
+    // Disable SASL-authenticated for starting local zk server
+    String isSaslEnable = System.setProperty(ZK_SASL_CLIENT, "false");
```
sunithabeeram: Why are we doing this?

pinot-broker/src/test/java/com/linkedin/pinot/broker/queryquota/TableQueryQuotaManagerTest.java
```diff
@@ -51,9 +55,18 @@
...
   private static String REALTIME_TABLE_NAME = RAW_TABLE_NAME + "_REALTIME";
 
-
   @BeforeTest
   public void beforeTest() {
+    Enumeration<Logger> loggers = Logger.getRootLogger().getLoggerRepository().getCurrentLoggers();
+    while (loggers.hasMoreElements()) {
+      Logger logger = loggers.nextElement();
+      System.out.println("Log name: " + logger.getName());
+      if (logger.getName().startsWith("com.linkedin.pinot.common.utils") ||
```
sunithabeeram: I think it will be sufficient if we enabled it only for ZkStarter right? Also, can this be added in a common/base class instead of setup for multiple tests? Is logging initialized per test? 
Also, should this be reset in a teardown method?

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ZkStarter.java
jackjlli: There's no need to enable SASL--authenticated if we start a local zk server. So I think there's no harm for us to reduce one possible cause for the timeout.

PR #2949, opened 2018-07-12
Retrieve hadoop parameters from controller

jenniferdai:
* In the offline hadoop job, we do not have the option to retrieve the schema and table config from the controller host in order to create segments. We depend on the user to pass in parameters regarding whether their table is append/refresh and we serialize the schema file they pass in. It is better to have the control on the Pinot side so that their parameters can be consistent with what they set in table config. For now, I have added append/refresh from table config and using the schema we retrieve.
* This change paves the way for us to add startree/hll/other parameters from table config into segment generation. This will also allow Pinot admins to set table-level control according to parameters on their side (eg: if inverted index shouldn't be generated, etc)
* Will add option for https/http into the PushLocation object along with other options

### DIFF
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/JobConfigConstants.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/JobConfigConstants.java
@@ -21,4 +21,11 @@ public class JobConfigConstants {
   public static final String PATH_TO_OUTPUT = "path.to.output";
 
   public static final String TARGZ = ".tar.gz";
+
+  public static final String TIME_COLUMN_NAME = "table.time.column.name";
+  public static final String TIME_COLUMN_TYPE = "table.time.column.type";
+  public static final String TABLE_PUSH_TYPE = "table.push.type";
+  public static final String SCHEMA = "data.schema";
+  public static final String SEGMENT_TABLE_NAME = "segment.table.name";
+
 }
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/JobConfigConstants.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/JobConfigConstants.java
@@ -28,4 +28,7 @@ public class JobConfigConstants {
   public static final String SCHEMA = "data.schema";
   public static final String SEGMENT_TABLE_NAME = "segment.table.name";
 
+  public static final String PUSH_TO_HOSTS = "push.to.hosts";
+  public static final String PUSH_TO_PORT = "push.to.port";
+
 }
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
@@ -16,8 +16,10 @@
 package com.linkedin.pinot.hadoop.job;
 
 import com.linkedin.pinot.common.Utils;
+import com.linkedin.pinot.common.config.TableConfig;
 import com.linkedin.pinot.common.data.Schema;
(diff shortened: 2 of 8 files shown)


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -190,7 +199,31 @@ public void run() throws Exception {
...
+    for (String host : _hosts) {
+      pushLocations.add(new PushLocation.PushLocationBuilder()
+          .setHost(host)
+          .setPort(_port)
+          .build());
+    }
+
+    ControllerRestApi controllerRestApiObject = new ControllerRestApi(pushLocations, _tableName);
+
+    TableConfig tableConfig = controllerRestApiObject.getTableConfig();
```
kishoreg: why do we have to do this, why cant we set the entire table config. This is adding unnecessary constants for every field in table config

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/TableConfigConstants.java
```diff
@@ -0,0 +1,5 @@
+package com.linkedin.pinot.hadoop.job;
+
+public class TableConfigConstants {
```
kishoreg: this class is not needed

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
jenniferdai: In the next one, I plan to add a new constructor in SegmentGeneratorConfig that takes in table config and sets parameters (startree, etc, so we don't have to serializet he whole thing). However, I like the extra logic/control that I feel should not go into the constructor for this parameter in particular in particular because I want the segment names to be correct.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/TableConfigConstants.java
jenniferdai: Reasoning is to separate out constants that we get from table config vs hadoop constants; there will be more coming. It looks sparse for now because it's piecemeal/haven't put in the majority of table config constants. I can combine them if you prefer though!

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/JobConfigConstants.java
```diff
@@ -21,4 +21,11 @@
   public static final String PATH_TO_OUTPUT = "path.to.output";
 
   public static final String TARGZ = ".tar.gz";
+
```
snleee: Can you clean up the constants from all hadoop jobs (`SegmentCreation, SegmentTarPush, SegmentUriPush`)? We use some of constants from this class and we use hard coded values for some other constants (e.g. `push.to.host`, `push.to.port`)

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -69,6 +75,9 @@ public SegmentCreationJob(String jobName, Properties properties) throws Exceptio
     _outputDir = getOutputDir();
     _stagingDir = new File(_outputDir, TEMP).getAbsolutePath();
     _depsJarPath = _properties.getProperty(PATH_TO_DEPS_JAR, null);
+    _hosts = _properties.getProperty("push.to.hosts").split(",");
```
snleee: I think that you need to provide the backward compatibility in case when no push location is provided. This will affect to people outside who are using this push job.

jenniferdai: good call, fixed

PR #2950, opened 2018-07-12
Raise a metric indicating the response from the controller for segment commit

npawar:
We were experiencing socket timeouts from the controller when posting segment commit message. This issue went unnoticed because we never did any reporting of the status of the post, and because if the other replicas committed successfully, the problem never has a chance to leave a trail on the consumption or segments.


npawar:
Moved the metric logic to ServerSegmentCompletionProtocol

PR #2952, opened 2018-07-13
Fix NettyCloseChannelTest

jackjlli:
This PR fixes NettyCloseChannelTest. 
Previously it failed occasionally because of timing out in Travis.


pinot-transport/src/test/java/com/linkedin/pinot/transport/netty/NettyCloseChannelTest.java
```diff
@@ -32,6 +34,7 @@
   private CountDownLatch _countDownLatch;
   private NettyTCPServer _nettyTCPServer;
   private NettyTCPClientConnection _nettyTCPClientConnection;
+  private final long CONNECTION_TIMEOUT = 10 * 1000L; // 10 seconds
```
sunithabeeram: It would be better to rename this as READ/RESONSE timeout and not CONNECTION (typically thats used to indicate just the timeout to 'connect' to the server). Also, indicate the duration in the variable name - READ_TIMEOUT_SECONDS. You can pass in seconds directly got getOne and specify the TimeUnit as seconds.

PR #2953, opened 2018-07-13
Option for segment metadata only upload

jenniferdai:
* Adding a new API that allows users to upload segments by only sending their segment metadata files
* Putting up for early feedback, tests are in progress

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
@@ -61,7 +61,7 @@ public class ControllerConf extends PropertiesConfiguration {
 
   // Defines the kind of storage and the underlying PinotFS implementation
   private static final String PINOT_FS_FACTORY_CLASS_PREFIX = "controller.storage.factory.class";
-  private static final String PINOT_FS_FACTORY_CLASS_LOCAL = "controller.storage.factory.class.file";
+  private static final String PINOT_FS_FACTORY_CLASS_LOCAL = "controller.storage.factory.class.localFS";
 
   private static final int DEFAULT_RETENTION_CONTROLLER_FREQUENCY_IN_SECONDS = 6 * 60 * 60; // 6 Hours.
   private static final int DEFAULT_VALIDATION_CONTROLLER_FREQUENCY_IN_SECONDS = 60 * 60; // 1 Hour.
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
@@ -17,11 +17,13 @@ package com.linkedin.pinot.controller;
 
 import com.linkedin.pinot.common.protocols.SegmentCompletionProtocol;
 import com.linkedin.pinot.common.utils.StringUtil;
+import com.linkedin.pinot.filesystem.LocalPinotFS;
 import java.io.File;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.List;
+import org.apache.commons.configuration.Configuration;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.PropertiesConfiguration;
 
@@ -57,6 +59,10 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final String SEGMENT_UPLOAD_TIMEOUT_IN_MILLIS = "controller.segment.upload.timeoutInMillis";
   private static final String REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS = "controller.realtime.segment.metadata.commit.numLocks";
 
+  // Defines the kind of storage and the underlying PinotFS implementation
+  private static final String PINOT_FS_FACTORY_CLASS_PREFIX = "controller.storage.factory.class";
+  private static final String PINOT_FS_FACTORY_CLASS_LOCAL = "controller.storage.factory.class.file";
+
   private static final int DEFAULT_RETENTION_CONTROLLER_FREQUENCY_IN_SECONDS = 6 * 60 * 60; // 6 Hours.
   private static final int DEFAULT_VALIDATION_CONTROLLER_FREQUENCY_IN_SECONDS = 60 * 60; // 1 Hour.
   private static final int DEFAULT_STATUS_CONTROLLER_FREQUENCY_IN_SECONDS = 5 * 60; // 5 minutes
@@ -74,6 +80,8 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final long DEFAULT_SEGMENT_UPLOAD_TIMEOUT_IN_MILLIS = 600_000L; // 10 minutes
   private static final int DEFAULT_REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS = 64;
 
+  private static final String DEFAULT_PINOT_FS_FACTORY_CLASS_LOCAL = LocalPinotFS.class.getName();
+
   public ControllerConf(File file) throws ConfigurationException {
     super(file);
   }
@@ -87,7 +95,15 @@ public class ControllerConf extends PropertiesConfiguration {
       return StringUtil.join("/", vip, "segments", tableName, URLEncoder.encode(segmentName, "UTF-8"));
     } catch (UnsupportedEncodingException e) {
       // Shouldn't happen
-      throw new AssertionError("UTF-8 encoding should always be supported", e);
+      throw new AssertionError("Encountered error while encoding in UTF-8 format", e);
+    }
+  }
+
+  public void setPinotFSFactoryClass(Configuration pinotFSFactoryClasses) {
+    setProperty(PINOT_FS_FACTORY_CLASS_LOCAL, DEFAULT_PINOT_FS_FACTORY_CLASS_LOCAL);
+
+    if (pinotFSFactoryClasses != null) {
+      pinotFSFactoryClasses.getKeys().forEachRemaining(key -> setProperty((String) key, pinotFSFactoryClasses.getProperty((String) key)));
     }
   }
 
@@ -209,6 +225,10 @@ public class ControllerConf extends PropertiesConfiguration {
   @Override
   public String toString() {
     return super.toString();
+  }
+
+  public Configuration getPinotFSFactoryClasses() {
+    return this.subset(PINOT_FS_FACTORY_CLASS_PREFIX);
   }
 
   public boolean getAcceptSplitCommit() {
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
@@ -61,7 +61,7 @@ public class ControllerConf extends PropertiesConfiguration {
 
   // Defines the kind of storage and the underlying PinotFS implementation
   private static final String PINOT_FS_FACTORY_CLASS_PREFIX = "controller.storage.factory.class";
-  private static final String PINOT_FS_FACTORY_CLASS_LOCAL = "controller.storage.factory.class.localFS";
+  private static final String PINOT_FS_FACTORY_CLASS_LOCAL = "controller.storage.factory.class.file";
 
   private static final int DEFAULT_RETENTION_CONTROLLER_FREQUENCY_IN_SECONDS = 6 * 60 * 60; // 6 Hours.
   private static final int DEFAULT_VALIDATION_CONTROLLER_FREQUENCY_IN_SECONDS = 60 * 60; // 1 Hour.
@@ -99,7 +99,7 @@ public class ControllerConf extends PropertiesConfiguration {
     }
   }
 
-  public void setPinotFSFactoryClass(Configuration pinotFSFactoryClasses) {
+  public void setPinotFSFactoryClasses(Configuration pinotFSFactoryClasses) {
     setProperty(PINOT_FS_FACTORY_CLASS_LOCAL, DEFAULT_PINOT_FS_FACTORY_CLASS_LOCAL);
 
     if (pinotFSFactoryClasses != null) {
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
@@ -253,23 +253,38 @@ public class PinotSegmentUploadRestletResource {
       @ApiParam(value = "Whether to enable parallel push protection") @DefaultValue("false") @QueryParam(FileUploadDownloadClient.QueryParameters.ENABLE_PARALLEL_PUSH_PROTECTION) boolean enableParallelPushProtection,
       @Context HttpHeaders headers, @Context Request request, @Suspended final AsyncResponse asyncResponse) {
     try {
-      asyncResponse.resume(uploadSegmentInternal(multiPart, null, enableParallelPushProtection, headers, request, null));
+      asyncResponse.resume(uploadSegmentInternal(multiPart, null, enableParallelPushProtection, headers, request, false));
     } catch (Throwable t) {
       asyncResponse.resume(t);
     }
   }
 
+  /**
+   * Will upload a Pinot segment without sending the file through the controller. If data is stored on a separate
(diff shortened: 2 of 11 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -56,6 +58,10 @@
   // protection is enabled. If the upload does not finish within the timeout, next upload can override the previous one.
   private static final String SEGMENT_UPLOAD_TIMEOUT_IN_MILLIS = "controller.segment.upload.timeoutInMillis";
 
+  // Defines the kind of storage and the underlying PinotFS implementation
+  private static final String PINOT_FS_FACTORY_CLASS_PREFIX = "controller.storage.factory.class";
+  private static final String PINOT_FS_FACTORY_CLASS_LOCAL = "controller.storage.factory.class.file";
```
mcvsubbu: controller.storage.factory.class.file reads like this is a file name.

How about "controller.storage.factory.class.localFS"?
Also, this is default, and does not need any additional config (or, are we planning to migrate the configs for segmentDir etc underneath this setting)?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -270,14 +287,14 @@ public void uploadSegmentAsJson(String segmentJsonStr,    // If segment is prese
...
-      asyncResponse.resume(uploadSegmentInternal(null, segmentJsonStr, enableParallelPushProtection, headers, request));
+      asyncResponse.resume(uploadSegmentInternal(null, segmentJsonStr, enableParallelPushProtection, headers, request, null));
     } catch (Throwable t) {
       asyncResponse.resume(t);
     }
   }
 
   private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, String segmentJsonStr,
-      boolean enableParallelPushProtection, HttpHeaders headers, Request request) {
+      boolean enableParallelPushProtection, HttpHeaders headers, Request request, FormDataMultiPart segmentMetadata) {
```
mcvsubbu: Why do we need two FormDataMultippart arguments?
We should know from 'request' whether this is a metadata update or segment update.  If we don't, then we can add a boolean saying that the multiPart has metadata instgead of segment data.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -253,7 +254,23 @@ public void uploadSegmentAsMultiPart(FormDataMultiPart multiPart,
...
+      asyncResponse.resume(t);
+    }
+  }
+
+  @POST
+  @ManagedAsync
+  @Produces(MediaType.APPLICATION_JSON)
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  @Path("/segmentmetadata")
+  @ApiOperation(value = "Upload a segment", notes = "Upload segment already in its final location by sending segment metadata file")
```
mcvsubbu: Please fix "value" and "notes".

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -253,7 +254,23 @@ public void uploadSegmentAsMultiPart(FormDataMultiPart multiPart,
...
+  public void uploadSegmentAsMultiPartWithMetadata(FormDataMultiPart multiPart,
+      @ApiParam(value = "Whether to enable parallel push protection") @DefaultValue("false") @QueryParam(FileUploadDownloadClient.QueryParameters.ENABLE_PARALLEL_PUSH_PROTECTION) boolean enableParallelPushProtection,
```
mcvsubbu: Are we protecting against parallel push of metadata?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -253,7 +254,23 @@ public void uploadSegmentAsMultiPart(FormDataMultiPart multiPart,
...
       @Context HttpHeaders headers, @Context Request request, @Suspended final AsyncResponse asyncResponse) {
     try {
-      asyncResponse.resume(uploadSegmentInternal(multiPart, null, enableParallelPushProtection, headers, request));
+      asyncResponse.resume(uploadSegmentInternal(multiPart, null, enableParallelPushProtection, headers, request, null));
+    } catch (Throwable t) {
+      asyncResponse.resume(t);
+    }
+  }
+
+  @POST
```
mcvsubbu: Please add a block comment here indicating the circumstances when a client would resort to this method (of pushing only segment metadata)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -328,9 +346,26 @@ private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, Strin
             throw new ControllerApplicationException(LOGGER,
                 "Failed to get segment fetcher for download URI: " + downloadURI, Response.Status.BAD_REQUEST);
           }
-
-          // Download segment tar file to local
-          segmentFetcher.fetchSegmentToLocal(downloadURI, tempTarredSegmentFile);
+          if (segmentMetadata == null) {
+            // Download segment tar file to local
+            segmentFetcher.fetchSegmentToLocal(downloadURI, tempTarredSegmentFile);
```
mcvsubbu: You need the segmentFetcher only if metadata is null, right? Can you push that code inside the if?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -328,9 +346,26 @@ private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, Strin
...
+          } else {
+            // Read segment metadata file and directly use that information to update zk
```
mcvsubbu: How do we guard against a client bug (or worse, malicious client) that may send wrong metadata? (a design question, I suppose)

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
jenniferdai: no, this is only for different kinds of filesystems. For each fs we configure (for example, internally, we can configure dali), we will have one of these strings. Basically, this will jsut like the immutable map in segment fetcher

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: yes because the same race condition when adding to zk can happen (minion pulls, pbnj pushes, minion pushes finished segment)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: We decided during design review that we wouldn't take care of this case. Have some notes in the design document about this (basically the problem is difficult because without downloading segment, can always be injected)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -252,7 +253,23 @@ public void uploadSegmentAsMultiPart(FormDataMultiPart multiPart,
...
+  @ManagedAsync
+  @Produces(MediaType.APPLICATION_JSON)
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  @Path("/segmentmetadata")
+  @ApiOperation(value = "Upload a segment", notes = "Upload segment already in its final location by sending segment metadata file")
+  public void uploadSegmentAsMultiPartWithMetadata(FormDataMultiPart multiPart,
+      @ApiParam(value = "Whether to enable parallel push protection") @DefaultValue("false") @QueryParam(FileUploadDownloadClient.QueryParameters.ENABLE_PARALLEL_PUSH_PROTECTION) boolean enableParallelPushProtection,
+      @Context HttpHeaders headers, @Context Request request, @Suspended final AsyncResponse asyncResponse) {
+    try {
+      asyncResponse.resume(uploadSegmentInternal(null, null, enableParallelPushProtection, headers, request, multiPart));
```
mcvsubbu: we should check here that the uploadType is URI. If not. return 4xx. 
Also, if it is the supported uploadType, then we can call a different method that does only metadata upload?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -327,9 +345,26 @@ private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, Strin
...
+              throw new ControllerApplicationException(LOGGER, "Invalid multi-part form for segment metadata", Response.Status.BAD_REQUEST);
+            }
+            File segmentMetadataFile = null;
+            FormDataBodyPart segmentMetadataBodyPart = segmentMetadataMap.values().iterator().next().get(0);
+            try (InputStream inputStream = segmentMetadataBodyPart.getValueAs(InputStream.class);
+                OutputStream outputStream = new FileOutputStream(segmentMetadataFile)) {
+                IOUtils.copyLarge(inputStream, outputStream);
+            }
+            finally {
+              segmentMetadata.cleanup();
```
mcvsubbu: multiPart.cleanup

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -346,26 +381,37 @@ private SuccessResponse uploadSegmentInternal(FormDataMultiPart multiPart, Strin
...
       // in the segment in order to ensure the segment is not corrupted
-      TarGzCompressionUtils.unTar(tempTarredSegmentFile, tempSegmentDir);
-      File[] files = tempSegmentDir.listFiles();
-      Preconditions.checkState(files != null && files.length == 1);
-      File indexDir = files[0];
-
-      SegmentMetadata segmentMetadata = new SegmentMetadataImpl(indexDir);
-      String segmentName = segmentMetadata.getName();
-      String offlineTableName = TableNameBuilder.OFFLINE.tableNameWithType(segmentMetadata.getTableName());
+      if (tempTarredSegmentFile != null) {
```
mcvsubbu: Are we doing this twice in some cases?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -412,21 +458,32 @@ private void uploadSegment(File indexDir, SegmentMetadata segmentMetadata, File
...
-      throw new ControllerApplicationException(LOGGER, "Quota check failed for segment: " + segmentName + " of table: " + offlineTableName + ", reason: "
-          + e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR);
-    }
-    if (!quotaResponse.isSegmentWithinQuota) {
-      throw new ControllerApplicationException(LOGGER,
-          "Quota check failed for segment: " + segmentName + " of table: " + offlineTableName + ", reason: "
-              + quotaResponse.reason, Response.Status.FORBIDDEN);
+    // Check quota when segment file is locally downloaded
+    // For metadata-only upload, we have no way to check quota because we don't send the file itself and
+    // we can't access length of zipped files in a remote filesystem
```
mcvsubbu: so then maybe the fs can have a getLength API as optional implementation?

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/PinotFSFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+
+/**
+ * Initializes the PinotFS implementation.
+ */
+public class PinotFSFactory {
+  public static final Logger LOGGER = LoggerFactory.getLogger(PinotFSFactory.class);
+  private static final PinotFSFactory INSTANCE = new PinotFSFactory();
+
+  private PinotFSFactory() {
```
mcvsubbu: Create the factory with the config

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/PinotFSFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+  }
+
+  public static PinotFSFactory getInstance() {
+    return INSTANCE;
+  }
+
+  public PinotFS init(Configuration config, URI uri) throws ClassNotFoundException, IllegalAccessException, InstantiationException {
```
mcvsubbu: method should be create(uri) not init(config, uri)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: Agree, can add that in a future one

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -252,7 +253,38 @@ public void uploadSegmentAsMultiPart(FormDataMultiPart multiPart,
...
+   * @param asyncResponse
+   */
+  @POST
+  @ManagedAsync
+  @Produces(MediaType.APPLICATION_JSON)
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  @Path("/segmentmetadata")
+  @ApiOperation(value = "Upload a segment by sending metadata", notes = "Does not download segment file")
+  public void uploadSegmentAsMultiPartWithMetadata(FormDataMultiPart multiPart,
+      @ApiParam(value = "Whether to enable parallel push protection") @DefaultValue("false") @QueryParam(FileUploadDownloadClient.QueryParameters.ENABLE_PARALLEL_PUSH_PROTECTION) boolean enableParallelPushProtection,
```
mayankshriv: Why is enabling parallel push protection a param for POST? What happens if multiple posts are called concurrently and some of them enable this param and others disable it? Would it make more sense to have that as a controller/table config instead of per POST request?

jenniferdai: We've done this for the tar segment file upload as well as the JSON file upload. The idea is that we don't do the extra zk write (that would protect against mutant metadata) unless clients send that header to protect against zk problems. We can change this in a future pr, but for now, this is consistent with other segment uploads

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -415,21 +470,32 @@ private void uploadSegment(File indexDir, SegmentMetadata segmentMetadata, File
...
+      if (!quotaResponse.isSegmentWithinQuota) {
+        throw new ControllerApplicationException(LOGGER,
+            "Quota check failed for segment: " + segmentName + " of table: " + offlineTableName + ", reason: " + quotaResponse.reason, Response.Status.FORBIDDEN);
+      }
     }
 
+    uploadSegmentMetadata(downloadUrl, provider, tempTarredSegmentFile, segmentMetadata, headers, enableParallelPushProtection);
+  }
+
+  private void uploadSegmentMetadata(String downloadUrl, FileUploadPathProvider provider, File tempTarredSegmentFile, SegmentMetadata segmentMetadata,
```
mayankshriv: Seems like this method is called for both data and metadata push flows. I think the reason is to protect concurrency, if you could add comments explaining the reasoning behind it, that would be great.

PR #2955, opened 2018-07-13
Added dummy idealstate-updater utility

mcvsubbu:
We can write custom code to apply any changes we want to idealstate on a live cluster


pinot-tools/src/main/java/com/linkedin/pinot/tools/PinotIdealstateChanger.java
```diff
@@ -0,0 +1,48 @@
...
+          @Override
+          public IdealState apply(@Nullable IdealState input) {
+            return applyChangesToIdealState(input);
+          }
+        },
+    RetryPolicies.exponentialBackoffRetryPolicy(5, 100, 1.2));
+  }
+
+  private IdealState applyChangesToIdealState(IdealState input) {
+    // Add code here as needed to change idealstate
```
sunithabeeram: Could we print a message here by default - so someone just trying out the code knows whats happening (or not)?

PR #2967, opened 2018-07-18
Emitting storage quota utilization

jackjlli:
This PR emits storage quota utilization.
If the controller is the leader, emit the real percentage; otherwise, emit 0.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/StorageQuotaChecker.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/StorageQuotaChecker.java
@@ -134,8 +134,8 @@ public class StorageQuotaChecker {
         tableSubtypeSize.reportedSizeInBytes);
 
     // Only emit the real percentage of storage quota usage by lead controller, otherwise emit 0L.
-    if (_pinotHelixResourceManager.isLeader()) {
-      long existingStorageQuotaUtilization = tableSubtypeSize.estimatedSizeInBytes / allowedStorageBytes * 100;
+    if (_pinotHelixResourceManager.isLeader() && allowedStorageBytes != 0L) {
+      long existingStorageQuotaUtilization = tableSubtypeSize.estimatedSizeInBytes  * 100 / allowedStorageBytes;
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/StorageQuotaChecker.java
```diff
@@ -130,6 +133,15 @@ public QuotaCheckerResponse isSegmentStorageWithinQuota(@Nonnull File segmentFil
         tableSubtypeSize.estimatedSizeInBytes,
         tableSubtypeSize.reportedSizeInBytes);
 
+    // Only emit the real percentage of storage quota usage by lead controller, otherwise emit 0L.
+    if (_pinotHelixResourceManager.isLeader()) {
+      long existingStorageQuotaUtilization = tableSubtypeSize.estimatedSizeInBytes / allowedStorageBytes * 100;
```
sunithabeeram: We should be careful about allowedStorageBytes == 0. There is a < 0 check for that earlier in the code, but not for == 0. Please add a check to avoid div by zero errors.

pinot-controller/src/test/java/com/linkedin/pinot/controller/validation/StorageQuotaCheckerTest.java
```diff
@@ -51,8 +53,10 @@ public void setUp() {
     _quotaConfig = mock(QuotaConfig.class);
     _controllerMetrics = mock(ControllerMetrics.class);
     _validationConfig = mock(SegmentsValidationAndRetentionConfig.class);
+    _pinotHelixResourceManager = mock(PinotHelixResourceManager.class);
     when(_tableConfig.getValidationConfig()).thenReturn(_validationConfig);
     when(_validationConfig.getReplicationNumber()).thenReturn(2);
+    when(_pinotHelixResourceManager.isLeader()).thenReturn(true);
```
sunithabeeram: Can you also add a test to check that the emitted % is correct?

jackjlli: Test added in the latest push.

PR #2968, opened 2018-07-18
Adding https option to hadoop segment upload

jenniferdai:
* Clients can declare enable.https in their job.properties file. If this parameter is declared, https will be used by default for all segment upload requests and tableconfig/schema get requests.

### DIFF
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/ControllerRestApi.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/ControllerRestApi.java
@@ -16,6 +16,7 @@
 package com.linkedin.pinot.hadoop.job;
 
 import com.linkedin.pinot.common.config.TableConfig;
+import com.linkedin.pinot.common.utils.ClientSSLContextGenerator;
 import com.linkedin.pinot.common.utils.FileUploadDownloadClient;
 import com.linkedin.pinot.common.utils.SimpleHttpResponse;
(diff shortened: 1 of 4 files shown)


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/ControllerRestApi.java
```diff
@@ -32,21 +35,39 @@
   private static final Logger LOGGER = LoggerFactory.getLogger(ControllerRestApi.class);
   private final List<PushLocation> _pushLocations;
   private final String _tableName;
+  private SSLContext _sslContext = null;
```
snleee: Is there a particular reason to explicitly set this to `null` initially?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -86,6 +87,10 @@ public SegmentCreationJob(String jobName, Properties properties) throws Exceptio
       _port = Integer.parseInt(portString);
     }
     _tableName = _properties.getProperty(JobConfigConstants.SEGMENT_TABLE_NAME);
+    String enableHttps = _properties.getProperty(JobConfigConstants.ENABLE_HTTPS);
+    if (enableHttps != null) {
```
snleee: Can you add `LOGGER.info` below where we print out all configs? (after `*********************`)

PR #2970, opened 2018-07-18
[PINOT-6677] Clean up any existing segment dir before replacing a segment

mcvsubbu:
In LLC, it is possible that a server tries to commit a segment, the commit fails, and eventually
another server commits the same segment with a different end offset. In that case, the
server that originally tried to commit the segment will get a DISCARD on for the consumer,
and eventually a state transition to ONLINE. In that case, we should clean out the original
segment before replacing it.


mcvsubbu:
@sunithabeeram  correct, it is always called in the helix state transition thread : either CONSUMING to ONLINE or OFFLINE to ONLINE. These cannot happen simultaneously. Further, helix also serializes calls to the same resource-partition.

PR #2972, opened 2018-07-19
Log uncompressed/compressed segment sizes during segment creation time

jackjlli:
This PR logs uncompressed/compressed segment sizes during segment creation time.

### DIFF
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
@@ -16,6 +16,7 @@
 package com.linkedin.pinot.hadoop.job.mapper;
 
 import com.linkedin.pinot.common.data.Schema;
+import com.linkedin.pinot.common.utils.DataSize;
 import com.linkedin.pinot.common.utils.TarGzCompressionUtils;
 import com.linkedin.pinot.core.data.readers.CSVRecordReaderConfig;
 import com.linkedin.pinot.core.data.readers.FileFormat;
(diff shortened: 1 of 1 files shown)


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -220,13 +221,20 @@ protected String createSegment(String dataFilePath, Schema schema, Integer seqId
...
+      File localSegmentFile = new File(_localDiskSegmentDirectory, segmentName);
+      String localSegmentPath = localSegmentFile.getAbsolutePath();
 
       String localTarPath = _localDiskSegmentTarPath + "/" + segmentName + JobConfigConstants.TARGZ;
       LOGGER.info("Trying to tar the segment to: {}", localTarPath);
       TarGzCompressionUtils.createTarGzOfDirectory(localSegmentPath, localTarPath);
       String hdfsTarPath = _localHdfsSegmentTarPath + "/" + segmentName + JobConfigConstants.TARGZ;
 
+      // Log segment size.
+      long uncompressedSegmentSize = localSegmentFile.length();
```
jenniferdai: If we want all the files in the directory, we might have to recursively go through the directory and get the size of all files. Unfortunately, java's length() only handles length of an individual file. We have some examples of recursive directory iteration when we gather input paths.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -220,13 +221,20 @@ protected String createSegment(String dataFilePath, Schema schema, Integer seqId
...
+      long compressedSegmentSize = new File(_localDiskSegmentTarPath, segmentName + JobConfigConstants.TARGZ).length();
```
jenniferdai: Can we use the above constant (hdfsTarPath) instead of constructing it again?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
jackjlli: Seems `FileUtils.sizeOfDirectory()` would be the best option. Thanks for pointing out! 👍

PR #2974, opened 2018-07-19
Adding table name and segment name headers to segment push

jenniferdai:
* Segment upload will print table name and segment name headers, can help to debug network latency.

### DIFF
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentTarPushJob.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentTarPushJob.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.hadoop.job;
 
+import com.linkedin.pinot.common.utils.CommonConstants;
 import com.linkedin.pinot.common.utils.FileUploadDownloadClient;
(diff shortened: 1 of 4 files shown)


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentTarPushJob.java
```diff
@@ -33,6 +37,10 @@
   private String _segmentPath;
   private String[] _hosts;
   private int _port;
+  private String _tableName;
+
+  public static final String HDR_SEGMENT_NAME = "Pinot-Segment-Name";
```
mcvsubbu: Use the same final Strings that the controller uses

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentUriPushJob.java
```diff
@@ -34,6 +38,10 @@
   private String _segmentPath;
   private String[] _hosts;
   private int _port;
+  private String _tableName;
+
+  public static final String HDR_SEGMENT_NAME = "Pinot-Segment-Name";
```
mcvsubbu: same here

PR #2975, opened 2018-07-19
Fixing setAdditionalProperties bug

jenniferdai:
* Internally, we override this method, so moving the conditional into the method for backwards compatibility handling.


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -211,6 +206,11 @@ public void run() throws Exception {
   }
 
   protected void setAdditionalJobProperties(Job job) throws Exception {
+    if (_hosts == null && _port == 0) {
+      LOGGER.warn("Unable to set TableConfig-dependent properties. Please set {} and {}", JobConfigConstants.PUSH_TO_HOSTS, JobConfigConstants.PUSH_TO_PORT);
```
jackjlli: It'd be good to say something like `Please set host {} and port {}`

PR #2977, opened 2018-07-19
Remove leadership check when setting up metadata for new table

npawar:
Adding a new table can be done by any controller, not just the leader. We should not check for leadership when adding metadata in this case, else we will end up not creating metadata but updating the ideal state. This results in an incorrect setup for an hour until the validation anager fixes it

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -234,6 +234,10 @@ public class PinotLLCRealtimeSegmentManager {
       }
     }
     _flushThresholdUpdateManager.clearFlushThresholdUpdater(tableConfig);
+    if (!isConnected()) {
+      throw new RuntimeException(
+          "Lost connection while setting up new table " + tableConfig.getTableName() + " isConnected=" + isConnected());
+    }
     IdealState idealState = setupTable(tableConfig, emptyIdealState, partitionCount);
     setTableIdealState(tableConfig.getTableName(), idealState);
  }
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -572,12 +578,15 @@ protected boolean createNewSegmentMetadataZNRecord(TableConfig realtimeTableConf
     final String newZnodePath =
         ZKMetadataProvider.constructPropertyStorePathForSegment(realtimeTableName, newSegmentNameStr);
 
-    if (!isConnected() || !isLeader()) {
-      // We can potentially log a different value than what we saw ....
-      LOGGER.warn("Lost leadership while committing new segment metadata for {} for table {}: isLeader={}, isConnected={}",
-          newSegmentNameStr, rawTableName, isLeader(), isConnected());
-      _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_NOT_LEADER, 1L);
-      return false;
+    if (!isNewTableSetup) {
```
kishoreg: I am not able to follow the need for this new boolean. It looks like this method is called from only one place. If thats the case, why don't we remove the isLeader check completely

kishoreg: Saw the other place where its called from but I still don't get why the check is not done by the caller before invoking this method.
 Are we using isNewTableSetup as a proxy for isLeader() check ?

npawar: The intention is to do it just before we write the metadata to the property store, which happens to be the last step of the method

PR #2981, opened 2018-07-20
Fix a bug in OffHeapStarTreeBuilder where file is not properly copied from data file to temp file

Jackie-Jiang:
Also add extra check to ensure that

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/startree/OffHeapStarTreeBuilder.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/startree/OffHeapStarTreeBuilder.java
@@ -30,9 +30,9 @@ import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
 import java.io.BufferedOutputStream;
 import java.io.DataOutputStream;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.RandomAccessFile;
 import java.nio.ByteOrder;
 import java.nio.channels.FileChannel;
 import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/startree/OffHeapStarTreeBuilder.java
```diff
@@ -645,9 +645,12 @@ private void constructStarTree(TreeNode node, int startDocId, int endDocId, int
...
       File tempFile = new File(_tempDir, startDocId + "_" + endDocId + ".unique.tmp");
-      try (FileChannel src = new FileInputStream(_dataFile).getChannel();
-          FileChannel dest = new FileOutputStream(tempFile).getChannel()) {
-        dest.transferFrom(src, startDocId * _docSize, tempBufferSize);
+      try (FileChannel src = new RandomAccessFile(_dataFile, "r").getChannel();
+          FileChannel dest = new RandomAccessFile(tempFile, "rw").getChannel()) {
+        long numBytesTransferred = src.transferTo(startDocId * _docSize, tempBufferSize, dest);
+        Preconditions.checkState(numBytesTransferred == tempBufferSize,
+            "Error transferring data from data file to temp file, transfer size mis-match");
+        dest.force(false);
```
kishoreg: what is the idea behind calling force and especially with false?

Jackie-Jiang: Ensure the data is flushed to disk (false means we don't enforce metadata to be flushed).

kishoreg: this is in try catch block, which means close would be called at the end of it and that would internally call flush right?. btw, this if for my own understanding. I know having this additional line is probably harmless.

Jackie-Jiang: @kishoreg Based on the javadoc, seems there is no guarantee on that.

PR #2983, opened 2018-07-22
Replace LBufferAPI in star-tree with PinotDataBuffer

Jackie-Jiang:
NOTE:
For temporary buffer, always use native order
For star-tree file, always use little-endian for backward compatibility

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/startree/DimensionBuffer.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/startree/DimensionBuffer.java
@@ -15,47 +15,44 @@
  */
 package com.linkedin.pinot.core.startree;
 
-import com.linkedin.pinot.core.segment.creator.impl.V1Constants;
+import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
 import java.nio.ByteBuffer;
-import java.nio.ByteOrder;
-import java.nio.IntBuffer;
 
 
 public class DimensionBuffer {
-  private static final ByteOrder NATIVE_ORDER = ByteOrder.nativeOrder();
-
-  private final int _numDimensions;
-  private final IntBuffer _intBuffer;
-
-  public DimensionBuffer(int numDimensions) {
-    _numDimensions = numDimensions;
-    _intBuffer = ByteBuffer.wrap(new byte[numDimensions * V1Constants.Numbers.INTEGER_SIZE]).asIntBuffer();
-  }
-
-  public DimensionBuffer(byte[] bytes) {
-    _numDimensions = bytes.length / V1Constants.Numbers.INTEGER_SIZE;
-    // NOTE: the byte array is returned from Unsafe which uses native order
-    _intBuffer = ByteBuffer.wrap(bytes).order(NATIVE_ORDER).asIntBuffer();
-  }
+  private final ByteBuffer _buffer;
 
   public static DimensionBuffer fromBytes(byte[] bytes) {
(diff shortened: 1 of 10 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/startree/DimensionBuffer.java
```diff
@@ -15,47 +15,42 @@
...
+    return _buffer.array();
   }
 
-  @SuppressWarnings({"EqualsWhichDoesntCheckParameterClass", "CheckStyle"})
-  @Override
-  public boolean equals(Object obj) {
-    DimensionBuffer that = (DimensionBuffer) obj;
-    for (int i = 0; i < _numDimensions; i++) {
-      if (_intBuffer.get(i) != that._intBuffer.get(i)) {
+  public boolean hasSameBytes(byte[] dimensionBytes) {
```
mayankshriv: If the assumption is that both the buffers should be of same length, please document it.

pinot-core/src/main/java/com/linkedin/pinot/core/startree/DimensionBuffer.java
```diff
@@ -66,8 +61,8 @@ public boolean equals(Object obj) {
   public String toString() {
     StringBuilder builder = new StringBuilder("[");
     String delimiter = "";
-    for (int i = 0; i < _numDimensions; i++) {
-      builder.append(delimiter).append(_intBuffer.get(i));
+    while (_buffer.remaining() > 0) {
```
mayankshriv: Seems like stateful toString(), which is incorrect. Please fix.

pinot-core/src/main/java/com/linkedin/pinot/core/startree/StarTreeDataTable.java
```diff
@@ -211,9 +192,9 @@ public boolean hasNext() {
       public Pair<byte[], byte[]> next() {
         byte[] dimensionBytes = new byte[_dimensionSize];
         byte[] metricBytes = new byte[_metricSize];
-        ByteBuffer byteBuffer = _dataBuffer.toDirectByteBuffer(_currentIndex++ * _docSize, (int) _docSize);
-        byteBuffer.get(dimensionBytes);
-        byteBuffer.get(metricBytes);
+        ByteBuffer byteBuffer = _dataBuffer.toDirectByteBuffer(_currentIndex++ * _docSizeLong, (int) _docSizeLong);
```
mayankshriv: Seems we changed _docSize name to _docSizeLong, and then casting it here to int. I am sure the logic is correct, but makes this part of the code a bit confusing.

pinot-core/src/main/java/com/linkedin/pinot/core/startree/DimensionBuffer.java
Jackie-Jiang: Added.

pinot-core/src/main/java/com/linkedin/pinot/core/startree/DimensionBuffer.java
Jackie-Jiang: Good catch

pinot-core/src/main/java/com/linkedin/pinot/core/startree/StarTreeDataTable.java
Jackie-Jiang: Fixed

PR #2986, opened 2018-07-24
Verify hybrid table contains the same table configs in its both subtype tables

jackjlli:
Previously for a hybrid table, it's allowable to have different metadata in table configs, which makes the time column name and type could be different. This leads to an issue that if one single sub-table is queried, the result will be returned. If a query uses raw table name, which means querying on both sub-tables, the result would be empty. Thus, in order to avoid this, we should validate the table config when we upload/update it for the hybrid table use case.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -439,7 +439,7 @@ public class PinotTableRestletResource {
 
     String newTimeColumnType = newSegmentConfig.getTimeType();
     String existingTimeColumnType = SegmentConfigToCompare.getTimeType();
-    if (!existingTimeColumnType.equals(newTimeColumnType)) {
+    if (!existingTimeColumnType.equalsIgnoreCase(newTimeColumnType)) {
       throw new PinotHelixResourceManager.InvalidTableConfigException(
           String.format("Time column types are different! Existing time column type: %s. New time column type: %s",
               existingTimeColumnType, newTimeColumnType));
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -420,7 +420,9 @@ public class PinotTableRestletResource {
 
     // Check if it is a hybrid table or not. If not, there's no need to verify anything.
     if (tableConfigToCompare == null) {
-      LOGGER.info("Table: {} is not a hybrid table. Skip validating table configs.", rawTableName);
+      LOGGER.info(
+          "Table: {} is not a hybrid table. Skipping consistency check across realtime and offline parts of the table.",
+          rawTableName);
       return;
     }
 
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -403,49 +403,48 @@ private void ensureMinReplicas(TableConfig config) {
...
       if (_pinotHelixResourceManager.hasRealtimeTable(rawTableName)) {
-        realtimeTableConfig = _pinotHelixResourceManager.getRealtimeTableConfig(rawTableName);
+        tableConfigToCompare = _pinotHelixResourceManager.getRealtimeTableConfig(rawTableName);
       }
     }
 
     // Check if it is a hybrid table or not. If not, there's no need to verify anything.
-    if (offlineTableConfig == null || realtimeTableConfig == null) {
+    if (tableConfigToCompare == null) {
+      LOGGER.info("Table: {} is not a hybrid table. Skip validating table configs.", rawTableName);
```
mcvsubbu: The log message is misleading. We are not skipping validation of table configs. We don't have anything to check for consistency. I suggest making the log message as "Skipping consistency check across realtime and offline parts of the table" or you can just remove the log message altogether.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -403,49 +403,48 @@ private void ensureMinReplicas(TableConfig config) {
...
-    if (!realtimeTimeColumnName.equalsIgnoreCase(offlineTimeColumnName)) {
-      throw new PinotHelixResourceManager.InvalidTableConfigException(String.format(
-          "Time column names are different! Real-time table time column name: %s. Offline table time column name: %s",
-          realtimeTimeColumnName, offlineTimeColumnName));
+    SegmentsValidationAndRetentionConfig newSegmentConfig = newTableConfig.getValidationConfig();
+    SegmentsValidationAndRetentionConfig SegmentConfigToCompare = tableConfigToCompare.getValidationConfig();
+
+    String newTimeColumnName = newSegmentConfig.getTimeColumnName();
+    String existingTimeColumnName = SegmentConfigToCompare.getTimeColumnName();
+    if (!existingTimeColumnName.equalsIgnoreCase(newTimeColumnName)) {
```
mcvsubbu: Column names are case sensitive, so it should be .equals, not .equalsIgnoreCase

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
jackjlli: Great! Thanks for the suggestion. I think it's good to keep this message log, since this message wouldn't be printed quite frequent, and it's beneficial for us to know how the request goes. I've adjusted the message in the latest push.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -398,6 +400,55 @@ private void ensureMinReplicas(TableConfig config) {
...
+    String existingTimeColumnName = SegmentConfigToCompare.getTimeColumnName();
+    if (!existingTimeColumnName.equals(newTimeColumnName)) {
+      throw new PinotHelixResourceManager.InvalidTableConfigException(
+          String.format("Time column names are different! Existing time column name: %s. New time column name: %s",
+              existingTimeColumnName, newTimeColumnName));
+    }
+
+    String newTimeColumnType = newSegmentConfig.getTimeType();
+    String existingTimeColumnType = SegmentConfigToCompare.getTimeType();
+    if (!existingTimeColumnType.equals(newTimeColumnType)) {
```
mcvsubbu: this can be equalsIgnoreCase, unless you are using to get the TimeUtils and turn it into an enum before comparison.

PR #2990, opened 2018-07-25
Print message in TableSizeReader only when num of errors > 0

jackjlli:
Currently it prints the message even when there's no error. 
Adding this if statement can help reduce most of the noises in controller log.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableSizeReader.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableSizeReader.java
@@ -193,8 +193,8 @@ public class TableSizeReader {
       }
       // after iterating over all servers update summary reported and estimated size of the segment
       if (errors != segmentSizes.serverInfo.size()) {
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableSizeReader.java
```diff
@@ -193,9 +193,11 @@ public TableSubTypeSizeDetails getTableSubtypeSize(String tableNameWithType, int
       }
       // after iterating over all servers update summary reported and estimated size of the segment
       if (errors != segmentSizes.serverInfo.size()) {
-        // atleast one server reported size for this segment
-        LOGGER.info("Could not get size for segment {} from {} servers. Using segmentLevelMax {} to estimate the size",
-            segmentEntry.getKey(), errors, segmentLevelMax);
+        if (errors > 0) {
+          // at least one server reported size for this segment
```
sunithabeeram: Move the comment outside of the if-block.

PR #2992, opened 2018-07-25
Change SegmentMetadataImpl.getPropertiesConfiguration() so that we don't need to cache the properties configuration

Jackie-Jiang:


Jackie-Jiang:
@kishoreg Is this good enough? Or should I also remove the reference to properties configuration as well?

PR #2995, opened 2018-07-25
Adding server side download methods for azure

jenniferdai:
* Server side fetcher methods for azure fs to support server download


pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/AzureSegmentFetcher.java
```diff
@@ -0,0 +1,86 @@
...
+    Configuration azureConfigs = config.subset(AZURE_KEY);
+    // The ADL account id. Example: {@code mystore.azuredatalakestore.net}.
+    String account = azureConfigs.getString(ACCOUNT_KEY);
+    // The endpoint that should be used for authentication.
+    // Usually of the form {@code https://login.microsoftonline.com/<tenant-id>/oauth2/token}.
+    String authEndpoint = azureConfigs.getString(AUTH_ENDPOINT_KEY);
+    // The clientId used to authenticate this application
+    String clientId = azureConfigs.getString(CLIENT_ID_KEY);
+    // The secret key used to authenticate this application
+    String clientSecret = azureConfigs.getString(CLIENT_SECRET_KEY);
```
mayankshriv: Should the key be encrypted?

pinot-common/pom.xml
```diff
@@ -108,6 +108,10 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
+    <dependency>
```
kishoreg: why are we adding this to common?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/AzureSegmentFetcher.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.commons.configuration.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Azure implementation for the Pinot Segment Fetcher interface, used by the server to download segments and also
+ * used by controller when we want the full segment payload to go through the controller during upload.
+ */
+public class AzureSegmentFetcher implements SegmentFetcher {
```
kishoreg: Why do we need separate implementation of SegmentFetcher for Azure. Can we not hide all storage specific details under FS implementation.

kishoreg:
At a high level, having segment fetcher and FS interface is redundant. We need to make SegmentFetcher work on top of FS interface and pass an instance of FS at runtime.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/GenericSegmentFetcher.java
```diff
@@ -0,0 +1,58 @@
...
+import com.linkedin.pinot.filesystem.PinotFSFactory;
+import java.io.File;
+import java.net.URI;
+import java.util.Collections;
+import java.util.Set;
+import org.apache.commons.configuration.Configuration;
+
+
+public class GenericSegmentFetcher implements SegmentFetcher {
+  private PinotFS _pinotFS;
```
bschnack: Do you want to make this final?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -44,45 +44,57 @@ public static SegmentFetcherFactory getInstance() {
...
   }
 
-  public SegmentFetcher getSegmentFetcherBasedOnURI(String uri) throws URISyntaxException {
+  public SegmentFetcher getSegmentFetcherBasedOnURI(String uri) throws Exception {
     String protocol = new URI(uri).getScheme();
-    return _segmentFetcherMap.get(protocol);
+    SegmentFetcher segmentFetcher = _segmentFetcherMap.get(protocol);
+    if (segmentFetcher instanceof GenericSegmentFetcher) {
+      LOGGER.info("Initializing pinotFS for segment fetcher with uri {}", uri);
+      if (_pinotFSConfig.getProperty("file") == null) {
```
bschnack: Is the else statement expected? If not, would you like to log it?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/GenericSegmentFetcher.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public void init(Configuration config) {
+  }
+
+  public GenericSegmentFetcher() {
```
mayankshriv: Why do we need this constructor?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/GenericSegmentFetcher.java
```diff
@@ -0,0 +1,58 @@
...
+
+  }
+
+  public GenericSegmentFetcher(URI uri, Configuration configuration) throws Exception {
```
mayankshriv: Given that there's an init that takes the configuration, why do we have a constructor that takes configuration as well? What's the contract of the interface, initializations in constructor or in init?

Also, if this is a generic fetcher, why does it need a uri in the constructor?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/GenericSegmentFetcher.java
```diff
@@ -0,0 +1,58 @@
...
+    PinotFSFactory factory = new PinotFSFactory(configuration);
+    _pinotFS = factory.create(uri);
```
mayankshriv: Should this class create and hold the FS upfront? Or should it be flexible enough that it can use the appropriate FS per invocation? If latter, we would need to avoid creating FS object per invocation of fetch.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
mayankshriv: What does the key "file" stand for? Could you use a private static final with meaningful name instead?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -44,45 +44,57 @@ public static SegmentFetcherFactory getInstance() {
...
+        _pinotFSConfig.addProperty("file", LocalPinotFS.class.getName());
```
mayankshriv: Modifying properties within this function seems incorrect. If this is a default property that is expected to be present, lets add it during initialization and not here.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/GenericSegmentFetcher.java
jenniferdai: Discussed this offline, this isn't possible at this point if we want backwards compatibility

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/GenericSegmentFetcher.java
jenniferdai: Talked about this offline. This is needed for all fs related classes. Http/Https/etc will all be separate.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/GenericSegmentFetcher.java
jenniferdai: This is needed because in the init for SegmentFetcher, it requires an instance of GenericSegmentFetcher, so it requires the default constructor

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -44,45 +43,50 @@ public static SegmentFetcherFactory getInstance() {
...
     }
+    _pinotFSConfig = pinotFSConfig;
   }
 
   public boolean containsProtocol(String protocol) {
     return _segmentFetcherMap.containsKey(protocol);
   }
 
-  public SegmentFetcher getSegmentFetcherBasedOnURI(String uri) throws URISyntaxException {
+  public SegmentFetcher getSegmentFetcherBasedOnURI(String uri) throws Exception {
```
mayankshriv: The change seems functionally equivalent, why did we need to repalce URISyntaxException to Exception?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -44,45 +43,50 @@ public static SegmentFetcherFactory getInstance() {
...
     String protocol = new URI(uri).getScheme();
-    return _segmentFetcherMap.get(protocol);
+    SegmentFetcher segmentFetcher = _segmentFetcherMap.get(protocol);
```
mayankshriv: We could probably just return _segmentFetcherMap.get(protocol).

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/PinotFSFactory.java
```diff
@@ -26,28 +28,44 @@
  * This factory class initializes the PinotFS class. It creates a PinotFS object based on the URI found.
  */
 public class PinotFSFactory {
-  public static final Logger LOGGER = LoggerFactory.getLogger(PinotFSFactory.class);
-  private static Configuration _schemeConfig;
+  private static final Logger LOGGER = LoggerFactory.getLogger(PinotFSFactory.class);
```
mayankshriv: Could we add unit test for this class?

PR #2996, opened 2018-07-25
Freeing up the reference to PropertiesConfiguration after loading seg…

kishoreg:
…ment metadata

We were holding on to the propertiesConfiguration object after loading the segment metadata which consumed extra memory. Even though the overhead is quite small for each segment, it adds up. In one of the environment with 25k segments, this attributed to 3G. With this change, the memory footprint reduced to ~250mb. This is lot of saving not just in terms of memory usage but also impact on GC since these are long lived and get moved around every time there is a major compaction.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
@@ -65,8 +65,8 @@ public class SegmentMetadataImpl implements SegmentMetadata {
   private static final Logger LOGGER = LoggerFactory.getLogger(SegmentMetadataImpl.class);
 
   private final File _indexDir;
-  private final PropertiesConfiguration _segmentMetadataPropertiesConfiguration;
   private final Map<String, ColumnMetadata> _columnMetadataMap;
+  private String _tableName;
   private String _segmentName;
   private final Set<String> _allColumns;
   private final Schema _schema;
@@ -90,6 +90,7 @@ public class SegmentMetadataImpl implements SegmentMetadata {
   private long _segmentStartTime;
   private long _segmentEndTime;
 
+ 
   /**
    * For segments on disk.
    * <p>Index directory passed in should be top level segment directory.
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
```diff
@@ -120,7 +121,7 @@ public SegmentMetadataImpl(File indexDir) throws IOException {
    */
   public SegmentMetadataImpl(RealtimeSegmentZKMetadata segmentMetadata, Schema schema) {
     _indexDir = null;
-    _segmentMetadataPropertiesConfiguration = new PropertiesConfiguration();
+    PropertiesConfiguration _segmentMetadataPropertiesConfiguration = new PropertiesConfiguration();
```
Jackie-Jiang: Remove the "_" prefix

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
```diff
@@ -173,7 +175,7 @@ public static PropertiesConfiguration getPropertiesConfiguration(File indexDir)
    *   <li> Start and End time. </li>
    * </ul>
    */
-  private void setTimeInfo() {
+  private void setTimeInfo(PropertiesConfiguration _segmentMetadataPropertiesConfiguration) {
```
Jackie-Jiang: Remove the "_" prefix

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
```diff
@@ -212,7 +214,7 @@ private void loadCreationMeta(File crcFile) throws IOException {
     return _allColumns;
   }
 
-  private void init() {
+  private void init(PropertiesConfiguration _segmentMetadataPropertiesConfiguration) {
```
Jackie-Jiang: Remove the "_" prefix

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
```diff
@@ -291,14 +295,14 @@ private void init() {
...
-      initStarTreeMetadata();
+      initStarTreeMetadata(_segmentMetadataPropertiesConfiguration);
     }
   }
 
   /**
    * Reads and initializes the star tree metadata from segment metadata properties.
    */
-  private void initStarTreeMetadata() {
+  private void initStarTreeMetadata(PropertiesConfiguration _segmentMetadataPropertiesConfiguration) {
```
Jackie-Jiang: Remove the "_" prefix

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/SegmentMetadataImpl.java
```diff
@@ -271,6 +273,8 @@ private void init() {
         _allColumns.add(columnName);
       }
     }
+    //set the table name
+    _tableName = _segmentMetadataPropertiesConfiguration.getString(V1Constants.MetadataKeys.Segment.TABLE_NAME);
```
mcvsubbu: The init() method seems to be private and called only from one of the constructors. Maybe we can make _tableName as final and initalize it in the constructor? Even move the logic from init to the constructor and make more vars final, but that will be a cleanup operation, so your choice.

PR #2999, opened 2018-07-26
[TE] Set the size of heatmap using other metric

jihaozh:
This PR allows the user to select a different metric than the primary metric to set the cell size of RCA heatmap.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-heatmap/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-heatmap/component.js
@@ -1,11 +1,16 @@
 import Component from '@ember/component';
-import { getProperties, computed } from '@ember/object';
+import { observer, getProperties, computed } from '@ember/object';
 import {
   toCurrentUrn,
   toBaselineUrn,
   appendFilters,
   isInverse,
-  isAdditive
+  isAdditive,
+  filterPrefix,
+  toMetricLabel,
+  stripTail,
+  extractTail,
+  appendTail
 } from 'thirdeye-frontend/utils/rca-utils';
 import {
   humanizeChange,
@@ -33,6 +38,7 @@ const ROOTCAUSE_MODE_MAPPING = {
   'Contribution to Overall Change': ROOTCAUSE_ROLLUP_MODE_CONTRIBUTION_TO_DIFF
 };
 
+
 export default Component.extend({
   //
   // external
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-heatmap/component.js
```diff
@@ -41,10 +47,16 @@ export default Component.extend({
 
   breakdowns: null, // {}
 
-  selectedUrn: null, // Set
+  selectedUrn: null, // ""
```
aaronucsd: Can we remove all these comments? Do we need them?

thirdeye/thirdeye-frontend/app/pods/components/rootcause-heatmap/component.js
```diff
@@ -56,6 +68,29 @@ export default Component.extend({
    */
   selectedMode: Object.keys(ROOTCAUSE_MODE_MAPPING)[1],
 
+  /**
+   * Selected option for heatmap size
+   */
+  sizeOptionSelected: computed('sizeMetricUrn', 'sizeOptions', {
```
aaronucsd: can you not use this style for computed? Especially if you're not doing anything with set(). Try this
```
  sizeOptionSelected: computed(
    'sizeMetricUrn', 'sizeOptions',
    function() {
      const { sizeMetricUrn, sizeOptions } =
        this.getProperties('sizeMetricUrn', 'sizeOptions');
      return sizeOptions.find(opt => opt.urn === sizeMetricUrn);
    }
  ),
```

aaronucsd: try

```
 sizeOptionSelected: computed(
    'sizeMetricUrn', 'sizeOptions',
    function() {
      const { sizeMetricUrn, sizeOptions } =
        this.getProperties('sizeMetricUrn', 'sizeOptions');
      return sizeOptions.find(opt => opt.urn === sizeMetricUrn);
    }
  ).readOnly(),
```

aaronucsd: https://m.alphasights.com/the-risks-you-take-when-you-use-ember-computed-oneway-f2c008a14404

jihaozh: this works. thanks for the example!

aaronucsd: Ideally we should use `oneWay` as readOnly will throw an error if someone (like a component) tries to mutate the object. OneWay makes a copy of the computed object and pass it on.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-heatmap/component.js
jihaozh: removed

PR #3005, opened 2018-07-27
Move kafka connection provider code to a separate class

npawar:
We plan to separate the StreamMetadataProvider and StreamConsumer. In order to achieve that, we would be extracting out SimpleMetadataProvider from SimpleConsumerWrapper. 
This change attempts to take out the code related to kafka connection that would be shared across the two SimpleWrappers, into a common class

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaConnectionProvider.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaConnectionProvider.java
@@ -1,0 +1,411 @@
+/**
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaConnectionProvider.java
```diff
@@ -0,0 +1,409 @@
...
+import kafka.javaapi.consumer.SimpleConsumer;
+import org.apache.kafka.common.protocol.Errors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Provides the kafka connection after ensuring that we're connected to the appropriate broker for consumption.
+ */
+public class KafkaConnectionProvider {
```
mcvsubbu: Maybe KafkaConnector? KafkaConnectionFSM? KafkaConnectionHandler? These are suggestions, and naming is not very important here. Can be modified easily without any externally visible changes.

npawar: I think I like KafkaConnectionHandler better, renamed

npawar:

PR #3008, opened 2018-07-28
Support UDF in aggregation function

Jackie-Jiang:
Add PQL compiler support for UDF in aggregation function
Aggregation function should always work on single argument, remove the redundant split on ',' (split on ',' will also break the UDF)

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
@@ -16,8 +16,10 @@
 package com.linkedin.pinot.pql.parsers.pql2.ast;
 
 import com.linkedin.pinot.common.request.AggregationInfo;
+import com.linkedin.pinot.common.request.transform.TransformExpressionTree;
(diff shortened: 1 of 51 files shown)


pinot-common/src/main/java/com/linkedin/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -71,37 +73,33 @@ public String getExpression() {
...
-    String identifier = null;
-    for (AstNode astNode : getChildren()) {
+    String expression;
+    // COUNT aggregation function always works on '*'
+    if (_name.equalsIgnoreCase("count")) {
+      expression = "*";
+    } else {
+      List<? extends AstNode> children = getChildren();
+      if (children == null || children.size() != 1) {
+        throw new Pql2CompilationException("Aggregation function expects exact 1 argument");
```
mayankshriv: We should think about removing this limitation. For example percentile(column, 95).

Jackie-Jiang: We should enhance our compiler so that it can identify UDF vs Aggregation, and directly compile the query into final format.

PR #3009, opened 2018-07-28
[PINOT-6736] Fix min size for hash-set in In/NotIn predicate impl

sunithabeeram:
In/NotIn predicate implementations typically use small hash sets. Profiling shows non-trivial amount of time spent in hash-set contains() calls. Testing showed that fixing a min hash set size improves the performance at little extra memory overhead. for one of the queries we see in production which averages about 580ms, this change reduced the latency down to 430ms.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/HashUtil.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/HashUtil.java
@@ -20,6 +20,22 @@ import java.nio.IntBuffer;
 
 
 public class HashUtil {
+
+  /** Tests show that even for smaller set sizes, setting the hash size to this min value
+   * improves performance at an insignificant increase of memory footprint.
+   */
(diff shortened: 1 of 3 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/HashUtil.java
```diff
@@ -20,6 +20,22 @@
...
+  public static int MIN_FASTUTIL_HASHSET_SIZE = 25;
+
+  /**
+   * Returns the min size for the fast-util hash-set given the expected size of
+   * values stored in the hash-set.
+   * @param expected the expected/actual number of values to be stored
+   * @return the optimal min value
+   */
+  public static int getMinHashSetSize(int expected) {
+    return Math.min(MIN_FASTUTIL_HASHSET_SIZE, expected);
```
Jackie-Jiang: Missed this pr, but this min should be max right?

PR #3010, opened 2018-07-28
[PINOT-6738] Deprioritize multivalue columns when evaluating filter expressions

sunithabeeram:
Phase 1 changes.
Changes include:
- Exposing datasource metadata and predicate evaluator through BaseFilterOperator. PredicateEvaluator is not currently used. Will be used in future revisions.
- Updating FilterOperatorUtils to return priority for scanning based single-value/multivalue columns. Single value cols have lower priority. 
- The optimization can be disabled per query using "useMVOpt" flag.

Testing done:
Ran a production query which has many multivalue columns in filter expression. Without the change, the latency is about 580ms on average, with the change we see it reduce to ~250ms


pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -79,7 +82,21 @@ int getPriority(BaseFilterOperator filterOperator) {
           return 3;
         }
         if (filterOperator instanceof ScanBasedFilterOperator) {
-          return 4;
+          DataSourceMetadata metadata = filterOperator.getDataSourceMetadata();
```
mayankshriv: Can we factor this out into a function? And also document the logic behind this heuristic, and the priority computation?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -79,7 +82,21 @@ int getPriority(BaseFilterOperator filterOperator) {
...
+          int basePriority = 4;
+          if (metadata == null) {
+            // shouldn't happen
+            return basePriority;
+          }
+          // lower priority for multivalue columns
+          if (! metadata.isSingleValue()) {
```
mayankshriv: Please format the changes before committing (referring to: ! metadata)

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -79,7 +82,21 @@ int getPriority(BaseFilterOperator filterOperator) {
...
+            // ensure priority for multivalue cols is higher than singvalue
+            // (accounting for the cardinality for single value)
+            basePriority *= DEFAULT_MAX_CARDINALITY;
```
mayankshriv: How did we decide upon the value for DEFAULT_MAX_CARDINALITY?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/data/source/ColumnDataSource.java
```diff
@@ -61,12 +62,12 @@ public ColumnDataSource(ColumnIndexContainer indexContainer, ColumnMetadata meta
   public ColumnDataSource(FieldSpec fieldSpec, int numDocs, int maxNumMultiValues, DataFileReader forwardIndex,
       InvertedIndexReader invertedIndex, MutableDictionary dictionary) {
     this(fieldSpec.getName(), fieldSpec.getDataType(), fieldSpec.isSingleValueField(), false, numDocs,
-        maxNumMultiValues, forwardIndex, invertedIndex, dictionary);
+        maxNumMultiValues, forwardIndex, invertedIndex, dictionary, -1);
```
mayankshriv: Use static final variable instead of -1, for readability. Also, unclear from the context why -1 is being passed for cardinality instead of the actual value?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/data/source/ColumnDataSource.java
```diff
@@ -52,7 +53,7 @@
   public ColumnDataSource(ColumnIndexContainer indexContainer, ColumnMetadata metadata) {
     this(metadata.getColumnName(), metadata.getDataType(), metadata.isSingleValue(), metadata.isSorted(),
```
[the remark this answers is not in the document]
sunithabeeram: That surprised me too. Will followup with Jackie.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
sunithabeeram: Done.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
sunithabeeram: Done.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
sunithabeeram: Its a guesstimate at this point - let me know if we can use a better value or a better way to determine the value.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/data/source/ColumnDataSource.java
sunithabeeram: I've updated the code to use UNKOWN_CARDINALITY for now mainly because it was not directly available for realtime segments. Will followup with Jackie on the different constructors for datasource and if/how we can get the cardinality in such cases.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -25,6 +25,10 @@
 
 
 public class FilterOperatorUtils {
+
+  // empirically derived max cardinality value; can be fine tuned/learnt in future iterations
+  private static final int DEFAULT_MAX_CARDINALITY = 1000_000;
```
mcvsubbu: Should we make this configurable (if we expect to tune this in future, or reset it to a different value depending on use case)

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -79,11 +83,47 @@ int getPriority(BaseFilterOperator filterOperator) {
...
+  private static int getScanBasedFilterPriority(BaseFilterOperator filterOperator) {
+    DataSourceMetadata metadata = filterOperator.getDataSourceMetadata();
+    int basePriority = 4;
+    if (metadata == null) {
+      // shouldn't happen
+      return basePriority;
+    }
+    // deprioritize multivalue columns
+    if (!metadata.isSingleValue()) {
+      // ensure priority for multivalue cols is higher than singvalue
```
mcvsubbu: typo

sunithabeeram:
Closing this as I had trouble updating it to the latest.

PR #3011, opened 2018-07-28
Verify table type when deleting table

jackjlli:
Since the param `tableName` may contain table type as suffix, this PR adds logic to check whether this case exists. 
++++++++
Use case 1:
tableName: TestTable, tableTypeStr: null
Both `deleteOfflineTable` and `deleteRealtimeTable ` will be called.

Use case 2:
tableName: TestTable_OFFLINE, tableTypeStr: null
Only `deleteOfflineTable` will be called.

Use case 3:
tableName: TestTable_OFFLINE, tableTypeStr: OFFLINE
Only `deleteOfflineTable` will be called.

Use case 4:
tableName: TestTable_OFFLINE, tableTypeStr: REALTIME
Throws 400 exception.

Use case 5:
tableName: TestTable, tableTypeStr: OFFLINE
Only `deleteOfflineTable` will be called.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -238,6 +238,17 @@ public class PinotTableRestletResource {
   public SuccessResponse deleteTable(
       @ApiParam(value = "Name of the table to delete", required = true) @PathParam("tableName") String tableName,
       @ApiParam(value = "realtime|offline", required = false) @QueryParam("type") String tableTypeStr) {
+    TableType tableType = TableNameBuilder.getTableTypeFromTableName(tableName);
+    if (tableType != null) {
+      if (tableTypeStr == null) {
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -238,6 +238,17 @@ public String alterTableStateOrListTableConfig(
   public SuccessResponse deleteTable(
       @ApiParam(value = "Name of the table to delete", required = true) @PathParam("tableName") String tableName,
       @ApiParam(value = "realtime|offline", required = false) @QueryParam("type") String tableTypeStr) {
+    TableType tableType = TableNameBuilder.getTableTypeFromTableName(tableName);
```
mcvsubbu: We have generally followed the convention of specifying a 'type', keeping the suffix _REALTIME and _OFFLINE internal to pinot. Granted there are a couple of APIs that use the suffix, but let us not propagate it (unless there is a specific issue).

PR #3018, opened 2018-07-31
[TE] frontend - aaronucsd/Added new shared summary email feature

aaronucsd:
Summary:
Allow for a share feature from the dashboard. THe user can export the shared summary as pdf,
copy to clipboard and share a snapshot link. There is a way to hide metric and alert level from the final shared content.
The shared feature allows for high level comment and metric level comments.

1. fixed Wow to be predicted instead on dashbaord and shared summary
2. Added new ember data for shared summary
3. added new libs to support PDF download
4. added cryptojs for hash support

screen shot 2018-07-30 at 4 58 42 pm
screen shot 2018-07-30 at 4 58 50 pm


thirdeye/thirdeye-frontend/app/adapters/base.js
```diff
@@ -33,22 +40,51 @@ export default DS.RESTAdapter.extend({
   },
 
   // TODO: Will keep for reference - lohuynh
```
apucher: Is this meant to stay here?

aaronucsd: Removed.

aaronucsd:
Added new route and controller unit tests for home/share-dashboard

PR #3024, opened 2018-07-31
Don't allow default column update in V3 format

Jackie-Jiang:
The reason for this is that, updating default column in V3 format requires changing the existing index
We should avoid this because in HEAP mode this has no effect to the original file
In case of V3 format default column change, drop the segment and download a new one to add the default column


pinot-core/src/test/java/com/linkedin/pinot/core/segment/index/loader/SegmentPreProcessorTest.java
```diff
@@ -260,22 +260,6 @@ public void testV1UpdateDefaultColumns() throws Exception {
     }
   }
 
-  @Test
-  public void testV3UpdateDefaultColumns() throws Exception {
```
sunithabeeram: Please add a test to assert the new behavior.

PR #3027, opened 2018-08-01
[TE] dashboard - fix sharing link

apucher:
* fix sharing namespace link

* auto-copy still broken (omits server url)


aaronucsd:
Ah yeah. Thanks for catching this @apucher !

PR #3033, opened 2018-08-03
Bump zookeeper version to 3.4.11

jackjlli:
This PR bumps zookeeper version to 3.4.11.


kishoreg:
Why not higher. What’s the latest version?

jackjlli:
@kishoreg The reason why I pick 3.4.11 is that Helix is also using this version. 
The latest non-beta version is 3.4.13. 
https://mvnrepository.com/artifact/org.apache.zookeeper/zookeeper

PR #3037, opened 2018-08-03
Retrieve routing table by exact table name instead of by prefix

haibow:
Currently in routing table debug endpoint, it's returning table names that matches by prefix.

e.g. for `table1`, it'll return
```
table1_REALTIME
table1_OFFLINE
table12_REALTIME
table1_staging_OFFLINE
```

Instead, proposing to do the match with removing table type suffix


haibow:
@Jackie-Jiang would you mind reviewing this?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/HelixExternalViewBasedRouting.java
```diff
@@ -555,7 +555,7 @@ public String dumpSnapshot(String tableName)
     JSONArray routingTableSnapshot = new JSONArray();
 
     for (String currentTable : _routingTableBuilderMap.keySet()) {
-      if (tableName == null || currentTable.startsWith(tableName)) {
+      if (tableName == null || tableName.equals(TableNameBuilder.extractRawTableName(currentTable))) {
```
kishoreg: why cant we do both? currentTable.startWith || equals. This should preserve the existing behavior

haibow: Then it'll still return routingTable for `table12` when I only want the routingTable for `table1`, right? I'm not exactly sure whether the existing behavior is a 'feature' or a 'bug' - it is confusing to me when I ask for the routingTable for a table, it's returning information for other unrelated tables. On the UI, I have to look through a whole bunch of unrelated routingTable info in the result to find what I want; when querying from application, I need to do another layer of filtering post getting the result. 

Any anticipated use cases for the current less-strict, prefix-matching logic?

haibow: Right, we don't really control how users name their tables, and we have tables named `eats_order` and `eats_order_with_count` etc. We are using this endpoint for the Presto-Pinot connector, and this behavior caused some confusion to us.

jfim:
This is a debugging feature (which is why it lives at something like `/debug/routingTable/<tableName>`), and the prefix match is intentional; it's not really meant to be used as an actual API.

What's the use case for this, and what data do you need for your integration? You mention that you're using this as part of your Presto-Pinot connector, it should probably use a stable API instead of a debugging endpoint (which could go away at any point in time).

kishoreg:
@jfim They use it in their Presto/Pinot connector. I dont think prefix was intentional in our original design, it was more avoiding check for _OFFLINE and _REALTIME

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedRouting.java
```diff
@@ -594,7 +594,7 @@ public String dumpSnapshot(String tableName)
     ArrayNode routingTableSnapshot = JsonUtils.newArrayNode();
 
     for (String currentTable : _routingTableBuilderMap.keySet()) {
-      if (tableName == null || currentTable.startsWith(tableName)) {
+      if (tableName == null || tableName.equals(TableNameBuilder.extractRawTableName(currentTable))) {
```
Jackie-Jiang: By Pinot naming convention, `tableName` means table name with or without type suffix. If the table name provided has the suffix, we should not return the one with the different suffix.
`table1` will match `table12` is a great example why we should not use `startsWith`. To solve that, the check should be on the type. You can us `TableNameBuilder.getTableTypeFromTableName()` to get the type. If type is null, compare the raw table name; if type is not null, check if the name is the same.

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedRouting.java
```diff
@@ -592,9 +592,12 @@ public String dumpSnapshot(String tableName)
       throws Exception {
     ObjectNode ret = JsonUtils.newObjectNode();
     ArrayNode routingTableSnapshot = JsonUtils.newArrayNode();
+    boolean tableNameHasTypeSuffix = TableNameBuilder.getTableTypeFromTableName(tableName) != null;
```
Jackie-Jiang: This might throw NPE

haibow: Thanks for catching this!

haibow:
@jfim sorry I missed your message earlier. I'm working on open sourcing the Presto Pinot connector and just came back to this diff. I'm relieved to find this endpoint is preserved on Apache Pinot 0.1.0 release :grinning: Would be awesome if we can make the API stable/public.

@kishoreg @Jackie-Jiang thanks for the comments!

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedRouting.java
```diff
@@ -594,7 +594,9 @@ public String dumpSnapshot(String tableName)
     ArrayNode routingTableSnapshot = JsonUtils.newArrayNode();
 
     for (String currentTable : _routingTableBuilderMap.keySet()) {
-      if (tableName == null || currentTable.startsWith(tableName)) {
+      if (tableName == null
```
Jackie-Jiang: Can you reformat using Pinot Style?
Should be like:
```
      if (tableName == null || (TableNameBuilder.getTableTypeFromTableName(tableName) != null && tableName
          .equals(currentTable)) || (TableNameBuilder.getTableTypeFromTableName(tableName) != null && tableName
          .equals(currentTable))) {
```

Jackie-Jiang:
Still not correct lol

PR #3039, opened 2018-08-07
Add inverted index option for offline data gen

jenniferdai:
* Currently, if the client's table config sets inverted index columns, there is no way to disable inverted index generation during segment generation. If we add this ability, we can configure this in our table config. Because generating an inverted index can lead to inflated segment sizes, this parameter will be useful if the user has trouble pushing their segment because of its size/if push takes too long and the user would rather take the hit to generate on the server side.


pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
```diff
@@ -37,6 +37,9 @@
   @ConfigKey("autoGeneratedInvertedIndex")
   private boolean _autoGeneratedInvertedIndex;
 
+  @ConfigKey("addInvertedIndexDuringDataGeneration")
+  private boolean _addInvertedIndexDuringDataGeneration;
```
Jackie-Jiang: "createInvertedIndexDuringSegmentGeneration"? Your call

PR #3041, opened 2018-08-07
Changed server api end points to return string instead of java objects

snleee:
Server side APIs for the internal release are broken due to denpendency
issues. Returning string value instead of java object for jersey API
resolves the issue.


mayankshriv:
It would be great to understand what exactly fails, because we are fixing one instance of this problem, and I am worried we might run into other issues in future.

snleee:
@sunithabeeram As we discussed offline,

PR #3047, opened 2018-08-07
Only log SEVERE message in travis

Jackie-Jiang:


pinot-common/src/main/resources/logging-severe-only.properties
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (C) 2014-2015 LinkedIn Corp. (pinot-core@linkedin.com)
```
jackjlli: It'd be better to update the time range to `2014-2018`. The same in `log4j-fatal-only.properties` as well.

Jackie-Jiang: Seems we don't check license in config files haha

PR #3051, opened 2018-08-08
[PINOT-6696] Remove http connection pool in table-size api path

sunithabeeram:
It appears that MultiThreadedHttpConnectionManager is buggy - especially when it encounters a connection refused error on any connection, it fails to "release" the underlying connection and subsequent connections to other servers over the pool encounter connection-refused due to the stale connection. This causes failures in the TableSize API which is invoked during segment upload, failing the upload.

Also, in the current code, we use default size for max connections which is 20; for one of large clusters (with ~80 servers hosting segments of a table), the 20 max connection essentially means that the connections are established and torn down, essentially rendering the pool ineffective. 

If we do need to use a pool, we should consider using PoolingHttpConnectionManager - this requires a httpclient version bump to 4.x.x from the existing 3.1. That is a larger change - for this phase, we'll consider the simpler case of bypassing the pool completely.

Testing done:
Deployed updated binary in integration cluster and hit table/size api on all tables hosted on the controller - checked that the basic functionality worked, there are no resource leaks (via netstat/lsof). Also, checked that there are no connection refused errors to hosts that are up.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
@@ -23,7 +23,6 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorCompletionService;
 import javax.annotation.Nonnull;
 import org.apache.commons.httpclient.HttpClient;
-import org.apache.commons.httpclient.HttpConnectionManager;
 import org.apache.commons.httpclient.methods.GetMethod;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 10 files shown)


sunithabeeram:
@mcvsubbu good point about deleteClosedConnections()! I looked through the code a bit more and it seems like the issue is more with how the "freeconnection" list is handled. When the application tries to execute a method, the HttpMethodDirector fetches a connection to satisfy the request. The pool-mgr should return a connection with the host-configuration passed in. But it looks like thats not happening correctly in some cases. This a bit of FUD at this point, but its possible that we get an incorrect connection when we don't see the error too (as there is nothing so far that points me to the fact that this is related to conn-refused; its just more noticeable in that particular case).

I think we should upgrade httpclients and work with newer connection pools than chase issues with the older clients.

Appreciate the review and suggestions!

PR #3052, opened 2018-08-08
Returning default fs on instantiation

jenniferdai:


sunithabeeram:
Can we please add test coverage for this? Did our tests always set a conf?

jenniferdai:
The conf is set in our tests

sunithabeeram:
Please do add a test without the conf set and ensure things are intialized correctly. We don't want future code changes/refactoring to knock the check you added resulting in the same issue again.

PR #3055, opened 2018-08-09
Testing pinotFS instantiation

jenniferdai:
Ensuring default controller starter behavior works as expected without having to add a config for the PinotFSFactory classes by removing the config setting in all our test paths and throughout our code. Verified that local implem is returned by default


snleee:
LGTM

PR #3057, opened 2018-08-09
Reduce heap allocation during realtime segment creation

mcvsubbu:
In order to calculate min and max values of a no-dictionary column we walk through the entire value set,
allocating a good bit of heap memory. But then we really don't use the value generated.
The getMin() and getMax() methods are called only to construct ColumnIndexCreationInfo (the values
gathered from columnStatistics are stored in the ColumnIndexCreationInfo object). But then
the min/max values from ColumnIndexCreationInfo are retrieved only for the time column and not for
any other column. Time column is never a no-dictionary column, so we can remove all the code in the
no-dictionary column to compute min-max.

This essentially reverts #1468, which was submitted because of review comments on #1462.
Comment on #1462 should now be addressed, because we now throw exception instead of returning
some invalid value.


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeNoDictionaryColStatistics.java
```diff
@@ -29,120 +29,23 @@
...
-    _maxValue = Double.valueOf(max);
-  }
-
   @Override
   public Object getMinValue() {
-    if (_minValue == null) {
-      gatherStatistics();
-    }
-    return _minValue;
+    throw new RuntimeException("Cannot get min value for no dictionary column " + _operatorName);
```
sunithabeeram: Whats the impact of this change? Does the calling code appropriately bracket this in try/catch? This shouldn't result in query time exceptions.

sunithabeeram:
Can we quantify how much of an improvement we see with this?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/stats/RealtimeNoDictionaryColStatistics.java
sunithabeeram: Looks like this is used only in SegmentColumnarIndexCreator and does not get used in the query path. As the description notes SegmentColumnarIndexCreator calls min/max only for time-column in writeMetadata().

mcvsubbu: Even in segment creation, this call never happens (with my change). Prior to my change, the call used to happen but the returned value never used.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/ColumnIndexCreationInfo.java
```diff
@@ -24,44 +24,19 @@
...
-  private final int lengthOfShortestEntry;
-  private final int lengthOfLongestEntry;
   private final boolean isAutoGenerated;
-  private final PartitionFunction partitionFunction;
-  private final List<IntRange> partitionRanges;
   private final Object defaultNullValue;
-  private int numPartitions;
+  private final ColumnStatistics columnStatistics;
 
-  public ColumnIndexCreationInfo(boolean createDictionary, Object min, Object max, Object sortedUniqueElementsArray,
```
sunithabeeram: That probably has to be the constructor with most args that I've seen. Glad we reduced that!

mcvsubbu:
I ran "jstat -gcutil", starting it before we build the segment, and terminating it after. I did this for building segment for an internal use case that had 10 no-dictionary columns. The code I invoked was to build a segment for one partition of this use case. A recent random segment form production was taken and replayed for this experiment.

I ran the jvm with "-Xmx24G -Xms24G -XX:NewSize=16g" so that we do not have a GC at all during the entire experiment.

The output of jstat showed that we decreased heap usage by 2.6% for this use case.

That works out to be 425M of heap for this particular use case, per partition.

PR #3058, opened 2018-08-09
[TE] Add support for data filter in detection pipeline emulation layer

jihaozh:
Add support for legacy data filter in detection pipeline emulation layer


jihaozh:
The old pipeline skips anomaly detection if the time series does not pass data filter with in a time range.

PR #3059, opened 2018-08-09
Changing pinotFS to use temp file instead of uri for copyToLocal

jenniferdai:
* Currently, we use URI. This doesn't make sense because it will always be copied to the local filesystem, so we should use a File object.

### DIFF
--- a/pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
+++ b/pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
@@ -110,13 +110,13 @@ public class LocalPinotFS extends PinotFS {
   }
 
   @Override
-  public void copyToLocalFile(URI srcUri, File tempFile) throws Exception {
-    copy(srcUri, tempFile.toURI());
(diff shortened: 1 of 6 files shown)


pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
```diff
@@ -107,8 +110,9 @@ public long length(URI fileUri) throws IOException {
   }
 
   @Override
-  public void copyToLocalFile(URI srcUri, URI dstUri) throws IOException {
-    copy(srcUri, dstUri);
+  public void copyToLocalFile(URI srcUri, File tempFile) throws Exception {
```
mcvsubbu: suggest rename variable from tempFile to destFile.

It is the application logic that decides whether the file is temp or permanent.

Also, why has IOException changed to Exception?

jenniferdai: Renamed.

In my next one,  I will add retries, so I am preparing for that.

PR #3060, opened 2018-08-09
[PINOT-6845] Provide a controller override for storage quota checks

sunithabeeram:
Changes include:
- add a controller level config to disable storage quota checks. Enabled by default. Will flip this around when we are ready to reenable the storage checks.
- Use the conf to determine if quota needs to be checked during upload.

Testing done:
Pushed changes to integration test controller and confirmed that quota tests are not done.

Currently there is no test coverage for the APIs. Will look into adding them in subsequent PRs.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
@@ -80,6 +80,9 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final long DEFAULT_SEGMENT_UPLOAD_TIMEOUT_IN_MILLIS = 600_000L; // 10 minutes
   private static final int DEFAULT_REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS = 64;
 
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -80,6 +80,9 @@
   private static final long DEFAULT_SEGMENT_UPLOAD_TIMEOUT_IN_MILLIS = 600_000L; // 10 minutes
   private static final int DEFAULT_REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS = 64;
 
+  private static final String ENABLE_STORAGE_QUOTA_CHECK = "controller.enable.storage.quota.check";
+  private static final boolean DEFAULT_STORAGE_QUOTA_CHECK = false;
```
snleee: If the storage quota was enabled by default before this change, I think we should turn this on by default. Otherwise, it's not backward compatible and may affect to users outside.

sunithabeeram: Updated the default.

PR #3061, opened 2018-08-09
Add a tool to clean up non-live instances

Jackie-Jiang:
The tool will simply remove all instances that are not in LIVEINSTANCES from INSTANCES and CONFIGS/PARTICIPANT.
Use with caution!


Jackie-Jiang:
@kishoreg Yes, we should call Helix.deleteInstance() ideally. The reason I added this small tool is because when we replace hardware we don't call that method, and ZK **INSTANCES** dir will never be cleaned up. Before running this tool, we need to make sure all instances in IdealStates are live. I agree that this tool is a little risky to use. Will close the PR and use it just for a record.

PR #3064, opened 2018-08-10
Revert "Remove using http connection pool in table-size api path

sunithabeeram:
This reverts commit d4fb2e0cb4c6c1bee69aa6452e4690e30f308829.

As silly as this sounds, this commit was based on an incorrect log line :(. Will fix the logs in another PR.

Also, turns out that we don't have any limits on the thread-pool used for making out bound connections, so removing the connection pool as well results in unbounded connections being made.


mcvsubbu:
Looks like the pool at least limits the number of outgoing connections we make per segments. Agree that we should keep the pool for now (even though it is churning all the time) while we look for alternative solutions.

sunithabeeram:
Yes, this is a clean revert of the commit.

PR #3067, opened 2018-08-11
[PINOT-6856]: Fix controller storage check logic when it is unable to get reports from servers

sunithabeeram:
Currently, when the controller fails to get a segment size report from any replica for a given segment, it falls back to using the table-level max. The problem with that is that for highly skewed tables, we can overestimate by a lot and cause segment uploads to fail.

It appears that it might be safer to let the upload go through in such cases, than try to extrapolate sizes based on existing sizes. Depending on which segments we actually get reports back from using, mix/max/average/median etc can always be vulnerable to overestimation.

With this change, we skip segments where we have no report from any replica. For segments that have report from some, but not all, we fill in for the other replicas using the segment-level-max (it should be extremely rare that replicas report different sizes for segments!)

Changes include:
- Firing off a metric when we skip quota enforcement when we don't hear back from any server at all.
- Firing off a metric when we go through with quota enforcement with partial reports. We still go through with the check as it can help catch cases where the table is already over quota even with the existing reports.
- Added a missingSegments count to track how many segments we don't have reports for at all.
- Fixed test to add coverage for missing segments.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
@@ -81,7 +81,8 @@ public class SegmentStatusChecker {
     _waitForPushTimeSeconds = config.getStatusCheckerWaitForPushTimeInSeconds();
     _metricsRegistry = metricsRegistry;
     HttpConnectionManager httpConnectionManager = new MultiThreadedHttpConnectionManager();
-    _tableSizeReader = new TableSizeReader(Executors.newCachedThreadPool(), httpConnectionManager, _pinotHelixResourceManager);
+    _tableSizeReader = new TableSizeReader(Executors.newCachedThreadPool(), httpConnectionManager,
+        _metricsRegistry, _pinotHelixResourceManager);
 
     _executorService = Executors.newSingleThreadScheduledExecutor(new ThreadFactory() {
       @Override
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
@@ -23,7 +23,6 @@ import com.linkedin.pinot.common.metrics.ControllerMetrics;
 import com.linkedin.pinot.common.utils.CommonConstants.Helix.TableType;
 import com.linkedin.pinot.controller.ControllerConf;
 import com.linkedin.pinot.controller.helix.core.PinotHelixResourceManager;
-import com.linkedin.pinot.controller.util.TableSizeReader;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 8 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableSizeReader.java
```diff
@@ -199,35 +208,36 @@ public TableSubTypeSizeDetails getTableSubtypeSize(String tableNameWithType, int
...
+            0);
+
+      } else {
+        LOGGER.warn("Missing size report for {} out of {} segments for table {}", subTypeSizeDetails.missingSegments,
+            subTypeSizeDetails.segments.size(), table);
+        _controllerMetrics.setValueOfTableGauge(table, ControllerGauge.TABLE_STORAGE_ESTIMATION_PARTIAL,
+            1);
+        // reset metric tracking failure
+        _controllerMetrics.setValueOfTableGauge(table, ControllerGauge.TABLE_STORAGE_ESTIMATION_FAILURE,
+            0);
```
mcvsubbu: You need to reset both metrics to 0 if there are no missing segments.

Would it make sense to have one metric ControllerGauge.TABLE_STORAGE_EST_FAILED_SEGMENTS, and report the number of segments failed? If it is 0 failed, then the metric automatically becomes 0.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -81,7 +81,8 @@ public SegmentStatusChecker(PinotHelixResourceManager pinotHelixResourceManager,
     _waitForPushTimeSeconds = config.getStatusCheckerWaitForPushTimeInSeconds();
     _metricsRegistry = metricsRegistry;
     HttpConnectionManager httpConnectionManager = new MultiThreadedHttpConnectionManager();
-    _tableSizeReader = new TableSizeReader(Executors.newCachedThreadPool(), httpConnectionManager, _pinotHelixResourceManager);
+    _tableSizeReader = new TableSizeReader(Executors.newCachedThreadPool(), httpConnectionManager,
```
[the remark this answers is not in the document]
sunithabeeram: Done.

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableSizeReader.java
sunithabeeram: Good point. Incorporated both comments.

pinot-controller/src/main/java/com/linkedin/pinot/controller/util/TableSizeReader.java
```diff
@@ -199,35 +208,41 @@ public TableSubTypeSizeDetails getTableSubtypeSize(String tableNameWithType, int
...
-        subTypeSizeDetails.estimatedSizeInBytes += sizeDetails.estimatedSizeInBytes;
+    if (subTypeSizeDetails.missingSegments > 0) {
+      if (subTypeSizeDetails.missingSegments == subTypeSizeDetails.segments.size()) {
+        LOGGER.warn("Could not get size reports from any server for all segments of the table {}", table);
+        subTypeSizeDetails.reportedSizeInBytes = -1;
+        subTypeSizeDetails.estimatedSizeInBytes = -1;
+        _controllerMetrics.setValueOfTableGauge(table, ControllerGauge.TABLE_STORAGE_ESTIMATION_FAILURE,
+            1);
+        // reset metric tracking partial estimation
+        _controllerMetrics.setValueOfTableGauge(table, ControllerGauge.TABLE_STORAGE_EST_FAILED_SEGMENTS,
```
mcvsubbu: shouldn't this metric be set to segments.size()? we failed to get size for all segments, right? And then maybe we don't need the other metric at all?

PR #3068, opened 2018-08-11
[PINOT-6701] Controller API to check storage quota before segment upload

sunithabeeram:
Expose a controller API to check storage quota before segments are uploaded. The API expects, as a query parameter, the uncompressed size of all segments that are to be uploaded. The API accounts for the appropriate push-type (refresh vs append) and whether the segments to be uploaded have indices or not and returns whether the upload can go through or not.

Eventually, we will drop the quota checks that are done per segment and expect all push jobs to call this API before starting the upload. That will save a lot of unnecessary calls to the servers for get table sizes as well as prevent partial uploads if the size is exceeded mid way through the segments push.

Testing Done: added integration test that covers the append use-case upload. Plan to add more after initial review.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/TableUploadQuotaCheck.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/TableUploadQuotaCheck.java
@@ -1,0 +1,156 @@
+/**
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.controller.api.resources;
+
+import com.linkedin.pinot.common.config.TableConfig;
+import com.linkedin.pinot.common.config.TableNameBuilder;
+import com.linkedin.pinot.common.metadata.ZKMetadataProvider;
+import com.linkedin.pinot.common.utils.DataSize;
+import com.linkedin.pinot.controller.ControllerConf;
+import com.linkedin.pinot.controller.helix.core.PinotHelixResourceManager;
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/TableUploadQuotaCheck.java
```diff
@@ -0,0 +1,156 @@
...
+      throw new ControllerApplicationException(LOGGER,
+          "size in bytes: " + size + " invalid", Response.Status.BAD_REQUEST);
+    }
+
+    String offlineTableName = TableNameBuilder.OFFLINE.tableNameWithType(tableName);
+    TableConfig offlineTableConfig =
+        ZKMetadataProvider.getOfflineTableConfig(_pinotHelixResourceManager.getPropertyStore(), offlineTableName);
+    if (offlineTableConfig == null) {
+      throw new ControllerApplicationException(LOGGER,
+          "Table: " + tableName + "not found", Response.Status.NOT_FOUND);
```
jackjlli: Add a space before "not found".

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/TableUploadQuotaCheck.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    if (!offlineTableConfig.getIndexingConfig().isCreateInvertedIndexDuringSegmentGeneration()) {
+      if (!offlineTableConfig.getIndexingConfig().getInvertedIndexColumns().isEmpty()) {
+        size = size + (long)(size * INDEX_RATIO);
+        LOGGER.info("Size (in bytes) estimate with indices included is ", size);
```
jackjlli: Please log the tableName here as well.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/TableUploadQuotaCheck.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    if (tableSizeDetails.reportedSizeInBytes == -1) {
+      // TODO: handle partial errors too once #3067 is merged
+      throw new ControllerApplicationException(LOGGER, String.format("Failed to read table size for %s", tableName),
+          Response.Status.INTERNAL_SERVER_ERROR);
+    }
+
+    if (tableSizeDetails.estimatedSizeInBytes + uploadSizeForAllReplicas > configured) {
+      String message = "Upload of " + DataSize.fromBytes(size) + " will exceed quota. Estimated append size "
```
jackjlli: I think it's best to use String.format() to form the String.

npawar:

PR #3070, opened 2018-08-14
[TE] detection - support active flag

apucher:


jihaozh:
One question. How will this affect our existing detection configs without the new active flag?

apucher:
They'll be disabled unless specifically activated. We're already seeing a bunch of unused prototypes lying around

PR #3073, opened 2018-08-14
Reduce heap usage during segment builds

mcvsubbu:
While measuring heap allocations during segment builds, jmap showed
sun.nio.cs.UTF_8$Decoder and sun.nio.cs.UTF_8$Encoder among the top
40 objects in both size and number of instances. In 4 use internal use cases
the jmap output had these numbers, allocated during segment creation
(in most cases, there were in the top 5 in terms of memory)

Use case 1:
 num     #instances         #bytes  class name
----------------------------------------------
   3:      26459379     1058375160  sun.nio.cs.UTF_8$Decoder
   7:       5592818      313197808  sun.nio.cs.UTF_8$Encoder

Use case 2:
   2:        950923       38036920  sun.nio.cs.UTF_8$Decoder
   30:          2349         131544  sun.nio.cs.UTF_8$Encoder

Use case 3:
   3:       5385769      215430760  sun.nio.cs.UTF_8$Decoder

Use case 4:
   5:       5407753      216310120  sun.nio.cs.UTF_8$Decoder
   7:       2016142      112903952  sun.nio.cs.UTF_8$Encoder

Upon closer look, it appears that the call String.getBytes("UTF-8") caches the
decoder in a ThreaLocal variable in StringCoding class, whereas the call to String.getBytes(Charset)
ends up creating a new decoder everytime.

Same with encoder calls.

It should help to cache the encoder/decoder in query path as well.

This change replaces calls to encode/decode Strings to utf8 with a method in
StringUtil class.

After the change, the decoder/encoder do not appear in the top objects.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
@@ -41,7 +41,6 @@ import org.apache.commons.codec.binary.Hex;
  */
 @SuppressWarnings("unused")
 public abstract class FieldSpec implements Comparable<FieldSpec>, ConfigNodeLifecycleAware {
-
   private static final Integer DEFAULT_DIMENSION_NULL_VALUE_OF_INT = Integer.MIN_VALUE;
   private static final Long DEFAULT_DIMENSION_NULL_VALUE_OF_LONG = Long.MIN_VALUE;
   private static final Float DEFAULT_DIMENSION_NULL_VALUE_OF_FLOAT = Float.NEGATIVE_INFINITY;
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
@@ -21,10 +21,8 @@ import com.google.gson.JsonPrimitive;
 import com.linkedin.pinot.common.Utils;
 import com.linkedin.pinot.common.config.ConfigKey;
 import com.linkedin.pinot.common.config.ConfigNodeLifecycleAware;
-import com.linkedin.pinot.common.utils.DataSchema;
 import com.linkedin.pinot.common.utils.EqualityUtils;
 import com.linkedin.pinot.common.utils.primitive.ByteArray;
-import java.nio.charset.Charset;
 import javax.annotation.Nonnull;
 import org.apache.avro.Schema.Type;
 import org.apache.commons.codec.DecoderException;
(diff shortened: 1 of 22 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/StringUtil.java
```diff
@@ -15,12 +15,14 @@
...
 package com.linkedin.pinot.common.utils;
 
+import java.io.UnsupportedEncodingException;
 import javax.annotation.Nonnull;
 import org.apache.commons.lang.StringUtils;
 
 
 public class StringUtil {
   private static final char NULL_CHARACTER = '\0';
+  private static String charSet = "UTF-8";
```
Jackie-Jiang: Please make this private static **final** with name **UTF-8**

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
```diff
@@ -43,7 +41,6 @@
  */
 @SuppressWarnings("unused")
 public abstract class FieldSpec implements Comparable<FieldSpec>, ConfigNodeLifecycleAware {
-  private static final Charset UTF_8 = Charset.forName("UTF-8");
```
Jackie-Jiang: (OCD) Can you also remove the empty line followed by the removed line? Same for other files

pinot-common/src/main/java/com/linkedin/pinot/common/utils/StringUtil.java
mcvsubbu: meant to do that.

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
mcvsubbu: really? i was just deleting stuff that is not used....
oh well..

pinot-common/src/main/java/com/linkedin/pinot/common/utils/StringUtil.java
Jackie-Jiang: I prefer change the variable name to "UTF_8" because based on the method name, it has to be UTF-8 instead of any other char set. and also the convention for private static final variable should be all capital letters.

mcvsubbu: woops, missed this last comment of yours. Will fix in the next code checkin. Sorry.

sunithabeeram:
Nice find! 1G per segment for one of the use cases is a nice saving.

PR #3074, opened 2018-08-14
[TE] rootcause - remove wo3w and wo4w by default

apucher:
* remove metric table wo3w and wo4w by default to reduce data source load

* hide predicted value header box if not available

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
@@ -317,6 +317,7 @@ export default Controller.extend({
       //
       // related metrics
       //
+      const anomalyMetricUrns = new Set();
       const relatedMetricUrns = new Set();
 
       if (activeTab === ROOTCAUSE_TAB_METRICS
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -371,11 +372,17 @@ export default Controller.extend({
...
       // aggregates
       //
-      const offsets = ['current', 'baseline', 'wo1w', 'wo2w', 'wo3w', 'wo4w'];
+      const offsets = ['current', 'baseline', 'wo1w', 'wo2w'];
       const offsetUrns = [...relatedMetricUrns]
         .map(urn => [].concat(offsets.map(offset => toOffsetUrn(urn, offset))))
         .reduce((agg, l) => agg.concat(l), []);
-      aggregatesService.request(context, new Set(offsetUrns));
+
+      const anomalyOffsets = ['current', 'baseline', 'wo1w', 'wo2w', 'wo3w', 'wo4w'];
```
aaronucsd: We still need `'wo3w', 'wo4w'` here?

PR #3076, opened 2018-08-14
Bump helix version to 0.8.2

jackjlli:
This PR bumps helix version to 0.8.2.
The method `waitUntilConnected()` has been removed. Use `waitUntilConnected(long time, TimeUnit timeUnit)` instead.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotRealtimeSegmentManager.java
@@ -43,6 +43,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import org.I0Itec.zkclient.IZkChildListener;
 import org.I0Itec.zkclient.IZkDataListener;
 import org.apache.helix.ControllerChangeListener;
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotRealtimeSegmentManager.java
```diff
@@ -92,7 +93,7 @@ public void start(ControllerMetrics controllerMetrics) {
     String zkUrl = _pinotHelixResourceManager.getHelixZkURL();
     _zkClient = new ZkClient(zkUrl, ZkClient.DEFAULT_SESSION_TIMEOUT, ZkClient.DEFAULT_CONNECTION_TIMEOUT);
     _zkClient.setZkSerializer(new ZNRecordSerializer());
-    _zkClient.waitUntilConnected();
+    _zkClient.waitUntilConnected(20, TimeUnit.SECONDS);
```
sunithabeeram: I don't think waitUntilConnected() and waitUntilConnected(20s) are the same thing. What happens after 20s timeout? Is the code able to recover from that gracefully?

mcvsubbu: No, the controller will fail to start.

New zk client in helix 0.8.2 does not support the old API.

This is called for the HLC segment manager in Pinot controller during controller start.

jackjlli: @sunithabeeram From the code perspective, the method `waitUntilConnected()` in 0.8.1 calls 
the method `waitUntilConnected(Integer.MAX_VALUE, TimeUnit.MILLISECONDS)`, which literally means waiting there forever.

sunithabeeram: I understand that the API support has changed. But we could technically do waitUntilConnected(Integer.MAX_VALUE,...) to keep the old behavior. Just making sure that this new behavior is indeed desirable.

Also, nit: it would be good to use a named constant instead of "20".

jackjlli: SGTM. I'll create another PR for that.

jackjlli: https://github.com/linkedin/pinot/pull/3078

PR #3078, opened 2018-08-15
Put zk client timeout param as constant

jackjlli:


pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
```diff
@@ -59,6 +59,10 @@
       }
     }
 
+    public static class ZkClient {
+      public static final long DEFAULT_TIMEOUT_IN_SECOND = 20L;
```
sunithabeeram: Would be good to include connect in the name; DEFAULT_CONNECT_TIMEOUT_SEC

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotRealtimeSegmentManager.java
```diff
@@ -93,7 +94,7 @@ public void start(ControllerMetrics controllerMetrics) {
     String zkUrl = _pinotHelixResourceManager.getHelixZkURL();
     _zkClient = new ZkClient(zkUrl, ZkClient.DEFAULT_SESSION_TIMEOUT, ZkClient.DEFAULT_CONNECTION_TIMEOUT);
     _zkClient.setZkSerializer(new ZNRecordSerializer());
-    _zkClient.waitUntilConnected(20, TimeUnit.SECONDS);
+    _zkClient.waitUntilConnected(CommonConstants.Helix.ZkClient.DEFAULT_TIMEOUT_IN_SECOND, TimeUnit.SECONDS);
```
kishoreg: Shouldn't we wait until we get connected without any timeout? Throwing exception in background threads is not of much use

jackjlli: The method `waitUntilConnected` has been called in the constructor of HelixAdmin in Class `HelixSetupUtils` before it hits this line of code. So it should be good.

sunithabeeram: I had a similar concern about waiting till connected here: https://github.com/linkedin/pinot/pull/3076#discussion_r210147713

PR #3079, opened 2018-08-15
Fix TDigest for cases where only one value in TDigest

Jackie-Jiang:
TDigest.quantile() does not support one value case, directly return value when there is only one value

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/PercentileTDigestAggregationFunction.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/PercentileTDigestAggregationFunction.java
@@ -200,7 +200,20 @@ public class PercentileTDigestAggregationFunction implements AggregationFunction
   @Nonnull
   @Override
   public Double extractFinalResult(@Nonnull TDigest intermediateResult) {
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/PercentileTDigestAggregationFunction.java
```diff
@@ -200,7 +200,20 @@ public boolean isIntermediateResultComparable() {
...
+  }
+
+  /**
+   * Calculates percentile from {@link TDigest}.
+   * <p>Handles cases where only one value in TDigest object.
+   */
+  public static double calculatePercentile(@Nonnull TDigest tDigest, int percentile) {
+    if (tDigest.size() == 1) {
+      // Specialize cases where only one value in TDigest (cannot use quantile method)
+      return tDigest.centroids().iterator().next().mean();
```
mayankshriv: Is it guaranteed that centriods() iterator() and next() return a non-null object? If not, lets add null check.

Jackie-Jiang: Yes, it is guaranteed.

PR #3094, opened 2018-08-21
Adding azure FS improvements

jenniferdai:
* Supporting copy on ADL
* Ignoring old file during fs methods if it exists
* Validated through local testing on ADL

### DIFF
--- a/pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
+++ b/pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
@@ -16,6 +16,8 @@
 package com.linkedin.pinot.filesystem;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.microsoft.azure.datalake.store.ADLFileInputStream;
+import com.microsoft.azure.datalake.store.ADLFileOutputStream;
 import com.microsoft.azure.datalake.store.ADLStoreClient;
 import com.microsoft.azure.datalake.store.DirectoryEntry;
 import com.microsoft.azure.datalake.store.oauth2.AccessTokenProvider;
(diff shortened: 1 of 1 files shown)


pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
```diff
@@ -88,8 +91,15 @@ public boolean move(URI srcUri, URI dstUri) throws IOException {
 
   @Override
   public boolean copy(URI srcUri, URI dstUri) throws IOException {
-    // TODO: Add this method. Not needed for now because file will be in final location
-    throw new UnsupportedOperationException("Cannot copy from AzureFS to AzureFS");
+    if (exists(dstUri)) {
```
mayankshriv: Should we have overwrite as parameter passed? So we don't always overwrite.

pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
```diff
@@ -117,6 +127,13 @@ public long length(URI fileUri) throws IOException {
 
   @Override
   public void copyToLocalFile(URI srcUri, File dstFile) throws Exception {
+    if (dstFile.exists()) {
```
mayankshriv: Same here.

pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
jenniferdai: I think we don't have that use case in Pinot and I don't anticipate us needing that as these are used for segment upload, refreshing segments on server side, etc.

jenniferdai: When we designed, other filesystems default overwrite, so I wanted to keep it consistent.

PR #3098, opened 2018-08-22
[TE] monitoring - add alert and detection task counters

apucher:


thirdeye/thirdeye-pinot/src/test/java/com/linkedin/thirdeye/detection/algorithm/MovingWindowAlgorithmTest.java
```diff
@@ -21,9 +21,6 @@
 import java.util.List;
```
jihaozh: Looks like a unrelated file change.

apucher: Good catch. I wonder if it's worth making a separate PR though

jihaozh: Let's just put it here. Thank you for cleaning up the tests!

PR #3111, opened 2018-08-24
[PINOT-6738] Deprioritize multi-value columns during scan-based filtering

sunithabeeram:
This is the same change as in #3010 - with change to PerfBenchmarkRunner removed (didn't belong with the change). (I had trouble with that branch locally so I started with master afresh and applied the diffs from 3010).

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BaseFilterOperator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BaseFilterOperator.java
@@ -15,9 +15,12 @@
  */
 package com.linkedin.pinot.core.operator.filter;
 
+import com.linkedin.pinot.core.common.DataSource;
+import com.linkedin.pinot.core.common.DataSourceMetadata;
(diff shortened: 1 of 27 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BaseFilterOperator.java
```diff
@@ -15,18 +15,41 @@
...
+
   /**
    * Return whether the result is empty.
    */
   public abstract boolean isResultEmpty();
+
+  /**
+   * Returns the datasource metadata
+   */
+  public @Nullable DataSourceMetadata getDataSourceMetadata() {
```
kishoreg: can we expose columnmetadata instead of datasourcemetadata. We dont want to expose datasourcemetadata outside of datasource.

sunithabeeram: Will address this in a followup PR. Thanks!

PR #3112, opened 2018-08-27
Adding mkdir to PinotFS

jenniferdai:
* In order to support a move in pluggable storage during segment upload, we want to configure baseDataDirs with different schemes. Therefore, we need FileUploadPathProvider, which creates the base data directory, to use PinotFS for making a directory.


pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
```diff
@@ -40,6 +40,11 @@ public LocalPinotFS() {
   public void init(Configuration configuration) {
   }
 
+  @Override
+  public void mkdir(URI uri) throws IOException {
+    FileUtils.forceMkdir(new File(uri));
```
mayankshriv: What's the behavior for forceMkdir if the directory already exists? We should not delete existing directory.

jenniferdai: Agree that we should collect all necessary apis, but I missed this one. I think after this one, we should be complete. forceMkDir is what we currently use in our code, so I wanted to keep the behavior the same. It doesn't overwrite the directory if it already exists.

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
```diff
@@ -40,6 +40,11 @@ public LocalPinotFS() {
   public void init(Configuration configuration) {
   }
 
+  @Override
+  public void mkdir(URI uri) throws IOException {
```
mayankshriv: Also, noticed that most other api's return a boolean but this one is void, any reason for that? If not, lets try to be consistent.

pinot-filesystem/src/test/java/com/linkedin/pinot/filesystem/LocalPinotFSTest.java
```diff
@@ -99,7 +99,8 @@ public void testFS() throws Exception {
 
     File firstTempDir = new File(_absoluteTmpDirPath, "firstTempDir");
     File secondTempDir = new File(_absoluteTmpDirPath, "secondTempDir");
-    Assert.assertTrue(firstTempDir.mkdirs(), "Could not make directory " + firstTempDir.getPath());
+    localPinotFS.mkdir(firstTempDir.toURI());
```
mayankshriv: Does this test leave the temp directories around after it is done? It should clean them. And that requires a new api in the interface (rmdir).

jenniferdai: no, it cleans all tempdirs, i do have delete in pinotfs to support this

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
jenniferdai: wanted to migrate all to void and throw exception always to be consistent across filesystems but can leave it as boolean for now

pinot-filesystem/src/test/java/com/linkedin/pinot/filesystem/LocalPinotFSTest.java
mayankshriv: Oh I see, delete is for file as well as directory I guess.

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
```diff
@@ -40,6 +40,12 @@ public LocalPinotFS() {
   public void init(Configuration configuration) {
   }
 
+  @Override
+  public boolean mkdir(URI uri) throws IOException {
+    FileUtils.forceMkdir(new File(uri));
+    return true;
```
mayankshriv: Always returning true is not useful.

jenniferdai: agree, that's why i made the interface defn void. For now, will leave as boolean with the rest as you mentioned in an earlier comment for consistency. Will clean all of them up together after doing some more careful investigation in our code.

PR #3114, opened 2018-08-27
Making segment upload changes for move

jenniferdai:
* During segment upload, we want URI upload to always move segment to a permanent directory so that we can have full control over deletion, operations, etc
* Also, we want all segment uploads to come through the controller if there is any movement that needs to occur, aka, if the segment is not in its final resting place in remote storage. Reasoning for this is we want the flexibility to upload to/from different kinds of filesystems. Eventually, we will also want to support download from https since some remote storages support that as well.
* Integration test is WIP, but putting up for early review


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -504,9 +504,10 @@ private void uploadSegment(File indexDir, SegmentMetadata segmentMetadata, File
     // Brand new segment, not refresh, directly add the segment
     if (znRecord == null) {
       LOGGER.info("Adding new segment: {}", segmentName);
-      if (downloadUrl == null) {
+      // We will only assume that segment metadata upload doesn't have to do a move through the controller because
```
mayankshriv: Why does segment metadata upload imply that the segment is in its final location? These two seem unrelated, and should not be tied together. If that is because controller moving the segment requires segment to download, then we should revisit why that is required and try to fix that, instead of building on top of that.

PR #3115, opened 2018-08-27
Adding PinotCrypter interface

jenniferdai:
* If we want the option to store data in encrypted form in remote pluggable storage, we need ways to decrypt/encrypt that data. Adding the initial interfaces that will take care of that

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/DefaultPinotCrypter.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/DefaultPinotCrypter.java
@@ -18,14 +18,14 @@ package com.linkedin.pinot.common.segment.crypt;
 import java.net.URI;
 
 
-public class DefaultPinotCrypter extends PinotCrypter {
+public class DefaultPinotCrypter implements PinotCrypter {
   @Override
-  public void encrypt(URI uri) {
+  public void encrypt(URI uri) throws Exception {
 
   }
 
   @Override
-  public void decrypt(URI uri) {
+  public void decrypt(URI uri) throws Exception {
 
   }
 }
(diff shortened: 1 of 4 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/DefaultPinotCrypter.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.segment.crypt;
+
+import java.net.URI;
+
+
+public class DefaultPinotCrypter extends PinotCrypter {
+  @Override
+  public void encrypt(URI uri) {
```
mayankshriv: Should the apis return Boolean and/or throw exception?

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/PinotCrypter.java
```diff
@@ -0,0 +1,37 @@
...
+package com.linkedin.pinot.common.segment.crypt;
+
+import java.net.URI;
+
+
+/**
+ * The PinotCrypter will encrypt and decrypt segments when they are downloaded. This class is especially useful in cases
+ * where segments cannot be stored unencrypted in storage.
+ */
+public abstract class PinotCrypter {
```
mayankshriv: This should be an interface not abstract class.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/PinotCrypterFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+  // Prevent factory from being instantiated
+  private PinotCrypterFactory() {
+
+  }
+  public static void init(Configuration config) {
+    _crypterClass = config.getString(CRYPTER_CLASS_CONFIG);
+  }
+
+  public static PinotCrypter create() {
```
mayankshriv: Create() in a factory should take an argument on what concrete implementation to instantiate, otherwise it is not a factory pattern.

jenniferdai: We will instantiate the PinotCrypter class at time of starting controllers, servers, minions so that when we download segments, we can decrypt. This class will be wired in by reflection, and during init, we will take that parameter in, so we don't need another parameter during create.

PR #3117, opened 2018-08-27
[TE] frontend - aaronucsd/added 401 error handling and login session …

aaronucsd:
…expired catches

1. Added 401 handling for ember data and non-ember data routes
2. Added better login message for session expired
3. Retained force url or deep links (fromUrl) to support things like bookmarks and emails
4. Added the logic to handle saving intented transition prior to session timed out
5. Added error substates in all routes. Most children route will bubble the error to its parent and
parents to application route.
6. manual tests
7. Important note: We need to still find all fetch() that isnt' using `checkStatus` to use it.
However, it is important to note that `checkStatus` already does the response.json so we do not need to call it explicitly like below.
```
        fetch(url)
          .then(checkStatus)
          .then(res => res.json())
          .then(res => this.set('selectedMetric', res));
```

The line `.then(res => res.json())` needs to be removed.

screen shot 2018-08-27 at 4 01 54 pm


aaronucsd:
Updated all `fetch()` calls to use `checkStatus` to make sure we catch 401 errors.

PR #3118, opened 2018-08-27
Decrypt files during segment download

jenniferdai:
* Add file decryption during segment download so that we can store decrypted files only on local disk and pluggable storage can keep all files as encrypted.

### DIFF
--- a/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
+++ b/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
@@ -22,6 +22,7 @@ import com.linkedin.pinot.common.data.DataManager;
 import com.linkedin.pinot.common.metadata.ZKMetadataProvider;
 import com.linkedin.pinot.common.metadata.segment.OfflineSegmentZKMetadata;
 import com.linkedin.pinot.common.segment.SegmentMetadata;
+import com.linkedin.pinot.common.segment.crypt.PinotCrypterFactory;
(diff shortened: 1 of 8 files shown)


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentFetcherAndLoader.java
```diff
@@ -55,8 +56,10 @@ public SegmentFetcherAndLoader(@Nonnull Configuration config, @Nonnull InstanceD
     Configuration pinotFSConfig = config.subset(CommonConstants.Controller.PREFIX_OF_CONFIG_OF_PINOT_FS_FACTORY);
     Configuration segmentFetcherFactoryConfig =
         config.subset(CommonConstants.Server.PREFIX_OF_CONFIG_OF_SEGMENT_FETCHER_FACTORY);
+    Configuration pinotCrypterConfig = config.subset(CommonConstants.Controller.PREFIX_OF_CONFIG_OF_PINOT_CRYPTER_FACTORY);
 
     PinotFSFactory.init(pinotFSConfig);
+    PinotCrypterFactory.init(pinotCrypterConfig);
```
mayankshriv: Should these factories (FS and Crypter) be initialized in the Controller/Server starters? Why is it being initialized here?

jenniferdai: This is where we initialize pinotfsfactory, segment fetcher, it is called in server starter.
unit tests were added in the last pr for pinot crypter factory initialization

PR #3124, opened 2018-08-29
Refactor ServerQueryRequest to pre-compute and store segment independent information about the request

Jackie-Jiang:
Compile all expressions in aggregation and group-by
Extract columns in filter, aggregation, group-by and filter
Do not cache InstanceRequest
Reduce the work in segment pruner from per segment basis to only once

Next step: use ServerQueryRequest for the query execution instead of BrokerRequest


pinot-common/src/main/java/com/linkedin/pinot/common/utils/request/RequestUtils.java
```diff
@@ -149,18 +151,58 @@ public static FilterQueryTree buildFilterQuery(Integer id, Map<Integer, FilterQu
...
+    Set<String> filterColumns = new HashSet<>();
+    if (root.getChildren() == null) {
+      filterColumns.add(root.getColumn());
+    } else {
+      Stack<FilterQueryTree> stack = new Stack<>();
+      stack.add(root);
+      while (!stack.empty()) {
+        FilterQueryTree node = stack.pop();
+        for (FilterQueryTree child : node.getChildren()) {
+          if (child.getChildren() == null) {
```
kishoreg: check for both null and empty?

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/DataSchemaSegmentPruner.java
```diff
@@ -15,111 +15,38 @@
...
-          }
-        }
-      }
-      return false;
-    }
-    // unsupported query type.
-    return true;
+  @Override
+  public boolean prune(@Nonnull IndexSegment segment, @Nonnull ServerQueryRequest queryRequest) {
+    Set<String> ColumnsInSchema = segment.getSegmentMetadata().getSchema().getColumnNames();
```
kishoreg: ColumnsInSchema --> columnsInSchema

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/DataSchemaSegmentPruner.java
```diff
@@ -15,111 +15,38 @@
...
+    return !(containsAll(ColumnsInSchema, queryRequest.getFilterColumns()) && containsAll(ColumnsInSchema,
```
kishoreg: should we not just construct columnsInQuery and do one check against that?

pinot-common/src/main/java/com/linkedin/pinot/common/utils/request/RequestUtils.java
Jackie-Jiang: **children** is non-null only for AND and OR, and will never be empty

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/DataSchemaSegmentPruner.java
Jackie-Jiang: Fixed

pinot-core/src/main/java/com/linkedin/pinot/core/query/pruner/DataSchemaSegmentPruner.java
Jackie-Jiang: Done

Jackie-Jiang:
@kishoreg For the **Nonnull** and **Nullable** annotations, it's for the clarity of the public APIs. With the annotations, when we call the methods, we know whether it expects a null value, and thus prevent causing NPE.
Also, IntelliJ can give out warnings when it detects potential problems: https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html

PR #3126, opened 2018-08-30
Clean up filter operators, blocks, and evaluators

Jackie-Jiang:
Removed redundant filter blocks
Implement new method "getNum(Non)MatchingDictIds" in base predicate evaluator
Only expose predicate evaluator and data source metadata in scan-based filter
BrokerRequest debug options should be nullable


sunithabeeram:
Thanks for noting the changes in the description. Could you please include the context/motivation for this change?

pinot-core/src/main/java/com/linkedin/pinot/core/startree/operator/StarTreeFilterOperator.java
```diff
@@ -207,8 +205,9 @@ public BaseFilterBlock getNextBlock() {
     } else if (numChildFilterOperators == 1) {
       return childFilterOperators.get(0).nextBlock();
     } else {
-      FilterOperatorUtils.reOrderFilterOperators(childFilterOperators, _debugOptions);
-      return new AndOperator(childFilterOperators).nextBlock();
+      FilterOperatorUtils.reorderFilterOperators(childFilterOperators,
+          FilterOperatorUtils.enableScanReorderOptimization(_debugOptions));
```
sunithabeeram: Reading this in the context of reorderFilterOperators definitely feels odd to me. Why not pass debugOptions and let reorderFilterOperator decide what it wants to look at?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/ScanBasedFilterOperator.java
```diff
@@ -73,32 +77,17 @@ public String getOperatorName() {
...
+  }
 
-    @Override
-    public FilterBlockDocIdSet getFilteredBlockDocIdSet() {
-      return docIdSet;
-    }
+  /**
+   * Returns the metadata of the data source associated with the scan filter.
+   */
+  public DataSourceMetadata getDataSourceMetadata() {
```
sunithabeeram: Just FYI. Kishore had a comment on my earlier PR that this should be changed to ColumnMetadata
https://github.com/linkedin/pinot/pull/3111#discussion_r213012688

kishoreg:
@Jackie-Jiang, thanks for making these changes. One suggestion is to break this PR into 4 different ones - one for each task

- Removed redundant filter blocks
- Implement new method "getNum(Non)MatchingDictIds" in base predicate evaluator
- Only expose predicate evaluator and data source metadata in scan-based filter
- BrokerRequest debug options should be nullable

Overall, I think "Removing redundant filter blocks" is a good change. The new method is nice to have , though I am worried if we start using it in different places without knowing the cost of computing non-matching ids. Other two seems unnecessary right now. @sunithabeeram is trying out various optimizations for scan based filters and we might get rid of debugOptions once we are comfortable.

Jackie-Jiang:
@sunithabeeram The reason why I started these changes is because my IDE has annotation check, and then the tests failed because **debugOptions** is nullable. Some changes are added as I look over the code. Will split into multiple PRs as Kishore suggested.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -58,13 +57,21 @@ public static BaseFilterOperator getLeafFilterOperator(PredicateEvaluator predic
   }
 
   /**
-   * Re-order filter operators based on the their cost. Put the ons with inverted index first so we can process less
+   * Returns whether to use scan re-order optimization (ON by default) based on the given debug options.
+   */
+  public static boolean enableScanReorderOptimization(@Nullable Map<String, String> debugOptions) {
```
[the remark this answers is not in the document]
Jackie-Jiang: I understand why you pass the whole debugOptions instead of just a boolean value. I keep the debugOptions in the operator, but added a method to extract the scanReorderOptimization flag from the debugOptions. For other options, we can add other methods that take debugOptions as input and extract the option from it.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -19,22 +19,21 @@
...
-
-  // Debug option to enable or disable multi-value optimization
-  public static final String USE_SCAN_REORDER_OPTIMIZATION = "useScanReorderOpt";
 
+public class FilterOperatorUtils {
   private FilterOperatorUtils() {
   }
 
+  // Debug option to enable or disable the optimization of reordering the scan filters
+  public static final String SCAN_REORDER_OPTIMIZATION = "scanReorderOpt";
```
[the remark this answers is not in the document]
Jackie-Jiang: I also do that except for util class or singleton class where we need to put a private constructor to prevent people constructing it.

pinot-core/src/main/java/com/linkedin/pinot/core/startree/operator/StarTreeFilterOperator.java
Jackie-Jiang: IMO, reorderFilterOperators method should be independent of debugOptions, and thus make future usage or testing easier.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/ScanBasedFilterOperator.java
Jackie-Jiang: I noticed that, but it's a little bit hard to access ColumnMetadata in this layer. I just kept the current API but only for ScanBasedFilterOperator.

Jackie-Jiang:
Split this PR into #3128 #3129 #3130

PR #3129, opened 2018-08-30
[Clean up] Add default implementation of new API getNum(Non)MatchingDictIds in base class

Jackie-Jiang:
Reduce the extra override methods needed.


sunithabeeram:
@mayankshriv, I introduced the the num-non-matchind method in my previous commit - backed out a change that relied on it to keep the earlier PR simpler, but decided to keep the method exposed. I plan to revisit that soon and will add necessary test coverage then.

PR #3130, opened 2018-08-30
Fix the annotation where debug options should be nullable

Jackie-Jiang:


pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -63,8 +64,9 @@ public static BaseFilterOperator getLeafFilterOperator(PredicateEvaluator predic
    * <p>Special filter operators such as {@link MatchEntireSegmentOperator} and {@link EmptyFilterOperator} should be
    * removed from the list before calling this method.
    */
-  public static void reOrderFilterOperators(List<BaseFilterOperator> filterOperators, Map<String, String> debugOptions) {
-    Collections.sort(filterOperators, new Comparator<BaseFilterOperator>() {
+  public static void reOrderFilterOperators(List<BaseFilterOperator> filterOperators,
```
snleee: I think that `reorderFilterOperators` should be better

Jackie-Jiang: Done

PR #3135, opened 2018-08-31
Invalidate controller leader cached value if exceptions occur during segment completion calls

npawar:
We recently faced an error scenario on perf controller. The leader controller was taken down. The servers cache the leader controller value, until a NOT_LEADER response is received, after which the cached value gets refreshed. 
But in this scenario, since the controller was down, it could not send NOT_LEADER. We threw an exception, but never refreshed the controller leader cached value.
The servers kept trying to send requests to the controller who they thought was the leader, and we never recovered from this. The only way out was a server restart.

Adding logic to invalidate and refresh caches in exception scenarios

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ControllerLeaderLocator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ControllerLeaderLocator.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.server.realtime;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.linkedin.pinot.core.query.utils.Pair;
 import org.apache.helix.AccessOption;
 import org.apache.helix.BaseDataAccessor;
@@ -23,7 +24,6 @@ import org.apache.helix.ZNRecord;
 import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 
 // Helix keeps the old controller around for 30s before electing a new one, so we will keep getting
 // the old controller as leader, and it will keep returning NOT_LEADER.
@@ -39,8 +39,12 @@ public class ControllerLeaderLocator {
   // Co-ordinates of the last known controller leader.
   private Pair<String, Integer> _controllerLeaderHostPort = null;
 
-  // Should we refresh the controller leader co-ordinates?
-  private volatile boolean _refresh = true;
+  // Indicates whether cached controller leader value is invalid
+  private volatile boolean _cachedControllerLeaderInvalid = true;
+  // Time in millis when cache invalidate was last set
+  private volatile long _lastCacheInvalidateMillis = 0;
+  // Minimum millis which must elapse between consecutive invalidation of cache
+  private static final long MILLIS_BETWEEN_INVALIDATE = 30_000;
 
   ControllerLeaderLocator(HelixManager helixManager) {
     _helixManager = helixManager;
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ControllerLeaderLocator.java
```diff
@@ -70,33 +74,67 @@ public static ControllerLeaderLocator getInstance() {
...
+  public synchronized void invalidateCachedControllerLeader() {
+    long millisSinceLastInvalidate = System.currentTimeMillis() - _lastCacheInvalidateMillis;
+    if (millisSinceLastInvalidate < MILLIS_BETWEEN_INVALIDATE) {
+      LOGGER.info(
+          "Millis since last controller cache value invalidate {} is less than allowed frequency {}. Skipping invalidate.",
+          millisSinceLastInvalidate, MILLIS_BETWEEN_INVALIDATE);
+    } else {
+      LOGGER.info("Invalidating cached controller leader value");
+      _cachedControllerLeaderInvalid = true;
+      _lastCacheInvalidateMillis = System.currentTimeMillis();
```
mcvsubbu: You can avoid getting time twice

pinot-core/src/test/java/com/linkedin/pinot/server/realtime/ControllerLeaderLocatorTest.java
```diff
@@ -32,6 +32,66 @@
...
+    // invalidate within {@link ControllerLeaderLocator::getMillisBetweenInvalidate()} millis
+    // values should remain unchanged
+    controllerLeaderLocator.invalidateCachedControllerLeader();
+    Assert.assertFalse(controllerLeaderLocator.isCachedControllerLeaderInvalid());
+    Assert.assertEquals(controllerLeaderLocator.getLastCacheInvalidateMillis(), lastCacheInvalidateMillis);
+    lastCacheInvalidateMillis = controllerLeaderLocator.getLastCacheInvalidateMillis();
+
+    // invalidate after {@link ControllerLeaderLocator::getMillisBetweenInvalidate()} millis have elapsed
+    // cache should be invalidated and last cache invalidation time should get updated
+    Thread.sleep(controllerLeaderLocator.getMillisBetweenInvalidate());
```
mcvsubbu: we should avoid this sleep, and any time-dependent test, since that can cause flaky tests.

kishoreg:
why are we not using ControllerChangeListener provided by Helix? We should get notified when there is a change in controller leadership.

mcvsubbu:
@kishoreg we did this piece of code at the time we implemented LLC. Since we had issues with zk watches (too many of them caused re-connect problems, etc.), the goal in LLC was ZERO zk watches. Along those lines, the segment completion protocol (which is what runs over this particular channel from server to the lead controller) uses this piece of code to locate the lead controller. The segment completion protocol has been designed to to be resilient to controller leadership changes, and a contoller responds with NOT_LEADER message if the server ever gets to the wrong leader. Since contrloller leadership changes are rare, the server just invalidates any cached controller co-ordinates if it gets a NOT_LEADER response.

sunithabeeram:
Are the lead controller coordinates cached in other parts of the code?

mcvsubbu:
@sunithabeeram  no, it is not cached in other parts of the code.

pinot-core/src/main/java/com/linkedin/pinot/server/realtime/ControllerLeaderLocator.java
npawar: Fixed

pinot-core/src/test/java/com/linkedin/pinot/server/realtime/ControllerLeaderLocatorTest.java
npawar: Removed time dependency from tests

pinot-core/src/test/java/com/linkedin/pinot/server/realtime/ControllerLeaderLocatorTest.java
```diff
@@ -32,6 +32,69 @@
...
+
+    // check values at startup
+    Assert.assertTrue(controllerLeaderLocator.isCachedControllerLeaderInvalid());
+    Assert.assertEquals(controllerLeaderLocator.getLastCacheInvalidateMillis(), 0);
+
+    // very first invalidate
+    controllerLeaderLocator.invalidateCachedControllerLeader();
+    Assert.assertTrue(controllerLeaderLocator.isCachedControllerLeaderInvalid());
+    long lastCacheInvalidateMillis = controllerLeaderLocator.getLastCacheInvalidateMillis();
+    Assert.assertTrue(lastCacheInvalidateMillis > 0);
```
mcvsubbu: a better test will be to take the millis before calling invaidate, and assert that the invalidation time is >= that time stamp,

npawar: that is being done for one of the following tests, where we want to test for lastCacheInvalidMillis being updated wrt the older value. For this particular one, I wanted to test that initial value has been updated from 0

PR #3137, opened 2018-09-04
Enhance service status check

Jackie-Jiang:
Refactor ServiceStatus for the following fix and enhancement:
1. When ideal state is null (resource has been removed), should return GOOD (fix)
2. Do not fail when partitionStateMap is empty because the instance might be removed from the resource (fix)
3. Log the correct number of resources left to be checked (fix)
4. Keep only the resources not checked in memory to reduce garbage (enhancement)
5. Log an ERROR message when partition get into ERROR state (enhancement)

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
@@ -16,9 +16,9 @@
 package com.linkedin.pinot.common.utils;
 
 import com.linkedin.pinot.common.config.TableNameBuilder;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -26,7 +26,6 @@ import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixManager;
 import org.apache.helix.PropertyKey;
-import org.apache.helix.PropertyKey.Builder;
 import org.apache.helix.model.CurrentState;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.IdealState;
@@ -34,9 +33,11 @@ import org.apache.helix.model.LiveInstance;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
 /**
  * Utility class to obtain the status of the Pinot instance running in this JVM.
  */
+@SuppressWarnings("unused")
 public class ServiceStatus {
   private static final Logger LOGGER = LoggerFactory.getLogger(ServiceStatus.class);
 
@@ -51,7 +52,9 @@ public class ServiceStatus {
    * Callback that returns the status of the service.
    */
   public interface ServiceStatusCallback {
+
     Status getServiceStatus();
+
     String getStatusDescription();
   }
 
@@ -107,12 +110,14 @@ public class ServiceStatus {
       return Status.GOOD;
     }
 
+    @Override
     public String getStatusDescription() {
       StringBuilder statusDescription = new StringBuilder();
       for (ServiceStatusCallback statusCallback : _statusCallbacks) {
-        statusDescription.append(
-            statusCallback.getClass().getSimpleName() + ":" + statusCallback.getStatusDescription() + ";"
-        );
+        statusDescription.append(statusCallback.getClass().getSimpleName())
+            .append(":")
+            .append(statusCallback.getStatusDescription())
+            .append(";");
       }
       return statusDescription.toString();
     }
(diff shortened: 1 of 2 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
```diff
@@ -123,147 +128,130 @@ public String getStatusDescription() {
...
+              _resourcesToMonitor.add(resourceName);
+              break;
+            }
           }
         }
       }
+      _numTotalResourcesToMonitor = _resourcesToMonitor.size();
 
-      LOGGER.info("Monitoring resources {} for start up of instance {}", _resourcesToMonitor, _instanceName);
+      LOGGER.info("Monitoring {} resources: {} for start up of instance {}", _numTotalResourcesToMonitor,
```
mcvsubbu: Logging of complete resource list was omitted from this log message to save the length of the log line. You may want to log a truncated version of the list/set or not log it at all (i.e. leave the log line as is)

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
```diff
@@ -123,147 +128,130 @@ public String getStatusDescription() {
...
-
-      LOGGER.info("Monitoring resources {} for start up of instance {}", _resourcesToMonitor, _instanceName);
-    }
+      _resourcesToMonitor = new HashSet<>(resourcesToMonitor);
+      _numTotalResourcesToMonitor = _resourcesToMonitor.size();
 
-    public String getStatusDescription() {
-      return _statusDescription;
+      LOGGER.info("Monitoring {} resources: {} for start up of instance {}", _numTotalResourcesToMonitor,
+          _resourcesToMonitor, _instanceName);
```
mcvsubbu: same as above comment. do not log the list/set of resources

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
```diff
@@ -123,147 +128,130 @@ public String getStatusDescription() {
...
+          iterator.remove();
           continue;
         }
 
-        Map<String, String> statePartitionStateMap = getPartitionStateMap(helixState);
-        if(statePartitionStateMap.isEmpty() && !idealState.getPartitionSet().isEmpty()) {
-          _statusDescription = "statePartitionStateMapSize=" + statePartitionStateMap.size() + ", idealStateSize="
-              + idealState.getPartitionSet().size() + completedCountStr;
+        String descriptionSuffix =
+            String.format("(%s, %d/%d left)", resourceName, _resourcesToMonitor.size(), _numTotalResourcesToMonitor);
```
mcvsubbu: If you log the status as a comma-separate "key=value" format, then we can write scripts to parse the logs.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
```diff
@@ -123,147 +128,130 @@ public String getStatusDescription() {
...
           // If the instance state is not ERROR and is not the same as what's expected from the ideal state, then it
           // hasn't finished starting up
-          if (!"ERROR".equals(currentStateStatus) && !idealStateStatus.equals(currentStateStatus)) {
-            _statusDescription = partition +" currentStateStatus=" + currentStateStatus + ", idealStateStatus="
-                + idealStateStatus + completedCountStr;
-            return Status.STARTING;
+          String currentStateStatus = partitionStateMap.get(partitionName);
+          if (!idealStateStatus.equals(currentStateStatus)) {
+            if ("ERROR".equals(currentStateStatus)) {
+              LOGGER.error(String.format("Resource: %s, partition: %s is in ERROR state", resourceName, partitionName));
```
mcvsubbu: This log message will be printed every time we call service status. Not fully convinced that we want that. Hopefully there are enough logs when the partition actually goes to ERROR state.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
Jackie-Jiang: The current one lists the complete list. This one just logs the number of resources as well.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
Jackie-Jiang: Done

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
```diff
@@ -123,147 +128,131 @@ public String getStatusDescription() {
...
+          iterator.remove();
           continue;
         }
 
-        Map<String, String> statePartitionStateMap = getPartitionStateMap(helixState);
-        if(statePartitionStateMap.isEmpty() && !idealState.getPartitionSet().isEmpty()) {
-          _statusDescription = "statePartitionStateMapSize=" + statePartitionStateMap.size() + ", idealStateSize="
-              + idealState.getPartitionSet().size() + completedCountStr;
+        String descriptionSuffix =
+            String.format("(resource=%s, numResourcesLeft=%d, numTotalResources=%d)", resourceName,
```
mcvsubbu: why do we have starting "(" and ending ")"? It will really be useful if we see a log message like:

<some descriptor>:a=x, b=y, c=z, d=uu, ...

then it is easy for us to script looking for certain values or segments that do not load correctly.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
```diff
@@ -123,147 +128,131 @@ public String getStatusDescription() {
...
+                _resourcesToMonitor.size(), _numTotalResourcesToMonitor);
+        T helixState = getState(resourceName);
+        if (helixState == null) {
+          _statusDescription = "Helix state does not exist " + descriptionSuffix;
```
mcvsubbu: Maybe introduce a ":" after "exist"

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
Jackie-Jiang: Done

pinot-common/src/main/java/com/linkedin/pinot/common/utils/ServiceStatus.java
Jackie-Jiang: Done

PR #3139, opened 2018-09-05
[TE] detection - add support for hourly, daily dimension filters

apucher:
* support convenience filter by mean hourly value
* support convenience filter by mean daily value


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -31,23 +34,6 @@
  */
 public class DimensionWrapper extends DetectionPipeline {
 
-  // exploration
```
jihaozh: No more static final constant?

apucher: yeah. doesn't add a lot of value

PR #3140, opened 2018-09-05
Support null values in BYTES data type.

mayankshriv:
1. Use default null value of byte[0] for BYTES data type.
2. Convert ByteBuffer type (from avro) into byte[] transform step,
instead of field extractor step.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
@@ -22,7 +22,6 @@ import com.linkedin.pinot.common.Utils;
 import com.linkedin.pinot.common.config.ConfigKey;
 import com.linkedin.pinot.common.config.ConfigNodeLifecycleAware;
 import com.linkedin.pinot.common.utils.EqualityUtils;
-import com.linkedin.pinot.common.utils.primitive.ByteArray;
 import javax.annotation.Nonnull;
 import org.apache.avro.Schema.Type;
 import org.apache.commons.codec.DecoderException;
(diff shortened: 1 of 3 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
```diff
@@ -154,8 +154,8 @@ private static Object getDefaultNullValue(@Nonnull FieldType fieldType, @Nonnull
           return stringDefaultNullValue;
         case BYTES:
           try {
-            return new ByteArray(Hex.decodeHex(stringDefaultNullValue.toCharArray()));
-          } catch (DecoderException e) {
+            return stringDefaultNullValue.getBytes("UTF-8");
```
Jackie-Jiang: This looks not correct, we still need to decode hex right?

pinot-core/src/main/java/com/linkedin/pinot/core/util/AvroUtils.java
```diff
@@ -291,6 +293,14 @@ public static Object transformAvroValueToObject(Object avroValue, FieldSpec fiel
     }
     if (fieldSpec.getDataType() == FieldSpec.DataType.STRING) {
       return avroValue.toString();
+    } else if (fieldSpec.getDataType() == FieldSpec.DataType.BYTES && avroValue instanceof ByteBuffer) {
+      // Avro ByteBuffer maps to ByteArray
```
Jackie-Jiang: maps to **byte[]**

pinot-core/src/main/java/com/linkedin/pinot/core/util/AvroUtils.java
```diff
@@ -291,6 +293,14 @@ public static Object transformAvroValueToObject(Object avroValue, FieldSpec fiel
...
+      ByteBuffer byteBuffer = (ByteBuffer) avroValue;
+
+      // Assumes byte-buffer is ready to read. Also, avoid getting underlying array, as it may be over-sized.
+      byte[] bytes = new byte[byteBuffer.limit()];
```
Jackie-Jiang: Basically the same, but should use **byteBuffer.remaining()**

pinot-core/src/main/java/com/linkedin/pinot/core/data/extractors/PlainFieldExtractor.java
```diff
@@ -210,15 +209,6 @@ public GenericRow transform(GenericRow row, GenericRow destinationRow) {
             hasError = true;
             _errorCount.put(column, _errorCount.get(column) + 1);
           }
-        } else if (value instanceof ByteBuffer) { // TODO: Need better handle for ByteBuffers.
```
kishoreg: why did we remove this?

mayankshriv: This is moved to transform layer in AvroUtils.

PR #3144, opened 2018-09-05
Move to Ted Dunning's implementation of TDigest.

mayankshriv:
Moved from streamlib's implementation of TDigest to Ted Dunning's, as the former does not handle empty TDigest objects, and has poor performance as compared to the latter. to Move to Ted Dunning's implementation of TDigest.


mayankshriv:
@mcvsubbu This feature has not been used yet, and the first usage showed the issue with original library. So we are intentionally making this backward incompatible change.

PR #3147, opened 2018-09-06
[TE] detection - update migration endpoint for unified injection

apucher:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/LegacyAnomalyFunctionTranslator.java
```diff
@@ -26,83 +19,44 @@
...
-    Map<String, String> alertFilter = anomalyFunctionDTO.getAlertFilter();
-    if (alertFilter != null) {
-      Map<String, Object> alertFilterProp = new HashMap<>();
-      alertFilterProp.put(PROP_CLASS_NAME, LegacyAlertFilterWrapper.class);
-      String alertFilterType = alertFilter.get("type");
-
-      if (alertFilterType != null) {
-        String alertFilterClassName = this.alertFilterFactory.getClassNameForAlertFilterType(alertFilterType.toUpperCase());
-        if (alertFilterClassName != null) {
-          alertFilterProp.put(PROP_LEGACY_ALERT_FILTER_CLASS_NAME, alertFilterClassName);
```
jihaozh: Looks like still need an alert filter class name for the regression alert filter to work.

apucher: good catch. I re-introduced the logic for this. Also, there's #3149 to add support for global impact filters to the legacy emulation layer

PR #3153, opened 2018-09-07
Refactoring Pinot segment upload

jenniferdai:
* Refactoring the upload code to be a set of interfaces that have different implementations across different upload types and universal utils classes that are used for all types.
* Makes it easier to add new features, such as ways to encrypt/decrypt segments.
* Added an integration test for URI upload. Note that this upload method intentionally behaves differently from the old URI upload in that it always will move the segment to a controller-dictated final directory.
* URI upload will always go through the JSON content-part endpoint since the multipart endpoint will throw errors
* Changed the current multipart endpoint and json segment upload endpoints to go through my new upload method.

* Future work
1. Add metadata upload type as an optimization
2. In this PR, I removed support for segment names with strange characters. The reason for this is because we are now using a URI abstraction. That means that the final segment resting place will have to fit into the URI abstraction and will require the filename to be encoded as well. Will need to do a bigger change to support that, aka, store segments as their encoded names both in zk metadata and in their file paths/look into that more and did not want to do that as part of this pr.


pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/PinotCrypterFactory.java
```diff
@@ -24,25 +23,20 @@
...
   }
-  public static void init(Configuration config) {
-    _crypterClass = config.getString(CRYPTER_CLASS_CONFIG);
-  }
 
-  public static PinotCrypter create() {
-    String crypterClassName = _crypterClass;
+  public static PinotCrypter create(String crypterClassName) {
     try {
       LOGGER.info("Instantiating PinotCrypter class {}", crypterClassName);
```
mcvsubbu: You may want to also consider an init() method into the crypter class, perhaps with a config class. In that case, an init() method may have to be added to the interface. Alternatively, consider implementing an abstract base class that you can add default methods in the future, and inheriting from that one.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/PinotCrypterFactory.java
```diff
@@ -24,25 +23,20 @@
...
       PinotCrypter pinotCrypter =  (PinotCrypter) Class.forName(crypterClassName).newInstance();
       return pinotCrypter;
     } catch (Exception e) {
+      LOGGER.warn("Unable to instantiate {}, creating default crypter {}", crypterClassName, DefaultPinotCrypter.class.getName());
```
mcvsubbu: Log the exception here

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcher.java
```diff
@@ -15,11 +15,10 @@
  */
 package com.linkedin.pinot.common.segment.fetcher;
 
+import java.io.File;
 import java.util.Set;
 import org.apache.commons.configuration.Configuration;
 
-import java.io.File;
```
mcvsubbu: Can you please add some javadocs to the interface methods? If I understand right, we will have more implementations of this interface across file systems, right? It will be really useful to implementors if the expected behavior is documented clearly. Thanks.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
```diff
@@ -42,14 +42,15 @@ public static SegmentFetcherFactory getInstance() {
   }
 
   public static final String PROTOCOLS_KEY = "protocols";
-  public static final List<String> DEFAULT_PROTOCOLS = Collections.unmodifiableList(Arrays.asList("file", "http"));
+  public static final List<String> DEFAULT_PROTOCOLS = Collections.unmodifiableList(Arrays.asList("file", "http", "multipart"));
```
mcvsubbu: not sure I understand the difference between "http" and "multipart". Should we call it multipart-http? Is the idea of introducing a multi-part fetcher to enable fetchign multiple segments in one go?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -244,74 +250,123 @@ public SuccessResponse deleteAllSegments(
...
+
+    File tempEncryptedFile = null;
+    File tempDecryptedFile = null;
+    File tempSegmentDir = null;
     try {
-      asyncResponse.resume(uploadSegmentInternal(multiPart, null, enableParallelPushProtection, headers, request, false));
-    } catch (Throwable t) {
-      asyncResponse.resume(t);
+      FileUploadPathProvider provider = new FileUploadPathProvider(_controllerConf);
+      String tempFileName = "tmp-" + System.nanoTime();
```
mcvsubbu: Do we want to add thread-ID as well here for uniqueness? We do need to be careful about file name length, though

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -244,74 +250,123 @@ public SuccessResponse deleteAllSegments(
...
+  private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enableParallelPushProtection, HttpHeaders headers, Request request) throws Exception {
+    // Get upload type
+    String uploadTypeStr = headers.getHeaderString(FileUploadDownloadClient.CustomHeaders.UPLOAD_TYPE);
+    FileUploadDownloadClient.FileUploadType uploadType = getUploadType(uploadTypeStr);
+
+    // Get crypter class
+    String crypterClassHeader = headers.getHeaderString(FileUploadDownloadClient.CustomHeaders.CRYPTER);
+
+    // Get Download URI
+    String downloadURI = headers.getHeaderString(FileUploadDownloadClient.CustomHeaders.DOWNLOAD_URI);
```
mcvsubbu: Not sure if this header can be changed, but if we are re-architecting things consider making  it SEGMENT_LOCATION instead of DOWNLOAD_URI

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -244,74 +250,123 @@ public SuccessResponse deleteAllSegments(
...
+            crypterClassHeader = DefaultPinotCrypter.class.getName();
+          }
+          encryptedObject = tempEncryptedFile;
+          break;
+        case TAR:
+          // In the case of TAR, we will use local file
+          if (crypterClassHeader == null) {
+            crypterClassHeader = MultipartPinotCrypter.class.getName();
+          }
+          downloadURI = "multipart://dummy";
```
mcvsubbu: ??

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -244,74 +250,123 @@ public SuccessResponse deleteAllSegments(
...
-  @ManagedAsync
-  @Produces(MediaType.APPLICATION_JSON)
-  @Consumes(MediaType.MULTIPART_FORM_DATA)
-  @Path("/segments")
-  @ApiOperation(value = "Upload a segment", notes = "Upload a segment as binary")
-  public void uploadSegmentAsMultiPart(FormDataMultiPart multiPart,
-      @ApiParam(value = "Whether to enable parallel push protection") @DefaultValue("false") @QueryParam(FileUploadDownloadClient.QueryParameters.ENABLE_PARALLEL_PUSH_PROTECTION) boolean enableParallelPushProtection,
-      @Context HttpHeaders headers, @Context Request request, @Suspended final AsyncResponse asyncResponse) {
+  private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enableParallelPushProtection, HttpHeaders headers, Request request) throws Exception {
+    // Get upload type
```
mcvsubbu: Do we still print all the headers on the upload, or is that  happening some place else?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -244,74 +250,123 @@ public SuccessResponse deleteAllSegments(
...
-  @POST
-  @ManagedAsync
-  @Produces(MediaType.APPLICATION_JSON)
-  @Consumes(MediaType.MULTIPART_FORM_DATA)
-  @Path("/segments")
-  @ApiOperation(value = "Upload a segment", notes = "Upload a segment as binary")
-  public void uploadSegmentAsMultiPart(FormDataMultiPart multiPart,
-      @ApiParam(value = "Whether to enable parallel push protection") @DefaultValue("false") @QueryParam(FileUploadDownloadClient.QueryParameters.ENABLE_PARALLEL_PUSH_PROTECTION) boolean enableParallelPushProtection,
-      @Context HttpHeaders headers, @Context Request request, @Suspended final AsyncResponse asyncResponse) {
+  private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enableParallelPushProtection, HttpHeaders headers, Request request) throws Exception {
```
mcvsubbu: I suppose we intend to have (eventually) the realtime segment do the "upload" step via this method. Let us think through if there are any architectural changes needed for that. We can implement that eventually.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -244,74 +250,123 @@ public SuccessResponse deleteAllSegments(
...
+      String finalSegmentPath = StringUtil.join("/", provider.getBaseDataDirURI().toString(), segmentMetadata.getTableName(),
+          URLEncoder.encode(segmentName, "UTF-8"));
+      URI finalSegmentLocationURI = new URI(finalSegmentPath);
+      ZKOperator zkOperator = new ZKOperator(_pinotHelixResourceManager, _controllerConf, _controllerMetrics);
+      zkOperator.completeSegmentOperations(segmentMetadata, finalSegmentLocationURI, tempDecryptedFile, enableParallelPushProtection, headers,
+          provider);
+
+      return new SuccessResponse("Successfully uploaded segment: " + segmentMetadata.getName() + " of table: " + segmentMetadata.getTableName());
+    } catch (WebApplicationException e) {
+      throw e;
```
mcvsubbu: Utils.rethrow?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -244,74 +250,123 @@ public SuccessResponse deleteAllSegments(
...
-  @ApiOperation(value = "Upload a segment", notes = "Upload a segment as json")
-  // TODO Does it even work if the segment is sent as a JSON body? Need to compare with the other API
-  public void uploadSegmentAsJson(String segmentJsonStr,    // If segment is present as json body
+  @ApiOperation(value = "Upload a segment", notes = "Upload a segment as binary")
+  public void uploadSegmentAsMultiPart(FormDataMultiPart multiPart,
       @ApiParam(value = "Whether to enable parallel push protection") @DefaultValue("false") @QueryParam(FileUploadDownloadClient.QueryParameters.ENABLE_PARALLEL_PUSH_PROTECTION) boolean enableParallelPushProtection,
       @Context HttpHeaders headers, @Context Request request, @Suspended final AsyncResponse asyncResponse) {
     try {
-      asyncResponse.resume(uploadSegmentInternal(null, segmentJsonStr, enableParallelPushProtection, headers, request, false));
+//      asyncResponse.resume(uploadSegmentInternal(multiPart, null, enableParallelPushProtection, headers, request, false));
```
mcvsubbu: pls remove

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -0,0 +1,195 @@
...
+    ZNRecord znRecord = _pinotHelixResourceManager.getSegmentMetadataZnRecord(offlineTableName, segmentName);
+
+    // Brand new segment, not refresh, directly add the segment
+    if (znRecord == null) {
+      LOGGER.info("Adding new segment: {}", segmentName);
+      try {
+        moveSegmentToPermanentDirectory(currentSegmentLocation, finalSegmentLocationURI);
+        LOGGER.info("Moved segment {} from temp location {} to {}", segmentName, currentSegmentLocation.getAbsolutePath(), finalSegmentLocationURI.getPath());
+      } catch (Exception e) {
+        LOGGER.error("Could not move segment {} from table {} to permanent directory", segmentName, rawTableName);
```
mcvsubbu: add exception to the log

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -0,0 +1,195 @@
...
+        throw new RuntimeException(e);
+      }
+      if (finalSegmentLocationURI.getScheme().equals("file")) {
```
mcvsubbu: Don't we need to set the download URI irrespective of whether the scheme is "file"?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -0,0 +1,195 @@
...
+  public void completeSegmentOperations(SegmentMetadata segmentMetadata, URI finalSegmentLocationURI, File currentSegmentLocation, boolean enableParallelPushProtection, HttpHeaders headers,
+      FileUploadPathProvider provider) throws Exception {
+    String rawTableName = segmentMetadata.getTableName();
+    String offlineTableName = TableNameBuilder.OFFLINE.tableNameWithType(rawTableName);
+    String segmentName = segmentMetadata.getName();
+
+    ZNRecord znRecord = _pinotHelixResourceManager.getSegmentMetadataZnRecord(offlineTableName, segmentName);
+
+    // Brand new segment, not refresh, directly add the segment
+    if (znRecord == null) {
```
mcvsubbu: May make it more readable if you break out the 'if' part into a method, and return ?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -0,0 +1,195 @@
...
+          throw new RuntimeException(
+              "Failed to update ZK metadata for segment: " + segmentName + " of table: " + offlineTableName);
+        }
+      } else {
+        // New segment is different with the existing one, update ZK metadata and refresh the segment
+        LOGGER.info("New segment crc {} is different than the existing segment crc {}. Updating ZK metadata and refreshing segment {}",
+            newCrc, existingCrc, segmentName);
+        moveSegmentToPermanentDirectory(currentSegmentLocation, finalSegmentLocationURI);
+        LOGGER.info("Moved segment {} from temp location {} to {}", segmentName, currentSegmentLocation.getAbsolutePath(), finalSegmentLocationURI.getPath());
+        if (finalSegmentLocationURI.getScheme().equals("file")) {
```
mcvsubbu: Can we have a static final String "file" some place?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -0,0 +1,195 @@
...
+import org.apache.helix.ZNRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The ZKOperator is a util class that is used during segment upload to set relevant metadata fields in zk. It will currently
+ * also perform the data move. In the future when we introduce versioning, we will decouple these two steps.
+ */
+public class ZKOperator {
```
mcvsubbu: Breaking this logic into a separate class seems to be the right direction. Good job!

Perhaps we can consider sub-classing this to handle different variations, not sure.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -0,0 +1,195 @@
...
+        String downloadUrl = ControllerConf.constructDownloadUrl(rawTableName, segmentName, provider.getVip());
+        _pinotHelixResourceManager.addNewSegment(segmentMetadata, downloadUrl);
+
+      } else {
+        _pinotHelixResourceManager.addNewSegment(segmentMetadata, finalSegmentLocationURI.toString());
+      }
+      return;
+    }
+
+    LOGGER.info("Segment {} already exists, refreshing if necessary", segmentName);
```
mcvsubbu: Will be really nice if we can break parts of the code below into different methods to make this method shorter. Or, sub-class of ZkOperator class.

pinot-core/src/main/java/com/linkedin/pinot/core/metadata/MetadataProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface MetadataProvider {
+  /**
+   * Returns a SegmentMetadata object from a tarred file
+   * @param tarredSegmentFile
+   * @param unzippedSegmentDir
+   * @return
+   * @throws Exception
+   */
+  SegmentMetadata extractMetadata(File tarredSegmentFile, File unzippedSegmentDir) throws Exception;
```
mcvsubbu: Do we expect more classes for this interface? What else may they be doing? Looks to me that this method is pretty specific and does not warrant an interface unless I am missing something. Is the unzippedSegmentDir expected to be used as a tempDir? Or, is it that the method uses ONE of tarred file or unzippedsegmentdir to extract metadata? Some javadocs are needed, esp. if this is to be an interface.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/ClusterIntegrationTestUtils.java
```diff
@@ -247,7 +247,9 @@ public void run() {
                 SegmentTestUtils.getSegmentGeneratorConfig(avroFile, outputDir, TimeUnit.DAYS, tableName, pinotSchema);
 
             // Test segment with space and special character in the file name
-            segmentGeneratorConfig.setSegmentNamePostfix(segmentIndex + " %");
+            // TODO: Add back. Cannot support this right now because we have switched to URIs for paths and uri paths
```
mcvsubbu: I don't understand the comment, but if you are planning to add back either later in this commit or a commit following pretty soon, I am ok with this. Please do open an issue. Otherwise, I suggest adding support in this commit.

pinot-core/src/main/java/com/linkedin/pinot/core/metadata/MetadataProvider.java
jenniferdai: I want the segment upload to be as clean as possible and that means that the whole chain will have the same behavior with different classes. The other implementation once we add metadata upload is for the MetadataProvider to simply unzip return the metadata, similar to a noop.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
jenniferdai: I currently do not anticipate any other variations to ZK operations during upload but am open to discussing if you feel there is a need.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/PinotCrypter.java
```diff
@@ -24,16 +24,18 @@
  */
 public interface PinotCrypter {
   /**
-   * Encrypts data.
-   * @param uri location of data to be encrypted
-   * @Exception throws an Exception if encrypt fails
+   * Encrypts the object into the file location provided
+   * @param decryptedObject
```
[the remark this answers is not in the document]
jenniferdai: I like the File idea, but I had to make this a generic object to support multipart. Because multipart is also a type of decrypting, and the lowest common denominator class is Object (there's no way to turn multipart into a file without actually extracting and we want to do that step in the crypter), it became an object.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
jenniferdai: The special case scenario for file is that we will set the downloadUri to the http segment endpoint we download from. Otherwise, we will construct it from controller data dir (with remote storage scheme) and download from there. The downloadUri has been set in the previous calling method. I wanted to put it there because in my mind the controller should come up with that location. However, this downloadUri is reset here because we need to move the final to the constructed final dir first.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jenniferdai: they have returned, good catch.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/MultipartPinotCrypter.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @Override
+  public void decrypt(Object encryptedObject, File decryptedFile) throws Exception {
+    if (! (encryptedObject instanceof FormDataMultiPart)) {
+      throw new RuntimeException("Expected multipart object for decryption");
+    }
+
+    FormDataMultiPart multiPart = (FormDataMultiPart) encryptedObject;
+    // Read segment file or segment metadata file and directly use that information to update zk
+    Map<String, List<FormDataBodyPart>> segmentMetadataMap = multiPart.getFields();
```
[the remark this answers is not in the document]
jenniferdai: I might have misunderstood/missed something, but this is an exact copy of our current logic.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/MultipartPinotCrypter.java
```diff
@@ -0,0 +1,79 @@
...
+    if (!validateMultiPart(segmentMetadataMap, null)) {
```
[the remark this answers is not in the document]
jenniferdai: I might have misunderstood/missed something, but this is an exact copy of our current logic.

pinot-common/pom.xml
```diff
@@ -297,6 +297,10 @@
       <groupId>com.linkedin.pinot</groupId>
       <artifactId>pinot-filesystem</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.glassfish.jersey.media</groupId>
+      <artifactId>jersey-media-multipart</artifactId>
```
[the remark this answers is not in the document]
jenniferdai: It's for the metadata provider, which I had to put here because all the other segment metadata classes are here too. It was throwing dependency errors in core

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/MultipartPinotCrypter.java
```diff
@@ -0,0 +1,79 @@
...
+import org.glassfish.jersey.media.multipart.FormDataBodyPart;
+import org.glassfish.jersey.media.multipart.FormDataMultiPart;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * The MultiPartPinotCrypter takes in a multipart object and extracts the tarred segment file.
+ */
+public class MultipartPinotCrypter implements PinotCrypter {
```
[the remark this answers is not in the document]
jenniferdai: This one is untarring and unzipping the file in the decrypt call.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/SegmentFetcherFactory.java
jenniferdai: I see, what would you suggest? Not sure I can think of a better one.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -244,74 +250,123 @@ public SuccessResponse deleteAllSegments(
...
+          if (downloadURI == null || downloadURI.isEmpty()) {
+            throw new ControllerApplicationException(LOGGER, "Failed to get downloadURI, needed for URI upload",
+                Response.Status.BAD_REQUEST);
+          }
+          if (crypterClassHeader == null) {
+            crypterClassHeader = DefaultPinotCrypter.class.getName();
+          }
+          encryptedObject = tempEncryptedFile;
+          break;
+        case TAR:
```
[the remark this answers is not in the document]
jenniferdai: No, the metadata one is going to come later. Did not want to make this very long.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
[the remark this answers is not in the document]
jenniferdai: We want to plug all upload methods into a SegmentFetcher as we discussed during the design, so that's why I've created a multipart one that is a noop.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/MultipartPinotCrypter.java
```diff
@@ -0,0 +1,79 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(MultipartPinotCrypter.class);
+
+  @Override
+  public void encrypt(Object decryptedObject, File encryptedFile) throws Exception {
+    throw new UnsupportedOperationException("Multipart Pinot Crypter does not support encryption");
```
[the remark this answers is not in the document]
jenniferdai: I don't think we need it because we would never want to transform a pinot segment to a multipart in our code I think but because we have a crypter (we want the interface to support decrypt and encrypt), I left the encrypt method there.

pinot-common/src/main/java/com/linkedin/pinot/common/segment/crypt/MultipartPinotCrypter.java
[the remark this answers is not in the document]
mcvsubbu: just log it as part name instead

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
mcvsubbu: check if you can use File.createTempFile() if possible.

pinot-core/src/main/java/com/linkedin/pinot/core/metadata/MetadataProvider.java
mcvsubbu: In that case, perhaps rename 'tarredSegmentFile' to 'tarFile' and 'unzippedSegmentDir' to 'workingDir'?

mcvsubbu: Suggest not getting class by name, but constructing the exact MetadataExtractor class based on the upload type. Suggest rename MetadataProvider to MeadataExftractor.

pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/ClusterIntegrationTestUtils.java
mcvsubbu: As discussed offline, removing support for special chars will break some use cases. So we need to find a way to add that back.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/SegmentValidator.java
```diff
@@ -0,0 +1,145 @@
...
+import org.joda.time.Interval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * SegmentValidator is a util class used during segment upload. It does verification such as a quota check and validating
+ * that the segment time values stored in the segment are valid.
+ */
+public class SegmentValidator {
```
mayankshriv: Let's ensure that there are tests (existing or new) that protect against any accidental bug due to this refactor.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -0,0 +1,212 @@
...
+  private void processNewSegment(SegmentMetadata segmentMetadata, URI finalSegmentLocationURI,
+      File currentSegmentLocation, FileUploadPathProvider provider, String rawTableName, String segmentName) {
+    try {
+      moveSegmentToPermanentDirectory(currentSegmentLocation, finalSegmentLocationURI);
+      LOGGER.info("Moved segment {} from temp location {} to {}", segmentName, currentSegmentLocation.getAbsolutePath(), finalSegmentLocationURI.getPath());
+    } catch (Exception e) {
+      LOGGER.error("Could not move segment {} from table {} to permanent directory", segmentName, rawTableName, e);
+      throw new RuntimeException(e);
+    }
+    if (finalSegmentLocationURI.getScheme().equals(LOCAL_SEGMENT_SCHEME)) {
```
mayankshriv: I think we can get rid of this 'if' by introducing another parameter. Basically, ZKOperator will need the following:
1. Metadata
2. sourceUri: URI of the source
3. downloadURI: URI where servers download segments from
4. storageURI: URI where segments are stored.
3 and 4 can be same in all cases except for segment upload (with data).

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
mcvsubbu: Will be good if we can add some unit tests to cover the various conditions that this class handles. Not all of it can be covered in integration tests, I believe.

jenniferdai: I think this should be done in integration tests so that we can verify data as well on the server side. We have integration tests to cover the various conditions (different refresh cases in UploadRefreshIntegrationTest, etc)

PR #3168, opened 2018-09-11
[TE] Add support for TDigest metrics in rca

dylancharris:
* add enum values for supported TDigest pinot function syntax
* add logic to reduce TDigests in thirdeye dashboard app by using a count-weighted average (since pinot returns doubles for tdigest function aggregations)

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auto/onboard/ConfigGenerator.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/auto/onboard/ConfigGenerator.java
@@ -93,8 +93,7 @@ public class ConfigGenerator {
       metricConfigDTO.setDefaultAggFunction(MetricConfigBean.DEFAULT_TDIGEST_AGG_FUNCTION);
       metricConfigDTO.setDatatype(MetricType.DOUBLE);
     } else {
-      MetricType dataType = MetricType.valueOf(dataTypeStr);
-      metricConfigDTO.setDatatype(dataType);
+      metricConfigDTO.setDatatype(MetricType.valueOf(dataTypeStr));
     }
 
     return metricConfigDTO;


dylancharris:
That is a great idea. Did not think about the use case where we may use PCT queries on other datasets using the same MetricAggFunction class. I'll see about fixing that!

kishoreg:
+1 to what Alex said. Another question is how do we know which function to use based on the metric? Are we going to configure a different metric for each percentile and add it to the same dataset?

dylancharris:
Added the suggestion from Alex to make the MetricAggFunction generic for other data sources. Then added the MetricType file which I actually forgot from the original commit. Couple open questions here regarding MetricType that I will investigate:

1. What's the source of truth for byte size on Pinot BYTES columns
2. What should the toNumber method return?

dylancharris:
> +1 to what Alex said. Another question is how do we know which function to use based on the metric? Are we going to configure a different metric for each percentile and add it to the same dataset?

My thinking is the metric will specify the MetricAggFunction to use. I will have another PR to auto-onboard the BYTES columns to use the "PCTXX" MetricAggFunction types. Then the conversion code in PqlUtils will convert that to the function expected by Pinot. 

In terms of "which percentile to use," the idea was to use a single percentile as "default" and improve it in the future to configure percentile from the client. What do you think, Kishore & Alex?

apucher:
@kishoreg atm the only way to go about this in TE is setting up separate metrics. We may be able to enable flexible agg functions on the UI but we're already over-committed
@dylancharris afaik the BYTES MetricType isn't used by pinot data source. The result in TE is always either a double, long, or string.

kishoreg:
@apucher this should be similar to configuring a derived metric in ThirdEye right? for example latency_99 -> PCT95(latency) and so on. 

latency_99, latency_95 ... can be part of the same dataset so that they can be used in RCA

apucher:
They would have to me **native** metrics as derived metrics are computed in TE directly

dylancharris:
@apucher, thats true the pinot data source does not use them at the moment. However, BYTES type may be needed because tdigest columns are, in effect, bytearrays. without adding that enum val to MetricType, we will have tdigest metric configs with DOUBLE or LONG type and will not know whether to use SUM as the aggFunction or one of the percentileTDigestXX functions.

I guess the alternative would be to override a metric's defaultAggFunction when the metric configs are created/updated and preserve the LONG/DOUBLE type. This would work, but may imply that MetricType is supposed to represent the return value of the metric column _after aggregation_ rather than before. Is that the intention?

apucher:
@dylancharris Making MetricType part of ThirdEye's MetricConfig is really unnecessary. MetricType serves as configuration for MetricTimeSeries only - a predecessor to DataFrame. MetricTimeSeries was used to move tabular data around TE and all of this data **has already been processed** and aggregated by the data source.

Whenever we onboard metrics and datasets from Pinot schemas, we need to set the defaultAggFunction in the config correctly, based on each field's data type. The confusing part here is that MetricType uses the exact same names as the DataType in the FieldSpec. The FieldSpec's DataType however describes the data type **before** aggregation, whereas MetricType the one **after** aggregation. The MetricConfig in ThirdEye only cares about the latter, and in fact only handle numeric (long, double) values.

dylancharris:
> The confusing part here is that MetricType uses the exact same names as the DataType in the FieldSpec. The FieldSpec's DataType however describes the data type before aggregation, whereas MetricType the one after aggregation.

Thanks for the context, that's what I was missing - I will remove the MetricType now!

dylancharris:
Added the auto-onboarding logic. Default Percentile agg function is P90

apucher:
@jihaozh can you have a quick look as well?

PR #3170, opened 2018-09-12
[TE] report missing anomaly selector in dashboard

newsummit:
Added links to report a missing anomaly for each anomaly set by metric in home dashboard. In this first iteration, we link users to the corresponding Alert Page with report modal opened, where they can report the anomaly.
screen shot 2018-09-11 at 6 45 41 pm

- Made sure back button nav works as expected after user goes to Alert Page
- Took care of outlier edge case for baseline values in dashboard (`utils/utils.js : humanizeFloat`)


apucher:
Thanks for the image. I'd somehow indicate to the user that clicking the "GO" button changes the underlying page, rather than just open a modal. The alternative is to enable the modal to stand alone.

newsummit:
@apucher, I added an indication that you're going to the Alert Page. I'm consulting with our designer for ideas on this. The next iteration should be the standalone modal.

apucher:
awesome. thanks, steve

PR #3174, opened 2018-09-12
[TE] YAML endpoint fix

jihaozh:
Minor fix for YAML endpoint.


jihaozh:
Minor fix for YAML endpoint.

PR #3180, opened 2018-09-14
[PINOT-7011] Log errors caught during query processing

sunithabeeram:
Most server query processing errors get sent back to the broker; the broker logging may not always have the server context when it logs the errors. This change logs the errors locally on the server as well.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QueryScheduler.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QueryScheduler.java
@@ -122,6 +122,8 @@ public abstract class QueryScheduler {
     try {
       dataTable = queryExecutor.processQuery(queryRequest, executorService);
     } catch (Exception e) {
+      LOGGER.error("Encountered exception while processing requestId {} from broker {}",
+          queryRequest.getRequestId(), queryRequest.getBrokerId(), e);
       // For not handled exceptions
       serverMetrics.addMeteredGlobalValue(ServerMeter.UNCAUGHT_EXCEPTIONS, 1);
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QueryScheduler.java
```diff
@@ -122,6 +122,8 @@ public void stop() {
     try {
       dataTable = queryExecutor.processQuery(queryRequest, executorService);
     } catch (Exception e) {
+      LOGGER.error("Encountered exception while processing requestId {} from broker {}",
```
kishoreg: anyway we can rate limit this?

sunithabeeram: I did consider that, and arrived at this change for the following reasons:
- We have OmitStackTraceInFastThrow enabled, so hopefully that will control some of the damage.
- Broker currently logs the error for this (except that I couldn't find a good way to include server context there) and I dont' see it being a very high rate. This still doesn't mean we won't flood the logs IF something bad tickles the server, but I am guessing the chances of that are low.
- The linked jira has context on the specific error and I could log the error in processQuery for the specific case we are chasing. It required converting a precondition check into a log and throw, which seemed clunky to me and having the caller take care of logging exceptions seemed like a better approach.

Let me know if you think its better to track counts per interval and only log them once in a while. I am however not sure how useful that would be in debugging issues.

sunithabeeram: Will merge for now and update with rate limiting as necessary.

PR #3182, opened 2018-09-14
[TE] Fix thirdeye-frontend build failure on local machine

akshayrai:


apucher:
The Ember moves in mysterious ways

PR #3185, opened 2018-09-14
Update pom.xml

pizzaz93:


kishoreg:
@pizzaz93 Did you face any issues with previous netty version? Can you add some description on why we need to upgrade?

pizzaz93:
The newer version would provide better security coverage

PR #3191, opened 2018-09-18
[TE] Stage interfaces for detection pipeline

jihaozh:
- Add the anomaly detector/anomaly filter/grouper stage interfaces for detection pipeline.
- High-level static interface for each stage above.
- Add the wrappers for those interfaces to run in the new pipeline. 
- Adapt threshold and sitewide impact anomaly filters to use the new stage interface.
- Adapt YAML translator to stage interfaces.
- Testing


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/algorithm/stage/AnomalyDetectionStage.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface AnomalyDetectionStage extends BaseDetectionStage {
+  /**
+   * Run detection in the specified time range and return a list of anomalies
+   * @param startTime start time stamp
+   * @param endTime end time stamp
+   * @param provider centralized data source for time series, anomalies, events, etc.
+   * @return list of anomalies
+   */
+  List<MergedAnomalyResultDTO> runDetection(long startTime, long endTime, DataProvider provider);
```
kishoreg: is the interface an algorithm has to implement? why are we providing access to the dataProvider to the underlying implementation by default? Shouldn't the framework do the work to fetch the data and invoke the algorithm with the data?

jihaozh: We can have a higher level API where the user only needs to define what data it needs to fetch and the framework will fetch the data and then run the stage with the data.

kishoreg:
What would that look like?

jihaozh:
It would look like this. For example, the detection stage will have two methods:

  /**
   * Returns a data model describing all required data(time series, aggregates, existing anomalies) to perform a stage.
   * Data is retrieved in one pass and cached between executions if possible.
   * @param startTime start time of detection window
   * @param endTime end time of detection window
   * @return detection data model
   */
  StaticDetectionPipelineModel getModel(long startTime, long endTime);

  /**
   * Run detection in the specified time range and return a list of anomalies
   * @param data data(time series, anomalies, etc.) as described by data model
   * @return list of anomalies
   */
  List<MergedAnomalyResultDTO> runDetection(StaticDetectionPipelineData data);

In the first method, the user defines what data they need to fetch for that stage, like the time range for the time series, aggregates, existing anomalies, events. The framework will fetch the data required by the user and then run the stage with the data.

Very similar to our getTimeSeriesIntervals, analyze in the current interface: https://github.com/linkedin/pinot/blob/master/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomalydetection/function/AnomalyDetectionF
unction.java. But the new interfaces add more flexibility to fetch aggregate values and holiday events and fit into the new pipeline.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/InputDataSpec.java
```diff
@@ -21,20 +21,23 @@
 import java.util.Collections;
 
 
-public class StaticDetectionPipelineModel {
+/**
+ * The data spec to describe all the input data for a detection stage perform the detection.
+ */
+public class InputDataSpec {
   final Collection<MetricSlice> timeseriesSlices;
```
kishoreg: can you add java doc for each input? what is aggregateSlice

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/StaticDetectionPipeline.java
```diff
@@ -52,32 +52,32 @@ protected StaticDetectionPipeline(DataProvider provider, DetectionConfigDTO conf
...
-    Map<MetricSlice, DataFrame> aggregates = this.provider.fetchAggregates(model.aggregateSlices, Collections.<String>emptyList());
-    Multimap<AnomalySlice, MergedAnomalyResultDTO> anomalies = this.provider.fetchAnomalies(model.anomalySlices);
-    Multimap<EventSlice, EventDTO> events = this.provider.fetchEvents(model.eventSlices);
+    InputDataSpec dataSpec = this.getInputDataSpec();
+    Map<MetricSlice, DataFrame> timeseries = this.provider.fetchTimeseries(dataSpec.timeseriesSlices);
+    Map<MetricSlice, DataFrame> aggregates = this.provider.fetchAggregates(dataSpec.aggregateSlices, Collections.<String>emptyList());
+    Multimap<AnomalySlice, MergedAnomalyResultDTO> anomalies = this.provider.fetchAnomalies(dataSpec.anomalySlices);
+    Multimap<EventSlice, EventDTO> events = this.provider.fetchEvents(dataSpec.eventSlices);
 
     StaticDetectionPipelineData data = new StaticDetectionPipelineData(
```
kishoreg: we should rename staticDetectionPipelineData as well right?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/StaticDetectionPipeline.java
```diff
@@ -52,32 +52,32 @@ protected StaticDetectionPipeline(DataProvider provider, DetectionConfigDTO conf
...
   @Override
   public final DetectionPipelineResult run() throws Exception {
-    StaticDetectionPipelineModel model = this.getModel();
-    Map<MetricSlice, DataFrame> timeseries = this.provider.fetchTimeseries(model.timeseriesSlices);
-    Map<MetricSlice, DataFrame> aggregates = this.provider.fetchAggregates(model.aggregateSlices, Collections.<String>emptyList());
-    Multimap<AnomalySlice, MergedAnomalyResultDTO> anomalies = this.provider.fetchAnomalies(model.anomalySlices);
-    Multimap<EventSlice, EventDTO> events = this.provider.fetchEvents(model.eventSlices);
+    InputDataSpec dataSpec = this.getInputDataSpec();
+    Map<MetricSlice, DataFrame> timeseries = this.provider.fetchTimeseries(dataSpec.timeseriesSlices);
+    Map<MetricSlice, DataFrame> aggregates = this.provider.fetchAggregates(dataSpec.aggregateSlices, Collections.<String>emptyList());
```
kishoreg: why are we passing emptyList? is that the default value.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/algorithm/stage/AnomalyDetectionStageWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+import com.linkedin.thirdeye.detection.DetectionPipelineResult;
+import java.util.List;
+import java.util.Map;
+import org.apache.commons.collections.MapUtils;
+
+
+/**
+ * Anomaly Detection Stage Wrapper. This wrapper runs a anomaly detection stage and return the anomalies.
+ */
+public class AnomalyDetectionStageWrapper extends DetectionPipeline {
```
kishoreg: why are these called Wrapper's?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/algorithm/stage/AnomalyDetectionStageWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+  private static final String PROP_SPECS = "specs";
+
+  private final AnomalyDetectionStage anomalyDetectionStage;
+
+  public AnomalyDetectionStageWrapper(DataProvider provider, DetectionConfigDTO config, long startTime, long endTime)
+      throws Exception {
+    super(provider, config, startTime, endTime);
+
+    Map<String, Object> properties = config.getProperties();
+    Preconditions.checkArgument(properties.containsKey(PROP_STAGE_CLASSNAME), "Missing " + PROP_STAGE_CLASSNAME);
```
kishoreg: why cant we get the className from the detectionConfigDTO? converting a concrete object to properties to access something does not seem right. this can be done as part of another PR, b

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/InputDataSpec.java
jihaozh: sure. I'll add the java doc.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/StaticDetectionPipeline.java
jihaozh: will rename this to inputData.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/StaticDetectionPipeline.java
jihaozh: The empty list is the group-by dimensions parameter which meant to be used only for dimension exploration for now. So this is not exposed in the high-level interface.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/algorithm/stage/AnomalyDetectionStageWrapper.java
jihaozh: These are wrappers around the interfaces and work as the drivers for our interfaces.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/algorithm/stage/AnomalyDetectionStageWrapper.java
jihaozh: Yeah, but in the current design, the class name has to come from the properties because there are multiple nested wrapper class names we need to pull out from one detectionConfigDTO.

PR #3192, opened 2018-09-18
CSV Reader Config when Creating Segments from CSV using Hadoop

fredxfred:
Currently you can specify a readerconfig when creating segments using pinot-admin.sh, but this argument is not available when creating segments using the hadoop jar command. I added code so that you can specify a readerconfig filepath in job.properties. Not sure why the spacing ended up off within getReaderConfig.

It may be a good idea to also allow job.properties to specify the expected file format. The current way that Hadoop SegmentCreation reads file formats gives problems e.g. when uploading a .tsv file; this PR now allows reading files with any delimiter, but they still have to have a .csv file extension.

### DIFF
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
@@ -39,7 +39,7 @@ import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat;
 import org.json.JSONException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import org.codehaus.jackson.map.ObjectMapper;
 
 public class HadoopSegmentCreationMapReduceJob {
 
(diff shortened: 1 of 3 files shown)


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -273,12 +274,18 @@ protected String createSegment(String dataFilePath, Schema schema, Integer seqId
...
       RecordReaderConfig readerConfig = null;
       switch (fileFormat) {
         case CSV:
-          readerConfig = new CSVRecordReaderConfig();
-          break;
+          if(_readerConfigFile == null) {
+            readerConfig = new CSVRecordReaderConfig();
+	  }
+	  else {
+	    LOGGER.info("Reading CSV Record Reader Config from: {}", _readerConfigFile);
```
kishoreg: is the readerconfig logged somewhere else? if not, I would log the deserialized readerConfig and readerConfigFile together.

fredxfred: No it is not logged anywhere. I can log it in getReaderConfig

fredxfred:
Sure thing, I'll add a test case. Might take me a bit of time to figure it out

fredxfred:
Logged deserialized reader config. Not sure where/how to test

PR #3195, opened 2018-09-19
Handle the case where num_scanned metrics might be missing when encountering exceptions

Jackie-Jiang:
When the num_scanned metrics are missing (exception case), log -1 instead of 0 so that we know the query encountered exception
Emit values only if they are larger than 0

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QueryScheduler.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QueryScheduler.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
  */
 public abstract class QueryScheduler {
   private static final Logger LOGGER = LoggerFactory.getLogger(QueryScheduler.class);
+  private static final String INVALID_NUM_SCANNED = "-1";
 
   protected final ServerMetrics serverMetrics;
   protected final QueryExecutor queryExecutor;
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/scheduler/QueryScheduler.java
```diff
@@ -135,39 +136,38 @@ public void stop() {
...
-    long numEntriesScannedPostFilter = 0;
     String tableNameWithType = queryRequest.getTableNameWithType();
-    try {
-      numDocsScanned = Long.parseLong(getMetadataValue(dataTableMetadata, DataTable.NUM_DOCS_SCANNED_METADATA_KEY));
-      numEntriesScannedInFilter = Long.parseLong(getMetadataValue(dataTableMetadata, DataTable.NUM_ENTRIES_SCANNED_IN_FILTER_METADATA_KEY));
-      numEntriesScannedPostFilter = Long.parseLong(getMetadataValue(dataTableMetadata, DataTable.NUM_ENTRIES_SCANNED_POST_FILTER_METADATA_KEY));
+    long numDocsScanned =
+        Long.parseLong(dataTableMetadata.getOrDefault(DataTable.NUM_DOCS_SCANNED_METADATA_KEY, INVALID_NUM_SCANNED));
+    long numEntriesScannedInFilter = Long.parseLong(
+        dataTableMetadata.getOrDefault(DataTable.NUM_ENTRIES_SCANNED_IN_FILTER_METADATA_KEY, INVALID_NUM_SCANNED));
```
sunithabeeram: We are no longer catching parse exception. This seems fine for the most part (ie, it shouldn't really happen), but if it does then we fail the query while trying to do something non-essential. Something to consider - do we need to be more defensive?

Jackie-Jiang: This should be okay, unless we explicitly put non-number into this field, which is unlikely to happen.

PR #3197, opened 2018-09-19
Forcing file permissions of staging dir to be 755 in hadoop

jenniferdai:
* Forcing the staging directory upon segment creation to have the permission that the user sets in fs.permissions.umask-mode. Defaults to null if not set.
* Thanks to Jonathan Hung for the investigation:
Due to some FileSystem caching that hadoop does, we have problems setting permissions on files. When FileSystem.get is called, it will search for a cached instance in its internal map, but it doesn’t use the conf when generating the key. So probably somewhere before this code is called there was already a FileSystem created (with default umask 027) which is being used when the staging dir is created.

### DIFF
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
@@ -62,6 +62,8 @@ public class SegmentCreationJob extends Configured {
   private final String _outputDir;
   private final String _tableName;
 
+  private final String _defaultPermissions;
+
   private String[] _hosts;
   private int _port;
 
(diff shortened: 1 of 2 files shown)


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -119,19 +120,22 @@ public void run() throws Exception {
...
 
-    if (fs.exists(new Path(_stagingDir))) {
+    if (fs.exists(stagingDir)) {
       LOGGER.warn("Found the temp folder, deleting it");
-      fs.delete(new Path(_stagingDir), true);
+      fs.delete(stagingDir, true);
     }
-    fs.mkdirs(new Path(_stagingDir));
+    fs.mkdirs(stagingDir);
+    FileSystem.get(getConf()).setPermission(stagingDir, new FsPermission("755"));
```
mayankshriv: We should take the FsPermission as an argument, which either comes from a config or defaults to permission of the parent of destination (where the temp is moved to).

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -119,19 +124,24 @@ public void run() throws Exception {
...
     FileSystem fs = FileSystem.get(getConf());
     Path inputPathPattern = new Path(_inputSegmentDir);
+    Path stagingDir = new Path(_stagingDir);
+    Path outputDir = new Path(_outputDir);
 
-    if (fs.exists(new Path(_stagingDir))) {
+    if (fs.exists(stagingDir)) {
       LOGGER.warn("Found the temp folder, deleting it");
-      fs.delete(new Path(_stagingDir), true);
+      fs.delete(stagingDir, true);
```
mcvsubbu: always useful to log the name of the folder that is being deleted

PR #3199, opened 2018-09-19
[TE] Check for at least one valid email in to-recipients

akshayrai:


thirdeye/thirdeye-frontend/app/pods/components/modals/manage-groups-modal/template.hbs
```diff
@@ -191,8 +191,11 @@
         value=toAddresses
         class="te-modal__input te-modal__input--fullwidth"
         disabled=cannotEdit
-        change=(action "onToAddresses")
+        key-up=(action "onToAddresses")
```
apucher: is there a specific reason to use **key-up**?

akshayrai: We are checking for at least one valid email address. As "change" gets triggered only when losing the focus from the recipient box, there is a chance that users will edit the recipient box and click on submit without even seeing the warning message.

Which means users can edit the recipient box and click on submit without the warning message being displayed to them.

apucher: :+1:

PR #3200, opened 2018-09-19
Add new configuration format command line tool

jfim:
Add a tool to create and update tables using the new table
configuration format.

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/ApplyTableConfigCommand.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/ApplyTableConfigCommand.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
  *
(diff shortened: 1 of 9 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/ApplyTableConfigCommand.java
```diff
@@ -0,0 +1,143 @@
...
+
+  @Option(name = "-showComputedConfig", required = false, usage = "Only display the computed configuration, without attempting to upload it")
+  private boolean _showComputedConfig = false;
+
+  @Option(name = "-help", required = false, help = true, aliases = { "-h", "--h", "--help" },
+      usage = "Print this message.")
+  private boolean _help = false;
+
+  @Override
+  public boolean execute() throws Exception {
```
mayankshriv: Does it make sense to separate out the logic to apply table config from the CLI? Doing so allows the former to be reusable (have found that to be useful in the past).

jfim: After discussing this offline, we'll leave it this way since most of the logic (config profile handling, generation of the upload url) is already factored out. It's not exactly clear what would be the right API for people to reuse this, so we can extract a proper API at the time at which it would be needed.

PR #3202, opened 2018-09-20
Changing permissions parameter to subtract the umask

jenniferdai:


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -84,7 +84,7 @@ public SegmentCreationJob(String jobName, Properties properties) throws Exceptio
     String hostsString = _properties.getProperty(JobConfigConstants.PUSH_TO_HOSTS);
     String portString = _properties.getProperty(JobConfigConstants.PUSH_TO_PORT);
 
-    _defaultPermissions = _properties.getProperty(JobConfigConstants.DEFAULT_PERMISSIONS, null);
+    _defaultPermissionsMask = _properties.getProperty(JobConfigConstants.DEFAULT_PERMISSIONS_MASK, null);
```
mayankshriv: Lets log the _defaultPermissionsMask.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -134,12 +134,15 @@ public void run() throws Exception {
...
+      LOGGER.warn("Found the temp folder {}, deleting it", stagingDir);
       fs.delete(stagingDir, true);
     }
     fs.mkdirs(stagingDir);
-    if (_defaultPermissions != null) {
-      FileSystem.get(getConf()).setPermission(stagingDir, new FsPermission(_defaultPermissions));
+
+    if (_defaultPermissionsMask != null) {
+      FsPermission umask = new FsPermission(_defaultPermissionsMask);
+      FsPermission permission = FsPermission.getDefault().applyUMask(umask);
```
mayankshriv: What does FsPermission.getDefault() return? Is it guaranteed to be 777? If not, this will not work.

Also, lets log the permissions being set.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -134,12 +134,15 @@ public void run() throws Exception {
...
+      FileSystem.get(getConf()).setPermission(stagingDir, permission);
     }
     fs.mkdirs(new Path(_stagingDir + "/input/"));
```
mayankshriv: Lets log the full path being created.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
mayankshriv: Does the integration test exercise the flow where the mask value is set in _properties from outside?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -134,12 +134,15 @@ public void run() throws Exception {
...
-      LOGGER.warn("Found the temp folder, deleting it");
+      LOGGER.warn("Found the temp folder {}, deleting it", stagingDir);
       fs.delete(stagingDir, true);
     }
     fs.mkdirs(stagingDir);
-    if (_defaultPermissions != null) {
-      FileSystem.get(getConf()).setPermission(stagingDir, new FsPermission(_defaultPermissions));
+
+    if (_defaultPermissionsMask != null) {
+      FsPermission umask = new FsPermission(_defaultPermissionsMask);
```
mayankshriv: Creation of directory with specified can be extracted out as a method that can be unit tested that compares the umask value specified in the test with the permissions created for the directory.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
jenniferdai: I ran the integration test with/without it set.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/SegmentCreationJob.java
jenniferdai: I exercised this manually in the integration test. Unit tests for this portion would include a lot of mocking because I need to get a hadoop filesystem, so I think the first way is preferable.

mayankshriv: It can be unit tested for PinotLocalFs, which should not require mocking?

PR #3208, opened 2018-09-21
Simplify the parameter for forward index creators

Jackie-Jiang:
Also move method getNumBitsPerValue() into PinotDataBitSet and re-implement it without using double

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/util/PinotDataBitSet.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/util/PinotDataBitSet.java
@@ -39,6 +39,34 @@ public final class PinotDataBitSet implements Closeable {
     }
   }
 
+  /**
+   * Returns the number of bits required to encode the value.
+   * <p>NOTE: Use at least one bit even there is only one possible value
+   * <p>Examples: (maximum value (binary format) -> number of bits)
+   * <ul>
+   *   <li>0(0) -> 1</li>
+   *   <li>1(1) -> 1</li>
+   *   <li>2(10) -> 2</li>
+   *   <li>9(1001) -> 4</li>
+   *   <li>113(1110001) -> 7</li>
+   * </ul>
+   *
+   * @param maxValue Maximum possible value
+   * @return Number of bits required to encode the value
+   */
+  public static int getNumBitsPerValue(int maxValue) {
+    // Use at least one bit even there is only one possible value
+    if (maxValue <= 1) {
+      return 1;
+    }
+    int numBitsPerValue = Byte.SIZE;
+    while (maxValue > 0xFF) {
+      maxValue >>>= Byte.SIZE;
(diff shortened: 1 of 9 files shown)


pinot-core/src/test/java/com/linkedin/pinot/core/io/util/PinotDataBitSetTest.java
```diff
@@ -27,6 +27,25 @@
   private static final Random RANDOM = new Random();
   private static final int NUM_ITERATIONS = 1000;
 
+  @Test
+  public void testGetNumBitsPerValue() {
+    for (int i = 0; i < NUM_ITERATIONS; i++) {
+      int cardinality = RANDOM.nextInt(Integer.MAX_VALUE) + 1;
+      int actual = PinotDataBitSet.getNumBitsPerValue(cardinality);
+      if (cardinality == 1) {
```
snleee: The probability of random picking `cardinatlity = 1` is too low. Maybe it's better to test this separately outside of this loop since it's an edge case?

snleee: Since this is an edge case, can we check separately when cardinality = 1? Currently, it will hit 1 with very very low chance.

pinot-core/src/main/java/com/linkedin/pinot/core/io/util/PinotDataBitSet.java
```diff
@@ -39,6 +39,20 @@
     }
   }
 
+  public static int getNumBitsPerValue(int cardinality) {
```
snleee: Can you put some comments on how we decide the number of bits per value in high level with some simple examples?

pinot-core/src/test/java/com/linkedin/pinot/core/io/util/PinotDataBitSetTest.java
Jackie-Jiang: Good point. Added

pinot-core/src/main/java/com/linkedin/pinot/core/io/util/PinotDataBitSet.java
Jackie-Jiang: Added

pinot-core/src/main/java/com/linkedin/pinot/core/io/util/PinotDataBitSet.java
```diff
@@ -39,6 +39,20 @@
...
+    // Use at least one bit even there is only one possible value
```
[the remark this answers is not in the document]
Jackie-Jiang: There is no **Math.log2()**. In order to do so, we have to do **Math.ceil(Math.log(cardinality) / Math.log(2))**, while I am afraid that using double division could cause wrong result because of the precision error.
As for the **Math.max()**, we can avoid the actual calculation by using the if condition check.

PR #3209, opened 2018-09-21
Changing AzurePinotFS list files to be recursive

jenniferdai:


pinot-azure-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
```diff
@@ -116,13 +119,22 @@ public long length(URI fileUri) throws IOException {
 
   @Override
   public String[] listFiles(URI fileUri) throws IOException {
-    // list directory contents
-    List<DirectoryEntry> list = _adlStoreClient.enumerateDirectory(fileUri.getPath(), Integer.MAX_VALUE);
-    List<String> fileNames = new ArrayList<>();
-    for (DirectoryEntry directoryEntry : list) {
-      fileNames.add(directoryEntry.name);
+    // list directory contents recursively
+    List<String> finalFileList = new ArrayList<>();
```
mayankshriv: This seems quite inefficient. There's a new URI call and ArrayList <-> Array conversion at each level of recursion. These can be completely avoided by writing a helper method that works purely on DirectoryEntry (instead of URI's) and ArrayList (instead of converting from/to String[]).

pinot-azure-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
```diff
@@ -116,13 +117,24 @@ public long length(URI fileUri) throws IOException {
...
   @Override
   public String[] listFiles(URI fileUri) throws IOException {
-    // list directory contents
-    List<DirectoryEntry> list = _adlStoreClient.enumerateDirectory(fileUri.getPath(), Integer.MAX_VALUE);
-    List<String> fileNames = new ArrayList<>();
-    for (DirectoryEntry directoryEntry : list) {
-      fileNames.add(directoryEntry.name);
+    DirectoryEntry rootDir = _adlStoreClient.getDirectoryEntry(fileUri.getPath());
+    if (rootDir == null) {
+      return new String[0];
```
mayankshriv: Save the empty string in a static final, so we don't new a garbage with each call.

PR #3210, opened 2018-09-21
[PINOT-7088]Provide a way to no downtime table rebalancing

sunithabeeram:
Currently, when the controller's rebalance API is called, the idealstate is updated in one go and there are no guarantees around replica availability.

This change provides a way to do no downtime rebalancing. Key changes include:
1. Option to specify via rebalance API whether a downtime is acceptable or not. (By default the code does the safe thing: no-downtime rebalance).
2. Provide a helper class to handle the actual ideal-state update actions. This is moved out of the Strategy classes - perhaps a clearer approach as the Strategy classes just inform what needs to be done and the actual execution is handled outside of the Strategy.
3. IdealStateUpdater class: This class does the no-downtime update using a fixed thread pool - this allows for the API to return the target IdealState to the user and handle the execution in the background. The no downtime aspect is handled thus:
- The ideal state is updated to reflect the target state first by doing one change at a time. For ex, if current state is Segment 1 -> {host1:online, host2:online} and new state is Segment 1 -> {host3:online, host4:online}, we transition through the following changes:
Segment 1 -> {host1:online, host2:online, host3:online} => Segment 1 -> {host1:online, host2:online,host3:online, host4:online}
Transitions for all segments are triggered in each pass, but each segment will handle just one transition.
- After this the ideal state is set to the target which will result in removing segments from the current servers: Segment1 -> {host3:online, host4:online}

Testing done:
- Manual testing at this point to ensure:
1. API return immediately
2. Table with ReplicaGroupAssignmentStrategy can be rebalanced correctly when instances per partition are changed (ie, added or removed).

TODO: Update unit tests to cover the new logic. Subsequent commits will address this.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/IdealStateUpdater.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/IdealStateUpdater.java
@@ -1,0 +1,220 @@
+/**
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.controller.helix.core;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.linkedin.pinot.common.config.TableConfig;
+import com.linkedin.pinot.common.utils.helix.HelixHelper;
+import com.linkedin.pinot.common.utils.retry.RetryPolicies;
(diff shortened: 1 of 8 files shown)


sunithabeeram:
This PR is based on some assumptions (helix resource/replication behavior currently seen) that may/should not hold. Closing this and will create another PR with a different assignment approach.

mcvsubbu:
What happens if the controller that is doing the idealstate update in a no-downtime mode gets restarted?

sunithabeeram:
@mcvsubbu thats a good point. I hadn't explicitly considered that, but the other thing that I was thinking about and the restart point may have same solution: what happens if another attempt to rebalance is kicked off either on the same controller or a different one? If the final ideal state mapping is different between the attempts, are convergence and correctness guaranteed?

I guess this will involve some sort of state maintenance to ensure things don't step on each other/have silent failures. We could have the leader handle rebalances and maintain local state (but won't have restart guarantees) or we have zk state and either a leader or other nodes handle rebalancing with heartbeats.

Its possible that a first cut of this approach still doesn't handle these cases and we build this in later. Currently, the operational overhead of ensuring no downtime rebalances (specifically where host count within a replica group changes) is immense (often requiring another table to be setup), so a solution that can alleviate the "happy" path issues may still take us a long way.

kishoreg:
The rebalance strategy in Helix is idempotent. We dont have to worry about
the edge cases as long as we compare and set api when we change the ideal
state

mcvsubbu:
I thought helix takes care of assuring no downtime, so I am not sure why we had to do this
-Subbu
________________________________

PR #3214, opened 2018-09-21
Adding local temp dir parameter to distinguish between local and remote fs

jenniferdai:
* Adding a new local temp dir parameter that determines the final location of segment upload
* If local.temp.dir is not set, we will default to data.dir
* Adding support for copyFromLocal in Azure
* Changing fileUploadPath provider to allow different schemes for final storage directory and local temp dir
* Fixing server pinot fs factory config
* Changing variable names to clarify intent

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
@@ -18,6 +18,7 @@ package com.linkedin.pinot.controller.api.resources;
 import com.linkedin.pinot.common.Utils;
 import com.linkedin.pinot.common.utils.CommonConstants;
 import com.linkedin.pinot.controller.ControllerConf;
+import com.linkedin.pinot.filesystem.LocalPinotFS;
 import com.linkedin.pinot.filesystem.PinotFS;
 import com.linkedin.pinot.filesystem.PinotFSFactory;
 import java.io.File;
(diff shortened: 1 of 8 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
```diff
@@ -66,22 +67,32 @@ public FileUploadPathProvider(ControllerConf controllerConf) throws InvalidContr
...
     PinotFS pinotFS = PinotFSFactory.create(scheme);
 
     try {
+      _localTempDirURI = new URI(localTempDir);
+      if (!pinotFS.exists(_localTempDirURI)) {
+        pinotFS.mkdir(_localTempDirURI);
+      }
       _baseDataDirURI = new URI(dataDir);
       if (!pinotFS.exists(_baseDataDirURI)) {
         pinotFS.mkdir(_baseDataDirURI);
```
mayankshriv: As a side note, I see that mkdir() will forcefully remove the directory if it already exists. This seems quite dangerous to me. For example, if there's a bug in pinotFs.exists(), we will end up deleting the data dir. This might be acceptable behavior for a file, but dangerous for deleting the entire directory. Let's open an issue to visit this outside of this PR.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
```diff
@@ -66,22 +67,32 @@ public FileUploadPathProvider(ControllerConf controllerConf) throws InvalidContr
...
       }
-      _fileUploadTmpDirURI = new URI(_baseDataDirURI + FILE_UPLOAD_TEMP_PATH);
+      _fileUploadTmpDirURI = new URI(_localTempDirURI + FILE_UPLOAD_TEMP_PATH);
```
mayankshriv: Do we need a '/' between tempdir and upload path?

jenniferdai: my paths include the slash

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -291,26 +291,28 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
...
-              getSegmentMetadata(crypterClassHeader, storageURI, tempEncryptedFile, tempDecryptedFile, tempSegmentDir,
+              getSegmentMetadata(crypterClassHeader, currentSegmentLocationURI, tempEncryptedFile, tempDecryptedFile, tempSegmentDir,
                   metadataProviderClass);
-          if (new URI(storageURI).getScheme().equals(CommonConstants.Segment.LOCAL_SEGMENT_SCHEME)) {
-            downloadURI = ControllerConf.constructDownloadUrl(segmentMetadata.getTableName(), segmentMetadata.getName(), provider.getVip());
+          if (new URI(currentSegmentLocationURI).getScheme().equals(CommonConstants.Segment.LOCAL_SEGMENT_SCHEME)) {
+            zkDownloadUri = ControllerConf.constructDownloadUrl(segmentMetadata.getTableName(), segmentMetadata.getName(), provider.getVip());
           } else {
-            downloadURI = storageURI;
+            LOGGER.info("Using configured data dir {}", _controllerConf.getDataDir());
```
fx19880617: Why we need to do this? This is a backward-incompatible change and breaks the our production pipeline.
When we upload a segment with uri type, I need to save zkDownloadUri same as currentSegmentLocationURI.

mayankshriv: I think the intent here is for storageUri to be something purely under the control of controller, as the currentSegmentUri could be a path that controller might not have any control over (eg write permissions, segments of same table could in different locations, etc). Is this failing because datadir is not set? Or are you trying to avoid additional copy? One option could be an optimization in the code that infers that the current segment is already in the datadir and hence no move required.

fx19880617: Yes, the major problem for us is that we don't have NAS mounting to all the controllers/servers for sharing segments.  All the segments are saved remotely on HDFS.

That's why our upload segment request will only contain segment download uri. We also expect the server to directly download segments from HDFS directly. Also for now we don't expect controller to save those data to HDFS, major concern is our HDFS quota.

PR #3223, opened 2018-09-25
Improve the way we get instances with a tag from helix

npawar:
Get all instance configs in one shot, and then filter out the instance configs based on tags/enabled. This will avoid exceptions we encounter if an instances is in an inconsistent state (for example present in INSTANCES but not in PARTICIPANTS). This is also more efficient than fetching all instances from INSTANCES first and then fetching each instance config

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
@@ -410,4 +410,22 @@ public class HelixHelper {
     }
     return enabledInstances;
   }
+
+  public static List<String> getInstancesWithTag(HelixAdmin helixAdmin, String helixClusterName,
+      String instanceTag) {
+    List<String> instances = helixAdmin.getInstancesInCluster(helixClusterName);
+    List<String> instancesWithTag = new ArrayList<>();
+    for (String instance : instances) {
+      try {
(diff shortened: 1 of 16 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/helix/HelixHelper.java
```diff
@@ -410,4 +410,22 @@ public IdealState apply(IdealState idealState) {
     }
     return enabledInstances;
   }
+
+  public static List<String> getInstancesWithTag(HelixAdmin helixAdmin, String helixClusterName,
```
kishoreg: lets get all instanceConfigs in one shot. that should not throw any exceptions and is more efficient.

npawar: I don't think I understand. That's exactly what this method is doing

npawar: Understood what you're referring to from the conversation offline. Will change it accordingly

npawar: @kishoreg made the changes to get all instance configs in one shot

npawar:
Moved code from both methods to a common private method

PR #3226, opened 2018-09-26
Log exception on controller for realtime segment commit operations

npawar:


npawar:
Planning to follow this up with an exponential back off (with upper limit on backoff time). Is that what you meant?

PR #3229, opened 2018-09-26
Fix the bug for "select * ... order by" queries

snleee:
After we introduced virtual column, we did not filter out virtual columns
in SelectionOrderByOperator when computing selection columns. This led
ArrayIndexOutOfBoundsException for queries with select * ... order by.
This pr fixes the issue.


jfim:
I wonder how that passed the integration tests though. In theory, the query generator should have generated at least one query that has the `select * ... order by` shape, so this should have been caught earlier.

Can you look into how that happened? The code fix, if need be, can be done as part of a different PR.

snleee:
@jfim I spent some time on adding this to existing test code. However, it will require a lot of code change because we perform sorting on column names when deciding the order for projection. This causes the comparison to fail. (I checked that the returning results are the same rows but columns are in different order) I will address this as a separate pr.

PR #3236, opened 2018-09-28
[Star-Tree V2] Value Aggregator

Jackie-Jiang:
Added a new interface Value Aggregator to pre-aggregate on  the input values for a specific type of aggregation
Implemented value aggregators for COUNT, MIN, MAX, SUM, AVG, MIN_MAX_RANGE, DISTINCT_COUNT_HLL, PERCENTILE_EST, PERCENTILE_TDIGEST
The value aggregators can be used by star-tree to generate aggregated values

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/aggregator/AvgValueAggregator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/aggregator/AvgValueAggregator.java
@@ -1,0 +1,81 @@
+/**
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/startree/v2/function/StarTreeAggregationFunction.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.IOException;
+
+
+/**
+ * Aggregation function for star-tree to aggregate on raw values as well as aggregated values.
+ *
+ * @param <R> Type of the raw value (non-aggregated)
+ * @param <A> Type of the aggregated value
+ */
+public interface StarTreeAggregationFunction<R, A> {
```
mayankshriv: Using AggregationFunction suffix makes it a bit confusing with the aggregation functions used during query execution time. Since this is completely different from that, may be using a different suffix like 'Aggregator' might help avoid that confusion.

Jackie-Jiang: Will rename to **StarTreeValueAggregator**

PR #3240, opened 2018-09-28
[TE] dimensions algorithm table fixes

newsummit:
Several closely related fixes and UX modifications for the dimensions algorithm table:

- Selection of a dimension exploration response row will generate a graph in the RCA UI. Previously, this graph was ignoring filters and just displaying the unfiltered metric. Now the filters are respected.
- The configuration in the table settings modal are now resetting in the event of:
-- modal cancellation prior to hitting "save"
-- a change in primary metric from the page context
- Added 2 functions needed in order to "roll up" dimension sub-totals into the table if needed. Currently, this feature is turned off until we get further clarity on it.

screen shot 2018-09-28 at 4 10 10 pm

Tests passing

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-dimensions-algorithm/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-dimensions-algorithm/component.js
@@ -116,6 +116,7 @@ export default Component.extend({
     summarySize: 20,
     orderType: 'auto',
     oneSideError: 'false',
+    enableSubTotals: true,
     excludedDimensions: []
   },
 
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-dimensions-algorithm/component.js
```diff
@@ -330,17 +420,33 @@ export default Component.extend({
       if (customTableSettings.excludedDimensions.length) {
         Object.assign(requestObj, { excludedDimensions: customTableSettings.excludedDimensions.join(',') });
       }
+      // Fetch dimension algorithm data
```
aaronucsd: Some async/awaits on the fetches?

newsummit: good idea. Thanks.

PR #3249, opened 2018-10-02
[TE] RCA Trend table fix

newsummit:
Trend table was unable to match selected start time to existing array of start times in component due to unnecessary nesting of the computed property.


aaronucsd:
I'm not able to grok the fix. Which CP is causing the issue? All i see is replacements of `getProperties` with `get`.

PR #3252, opened 2018-10-02
Cleanup PinotFS

jenniferdai:
* Editing decode/encode methods to be only for local fs
* Fixing AzureFS copy method
* Updating javadoc for mkdir


pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
```diff
@@ -120,17 +121,20 @@ public long length(URI fileUri) throws IOException {
 
   @Override
   public void copyToLocalFile(URI srcUri, File dstFile) throws Exception {
-    copy(srcUri, new URI(URLEncoder.encode(dstFile.getAbsolutePath(), "UTF-8")));
+    copy(srcUri, new URI(encodeURI(dstFile.getAbsolutePath())));
```
mayankshriv: May be the logging should be outside and not within the library.

PR #3253, opened 2018-10-03
Add flags to ensure server loading all segments then start serving query traffic

fx19880617:
1. Adding flags to start serving queries after all segments are loaded.
2. Adding a health check endpoint for pinot server.

### DIFF
--- a/pinot-server/src/main/java/com/linkedin/pinot/server/api/resources/HealthCheckResource.java
+++ b/pinot-server/src/main/java/com/linkedin/pinot/server/api/resources/HealthCheckResource.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.server.api.resources;
+
+import com.linkedin.pinot.common.utils.ServiceStatus;
+import com.linkedin.pinot.common.utils.ServiceStatus.Status;
+import com.linkedin.pinot.server.starter.ServerInstance;
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+/**
+ * REST API to do health check through ServiceStatus.
+ */
+@Api(tags = "Health")
+@Path("/")
+public class HealthCheckResource {
+
+  @GET
+  @Path("/health")
+  @Produces(MediaType.TEXT_PLAIN)
+  @ApiOperation(value = "Checking server health")
+  @ApiResponses(value = { @ApiResponse(code = 200, message = "Server is healthy"), @ApiResponse(code = 503, message = "Server is not healthy")
+  })
+  public String checkHealth() {
+    Status status = ServiceStatus.getServiceStatus();
+    if (status == Status.GOOD) {
+      return "OK";
+    }
+    throw new WebApplicationException(String.format("Pinot server status is %s", status), Response.Status.SERVICE_UNAVAILABLE);
+  }
+}
--- a/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
+++ b/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
@@ -26,6 +26,8 @@ import com.linkedin.pinot.common.metrics.ServerMetrics;
 import com.linkedin.pinot.common.utils.CommonConstants;
 import com.linkedin.pinot.common.utils.NetUtil;
 import com.linkedin.pinot.common.utils.ServiceStatus;
+import com.linkedin.pinot.common.utils.ServiceStatus.Status;
+import com.linkedin.pinot.core.data.manager.InstanceDataManager;
 import com.linkedin.pinot.core.segment.memory.PinotDataBuffer;
 import com.linkedin.pinot.server.conf.ServerConf;
 import com.linkedin.pinot.server.realtime.ControllerLeaderLocator;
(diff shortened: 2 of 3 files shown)


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -164,9 +167,32 @@ public HelixServerStarter(String helixClusterName, String zkServer, Configuratio
     serverMetrics.addCallbackGauge("memory.allocationFailureCount", PinotDataBuffer::getAllocationFailureCount);
   }
 
-  private void updateInstanceConfigInHelix(int adminApiPort, boolean shuttingDown) {
+  private void waitForAllSegmentsLoaded() {
+    if (_helixServerConfig.getBoolean(CommonConstants.Server.CONFIG_OF_STARTER_ENABLE_SEGMENTS_LOADING_CHECK, CommonConstants.Server.DEFAULT_STARTER_ENABLE_SEGMENTS_LOADING_CHECK)) {
+      long startTime = System.currentTimeMillis();
+      int serverStarterTimeout = _helixServerConfig.getInt(CommonConstants.Server.CONFIG_OF_STARTER_TIMEOUT_IN_SECONDS, CommonConstants.Server.DEFAULT_STARTER_TIMEOUT_IN_SECONDS);
+      boolean allSegmentsLoaded = false;
+      while (System.currentTimeMillis() - startTime < 1000 * serverStarterTimeout) {
```
sunithabeeram: Use TimeUnit to convert serverStarterTimeout to ms once outside of the loop. (No need to 1000 * serverStarterTimeout repeatedly).

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -164,9 +167,32 @@ public HelixServerStarter(String helixClusterName, String zkServer, Configuratio
...
+        if (ServiceStatus.getServiceStatus() == Status.GOOD) {
+          LOGGER.info("All the segments are fully loaded into Pinot server.");
+          allSegmentsLoaded = true;
+          break;
+        }
+        try {
+          LOGGER.info("Waiting for all segments loaded, sleep 10 seconds...");
+          Thread.sleep(10000);
```
sunithabeeram: For the currently configured value of 600s, this sleep seems ok, but it may not be if its configured lower. Would be good to:
a) Use a named constant 
b) sleep for min of that constant or remaining time

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -164,9 +167,32 @@ public HelixServerStarter(String helixClusterName, String zkServer, Configuratio
...
+        } catch (Exception e) {
+          // Do nothing
```
sunithabeeram: Please catch interrupted exception separately. 
When you catch interrupted exception, do:
{
Thread.currentThread.interrupt();
return;
}
This allows the interrupt to propagate - the method should not continue on ignoring the interrupt.
For all other exceptions please log the exception.

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -164,9 +167,32 @@ public HelixServerStarter(String helixClusterName, String zkServer, Configuratio
...
+        }
+      }
+      if (!allSegmentsLoaded) {
+        LOGGER.info("Segments are not fully loaded within {} seconds...", serverStarterTimeout);
```
sunithabeeram: Log this as a warning.

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
jackjlli: It'd be good to firstly calculate the estimated end time beforehand, and then check `System.currentTimeMillis() < estimatedEndTime` in the while loop.

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -164,9 +167,32 @@ public HelixServerStarter(String helixClusterName, String zkServer, Configuratio
...
+      }
+    }
+  }
+
+  private void setAdminApiPort(int adminApiPort) {
     Map<String, String> propToUpdate = new HashMap<>();
-    propToUpdate.put(CommonConstants.Helix.IS_SHUTDOWN_IN_PROGRESS, String.valueOf(shuttingDown));
```
jackjlli: Shouldn't remove this flag. Otherwise Pinot broker won't be notified that Server has already been back up.

pinot-server/src/main/java/com/linkedin/pinot/server/api/resources/HealthCheckResource.java
```diff
@@ -0,0 +1,59 @@
...
+  @Inject
+  ServerInstance server;
+
+  @GET
+  @Path("/health")
+  @Produces(MediaType.TEXT_PLAIN)
+  @ApiOperation(value = "Checking server health")
+  @ApiResponses(value = { @ApiResponse(code = 200, message = "Server is healthy"), @ApiResponse(code = 503, message = "Server is not healthy")
+  })
+  public String checkHealth() {
```
kishoreg: can you add another api that spits out the tables/segments that are not healthy.

Something like
T1:
 status:OK
 ONLINE:X, OFFLINE:Y, ERROR:Z....
T2:NOT OK
 ONLINE:X, OFFLINE:Y, ERROR:Z....


...

unhealthy segments
T1:<number

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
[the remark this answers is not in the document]
fx19880617: No, ServiceStatus.getServiceStatus() doesn't provide such functionalities.

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
fx19880617: Will calculate the time to sleep.

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
fx19880617: will do

fx19880617: will do

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
fx19880617: will do

pinot-server/src/main/java/com/linkedin/pinot/server/api/resources/HealthCheckResource.java
fx19880617: How about adding this feature in a separated PR, since health check is a blocker for server starter. This is a feature for easy troubleshooting.

fx19880617:
@kishoreg could you take another pass?

pinot-server/src/main/java/com/linkedin/pinot/server/api/resources/HealthCheckResource.java
kishoreg: Makes sense!.

pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -177,9 +186,97 @@ public HelixServerStarter(String helixClusterName, String zkServer, Configuratio
...
+        if (ServiceStatus.getServiceStatus() == Status.GOOD) {
+          LOGGER.info("All the segments are fully loaded into Pinot server, time taken: {} seconds", TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis() - startTime));
+          allSegmentsLoaded = true;
+          break;
+        }
+        try {
+          int numSegmentsLoaded = getNumSegmentLoaded();
+          int numSegmentsToLoad = getNumSegmentsToLoad();
+          LOGGER.warn("Waiting for all segments to be loaded, current progress: [ {} / {} ], sleep {} seconds...", numSegmentsLoaded, numSegmentsToLoad, timeToSleep);
+          logSegmentsLoadingInfo();
```
sunithabeeram: It might be good to log this only at the end when we are bailing out - else it might result in too much logging.

fx19880617: Only logging per table level progress, when full segments bootstrap is not completed on time.

sunithabeeram: Thanks for fixing it!

PR #3254, opened 2018-10-03
[Star-Tree V2] Enhance AVG, MIN_MAX_RANGE aggregatin function to work on serialized bytes

Jackie-Jiang:
Enhance AVG, MIN_MAX_RANGE aggregation function to work on serialized bytes
Star-Tree can generate pre-aggregated value for these aggregations

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/AvgAggregationFunction.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/AvgAggregationFunction.java
@@ -15,8 +15,10 @@
  */
 package com.linkedin.pinot.core.query.aggregation.function;
 
+import com.linkedin.pinot.common.data.FieldSpec;
 import com.linkedin.pinot.common.utils.DataSchema;
 import com.linkedin.pinot.core.common.BlockValSet;
+import com.linkedin.pinot.core.common.ObjectSerDeUtils;
 import com.linkedin.pinot.core.query.aggregation.AggregationResultHolder;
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/aggregation/function/AvgAggregationFunction.java
```diff
@@ -60,12 +62,34 @@ public GroupByResultHolder createGroupByResultHolder(int initialCapacity, int ma
...
+        }
+        setAggregationResult(aggregationResultHolder, sum, (long) length);
+        break;
+      case BYTES:
+        // Serialized AvgPair
+        byte[][] bytesValues = blockValSets[0].getBytesValuesSV();
+        sum = 0.0;
+        long count = 0L;
+        for (int i = 0; i < length; i++) {
+          AvgPair value = ObjectSerDeUtils.AVG_PAIR_SER_DE.deserialize(bytesValues[i]);
```
mayankshriv: How do we guarantee that the bytes are deserializable to the expected class?

Jackie-Jiang: The bytes should be generated from star-tree (or the aggregator for this aggregation function). If client aggregates on a random bytes column, it will throw exception.

mayankshriv: For the latter, we probably should think about having a query sanitizer to block bad queries reaching this deep in the execution stack. But that I guess is outside the scope of this PR.

PR #3255, opened 2018-10-03
Have server wait until no incoming queries nor online table resources during shutdown

jackjlli:
This PR adds code to have pinot server wait until it doesn't get any incoming queries within 60 seconds during shutdown. The max waiting time is 10 mins. Add `_latestQueryTime` in ServerInstance to track the latest Query time.
This PR also adds code to wait for the EV update as well. A spectator helix manager is needed to keep track of the EV change.
Unit test added.


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -220,10 +223,38 @@ public void stop() {
     if (_helixServerConfig.getBoolean(CommonConstants.Server.CONFIG_OF_ENABLE_SHUTDOWN_DELAY, true)) {
       Uninterruptibles.sleepUninterruptibly(_maxQueryTimeMs, TimeUnit.MILLISECONDS);
     }
+    waitUntilNoIncomingQueries();
```
sunithabeeram: From our observations and evidence from logs, broker sees instance config updates soon enough and the server will stop getting requests soon after shuttingdown is set to true. I am not sure this change will actually get what we want.

PR #3260, opened 2018-10-04
Adding an exception to server response when there are missing segments

snleee:
When broker sends a request to server for a segment that does not
exist on server, no message or indication on this is sent back
to broker. This pr adds the exception information in DataTable that
is sent back to broker that it can know if there was any missing
segment from server side while processing the query.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
@@ -38,11 +38,14 @@ import com.linkedin.pinot.core.query.pruner.SegmentPrunerService;
 import com.linkedin.pinot.core.query.request.ServerQueryRequest;
 import com.linkedin.pinot.core.query.request.context.TimerContext;
 import com.linkedin.pinot.core.util.trace.TraceContext;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import javax.annotation.Nonnull;
 import javax.annotation.concurrent.ThreadSafe;
 import org.apache.commons.configuration.Configuration;
 import org.apache.commons.configuration.ConfigurationException;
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -221,6 +233,46 @@ private long pruneSegments(TableDataManager tableDataManager, List<SegmentDataMa
...
   }
 
+  /**
+   * Helper method to compute missing segments
+   *
+   * @param segmentsToQuery segments that need to be queried according to server query request
+   * @param segmentDataManagers acquired segment data managers from table data manager
+   * @return a list of missing segments
+   */
+  @Nonnull
```
Jackie-Jiang: I think the new convention is to not use Nonnull annotation

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -221,6 +233,46 @@ private long pruneSegments(TableDataManager tableDataManager, List<SegmentDataMa
...
+  private Set<String> computeMissingSegments(List<String> segmentsToQuery,
+      List<SegmentDataManager> segmentDataManagers) {
+    Set<String> missingSegments = new HashSet<>(segmentsToQuery);
+    for (SegmentDataManager segmentDataManager : segmentDataManagers) {
+      String segmentName = segmentDataManager.getSegmentName();
+      if (missingSegments.contains(segmentName)) {
```
Jackie-Jiang: Can be simplified by directly calling missingSegments.remove(segmentName)

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
snleee: removed the annotation

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
snleee: fixed

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -221,6 +233,46 @@ private long pruneSegments(TableDataManager tableDataManager, List<SegmentDataMa
...
+
+  /**
+   * Helper method to handle missing segments
+   *
+   * @param missingSegments a list of missing segments
+   * @param tableNameWithType table name with type
+   * @param dataTable data table result
+   */
+  private void handleMissingSegments(Set<String> missingSegments, String tableNameWithType, DataTable dataTable) {
+    String errorMsg = "Could not find segment " + String.join(",", missingSegments) + " for table " + tableNameWithType;
```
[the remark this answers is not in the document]
kishoreg: This will allow us to add a feature in broker to only query the missing segments and merge with current response.

sunithabeeram:
Will the handling of broker setting partial-response based on the missing segments indication be done in a different PR?

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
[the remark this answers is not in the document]
Jackie-Jiang: One way is using the Exception, but just put a ',' joined segments as the exception message

PR #3264, opened 2018-10-04
Add abstract class PeriodicTask to unify duplicate code

jackjlli:
This RB:

- adds abstract class PeriodicTask to unify some duplicate code.
- adds another class to specify Controller's PeriodicTask, which can only be run by Pinot controller only.
- adds periodic task scheduler, so that all the periodic tasks share the same executor with different task interval times.
- refactors logic so that some cleanup code would be called only once, i.e. leadership transition.
- adds unit tests.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
@@ -15,74 +15,34 @@
  */
 package com.linkedin.pinot.common.utils;
 
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 
 public abstract class PeriodicTask {
-  private static final Logger LOGGER = LoggerFactory.getLogger(PeriodicTask.class);
-  protected ScheduledExecutorService _executorService;
-  private String _taskName;
+  private final String _taskName;
+  private long _executionTime;
   private long _intervalSeconds;
   private long _initialDelay;
 
+  public static final long DEFAULT_RUN_FREQUENCY_IN_SECOND = 3600L;
+  public static final long DEFAULT_INITIAL_DELAY = 120L;
+
   public PeriodicTask(String taskName) {
-    this(taskName, 3600L);
+    this(taskName, DEFAULT_RUN_FREQUENCY_IN_SECOND);
   }
 
   public PeriodicTask(String taskName, long runFrequencyInSeconds) {
-    this(taskName, runFrequencyInSeconds, 120L);
+    this(taskName, runFrequencyInSeconds, DEFAULT_INITIAL_DELAY);
   }
 
   public PeriodicTask(String taskName, long runFrequencyInSeconds, long initialDelay) {
     _taskName = taskName;
     _intervalSeconds = runFrequencyInSeconds;
     _initialDelay = initialDelay;
-    _executorService = Executors.newSingleThreadScheduledExecutor(runnable -> {
-      Thread thread = new Thread(runnable);
-      thread.setName(_taskName + "ExecutorService");
-      return thread;
-    });
+    _executionTime = System.currentTimeMillis() + initialDelay;
   }
 
   public abstract void runTask();
 
-  /**
-   * Start the periodic task.
-   */
-  public void start() {
-    LOGGER.info("Starting {}", _taskName);
-
-    // Set up an executor that executes validation tasks periodically
-    _executorService.scheduleWithFixedDelay(() -> {
-      try {
-        runTask();
-      } catch (Throwable e) {
-        // catch all errors to prevent subsequent exeuctions from being silently suppressed
-        // Ref: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay-java.lang.Runnable-long-long-java.util.concurrent.TimeUnit-
-        LOGGER.warn("Caught exception while running {}", _taskName, e);
-      }
-    }, _initialDelay, _intervalSeconds, TimeUnit.SECONDS);
-  }
-
-  /**
-   * Stop the periodic task.
-   */
-  public void stop() {
-    LOGGER.info("Stopping {}", _taskName);
-    if (_executorService == null) {
-      return;
-    }
-    _executorService.shutdown();
-    try {
-      _executorService.awaitTermination(_initialDelay, TimeUnit.SECONDS);
-    } catch (InterruptedException e) {
-      // Ignored
-    }
-    _executorService = null;
+  public void initTask() {
   }
 
   protected void setIntervalSeconds(long intervalSeconds) {
@@ -91,5 +51,21 @@ public abstract class PeriodicTask {
 
   protected long getIntervalSeconds() {
     return _intervalSeconds;
+  }
+
+  protected long getInitialDelay() {
+    return _initialDelay;
+  }
+
+  public String getTaskName() {
+    return _taskName;
+  }
+
+  public long getExecutionTime() {
+    return _executionTime;
+  }
+
+  public void updateExecutionTime() {
+    _executionTime += _intervalSeconds;
   }
 }
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
@@ -1,0 +1,95 @@
+/**
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.utils;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public abstract class PeriodicTask {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PeriodicTask.class);
+  protected ScheduledExecutorService _executorService;
+  private String _taskName;
+  private long _intervalSeconds;
+  private long _initialDelay;
+
+  public PeriodicTask(String taskName) {
+    this(taskName, 3600L);
+  }
+
+  public PeriodicTask(String taskName, long runFrequencyInSeconds) {
+    this(taskName, runFrequencyInSeconds, 120L);
+  }
+
+  public PeriodicTask(String taskName, long runFrequencyInSeconds, long initialDelay) {
+    _taskName = taskName;
+    _intervalSeconds = runFrequencyInSeconds;
+    _initialDelay = initialDelay;
+    _executorService = Executors.newSingleThreadScheduledExecutor(runnable -> {
+      Thread thread = new Thread(runnable);
+      thread.setName(_taskName + "ExecutorService");
+      return thread;
+    });
+  }
+
+  public abstract void runTask();
+
+  /**
+   * Start the periodic task.
+   */
+  public void start() {
+    LOGGER.info("Starting {}", _taskName);
+
+    // Set up an executor that executes validation tasks periodically
+    _executorService.scheduleWithFixedDelay(() -> {
+      try {
+        runTask();
+      } catch (Throwable e) {
+        // catch all errors to prevent subsequent exeuctions from being silently suppressed
+        // Ref: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay-java.lang.Runnable-long-long-java.util.concurrent.TimeUnit-
+        LOGGER.warn("Caught exception while running {}", _taskName, e);
+      }
+    }, _initialDelay, _intervalSeconds, TimeUnit.SECONDS);
+  }
+
+  /**
+   * Stop the periodic task.
+   */
+  public void stop() {
+    LOGGER.info("Stopping {}", _taskName);
+    if (_executorService == null) {
+      return;
+    }
+    _executorService.shutdown();
+    try {
+      _executorService.awaitTermination(_initialDelay, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      // Ignored
+    }
+    _executorService = null;
+  }
+
+  protected void setIntervalSeconds(long intervalSeconds) {
+    _intervalSeconds = intervalSeconds;
+  }
+
+  protected long getIntervalSeconds() {
+    return _intervalSeconds;
+  }
+}
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
@@ -20,24 +20,24 @@ public abstract class PeriodicTask {
   private final String _taskName;
   private long _executionTime;
   private long _intervalSeconds;
-  private long _initialDelay;
+  private long _initialDelaySeconds;
 
   public static final long DEFAULT_RUN_FREQUENCY_IN_SECOND = 3600L;
-  public static final long DEFAULT_INITIAL_DELAY = 120L;
+  public static final long DEFAULT_INITIAL_DELAY_IN_SECOND = 120L;
 
   public PeriodicTask(String taskName) {
     this(taskName, DEFAULT_RUN_FREQUENCY_IN_SECOND);
   }
 
   public PeriodicTask(String taskName, long runFrequencyInSeconds) {
-    this(taskName, runFrequencyInSeconds, DEFAULT_INITIAL_DELAY);
+    this(taskName, runFrequencyInSeconds, DEFAULT_INITIAL_DELAY_IN_SECOND);
   }
 
-  public PeriodicTask(String taskName, long runFrequencyInSeconds, long initialDelay) {
+  public PeriodicTask(String taskName, long runFrequencyInSeconds, long initialDelaySeconds) {
     _taskName = taskName;
     _intervalSeconds = runFrequencyInSeconds;
-    _initialDelay = initialDelay;
-    _executionTime = System.currentTimeMillis() + initialDelay;
+    _initialDelaySeconds = initialDelaySeconds;
+    _executionTime = System.currentTimeMillis() + initialDelaySeconds;
   }
 
   public abstract void runTask();
@@ -53,8 +53,8 @@ public abstract class PeriodicTask {
     return _intervalSeconds;
   }
 
-  protected long getInitialDelay() {
-    return _initialDelay;
+  protected long getInitialDelaySeconds() {
+    return _initialDelaySeconds;
   }
 
   public String getTaskName() {
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTaskScheduler.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTaskScheduler.java
@@ -1,0 +1,103 @@
+/**
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.common.utils;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.PriorityBlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class PeriodicTaskScheduler {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PeriodicTaskScheduler.class);
+  private PriorityBlockingQueue<PeriodicTask> _periodicTasks;
+  protected ScheduledExecutorService _executorService;
+  private long _intervalSeconds;
+  private long _initialDelay;
+
+  public PeriodicTaskScheduler() {
+    _executorService = Executors.newSingleThreadScheduledExecutor(runnable -> {
+      Thread thread = new Thread(runnable);
+      thread.setName("PeriodicTaskSchedulerExecutorService");
+      return thread;
+    });
+    _initialDelay = PeriodicTask.DEFAULT_INITIAL_DELAY;
+    _intervalSeconds = PeriodicTask.DEFAULT_RUN_FREQUENCY_IN_SECOND;
+    _periodicTasks = new PriorityBlockingQueue<>(10, (o1, o2) -> {
+      if (o1.getExecutionTime() == o2.getExecutionTime()) {
+        return 0;
+      }
+      return o1.getExecutionTime() < o2.getExecutionTime() ? -1 : 1;
+    });
+  }
+
+  public void start() {
+    LOGGER.info("Starting PeriodicTaskScheduler. Run frequency in seconds: {}", _intervalSeconds);
+
+    // Set up an executor that executes tasks periodically
+    _executorService.scheduleWithFixedDelay(() -> {
+      if (_periodicTasks.isEmpty()) {
+        LOGGER.warn("No task assigned to PeriodicTaskScheduler");
+        return;
+      }
+      PeriodicTask task = _periodicTasks.poll();
+      try {
+        task.runTask();
+      } catch (Throwable e) {
+        // catch all errors to prevent subsequent exeuctions from being silently suppressed
+        // Ref: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay-java.lang.Runnable-long-long-java.util.concurrent.TimeUnit-
+        LOGGER.warn("Caught exception while running PeriodicTaskScheduler", e);
+      } finally {
+        task.updateExecutionTime();
+        _periodicTasks.offer(task);
+      }
+    }, _initialDelay, _intervalSeconds, TimeUnit.SECONDS);
+  }
+
+  public void stop() {
+    LOGGER.info("Stopping PeriodicTaskScheduler");
+    if (_executorService == null) {
+      return;
+    }
+    _executorService.shutdown();
+    try {
+      _executorService.awaitTermination(_initialDelay, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      // Ignored
+    }
+    _executorService = null;
+  }
+
+  /**
+   * Add periodic task before starting the scheduler.
+   */
+  public void addPeriodicTask(PeriodicTask task) {
+    if (task.getIntervalSeconds() > 0L) {
+      LOGGER.info("Adding {} to PeriodicTaskScheduler, run frequency in seconds: {}", task.getTaskName(),
+          task.getIntervalSeconds());
+      task.initTask();
+      _periodicTasks.offer(task);
+      _intervalSeconds = Math.min(_intervalSeconds, task.getIntervalSeconds());
+      _initialDelay = Math.min(_initialDelay, task.getInitialDelay());
+    }
+  }
+
+  public void removePeriodicTask(PeriodicTask task) {
+    _periodicTasks.remove(task);
+  }
+}
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTaskScheduler.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTaskScheduler.java
@@ -28,7 +28,7 @@ public class PeriodicTaskScheduler {
   private PriorityBlockingQueue<PeriodicTask> _periodicTasks;
   protected ScheduledExecutorService _executorService;
   private long _intervalSeconds;
-  private long _initialDelay;
+  private long _initialDelaySeconds;
 
   public PeriodicTaskScheduler() {
     _executorService = Executors.newSingleThreadScheduledExecutor(runnable -> {
@@ -36,7 +36,7 @@ public class PeriodicTaskScheduler {
       thread.setName("PeriodicTaskSchedulerExecutorService");
       return thread;
     });
-    _initialDelay = PeriodicTask.DEFAULT_INITIAL_DELAY;
+    _initialDelaySeconds = PeriodicTask.DEFAULT_INITIAL_DELAY_IN_SECOND;
     _intervalSeconds = PeriodicTask.DEFAULT_RUN_FREQUENCY_IN_SECOND;
     _periodicTasks = new PriorityBlockingQueue<>(10, (o1, o2) -> {
       if (o1.getExecutionTime() == o2.getExecutionTime()) {
@@ -50,12 +50,16 @@ public class PeriodicTaskScheduler {
     LOGGER.info("Starting PeriodicTaskScheduler. Run frequency in seconds: {}", _intervalSeconds);
 
     // Set up an executor that executes tasks periodically
-    _executorService.scheduleWithFixedDelay(() -> {
+    _executorService.scheduleAtFixedRate(() -> {
       if (_periodicTasks.isEmpty()) {
-        LOGGER.warn("No task assigned to PeriodicTaskScheduler");
+        LOGGER.info("No task assigned to PeriodicTaskScheduler");
         return;
       }
-      PeriodicTask task = _periodicTasks.poll();
+      PeriodicTask task = _periodicTasks.peek();
+      if (System.currentTimeMillis() < task.getExecutionTime()) {
+        return;
+      }
+      _periodicTasks.poll();
       try {
         task.runTask();
       } catch (Throwable e) {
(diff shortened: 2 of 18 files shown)


Jackie-Jiang:
We should have a single top level periodic task scheduler, then plug in multiple periodic tasks. This way, we can easily manage the order of all tasks, and avoid any race condition between tasks (e.g. retention manager and validation manager battling with each other). Also, we can fetch tables only once in the top level task scheduler and pass them to each individual tasks (when we implement the controller resource).

pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
```diff
@@ -15,74 +15,34 @@
...
-    try {
-      _executorService.awaitTermination(_initialDelay, TimeUnit.SECONDS);
-    } catch (InterruptedException e) {
-      // Ignored
-    }
-    _executorService = null;
+  public void initTask() {
   }
 
   protected void setIntervalSeconds(long intervalSeconds) {
```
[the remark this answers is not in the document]
jackjlli: That's because in `RealtimeSegmentRelocator`, we need to convert the frequency on the fly, e.g. converting `1h`(string format) to `3600`(long format). 
This conversion cannot be done until constructor gets called. And this setting intervalSeconds is what we're doing in the current code for `RealtimeSegmentRelocator`.

mcvsubbu:
> We should have a single top level periodic task scheduler, then plug in multiple periodic tasks. This way, we can easily manage the order of all tasks, and avoid any race condition between tasks (e.g. retention manager and validation manager battling with each other). Also, we can fetch tables only once in the top level task scheduler and pass them to each individual tasks (when we implement the controller resource).

I am not sure I agree that we should isolate the tasks. We don't do that now. The tasks need to be aware that other tasks may run, and should protect themselves appropriately.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
```diff
@@ -92,4 +52,20 @@ protected void setIntervalSeconds(long intervalSeconds) {
...
+
+  public String getTaskName() {
+    return _taskName;
+  }
+
+  public long getExecutionTime() {
+    return _executionTime;
+  }
+
+  public void updateExecutionTime() {
```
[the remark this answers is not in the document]
jackjlli: The variable `_executionTime` is used to calculate the priority of each task in priority queue. 
Right now we use `scheduleWithFixedDelay` for each of the tasks.
If we schedule a task to run every minute, but it takes 1.5 mins to run, then the next task would be ran after 1min, i.e. the 2nd task will start at 2.5th mins.
If we want the next task to run right after the previous task(which took 1.5mins) finished, then `scheduleAtFixedRate` should be used.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,103 @@
...
+
+public class PeriodicTaskScheduler {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PeriodicTaskScheduler.class);
+  private PriorityBlockingQueue<PeriodicTask> _periodicTasks;
+  protected ScheduledExecutorService _executorService;
+  private long _intervalSeconds;
+  private long _initialDelay;
+
+  public PeriodicTaskScheduler() {
+    _executorService = Executors.newSingleThreadScheduledExecutor(runnable -> {
```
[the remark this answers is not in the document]
jackjlli: I think it's preferable to use just 1 thread, so that multiple tasks won't be run at the same time (which may lead to race condition).

pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,103 @@
...
+      Thread thread = new Thread(runnable);
+      thread.setName("PeriodicTaskSchedulerExecutorService");
```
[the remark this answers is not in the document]
jackjlli: This is a scheduler level thread. If the thread was in task level, we could rename that.
Plus, the log message prints out the current class name. If some exception happens, we can find out which class this exception comes from according to the log message.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/PeriodicTask.java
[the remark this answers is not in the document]
jackjlli: Actually I like the idea of tracking the run time of each task as you mentioned in the first comment. I've added an abstract method in `PeriodicTask` called `runTask()`. We can easily calculate the run time of each task in that method and emit a metric out of it.

jackjlli:
Right now validation manager and task manager run every 1 hour by default. In order to stagger the tasks, validation manager is set to start at 30th min. Task manager remains the same.

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,109 @@
...
+  private long _intervalSeconds;
+  private long _initialDelaySeconds;
+
+  public PeriodicTaskScheduler() {
+    _executorService = Executors.newSingleThreadScheduledExecutor(runnable -> {
+      Thread thread = new Thread(runnable);
+      thread.setName("PeriodicTaskSchedulerExecutorService");
+      return thread;
+    });
+    _initialDelaySeconds = 120L;
```
Jackie-Jiang: Make this configurable

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,109 @@
...
+    _intervalSeconds = 300L;
```
Jackie-Jiang: Seems this one is not used

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,109 @@
...
+    _periodicTasks = new PriorityBlockingQueue<>(10, (o1, o2) -> {
+      if (o1.getExecutionTime() == o2.getExecutionTime()) {
+        return 0;
+      }
+      return o1.getExecutionTime() < o2.getExecutionTime() ? -1 : 1;
+    });
+  }
+
+  public void start() {
```
Jackie-Jiang: Recommend changing this to **start(List\<PeriodicTask\> periodicTasks)** so that we don't need to add/remove each of the tasks

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,109 @@
...
+            _periodicTasks.offer(task);
+          }
+        }
+      }
+    }, _initialDelaySeconds, TimeUnit.SECONDS);
+  }
+
+  public void stop() {
+    LOGGER.info("Stopping PeriodicTaskScheduler");
+    if (_executorService == null) {
```
Jackie-Jiang: This can never be null

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class PeriodicTaskScheduler {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PeriodicTaskScheduler.class);
+  private PriorityBlockingQueue<PeriodicTask> _periodicTasks;
+  private ScheduledExecutorService _executorService;
```
Jackie-Jiang: Make this final

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,109 @@
...
+    } catch (InterruptedException e) {
+      // Ignored
+    }
+    _executorService = null;
+  }
+
+  /**
+   * Add periodic task before starting the scheduler.
+   */
+  public void addPeriodicTask(PeriodicTask task) {
```
Jackie-Jiang: With the change of **start**, no need for add/remove

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,109 @@
...
+      }
+    }, _initialDelaySeconds, TimeUnit.SECONDS);
+  }
+
+  public void stop() {
+    LOGGER.info("Stopping PeriodicTaskScheduler");
+    if (_executorService == null) {
+      return;
+    }
+    _executorService.shutdown();
```
Jackie-Jiang: This will never terminate for an infinite loop, use **shutdownNow** instead

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -72,68 +67,30 @@
...
-      }
-    });
+  @Override
+  public void run() {
+    updateSegmentMetrics();
   }
 
   /**
-   * Starts the segment status checker.
+   * Init the segment status checker.
```
Jackie-Jiang: You can remove the javadoc for override method, same for other tasks

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,119 @@
...
+import org.slf4j.LoggerFactory;
+
+
+public class PeriodicTaskScheduler {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PeriodicTaskScheduler.class);
+  private final ScheduledExecutorService _executorService;
+  private long _initialDelayInSeconds;
+  private volatile boolean _running;
+
+  public static class PeriodicTaskEntry implements Comparable<PeriodicTaskEntry> {
```
Jackie-Jiang: Keep this as a private inner class

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTask.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package com.linkedin.pinot.core.periodictask;
+
+public interface PeriodicTask extends Runnable {
+
+  void init();
+
+  @Override
+  void run();
```
Jackie-Jiang: Why overriding this one?

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTask.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.core.periodictask;
+
+public interface PeriodicTask extends Runnable {
```
Jackie-Jiang: Please well document the interface

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -0,0 +1,56 @@
...
+
+public abstract class BasePeriodicTask implements PeriodicTask {
+  private final String _taskName;
+  private long _intervalInSeconds;
+  private long _initialDelayInSeconds;
+
+  private static final long DEFAULT_INITIAL_DELAY_IN_SECOND = 120L;
+
+  public BasePeriodicTask(String taskName, long runFrequencyInSeconds) {
+    this(taskName, runFrequencyInSeconds, DEFAULT_INITIAL_DELAY_IN_SECOND);
```
Jackie-Jiang: Don't use constant value for initial delay. **Random.nextInt(runFrequencyInSeconds)** might be better

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  public BasePeriodicTask(String taskName, long runFrequencyInSeconds, long initialDelaySeconds) {
+    _taskName = taskName;
+    _intervalInSeconds = runFrequencyInSeconds;
+    _initialDelayInSeconds = initialDelaySeconds;
+  }
+
+  public abstract void run();
```
Jackie-Jiang: Remove this override

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
jackjlli: @Jackie-Jiang Nope, as long as we set `_running` to false, it will terminate after finishing a task.
And using `shutdownNow` will throw a lot of unnecessary exceptions from either `take()` method or `sleep()` method.

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTask.java
jackjlli: javadoc added.

pinot-core/src/test/java/com/linkedin/pinot/core/periodictask/PeriodicTaskSchedulerTest.java
```diff
@@ -0,0 +1,181 @@
...
+      }
+    };
+    periodicTasks.add(task2);
+
+    long start = System.currentTimeMillis();
+    periodicTaskScheduler.start(periodicTasks);
+    try {
+      Thread.sleep(totalRunTimeInMilliseconds);
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
sunithabeeram: I am not sure catching interrupted exception is a good idea. You can add throws to the method signature and let the test fail and bail instead of continue on.

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,126 @@
...
+    PeriodicTask getPeriodicTask() {
+      return _periodicTask;
+    }
+
+    long getExecutionTime() {
+      return _executionTime;
+    }
+
+    void updateExecutionTime() {
+      _executionTime = System.currentTimeMillis() + _periodicTask.getIntervalInSeconds() * 1000L;
```
sunithabeeram: Shouldn't we check if interval is > 0 before doing this math?

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,126 @@
...
+  public PeriodicTaskScheduler(long initialDelayInSeconds) {
+    LOGGER.info("Initializing PeriodicTaskScheduler. Initial delay in seconds: {}", initialDelayInSeconds);
+    _executorService = Executors.newScheduledThreadPool(1);
+    _initialDelayInSeconds = initialDelayInSeconds;
+    _running = true;
+  }
+
+  public void start(List<PeriodicTask> periodicTasks) {
+    LOGGER.info("Starting PeriodicTaskScheduler.");
+    PriorityBlockingQueue<PeriodicTaskEntry> queue = new PriorityBlockingQueue<>(Math.max(1, periodicTasks.size()));
```
sunithabeeram: If periodicTasks is empty, why even bother with the rest? We can log a message and bail out?

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
jackjlli: The check has been done before the task gets pushed to the list.
See line 61 of `ControllerStarter.java`

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -72,68 +67,27 @@
...
+    updateSegmentMetrics();
   }
 
-  /**
-   * Starts the segment status checker.
-   */
-  public void start() {
-    if (_segmentStatusIntervalSeconds == -1) {
+  @Override
+  public void init() {
```
Jackie-Jiang: Remove this because it's already checked?

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+      return _executionTime;
+    }
+
+    void updateExecutionTime() {
+      _executionTime = System.currentTimeMillis() + _periodicTask.getIntervalInSeconds() * 1000L;
+    }
+
+    @Override
+    public int compareTo(PeriodicTaskEntry o) {
+      if (this._executionTime == o._executionTime) {
```
Jackie-Jiang: return Long.compare(this,_executionTime, o.executionTime);

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,124 @@
...
+        PeriodicTaskScheduler.PeriodicTaskEntry taskEntry = null;
+        try {
+          taskEntry = queue.take();
+          long currentTime = System.currentTimeMillis();
+          if (currentTime < taskEntry.getExecutionTime()) {
+            Thread.sleep(taskEntry.getExecutionTime() - currentTime);
+          }
+          LOGGER.info("Running Task {}", taskEntry.getPeriodicTask().getTaskName());
+          taskEntry.getPeriodicTask().run();
+        } catch (InterruptedException ie) {
```
Jackie-Jiang: Should we ignore the interrupt? We have a global Throwable catch block

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
jackjlli: Yeah sure. The check can be removed while `setStatusToDefault()` needs to be called here.

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
jackjlli: I think it's essential to keep it there. One the one hand, this allows the interrupt to propagate - the method should not continue on ignoring the interrupt, which is beneficial for us to test. On the other hand, the global throwable block just simply logs the exception/error without stopping it, so that it won't block the next following task.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
jackjlli: The check has been removed in the last push.

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
Jackie-Jiang: **Thread.currentThread().interrupt();** has no effect because the thread will terminate just after this.

jackjlli:
Added ControllerPeriodicTask abstract class in the latest push, so that Pinot lead controller's logic can be unified in the same code, and can be easily migrated in future.

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -240,7 +217,7 @@ static int computeNumMissingSegments(List<Interval> segmentIntervals, Duration f
     }
 
     // Sort the intervals by ascending starting time
-    segmentIntervals.sort((o1, o2) -> Long.compare(o1.getStartMillis(), o2.getStartMillis()));
+    segmentIntervals.sort(Comparator.comparingLong(BaseInterval::getStartMillis));
```
Jackie-Jiang: This might be slower than the existing one due to run-time compilation.

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * A base class to implement periodic task interface.
+ */
+public abstract class BasePeriodicTask implements PeriodicTask {
+  private final String _taskName;
+  private long _intervalInSeconds;
+  private long _initialDelayInSeconds;
+
+  public BasePeriodicTask(String taskName, long runFrequencyInSeconds, long initialDelaySeconds) {
```
Jackie-Jiang: initialDelay**In**Seconds?

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -0,0 +1,50 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.linkedin.pinot.core.periodictask;
+
+/**
+ * A base class to implement periodic task interface.
+ */
+public abstract class BasePeriodicTask implements PeriodicTask {
```
Jackie-Jiang: Since we have the ControllerPeriodicTask, I don't think we still need this abstract class.

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -0,0 +1,71 @@
...
+
+    if (periodicTasks.size() > CORE_POOL_SIZE) {
+      LOGGER.warn("The number of tasks:{} is more than the default number of threads:{}.", periodicTasks.size(),
+          CORE_POOL_SIZE);
+    }
+
+    LOGGER.info("Starting PeriodicTaskScheduler.");
+    // Set up an executor that executes tasks periodically
+    for (PeriodicTask periodicTask : periodicTasks) {
+      _executorService.scheduleAtFixedRate(() -> {
```
Jackie-Jiang: Rate or Delay?

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/BasePeriodicTask.java
jackjlli: I think we'll still need this class. `PeriodicTask` and `BasePeriodicTask` are so generic. They can be used in other module in future as well.

Jackie-Jiang: Make sense

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -0,0 +1,50 @@
...
+  private long _initialDelayInSeconds;
+
+  public BasePeriodicTask(String taskName, long runFrequencyInSeconds, long initialDelayInSeconds) {
+    _taskName = taskName;
+    _intervalInSeconds = runFrequencyInSeconds;
+    _initialDelayInSeconds = initialDelayInSeconds;
+  }
+
+  @Override
+  public void init() {
```
Jackie-Jiang: Let's remove this empty init() implementation, and always let child classes override this

pinot-core/src/test/java/com/linkedin/pinot/core/periodictask/PeriodicTaskSchedulerTest.java
```diff
@@ -0,0 +1,165 @@
...
+    periodicTasks.add(task);
+
+    long start = System.currentTimeMillis();
+    periodicTaskScheduler.start(periodicTasks);
+    Thread.sleep(totalRunTimeInMilliseconds);
+
+    periodicTaskScheduler.stop();
+
+    Assert.assertTrue(count.get() > 0);
+    Assert.assertTrue(count.get() == (totalRunTimeInMilliseconds / (runFrequencyInSeconds * 1000)));
```
Jackie-Jiang: Use assertEquals(), same for other places

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -0,0 +1,62 @@
...
+      return;
+    }
+    List<String> allTableNames = _pinotHelixResourceManager.getAllTables();
+    process(allTableNames);
+  }
+
+  /**
+   * Does the following logic when not being a lead controller.
+   */
+  public abstract void onBecomeNotLeader();
```
Jackie-Jiang: We should somehow rename this method because it's not **become** not leader from leader. Maybe nonLeaderCleanUp()?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -33,16 +33,20 @@ public ControllerPeriodicTask(String taskName, long runFrequencyInSeconds,
...
 
+  @Override
+  public void init() {
+  }
+
   @Override
   public void run() {
     if (!_pinotHelixResourceManager.isLeader()) {
-      onBecomeNotLeader();
+      nonLeaderCleanUp();
```
mcvsubbu: We should call this only once in the lifetime until the controller assumes leadership again. So, the base class needs to remember if it has already invoked the method. 

I would also advocate having a similar method onBecomeLeader() so that a task can implement one call that needs to happen when it gains leadership.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -71,98 +62,58 @@
...
-          // Ref: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay-java.lang.Runnable-long-long-java.util.concurrent.TimeUnit-
-          LOGGER.warn("Caught exception while running segment status checker", e);
-        }
-      }
-    }, SegmentCheckerDefaultIntervalSeconds, _segmentStatusIntervalSeconds, TimeUnit.SECONDS);
   }
 
+  @Override
+  public void onBecomeNotLeader() {
+    super.onBecomeNotLeader();
```
mcvsubbu: why invoke super? Is every sub-class required to call super? why?
In the base class, do whatever is needed on gaining leadership, and then call the subclass method.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/minion/PinotTaskManager.java
```diff
@@ -163,15 +135,41 @@ public void stopScheduler() {
...
+  /**
+   * Public API to schedule tasks. It doesn't matter whether current pinot controller is leader.
+   */
+  public Map<String, String> scheduleTasks() {
+    return scheduleTasks(_pinotHelixResourceManager.getAllTables());
+  }
+
+  @Override
+  public void init() {
+    LOGGER.info("Starting task manager with running frequency of {} seconds", getIntervalInSeconds());
```
mcvsubbu: shouldn't this be printed in onBecomeLeader() method ?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/minion/PinotTaskManager.java
```diff
@@ -163,15 +135,41 @@ public void stopScheduler() {
...
+  }
+
   /**
    * Performs necessary cleanups (e.g. remove metrics) when the controller leadership changes.
    */
-  private void nonLeaderCleanUp() {
+  @Override
+  public void onBecomeNotLeader() {
+    super.onBecomeNotLeader();
```
mcvsubbu: remove

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -0,0 +1,101 @@
...
+      processLeaderTask(allTableNames);
+    }
+  }
+
+  private void skipLeaderTask() {
+    if (_amILeader) {
+      onBecomeNotLeader();
+    }
+    _amILeader = false;
+    printSkipMessage();
```
mcvsubbu: As long as we have the task name, we can make printSkipMessage() a private method, with the task name being printed.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  private void processLeaderTask(List<String> allTableNames) {
```
mcvsubbu: nit: It is not always 'allTableNames', right? As we move to a distributed controller model? Maybe change the var name?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Processes the periodic task as lead controller.
+   * @param allTableNames List of all the table names
+   */
+  public abstract void process(List<String> allTableNames);
+
+  /**
+   * Prints message of skipping the periodic task.
+   */
+  public abstract void printSkipMessage();
```
mcvsubbu: You should not need this method to be abstract. The base class can log this message, right?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
```diff
@@ -50,55 +49,39 @@
...
-    }, 120, _runFrequencySeconds, TimeUnit.SECONDS);
+  @Override
+  public void printSkipMessage() {
+    LOGGER.info("Skipping realtime segment relocation, not leader!");
   }
 
-  public void stop() {
-    _executorService.shutdown();
+  @Override
+  public void process(List<String> allTableNames) {
```
mcvsubbu: rename allTableNames

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/retention/RetentionManager.java
```diff
@@ -43,53 +42,56 @@
...
+  }
+
+  @Override
+  public void printSkipMessage() {
+    LOGGER.info("Controller is not leader, skip");
   }
 
-  public void execute() {
+  @Override
+  public void process(List<String> allTableNames) {
```
mcvsubbu: rename allTableNames

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public void init() {
+  }
+
+  @Override
+  public void run() {
+    if (!_pinotHelixResourceManager.isLeader()) {
+      skipLeaderTask();
+    } else {
+      List<String> allTableNames = _pinotHelixResourceManager.getAllTables();
```
mcvsubbu: I would suggest adding the iterator to the base class, and handling any exception here after calling a method in the sub-class to process a single table. We have had bugs before where exceptions were not handled, so the periodic tasks stopped at the table that encountered the exception.  Ideally the state should be maintained in the base class and the sub-class methods invoked when the state changes. It also gets easy to unit test sub-classes with this framework.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
jackjlli: We can rename it once we added the logic of distributed controller model.
For now it does handle all the tables.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
jackjlli: I think what you're concerned about is the logic in `retention manager`, where the retention manager stopped if any exception occurred in **one** table. We can handle it in another PR.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/retention/RetentionManager.java
jackjlli: No need to rename it at this time.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
jackjlli: No need to rename it at this time.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
mcvsubbu: I am concerned about any new task we add failing to handle exceptions, and in genral, about unifying the task handling as much as possible. I am fine doing this in another PR.

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/retention/RetentionManager.java
```diff
@@ -43,53 +42,44 @@
...
 
-  public void start() {
+  @Override
+  public void onBecomeLeader() {
     LOGGER.info("Starting RetentionManager with runFrequencyInSeconds: {}, deletedSegmentsRetentionInDays: {}",
-        _runFrequencyInSeconds, _deletedSegmentsRetentionInDays);
-    _executorService.scheduleWithFixedDelay(this::execute, Math.min(60, _runFrequencyInSeconds), _runFrequencyInSeconds,
-        TimeUnit.SECONDS);
-    LOGGER.info("RetentionManager started");
+        getIntervalInSeconds(), _deletedSegmentsRetentionInDays);
```
mcvsubbu: We should not need run-frequency here, right? It can be logged in the base class for all the periodic tasks.

jackjlli: The purpose of the log here is to log the value of `_deletedSegmentsRetentionInDays` rather than run-frequency. The latter one has been added to the base class.

PR #3270, opened 2018-10-05
[Star-Tree V2] Added unit tests for all supported aggregations

Jackie-Jiang:
Added unit test for COUNT, MIN, MAX, SUM, AVG, MIN_MAX_RANGE, DISTINCT_COUNT_HLL, PERCENTILE_EST, PERCENTILE_TDIGEST
For AVG, MIN_MAX_RANGE, DISTINCT_COUNT_HLL, PERCENTILE_EST, PERCENTILE_TDIGEST, also tested pre-aggregated raw values


Jackie-Jiang:
@mayankshriv All of them extends the base test, and seems testng does not automatically include inner class test by default

PR #3272, opened 2018-10-05
Fix the bug in SegmentGeneratorConfig where columnSortOrder is not copied

Jackie-Jiang:


sunithabeeram:
Can you add a unit-test for this?

Jackie-Jiang:
@sunithabeeram I don't think we need a unit-test for config... Also, that method should be deprecated and replaced with the builder

PR #3273, opened 2018-10-05
[TE] frontend - aaronucsd/Added first draft of recommended metric boo…

aaronucsd:
screen shot 2018-10-05 at 9 15 10 am
…kmarking

1. added embedded bookmarking for recommended metrics for quick investigation selection.
2. Made the search for metrics wider.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-metric-dimension/template.hbs
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-metric-dimension/template.hbs
@@ -11,6 +11,8 @@
     </label>
     {{rootcause-select-metric
       selectedUrn=baseUrn
+      entities=entities
+      selectedUrns=selectedUrns
       onSelection=(action "onMetric")
     }}
   </div>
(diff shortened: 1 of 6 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-metric-dimension/template.hbs
```diff
@@ -11,6 +11,8 @@
     </label>
     {{rootcause-select-metric
       selectedUrn=baseUrn
+      metricEntities=entities
```
apucher: I'd call this "entities" throughout. Most RCA components use that convention

aaronucsd: Good catch. I had it as `entities=entities` but changed it to test passing in only metrics only to the components vs having the component derive the metrics but seems better having it in the component for separation of concerns.

aaronucsd: reverted it back to `entities=entities`

PR #3279, opened 2018-10-06
[TE] frontend - setting limits on custom MTTD in alert tuning

newsummit:
MTTD custom settings in alert tuning flow that are less than the metric's granularity will result in no effective alert filtering. Therefore, we're setting a constraint on the minimum MTTD (Minimum Time to Detect) setting.

screen shot 2018-10-05 at 7 55 58 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/manage/alert/tune/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/manage/alert/tune/route.js
@@ -65,7 +65,8 @@ const sensitivityDefaults = {
   defaultPercentChange: '0.3',
   mttdGranularityMinimums: {
     days: 24,
-    hours: 1
+    hours: 1,
+    minutes: 0.25
   }
 };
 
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/pods/manage/alert/tune/route.js
```diff
@@ -59,10 +59,14 @@ const severityMap = {
  * If no filter data is set for sensitivity, use this
  */
 const sensitivityDefaults = {
+  defaultMttdVal: '5',
   selectedSeverityOption: 'Percentage of Change',
   selectedTunePattern: 'Up and Down',
   defaultPercentChange: '0.3',
-  defaultMttdChange: '5'
+  mttdGranularityMinimums: {
+    days: 24,
```
cecilynie: Hard to understand here, why we set the days to be "24"?

newsummit: Oh, I see what you're saying here. I need to base these limits on `windowSize` and `windowUnit`. Ok - will modify.

newsummit: The granularity minimums are max number of hours (MTTD)

PR #3277, opened 2018-10-05
[TE] fronted - rca metric select editable text

newsummit:
Allowing user a way to quickly modify input search text for selected metric.

screen shot 2018-10-05 at 6 55 53 pm

PR #3281, opened 2018-10-08
[TE] rootcause - unbreak rca coldstart page

apucher:
Recent changes in #3273 removed the metric selector from the blank RCA page. This PR re-creates the original state an fixes an error on empty selection introduced in #3277

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-metric/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-metric/component.js
@@ -41,6 +41,7 @@ export default Component.extend({
     function() {
       const { selectedUrns, entities } = this.getProperties('selectedUrns', 'entities');
       const result = filterPrefix(selectedUrns, 'thirdeye:metric:')
+        .filter(urn => urn in entities)
         .map((urn) => {
           const entity = entities[urn];
           const agg = entity ? { alias: entity.label, id: entity.urn.split(':')[2] } : {};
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/rootcause/template.hbs
```diff
@@ -133,14 +133,12 @@
...
         <div class="col-xs-10">
-          {{#if selectedUrns.size}}
-            {{rootcause-select-metric-dimension
-              selectedUrn=metricUrn
-              selectedUrns=selectedUrns
-              entities=entities
-              onSelection=(action "onPrimaryChange")
-            }}
-          {{/if}}
+          {{rootcause-select-metric-dimension
```
aaronucsd: Ah i see. Makes sense.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-select-metric/component.js
```diff
@@ -41,6 +41,7 @@ export default Component.extend({
     function() {
       const { selectedUrns, entities } = this.getProperties('selectedUrns', 'entities');
       const result = filterPrefix(selectedUrns, 'thirdeye:metric:')
+        .filter(urn => urn in entities)
```
aaronucsd: If you do this check here to confirm the urn is in `entitites`, then we can remove the condition check here for `entity`.

`const agg = entity ? { alias: entity.label, id: entity.urn.split(':')[2] } : {};`

PR #3284, opened 2018-10-08
Rename StreamMetadata to StreamConfig

npawar:
StreamConfig object is the wrapper around the TableConfig.indexingConfig.streamConfig map. The next step will be to introduce stream specific stream configs eg. kafkaStreamConfig and also to unify StreamProviderConfig and StreamConfig. This renaming also ensures that we do not confuse StreamConfig with the StreamMetadataProviders

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
@@ -38,10 +38,10 @@ import com.linkedin.pinot.core.realtime.StreamProviderConfig;
 import com.linkedin.pinot.core.realtime.converter.RealtimeSegmentConverter;
 import com.linkedin.pinot.core.realtime.impl.RealtimeSegmentConfig;
 import com.linkedin.pinot.core.realtime.impl.kafka.KafkaHighLevelStreamProviderConfig;
+import com.linkedin.pinot.core.realtime.stream.StreamConfig;
 import com.linkedin.pinot.core.realtime.stream.StreamConsumerFactory;
 import com.linkedin.pinot.core.realtime.stream.StreamConsumerFactoryProvider;
 import com.linkedin.pinot.core.realtime.stream.StreamLevelConsumer;
-import com.linkedin.pinot.core.realtime.stream.StreamMetadata;
 import com.linkedin.pinot.core.segment.index.loader.IndexLoadingConfig;
 import java.io.File;
 import java.util.ArrayList;
@@ -76,7 +76,7 @@ public class HLRealtimeSegmentDataManager extends RealtimeSegmentDataManager {
   private final File resourceTmpDir;
   private final MutableSegmentImpl realtimeSegment;
   private final String tableStreamName;
-  private final StreamMetadata _streamMetadata;
+  private final StreamConfig _streamConfig;
 
   private final long start = System.currentTimeMillis();
   private long segmentEndTimeThreshold;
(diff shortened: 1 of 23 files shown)


npawar:

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -325,7 +325,7 @@ private void handleTransientStreamErrors(Exception e) throws Exception {
   protected boolean consumeLoop() throws Exception {
     _fieldExtractor.resetCounters();
     final long idlePipeSleepTimeMillis = 100;
-    final long maxIdleCountBeforeStatUpdate = (3 * 60 * 1000)/(idlePipeSleepTimeMillis + _streamMetadata.getKafkaFetchTimeoutMillis());  // 3 minute count
+    final long maxIdleCountBeforeStatUpdate = (3 * 60 * 1000)/(idlePipeSleepTimeMillis + _streamConfig.getKafkaFetchTimeoutMillis());  // 3 minute count
```
mcvsubbu: I suppose you are going to change the methods in streamconfig to not say getKafkaFetchTimeout? Does that change make sense in this PR or in a later one? Since LLRealtimeSegmentDataManager is stream independent, we should not make kafka-specific calls from this class.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
```diff
@@ -165,12 +165,12 @@ public HLRealtimeSegmentDataManager(final RealtimeSegmentZKMetadata realtimeSegm
       resourceTmpDir.mkdirs();
     }
     // create and init stream level consumer
-    _streamMetadata = new StreamMetadata(tableConfig.getIndexingConfig().getStreamConfigs());
-    _streamConsumerFactory = StreamConsumerFactoryProvider.create(_streamMetadata);
+    _streamConfig = new StreamConfig(tableConfig.getIndexingConfig().getStreamConfigs());
+    _streamConsumerFactory = StreamConsumerFactoryProvider.create(_streamConfig);
     _streamLevelConsumer = _streamConsumerFactory.createStreamLevelConsumer(
-        HLRealtimeSegmentDataManager.class.getSimpleName() + "-" + _streamMetadata.getKafkaTopicName());
+        HLRealtimeSegmentDataManager.class.getSimpleName() + "-" + _streamConfig.getKafkaTopicName());
```
mcvsubbu: Note to change getKafkaTopicName() to a more generic name

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
npawar: I will make this change in the next couple of PRs, which will introduce stream specific StreamConfigs as wrappers around StreamConfig eg. 
`KakfaStreamConfig config = KafkaStreamConfig(streamConfig)`
As part of that, will go through all methods, and split them up into generic method calls vs specific calls. I was trying to avoid logic changes in this one.

mcvsubbu: Perfect, thanks.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
npawar: Will take care of this in a following PR, which will 
1. Unify the StreamConfig and StreamProviderConfig, so that we have only a single config object moving around
2. Split up the methods in StreamConfig, keeping only generic ones in the base class. All specific ones will go into StreamConfig wrappers specific to the stream implementation

PR #3289, opened 2018-10-09
[TE] frontend - rootcause chart for daily metrics

newsummit:
Changes to RCA graph:
- If metric is daily, back up the anomaly range inspector window by 1 day (current day usually will have no data)
- Dried up the duplicate analysis range prep code a bit in rootcause/route
- Removed white space under graph
- Gave more width to anomaly range selector for minute metrics
- Preserving order of user-added metrics (new graphs would occasionally appear at top of stack)

Before (daily metric)
screen shot 2018-10-09 at 3 05 47 pm

After (daily metric)
screen shot 2018-10-09 at 4 28 58 pm

In latest commit:
- removed anomalyRange adjustment logic (need a more wholistic approach)

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
@@ -239,10 +239,11 @@ export default Component.extend({
         return {};
       }
 
-      return filterPrefix(displayableUrns, 'frontend:metric:current:')
+      const charts = filterPrefix(displayableUrns, 'frontend:metric:current:')
         .map(urn => [toMetricLabel(toMetricUrn(urn), entities), urn])
-        .sort()
         .map(t => t[1]);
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-chart/component.js
```diff
@@ -241,7 +241,6 @@ export default Component.extend({
 
       return filterPrefix(displayableUrns, 'frontend:metric:current:')
         .map(urn => [toMetricLabel(toMetricUrn(urn), entities), urn])
-        .sort()
```
apucher: is this intentional? they should be sorted

newsummit: When sorted (by id) the metrics loaded in graph don't match the order in the legend.

PR #3291, opened 2018-10-09
[TE] rootcause - anomaly end time quick fix

apucher:
Removes the artificial anomaly end time expansion. Anomalies not aligned correctly should be handled in the backend - or not at all.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
@@ -275,7 +275,7 @@ export default Route.extend(AuthenticatedRouteMixin, {
         const granularity = adjustGranularity(anomalyEntity.attributes.metricGranularity[0]);
         const metricGranularity = toMetricGranularity(granularity);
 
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -275,7 +275,7 @@ export default Route.extend(AuthenticatedRouteMixin, {
         const granularity = adjustGranularity(anomalyEntity.attributes.metricGranularity[0]);
         const metricGranularity = toMetricGranularity(granularity);
 
-        const anomalyRange = [parseInt(anomalyEntity.start, 10), parseInt(anomalyEntity.end, 10) + 1];
+        const anomalyRange = [parseInt(anomalyEntity.start, 10), parseInt(anomalyEntity.end, 10)];
```
cyy0714: Do you know why do we + 1 before? It would be good to leave a comment on the change. So next time people who changes this part knows the logic behind.

cyy0714: In the new version, the anomalies are from 12am to 12am. The endtime exclusive should be the same for all granularity.

PR #3292, opened 2018-10-09
[TE] rootcause - prevent debounce error message

apucher:


thirdeye/thirdeye-frontend/app/pods/rootcause/controller.js
```diff
@@ -786,6 +788,8 @@ export default Controller.extend({
     const { anomalyUrn, anomalyFeedback, sessionText } =
       this.getProperties('anomalyUrn', 'anomalyFeedback', 'sessionText');
 
+    if (!anomalyUrn) { return; }
+
     // debounce: do not run if destroyed
     if (this.isDestroyed) { return; }
```
aaronucsd: this vs line 764 `if (this.isDestroyed)`?

apucher: hmm?

aaronucsd: maybe dup. We can remove one?

apucher: why would this be a dup? sometime rca loads without an anomaly id, bt controller is not destroyed either

apucher: resolved offline

PR #3294, opened 2018-10-10
Removing extra copy in default crypter during segment upload

jenniferdai:
* Currently, we have two unnecessary copy calls during segment upload if we are using the default crypter. Essentially, "decrypting" the segment is not needed. Also, we do not need segment fetcher to do an extra copy; in the case of segment, we will simply start with the segment in its final location.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
@@ -36,7 +36,7 @@ import com.linkedin.pinot.controller.helix.core.PinotHelixResourceManager;
 import com.linkedin.pinot.controller.helix.core.PinotHelixSegmentOnlineOfflineStateModelGenerator;
 import com.linkedin.pinot.controller.util.TableSizeReader;
 import com.linkedin.pinot.controller.validation.StorageQuotaChecker;
-import com.linkedin.pinot.core.crypt.DefaultPinotCrypter;
+import com.linkedin.pinot.core.crypt.NoOpPinotCrypter;
 import com.linkedin.pinot.core.crypt.PinotCrypter;
 import com.linkedin.pinot.core.crypt.PinotCrypterFactory;
 import com.linkedin.pinot.core.metadata.DefaultMetadataExtractor;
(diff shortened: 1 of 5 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/crypt/DefaultPinotCrypter.java
```diff
@@ -36,21 +34,11 @@ public void init(Configuration config) {
 
   @Override
   public void encrypt(File decryptedFile, File encryptedFile) {
-    try {
-      FileUtils.copyFile(decryptedFile, encryptedFile);
-    } catch (IOException e) {
-      LOGGER.warn("Could not encrypt file");
-      FileUtils.deleteQuietly(encryptedFile);
-    }
+    return;
```
mayankshriv: With this, do we even need the DefaultCrypter?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -321,8 +321,13 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
           .validateSegment(segmentMetadata, tempSegmentDir);
 
       // Zk operations
-      completeZkOperations(enableParallelPushProtection, headers, tempEncryptedFile, provider, segmentMetadata,
-          segmentName, zkDownloadUri);
+      if (crypterClassHeader.equalsIgnoreCase(DefaultPinotCrypter.class.getName())) {
```
mayankshriv: Would like to reduce the number of 'if', if not fully eliminate them.

mcvsubbu: +1
Let us at least add a TODO here if we are in a hurry to ship the fix

pinot-core/src/main/java/com/linkedin/pinot/core/crypt/DefaultPinotCrypter.java
jenniferdai: The reason I put defaultcrypter is so that we can take the same code path for both cases ( decrypt)

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -273,36 +271,31 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
     try {
       FileUploadPathProvider provider = new FileUploadPathProvider(_controllerConf);
       String tempFileName = TMP_DIR_PREFIX + System.nanoTime();
-      tempEncryptedFile = new File(provider.getFileUploadTmpDir(), tempFileName + ENCRYPTED_SUFFIX);
       tempDecryptedFile = new File(provider.getFileUploadTmpDir(), tempFileName);
       tempSegmentDir = new File(provider.getTmpUntarredPath(), tempFileName);
 
+      // Set default crypter and encrypted file accordingly
```
mayankshriv: Update comment to reflect why encrypted file is named differently.

pinot-core/src/main/java/com/linkedin/pinot/core/crypt/DefaultPinotCrypter.java
mayankshriv: May be it should be named NoOpCrypter then?

PR #3295, opened 2018-10-10
Cleanup segment fetcher instantiation

jenniferdai:
* Cleaning up constructor of segment fetcher initialization to remove unneeded items


jenniferdai:
Discussed offline, we will leave this in the top level since we need to instantiate pinotfs anyway for other operations.

PR #3297, opened 2018-10-10
[TE] Deletion endpoint for anomaly functions

akshayrai:


akshayrai:
@apucher, deprecated the older endpoint which took only one function id as a param.

PR #3298, opened 2018-10-10
[ThirdEye][s] Fix StringUtils on empty property value

cyy0714:
An exception is thrown when encountering empty property value. The pr is to fix the exception when input property has empty property value. For example, we may have a property string, a=a;b=. We expect to have {"a":"a","b":""}. Before the fix, the StringUtils will throw an exception because it gets nothing from b's value.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
@@ -17,14 +17,13 @@
 package com.linkedin.thirdeye.datalayer.dto;
 
 import com.linkedin.thirdeye.datalayer.pojo.AnomalyFunctionBean;
-import java.io.ByteArrayInputStream;
+import com.linkedin.thirdeye.datalayer.util.StringUtils;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import org.apache.commons.collections.CollectionUtils;
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -61,7 +60,7 @@ public class AnomalyFunctionDTO extends AnomalyFunctionBean {
    * true if this function should get total metric
    */
   public boolean isToCalculateGlobalMetric() {
-    return StringUtils.isNotEmpty(this.getGlobalMetric());
+    return org.apache.commons.lang3.StringUtils.isNotEmpty(this.getGlobalMetric());
   }
 
   /**
@@ -75,7 +74,6 @@ public class AnomalyFunctionDTO extends AnomalyFunctionBean {
       props = toProperties(getProperties());
     } catch (IOException e) {
       LOGGER.warn("Failed to parse property string ({}) for anomaly function: {}", getProperties(), getId());
-
     }
     return props;
   }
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
```diff
@@ -61,7 +59,7 @@ public void setTopicMetric(String topicMetric) {
    * true if this function should get total metric
    */
   public boolean isToCalculateGlobalMetric() {
-    return StringUtils.isNotEmpty(this.getGlobalMetric());
+    return org.apache.commons.lang3.StringUtils.isNotEmpty(this.getGlobalMetric());
```
akshayrai: strike off! --org.apache.commons.lang3.--

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
```diff
@@ -70,48 +68,10 @@ public boolean isToCalculateGlobalMetric() {
...
    */
   public Properties toProperties() {
-    Properties props = null;
-    try {
-      props = toProperties(getProperties());
-    } catch (IOException e) {
-      LOGGER.warn("Failed to parse property string ({}) for anomaly function: {}", getProperties(), getId());
-
-    }
+    Properties props = StringUtils.decodeCompactedProperties(getProperties());
```
akshayrai: Please enclose in try catch

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
```diff
@@ -70,48 +68,10 @@ public boolean isToCalculateGlobalMetric() {
...
     return props;
   }
 
-  /**
-   * Parses the properties of String and returns the corresponding Properties object.
-   *
-   * @return a Properties object corresponds to the properties String of this anomaly function.
-   */
-  public static Properties toProperties(String properties) throws IOException {
```
akshayrai: This is a public method. Are you sure that it is not called by some other code?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/util/StringUtils.java
```diff
@@ -53,7 +53,13 @@ public static Properties decodeCompactedProperties(String propStr) {
     Properties props = new Properties();
     for (String part : SEMICOLON_SPLITTER.split(propStr)) {
       List<String> kvPair = EQUALS_SPLITTER.splitToList(part);
-      props.setProperty(kvPair.get(0), kvPair.get(1));
+      if (kvPair.size() == 2) {
+        props.setProperty(kvPair.get(0), kvPair.get(1));
```
akshayrai: Can you trim the result before inserting it into the map? That way we can make sure white-spaces are also cleaned up.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
cyy0714: There are two StringUtils used in the class: one is from ThirdEye and one is from apache. I'm not able to strike out org.apache.commons.lang3

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
cyy0714: Why do we want to catch the exception from decodeCompactedProperties? If the exception is caught, what is the expected behavior if there is an exception?

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
cyy0714: Thanks for pointing out.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
cyy0714: Fixed

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/dto/AnomalyFunctionDTO.java
cyy0714: Added back.

thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/datalayer/util/StringUtils.java
cyy0714: Added.

PR #3299, opened 2018-10-10
Changing SegmentFetcher init to include protocol

jenniferdai:
* Currently, we create the correct pinotFS inside the fetchSegmentToLocal method in SegmentFetcher. Instead, we want the init method to create the correct class and place that inside the segment fetcher map. This PR will add a protocol argument to the init constructor of SegmentFetcher to accomplish this.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/segment/fetcher/HdfsSegmentFetcher.java
@@ -41,7 +41,7 @@ public class HdfsSegmentFetcher implements SegmentFetcher {
   private int _retryWaitMs = RETRY_WAITIME_MS_DEFAULT;
 
   @Override
-  public void init(org.apache.commons.configuration.Configuration configs) {
+  public void init(org.apache.commons.configuration.Configuration configs, String protocol) {
     try {
       _retryCount = configs.getInt(RETRY, _retryCount);
       _retryWaitMs = configs.getInt(RETRY_WAITIME_MS, _retryWaitMs);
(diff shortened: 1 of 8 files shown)


Jackie-Jiang:
I don't think protocol should be tied to the SegmentFetcher, can you please justify this?

jenniferdai:
I think in the case of segment fetchers that are tied to fs (not http, etc), it should be tied to pinotfs, aka, pinotfs should be initialized first, and we shoul duse those same instances. In the case people override segment fetcher with non pinot fs classes, they do not need to depend on this. Segment fetchers that are tied to fs will use pinot fs methods, so should depend on the appropriate pinotfs to be instantiated appropriately

Jackie-Jiang:
There should be no enforcement such as only using HTTPS for HDFS. IMO, we should de-couple them.

jenniferdai:
I might have phrased what I was trying to say incorrectly. What I'm saying is if you're going to use the pinotfs methods, you should initialize pinotfs so that we can use those same instances in SegmentFetcher. If not, and you have configured some other non-pinot fs class, then your code will work still.

Jackie-Jiang:
I'm not talking about implementation details but high level abstraction. SegmentFetcher interface should not take **protocol** as an argument for init() method.

jenniferdai:
I think it is right now because the fetchSegmentToLocal today depends on protocol, so each segment fetcher works that way today

Jackie-Jiang:
For the new design (especially for interface design), we should not change API because of some old dependency. This change will pollute the interface and make future development harder. We need to come up with another work around.

jenniferdai:
Each fetcher can only fetch for a single protocol. SegmentFetcher should not need to be smart enough to make that determination, even in the future. Eventually, even if we will support http fetch for hdfs, the protocol is still http, so segmentfetcher should still be the http-configured one.

Jackie-Jiang:
I think you are mixing the concept of **scheme** and **protocol**. IMO they are two different concept, and we should not pass protocol as scheme into PinotFSFactory

jenniferdai:
What do you think is a better approach?

Jackie-Jiang:
Decouple FS and protocol. We might fetch from HDFS using hdfs:// or https://, so we should not bind them together

jenniferdai:
i think those need to be two separate segment fetchers (same copytolocal wont work in that case)

PR #3300, opened 2018-10-10
Adding close to PinotFS to support different fs reqs

jenniferdai:
* Adding a close method inside PinotFS and making the abstract class closeable in case certain filesystems need to do certain operations inside close.
* Also adding close to minion, server, controller shutdown


pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/PinotFSFactory.java
```diff
@@ -75,4 +75,8 @@ public static PinotFS create(String scheme) {
     }
     return pinotFS;
   }
+
+  public static Map<String, PinotFS> getFileSystemMap() {
```
Jackie-Jiang: Don't directly expose the map, add a shutDown() method instead

PR #3302, opened 2018-10-11
Fix ValidationManager, reduce logs and ZK accesses

Jackie-Jiang:
1. Fix the bug where we should use push frequency instead of time unit to count missing segments
2. Remove the log for missing segments because we already emit the metrics
3. Reduce ZK accesses by caching the instance configs

Once we detect some segments might be missing, we can list the segments and find the missing segments

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
@@ -15,9 +15,10 @@
  */
 package com.linkedin.pinot.controller.validation;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.linkedin.pinot.common.config.SegmentsValidationAndRetentionConfig;
 import com.linkedin.pinot.common.config.TableConfig;
 import com.linkedin.pinot.common.config.TableNameBuilder;
-import com.linkedin.pinot.common.metadata.ZKMetadataProvider;
 import com.linkedin.pinot.common.metadata.segment.OfflineSegmentZKMetadata;
 import com.linkedin.pinot.common.metadata.segment.RealtimeSegmentZKMetadata;
 import com.linkedin.pinot.common.metrics.ValidationMetrics;
@@ -30,15 +31,13 @@ import com.linkedin.pinot.controller.helix.core.PinotHelixResourceManager;
 import com.linkedin.pinot.controller.helix.core.realtime.PinotLLCRealtimeSegmentManager;
 import com.linkedin.pinot.core.realtime.stream.StreamConfig;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-import org.apache.helix.ZNRecord;
-import org.apache.helix.store.zk.ZkHelixPropertyStore;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.helix.model.InstanceConfig;
 import org.joda.time.Duration;
 import org.joda.time.Interval;
 import org.slf4j.Logger;
@@ -49,15 +48,13 @@ import org.slf4j.LoggerFactory;
  * Manages the segment validation metrics, to ensure that all offline segments are contiguous (no missing segments) and
  * that the offline push delay isn't too high.
  */
-
 public class ValidationManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(ValidationManager.class);
   private final ValidationMetrics _validationMetrics;
-  private final ScheduledExecutorService _executorService;
   private final PinotHelixResourceManager _pinotHelixResourceManager;
   private final long _validationIntervalSeconds;
-  private final boolean _autoCreateOnError;
   private final PinotLLCRealtimeSegmentManager _llcRealtimeSegmentManager;
+  private final ScheduledExecutorService _executorService;
 
   /**
    * Constructs the validation manager.
@@ -71,17 +68,9 @@ public class ValidationManager {
     _validationMetrics = validationMetrics;
     _pinotHelixResourceManager = pinotHelixResourceManager;
     _validationIntervalSeconds = config.getValidationControllerFrequencyInSeconds();
-    _autoCreateOnError = true;
     _llcRealtimeSegmentManager = llcRealtimeSegmentManager;
-
-    _executorService = Executors.newSingleThreadScheduledExecutor(new ThreadFactory() {
-      @Override
-      public Thread newThread(Runnable runnable) {
-        Thread thread = new Thread(runnable);
-        thread.setName("PinotValidationManagerExecutorService");
-        return thread;
-      }
-    });
+    _executorService =
+        Executors.newSingleThreadScheduledExecutor(runnable -> new Thread(runnable, "ValidationManagerThread"));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -264,120 +280,45 @@ public static long computeRealtimeTotalDocumentInSegments(
...
-          expectedIntervalCount = intervalCount + 1;
-          ++currentIndex;
-        }
-      }
-    }
-
-    return missingSegments;
-  }
-}
+}
```
jackjlli: Missing a new line at the end.

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -264,120 +280,45 @@ public static long computeRealtimeTotalDocumentInSegments(
...
-    if (frequency == null) {
-      return Collections.emptyList();
-    }
-
-    // Default segment granularity to day level if its small than hours.
-    if (frequency.getMillis() < Duration.standardHours(1).getMillis()) {
-      frequency = Duration.standardDays(1);
-    }
+  @VisibleForTesting
+  static List<Interval> computeMissingIntervals(List<Interval> intervals, Duration frequency) {
```
jackjlli: Can we mark `intervals` as `@nonnull` here?

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -264,120 +280,45 @@ public static long computeRealtimeTotalDocumentInSegments(
...
-      public int compare(Interval first, Interval second) {
-        if (first.getStartMillis() < second.getStartMillis()) {
-          return -1;
-        } else if (second.getStartMillis() < first.getStartMillis()) {
-          return 1;
+    intervals.sort((o1, o2) -> Long.compare(o1.getStartMillis(), o2.getStartMillis()));
+
+    List<Interval> missingIntervals = new ArrayList<>();
+    long frequencyMs = frequency.getMillis();
+    long lastStartTimeMs = 0;
```
jackjlli: It'd be better to use `0L`.

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -264,120 +280,45 @@ public static long computeRealtimeTotalDocumentInSegments(
...
+    for (Interval segmentInterval : intervals) {
+      long startTimeMs = segmentInterval.getStartMillis();
+      if (lastStartTimeMs != 0) {
```
jackjlli: `0L`

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -264,120 +280,45 @@ public static long computeRealtimeTotalDocumentInSegments(
   }
 
   /**
-   * Computes a list of missing intervals, given a list of existing intervals and the expected frequency of the
-   * intervals.
+   * Computes missing intervals based on the given existing intervals and the expected frequency of the intervals.
```
snleee: Can you briefly add the explanation on how we define missing intervals here?

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -122,71 +109,80 @@ public void runValidation() {
...
-        ZKMetadataProvider.getOfflineSegmentZKMetadataListForTable(propertyStore, offlineTableName);
+        _pinotHelixResourceManager.getOfflineSegmentMetadata(offlineTableName);
 
-    // Compute the missing segments if there are at least two
+    // Compute the missing segments if there are at least two segments
     int numMissingSegments = 0;
     int numSegments = offlineSegmentZKMetadataList.size();
-    if (numSegments >= 2) {
+    SegmentsValidationAndRetentionConfig validationConfig = tableConfig.getValidationConfig();
+    if (numSegments >= 2 && StringUtils.isNotEmpty(validationConfig.getTimeColumnName())) {
```
jackjlli: Why do we need to import `StringUtils`?

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
Jackie-Jiang: Added

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
Jackie-Jiang: I want to go with the convention of only annotating **nullable** because most of the arguments should be nonnull by default.

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
Jackie-Jiang: Changed to -1L

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
Jackie-Jiang: Same here

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
Jackie-Jiang: It's a null + empty check for simplicity

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
Jackie-Jiang: Added

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -122,71 +109,80 @@ public void runValidation() {
...
-          timeGranularity = offlineSegmentZKMetadata.getTimeGranularity();
+        } else {
+          segmentsWithInvalidInterval.add(offlineSegmentZKMetadata.getSegmentName());
         }
       }
-      List<Interval> missingIntervals = computeMissingIntervals(segmentIntervals, timeGranularity);
-      for (Interval missingInterval : missingIntervals) {
-        LOGGER.warn("Missing data in table {} for time interval {}", offlineTableName, missingInterval);
+      if (!segmentsWithInvalidInterval.isEmpty()) {
+        LOGGER.warn("Table: {} has segments with invalid interval: {}", offlineTableName, segmentsWithInvalidInterval);
```
mcvsubbu: I am wondering what the utlity of this log message is, given that we already set a metric with the number of missing intervals. it could also end up being a very large log message.

If the metric is non-zero, we can list the segments of the table and detect the missing segments (if the owner of the table so desires).

In any case, I am not sure this is a WARN. It should be INFO? (we will log WARN and ERROR messages if a segment push failed on the pinot controller, I suppose).

Jackie-Jiang: I would prefer WARN because this message denotes some abnormal for the table.
If we don't need the interval for missing segments, maybe we can just log the number of missing segments. What's your opinion on this?

mcvsubbu: A WARN indicates something is wrong with pinot (controller). That is not necessarily the case here (I know that it _can_ be the case and hence the question mark on should it be an info). This can be due to a failure while pushing segment on the offline side (e.g. hadoop) and pinot never got the segment. It seems to me that info is enough. 

Also, we can probably remove the log altogether, since we already have a metric and can monitor it. If we list the segments we can find the missing intervals.

Jackie-Jiang: IMO, INFO is purely for debugging purpose(we don't enable DEBUG usually), WARN is something we should keep an eye on (not necessary something goes wrong), ERROR means something unexpected happens, which probably needs bug fix.
For the log, if we got too many tables, the table name might be replaced with "allTable" to reduce the number of metrics (not happening right now for ValidationMetrics). I prefer at least keep a one line log for number of missing segments.

PR #3304, opened 2018-10-11
Changing segment fetcher init to use correct constant in controller

jenniferdai:
* Currently, we use the server constant to initialize Segment Fetcher, changing it to use the controller prefix inside ControllerStarter

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
@@ -294,6 +294,7 @@ public class CommonConstants {
   }
 
   public static class Controller {
+    public static final String PREFIX_OF_CONFIG_OF_SEGMENT_FETCHER_FACTORY = "pinot.controller.segment.fetcher";
     public static final String PREFIX_OF_CONFIG_OF_PINOT_FS_FACTORY = "pinot.controller.storage.factory";
     public static final String HOST_HTTP_HEADER = "Pinot-Controller-Host";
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerStarter.java
```diff
@@ -124,7 +124,7 @@ public void start() {
 
     Configuration pinotFSConfig = _config.subset(CommonConstants.Controller.PREFIX_OF_CONFIG_OF_PINOT_FS_FACTORY);
```
jackjlli: Same as `PREFIX_OF_CONFIG_OF_SEGMENT_FETCHER_FACTORY`. Other controller configs in `ControllerConf` are eager to have them reunion.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
```diff
@@ -294,6 +294,7 @@ public ServerType getServerType() {
   }
 
   public static class Controller {
+    public static final String PREFIX_OF_CONFIG_OF_SEGMENT_FETCHER_FACTORY = "pinot.controller.segment.fetcher";
```
jackjlli: Can we move it to `ControllerConf.java`?

jenniferdai: Talked to Jack offline, I agree that we should either put all controller-specific configs in controllerconf, server specific in server conf, etc, or put them all in common constants. Will create a separate ticket for that. I want that refactor to be separate from fixing incorrect configurations such as this one.

PR #3310, opened 2018-10-11
Adding logs for segment movement

jenniferdai:
* Adding logs for segment copyToLocal and segment copyFromLocal methods in external places rather than in fs


pinot-minion/src/main/java/com/linkedin/pinot/minion/executor/BaseMultipleSegmentsConversionExecutor.java
```diff
@@ -80,6 +80,7 @@
       for (int i = 0; i < downloadURLs.length; i++) {
```
Jackie-Jiang: for (String downloadURL : downloadURLs) {
    ...
}

jenniferdai: I like that more too, but the i is used to name the tarredSegmentFile

Jackie-Jiang: Oh, missed that, my bad

PR #3311, opened 2018-10-12
Move StreamMessageDecoder into StreamConsumer

npawar:
The data manager level doesn't need to manage the decoding of the messages. It can be handled by the stream consumer. The StreamLevelConsumer already has decoder inside the consumer. Adding a method in PartitionLevelConsumer to decode a message. This will allow us to remove decoders from LLRealtimeDataManager. The consumers are responsible for creating the decoder they need using the stream configs and StreamDecoderProvider.
The classes StreamConsumerFactory, StreamLevelConsumer and PartitionLevelConsumer show how the updated interfaces look like.


npawar:

mcvsubbu:
As discussed let us hold off on this change, and do the other change of removing StreamProviderConfig.

PR #3314, opened 2018-10-12
Fix a breaking change that download uri in zk is not the same as upload uri

fx19880617:
Hi @jenniferdai,

Our offline pipeline are all based on URI upload type. And we expected downloadUri in SegmentMetadata in Zookeeper is the same as the upload uri we set in the request header.

Currently we have segments serving by two hdfs clusters: unsecure hdfs(Example URI: _hdfs://path/to/unsecure/segment_) and secure hdfs(Example URI:  _viewfs://path/to/secure/segment_).

I put a quick fix here, but maybe you have more thoughts on the "FileUploadPathProvider", please advise.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
@@ -255,6 +255,11 @@ public class PinotSegmentUploadRestletResource {
     // Get URI of current segment location
     String currentSegmentLocationURI = headers.getHeaderString(FileUploadDownloadClient.CustomHeaders.DOWNLOAD_URI);
 
+    // Get strategy of segment persistence
+    String segmentPersistenceStratey = headers.getHeaderString(FileUploadDownloadClient.CustomHeaders.SEGMENT_PERSISTENCE_STRATEGY);
+
+
+
     File tempEncryptedFile = null;
     File tempDecryptedFile = null;
     File tempSegmentDir = null;
@@ -284,7 +289,7 @@ public class PinotSegmentUploadRestletResource {
           segmentMetadata =
               getMetadataForURI(crypterClassHeader, currentSegmentLocationURI, tempEncryptedFile, tempDecryptedFile,
                   tempSegmentDir, metadataProviderClass);
-          zkDownloadUri = getZkDownloadURIForURIUpload(currentSegmentLocationURI, segmentMetadata, provider);
+          zkDownloadUri = getZkDownloadURIForURIUpload(currentSegmentLocationURI, segmentMetadata, segmentPersistenceStratey, provider);
           break;
         case SEGMENT:
           getFileFromMultipart(multiPart, tempDecryptedFile);
@@ -327,15 +332,19 @@ public class PinotSegmentUploadRestletResource {
   }
 
   private String getZkDownloadURIForURIUpload(String currentSegmentLocationURI, SegmentMetadata segmentMetadata,
-      FileUploadPathProvider provider) throws URISyntaxException, UnsupportedEncodingException {
+      String segmentPersistenceStratey, FileUploadPathProvider provider) throws URISyntaxException, UnsupportedEncodingException {
     String zkDownloadUri;
     if (new URI(currentSegmentLocationURI).getScheme().equals(CommonConstants.Segment.LOCAL_SEGMENT_SCHEME)) {
       zkDownloadUri = ControllerConf.constructDownloadUrl(segmentMetadata.getTableName(), segmentMetadata.getName(),
           provider.getVip());
     } else {
-      LOGGER.info("Using configured data dir {}", _controllerConf.getDataDir());
-      zkDownloadUri = StringUtil.join("/", provider.getBaseDataDirURI().toString(), segmentMetadata.getTableName(),
-          URLEncoder.encode(segmentMetadata.getName(), "UTF-8"));
+      if (CommonConstants.Segment.PERSIST_TO_PINOT_MANAGED_FILESYSTEM.equalsIgnoreCase(segmentPersistenceStratey)) {
+        LOGGER.info("Using configured data dir {}", _controllerConf.getDataDir());
+        zkDownloadUri = StringUtil.join("/", provider.getBaseDataDirURI().toString(), segmentMetadata.getTableName(),
+            URLEncoder.encode(segmentMetadata.getName(), "UTF-8"));
+      } else {
+        zkDownloadUri = currentSegmentLocationURI;
+      }
     }
     return zkDownloadUri;
   }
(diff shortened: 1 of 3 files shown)


jenniferdai:
Hi @fx19880617 ,
I apologize for the backwards incompatible change but would like to discuss whether you would like to migrate to the new approach. I will happily support the migration effort/spend time with your team to help. Definitely feel free to chime in if you want the old way to still be supported. 

I have a couple of design reasons for making this backwards incompatible change. 
1. I wanted Pinot internals to have the final say about the directory structure where segments are deposited. This way, we can enable features (later) such as retentionmanager for deep storage, versioning, etc. If clients have the final say about how segments are stored, this will be much more difficult.
2. Because of the first assumption, our initial way to make sure we have the final say during segment upload is to first copy the segment to the local controller dir (thus why the local temp dir is needed) and then upload the segment to the final uri location based on the data dir configured and additional paths after that. 
3. The reason I have left it to download first to local and then upload is for example, if you are uploading across different hadoop clusters. There could be different authentication needed, so we may need to support multiple types of hadoop filesystems. I am leaving that flexibility for now and hope to potentially add a direct move later (though with encryption, etc, might be a little complex)

What are your thoughts?

fx19880617:
> Hi @fx19880617 ,
> I apologize for the backwards incompatible change but would like to discuss whether you would like to migrate to the new approach. I will happily support the migration effort/spend time with your team to help. Definitely feel free to chime in if you want the old way to still be supported.
> 
> I have a couple of design reasons for making this backwards incompatible change.
> 
> 1. I wanted Pinot internals to have the final say about the directory structure where segments are deposited. This way, we can enable features (later) such as retentionmanager for deep storage, versioning, etc. If clients have the final say about how segments are stored, this will be much more difficult.
> 2. Because of the first assumption, our initial way to make sure we have the final say during segment upload is to first copy the segment to the local controller dir (thus why the local temp dir is needed) and then upload the segment to the final uri location based on the data dir configured and additional paths after that.
> 3. The reason I have left it to download first to local and then upload is for example, if you are uploading across different hadoop clusters. There could be different authentication needed, so we may need to support multiple types of hadoop filesystems. I am leaving that flexibility for now and hope to potentially add a direct move later (though with encryption, etc, might be a little complex)
> 
> What are your thoughts?

Overall this approach sounds good, we also need to use hdfs for simple consumer segment upload. 

One concern about current provide is that it's per cluster based, not per table based. Currently we don't own users' data, which means that they are using their own hdfs quota, if we move all those stuffs under pinot, there could be some simple bad use cases quickly exhausted your namespace(e.g. user pushing too many small segments, small segment flush size for LLC segments).

For the quickfix, we could keep the original behavior and add a flag in header to indicate moving data to PinotFS for storage.

fx19880617:
@jenniferdai updated the PR for a quick fixing. This requires segment push job to add a customer header in order to use PinotFS for persistence, this should keep backward compatibility for existing segment push jobs.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -327,15 +332,19 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
...
     } else {
-      LOGGER.info("Using configured data dir {}", _controllerConf.getDataDir());
-      zkDownloadUri = StringUtil.join("/", provider.getBaseDataDirURI().toString(), segmentMetadata.getTableName(),
-          URLEncoder.encode(segmentMetadata.getName(), "UTF-8"));
+      if (CommonConstants.Segment.PERSIST_TO_PINOT_MANAGED_FILESYSTEM.equalsIgnoreCase(segmentPersistenceStratey)) {
+        LOGGER.info("Using configured data dir {}", _controllerConf.getDataDir());
+        zkDownloadUri = StringUtil.join("/", provider.getBaseDataDirURI().toString(), segmentMetadata.getTableName(),
+            URLEncoder.encode(segmentMetadata.getName(), "UTF-8"));
+      } else {
+        zkDownloadUri = currentSegmentLocationURI;
```
jenniferdai: even if you set this, inside "completeZkOperations," I again get the final segment directory and will move it there later on. Will that break your functionality?

fx19880617: hmm, so the downloadUri in zk will be override again ?

jenniferdai: nope, download uri will be right, but the segment will still be copied to local and copied to the remote place. segment will be in both places

fx19880617:
1. Temp directory is there for a while, and we use it to extract segment metadata right? What we don't want is segment movement. This actually changed the contract with our customers and involves a lot of effort to set up quota and chargeback.

2. Not really. Since moving segments to PinotFS is a new feature added, we should keep this change backward compatible and upgrade the customer pipelines gradually.

jenniferdai:
For 1, we can propagate that header through the code so that it doesn't do the move as well. 

I understand. We can keep #2 as is. Do you mind if I set that header inside the segment uri push job when we create the segment uri request?

fx19880617:
> For 1, we can propagate that header through the code so that it doesn't do the move as well.
> 
> I understand. We can keep #2 as is. Do you mind if I set that header inside the segment uri push job when we create the segment uri request?

For 1, I think so, however we expect default is to not move which is the major reason we do URI push.
For 2, we don't use existing open-source push job. It's hard to upgrade client then server.

jenniferdai:
we can take the same approach as with this header; the previous behavior we checked if downlaodUri == null before moving.

sure! I can update the open source with that header once this goes in. Should I assume the idea is to eventually phase out the header? In that case, might be nice to add a TODO also

jenniferdai:
i can help to implement that if you would like!

fx19880617:
Sounds good. Feel free to implement this.  

I feel we should keep this  *not moving" as one "feature" to support for a certain amount of time. 
There must be other users (heard slack is using S3) who also depends on this.

jenniferdai:
@fx19880617 Subbu and I talked offline about this. One option is to bring back the old endpoint and change the path of my new endpoint to /v2/segments instead so that you have exactly the same behavior as before and so we can deprecate that endpoint eventually. I want to do a design review to compare the approaches since this is an architectural change, so is it okay if I implement this on Monday? I will send out an internal email today with details.

jenniferdai:
Sorry about the delay. Really wanted to implement this today to unblock you but want agreement from the team.

fx19880617:
Adding a V2 endpoint is better. It will help client upgrade/migration.

kishoreg:
+1 on adding new endpoint. It’s ok to redirect old endpoint to new endpoint
with appropriate parameters to match the existing behavoiur

jenniferdai:
sounds good, i will combine both approaches

fx19880617:
Duplicated as https://github.com/linkedin/pinot/pull/3326

PR #3318, opened 2018-10-12
Refactor FileUploadPathProvider to be more robust

Jackie-Jiang:
1. Always append local scheme to URI if no scheme exists
2. Check whether all temp directory are local while constructing the class instance
3. Adding more logs for the location of all directories

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
@@ -15,13 +15,13 @@
  */
 package com.linkedin.pinot.controller.api.resources;
 
-import com.linkedin.pinot.common.Utils;
 import com.linkedin.pinot.common.utils.CommonConstants;
 import com.linkedin.pinot.controller.ControllerConf;
 import com.linkedin.pinot.filesystem.LocalPinotFS;
 import com.linkedin.pinot.filesystem.PinotFS;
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
```diff
@@ -30,79 +30,79 @@
...
+      mkdirIfNotExists(pinotFS, _schemasTmpDirURI);
+
+      // URIs that are always local
+      String localTempDir = controllerConf.getLocalTempDir();
+      if (localTempDir == null) {
+        LOGGER.info("Local temporary directory is not configured, use data directory as the local temporary directory");
+        _localTempDirURI = _baseDataDirURI;
       } else {
-        scheme = dataDirScheme;
+        _localTempDirURI = getUriFromPath(localTempDir);
```
jenniferdai: What input are you planning on allowing?
I see that normal directories are allowed but curious about different file:// prefixes? Do all amounts of slashes work?

Jackie-Jiang: None of them works except for file://. There is a check just after this

jenniferdai:
In that case, please update docs/microsoft folks

Jackie-Jiang:
@jenniferdai What do you mean? It's the same as current behavior, where local dir must use file://. The only difference is we allow address without scheme.

Jackie-Jiang:
All "file" scheme will work

PR #3324, opened 2018-10-15
[TE] frontend - isolate anomalyResponseObject modification to home controller

newsummit:
The home and index share controllers add an "All Resolutions" option to the globally defined anomalyResponseObj in anomaly utils. Here we are cloning that object to ensure the modified options are not used in the anomaly "labeling" options component, which needs to use the original array.

Also a tiny css tweak for the anomaly comments icon.

screen shot 2018-10-14 at 7 32 54 pm
screen shot 2018-10-14 at 7 33 04 pm

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/home/index/controller.js
+++ b/thirdeye/thirdeye-frontend/app/pods/home/index/controller.js
@@ -1,8 +1,9 @@
 import Controller from '@ember/controller';
 import floatToPercent from 'thirdeye-frontend/utils/float-to-percent';
-import { computed, set, get } from '@ember/object';
+import { computed, set, get, setProperties } from '@ember/object';
 import { isBlank } from '@ember/utils';
 import moment from 'moment';
+import _ from 'lodash';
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-frontend/app/pods/home/index/controller.js
```diff
@@ -26,20 +26,19 @@ export default Controller.extend({
...
-    get(this, 'anomalyResponseObj').push({
+    // Add ALL option to copy of global anomaly response object
+    const anomalyResponseObj = _.cloneDeep(anomalyUtil.anomalyResponseObj);
+    anomalyResponseObj.push({
       name: 'All Resolutions',
       value: 'ALL',
       status: 'All Resolutions'
     });
+    setProperties(this, {
+      anomalyResponseObj,
```
aaronucsd: i think we should rename this to something relate to `Filter By` since it was meant for that and not to be confused with the actual anomaly response list. If that makes sense.

PR #3326, opened 2018-10-15
Adding backwards compatibility for segment upload

jenniferdai:
* The new segment upload behavior is not backwards compatible with the old segment upload behavior in that the new one creates its own final location URI and will move the segment to this location while the old one took the header offered as the final location
* To remedy this, I am adding 2 endpoints, an old endpoint /segments and /v2/segments for the new segment upload (with the move) with an upload_version header.
* The old segment upload endpoint will still work if you call the old endpoint. I have also changed the default endpoints generated in our code to call the v2 endpoint.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
@@ -70,19 +70,10 @@ public class FileUploadDownloadClient implements Closeable {
     public static final String DOWNLOAD_URI = "DOWNLOAD_URI";
     public static final String SEGMENT_ZK_METADATA_CUSTOM_MAP_MODIFIER = "Pinot-SegmentZKMetadataCustomMapModifier";
     public static final String CRYPTER = "CRYPTER";
-    public static final String UPLOAD_VERSION = "UPLOAD_VERSION";
   }
 
   public static class QueryParameters {
     public static final String ENABLE_PARALLEL_PUSH_PROTECTION = "enableParallelPushProtection";
-  }
-
-  public enum UploadVersion {
-    V1, V2;
-
-    public static UploadVersion getDefaultUploadVersion() {
-      return V2;
-    }
   }
 
   public enum FileUploadType {
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
@@ -70,10 +70,19 @@ public class FileUploadDownloadClient implements Closeable {
     public static final String DOWNLOAD_URI = "DOWNLOAD_URI";
     public static final String SEGMENT_ZK_METADATA_CUSTOM_MAP_MODIFIER = "Pinot-SegmentZKMetadataCustomMapModifier";
     public static final String CRYPTER = "CRYPTER";
+    public static final String UPLOAD_VERSION = "UPLOAD_VERSION";
   }
 
   public static class QueryParameters {
     public static final String ENABLE_PARALLEL_PUSH_PROTECTION = "enableParallelPushProtection";
+  }
+
+  public enum UploadVersion {
+    V1, V2;
+
+    public static UploadVersion getDefaultUploadVersion() {
+      return V2;
+    }
   }
 
   public enum FileUploadType {
@@ -90,7 +99,7 @@ public class FileUploadDownloadClient implements Closeable {
   private static final String HTTP = "http";
   private static final String HTTPS = "https";
   private static final String SCHEMA_PATH = "/schemas";
-  private static final String SEGMENT_PATH = "/segments";
+  private static final String SEGMENT_PATH = "/v2/segments";
   private static final String SEGMENT_METADATA_PATH = "/segmentmetadata";
   private static final String TABLES_PATH = "/tables";
   private static final String TYPE_DELIMITER = "?type=";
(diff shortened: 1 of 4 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
```diff
@@ -90,7 +99,7 @@ public static FileUploadType getDefaultUploadType() {
   private static final String HTTP = "http";
   private static final String HTTPS = "https";
   private static final String SCHEMA_PATH = "/schemas";
-  private static final String SEGMENT_PATH = "/segments";
+  private static final String SEGMENT_PATH = "/v2/segments";
```
mcvsubbu: Has the old path been removed? The idea is to have both in place, right? Maybe I am missing something here. The idea is not to introduce a new header, but a new protocol endpoint.

mcvsubbu: thanks,

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -157,9 +158,10 @@ private void processExistingSegment(SegmentMetadata segmentMetadata, URI finalSe
         LOGGER.info(
             "New segment crc {} is different than the existing segment crc {}. Updating ZK metadata and refreshing segment {}",
             newCrc, existingCrc, segmentName);
-        moveSegmentToPermanentDirectory(currentSegmentLocation, finalSegmentLocationURI);
-        LOGGER.info("Moved segment {} from temp location {} to {}", segmentName,
-            currentSegmentLocation.getAbsolutePath(), finalSegmentLocationURI.getPath());
+        if (moveSegmentToFinalLocation) {
+          moveSegmentToPermanentDirectory(currentSegmentLocation, finalSegmentLocationURI);
+          LOGGER.info("Moved segment {} from temp location {} to {}", segmentName, currentSegmentLocation.getAbsolutePath(), finalSegmentLocationURI.getPath());
+        }
```
mcvsubbu: may be useful to add a log here in an else that we are skipping move to final location

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/upload/ZKOperator.java
```diff
@@ -191,14 +193,18 @@ private void checkCRC(HttpHeaders headers, String offlineTableName, String segme
...
+    // For v1 segment uploads, we will not move the segment
+    if (moveSegmentToFinalLocation) {
+      try {
+        moveSegmentToPermanentDirectory(currentSegmentLocation, finalSegmentLocationURI);
+        LOGGER.info("Moved segment {} from temp location {} to {}", segmentName, currentSegmentLocation.getAbsolutePath(),
+            finalSegmentLocationURI.getPath());
+      } catch (Exception e) {
+        LOGGER.error("Could not move segment {} from table {} to permanent directory", segmentName, rawTableName, e);
+        throw new RuntimeException(e);
+      }
```
mcvsubbu: same comment as before, may be useful to add an else log

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -326,12 +326,22 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
     }
   }
 
+  private FileUploadDownloadClient.UploadVersion getUploadVersion(String uploadVersionString) {
```
mcvsubbu: unused method?

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -326,12 +326,22 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
...
+  }
+
   private String getZkDownloadURIForURIUpload(String currentSegmentLocationURI, SegmentMetadata segmentMetadata,
-      FileUploadPathProvider provider) throws URISyntaxException, UnsupportedEncodingException {
+      FileUploadPathProvider provider, boolean moveSegmentToFinalLocation) throws URISyntaxException, UnsupportedEncodingException {
     String zkDownloadUri;
     if (new URI(currentSegmentLocationURI).getScheme().equals(CommonConstants.Segment.LOCAL_SEGMENT_SCHEME)) {
       zkDownloadUri = ControllerConf.constructDownloadUrl(segmentMetadata.getTableName(), segmentMetadata.getName(),
           provider.getVip());
+    } else if (moveSegmentToFinalLocation) {
```
fx19880617: shall this be if  (!moveSegmentToFinalLocation) ?

jenniferdai: good catch, fixed

fx19880617:
> @fx19880617 want to confirm one last time, are you using this method inside fileuploaddownloadclient? i've changed it to the new endpoint
> public static URI getUploadSegmentHttpURI(String host, int port) throws URISyntaxException {
> return getURI(HTTP, host, port, SEGMENT_PATH);

Hi, our push job will use this client, could we keep a backward compatible endpoint? Although it's probably fine for us to stick with old client for a while.

The scala code snippet of how we use it:

```
val controllerURI = FileUploadDownloadClient.getUploadSegmentHttpURI(host, Integer.parseInt(port))
val responseCode = fileUploadDownloadClient.sendSegmentUri(controllerURI, uri)
```

PR #3328, opened 2018-10-15
[Clean up] Enhance TableSizeReader to reduce the logs

Jackie-Jiang:
1. In MultiGetRequest, log a warning for failed GET request (there is no way to get the failed url info from the caller side because we don't keep the order of URLS in CompletionService)
2. In TableSizeReader, if there are missing segments from all servers, log only one lines with all missing segments
3. Also fixed the issue where getResponseBodyAsStream should be used instead of getResponseBodyAsString for large size response

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
@@ -15,16 +15,14 @@
  */
 package com.linkedin.pinot.common.http;
 
-import com.google.common.base.Preconditions;
 import java.util.List;
-import java.util.concurrent.Callable;
 import java.util.concurrent.CompletionService;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorCompletionService;
-import javax.annotation.Nonnull;
 import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.HttpConnectionManager;
 import org.apache.commons.httpclient.methods.GetMethod;
+import org.apache.commons.httpclient.params.HttpClientParams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -72,20 +70,16 @@ import org.slf4j.LoggerFactory;
 public class MultiGetRequest {
   private static final Logger LOGGER = LoggerFactory.getLogger(MultiGetRequest.class);
 
-  Executor executor;
-  HttpConnectionManager connectionManager;
+  private final Executor _executor;
+  private final HttpConnectionManager _connectionManager;
 
   /**
    * @param executor executor service to use for making parallel requests
    * @param connectionManager http connection manager to use.
    */
-  public MultiGetRequest(@Nonnull Executor executor,
-      @Nonnull HttpConnectionManager connectionManager) {
-    Preconditions.checkNotNull(executor);
-    Preconditions.checkNotNull(connectionManager);
-
-    this.executor = executor;
-    this.connectionManager = connectionManager;
+  public MultiGetRequest(Executor executor, HttpConnectionManager connectionManager) {
+    _executor = executor;
+    _connectionManager = connectionManager;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
```diff
@@ -95,24 +89,22 @@ public MultiGetRequest(@Nonnull Executor executor,
...
+    CompletionService<GetMethod> completionService = new ExecutorCompletionService<>(_executor);
+    for (String url : urls) {
+      completionService.submit(() -> {
+        try {
           GetMethod getMethod = new GetMethod(url);
           getMethod.getParams().setSoTimeout(timeoutMs);
-          // if all connections in the connection manager are busy this will wait to retrieve a connection
-          // set time to wait to retrieve a connection from connection manager
-          client.getParams().setConnectionManagerTimeout(timeoutMs);
           client.executeMethod(getMethod);
```
sunithabeeram: How is the sharing of the single client going to work? Earlier, the code was creating a new client for each url - which makes sense as they multiplex the request over the connections managed by the connection mgr. In the new code, the same client is shared - its not an async client, so I don't think this will work as expected. Let me know if I am missing something.

pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
```diff
@@ -95,24 +89,22 @@ public MultiGetRequest(@Nonnull Executor executor,
...
           return getMethod;
+        } catch (Exception e) {
+          LOGGER.warn("Caught {} while executing GET on URL: {}", e.getClass().getName(), url);
```
sunithabeeram: Might be better to log e.getMessage().

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/ServerTableSizeReader.java
```diff
@@ -80,29 +77,25 @@ public ServerTableSizeReader(Executor executor, HttpConnectionManager connection
...
-        } else if (Throwables.getRootCause(e) instanceof ConnectionPoolTimeoutException) {
-          LOGGER.warn("Server request to read table size timed out on getting a connection from pool, table: {}.", table, e);
-        } else {
-          LOGGER.warn("Execution exception while reading segment sizes for table: {}.", table, e);
-        }
-      } catch(Exception e) {
-        LOGGER.warn("Error while reading segment sizes for table: {}", table);
+        TableSizeInfo tableSizeInfo = OBJECT_MAPPER.readValue(getMethod.getResponseBodyAsStream(), TableSizeInfo.class);
+        serverToSegmentSizeInfoListMap.put(instance, tableSizeInfo.segments);
+      } catch (Exception e) {
```
sunithabeeram: Was any failure path testing done to ensure this works as expected?

pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
Jackie-Jiang: Changed to e.toString() so that class and message are both logged

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/ServerTableSizeReader.java
Jackie-Jiang: Yes, the current test covers error case.

pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
```diff
@@ -95,24 +89,23 @@ public MultiGetRequest(@Nonnull Executor executor,
...
           GetMethod getMethod = new GetMethod(url);
           getMethod.getParams().setSoTimeout(timeoutMs);
-          // if all connections in the connection manager are busy this will wait to retrieve a connection
-          // set time to wait to retrieve a connection from connection manager
-          client.getParams().setConnectionManagerTimeout(timeoutMs);
           client.executeMethod(getMethod);
           return getMethod;
+        } catch (Exception e) {
+          // Log only exception type and message instead of the whole stack trace
+          LOGGER.warn("Caught '{}' while executing GET on URL: {}", e.toString(), url);
```
jackjlli: Nice and clean here 👍

pinot-common/src/main/java/com/linkedin/pinot/common/http/MultiGetRequest.java
Jackie-Jiang: FYI https://stackoverflow.com/questions/1281219/best-practice-to-use-httpclient-in-multithreaded-environment

PR #3336, opened 2018-10-16
Add metadata info and metric when querying non-existing columns

jackjlli:
This PR adds metadata info and metric when querying non-existing columns. 
The metric is used to monitor how many tables do we currently query with non-existing columns, so that we can take some follow up actions, like including warning message in the response, etc.
The metadata info will be like ` "pruningReasons": ["DataSchemaSegmentPruner"]`. 
Thus the response output will be like:
```
{
    "selectionResults": {
        "columns": ["l_comments", "l_discount"],
        "results": []
    },
    "exceptions": [],
    "numServersQueried": 1,
    "numServersResponded": 1,
    "numSegmentsQueried": 1,
    "numSegmentsProcessed": 0,
    "numSegmentsMatched": 0,
    "numDocsScanned": 0,
    "numEntriesScannedInFilter": 0,
    "numEntriesScannedPostFilter": 0,
    "numGroupsLimitReached": false,
    "totalDocs": 7280281,
    "timeUsedMs": 8,
    "segmentStatistics": [],
    "traceInfo": {},
    "pruningReasons": ["DataSchemaSegmentPruner"]
}
```

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
@@ -34,13 +34,16 @@ import com.linkedin.pinot.core.plan.maker.InstancePlanMakerImplV2;
 import com.linkedin.pinot.core.plan.maker.PlanMaker;
 import com.linkedin.pinot.core.query.config.QueryExecutorConfig;
 import com.linkedin.pinot.core.query.exception.BadQueryRequestException;
+import com.linkedin.pinot.core.query.pruner.SegmentPruner;
 import com.linkedin.pinot.core.query.pruner.SegmentPrunerService;
 import com.linkedin.pinot.core.query.request.ServerQueryRequest;
 import com.linkedin.pinot.core.query.request.context.TimerContext;
 import com.linkedin.pinot.core.util.trace.TraceContext;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import javax.annotation.concurrent.ThreadSafe;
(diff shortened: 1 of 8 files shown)


sunithabeeram:
I haven't reviewed the full code yet, but the metadata we send back should be meaningful to the user. Can we name the field as "missing/nonExistentColumns"?

kishoreg:
Are we going to include every pruner that was run as part of the query processing in the response?

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -143,6 +147,22 @@ public DataTable processQuery(ServerQueryRequest queryRequest, ExecutorService e
         metadata.put(DataTable.NUM_ENTRIES_SCANNED_POST_FILTER_METADATA_KEY, "0");
         metadata.put(DataTable.NUM_SEGMENTS_PROCESSED, "0");
         metadata.put(DataTable.NUM_SEGMENTS_MATCHED, "0");
+
+        // adds pruning reasons to metadata.
+        StringBuilder pruningReasons = new StringBuilder();
+        for (SegmentPruner segmentPruner : matchedSegmentPruners) {
+          // emits metric when querying non-existing columns.
+          if ("DataSchemaSegmentPruner".equals(segmentPruner.toString())) {
```
sunithabeeram: This seems a bit involved to surface the information we are looking for. Can this be handled better on the broker side? Broker can cache the schema and check for columns queried against the schema?

sunithabeeram:
> Are we going to include every pruner that was run as part of the query processing in the response?

No, the intention is to provide information about non-existent columns if the user queries them. The response should list the missing columns.

pinot-core/src/main/java/com/linkedin/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
jackjlli: Yeah, putting the check to broker side seems to be the right place. 
In that case, `DataSchemaSegmentPruner` would be useless.

npawar:

PR #3339, opened 2018-10-17
Adding isDirectory method to pinotFS

jenniferdai:
* Adding isDirectory method so that we can tell whether the uri is a file or directory.
* This method is needed to support deletion / retention


pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/PinotFS.java
```diff
@@ -110,6 +110,13 @@
    */
   public abstract void copyFromLocalFile(File srcFile, URI dstUri) throws Exception;
 
+  /**
+   * Allows us the ability to determine whether the uri is a directory.
+   * @param uri location of file or directory
+   * @return true if uri is a directory, false otherwise.
+   */
+  public abstract boolean isDirectory(URI uri) throws IOException;
```
mayankshriv: Why does this need to be a URI?

jenniferdai: alternative is a string, and URI is better imo. Also consistent with the other apis

PR #3343, opened 2018-10-17
Adding way to construct old segment upload endpoint

jenniferdai:
* We want a migration path from the old segment endpoint to the new one. Adding the method back to use the old segment endpoint. Also added a new method to use the call the new endpoint.
* Changed all references in open source to use the new endpoint as well.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/utils/FileUploadDownloadClient.java
@@ -139,18 +139,22 @@ public class FileUploadDownloadClient implements Closeable {
 
   /**
    * This method calls the old segment upload endpoint. We will deprecate this behavior soon. Please call
-   * getUploadSegmentV2HttpURI to construct your request.
+   * getUploadSegmentHttpURI to construct your request.
    */
   @Deprecated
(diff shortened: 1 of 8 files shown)


jenniferdai:
@fx19880617

jenniferdai:
@Jackie-Jiang I think it may be nice to help users construct the request (in case other people wanted to migrate as well), thoughts? Maybe I can rename my new one

Jackie-Jiang:
@jenniferdai I would rather rename the old one instead, because the whole code base uses the new one.

jenniferdai:
@Jackie-Jiang makes sense, how about this most recent approach?

Jackie-Jiang:
@jenniferdai We should keep the new one as default. We can bring back the old one with name like **getDeprecatedUploadSegmentHttpURI()**. @fx19880617 Any thoughts?

fx19880617:
> @jenniferdai We should keep the new one as default. We can bring back the old one with name like **getDeprecatedUploadSegmentHttpURI()**. @fx19880617 Any thoughts?

I think my major concern is that if people blindly upgrade both server and client. Then segment URI push will still fail since v2 api on controller is not backward compatible with v1 api. 

I'm open to change my code to use the old api, however not sure if this will break other users.

PR #3344, opened 2018-10-17
[TE] rootcause - display step chart for daily metrics

apucher:
image


apucher:
still pending design review

PR #3359, opened 2018-10-22
[PINOT-7008] Include the brokerRequest in the log message when group-by operator times out

sunithabeeram:
Currently, when we see group-by timeouts on the server side, its unclear which request the server timed out on. At best we can correlate with broker logs to guess which request might have triggered the issue. This change includes the (deconstrcuted) broker request in the log to provide more clues. The code does not have access to the pql, but a log like this should still provide valuable info:
"Timed out while combining group-by results after 14985ms Broker request: BrokerRequest(querySource:QuerySource(tableName:Test_OFFLINE), aggregationsInfo:[AggregationInfo(aggregationType:count, aggregationParams:{column=*}, isInSelectList:true)], groupBy:GroupBy(topN:10, expressions:[id]))"

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/operator/CombineGroupByOperator.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/operator/CombineGroupByOperator.java
@@ -156,7 +156,8 @@ public class CombineGroupByOperator extends BaseOperator<IntermediateResultsBloc
       boolean opCompleted = operatorLatch.await(_timeOutMs, TimeUnit.MILLISECONDS);
       if (!opCompleted) {
         // If this happens, the broker side should already timed out, just log the error and return
-        String errorMessage = "Timed out while combining group-by results after " + _timeOutMs + "ms";
+        String errorMessage = "Timed out while combining group-by results after " + _timeOutMs + "ms"
+            + " Broker request: " + _brokerRequest;
         LOGGER.error(errorMessage);
         return new IntermediateResultsBlock(new TimeoutException(errorMessage));
       }


pinot-core/src/main/java/com/linkedin/pinot/core/operator/CombineGroupByOperator.java
```diff
@@ -156,7 +156,8 @@ public void runJob() {
       boolean opCompleted = operatorLatch.await(_timeOutMs, TimeUnit.MILLISECONDS);
       if (!opCompleted) {
         // If this happens, the broker side should already timed out, just log the error and return
-        String errorMessage = "Timed out while combining group-by results after " + _timeOutMs + "ms";
+        String errorMessage = "Timed out while combining group-by results after " + _timeOutMs + "ms"
+            + " Broker request: " + _brokerRequest;
```
mayankshriv: Does broker request have a requestId in it? If not, it might be better to stick in there, and just print the requestId.

kishoreg: +1 on just adding the request id. Also, can we not find the server that timed out in broker log along with pql and request Id

sunithabeeram: I looked to log just the requestId too, but broker request doesn't have it.

Typically we see a bunch of requests timeout when a bad query hits the server; in a high qps use case, its harder to see exactly what the offending request is. Sometimes, its obvious from the query pattern and sometimes its not so straightforward. This change takes the guess work away.

mayankshriv: I was proposing to find a way to have requestId available there. Also, I am unsure how this solves the problem when it is not obvious from the query pattern what the offending query is. Moreover, for long queries, it may be hard to match the brokerRequest to its corresponding PQL.

sunithabeeram: It looks like this other change that I made actually gets what we want: https://github.com/linkedin/pinot/pull/3180

That code wasn't deployed to production when the error that I noticed happened.

PR #3362, opened 2018-10-23
Limit the number of groups stored both inner segment and inter segment

Jackie-Jiang:
Motivation: when executing aggregation group-by queries, currently all
intermediate results are stored in a map on heap. When encountering
queries with huge scan and hit large number of groups, we might run
into issue of running out of heap memory (one map could end up being
multiple gigabytes). To prevent this, we need to limit the number of
groups stored in memory.
Solution: we set a threshold on number of groups allowed both inner
segment and inter segment (default 100,000). When the number of groups
go beyond this value, no new groups will be created, but we will keep
doing aggregation to the existing groups until all documents get
processed.
Heuristics: this is a best effort way of keeping the result correct
but reduce the memory footprint. The assumption is that, if the group
is not showing up in the first 100,000 groups, then this group is not
important, and probably will not show up in the final result. So it's
safe to just ignore this group.
NUM_GROUPS_LIMIT can be configured case by case, and if it's configured
as Integer.MAX_VALUE, it's the same as current behavior.

- Added configurable NUM_GROUPS_LIMIT
- Limit number of groups in GroupKeyGenerator (inner segment)
- Limit number of groups in CombineGroupByOperator (inter segments)

On-the-fly trim related code is removed for the following reason:
- With the current group-by semantic, no easy way to implement it, and
    it's never enabled
- It should not be implemented in ResultHolder because it's not
    aggregation type independent
- The new feature borrows the 'numGroupsLimit' config from it
- If we want to implement on-the-fly trim in the future, we can start
    a new design from scratch


sunithabeeram:
"Heuristics: this is a best effort way of keeping the result correct
but reduce the memory footprint. The assumption is that, if the group
is not showing up in the first 100,000 groups, then this group is not
important, and probably will not show up in the final result. So it's
safe to just ignore this group."

Wouldn't it be better if we returned an error at this point instead? Aren't we taking a chance with correctness in the current approach?

kishoreg:
> "Heuristics: this is a best effort way of keeping the result correct
> but reduce the memory footprint. The assumption is that, if the group
> is not showing up in the first 100,000 groups, then this group is not
> important, and probably will not show up in the final result. So it's
> safe to just ignore this group."
> 
> Wouldn't it be better if we returned an error at this point instead? Aren't we taking a chance with correctness in the current approach?

It's possible that the result is not exactly right even today. Currently, we only trim at the server level (before sending the result back to the broker). This PR is setting the limit at segment/intersegment level as well. We discussed the possibility of having a MAX_GROUP_BY_LIMIT_REACHED boolean (similar to partialResponse) in response.

kishoreg:
LGTM. lets add something to the response that tells the caller that maxGroupByLimit was reached.

I dont follow the explanation for removing trim related code.

Jackie-Jiang:
@kishoreg Will add a flag numGroupsLimitReached into response in another PR

PR #3366, opened 2018-10-23
[TE] rootcause - request scheduler

apucher:
This PR adds a request scheduler service to handle large data sets gracefully. The scheduler supports multi-level priority scheduling to respond to user-requests sensibly while under full load.


thirdeye/thirdeye-frontend/app/pods/services/rootcause-fetcher/service.js
```diff
@@ -0,0 +1,210 @@
...
+//
+/**
+ * Priority Queue via max-heap
+ */
+const top = 0;
+const parent = i => ((i + 1) >>> 1) - 1;
+const left = i => (i << 1) + 1;
+const right = i => (i + 1) << 1;
+
+class PriorityQueue {
```
jihaozh: I like Javascript. Have to implement heap from scratch!

thirdeye/thirdeye-frontend/app/pods/services/rootcause-fetcher/service.js
```diff
@@ -0,0 +1,210 @@
...
+      starting.pushObject(waiting.pop()); // modified in-place
+    }
+
+    setProperties(this, { running: [...remaining, ...starting] });
+
+    starting.forEach(p => {
+      fetch(p.url)
+        .then(res => { p.result = res; this._update(); })
+        .catch(err => { p.error = err; this._update(); });
+    });
```
aaronucsd: I would swap this forEach for `for..of` and add async/await if you want guarantee for the order of return responses and also if there is dependencies of something here (ex. p.result) in the codes after it. 

`
  async _update() {
..
 for (const p of starting) {
      await fetch(p.url)
        .then(res => { p.result = res; this._update(); })
        .catch(err => { p.error = err; this._update(); });
    });`

thirdeye/thirdeye-frontend/app/pods/services/rootcause-fetcher/service.js
aaronucsd: There a need for document for both Classes and also what this new service will deliver. It's good practice on all services and utils etc. So we know how we can use it elsewhere. A link to `stackoverflow` might always give us enough insight. Thanks.

thirdeye/thirdeye-frontend/app/pods/services/rootcause-fetcher/service.js
apucher: This is all about concurrency

thirdeye/thirdeye-frontend/app/pods/services/rootcause-fetcher/service.js
apucher: ok added more JS doc

PR #3369, opened 2018-10-23
Cleaning up crypter initialization

jenniferdai:
* Cleaning up crypter initialization so that the crypter classes are created upon controller/server/minion startup.
* Changed the way crypter is initialized in segment upload accordingly
* Configs will be prefixed with crypter.class.${CRYPTER_CLASS_NAME}, and each crypter will call init with the subset of configs from the ${CRYPTER_CLASS_NAME}
* Added unit test for custom crypter

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/crypt/PinotCrypterFactory.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/crypt/PinotCrypterFactory.java
@@ -15,6 +15,10 @@
  */
 package com.linkedin.pinot.core.crypt;
 
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import org.apache.commons.configuration.Configuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/crypt/PinotCrypterFactory.java
```diff
@@ -24,20 +28,54 @@
...
 
   }
 
+  /**
+   * Initializes map of crypter classes at startup time. Will initialize map with lower case simple class names.
+   * @param config
+   */
+  public static void init(Configuration config) {
+    // Get schemes and their respective classes
+    Iterator<String> keys = config.subset(CLASS).getKeys();
```
mayankshriv: Why does this method need to subset? Why doesn't the caller subset and pass the right config?

pinot-core/src/main/java/com/linkedin/pinot/core/crypt/PinotCrypterFactory.java
```diff
@@ -24,20 +28,54 @@
...
+  private static final String CLASS = "class";
 
   // Prevent factory from being instantiated
   private PinotCrypterFactory() {
 
   }
 
+  /**
+   * Initializes map of crypter classes at startup time. Will initialize map with lower case simple class names.
+   * @param config
```
mayankshriv: Would be great if you can add a sample config variable with value.

PR #3370, opened 2018-10-23
Supporting delete api for different storage options

jenniferdai:
* This PR adds support to delete Pinot segments in different storage systems using the delete api.
* We use PinotFS methods in order to do so.
* Test coverage in SegmentDeletionManagerTest and DeleteAPI Integration Test

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
@@ -173,19 +173,19 @@ public class SegmentDeletionManager {
   protected void removeSegmentFromStore(String tableNameWithType, String segmentId) {
     final String rawTableName = TableNameBuilder.extractRawTableName(tableNameWithType);
     if (_dataDir != null) {
-      URI deletedSegmentDirURI;
       URI fileToMoveURI;
       PinotFS pinotFS;
       URI dataDirURI = ControllerConf.getUriFromPath(_dataDir);
       fileToMoveURI = ControllerConf.constructSegmentLocation(_dataDir, rawTableName, segmentId);
+      URI deletedSegmentDestURI = ControllerConf.constructSegmentLocation(
+          StringUtil.join(File.separator, _dataDir, DELETED_SEGMENTS), rawTableName, segmentId);
       pinotFS = PinotFSFactory.create(dataDirURI.getScheme());
-      deletedSegmentDirURI =
-          ControllerConf.getUriFromPath(StringUtil.join(File.separator, _dataDir, DELETED_SEGMENTS, rawTableName));
+
       try {
         if (pinotFS.exists(fileToMoveURI)) {
           // Overwrites the file if it already exists in the target directory.
-          pinotFS.move(fileToMoveURI, deletedSegmentDirURI);
-          LOGGER.info("Moved segment {} from {} to {}", segmentId, fileToMoveURI.toString(), deletedSegmentDirURI.toString());
+          pinotFS.move(fileToMoveURI, deletedSegmentDestURI);
+          LOGGER.info("Moved segment {} from {} to {}", segmentId, fileToMoveURI.toString(), deletedSegmentDestURI.toString());
         } else {
           if (!SegmentName.isHighLevelConsumerSegmentName(segmentId)) {
             LOGGER.warn("Not found local segment file for segment {}" + fileToMoveURI.toString());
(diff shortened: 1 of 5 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
```diff
@@ -168,37 +174,51 @@ public void removeSegmentsFromStore(String tableNameWithType, List<String> segme
...
-        try {
+    if (_dataDir != null) {
+      URI deletedSegmentDirURI;
+      URI fileToMoveURI;
+      PinotFS pinotFS;
+      try {
+        URI dataDirURI = ControllerConf.getUriFromPath(_dataDir);
+        fileToMoveURI = ControllerConf.constructSegmentLocation(_dataDir, rawTableName, segmentId);
+        pinotFS = PinotFSFactory.create(dataDirURI.getScheme());
+        deletedSegmentDirURI = ControllerConf.getUriFromPath(StringUtil.join(File.separator, _dataDir, DELETED_SEGMENTS, rawTableName));
```
mcvsubbu: construct this in the ctor and make it final?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
```diff
@@ -168,37 +174,51 @@ public void removeSegmentsFromStore(String tableNameWithType, List<String> segme
...
+      try {
+        if (pinotFS.exists(fileToMoveURI)) {
           // Overwrites the file if it already exists in the target directory.
-          FileUtils.copyFileToDirectory(fileToMove, targetDir, false);
-          LOGGER.info("Moved segment {} from {} to {}", segmentId, fileToMove.getAbsolutePath(), targetDir.getAbsolutePath());
-          if (!fileToMove.delete()) {
-            LOGGER.warn("Could not delete file", segmentId, fileToMove.getAbsolutePath());
+          pinotFS.move(fileToMoveURI, deletedSegmentDirURI);
+          LOGGER.info("Moved segment {} from {} to {}", segmentId, fileToMoveURI.toString(), deletedSegmentDirURI.toString());
+          if (!pinotFS.delete(fileToMoveURI)) {
```
mcvsubbu: since you are calling pinotfs.move, you should not be deleting the old file here. you will always get an error. just make sure that the pinotfs.move does not throw an error if there is already a destination file

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
```diff
@@ -207,8 +227,8 @@ protected void removeSegmentFromStore(String tableNameWithType, String segmentId
    * @param retentionInDays: retention for deleted segments in days
    */
   public void removeAgedDeletedSegments(int retentionInDays) {
-    if (_localDiskDir != null) {
-      File deletedDir = new File(_localDiskDir, DELETED_SEGMENTS);
+    if (_dataDir != null) {
+      File deletedDir = new File(_dataDir, DELETED_SEGMENTS);
```
mcvsubbu: This should also be a pinotFs call, right?

jenniferdai: i will support remove aged segments in the next one, after my other PR for adding lastModified to PinotFS goes in

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
jenniferdai: I realized I need the table name to construct this, so will leave it here

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
```diff
@@ -168,37 +172,31 @@ public void removeSegmentsFromStore(String tableNameWithType, List<String> segme
...
+      deletedSegmentDirURI =
+          ControllerConf.getUriFromPath(StringUtil.join(File.separator, _dataDir, DELETED_SEGMENTS, rawTableName));
+      try {
+        if (pinotFS.exists(fileToMoveURI)) {
           // Overwrites the file if it already exists in the target directory.
-          FileUtils.copyFileToDirectory(fileToMove, targetDir, false);
-          LOGGER.info("Moved segment {} from {} to {}", segmentId, fileToMove.getAbsolutePath(), targetDir.getAbsolutePath());
-          if (!fileToMove.delete()) {
-            LOGGER.warn("Could not delete file", segmentId, fileToMove.getAbsolutePath());
+          pinotFS.move(fileToMoveURI, deletedSegmentDirURI);
```
mcvsubbu: Don't we need to construct a file URI to move to?  So, we need to include the segmentId in line 183?

jenniferdai: good catch, thank you

PR #3372, opened 2018-10-23
Reader config

mnandinic:
Added code to fetch the default file system from hadoop context. Earlier the code was working only for localFS. Also added code to set timeFormat as per the given timeFieldSpec in schema.

### DIFF
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
@@ -26,13 +26,9 @@ import com.linkedin.pinot.core.data.readers.ThriftRecordReaderConfig;
 import com.linkedin.pinot.core.indexsegment.generator.SegmentGeneratorConfig;
 import com.linkedin.pinot.core.segment.creator.impl.SegmentIndexCreationDriverImpl;
 import com.linkedin.pinot.common.data.TimeGranularitySpec;
-import com.linkedin.pinot.common.data.TimeGranularitySpec.TimeFormat;
 import com.linkedin.pinot.hadoop.job.JobConfigConstants;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
-import java.io.InputStream;
-import java.net.URL;
 import java.text.SimpleDateFormat;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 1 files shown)


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -191,26 +199,43 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
       context.write(new LongWritable(Long.parseLong(lineSplits[2])), new Text(
-          FileSystem.get(_properties).listStatus(new Path(_localHdfsSegmentTarPath + "/"))[0].getPath().getName()));
+              FileSystem.get(_properties).listStatus(new Path(_localHdfsSegmentTarPath + "/"))[0].getPath().getName()));
       LOGGER.info("Finished the job successfully");
     }
 
     protected void setSegmentNameGenerator(SegmentGeneratorConfig segmentGeneratorConfig, Integer seqId,
-        Path hdfsAvroPath, File dataPath) {
+                                           Path hdfsAvroPath, File dataPath) {
+      String timeFormatStr = segmentGeneratorConfig.getSchema().getTimeFieldSpec().getIncomingGranularitySpec().getTimeFormat();
```
jenniferdai: theoretically, if it's an offline-only use case, users may not set a schema or a time field spec or an incoming granularity spec. could we do a null check here?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -191,26 +199,43 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
+      String simpleDateTypeStr = timeFormatStr.split(":")[1];
```
jenniferdai: same here, may not be simple date format in which case we will only have one item in the array

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -191,26 +199,43 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
+
+      if (timeFormatStr.equals(TimeGranularitySpec.TimeFormat.EPOCH.toString())) {
```
jenniferdai: maybe this could go above the simple date logic to help

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -280,13 +305,17 @@ private RecordReaderConfig getReaderConfig(FileFormat fileFormat) throws IOExcep
...
-	    LOGGER.info("Reading CSV Record Reader Config from: {}", _readerConfigFile);
-	    readerConfig = new ObjectMapper().readValue(new File(_readerConfigFile), CSVRecordReaderConfig.class);
-	    LOGGER.info("CSV Record Reader Config: {}", readerConfig.toString());
           }
-	  break;
+          else {
+            LOGGER.info("Reading CSV Record Reader Config from: {}", _readerConfigFile);
+            // URL configURL = new URL(_readerConfigFile);
+
+            Path p = new Path(_readerConfigFile);
```
jenniferdai: is it possible to add a test case for csv generation/reading? What is the issue you ran into? We don't usually generate CSV internally, so I am not too familiar with this part.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -280,13 +305,17 @@ private RecordReaderConfig getReaderConfig(FileFormat fileFormat) throws IOExcep
...
-	  }
-	  else {
-	    LOGGER.info("Reading CSV Record Reader Config from: {}", _readerConfigFile);
-	    readerConfig = new ObjectMapper().readValue(new File(_readerConfigFile), CSVRecordReaderConfig.class);
-	    LOGGER.info("CSV Record Reader Config: {}", readerConfig.toString());
           }
-	  break;
+          else {
+            LOGGER.info("Reading CSV Record Reader Config from: {}", _readerConfigFile);
+            // URL configURL = new URL(_readerConfigFile);
```
jenniferdai: Can we remove these comments when it comes time to check in?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -280,13 +305,17 @@ private RecordReaderConfig getReaderConfig(FileFormat fileFormat) throws IOExcep
...
+
+            Path p = new Path(_readerConfigFile);
+            // InputStream inconfig = configURL.openStream();
```
jenniferdai: same here

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -99,20 +106,20 @@ public void setup(Context context) throws IOException, InterruptedException {
       _readerConfigFile = _properties.get(JobConfigConstants.PATH_TO_READER_CONFIG);
       if (_outputPath == null || _tableName == null) {
         throw new RuntimeException(
-            "Missing configs: " + "\n\toutputPath: " + _properties.get(JobConfigConstants.PATH_TO_OUTPUT)
-                + "\n\ttableName: " + _properties.get(JobConfigConstants.SEGMENT_TABLE_NAME));
+                "Missing configs: " + "\n\toutputPath: " + _properties.get(JobConfigConstants.PATH_TO_OUTPUT)
```
jenniferdai: which formatting scheme are you using for these changes?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
mnandinic: Not sure how and where to add the test case for this. Earlier the code was only compitable to read the config from the local file system. When the defaultFS of hadoop is other than file, this was breaking.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
mnandinic: I think it is been indented with 4 spaces. What is preferred ?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -191,26 +199,48 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
+              FileSystem.get(_properties).listStatus(new Path(_localHdfsSegmentTarPath + "/"))[0].getPath().getName()));
       LOGGER.info("Finished the job successfully");
     }
 
     protected void setSegmentNameGenerator(SegmentGeneratorConfig segmentGeneratorConfig, Integer seqId,
-        Path hdfsAvroPath, File dataPath) {
+                                           Path hdfsAvroPath, File dataPath) {
+      if(segmentGeneratorConfig.getSchema() == null ||
+              segmentGeneratorConfig.getSchema().getTimeFieldSpec() == null ||
+              segmentGeneratorConfig.getSchema().getTimeFieldSpec().getIncomingGranularitySpec() == null){
```
jenniferdai: could we use outgoing granularity spec?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -191,26 +199,48 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
 
       context.write(new LongWritable(Long.parseLong(lineSplits[2])), new Text(
-          FileSystem.get(_properties).listStatus(new Path(_localHdfsSegmentTarPath + "/"))[0].getPath().getName()));
+              FileSystem.get(_properties).listStatus(new Path(_localHdfsSegmentTarPath + "/"))[0].getPath().getName()));
       LOGGER.info("Finished the job successfully");
     }
 
     protected void setSegmentNameGenerator(SegmentGeneratorConfig segmentGeneratorConfig, Integer seqId,
-        Path hdfsAvroPath, File dataPath) {
+                                           Path hdfsAvroPath, File dataPath) {
```
jenniferdai: nit : could we store schema in a variable?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
jenniferdai: I think 2 is preferred; could we revert the formatting changes?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
jenniferdai: could we also add a check to ensure that the format is either epoch or simple date and throw an exception if it isn't?

jenniferdai:
I think a helpful test case would be ensuring that segment naming of simple date format and/or segment metadata is as expected.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
mnandinic: what happens to the case where there is only incoming granularity ?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
mnandinic: default is epoch and if simple date format is not right, it would anyways throw exception as its in try

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
jenniferdai: @Nonnull
  public TimeGranularitySpec getOutgoingGranularitySpec() {
    if (_outgoingGranularitySpec == null) {
      return _incomingGranularitySpec;
    } else {
      return _outgoingGranularitySpec;
    }
  }

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
jenniferdai: For codestyle reasons so that people can understand what the allowed formats are, I would prefer if you checked for this up front, with a comment that only epoch and simpledate time formats are allowed

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -205,8 +213,32 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
+        return;
+      }
+
+      String timeFormatStr = schema.getTimeFieldSpec().getIncomingGranularitySpec().getTimeFormat();
+
+      if (timeFormatStr.equals(TimeGranularitySpec.TimeFormat.EPOCH.toString())) {
+        return;
+      }
+
+      String simpleDateTypeStr = timeFormatStr.split(":")[1];
```
jenniferdai: could you refactor the check for valid simple date format logic into a separate method?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -205,8 +213,32 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
 
+      // Throws illegal argument exception if invalid
+      try{
+        new SimpleDateFormat(simpleDateTypeStr);
+      }
+      catch (IllegalArgumentException e) {
+        throw new RuntimeException("Could not parse simple date format " + simpleDateTypeStr, e);
+      }
+      segmentGeneratorConfig.setSimpleDateFormat(simpleDateTypeStr);
```
jenniferdai: can you refactor this line and the next into a separate method and comment why you have paired millis/with the simple date format, provide an example so outside pinot users can also follow this?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -205,8 +213,32 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
+      segmentGeneratorConfig.setSegmentTimeUnit(TimeUnit.MILLISECONDS);
```
jenniferdai: Can you also add logging statements here so we know what is beign set?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -191,8 +199,8 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
 
       try {
         String segmentName =
-            createSegment(_inputFilePath, schema, Integer.parseInt(lineSplits[2]), hdfsInputFilePath, localInputDataDir,
-                fs);
+          createSegment(_inputFilePath, schema, Integer.parseInt(lineSplits[2]), hdfsInputFilePath, localInputDataDir,
```
jenniferdai: Can we also revert this spacing?

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -205,8 +213,32 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
     }
 
     protected void setSegmentNameGenerator(SegmentGeneratorConfig segmentGeneratorConfig, Integer seqId,
-        Path hdfsAvroPath, File dataPath) {
+                                           Path hdfsAvroPath, File dataPath) {
+      Schema schema = segmentGeneratorConfig.getSchema();
+
+      if(schema == null ||
```
jenniferdai: Can we leave out the incoming granularity spec null check? That is not needed, think that is a malformed schema and should have already failed

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -205,8 +213,32 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
...
+        schema.getTimeFieldSpec() == null ||
+        schema.getTimeFieldSpec().getIncomingGranularitySpec() == null){
```
jenniferdai: Can we also add a info log that states schema is null or no time field spec, not setting time column or something aong those lines?

mnandinic:
Removed the date format conversion. The code changes for reading the config file from the default file system specified in hadoop. Tested in local and ADLS.

pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -154,7 +161,8 @@ protected void map(LongWritable key, Text value, Context context) throws IOExcep
 
       // To inherit from from the Hadoop Mapper class, you can't directly throw a general exception.
       Schema schema;
-      final FileSystem fs = FileSystem.get(new Configuration());
+      Configuration conf = context.getConfiguration();
+      final FileSystem fs = FileSystem.get(conf);
```
jenniferdai: This part seems not needed, can we revert?

PR #3374, opened 2018-10-23
Get quickstart-realtime working again

mcvsubbu:
Added the consumer factor in config, since this is now mandatory configuration for a stream.

Fixed the call to JSONObject.getString(). We bumped up org.json version from 20170516 to 20080701
in commit aa3a4c9b861bd5405a28ac6e6e86deeb16e6b809. During this time JSONObject changed to
throw exception if JSONObject.getString() is called on an object that is NOT a string type.
Used to be that the call automatically converts the value to a string, but not anymore.

Changed the call to get the object and call toString() on it (what 20170516 did)


mcvsubbu:
> Anyway, we can add an integration test case for the quickstart?

I tried adding one, but it needs a lot of surgery. The way this class is stopped is to kill the pid!!

PR #3377, opened 2018-10-24
Support .tar.gz upload for pluggable storage

jenniferdai:
* Currently, when we receive a segment file, we assume that the end destination is local. We assume that if we are uploading to deep storage, the upload will always come through URI upload.
* However, for minion uploads, where we will change the segment locally and re-upload, we will need to support uploading segment files.
* This PR adds the flexibility for the zkDownloadURI to be configured using deep storage rather than always having it be http/https

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
@@ -47,7 +47,6 @@ import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.InetAddress;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -326,23 +332,17 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
...
-    } else if (!moveSegmentToFinalLocation) {
-      LOGGER.info("Setting zkDownloadUri to {} for segment {} of table {}, skipping move", currentSegmentLocationURI,
-          segmentMetadata.getName(), segmentMetadata.getTableName());
-      zkDownloadUri = currentSegmentLocationURI;
     } else {
+      // Receiving .tar.gz segment upload for pluggable storage
       LOGGER.info("Using configured data dir {} for segment {} of table {}", _controllerConf.getDataDir(),
           segmentMetadata.getName(), segmentMetadata.getTableName());
-      zkDownloadUri = StringUtil.join("/", provider.getBaseDataDirURI().toString(), segmentMetadata.getTableName(),
+      return StringUtil.join(File.separator, provider.getBaseDataDirURI().toString(), segmentMetadata.getTableName(),
```
Jackie-Jiang: File.separator is not correct here. This is not the local file path, but the uri

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -284,19 +283,26 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
...
           segmentMetadata = getSegmentMetadata(crypterClassHeader, tempEncryptedFile, tempDecryptedFile, tempSegmentDir,
               metadataProviderClass);
-          zkDownloadUri = ControllerConf.constructDownloadUrl(segmentMetadata.getTableName(), segmentMetadata.getName(),
-              provider.getVip());
           break;
         default:
           throw new UnsupportedOperationException("Unsupported upload type: " + uploadType);
       }
 
+      zkDownloadUri = getZkDownloadURIForSegmentUpload(segmentMetadata, provider);
```
Jackie-Jiang: Move this into the "moveSegmentToFinalLocation" if check

PR #3378, opened 2018-10-24
Add a flag numGroupsLimitReached to mark responses that hit the number of groups limit

Jackie-Jiang:
1. Add the flag into the query response
2. Add the flag into the query log
3. Add a metric to track the number of queries hitting the limit

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -261,18 +261,25 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     _brokerMetrics.addPhaseTiming(rawTableName, BrokerQueryPhase.QUERY_EXECUTION,
         executionEndTimeNs - routingEndTimeNs);
 
+    // Track number of queries with number of groups limit reached
+    if (brokerResponse.isNumGroupsLimitReached()) {
+      _brokerMetrics.addMeteredTableValue(rawTableName, BrokerMeter.BROKER_RESPONSES_WITH_NUM_GROUPS_LIMIT_REACHED, 1);
+    }
+
     // Set total query processing time
     long totalTimeMs = TimeUnit.NANOSECONDS.toMillis(executionEndTimeNs - compilationStartTimeNs);
     brokerResponse.setTimeUsedMs(totalTimeMs);
 
     LOGGER.debug("Broker Response: {}", brokerResponse);
+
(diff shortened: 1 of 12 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -261,18 +261,25 @@ public BrokerResponse handleRequest(JSONObject request, @Nullable RequesterIdent
...
     // Set total query processing time
     long totalTimeMs = TimeUnit.NANOSECONDS.toMillis(executionEndTimeNs - compilationStartTimeNs);
     brokerResponse.setTimeUsedMs(totalTimeMs);
 
     LOGGER.debug("Broker Response: {}", brokerResponse);
+
     // Table name might have been changed (with suffix _OFFLINE/_REALTIME appended)
     LOGGER.info(
-        "RequestId:{}, table:{}, timeMs:{}, docs:{}/{}, entries:{}/{}, servers:{}/{}, exceptions:{}, serverStats:{}, query:{}",
+        "RequestId:{}, table:{}, timeMs:{}, docs:{}/{}, entries:{}/{}, servers:{}/{}, groupLimit:{}, exceptions:{}, serverStats:{}, query:{}",
```
kishoreg: groupLimit  -> exceededMaxGroupLimit or groupLimitReached

pinot-common/src/main/java/com/linkedin/pinot/common/response/BrokerResponse.java
```diff
@@ -91,6 +91,11 @@ String toJsonString()
    */
   long getTotalDocs();
 
+  /**
+   * Returns whether the number of groups limit has been reached.
```
kishoreg: can we link to the config that controls the max groups

kishoreg:
test case?

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
Jackie-Jiang: Changed to groupLimitReacher. Want to keep it short because it will be logged for each query.

pinot-common/src/main/java/com/linkedin/pinot/common/response/BrokerResponse.java
Jackie-Jiang: That is not possible because the config is in pinot-core... We hit this issue again lol

Jackie-Jiang:
@kishoreg Added test into the InterSegmentAggregation test set

Issue #2715, opened 2018-04-09
Segment Merge and Roll-up using Minion Framework


snleee:
# Motivation
A lot of tables in Pinot cluster have the large number of segments with the small size or few documents inside. Because Pinot uses segment level dictionary encoding and indexing, processing small segments are much more costly in terms of both space and performance.

We can improve the small segment problem by introducing the Minion merge/roll-up service that periodically picks up the multiple small segments and merge them into larger segments. In this way, Pinot can benefit both from disk storage and the query performance.

snleee:
1. Segment Merge Service (first doc by @Jackie-Jiang)
https://docs.google.com/document/d/1zoklHjbli-HIy0JAiBITABNsBOuC2jFthcjwGqsFVFQ/edit?usp=sharing

2. Segment Merge and Rollup
https://docs.google.com/document/d/1IS4h17JZot0FbVdL7WuWnSngp_3gXFHk8I27iYP9BZA/edit?usp=sharing

3. Broker Routing for Merged Segment
https://docs.google.com/document/d/1__6AZthhR8E0h-Y1C72exoIXgJQ4eFHpn0AVcgTX1sY/edit?usp=sharing

4. https://docs.google.com/document/d/1XqEdpg--ehbk4oaEP_VDNsbyVsytt2siFl06iSl0o10/edit?usp=sharing

snleee:
Updated design document
https://docs.google.com/document/d/1-AKCfXNXdoNjFIvJ87wjWwFM_38gS0NCwFrIYjYsqp8/edit#

PR #3379, opened 2018-10-24
Add segment selector and add the support for dynamic routing table

snleee:
1. Add segment selector interface and hooked in HelixExternalViewBasedRouting
2. Add support for dynamic approach for building a routing table
3. Refactoring routing table builders

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/selector/SegmentSelector.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/selector/SegmentSelector.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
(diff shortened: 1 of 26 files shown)


snleee:
#2715

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/selector/SegmentSelector.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Initiate the segment selector
+   */
+  void init(TableConfig tableConfig, ZkHelixPropertyStore<ZNRecord> propertyStore);
+
+  /**
+   * Compute and update the information required for a selector
+   * <p>Should be called whenever there is an external view change
+   */
+  void computeOnExternalViewChange(String tableNameWithType);
```
Jackie-Jiang: The segment selector is per table based right? In that case we don't need to pass in the table name.

PR #3387, opened 2018-10-26
Add LimitStringLengthRecordReader to limit the length of string columns

Jackie-Jiang:
Motivation:
Recently, we run into issues where bad data is getting indexed into Pinot.
The length of one single string value is more than 1M, and cause the
dictionary size to be huge (>2G) and also cause int overflow and then crash
the server. In most of the cases, super long string is not even queriable,
and trying to index them is usually caused by client code bug. So we decide
to add a configurable limit to the length of string columns. If the value
for the column goes over the limit, the value will be trimmed to be the
length limit.

1. Added "stringLengthLimit" into FieldSpec, so each field (column) can have
  different length limit
2. The default limit is 512 (string length but not decoded bytes length).
  The feature can be disabled by setting the limit to a value <= 0
3. Added LimitStringLengthRecordReader that works on the existing record
  readers but trim strings that are longer than the length limit

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
@@ -57,6 +57,8 @@ public abstract class FieldSpec implements Comparable<FieldSpec>, ConfigNodeLife
   private static final String DEFAULT_METRIC_NULL_VALUE_OF_STRING = "null";
   private static final byte[] DEFAULT_METRIC_NULL_VALUE_OF_BYTES = NULL_BYTE_ARRAY_VALUE;
 
+  private static final int DEFAULT_STRING_LENGTH_LIMIT = 512;
+
   @ConfigKey("name")
   protected String _name;
 
@@ -70,6 +72,9 @@ public abstract class FieldSpec implements Comparable<FieldSpec>, ConfigNodeLife
 
   @ConfigKey("defaultNullValue")
   private transient String _stringDefaultNullValue;
+
+  @ConfigKey("stringLengthLimit")
+  private int _stringLengthLimit = DEFAULT_STRING_LENGTH_LIMIT;
 
   // Transform function to generate this column, can be based on other columns
   protected String _transformFunction;
@@ -235,6 +240,15 @@ public abstract class FieldSpec implements Comparable<FieldSpec>, ConfigNodeLife
     }
   }
 
+  public int getStringLengthLimit() {
+    return _stringLengthLimit;
+  }
+
+  // Required by JSON de-serializer. DO NOT REMOVE.
+  public void setStringLengthLimit(int stringLengthLimit) {
+    _stringLengthLimit = stringLengthLimit;
+  }
+
   /**
    * Transform function if defined else null.
    * @return
@@ -260,6 +274,9 @@ public abstract class FieldSpec implements Comparable<FieldSpec>, ConfigNodeLife
     jsonObject.addProperty("dataType", _dataType.name());
     if (!_isSingleValueField) {
       jsonObject.addProperty("singleValueField", false);
+    }
+    if (_dataType == DataType.STRING && _stringLengthLimit != DEFAULT_STRING_LENGTH_LIMIT) {
+      jsonObject.addProperty("stringLengthLimit", _stringLengthLimit);
     }
     appendDefaultNullValue(jsonObject);
     return jsonObject;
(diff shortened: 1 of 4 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
```diff
@@ -71,6 +73,9 @@
   @ConfigKey("defaultNullValue")
   private transient String _stringDefaultNullValue;
 
+  @ConfigKey("stringLengthLimit")
+  private int _stringLengthLimit = DEFAULT_STRING_LENGTH_LIMIT;
```
kishoreg: Why call it string length limit? Can we not call it maxLength and use it for bytes as well. Also it’s better to check string. GetBytes size rt

pinot-core/src/main/java/com/linkedin/pinot/core/data/readers/LimitStringLengthRecordReader.java
```diff
@@ -0,0 +1,102 @@
...
+  @Override
+  public GenericRow next(GenericRow reuse) throws IOException {
+    GenericRow next = _recordReader.next(reuse);
+    for (Map.Entry<String, Integer> entry : _lengthLimits.entrySet()) {
+      String column = entry.getKey();
+      String value = (String) next.getValue(column);
+      if (value != null) {
+        int lengthLimit = entry.getValue();
+        if (value.length() > lengthLimit) {
+          next.putField(column, value.substring(0, lengthLimit));
```
kishoreg: Let’s trim by byte length to be independent of system encoding type

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
Jackie-Jiang: I don't know the behavior of trimming string in the middle of a UTF-8 encoded character. Also, we only support fixed length bytes for dictionary, so I don't know if we should trim bytes as well (then we might run into de-serialization exception issue).

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
```diff
@@ -358,23 +377,15 @@ public void postInject() {
...
    */
   public enum DataType {
-    INT,
-    LONG,
-    FLOAT,
-    DOUBLE,
-    BOOLEAN,  // Stored as STRING
-    STRING,
-    BYTES;
+    INT, LONG, FLOAT, DOUBLE, BOOLEAN,  // Stored as STRING
```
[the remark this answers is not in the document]
Jackie-Jiang: Historical issue, but no...

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
mcvsubbu: Since this is on a per-field basis, let us call it maxLength. We can choose not to implement it for BYTE_ARRAY since that can break encoding. And yes, agree on limiting the number of characters in string rather than number of encoded bytes.

pinot-core/src/main/java/com/linkedin/pinot/core/data/readers/LimitStringLengthRecordReader.java
```diff
@@ -0,0 +1,121 @@
...
+        // Multi-valued column
+        assert value instanceof Object[];
+        Object[] values = (Object[]) value;
+        int length = values.length;
+        for (int i = 0; i < length; i++) {
+          assert values[i] instanceof String;
+          String stringValue = (String) values[i];
+          if (stringValue.length() > lengthLimit) {
+            values[i] = stringValue.substring(0, lengthLimit);
+            _numStringsTrimmed++;
```
mcvsubbu: Do u want to keep this stat on a per column basis?

Jackie-Jiang:
@mcvsubbu Good point for real-time. I'll re-structure code to keep them the same behavior.

fx19880617:
Another point for realtime ingestion is that we could put trimming logic into message decoder to reduce memory footprint.

pinot-common/src/main/java/com/linkedin/pinot/common/data/FieldSpec.java
mcvsubbu: @fx19880617 disagree re maxBytes.  If we call it maxBytes it will somehow imply that we trim to that many bytes even if is a String. As pointed out by @Jackie-Jiang it does not make sense to trim bytes such that a String cannot be decoded. Let us call it maxLength and the semantics can be based on the data type. For ByteArray, it will translate to max length of the array.

mcvsubbu:
@fx19880617 the decoder is an configurable class, and trimming it there will need that every decoder implement the trimming. I think trimming it outside the decoder is better.

Jackie-Jiang:
Close and replace with #3445

PR #3388, opened 2018-10-26
[TE] rootcause - support query params for external setup of investigation

apucher:
This PR adds one-time query params to set up the RCA investigation view of ThirdEye via an external link. Supported arguments are:
* contextUrnsInit (comma separated urns - metrics under investigation)
* selectedUrnsInit (comma separated urns - selected auxiliary metrics)
* anomalyUrnsInit (comma separated urns - anomalies and associated metrics)
* anomalyRange (comma separated tuple, the start and end of the investigation period)
* analysisRange (comma separated tuple, the start and end of the display range)
* granularity (display granularity for time series)
* compareMode (default baseline for analysis and display)

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
@@ -95,6 +95,19 @@ const toAnalysisOffset = (granularity) => {
 };
 
 /**
+ * Returns a Set of urns augmented for virtual frontend metrics.
+ * @example ([thirdeye:metric:1, thirdeye:event:holiday:2]) =>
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/rootcause/route.js
```diff
@@ -369,7 +436,16 @@ export default Route.extend(AuthenticatedRouteMixin, {
       selectedUrns,
       sizeMetricUrns,
       setupMode,
-      context
+      context,
+
+      // reset overrides
+      contextUrnsInit: undefined,
```
aaronucsd: Could this be reset in resetController?

apucher: I'm not happy with this thing overall. It should wipe any query params off the RCA url once the RCA is set up

PR #3389, opened 2018-10-26
[TE] Read alerter configurations as a Map for easy extensibility 

akshayrai:
Changes:
* Removed explicit getters and setters for smtpConfiguration and introduced a generic alerterConfiguration which hold a Map of alerter to corresponding properties.

`cat detector.yml`
```
...
alerterConfiguration:
  smtpConfiguration:
    smtpHost: email.corp.linkedin.com
    smtpPort: 25
    smtpUser: n/a
    smtpPassword: n/a
  alerter_configuration:
    prop: value
...
```


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/common/ThirdEyeConfiguration.java
```diff
@@ -174,4 +168,11 @@ public void setDataSources(String dataSources) {
     this.dataSources = dataSources;
   }
 
+  public Map<String, Map<String, String>> getAlerterConfiguration() {
```
jihaozh: Maybe change the return type to Map<String, Object>, would add more flexibility.

akshayrai: I am not very confident of making it an Object. These are platform metadata. I don't foresee any benefits by making it an object. In addition, making it an object will require us to call toString each time we extract the properties.

jihaozh: just in case the configuration might be a list or something else later. such as 

>   alerter_configuration: prop: [value, value]

thirdeye usually uses a MapUtils library to get properties out of Map<String, Object> configs.

not a blocker.

PR #3399, opened 2018-10-30
[TE] yaml - detection annotation

jihaozh:
- annotations to mark detection pipeline stage implementations
- the endpoint to return all annotation information
- detection registry extract from stage class implementations


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/annotation/DetectionTags.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.linkedin.thirdeye.detection.annotation;
+
+public class DetectionTags {
```
apucher: I wonder if this could be an Enum instead

jihaozh: changed to the Enum type.

PR #3402, opened 2018-10-31
Introduce OffsetCriteria class to handle different offset types and operations

npawar:
The OffsetCriteria class will allow us to handle smallest, largest, period and custom offsets for all streams


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -95,7 +96,8 @@
 
 public class PinotLLCRealtimeSegmentManager {
   public static final Logger LOGGER = LoggerFactory.getLogger(PinotLLCRealtimeSegmentManager.class);
-  private static final String KAFKA_SMALLEST_OFFSET = "smallest";
+  private static final OffsetCriteria SMALLEST_OFFSET_CRITERIA =
```
mcvsubbu: Minor: Do you want to make SMALLEST_OFFSET_CRITERIA a public static in OffsetCriteria class?

npawar: yes that would be a better place. Will change it

npawar:

PR #3418, opened 2018-10-31
Segment name generator with _postfix

mnandinic:
If the time column name exists, _segmentNamePostfix = "postfix", and _sequenceId = 1, the segment name should be tableName_minDate_maxDate_postfix_1 , rather than ableName_minDate_maxDate_postfix-1

### DIFF
--- a/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
+++ b/pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
@@ -224,8 +224,10 @@ public class HadoopSegmentCreationMapReduceJob {
       segmentGeneratorConfig.setFormat(fileFormat);
       segmentGeneratorConfig.setOnHeap(true);
 
+//      If the time column name exists, _segmentNamePostfix = "postfix", and _sequenceId = 1, the segment name would be
+//      tableName_minDate_maxDate_postfix_1
       if (null != _postfix) {
-        segmentGeneratorConfig.setSegmentNamePostfix(String.format("%s-%s", _postfix, seqId));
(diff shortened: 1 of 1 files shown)


pinot-hadoop/src/main/java/com/linkedin/pinot/hadoop/job/mapper/HadoopSegmentCreationMapReduceJob.java
```diff
@@ -224,8 +224,10 @@ protected String createSegment(String dataFilePath, Schema schema, Integer seqId
       segmentGeneratorConfig.setFormat(fileFormat);
       segmentGeneratorConfig.setOnHeap(true);
 
+//      If the time column name exists, _segmentNamePostfix = "postfix", and _sequenceId = 1, the segment name would be
+//      tableName_minDate_maxDate_postfix_1
       if (null != _postfix) {
-        segmentGeneratorConfig.setSegmentNamePostfix(String.format("%s-%s", _postfix, seqId));
```
npawar: Not quite sure why it needs to be "_" and not "-". Could you please elaborate?

mnandinic: Everywhere in the code it is mentioned that this is the format. If you look at DefaultSegmentNameGenerator, the comments say

 * If the time column name exists, _segmentNamePostfix = "postfix", and _sequenceId = 1, the segment name would be
   * tableName_minDate_maxDate_postfix_1

But when I tried to create segment, it was _postfix-sequenceId. After looking into the code, found this.

jenniferdai: I think we should leave this as is. Uber guys committed this parameter for their use, and I do not want to change how their segment names are formatted. I might have added the wrong comment for the sample segment name, if you would like to fix that instead.

npawar:

PR #3421, opened 2018-10-31
Fix a potential typo in QuerySchedulerFactory

haibow:
Likely a typo. Similar to `tokenbucket`, we should also check schedulerName for `bounded_fcfs`.


haibow:
@atumbde could you help to take a look?

haibow:
Thanks @jackjlli ! Could you help to merge it? I don't have permission to do it.

PR #3424, opened 2018-11-01
Add RecordTransformer to perform transformation to incoming record (GenericRow)

Jackie-Jiang:
Replace the PlainFieldExtractor with record transformers for flexibility
When reading records from record reader or consuming realtime records, need to perform transformation.
When converting realtime segment into offline segment, no need to perform transformation because it's already done while consuming.

Default record transformers:
1. TimeTransformer: converts incoming time into outgoing time
2. ExpressionTransformer: evaluate expressions and insert values into record
3. DataTypeTransformer: converts values in record to data types specified in the schema
4. SanitationTransformer: sanitizes values (remove '\0' from string, put byte[] into ByteArray)

CompoundTransformer can take a list of transformers and perform all the transformations


pinot-core/src/main/java/com/linkedin/pinot/core/data/recordtransformer/DataTypeTransformer.java
```diff
@@ -0,0 +1,121 @@
...
+  @Override
+  public GenericRow transform(GenericRow record) {
+    for (Map.Entry<String, PinotDataType> entry : _dataTypes.entrySet()) {
+      String column = entry.getKey();
+      PinotDataType dest = entry.getValue();
+      Object value = record.getValue(column);
+
+      // NOTE: should not be null in normal case if combined with other transformers (without TimeTransformer, outgoing
+      // time column might be null)
+      if (value == null) {
```
mcvsubbu: Where is the value being set to default null (if incoming is null)?

pinot-core/src/test/java/com/linkedin/pinot/core/data/recordtransformer/RecordTransformerTest.java
```diff
@@ -0,0 +1,136 @@
...
+    TimeTransformer transformer = new TimeTransformer(SCHEMA);
+    GenericRow record = getRecord();
+    for (int i = 0; i < NUM_ROUNDS; i++) {
+      record = transformer.transform(record);
+      assertNotNull(record);
+      assertEquals(record.getValue("outgoing"), 123 * 6 * 3600 * 1000L);
+    }
+  }
+
+  @Test
```
mcvsubbu: I am not sure all the removed tests from PlainFieldExtractorTest are covered here. I see some tests there for a random class that has a toString method, and also multiple time-spec transformation tests

pinot-core/src/main/java/com/linkedin/pinot/core/data/recordtransformer/DataTypeTransformer.java
Jackie-Jiang: Default null is set before getting to the record transformer. Value should never be null here in normal case.

pinot-core/src/test/java/com/linkedin/pinot/core/data/recordtransformer/RecordTransformerTest.java
Jackie-Jiang: Here we test the random class with toString() in the "svString" column. Time converter has separate tests (DefaultTimeConverterTest).

pinot-core/src/main/java/com/linkedin/pinot/core/data/recordtransformer/CompoundTransformer.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class CompoundTransformer implements RecordTransformer {
+  private final List<RecordTransformer> _transformers;
+
+  /**
+   * Returns a record transformer that performs time transform, expressions transform and data type transform.
+   */
+  public static CompoundTransformer getDefaultTransformer(Schema schema) {
+    return new CompoundTransformer(
+        Arrays.asList(new TimeTransformer(schema), new ExpressionTransformer(schema), new DataTypeTransformer(schema)));
```
mcvsubbu: Can you add a comment (preferably a story) here about how certain transformers have to be before others? The CompoundTransformer class somehow implies that the order does not matter, but it does, doesn't it?

pinot-core/src/main/java/com/linkedin/pinot/core/data/recordtransformer/DataTypeTransformer.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  @Override
+  public GenericRow transform(GenericRow record) {
+    for (Map.Entry<String, PinotDataType> entry : _dataTypes.entrySet()) {
+      String column = entry.getKey();
+      PinotDataType dest = entry.getValue();
+      Object value = record.getValue(column);
+
```
mcvsubbu: should you continue if it is a time field-spec?
Somehow i feel it is better to fold in time transformation into record transformation (like before).

Jackie-Jiang: By this point, all fields should be treated the same (such as expression fields). DataTypeTransformer should be the last transformer which guarantees that all the values follow the field spec.

npawar:
What kind of flexibility are we achieving by moving from PlainFieldExtractor to RecordTransformer? What was the issue with the PlainFieldExtractor which prompted this change?

pinot-core/src/main/java/com/linkedin/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
```diff
@@ -318,16 +324,18 @@ private void buildRaw() throws Exception {
...
       recordReader.rewind();
       LOGGER.info("Start building IndexCreator!");
-      GenericRow readRow = new GenericRow();
-      GenericRow transformedRow = new GenericRow();
+      GenericRow readRow = null;
       while (recordReader.hasNext()) {
         long start = System.currentTimeMillis();
-        transformedRow = readNextRowSanitized(readRow, transformedRow);
+        readRow = GenericRow.createOrReuseRow(readRow);
+        GenericRow transformedRow = _recordTransformer.transform(recordReader.next(readRow));
```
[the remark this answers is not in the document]
Jackie-Jiang: Good point, and one problem this PR is trying to address is removing this extra transform. I added a constructor to specify the record transformer, and for realtime segment conversion, the record transformer is the pass-through one (no-op transformer).

pinot-core/src/test/java/com/linkedin/pinot/core/data/recordtransformer/RecordTransformerTest.java
```diff
@@ -0,0 +1,136 @@
...
+import com.linkedin.pinot.core.data.GenericRow;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.*;
+
+
+public class RecordTransformerTest {
```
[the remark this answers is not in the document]
Jackie-Jiang: Added

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
```diff
@@ -244,31 +242,8 @@ public void run() {
...
-            serverMetrics.addMeteredTableValue(tableStreamName,
-                ServerMeter.ROWS_WITH_ERRORS, (long) numErrors);
-          }
-          Map<String, Integer> errorCount = extractor.getErrorCount();
-          for (String column : errorCount.keySet()) {
-            if ((numErrors = errorCount.get(column)) > 0) {
-              segmentLogger.warn("Column {} had {} rows with errors", column, numErrors);
-            }
-          }
-          if ((numConversions = extractor.getTotalConversions()) > 0) {
```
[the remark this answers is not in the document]
Jackie-Jiang: We keep the number of rows errored metric, and remove the numNulls (always 0), and nulConversions (which IMO is not very useful). I don't think there is a good way to confirm with open source users.

pinot-core/src/main/java/com/linkedin/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
```diff
@@ -244,31 +242,8 @@ public void run() {
...
-            serverMetrics.addMeteredTableValue(tableStreamName,
-                ServerMeter.ROWS_NEEDING_CONVERSIONS, (long) numConversions);
-            segmentLogger.info("{} rows needed conversions ", numConversions);
-          }
-          if ((numNulls = extractor.getTotalNulls()) > 0) {
-            serverMetrics.addMeteredTableValue(tableStreamName,
-                ServerMeter.ROWS_WITH_NULL_VALUES, (long) numNulls);
-            segmentLogger.info("{} rows had null columns", numNulls);
-          }
-          if ((numNullCols = extractor.getTotalNullCols()) > 0) {
```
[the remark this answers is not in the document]
Jackie-Jiang: Done

Jackie-Jiang:
@npawar One major limitation with PlainFieldExtractor is all transforms are fixed and cannot be changed. With the abstraction of RecordTransformer, we can easily combine multiple RecordTransformers to achieve different requirement, and each RecordTransformer need to only focus on one functionality, such as time conversion, expression evaluation, data type conversion etc. This can also make adding new transformers way easier. For converting realtime segment to offline segment, we can use a PassThrough transformer because all the transform has already been applied.

PR #3425, opened 2018-11-01
Move group id to StreamConfig level

npawar:
The group id in the stream configs should be answered as a generic property from the StreamConfig class.


npawar:

PR #3426, opened 2018-11-01
Add stream agnostic metrics for realtime

npawar:
Some of the metrics in Controller and Server have KAFKA in their names. We cannot directly rename them, as we will lose out on alerts setup on the old names, if any. 
The plan is to introduce new metrics which have stream agnostic names. We will start using the new metrics, while also having the older one around. Once we have collected sufficient days of metrics for the enw one, we can phase out the old one.
We need to announce to the open source community that the older metrics will be going away eventually, and a shift to the new ones is required.


npawar:

PR #3433, opened 2018-11-06
PINOT-7296: Reduce lock contention by not sharing DecimalFormat instances

sunithabeeram:
We see very high latencies for requests that select a large number of rows from Pinot (ie, requests of type "select col1, ... coln from tableX where ..."). For one such query, we noticed that returning ~9k rows with 5 columns selected resulted in a latency of 2s, whereas selecting about 3k rows showed a latency of 21ms! The latency increase was due to broker code. Thread dumps showed most of the threads blocked on a lock used in DecimalFormat:

"qtp548554586-249275" #249275 prio=5 os_prio=0 tid=0x00007f372818e000 nid=0x3a53 waiting for monitor entry [0x00007f33df1ba000]
java.lang.Thread.State: BLOCKED (on object monitor)
at java.text.DecimalFormat.format(DecimalFormat.java:703)

waiting to lock <0x00000002c0eb6718> (a java.text.DigitList)
at java.text.DecimalFormat.format(DecimalFormat.java:646)
at java.text.NumberFormat.format(NumberFormat.java:316)
at com.linkedin.pinot.core.query.selection.SelectionOperatorUtils.getFormattedValue(SelectionOperatorUtils.java:580)

It is unclear from the code as to why shared instances of DecimalFormat were used. Creating new instances per query improved the latencies a lot (fetching 9k rows now takes ~60-80ms) - there was no increase in GC pressure due to new object creation as the objects of DecimalFormat seem to be pretty light-weight.

Testing done:
Performance tested the new code against the benchmark that had latency issues - we are able to scale better now.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/query/selection/SelectionOperatorUtils.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/query/selection/SelectionOperatorUtils.java
@@ -66,14 +66,11 @@ public class SelectionOperatorUtils {
   private SelectionOperatorUtils() {
   }
 
-  private static final DecimalFormat INT_FORMAT =
-      new DecimalFormat("##########", DecimalFormatSymbols.getInstance(Locale.US));
-  private static final DecimalFormat LONG_FORMAT =
-      new DecimalFormat("####################", DecimalFormatSymbols.getInstance(Locale.US));
-  private static final DecimalFormat FLOAT_FORMAT =
-      new DecimalFormat("#########0.0####", DecimalFormatSymbols.getInstance(Locale.US));
-  private static final DecimalFormat DOUBLE_FORMAT =
-      new DecimalFormat("###################0.0#########", DecimalFormatSymbols.getInstance(Locale.US));
+  private static final String INT_PATTERN = "##########";
+  private static final String LONG_PATTERN = "####################";
+  private static final String FLOAT_PATTERN = "#########0.0####";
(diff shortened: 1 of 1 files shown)


mcvsubbu:
Good find!
We could even use thread-local DecimalFormat object (if it comes to GC-ing)

sunithabeeram:
@mcvsubbu agreed. I did consider using thread-locals, but didn't find a specific need for it at this point. (Thread-locals tend to bite back at times, so didnt' want to prematurely optimize)

Also, its not clear why DecimalFormat was chosen in the first place. Simple String formatting could have done the job, IMO. There are some minor formatting differences between the 2 though, so I didn't attempt switching away from DecimalFormat usage as this point.

Jackie-Jiang:
Good finding. The lock is from DecimalFormat.java line 615 **synchronized(digitList)**. I recommend using either thread local or **clone()** instead of creating a format each time, as constructing DecimalFormat is quite expensive.

sunithabeeram:
Clone makes a shallow copy, so I don't think it will address the concern here. I would like to be careful about what we keep in thread-local - I've seen cases where its been abused and we take a hit due to that (holding on to more objects than needed; references not getting cleared when they ought to etc).

The suggestions are valid, but I would like to make sure the changes are warranted before making them. At this time, GC is able to keep up with the object allocation rate.

Jackie-Jiang:
@sunithabeeram The clone is not a shallow copy, but a deep copy for DecimalFormat. So it won't fall into the same issue (the clone uses a separate DigitList).

PR #3441, opened 2018-11-07
Replace ENABLE_SEGMENT_LEVEL_VALIDATION config with SEGMENT_LEVEL_VALIDATION_INTERVALS

Jackie-Jiang:
Segment level validation is useful to track missing segments and total document count
Instead of disabling it, we run segment level validation with intervals
By default, run segment level validation once per 24 rounds of validation


pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -93,7 +95,7 @@
   private static final long DEFAULT_SEGMENT_UPLOAD_TIMEOUT_IN_MILLIS = 600_000L; // 10 minutes
   private static final int DEFAULT_REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS = 64;
   private static final boolean DEFAULT_ENABLE_STORAGE_QUOTA_CHECK = true;
-  private static final boolean DEFAULT_ENABLE_SEGMENT_LEVEL_VALIDATION = true;
+  private static final int DEFAULT_SEGMENT_LEVEL_VALIDATION_INTERVALS = 24;
```
sunithabeeram: Should we keep this time bound instead of count bound? If the validation runs more frequently, then we'll reach the count sooner, right? We could track last segment validation time and define a segment-validation-interval-hours to control the interval?

Jackie-Jiang: Done

PR #3442, opened 2018-11-07
Allow generating new star-tree using Hadoop job

Jackie-Jiang:
Added class StarTreeIndexConfig for the new star-tree index config
Added starTreeIndexConfigs field in TableConfig.IndexingConfig
SegmentGeneratorConfig can extract the new star-tree builder configs from the table config automatically


pinot-common/src/main/java/com/linkedin/pinot/common/config/IndexingConfig.java
```diff
@@ -72,6 +72,9 @@
   @ConfigKey("starTreeIndexSpec")
   private StarTreeIndexSpec _starTreeIndexSpec;
 
+  @ConfigKey("starTreeIndexConfigs")
```
kishoreg: whats the difference between startreeindexSpec and starTreeIndexConfigs. Do we need both?

Jackie-Jiang: starTreeIndexSpec is for old star-tree, while starTreeIndexConfig is for new star-tree. Once we fully migrate to the new tree, we can remove the starTreeIndexSpec.

PR #3443, opened 2018-11-08
Add API to generate batch id for segment upload if batch upload is within quota.

jackjlli:
This PR:
- Add API to generate batch id for segment upload if batch upload is within quota.
- Modify upload segment API to include batch info to ZK metadata.

Details of the new API:
POST `/tables/{tableName}/batch`
Body:
```
{
  "segments": [
    {
      "name": "string",
      "size": 0
    }
  ]
}
```
Response:
```
{
  "status": "Segment upload is within quota. Successfully generated batch id for table:  testTable",
  "batchId": "b2483e32-342b-44ad-b1bb-0a40c067176b-1",
  "currentSizeAcrossReplicas": 500,
  "storageQuotaAcrossReplicas": 6000
}
```


kishoreg:
Can we come up with the better api path.

jackjlli:
@kishoreg Changed the path to `/tables/{tableName}/batch`. 
Basically it mainly checks storage quota and should be put under the table size resource.

jackjlli:
This PR is no more needed. I'm closing this PR.

PR #3445, opened 2018-11-08
Add max length limit to string values

Jackie-Jiang:
Motivation:
Recently, we run into issues where bad data is getting indexed into Pinot.
The length of one single string value is more than 1M, and cause the
dictionary size to be huge (>2G) and also cause int overflow and then crash
the server. In most of the cases, super long string is not even queriable,
and trying to index them is usually caused by client code bug. So we decide
to add a configurable limit to the length of string columns. If the value
for the column goes over the limit, the value will be trimmed to be the
length limit.

Added "maxLength" into FieldSpec, so each field (column) can have different
length limit.
The default limit is 512 (number of characters but not decoded bytes length).

NOTE: for string with null character, the part after the first null character
will be truncated because null character is reserved as padding character

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/data/recordtransformer/SanitationTransformer.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/data/recordtransformer/SanitationTransformer.java
@@ -19,8 +19,7 @@ import com.linkedin.pinot.common.data.FieldSpec;
 import com.linkedin.pinot.common.data.Schema;
 import com.linkedin.pinot.common.utils.StringUtil;
 import com.linkedin.pinot.core.data.GenericRow;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
 import java.util.Map;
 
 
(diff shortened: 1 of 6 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/data/recordtransformer/SanitationTransformer.java
```diff
@@ -19,50 +19,52 @@
...
-  private final List<String> _stringColumns = new ArrayList<>();
+  private final Map<String, Integer> _stringColumnMaxLengthMap = new HashMap<>();
 
   public SanitationTransformer(Schema schema) {
     for (Map.Entry<String, FieldSpec> entry : schema.getFieldSpecMap().entrySet()) {
-      FieldSpec.DataType dataType = entry.getValue().getDataType();
-      if (dataType == FieldSpec.DataType.STRING) {
-        _stringColumns.add(entry.getKey());
+      FieldSpec fieldSpec = entry.getValue();
+      if (fieldSpec.getDataType() == FieldSpec.DataType.STRING) {
```
mcvsubbu: may be useful to add a comment here that for now we sanitize only string types

Jackie-Jiang: Added

PR #3446, opened 2018-11-08
Fixed the naming bug for merge segment admin command

snleee:
1. Fixed the bug for min/max timestamp initialization that caused the wrong segment name generation
2. Added the support for nested directory + multiple directories
3. Removed 'inputSegmentsTarred' flag

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/SegmentMergeCommand.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/SegmentMergeCommand.java
@@ -55,7 +55,7 @@ public class SegmentMergeCommand extends AbstractBaseAdminCommand implements Com
   private static final String DEFAULT_MERGE_TYPE = "CONCATENATE";
   private static final int DEFAULT_SEQUENCE_ID = 0;
 
-  @Option(name = "-inputSegments", required = true, metaVar = "<String>", usage = "Comma separated input segment paths that need to be merged")
+  @Option(name = "-inputPaths", required = true, metaVar = "<String>", usage = "Comma separated input segment files or directories that contains input segments to be merged")
   private String _inputSegmentPaths;
 
   @Option(name = "-outputPath", required = true, metaVar = "<String>", usage = "Output segment path. This should be different from working directory.")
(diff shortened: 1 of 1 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/SegmentMergeCommand.java
```diff
@@ -112,6 +112,23 @@ public boolean execute() throws Exception {
       workingDir = new File(_workingDirectory);
     }
 
+    // Check if the input path is directory or comma separated files
+    String[] paths = _inputSegmentPaths.split(INPUT_SEGMENT_SEPARATOR);
```
Jackie-Jiang: Should we support nested directory or passing multiple directories?

pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/SegmentMergeCommand.java
```diff
@@ -210,8 +227,8 @@ private String getDefaultSegmentName(TableConfig tableConfig, Schema schema, Lis
     String tableName = TableNameBuilder.extractRawTableName(tableConfig.getTableName());
 
     // Compute mix/max time from segment metadata
-    long minStartTime = Integer.MAX_VALUE;
-    long maxEndTime = Integer.MIN_VALUE;
+    long minStartTime =  Long.MAX_VALUE;
```
Jackie-Jiang: (OCD) Extra space

pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/SegmentMergeCommand.java
```diff
@@ -112,6 +112,23 @@ public boolean execute() throws Exception {
...
+    if (paths.length == 1) {
+      File inputDirectory = new File(paths[0].trim());
+      if (inputDirectory.isDirectory()) {
+        for (File path : inputDirectory.listFiles()) {
+          inputPaths.add(path.getAbsolutePath());
+        }
+      }
+    } else {
+      for (String path : paths) {
+        inputPaths.add(path.trim());
```
mcvsubbu: good to check that the path is not a directory here and throw appropriate error. Or, support specifying comma-separated directory and/or paths

pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/SegmentMergeCommand.java
snleee: Since I support untarred segment file, I think that we should not support nested directory. I will update the code to support multiple paths and/or multiple directories (1-level).

pinot-tools/src/main/java/com/linkedin/pinot/tools/segment/converter/SegmentMergeCommand.java
snleee: I changed this to support comma separated files/directories + nested directories. Also, I added to throw the exception if the given input path is not valid.

PR #3449, opened 2018-11-09
[TE] rootcause - inconsistent data warning

harleyjj:
Rootcause route displays warning if the current value from the anomaly function and the current value retrieved from the backend are different by more than 1 percent.


xiaohui-sun:
Good job @harleyjj ! This is very useful.

PR #3452, opened 2018-11-09
[TE] anomalies - backend label translation fix

apucher:


apucher:
Absolutely not happy with the state of affairs here. There are 3 different representations for anomaly labels used throughout backend and frontend. Fixing this isn't trivial unfortunately.

PR #3453, opened 2018-11-09
Fix the bug for transferring more than 2G bytes

Jackie-Jiang:
FileChannel.transferTo() can transfer at most 2G bytes
Added FileUtils.transferBytes() to handle transferring more than 2G bytes


Jackie-Jiang:
@mcvsubbu Sadly we cannot because we already imported the apache one

PR #3470, opened 2018-11-13
[PINOT-7328] Reduce lock contention in physical planning phase by reducing the total number of tasks

sunithabeeram:
Performance tests show that for requests that require little server work (docs/entries scanned <100k), the latency increases quickly with QPS while CPU is hardly used. In one example, with a query that took ~2ms to execute, beyond 600QPS the latency shot up to 200+ms while CPU utilization was <20%. 

Profiling/thread dumps showed that majority of the time was spent in taking tasks off the task queue in the physical planning phase - queries were essentially serialized on the LinkedBlockingQueue's lock.

This change bounds the total number of tasks that can be created/put on the queue, thus reducing contention and doing more work per task.

With this change in place, we were able to push the example query to 2000+ qps.

Testing done:
- Added unit tests to cover functionality.
- Perf tests as noted above.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
@@ -93,7 +93,7 @@ public class CombinePlanNode implements PlanNode {
             List<Operator> operators = new ArrayList<>();
             int start = index * opsPerThread;
             int limit = Math.min(opsPerThread, numPlanNodes - start);
-            for(int count = index * opsPerThread; count < start + limit; count = count + 1) {
+            for(int count = start; count < start + limit; count++) {
               operators.add(_planNodes.get(count).run());
             }
             return operators;
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import org.apache.zookeeper.Op;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import org.apache.zookeeper.Op;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
```diff
@@ -36,7 +36,7 @@
 public class CombinePlanNode implements PlanNode {
   private static final Logger LOGGER = LoggerFactory.getLogger(CombinePlanNode.class);
 
-  private static final int NUM_PLAN_NODES_THRESHOLD_FOR_PARALLEL_RUN = 10;
+  private static final int MAX_PLAN_TASKS = Math.min(10, (int) (Runtime.getRuntime().availableProcessors() * .5));
```
kishoreg: Is this max plan tasks per thread or max threads?  Looks like it’s used as both. 

Is the goal to have 
1 - fixed number of threads, variable number of segments or 
2 - fixed number of segments per thread, variable threads?
3 - hybrid where we prefer option 2 but limit number of threads

kishoreg: Ok so there are really two variables - max threads and number of segments per thread. The code is using same variable to represent both of them MAX_PLAN_TASKS which is why I got confused. If possible lets create two separate variables.

Also, another concern is for 11 segments, we will still have 10 threads where as with 10 segments, we will be using only 1 thread. Is my understanding right?

This is where the hybrid approach might work better where number of threads can vary anywhere from 1 to MAX_THREADS=10.

//psuedo code
MAX_THREADS = 10;
MAX_SEGMENT_PER_THREAD = 100; //not sure whats the right number for this.
numThreads = Math.min(10, numSegments/MAX_SEGMENT_PER_THREAD); //anywhere from 1 to 10
numSegmentPerThread = numSegments/numThreads;

In your code Outer loop will go from 0 to numThreads and inner loop in the callJob will go from i*numSegmentPerThread to (i +1 ) * numSegmentPerThread 

This will allow us to control the two variables independently.

sunithabeeram: Sorry - I misunderstood your earlier comment. I've updated the code to use the min number of threads possible.

pinot-core/src/main/java/com/linkedin/pinot/core/plan/CombinePlanNode.java
```diff
@@ -78,15 +79,21 @@ public Operator run() {
...
+      for (int i = 0; i < threads; i++) {
         final int index = i;
         futures[i] = _executorService.submit(new TraceCallable<List<Operator>>() {
           @Override
           public List<Operator> callJob() throws Exception {
             List<Operator> operators = new ArrayList<>();
-            for(int count = index; count < numPlanNodes; count = count + MAX_PLAN_TASKS) {
+            int start = index * opsPerThread;
+            int limit = Math.min(opsPerThread, numPlanNodes - start);
+            for(int count = index * opsPerThread; count < start + limit; count = count + 1) {
```
mcvsubbu: ocd?
can use 'start' instead of 'index * opsPerThread' ? 
also, count++, maybe?

pinot-core/src/test/java/com/linkedin/pinot/core/plan/CombinePlanNodeTest.java
```diff
@@ -36,7 +36,7 @@
   @Test
   public void testParallelExecution() {
     AtomicInteger count = new AtomicInteger(0);
-    int numPlans = 42;
+    int numPlans = 99;
```
mcvsubbu: may also be useful to pick multiple numPlans values and run the test in a loop for a few different ones.

PR #3472, opened 2018-11-13
Misc Cleanup: Removed unused member variable from FixedByteSingleValueMultiColWriter.

mayankshriv:
1. Removed the unsed member variable 'rows' from
FixedByteSingleValueMultiColWriter.

2. Fixed various method signatures to throw appropriate exception.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/FixedByteSingleValueMultiColWriter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/FixedByteSingleValueMultiColWriter.java
@@ -24,14 +24,10 @@ import java.nio.ByteOrder;
 
 public class FixedByteSingleValueMultiColWriter {
   private int[] columnOffsets;
-  private int rows;
   private PinotDataBuffer indexDataBuffer;
   private int rowSizeInBytes;
 
-  public FixedByteSingleValueMultiColWriter(File file, int rows, int cols,
-      int[] columnSizes)
-      throws IOException {
-    this.rows = rows;
+  public FixedByteSingleValueMultiColWriter(File file, int rows, int cols, int[] columnSizes) throws IOException {
     this.columnOffsets = new int[cols];
     rowSizeInBytes = 0;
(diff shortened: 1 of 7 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/FixedByteSingleValueMultiColWriter.java
```diff
@@ -24,14 +24,10 @@
...
   private int[] columnOffsets;
-  private int rows;
   private PinotDataBuffer indexDataBuffer;
   private int rowSizeInBytes;
 
-  public FixedByteSingleValueMultiColWriter(File file, int rows, int cols,
-      int[] columnSizes)
-      throws IOException {
-    this.rows = rows;
+  public FixedByteSingleValueMultiColWriter(File file, int rows, int cols, int[] columnSizes) throws IOException {
```
snleee: You may want to remove `int rows` here as well.

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/FixedByteSingleValueMultiColWriter.java
```diff
@@ -24,14 +24,10 @@
 
 public class FixedByteSingleValueMultiColWriter {
   private int[] columnOffsets;
-  private int rows;
   private PinotDataBuffer indexDataBuffer;
   private int rowSizeInBytes;
 
-  public FixedByteSingleValueMultiColWriter(File file, int rows, int cols,
```
snleee: You probably wanted to throw `FileNotFoundException` instead of `IOException`?

pinot-core/src/main/java/com/linkedin/pinot/core/io/writer/impl/FixedByteSingleValueMultiColWriter.java
mayankshriv: That cannot be removed, because it is used to compute the total size for mmaping in this constructor.

PR #3473, opened 2018-11-13
[TE] Ability to configure reference links like oncall run-book in the email alert

akshayrai:
Changes:
* Added ability to configure a map of reference links (per subscription group) which will be included in the email alert (legacy + new).
* Sample config (Detection Alert Config or Alert Config):
```
    "referenceLinks": {
        "Oncall Runbook": "<link>"
    }
```


kishoreg:
This is amazing. Yaml config gives us the ability to move fast. Before it gets too complicated, lets make sure we document all the configs in one place? Maybe have a schema for yaml format? We made this mistake in Pinot and would like to avoid that in ThirdEye.

aaronucsd:
> This is amazing. Yaml config gives us the ability to move fast. Before it gets too complicated, lets make sure we document all the configs in one place? Maybe have a schema for yaml format? We made this mistake in Pinot and would like to avoid that in ThirdEye.

Second the document for the yaml config. We need to show in UI as well so the user know what options exits.

PR #3475, opened 2018-11-14
Enhance controller periodic task and scheduler

Jackie-Jiang:
For ControllerPeriodicTask:
1. Random pick an initial delay of range 2-5 mins so each task does not start at the same time
2. Make leadership information private to the class
For PeriodicTaskScheduler:
1. Filter out tasks with non-positive interval, so we can easily disable any task by setting the interval to an non-positive value
2. Set the size of thread pool to be the same as number of tasks
3. Reduce the run-time of PeriodicTaskSchedulerTest

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
@@ -15,10 +15,10 @@
  */
 package com.linkedin.pinot.controller.helix.core.periodictask;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.linkedin.pinot.controller.helix.core.PinotHelixResourceManager;
 import com.linkedin.pinot.core.periodictask.BasePeriodicTask;
 import java.util.List;
+import java.util.Random;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -60,38 +68,29 @@ public void run() {
   }
 
   private void skipLeaderTask() {
-    if (getAmILeader()) {
+    if (_wasLeader) {
```
mcvsubbu: nit: I would rename it as _isLeader

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -15,58 +15,63 @@
...
     }
 
-    if (periodicTasks.size() > CORE_POOL_SIZE) {
-      LOGGER.warn("The number of tasks:{} is more than the default number of threads:{}.", periodicTasks.size(),
-          CORE_POOL_SIZE);
+    List<PeriodicTask> tasksWithValidInterval = new ArrayList<>();
+    for (PeriodicTask periodicTask : periodicTasks) {
+      if (periodicTask.getIntervalInSeconds() > 0) {
+        tasksWithValidInterval.add(periodicTask);
+      }
```
mcvsubbu: Would be good to add a log whether or not you are adding a periodic task here (and remove other logs of "Adding XXX task")

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -15,58 +15,63 @@
...
     }
   }
 
   public void stop() {
-    LOGGER.info("Stopping PeriodicTaskScheduler");
-    _executorService.shutdown();
+    if (_executorService != null) {
+      LOGGER.info("Stopping periodic task scheduler");
+      _executorService.shutdown();
+      _executorService = null;
```
mcvsubbu: Do you need to await termination of tasks?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
Jackie-Jiang: Done

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
Jackie-Jiang: Done.

pinot-core/src/main/java/com/linkedin/pinot/core/periodictask/PeriodicTaskScheduler.java
Jackie-Jiang: That could be very slow. Once we shut down the executor, no new task will be scheduled.

pinot-core/src/test/java/com/linkedin/pinot/core/periodictask/PeriodicTaskSchedulerTest.java
```diff
@@ -16,171 +16,70 @@
...
-      }
-    };
-    periodicTasks.add(task1);
+  public void testTaskWithInvalidInterval() throws Exception {
+    AtomicBoolean initCalled = new AtomicBoolean();
+    AtomicBoolean runCalled = new AtomicBoolean();
 
-    // Stagger 2 tasks by delaying the 2nd task half of the frequency.
-    PeriodicTask task2 = new BasePeriodicTask("Task2", runFrequencyInSeconds, runFrequencyInSeconds / 2) {
+    List<PeriodicTask> periodicTasks = Collections.singletonList(new BasePeriodicTask("TestTask", 0L, 0L) {
```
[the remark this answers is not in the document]
Jackie-Jiang: Done

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/ValidationManager.java
```diff
@@ -60,8 +60,7 @@
 
   public ValidationManager(ControllerConf config, PinotHelixResourceManager pinotHelixResourceManager,
       PinotLLCRealtimeSegmentManager llcRealtimeSegmentManager, ValidationMetrics validationMetrics) {
-    super("ValidationManager", config.getValidationControllerFrequencyInSeconds(),
-        config.getValidationControllerFrequencyInSeconds() / 2, pinotHelixResourceManager);
```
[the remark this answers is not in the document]
Jackie-Jiang: There is no guarantee that any two tasks are not running in parallel, so using random initial delay should be fine. Also, the segment level validation is with a different frequency.

PR #3476, opened 2018-11-14
Fix the bug where realtime time conversion is skipped when incoming and outgoing time name are the same

Jackie-Jiang:
1. Always perform time transformation for REALTIME segment consumption
2. Always not perform time transformation for OFFLINE segment generation
  - The reason to disable time trasformation for OFFLINE is that, with the same incoming and outgoing time name, there is no way to determine whether the time conversion already happens. OFFLINE segment might be pre-aggregated and the time might have already been converted. If we need time transformation for OFFLINE segment generation in the future, we can add an extra flag for that.


Jackie-Jiang:
Propose the solution #3484 and close this one

PR #3484, opened 2018-11-15
Fix the bug where time conversion is skipped when incoming and outgoing time column name are the same

Jackie-Jiang:
1. Refactor TimeConverter to based on single time granularity spec
2. Refactor TimeTransformer so it can automatically detect whether the conversion is needed (based on the value in the record)
Added tests for the new TimeConverter and TimeTransformer


pinot-core/src/main/java/com/linkedin/pinot/core/data/recordtransformer/TimeTransformer.java
```diff
@@ -29,37 +29,64 @@
...
     }
-    // Skip transformation if outgoing value already exist
-    // NOTE: outgoing value might already exist for OFFLINE data
-    if (record.getValue(_outgoingTimeColumn) == null) {
-      record.putField(_outgoingTimeColumn, _timeConverter.convert(record.getValue(_incomingTimeColumn)));
+
+    // Use the first record for sanity check and determine whether the conversion is needed
+    Object incomingTimeValue = record.getValue(_incomingTimeColumn);
+    if (_isFirstRecord) {
+      _isFirstRecord = false;
```
mcvsubbu: Would it be better to set this only if we are not throwing an exception?

Jackie-Jiang: Make sense when use just ignore the exception and keep transforming. Will make the change, thanks.

PR #3485, opened 2018-11-15
[TE] frontend harleyjj/manage remove WoW and change buildMetricDataUr…

harleyjj:
…l to take user inputs.

This addresses the following problems:

- Only 1 month (for daily/hourly) or 1 week (for 5 minutes) level data is shown in the graph, which doesn't honor user selected time range.
- The wow comparison is misleading and should be removed.
- Primary metric is not used and should be hidden.

It does NOT fix this problem, as the trailing zero is in the response data, regardless of date range:

- The last point is zero if no data.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/anomaly-graph/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/anomaly-graph/component.js
@@ -159,12 +159,11 @@ export default Component.extend({
     const legendText = this.get('legendText');
 
     const {
-      dotted = { text: 'expected', color: 'blue'},
       solid = { text: 'current', color: 'blue' }
     }  = legendText;
 
     chart.insert('div', '.chart').attr('class', 'anomaly-graph__legend').selectAll('span')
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-frontend/app/pods/components/anomaly-graph/component.js
```diff
@@ -388,7 +388,7 @@ export default Component.extend({
     const showGraphLegend = this.get('showGraphLegend');
     return {
       position: 'inset',
-      show: showGraphLegend
+      show: false
```
aaronucsd: We should have left `showGraphLegend` here. As it has changed to false. Also we do like the option for user decide to show the legend by specifying the showGraphLegend's value. Now we force it to not show with ` show: false`

aaronucsd: Also if you don't use `const showGraphLegend = this.get('showGraphLegend');` we should have removed it.

PR #3494, opened 2018-11-16
adding StarTreeIndexv1 I/O for hadoop

fredxfred:
This allows a StarTreeIndexSpecFile (for StarTreeIndexv1) to be passed as a parameter to the hadoop segment creation job. Haven't gotten a chance to fully determine what needs to be done to pass a spec file for StarTreeIndexv2 but I want to try that in the future too


Jackie-Jiang:
To enable either star-tree v1 or v2, all you need to do is putting the spec into IndexingConfig inside TableConfig, then we will automatically load it into the generator config.

fredxfred:
One of our use cases is that we would like to change which columns are star tree indexed in a live cluster (to see the impact on performance). We haven't been able to get the REST command to update table config to work without reloading segments (possibly there is a bug there?), which brings our cluster down for too long. Having the StarTreeIndexSpecFile read as a parameter to the hadoop job gives us flexibility regarding which columns are indexed and avoids having to update the table config

npawar:

PR #3499, opened 2018-11-16
Add table config backup tool

jfim:
Add a tool that allows saving table configurations to disk using the new
configuration format.

### DIFF
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackupTableConfigsCommand.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackupTableConfigsCommand.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ * Copyright (C) 2014-2018 LinkedIn Corp. (pinot-core@linkedin.com)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.linkedin.pinot.tools.admin.command;
 
 import com.google.common.base.Charsets;
--- a/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackupTableConfigsCommand.java
+++ b/pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackupTableConfigsCommand.java
@@ -1,0 +1,125 @@
+/**
+ * Copyright (C) 2014-2016 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackupTableConfigsCommand.java
```diff
@@ -0,0 +1,124 @@
...
+  @Option(name = "-help", required = false, help = true, aliases = { "-h", "--h", "--help" },
+      usage = "Print this message.")
+  private boolean _help = false;
+
+  @Override
+  public boolean execute() throws Exception {
+    // Create the output directory if it does not exist
+    File outputDirectory = new File(_outputDir);
+    if (!outputDirectory.exists()) {
+      outputDirectory.mkdirs();
```
mayankshriv: Log a message?

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackupTableConfigsCommand.java
```diff
@@ -0,0 +1,124 @@
...
+        tablesToFetch = new ArrayList<>(tableNameArray.length());
+        for (int i = 0; i < tableNameArray.length(); i++) {
+          tablesToFetch.add(tableNameArray.getString(i));
+        }
+      } else {
+        throw new RuntimeException("Failed to fetch the table list, got HTTP status " + statusCode + ": " + responseEntity);
+      }
+    }
+
+    System.out.println("Will fetch " + tablesToFetch.size() + " table(s)");
```
mayankshriv: Unsure which is better System.out.println vs Logger?

pinot-tools/src/main/java/com/linkedin/pinot/tools/admin/command/BackupTableConfigsCommand.java
```diff
@@ -0,0 +1,124 @@
...
+
+    boolean allTablesSuccessful = true;
+    for (String tableName : tablesToFetch) {
+      HttpResponse response = client.execute(new HttpGet(requestURLBuilder.forNewUpdateTableConfig(tableName)));
+      int statusCode = response.getStatusLine().getStatusCode();
+      String responseEntity = IOUtils.toString(response.getEntity().getContent());
+      if (statusCode / 100 == 2) {
+        try {
+          File outputFile = new File(outputDirectory, tableName + ".conf");
```
mayankshriv: May want to use static private final instead of ".conf"?

mcvsubbu:
I assume that the idea behind backing up is to re-apply the table configs in case something goes wrong. The table configs downloaded via API cannot be used for re-creating the same table, so you may want to consider splitting them into OFFLINE and REALTIME parts from the json object returned from the controller. 

Otherwise, lgtm

jfim:
@mcvsubbu This can also be used for migration. PR #3506 discusses how one can migrate from the old config format to the new config format by backing up the configs with this tool and merging the different configs across Pinot clusters.

mcvsubbu:
Nice, thanks

npawar:

PR #3501, opened 2018-11-16
Fix NPE while unparsing single key DSL output

jfim:
Fix NPE that can happen if unparsing of a single key DSL fails.


jfim:
This has been tested using 4,310 table config files used in production at Linkedin, which have been tested for equality between the old and new configuration formats (both before and after merging) to ensure that the code behaves correctly across a large number of varied use cases. For obvious reasons, I'm not checking those in.

The overall code is also exercised as part of #2937.

jfim:
This will get done once the config format is fully stable.

PR #3505, opened 2018-11-16
Handle invalid durations in the duration DSL

jfim:
Handle invalid durations in the duration DSL by removing invalid values
when loading (eg. "null DAYS" or "null null"). These values come from
invalid old-style configurations that got migrated to the new config
format.

This does not fully round-trip. For example, an old config with timeUnit
of DAYS and timeValue of null will lose the DAYS qualifier, although
this configuration was invalid (and ignored) in the first place.

### DIFF
--- a/pinot-common/src/main/java/com/linkedin/pinot/common/config/DurationDsl.java
+++ b/pinot-common/src/main/java/com/linkedin/pinot/common/config/DurationDsl.java
@@ -16,26 +16,39 @@
 package com.linkedin.pinot.common.config;
 
 import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
  * DSL for durations, which turns "5 days" into a duration.
  */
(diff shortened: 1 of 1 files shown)


pinot-common/src/main/java/com/linkedin/pinot/common/config/DurationDsl.java
```diff
@@ -28,14 +28,22 @@ public Duration parse(String text) {
...
-    final String unit = parts[1].toUpperCase();
-    final String unitCount = parts[0];
-    return new Duration(TimeUnit.valueOf(unit), Integer.parseInt(unitCount));
+    try {
+      String[] parts = text.split(" ");
+      final String unit = parts[1].toUpperCase();
+      final String unitCount = parts[0];
+      return new Duration(TimeUnit.valueOf(unit), Integer.parseInt(unitCount));
+    } catch (Exception e) {
+      return null;
```
jenniferdai: do you want to log something here before returning?

PR #3514, opened 2018-11-19
[TE] Percentage and absolute change rule filter

jihaozh:
Percentage and absolute change rule filter in the new pipeline.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/components/AbsoluteChangeRuleAnomalyFilter.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/components/AbsoluteChangeRuleAnomalyFilter.java
@@ -16,7 +16,6 @@
 
 package com.linkedin.thirdeye.detection.components;
 
-import com.google.common.base.Strings;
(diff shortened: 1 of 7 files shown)


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/components/AbsoluteChangeRuleAnomalyFilter.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return true;
+  }
+
+  @Override
+  public void init(AbsoluteChangeRuleAnomalyFilterSpec spec, InputDataFetcher dataFetcher) {
+    this.dataFetcher = dataFetcher;
+    this.pattern = Pattern.valueOf(spec.getPattern().toUpperCase());
+    // customize baseline offset
+    if (!Strings.isNullOrEmpty(spec.getOffset())) {
```
apucher: If you get a chance, use Apache StringUtils instead

jihaozh: thanks. changed to use string utils.

PR #3516, opened 2018-11-19
Deleting extra check for deletedDir during retention

jenniferdai:


pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/SegmentDeletionManager.java
```diff
@@ -203,13 +203,6 @@ protected void removeSegmentFromStore(String tableNameWithType, String segmentId
    */
   public void removeAgedDeletedSegments(int retentionInDays) {
```
jackjlli: nit: It'd be good to log the messages like `Start/Finished removing blablabla` here and at the end of the method.

jenniferdai: There's a log in the caller of the method. LOGGER.info("Removing aged (more than {} days) deleted segments for all tables", _deletedSegmentsRetentionInDays);

PR #3532, opened 2018-11-20
[TE] frontend - harleyjj/rootcause - fix data inconsistency warning t…

harleyjj:
…o not be triggered by date range adjustments

The data inconsistency warning:

-will not be triggered by changing the date range
-will display only the original values that triggered it (value does not change based on date range)
-will remain displayed when the user changes the date range, if the original values merit the warning

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
@@ -3,7 +3,8 @@ import {
   computed,
   setProperties,
   getProperties,
-  get
+  get,
+  set
 } from '@ember/object';
 import moment from 'moment';
 import {
@@ -67,10 +68,38 @@ export default Component.extend({
   options: Object.keys(ANOMALY_OPTIONS_MAPPING),
 
   /**
+   * Can be set by isWarning if data inconsistent
+   * @type {boolean}
+   */
+  warningValue: false,
+
+  /**
    * A mapping of the status and a more human readable version
    * @type {Object}
    */
   optionsMapping: ANOMALY_OPTIONS_MAPPING,
+
+  /**
+   * Locally cache original time range on init
+   *
+   */
+  init() {
+    this._super(...arguments);
+    set(this, 'anomalyRangeOld', get(this, 'anomalyRange'));
+  },
+
+  /**
+   * Checks if anomalyRange has been changed since init
+   * @type {boolean}
+   */
+  isRangeChanged: computed(
+    'anomalyRange',
+    function () {
+      const anomalyRange = get(this, 'anomalyRange');
+      const anomalyRangeOld = get(this, 'anomalyRangeOld');
+      return (anomalyRange != anomalyRangeOld);
+    }
+  ),
 
   /**
    * Urn of an anomaly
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -330,16 +359,27 @@ export default Component.extend({
...
-    if (newCurrent && oldCurrent){
-      const diffCurrent = Math.abs((newCurrent-oldCurrent)/newCurrent);
-      return (diffCurrent > 0.01);
+  isWarning: computed('anomalyInfo', 'isRangeChanged', function () {
+    if(!get(this, 'isRangeChanged')) {
+      let oldCurrent = parseFloat(get(this, 'current'));
+      const newCurrent = this._getAggregate('current');
+      if (newCurrent && oldCurrent){
+        const diffCurrent = Math.abs((newCurrent-oldCurrent)/newCurrent);
+        if (diffCurrent > 0.01) {
```
xiaohui-sun: This is 1% but your comment says 5%.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -330,16 +359,27 @@ export default Component.extend({
...
-  isWarning: computed('anomalyInfo', function () {
-    const anomalyInfo = get(this, 'anomalyInfo');
-    let oldCurrent = parseFloat(get(this, 'current'));
-    const newCurrent = this._getAggregate('current');
-
-    if (newCurrent && oldCurrent){
-      const diffCurrent = Math.abs((newCurrent-oldCurrent)/newCurrent);
-      return (diffCurrent > 0.01);
+  isWarning: computed('anomalyInfo', 'isRangeChanged', function () {
+    if(!get(this, 'isRangeChanged')) {
```
xiaohui-sun: Based on our discussion we will always show the alert regardless range change or not.

harleyjj: Yes, this date range checker is in place because 'newCurrent' comes from _getAggregate, which will be updated if the user changes the date range.  So we make sure that the date range can't trigger the warning (In one case, the anomaly doesn't have the warning, but when the user slides the date picker, it activates the warning - 'isRangeChanged' is checked here to prevent this behavior).

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -66,12 +67,40 @@ export default Component.extend({
...
+  /**
+   * Checks if anomalyRange has been changed since init
+   * @type {boolean}
+   */
+  isRangeChanged: computed(
+    'anomalyRange',
+    function () {
+      const anomalyRange = get(this, 'anomalyRange');
+      const anomalyRangeOld = get(this, 'anomalyRangeOld');
+      return (anomalyRange != anomalyRangeOld);
```
apucher: This isn't gonna work, unfortunately. You'll need to compare the **start** and **end** of the anomaly entity with the anomaly range in the rca context

apucher: more background - the user may change the investigation range and save a session. If the session is loaded later on, the anomalyRange will be whatever the user saved, rather than the anomaly's original value.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
harleyjj: @xiaohui-sun Thank you, I corrected it.

thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
harleyjj: @apucher thank you for the help - I updated the code as specified

PR #3535, opened 2018-11-21
Optimize all filter predicates by adding isAlwaysTrue()

Jackie-Jiang:
When filter predicate is always evaluated to true, we can optimize the filter operator by replacing it with MatchAllFilterOperator.
This is especially useful for range filters on time column which spans a very long time range.
Also added method isResultMatchingAll() in BaseFilterOperator for the optimization.


pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/AndFilterOperator.java
```diff
@@ -28,6 +28,7 @@
   private final List<BaseFilterOperator> _filterOperators;
 
   public AndFilterOperator(List<BaseFilterOperator> filterOperators) {
+    // NOTE: EmptyFilterOperator and MatchAllFilterOperator should not be passed into the AndFilterOperator
```
sunithabeeram: Is that really necessary? Why can't we handle these operators if they are passed in more gracefully?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BitmapBasedFilterOperator.java
```diff
@@ -42,6 +42,8 @@
 
   public BitmapBasedFilterOperator(PredicateEvaluator predicateEvaluator, DataSource dataSource, int startDocId,
       int endDocId) {
+    // NOTE: predicate that is always false or true should not be passed into the ScanBasedFilterOperator
```
sunithabeeram: Same concern as above.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/predicate/NotEqualsPredicateEvaluatorFactory.java
```diff
@@ -75,8 +75,12 @@ public static BaseRawValueBasedPredicateEvaluator newRawValueBasedEvaluator(NEqP
       _nonMatchingDictId = dictionary.indexOf(nEqPredicate.getNotEqualsValue());
       if (_nonMatchingDictId >= 0) {
         _nonMatchingDictIds = new int[]{_nonMatchingDictId};
+        if (dictionary.length() == 1) {
```
sunithabeeram: Shouldn't this be _nonMatchingIds == dictionary.length()?

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/AndFilterOperator.java
kishoreg: +1

Jackie-Jiang: We can handle them, but we should not pass them in per performance concern (actually we won't pass them in because re-order operators won't take these 2 operators). I just added this comment for clarification. Will add more details to the comment.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/BitmapBasedFilterOperator.java
Jackie-Jiang: We can handle it, but we should not pass them in for performance concern.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/predicate/NotEqualsPredicateEvaluatorFactory.java
Jackie-Jiang: You mean **_nonMatchingDictIds.length == dictionary.length()**? **_nonMatchingDictIds.length** is always 1 here.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/filter/AndFilterOperator.java
sunithabeeram: Additional comments/context is great - thanks for adding that; however, it might be better to programatically enforce it via constructor checks (throw IllegalArg/State Exc for example) so the contract is not broken when code gets refactored/changed. In other places in this PR you have used asserts. These are weak checks as they can be disabled at the JVM level; would be good to consider throwing exceptions instead.

PR #3548, opened 2018-11-26
[TE] frontend - harleyjj/rootcause - warning value uses aggregate mul…

harleyjj:
…tiplier and can be turned off if aggregates change without slider being adjusted

The warning uses the aggregate multiplier to ensure oldCurrent and newCurrent are being compared in the same timeframe.  Once the date picker has been adjusted, isWarning's return value won't be adjusted anymore.  If aggregates update without the date being adjusted, the warning can be turned on or off, as appropriate.  The displayed value only changes when the warningValue variable is changed.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
@@ -327,7 +327,6 @@ export default Component.extend({
           };
         }
       });
-
       return anomalyInfo;
     }
   ),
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/pods/components/rootcause-anomaly/component.js
```diff
@@ -356,12 +355,16 @@ export default Component.extend({
...
+      const oldCurrent = parseFloat(get(this, 'current'));
+      let newCurrent = this._getAggregate('current');
+      const aggregateMultiplier = parseFloat(get(this, 'aggregateMultiplier'));
+      if (newCurrent && oldCurrent && aggregateMultiplier){
+        newCurrent = newCurrent * aggregateMultiplier;
         const diffCurrent = Math.abs((newCurrent-oldCurrent)/newCurrent);
         if (diffCurrent > 0.01) {
           set(this, 'warningValue', true);
+        } else {
+          set(this, 'warningValue', false);
```
apucher: this really shouldn't have side-effects

PR #3549, opened 2018-11-26
Cleaning data dir so URI construction works as expected

jenniferdai:
* Currently, in retention manager (and I am sure other places in the future), we have switched to deep storage. As a result, we have to construct URIs out of path strings, and we do not expect for these paths to already end with slashes. So that users can still declare whatever data directory they desire, we will clean this data directory before using it.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
@@ -23,6 +23,7 @@ import com.linkedin.pinot.filesystem.PinotFSFactory;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/FileUploadPathProvider.java
```diff
@@ -134,4 +135,18 @@ public File getTmpUntarredPath() {
...
     return new File(_schemasTmpDirURI);
   }
+
+  /**
+   * Cleans dataDir by getting rid of unnecessary slashes at the end of dataDir. Needed so other URIs are constructed
+   * as expected.
+   * @param dataDir data directory of pinot segments
+   * @return cleaned data directory
+   */
+  private String cleanDataDir(String dataDir) {
```
mayankshriv: The name seems to suggest data directory's content would be cleaned/removed. Why not just use existing utility functions such as trim()/strip() to do the same, instead of writing a new one?

jenniferdai: strip is a replace all or a strip leading/trialing spaces, and trim is just leading/trailing as far as i know, but I only want to remove the last slash. Is there another method I don't know about?

fx19880617: Maybe StringUtils.removeEnd(...)?
[https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#removeEnd-java.lang.String-java.lang.String-]

jenniferdai: thank you!

jenniferdai:
Good point. I think I would prefer if Pinot breaks since dataDir is manually configured, and if they put extra slashes there, it is most likely a typo, so I will leave it as is for now.

PR #3556, opened 2018-11-27
Fixing AzurePinotFS Move API

jenniferdai:
* Previously, we called a rename in the move method in AzurePinotFS. PinotFS assumes that the move command takes in the srcUri and that the file gets moved to dstUri. However, rename doesn't behave that way. I have changed move to be a copy and delete instead.

### DIFF
--- a/pinot-azure-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
+++ b/pinot-azure-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
@@ -99,8 +99,7 @@ public class AzurePinotFS extends PinotFS {
     if (exists(dstUri) && !overwrite) {
       return false;
     }
-    //rename the file
-    return _adlStoreClient.rename(srcUri.getPath(), dstUri.getPath());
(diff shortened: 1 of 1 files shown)


pinot-azure-filesystem/src/main/java/com/linkedin/pinot/filesystem/AzurePinotFS.java
```diff
@@ -99,8 +99,7 @@ public boolean move(URI srcUri, URI dstUri, boolean overwrite) throws IOExceptio
     if (exists(dstUri) && !overwrite) {
       return false;
     }
-    //rename the file
-    return _adlStoreClient.rename(srcUri.getPath(), dstUri.getPath());
+    return copy(srcUri, dstUri) && delete(srcUri, true);
```
mayankshriv: Could you elaborate how the rename() api is different from copy() && delete()

s-sahay: `rename()` with `overwrite` param set to `true` should give the same functionality.

PR #3558, opened 2018-11-27
Fixes to doc

mcvsubbu:


docs/architecture.rst
```diff
@@ -102,5 +140,3 @@ configured for any set of columns. Inverted indices, while taking up more storag
 
 Specialized indexes like StartTree index is also supported.
 
-**TODO Add/Link startree doc here**
```
jenniferdai: do we want to keep the startree todo so that we can remember to add a startree doc?

PR #3564, opened 2018-11-28
[TE] frontend - harleyjj/create-alert - make minor improvements to YA…

harleyjj:
…ML alert creator

1) Toggle button displays what you would toggle too (i.e. clicking when it says YAML changes to YAML)
2) If you click 'cancel' on YAML, it clears the info you typed and toggles back to form
3) If you toggle to the form after you typed into YAML, info you typed is preserved when you toggle back


harleyjj:
Thank you @apucher !

PR #3566, opened 2018-11-29
Add metric gauge to track the number of llc simultaneous segment builds

jackjlli:
This PR adds a metric gauge to track the number of LLC simultaneous segment builds, which could help us debug what happened during the real-time segment completion.


pinot-common/src/main/java/com/linkedin/pinot/common/metrics/AbstractMetrics.java
```diff
@@ -372,6 +371,34 @@ public Long call() throws Exception {
     }
   }
 
+  /**
+   * Logs a value to a table gauge.
```
mcvsubbu: s/Logs/Adds/

jackjlli: Done.

PR #3567, opened 2018-11-29
[TE] rootcause - rich metric selector

apucher:
Improve visualization of metrics selector on root cause page. Displays metric names similar to metrics table, with dataset subscript and dimension values. Also enables quick-selection of metrics with pre-existing dimensions.

image


harleyjj:
Very nice css

PR #3568, opened 2018-11-29
Add guava cache to cache table schema in pinot broker

jackjlli:
This PR adds a guava cache to cache table schemas in broker so that non-existence column check can directly be done on broker side.

When the EV of broker resource got changed, the guava cache will make an async call to ZK once to fetch the schema for the table (referring to the code in `BrokerResourceOnlineOfflineStateModelFactory`).

The cached schema will be expired after certain amount of time. The reason of using **expire** instead of **refresh** is to reduce the number of times of ZK access. E.g. a single cluster is serving for thousands of tables. If we use `refreshAfterWrite`(lets say 30 mins), there'll be thousands of ZK read to update schema in the cache every 30 mins. The tables with less qps shouldn't be kept in the cache unless it's queried. And the ones with more qps should stay in the cache and keep updated.

When processing a query, if the cache contains the table schema, just simply returns the schema and compares against the columns appeared in the query.
Otherwise, the cache calls ZK **only once** to load the schema asynchronously. The following queries asking for the same new table will ignore the schema check until guava cache successfully gets the schema from ZK.

If there're columns in the query missing in the table schema, broker emits a broker metric and returns an exception back to the client.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -208,8 +208,9 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     }
 
     // Validate the request
+    FilterQueryTree[] filterQueryTree = new FilterQueryTree[1];
     try {
-      validateRequest(brokerRequest, requestParams);
+      validateRequest(brokerRequest, requestParams, filterQueryTree);
     } catch (Exception e) {
       LOGGER.info("Caught exception while validating request {}: {}, {}", requestId, query, e.getMessage());
       requestStatistics.setErrorCode(QueryException.QUERY_VALIDATION_ERROR_CODE);
@@ -239,18 +240,18 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     BrokerRequest realtimeBrokerRequest = null;
     if ((offlineTableName != null) && (realtimeTableName != null)) {
       // Hybrid
-      offlineBrokerRequest = _brokerRequestOptimizer.optimize(getOfflineBrokerRequest(brokerRequest), timeColumn);
-      realtimeBrokerRequest = _brokerRequestOptimizer.optimize(getRealtimeBrokerRequest(brokerRequest), timeColumn);
+      offlineBrokerRequest = _brokerRequestOptimizer.optimize(getOfflineBrokerRequest(brokerRequest), timeColumn, filterQueryTree[0]);
+      realtimeBrokerRequest = _brokerRequestOptimizer.optimize(getRealtimeBrokerRequest(brokerRequest), timeColumn, filterQueryTree[0]);
       requestStatistics.setFanoutType(RequestStatistics.FanoutType.HYBRID);
     } else if (offlineTableName != null) {
       // OFFLINE only
       brokerRequest.getQuerySource().setTableName(offlineTableName);
-      offlineBrokerRequest = _brokerRequestOptimizer.optimize(brokerRequest, timeColumn);
+      offlineBrokerRequest = _brokerRequestOptimizer.optimize(brokerRequest, timeColumn, filterQueryTree[0]);
       requestStatistics.setFanoutType(RequestStatistics.FanoutType.OFFLINE);
     } else {
       // REALTIME only
       brokerRequest.getQuerySource().setTableName(realtimeTableName);
-      realtimeBrokerRequest = _brokerRequestOptimizer.optimize(brokerRequest, timeColumn);
+      realtimeBrokerRequest = _brokerRequestOptimizer.optimize(brokerRequest, timeColumn, filterQueryTree[0]);
       requestStatistics.setFanoutType(RequestStatistics.FanoutType.REALTIME);
     }
 
@@ -329,7 +330,7 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
    *   <li>Value for 'LIMIT' for selection query <= configured value</li>
    * </ul>
    */
-  private void validateRequest(BrokerRequest brokerRequest, RequestParams requestParams) {
+  private void validateRequest(BrokerRequest brokerRequest, RequestParams requestParams, FilterQueryTree[] filterQueryTree) {
     if (brokerRequest.isSetAggregationsInfo()) {
       if (brokerRequest.isSetGroupBy()) {
         long topN = brokerRequest.getGroupBy().getTopN();
@@ -352,7 +353,7 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
       String tableName = brokerRequest.getQuerySource().getTableName();
       Schema schema = _tableSchemaCache.getIfTableSchemaPresent(tableName);
       if (schema != null) {
-        Set<String> columnsFromBrokerRequest = getAllColumnsFromBrokerRequest(brokerRequest);
+        Set<String> columnsFromBrokerRequest = getAllColumnsFromBrokerRequest(brokerRequest, filterQueryTree);
         // Filters out virtual columns in the query.
         columnsFromBrokerRequest.removeIf(column -> column.startsWith("$"));
         columnsFromBrokerRequest.removeAll(schema.getColumnNames());
@@ -372,12 +373,12 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
    * Helper to get all the columns from broker request.
    * Returns the set of all the columns.
    */
-  private Set<String> getAllColumnsFromBrokerRequest(BrokerRequest brokerRequest) {
+  private Set<String> getAllColumnsFromBrokerRequest(BrokerRequest brokerRequest, FilterQueryTree[] filterQueryTree) {
     Set<String> allColumns = new HashSet<>();
     // Filter
-    FilterQueryTree filterQueryTree = RequestUtils.generateFilterQueryTree(brokerRequest);
-    if (filterQueryTree != null) {
-      allColumns.addAll(RequestUtils.extractFilterColumns(filterQueryTree));
+    filterQueryTree[0] = RequestUtils.generateFilterQueryTree(brokerRequest);
+    if (filterQueryTree[0] != null) {
+      allColumns.addAll(RequestUtils.extractFilterColumns(filterQueryTree[0]));
     }
 
     // Aggregation
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -22,7 +22,6 @@ import com.fasterxml.jackson.databind.JsonNode;
 import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -45,21 +44,15 @@ import org.apache.pinot.common.exception.QueryException;
 import org.apache.pinot.common.metrics.BrokerMeter;
 import org.apache.pinot.common.metrics.BrokerMetrics;
 import org.apache.pinot.common.metrics.BrokerQueryPhase;
-import org.apache.pinot.common.request.AggregationInfo;
 import org.apache.pinot.common.request.BrokerRequest;
 import org.apache.pinot.common.request.FilterOperator;
 import org.apache.pinot.common.request.FilterQuery;
 import org.apache.pinot.common.request.FilterQueryMap;
-import org.apache.pinot.common.request.GroupBy;
-import org.apache.pinot.common.request.Selection;
-import org.apache.pinot.common.request.transform.TransformExpressionTree;
 import org.apache.pinot.common.response.BrokerResponse;
 import org.apache.pinot.common.response.broker.BrokerResponseNative;
 import org.apache.pinot.common.utils.CommonConstants;
-import org.apache.pinot.common.utils.request.FilterQueryTree;
+import org.apache.pinot.common.utils.request.RequestInfo;
 import org.apache.pinot.common.utils.request.RequestUtils;
-import org.apache.pinot.core.query.aggregation.function.AggregationFunctionType;
-import org.apache.pinot.core.query.aggregation.function.AggregationFunctionUtils;
 import org.apache.pinot.core.query.reduce.BrokerReduceService;
 import org.apache.pinot.pql.parsers.Pql2Compiler;
 import org.slf4j.Logger;
@@ -208,9 +201,9 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     }
 
     // Validate the request
-    FilterQueryTree[] filterQueryTree = new FilterQueryTree[1];
+    RequestInfo requestInfo;
     try {
-      validateRequest(brokerRequest, requestParams, filterQueryTree);
+      requestInfo = validateRequest(brokerRequest, requestParams);
     } catch (Exception e) {
       LOGGER.info("Caught exception while validating request {}: {}, {}", requestId, query, e.getMessage());
       requestStatistics.setErrorCode(QueryException.QUERY_VALIDATION_ERROR_CODE);
@@ -225,9 +218,6 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     }
     Map<String, String> debugOptions = requestParams.getDebugOptions();
     if (debugOptions != null) {
-//    if (request.has(DEBUG_OPTIONS)) {
-//      Map<String, String> debugOptions = Splitter.on(';').omitEmptyStrings().trimResults().withKeyValueSeparator('=')
-//          .split(request.get(DEBUG_OPTIONS).asText());
       LOGGER.debug("Debug options are set to: {} for request {}: {}", debugOptions, requestId, query);
       brokerRequest.setDebugOptions(debugOptions);
     }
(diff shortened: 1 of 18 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -311,6 +329,68 @@ private void validateRequest(BrokerRequest brokerRequest) {
...
+  }
+
+  /**
+   * Helper to get all the columns from broker request.
+   * Returns the set of all the columns.
+   */
+  private Set<String> getAllColumnsFromBrokerRequest(BrokerRequest brokerRequest) {
+    Set<String> allColumns = new HashSet<>();
+    // Filter
+    FilterQueryTree filterQueryTree = RequestUtils.generateFilterQueryTree(brokerRequest);
```
mayankshriv: The query tree is already being generated at few places in the read path. We should think about solving it in a way that the query is parsed and query tree generated exactly once per query.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -311,6 +329,68 @@ private void validateRequest(BrokerRequest brokerRequest) {
             "Value for 'LIMIT' (" + limit + ") exceeds maximum allowed value of " + _queryResponseLimit);
       }
     }
+
+    // Checks whether the query contains non-existence columns.
+    // Table name has already been verified before hitting this line.
+    String tableName = brokerRequest.getQuerySource().getTableName();
```
mayankshriv: These seem like expensive operations being added in the query path. At the very least, this feature should be off by default.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
```diff
@@ -0,0 +1,67 @@
...
+
+
+public class TableSchemaCache {
+  private static final long DEFAULT_CACHE_SIZE = 50;
+  private static final long DEFAULT_CACHE_TIMEOUT_IN_MINUTE = 60;
+
+  private final LoadingCache<String, Schema> _tableSchemaCache;
+  private final ZkHelixPropertyStore<ZNRecord> _propertyStore;
+
+  TableSchemaCache(ZkHelixPropertyStore<ZNRecord> propertyStore) {
```
mayankshriv: Will each broker cache schema for all tables in the cluster, or just the tables it serves? If former, we should change it to the latter.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
```diff
@@ -0,0 +1,67 @@
...
+    _propertyStore = propertyStore;
+    _tableSchemaCache = CacheBuilder.newBuilder()
+        .maximumSize(DEFAULT_CACHE_SIZE)
+        .expireAfterWrite(DEFAULT_CACHE_TIMEOUT_IN_MINUTE, TimeUnit.MINUTES)
```
Jackie-Jiang: refreshAfterWrite might be better

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
```diff
@@ -0,0 +1,67 @@
...
+          }
+        });
+
+  }
+
+  /**
+   * Refreshes table schema.
+   * @param tableName Table name with or without type suffix.
+   */
+  public void refreshTableSchema(String tableName) {
```
Jackie-Jiang: Please make sure this is done async.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -78,13 +93,16 @@
...
+      TableQueryQuotaManager tableQueryQuotaManager, ZkHelixPropertyStore<ZNRecord> propertyStore,
+      BrokerMetrics brokerMetrics) {
     _config = config;
     _routingTable = routingTable;
     _timeBoundaryService = timeBoundaryService;
     _accessControlFactory = accessControlFactory;
     _tableQueryQuotaManager = tableQueryQuotaManager;
+    _propertyStore = propertyStore;
     _brokerMetrics = brokerMetrics;
+    _tableSchemaCache = new TableSchemaCache(_propertyStore);
```
sunithabeeram: It might be worth it to keep the tableSchemaCache as a top level citizen that gets passed in (instead of property-store). Its possible classes outside of the RequestHandler might need access to it.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
sunithabeeram: +1 for conf control.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -311,6 +329,68 @@ private void validateRequest(BrokerRequest brokerRequest) {
...
+    }
+
+
+    // Selection
+    Selection selection = brokerRequest.getSelections();
+    if (selection != null) {
+      allColumns.addAll(RequestUtils.extractSelectionColumns(selection));
+    }
+
+    return allColumns;
```
sunithabeeram: It looks like this functionality is very similar to ServerQueryRequest.java - It might make sense to extract the common code into RequestUtils itself so it can be used by both the broker and server side implementations.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
```diff
@@ -0,0 +1,67 @@
...
+import com.linkedin.pinot.common.config.TableNameBuilder;
+import com.linkedin.pinot.common.data.Schema;
+import com.linkedin.pinot.common.metadata.ZKMetadataProvider;
+import java.util.concurrent.TimeUnit;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
+
+
+public class TableSchemaCache {
+  private static final long DEFAULT_CACHE_SIZE = 50;
```
sunithabeeram: We might size this to be bigger so as to accommodate bigger clusters, else we'll have to expose a config option to do the same. 
It might be worth exploring 'softValues' option in the guava cache so the values can be GC'ed under memory pressure, but otherwise we can keep as many schemas/tables as are accessed by the broker - this can help remove yet another tunable for the cache size.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
sunithabeeram: +1; the behavior explained here is desirable: https://github.com/google/guava/wiki/CachesExplained#refresh

pinot-common/src/main/java/com/linkedin/pinot/common/metrics/BrokerMeter.java
```diff
@@ -94,7 +94,9 @@
   // Netty connection metrics
   NETTY_CONNECTION_REQUESTS_SENT("nettyConnection", true),
   NETTY_CONNECTION_BYTES_SENT("nettyConnection", true),
-  NETTY_CONNECTION_BYTES_RECEIVED("nettyConnection", true);
+  NETTY_CONNECTION_BYTES_RECEIVED("nettyConnection", true),
+
+  QUERY_NON_EXISTENCE_COLUMNS("queries", false);
```
sunithabeeram: EXISTENCE => EXISTENT

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
jackjlli: It is the latter.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
jackjlli: +1 on moving it to a config option.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: Done

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
jackjlli: Changed to `refreshAfterWrite`

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/TableSchemaCache.java
jackjlli: Thanks for reminding. 👍

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: Done

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: A extra field called `validateQuery` in query has been added to this PR. Only when it's specified and it's set to true do we validate the schema.

pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -120,6 +122,11 @@ public HelixBrokerStarter(String brokerHost, String helixClusterName, String zkS
     _helixExternalViewBasedRouting = new HelixExternalViewBasedRouting(_propertyStore, _spectatorHelixManager,
         pinotHelixProperties.subset(ROUTING_TABLE_PARAMS_SUBSET_KEY));
     _tableQueryQuotaManager = new TableQueryQuotaManager(_spectatorHelixManager);
+    _tableSchemaCache = new TableSchemaCache(_propertyStore,
+        _pinotHelixProperties.getInt(CommonConstants.Broker.CONFIG_OF_TABLE_SCHEMA_MAXIMUM_CACHE_SIZE,
```
kishoreg: none of these configs are useful. If we go beyond cache size, we will be hitting ZK for every call

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -181,7 +195,7 @@ public BrokerResponse handleRequest(JSONObject request, @Nullable RequesterIdent
 
     // Validate the request
     try {
-      validateRequest(brokerRequest);
+      validateRequest(brokerRequest, request);
```
kishoreg: can we parse the json request into a concrete class e.g. RequestParams. This will avoid having constant strings everywhere.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
kishoreg: +1

jfim:
Can you explain why does the broker needs to cache the schema in the first place?

jackjlli:
@jfim The purpose of keeping schemas in broker is to reduce the call to pinot servers if any columns in the query don't exist in table schema. Currently pinot brokers blindly send out the query to all the selected servers and get empty response without any extra info.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: I've added a new method called `preComputeRequestInfo` which extracts this part of logic, so that both broker and server can reuse this logic.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: Talked with @mayankshriv offline. It's better we parse the query only once end-to-end, i.e from broker to server. While this might need to have another design on changing the payload, which would be out of the scope of this PR. In the current PR we guarantee the query is parsed only once in pinot-broker. We can take this payload change in the near future.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -207,13 +213,15 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
...
+    if (requestParams.isTrace()) {
       LOGGER.debug("Enable trace for request {}: {}", requestId, query);
       brokerRequest.setEnableTrace(true);
     }
-    if (request.has(DEBUG_OPTIONS)) {
-      Map<String, String> debugOptions = Splitter.on(';').omitEmptyStrings().trimResults().withKeyValueSeparator('=')
-          .split(request.get(DEBUG_OPTIONS).asText());
+    Map<String, String> debugOptions = requestParams.getDebugOptions();
+    if (debugOptions != null) {
+//    if (request.has(DEBUG_OPTIONS)) {
```
sunithabeeram: Please remove commented out code if its not needed.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -332,6 +340,31 @@ private void validateRequest(BrokerRequest brokerRequest) {
...
+
+    // Checks whether the query contains non-existent columns.
+    // Table name has already been verified before hitting this line.
+    if (requestParams.isValidateQuery()) {
+      String tableName = brokerRequest.getQuerySource().getTableName();
+      Schema schema = _tableSchemaCache.getIfTableSchemaPresent(tableName);
+      if (schema != null) {
+        RequestInfo requestInfo = RequestUtils.preComputeRequestInfo(brokerRequest);
+        // gets all the columns from broker request
+        Set<String> columnsFromBrokerRequest = requestInfo.getAllColumns();
```
sunithabeeram: Consider adding getPhysicalColumns so the virtual column filtering can be done internally.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -332,6 +340,31 @@ private void validateRequest(BrokerRequest brokerRequest) {
...
+        // filters out virtual columns in the query.
+        columnsFromBrokerRequest.removeIf(column -> column.startsWith("$"));
+        columnsFromBrokerRequest.removeAll(schema.getColumnNames());
+        if (!columnsFromBrokerRequest.isEmpty()) {
+          _brokerMetrics.addMeteredTableValue(tableName, BrokerMeter.QUERY_NON_EXISTENT_COLUMNS, 1L);
+          throw new RuntimeException(
```
sunithabeeram: Can we throw a more specific exception so that we can throw 400 bad request instead of a 500 internal server error?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -197,8 +202,9 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
     }
 
     // Validate the request
+    FilterQueryTree[] filterQueryTree = new FilterQueryTree[1];
```
sunithabeeram: Why does this need to be an array of 1?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/RequestParams.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public Map<String, String> getDebugOptions() {
+    return _debugOptions;
+  }
+
+  public boolean isTrace() {
+    return _trace;
+  }
+
+  public boolean isValidateQuery() {
```
sunithabeeram: isValidate => shouldValidate

pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
```diff
@@ -118,10 +118,15 @@ public ServerType getServerType() {
     public static final String CONFIG_OF_BROKER_REFRESH_TIMEBOUNDARY_INFO_SLEEP_INTERVAL =
             "pinot.broker.refresh.timeBoundaryInfo.sleepInterval";
     public static final long DEFAULT_BROKER_REFRESH_TIMEBOUNDARY_INFO_SLEEP_INTERVAL_MS = 10000L;
+    public static final String CONFIG_OF_TABLE_SCHEMA_MAXIMUM_CACHE_SIZE = "pinot.broker.table.schema.cache.size";
```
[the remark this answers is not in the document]
sunithabeeram: @kishoreg I don't believe that is correct. We want to maintain a LRU cache - so if we have a lot of tables, the frequently queried ones will stay in memory and we'll hit ZK only for the ones not in memory. We should limit the in memory size to guard against pathological cases where the size grows unbounded.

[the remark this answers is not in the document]
kishoreg: I haven't looked at the PR in detail. The most important thing is we should never be hitting ZK in the request path. This violates that scenario. I prefer the broker going failing to serve a table if the number of tables is a lot vs going to ZK.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: We need an instance of FilterQueryTree to be returned if it's been calculated. And simply returning this instance out of the method `validateRequest` looks unreasonable. Thus, I use an array to return the instance. Another way to handle that is to add a private variable of `FilterQueryTree` in requestParams.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: This runtime exception will be captured in caller's exception block and thus throws a `QUERY_VALIDATION_EXCEPTIONS` back to the client. So we're good here.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/RequestParams.java
jackjlli: Done.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: Updated the PR by returning request info.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
jackjlli: Done.

pinot-common/src/main/java/com/linkedin/pinot/common/utils/CommonConstants.java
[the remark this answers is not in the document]
jackjlli: The design of using this guava cache is, if the schema doesn't exist, the cache makes an async call to fetch the schema. This async call will be done **only once** even if there's a burst of queries coming in (e.g N queries). The following queries (from 2nd to Nth) containing the new schema will ignore the query check until guava cache successfully gets the schema from ZK. So in this case, we guarantee the number of times to hit ZK is just once. There won't be N ZK reads when N queries received.

npawar:

PR #3575, opened 2018-12-01
[TE] frontend - harleyjj/rootcause - use humanizeFloat to change form…

harleyjj:
…atting of y-axis and prevent apparent duplicates


akshayrai:
Can we also do this for the graph in the anomaly page? Only if it is simple change like this.

harleyjj:
I'll look into it now

PR #3579, opened 2018-12-03
[TE] frontend - default redirect to home dashboard

apucher:


apucher:
@aaronucsd you are not worthy ... get your apache uid fixed

PR #3582, opened 2018-12-04
[TE] frontend - harleyjj/manage - align startDate at startOf() instea…

harleyjj:
…d of endOf()

This accomplishes two things:
1) Starting timestamp is now aligned at 00:00, which should invoke better response from endpoint
2) Defaults date picker time to 00:00

The trailing zero issue on the graph is resolved in some cases, but not all


xiaohui-sun:
Have you merged my change and tested?

PR #3583, opened 2018-12-04
Add segment selector for merged segments

snleee:
Once the merge lineage information is written to the property store,
broker will read the lineage and pick only segments that are required
for answering the query. This PR adds the segment selector for merged
segments.
1. Add segment selector interface
2. Add merged segment selector implementation
3. Add unit tests

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/BaseRoutingTableBuilder.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/BaseRoutingTableBuilder.java
@@ -16,6 +16,7 @@
 package com.linkedin.pinot.broker.routing.builder;
 
 import com.linkedin.pinot.broker.routing.RoutingTableLookupRequest;
+import com.linkedin.pinot.broker.routing.selector.SegmentSelector;
 import com.linkedin.pinot.common.config.RoutingConfig;
 import com.linkedin.pinot.common.config.TableConfig;
 import com.linkedin.pinot.common.metrics.BrokerMeter;
@@ -123,16 +124,16 @@ public abstract class BaseRoutingTableBuilder implements RoutingTableBuilder {
     }
   }
 
-  @Override
-  public Map<String, List<String>> getRoutingTable(RoutingTableLookupRequest request) {
+  public Map<String, List<String>> getRoutingTable(RoutingTableLookupRequest request, SegmentSelector segmentSelector) {
     if (_enableDynamicComputing) {
       // Copy the pointer for snapshot since the pointer for segment to servers map can change at anytime
       Map<String, List<String>> segmentToServersMap = _segmentToServersMap;
 
-      // Get all existing segments
+      // Selecting segments only required for processing a query
       Set<String> segmentsToQuery = segmentToServersMap.keySet();
-
-      // TODO: add the selection logic here
+      if (segmentSelector != null) {
+        segmentsToQuery = segmentSelector.selectSegments(request, segmentsToQuery);
+      }
 
       // Compute the final routing table
       return computeDynamicRoutingTable(segmentToServersMap, segmentsToQuery);
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/DefaultRealtimeRoutingTableBuilder.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/DefaultRealtimeRoutingTableBuilder.java
@@ -16,6 +16,7 @@
 package com.linkedin.pinot.broker.routing.builder;
 
 import com.linkedin.pinot.broker.routing.RoutingTableLookupRequest;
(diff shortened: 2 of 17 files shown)


snleee:
#2715

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/BaseRoutingTableBuilder.java
```diff
@@ -123,16 +124,16 @@ public void computeOnExternalViewChange(String tableName, ExternalView externalV
...
       // Copy the pointer for snapshot since the pointer for segment to servers map can change at anytime
       Map<String, List<String>> segmentToServersMap = _segmentToServersMap;
 
-      // Get all existing segments
+      // Selecting segments only required for processing a query
       Set<String> segmentsToQuery = segmentToServersMap.keySet();
-
-      // TODO: add the selection logic here
+      if (segmentSelector != null) {
+        segmentsToQuery = segmentSelector.selectSegments(request, segmentsToQuery);
```
mayankshriv: Does it make sense to merge the concepts of segment selector and pruner?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/DefaultRealtimeRoutingTableBuilder.java
```diff
@@ -68,7 +69,7 @@ public void computeOnExternalViewChange(String tableName, ExternalView externalV
   }
 
   @Override
-  public Map<String, List<String>> getRoutingTable(RoutingTableLookupRequest request) {
+  public Map<String, List<String>> getRoutingTable(RoutingTableLookupRequest request, SegmentSelector segmentSelector) {
```
mayankshriv: Should selector be passed into this api, or should it belong inside of the builder(s)?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/selector/SegmentSelectorFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.linkedin.pinot.common.config.TableConfig;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
+
+
+/**
+ * Factory for segment selector
+ */
+public class SegmentSelectorFactory {
```
mayankshriv: This does not adhere to factory pattern, may be better to name it as SegmentSelectorProvider?

pinot-broker/src/test/java/com/linkedin/pinot/broker/routing/selector/MergedSegmentSelectorTest.java
```diff
@@ -0,0 +1,102 @@
...
+  @Test
+  public void testMergedSegmentSelector() throws Exception {
+    MergedSegmentSelector segmentSelector = new MergedSegmentSelector();
+
+    TableConfig tableConfig =
+        new TableConfig.Builder(CommonConstants.Helix.TableType.OFFLINE).setTableName(TEST_TABLE_NAME).build();
+    ZkHelixPropertyStore<ZNRecord> fakePropertyStore = new FakePropertyStore();
+    segmentSelector.init(tableConfig, fakePropertyStore);
+
+    // Add G0 (segment0), G1 (segment1), G2 (segment2) that represent 3 base segments
```
mayankshriv: Could we programatically generate these trees, to improve coverage?

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/BaseRoutingTableBuilder.java
snleee: for now, I would want to keep them separate. Once the merge feature is done, I can work on cleaning up the pruner (currently, it's closely tied with `PartitionAwareRoutingTableBuilder` and we need to decouple them)

pinot-broker/src/main/java/com/linkedin/pinot/broker/routing/builder/DefaultRealtimeRoutingTableBuilder.java
snleee: As discussed offline, we concluded that broker routing builder requires the re-structuring on class hierarchy. I will add the comments about it and will address later with the separate clean-up pr.

pinot-broker/src/test/java/com/linkedin/pinot/broker/routing/selector/MergedSegmentSelectorTest.java
snleee: If we generate the tree programatically, i still won't have the result to test against. I can probably come up with another implementation of the same logic directly from hashmap representation (not using tree). I will add this as a separate pr for improving testability of this segment selection logic. For now, I expanded the testing case to cover more edge cases with a fixed example.

PR #3584, opened 2018-12-04
add kafka simple consumer buffer and timeout to stream config

jamesyfshao:
When the kafka broker cluster has max message size config bigger than our current hard-coded value of low-level consumer buffer size (512K), pinot kafka consumer could encounter messages from kafka that are bigger than our buffer size. When this happens, kafka low-level consumer will fail silently and returns an empty message result list to low-level message ingester. From pinot users' point of view, pinot kafka low-level consumer stuck at the large message offset. This change aims to allow users to define their kafka ingestor buffer size so it can handle the failure scenarios above.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaLowLevelStreamConfig.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaLowLevelStreamConfig.java
@@ -49,9 +49,9 @@ public class KafkaLowLevelStreamConfig {
     String llcTimeoutKey =
         KafkaStreamConfigProperties.constructStreamProperty(KafkaStreamConfigProperties.LowLevelConsumer.KAFKA_SOCKET_TIMEOUT);
     _bootstrapHosts = streamConfigMap.get(llcBrokerListKey);
-    _kafkaBufferSize = getConfigWithDefault(streamConfigMap, llcBufferKey,
+    _kafkaBufferSize = getIntConfigWithDefault(streamConfigMap, llcBufferKey,
         KafkaStreamConfigProperties.LowLevelConsumer.KAFKA_BUFFER_SIZE_DEFAULT);
-    _kafkaSocketTimeout = getConfigWithDefault(streamConfigMap, llcTimeoutKey,
+    _kafkaSocketTimeout = getIntConfigWithDefault(streamConfigMap, llcTimeoutKey,
         KafkaStreamConfigProperties.LowLevelConsumer.KAFKA_SOCKET_TIMEOUT_DEFAULT);
     Preconditions.checkNotNull(_bootstrapHosts,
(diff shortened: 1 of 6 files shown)


pinot-core/src/test/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaLowLevelStreamConfigTest.java
```diff
@@ -0,0 +1,89 @@
+package com.linkedin.pinot.core.realtime.impl.kafka;
```
snleee: License header is required at the top of all the files

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaLowLevelStreamConfig.java
```diff
@@ -53,12 +65,36 @@ public String getBootstrapHosts() {
...
 
+  public int getKafkaBufferSize() {
+    return _kafkaBufferSize;
+  }
+
+  public int getKafkaSocketTimeout() {
+    return _kafkaSocketTimeout;
+  }
+
+  private int getConfigWithDefault(Map<String, String> configMap, String key, int defaultValue) {
```
snleee: I think that it's better to indicate `int` to the function name since this is specific to getting an int value. e.g. `getIntConfigWithDefault`

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaPartitionLevelConsumer.java
```diff
@@ -45,6 +46,11 @@ public KafkaPartitionLevelConsumer(String clientId, StreamConfig streamConfig, i
     super(clientId, streamConfig, partition, kafkaSimpleConsumerFactory);
   }
 
+  @VisibleForTesting
```
snleee: I think that this function should locate in the base class that defines `simpleConsumer`. Can we move this to `KafkaConnectionHandler`?

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaLowLevelStreamConfig.java
jamesyfshao: updated in new diff

pinot-core/src/main/java/com/linkedin/pinot/core/realtime/impl/kafka/KafkaPartitionLevelConsumer.java
jamesyfshao: updated in new diff

npawar:
LGTM

snleee:
@npawar Thanks for the confirmation. I'm merging the pr.

PR #3593, opened 2018-12-08
Allow completing segments to finish before stopping LLC Segment Manager

mcvsubbu:
Added a stop() method to LLC segment manager, so that we can allow the segments
that are completing to be done, and not allow new segments to start their
completion process while controller is shutting down

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerStarter.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerStarter.java
@@ -287,6 +287,10 @@ public class ControllerStarter {
 
   public void stop() {
     try {
+      // Stop PinotLLCSegmentManager before stopping Jersey API. It is possible that stopping Jersey API
+      // may interrupt the handlers waiting on an I/O.
+      PinotLLCRealtimeSegmentManager.getInstance().stop();
+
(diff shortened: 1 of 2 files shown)


mcvsubbu:
Addresses issue #3594

mcvsubbu:
Occurred to me that if we start ValidationManager soon after a leadership gain, we don't need to do this PR?

sunithabeeram:
Looks like @npawar also attempted to fix this in https://github.com/apache/incubator-pinot/pull/3591 :)

> Occurred to me that if we start ValidationManager soon after a leadership gain, we don't need to do this PR?

I guess if the previous leader is not "dead" it could continue running validation-mgr - unless there is a periodic isLeader() check that lets the old controller give up.

mcvsubbu:
> Looks like @npawar also attempted to fix this in #3591 :)
> 
> > Occurred to me that if we start ValidationManager soon after a leadership gain, we don't need to do this PR?
> 
> I guess if the previous leader is not "dead" it could continue running validation-mgr - unless there is a periodic isLeader() check that lets the old controller give up.

Yes, @npawar and I talked about it. We need to make sure that we check the leadership before every new table we process. If we are not leader, stop the validationmanager right away.

PR #3599, opened 2018-12-10
[TE] frontend - update default yaml

jihaozh:
update the default yaml in the yaml editor


harleyjj:
This is very helpful!

PR #3602, opened 2018-12-11
Cleanup: update code to return only physical columns for select * queries

sunithabeeram:
Addressing pending comments from https://github.com/apache/incubator-pinot/pull/3229 as I am looking to fix https://github.com/apache/incubator-pinot/issues/3601.

Changes in this PR include:
- Adding isVirtualColumn to ColumnMetadata; this is not currently used, but included for completeness.
- Updated code to use getPhysicalColumnNames instead of getColumnNames where appropriate.

Testing done:
- Added integration tests to cover the select * queries and assertions to ensure that virtual columns are not included.


pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/ColumnMetadata.java
```diff
@@ -81,6 +82,7 @@
   private final String dateTimeFormat;
   private final String dateTimeGranularity;
 
+
```
jfim: Avoid having extra whitespace to classes. Ping me if you need the link for the code format standard, so that whitespace/names are consistently applied by your IDE.

PR #3603, opened 2018-12-11
[TE] Endpoints for create and edit alert yaml along with validators

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/detection/validators/DetectionAlertConfigValidator.java
```diff
@@ -0,0 +1,109 @@
...
+   * Perform validation on the alert config like verifying if all the required fields are set
+   */
+  @SuppressWarnings("unchecked")
+  public boolean validateConfig(DetectionAlertConfigDTO alertConfig,  Map<String, String> responseMessage) {
+    boolean isValid = true;
+
+    // Check for all the required fields in the alert
+    if (StringUtils.isEmpty(alertConfig.getName())) {
+      responseMessage.put("message", "Subscription group name field cannot be left empty.");
+      return false;
```
xiaohui-sun: Can we do all the check and return the message together instead of early return here?

thirdeye/thirdeye-pinot/src/test/resources/com/linkedin/thirdeye/detection/yaml/alertconfig/alert-config-2.yaml
```diff
@@ -0,0 +1,42 @@
...
+type: DIMENSIONAL_ALERTER_PIPELINE
+dimensionRecipients:
+ "android":
+  - "android-thirdeye@thirdeye.com"
+ "ios":
+  - "ios-thirdeye@thirdeye.com"
+dimension: app_name
+
+detectionConfigIds:
+ - 5773069
```
xiaohui-sun: Is there a unique name to identify the detection config? Using the name would be more user friendly instead of id. We can do the conversion in backend.

PR #3604, opened 2018-12-11
Add ControllerLeadershipManager as single place to check controller leadership changes

npawar:
Step 1 of https://github.com/apache/incubator-pinot/issues/3600


pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerLeadershipManager.java
```diff
@@ -0,0 +1,116 @@
...
+  private ControllerLeadershipManager(HelixManager helixManager) {
+    _helixManager = helixManager;
+    _helixManager.addControllerListener((ControllerChangeListener) notificationContext -> onControllerChange());
+  }
+
+  /**
+   * Create an instance of ControllerLeadershipManager
+   * @param helixManager
+   */
+  public static synchronized void createInstance(HelixManager helixManager) {
```
sunithabeeram: It might be simpler to expose just getInstance(); if its not created, the instance will get created. Does the user code really need to distinguish between create and get?

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerLeadershipManager.java
```diff
@@ -0,0 +1,116 @@
...
+      throw new RuntimeException("Instance of ControllerLeadershipManager already created");
+    }
+    INSTANCE = new ControllerLeadershipManager(helixManager);
+  }
+
+  /**
+   * Get the instance of ControllerLeadershipManager
+   * @return
+   */
+  public static ControllerLeadershipManager getInstance() {
```
sunithabeeram: This needs to be synchronized as well, else it might suffer improper publication.

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerLeadershipManager.java
mcvsubbu: I thought about that as well. I think the option is between create() and init(HelixManager). Perhaps the latter is a bit less confusing.

npawar: We need 2 calls, else everyone who wants to get an instance will have to provide helix manager

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerStarter.java
```diff
@@ -161,13 +161,15 @@ public void start() {
     _helixResourceManager.start();
     final HelixManager helixManager = _helixResourceManager.getHelixZkManager();
 
+    LOGGER.info("Creating ControllerLeadershipManager instance");
+    ControllerLeadershipManager.createInstance(helixManager);
```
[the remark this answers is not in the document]
npawar: I put it at this place because helixManager is initialized in pinotHelixResourceManager.start(). If we were to move it before the start, we would have to extract out the helixManager setup part

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerLeadershipManager.java
mcvsubbu: I think init() and getInstance() is less confusing and less prone to errors than create() and getInstance().

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotRealtimeSegmentManager.java
```diff
@@ -101,12 +101,7 @@ public void start(ControllerMetrics controllerMetrics) {
     _zkClient.subscribeDataChanges(_tableConfigPath, this);
 
     // Subscribe to leadership changes
-    _pinotHelixResourceManager.getHelixZkManager().addControllerListener(new ControllerChangeListener() {
-      @Override
-      public void onControllerChange(NotificationContext changeContext) {
-        processPropertyStoreChange(CONTROLLER_LEADER_CHANGE);
-      }
-    });
+    ControllerLeadershipManager.getInstance().subscribe(PinotLLCRealtimeSegmentManager.class.getName(), this);
```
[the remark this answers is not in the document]
mcvsubbu: Good point. All registrations should happen before we listen to becoming leader

mcvsubbu:
Eventually, we do plan to distribute the load across controllers. Right now, each periodic task relies on running on  the controler  when it becomes leader (and stopping when it loses leadership).

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/realtime/PinotRealtimeSegmentManager.java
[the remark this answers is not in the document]
kishoreg: Another way to handle this is to invoke the callback when its subscribed. This will enforce that the callbacks are idempotent.

npawar:
Agree that the subscribers should not really know "leader" or "not leader", eventually, when we get to a distributed load model. As of now, the periodic tasks do check for leader before executing, and this is one way to get them to start/stop on leadership changes.
Even in the distributed model, the periodic tasks would have to start/stop and switch between the controller responsible for it.

PR #3606, opened 2018-12-11
[PINOT-7476] Add metrics to track cases where segment refresh/reloads fail

sunithabeeram:
Given the async nature of reload/refresh (a fire and forget from the controller), there is little feedback/tracking when servers encounter failures during reload/refresh. Multi segment reload can fail after some segments are reloaded which might lead to consistency issues. 

This PR adds metrics to help track when such failures do occur.

### DIFF
--- a/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
+++ b/pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.server.starter.helix;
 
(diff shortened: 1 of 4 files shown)


pinot-server/src/main/java/com/linkedin/pinot/server/starter/helix/SegmentMessageHandlerFactory.java
```diff
@@ -146,6 +160,8 @@ public HelixTaskResult handleMessage() throws InterruptedException {
       try {
         if (_segmentName.equals("")) {
           acquireSema("ALL", _logger);
+          // NOTE: the method aborts if any segment reload encounters an unhandled exception - can lead to inconsisten
```
mcvsubbu: nit: typo

Issue #3601, opened 2018-12-11
Realtime segment reload fails due to NPE in VirtualColumn code


sunithabeeram:
Reloading segments for Realtime tables fails with the exception shared below. This happens because virtual columns are written out to the segment file.

Exception:
com.linkedin.pinot.server.starter.helix.SegmentMessageHandlerFactory$SegmentReloadMessageHandler.handleMessage(SegmentMessageHandlerFactory.java:159)
        at org.apache.helix.messaging.handling.HelixTask.call(HelixTask.java:96)
        at org.apache.helix.messaging.handling.HelixTask.call(HelixTask.java:47)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
        at java.lang.Class.forName0(Native Method)
        at java.lang.Class.forName(Class.java:264)
        at com.linkedin.pinot.core.segment.virtualcolumn.VirtualColumnProviderFactory.buildProvider(VirtualColumnProviderFactory.java:30)
        at com.linkedin.pinot.core.indexsegment.immutable.ImmutableSegmentLoader.load(ImmutableSegmentLoader.java:131)
        at com.linkedin.pinot.server.starter.helix.HelixInstanceDataManager.reloadSegment(HelixInstanceDataManager.java:224)
        at com.linkedin.pinot.server.starter.helix.HelixInstanceDataManager.reloadAllSegments(HelixInstanceDataManager.java:185)
        at com.linkedin.pinot.server.starter.helix.SegmentMessageHandlerFactory$SegmentReloadMessageHandler.handleMessage(SegmentMessageHandlerFactory.java:149)

PR #3607, opened 2018-12-11
Filter out virtual-columns in realtime segment conversion path

sunithabeeram:
Fixes issues identified in #3601 

Added a unit test to ensure the virtual cols are removed.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/RealtimeSegmentConverter.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/RealtimeSegmentConverter.java
@@ -15,6 +15,7 @@
  */
 package com.linkedin.pinot.core.realtime.converter;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.linkedin.pinot.common.config.ColumnPartitionConfig;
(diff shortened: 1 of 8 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/realtime/converter/RealtimeSegmentConverter.java
```diff
@@ -150,4 +137,33 @@ public void build(@Nullable SegmentVersion segmentVersion, ServerMetrics serverM
...
+  @VisibleForTesting
+  public
+  Schema getUpdatedSchema(Schema original) {
+
+    TimeFieldSpec tfs = original.getTimeFieldSpec();
+    // Use outgoing granularity for creating segment
+    TimeGranularitySpec outgoing = tfs.getOutgoingGranularitySpec();
+    TimeFieldSpec newTimeSpec = new TimeFieldSpec(outgoing);
+
+    Schema newSchema = new Schema();
```
mcvsubbu: Will iterating over schema.getPhysicalColumns() work here? (since you added that method)

sunithabeeram: Good point. Updated.

PR #3614, opened 2018-12-13
Adding a class to get statistics about a request (query) processed by…

mcvsubbu:
… Pinot

This will enable us to publish such statistics about each request to a stream.
This stream can eventually be consumed by Pinot, and data from the stream
analyzed on a per-table basis.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/api/RequestStatistics.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/api/RequestStatistics.java
@@ -1,0 +1,171 @@
+/**
+ * Copyright (C) 2014-2015 LinkedIn Corp. (pinot-core@linkedin.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.linkedin.pinot.broker.api;
+
+import com.linkedin.pinot.common.request.BrokerRequest;
+import com.linkedin.pinot.common.request.QueryType;
+import com.linkedin.pinot.common.response.BrokerResponse;
+
+
+/**
+ * A class to hold the details regarding a request and the statistics.
+ * This object can be used to publish the query processing statistics to a stream for
+ * post-processing at a finer level than metrics.
+ */
+public class RequestStatistics {
+  private int _errorCode = 0;
+  private String _pql;
+  private String _tableName = "NotYetParsed";
+  private long _processingTimeMillis = -1;
+
+  private long _totalDocs;
+  private long _numDocsScanned;
+  private long _numEntriesScannedInFilter;
+  private long _numEntriesScannedPostFilter;
+  private long _numSegmentsQueried;
+  private long _numSegmentsProcessed;
+  private long _numSegmentsMatched;
+  private int _numServersQueried;
+  private int _numServersResponded;
+  private boolean _isNumGroupsLimitReached;
+  private int _numExceptions;
+  private boolean _isGroupBy;
+  private boolean _isAggregation;
+
+  public enum FanoutType {
+    OFFLINE,
+    REALTIME,
+    HYBRID
+  }
+
+  private FanoutType _fanoutType;
+
+  public RequestStatistics() {
+  }
+
+  public void setErrorCode(int errorCode) {
+    _errorCode = errorCode;
+  }
+
+  public void setPql(String pql) {
+    _pql = pql;
+  }
+
+  public void setTableName(String tableName) {
+    _tableName = tableName;
+  }
+
+  public void setQueryProcessingTime(long processingTimeMillis) {
+    _processingTimeMillis = processingTimeMillis;
+  }
+
+  public void setStatistics(BrokerResponse brokerResponse) {
+    _totalDocs = brokerResponse.getTotalDocs();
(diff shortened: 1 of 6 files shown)


pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -228,6 +238,9 @@ public BrokerResponse handleRequest(JSONObject request, @Nullable RequesterIdent
     Map<String, List<String>> offlineRoutingTable = null;
     Map<String, List<String>> realtimeRoutingTable = null;
     if (offlineBrokerRequest != null) {
+      if (realtimeBrokerRequest == null) {
+        requestStatistics.setFanoutType(RequestStatistics.FanoutType.OFFLINE);
```
sunithabeeram: It looks like this can be set in the if-else block of line 222?

pinot-broker/src/main/java/com/linkedin/pinot/broker/api/RequestStatistics.java
```diff
@@ -0,0 +1,171 @@
...
+import com.linkedin.pinot.common.request.QueryType;
+import com.linkedin.pinot.common.response.BrokerResponse;
+
+
+/**
+ * A class to hold the details regarding a request and the statistics.
+ * This object can be used to publish the query processing statistics to a stream for
+ * post-processing at a finer level than metrics.
+ */
+public class RequestStatistics {
```
sunithabeeram: What might be the limits of the data that can be pushed through this?

There are other tidbits of info that I would like to see in here, so would be good to know what the limits (if any) are; for ex, the brokerProcessingTime (current processingtime includes server+broker latency); it would be good to include the server-stats that are normally logged, and perhaps some client side info as well. In anycase, it will be good if this is easily extensible.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -269,6 +288,8 @@ public BrokerResponse handleRequest(JSONObject request, @Nullable RequesterIdent
     // Set total query processing time
     long totalTimeMs = TimeUnit.NANOSECONDS.toMillis(executionEndTimeNs - compilationStartTimeNs);
     brokerResponse.setTimeUsedMs(totalTimeMs);
+    requestStatistics.setQueryProcessingTime(totalTimeMs);
+    requestStatistics.setStatistics(brokerResponse);
```
sunithabeeram: This is already  being set in line 131

pinot-broker/src/main/java/com/linkedin/pinot/broker/api/RequestStatistics.java
```diff
@@ -0,0 +1,171 @@
+/**
+ * Copyright (C) 2014-2015 LinkedIn Corp. (pinot-core@linkedin.com)
```
sunithabeeram: 2018

pinot-broker/src/main/java/com/linkedin/pinot/broker/api/RequestStatistics.java
mcvsubbu: The limits are not defined here, so we can always add more stuff to this object. Limitations will be (if any) on the streams channel in which events are published. The events will contain data out of this object.  The limitation will be defined by the schema chosen for publishing. The schema may or many not include all members of RequestStatistics.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
mcvsubbu: yes. Changed.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
mcvsubbu: The one in line 131 is set using brokerRequest (and this one is from brokerResponse).

pinot-broker/src/main/java/com/linkedin/pinot/broker/api/RequestStatistics.java
```diff
@@ -0,0 +1,174 @@
...
+
+  public void setRequestId(long requestId) {
+    _requestId = requestId;
+  }
+
+  public void setRequestArrivalTimeMillis(long requestArrivalTimeMillis) {
+    _requestArrivalTimeMillis = requestArrivalTimeMillis;
+  }
+
+  public void setReduceTimeNanos(long reduceTimeNanos) {
```
sunithabeeram: Although the name does indicate the unit, I think it would be better to report this is in millis. Its just easier to deal with all latency numbers in the stats as millis so you don't have to mentally switch between units.

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -29,5 +30,6 @@
 
   void shutDown();
 
-  BrokerResponse handleRequest(JSONObject request, @Nullable RequesterIdentity requesterIdentity) throws Exception;
+  BrokerResponse handleRequest(JSONObject request, @Nullable RequesterIdentity requesterIdentity,
```
sunithabeeram: Did you want to deprecate the earlier one and add a new one instead?

pinot-broker/src/main/java/com/linkedin/pinot/broker/api/RequestStatistics.java
mcvsubbu: Done

pinot-broker/src/main/java/com/linkedin/pinot/broker/requesthandler/BrokerRequestHandler.java
mcvsubbu: yes, done

PR #3618, opened 2018-12-18
Refactor ControllerPeriodicTask to iterate over tables 

npawar:
Currently each periodic task is responsible for iterating over all tables. We plan to introduce checks for leadership before proceeding with the task for a table. In the current design, we would have to have a check in each periodic task. 
Refactoring this to pull up the iteration into ControllerPeriodicTask. This way we can have the checks just in the ControllerPeriodicTask, and each PeriodicTask is only responsible for executing the task on a given table.

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
@@ -124,7 +124,7 @@ public abstract class ControllerPeriodicTask extends BasePeriodicTask {
   protected abstract void preprocess();
 
   /**
-   * Process the controller periodic task for the given table
+   * Execute the controller periodic task for the given table
    * @param tableNameWithType
    */
   protected abstract void process(String tableNameWithType);
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
@@ -110,7 +110,29 @@ public abstract class ControllerPeriodicTask extends BasePeriodicTask {
    *
    * @param tables List of table names
    */
-  public abstract void process(List<String> tables);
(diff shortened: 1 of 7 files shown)


npawar:
Step 2 of https://github.com/apache/incubator-pinot/issues/3600

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -110,7 +110,29 @@ public void onBecomeLeader() {
...
+  /**
+   * This method runs before processing all tables
+   */
+  protected abstract void preprocess();
+
+  /**
+   * Execute the controller periodic task for the given table
+   * @param tableNameWithType
+   */
+  protected abstract void process(String tableNameWithType);
```
mcvsubbu: suggest rename to processTable

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
```diff
@@ -58,51 +58,59 @@ public RealtimeSegmentRelocator(PinotHelixResourceManager pinotHelixResourceMana
...
+  }
+
+  @Override
+  protected void process(String tableNameWithType) {
+    runRelocation(tableNameWithType);
+  }
+
+  @Override
+  protected void postprocess() {
+
```
mcvsubbu: Not related to this, but should we add some metrics on number of segments relocated?

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
npawar: done

pinot-controller/src/main/java/com/linkedin/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
npawar: Yes, that is a good idea. Will do so in a future PR

PR #3620, opened 2018-12-19
[TE] frontend - harleyjj/rootcause-visual - componentize rootcause gr…

harleyjj:
…aph for reusability in other routes

This wraps the rootcause chart, legend, toolbar, and sliding date picker into a component called rootcause-visual for future usage in other routes, such as manage/alert.  Adds booleans for legend, toolbar, and slider so that they can be rendered or not depending on needs.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/rootcause-visual/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/rootcause-visual/component.js
@@ -1,0 +1,106 @@
+import { get } from '@ember/object';
+import Component from '@ember/component';
+
+export default Component.extend({
+  showLegend: false, // boolean
+
+  showToolbar: false, // boolean
+
+  showSlider: false, // boolean
+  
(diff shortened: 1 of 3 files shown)


apucher:
@harleyjj I be glad to get more background on this. I understand you want to integrate the chart with the "Manage Alert" page, but I don't get how comparison range and legend are useful for this. Ultimately I wonder what's the value-add of this component. It doesn't seem to add any kind of shared state or interaction between the sub-components.

aaronucsd:
> @harleyjj I be glad to get more background on this. I understand you want to integrate the chart with the "Manage Alert" page, but I don't get how comparison range and legend are useful for this. Ultimately I wonder what's the value-add of this component. It doesn't seem to add any kind of shared state or interaction between the sub-components.

@apucher I think @harleyjj is trying to make the two charts or all the charts be consistent with using the same chart component. I agree that the range comparison doesn't make sense. We might want to exempt that from this component.

PR #3621, opened 2018-12-20
[TE] frontend - harleyjj/rootcause - add Anomaly to labels for rootca…

harleyjj:
…use legend and tooltip

In the case that an event is an anomaly, Anomaly will be included in the label in the Rootcause route, where toEventLabel is called


aaronucsd:
Screenshot if possible mask.

thirdeye/thirdeye-frontend/app/utils/rca-utils.js
```diff
@@ -309,7 +309,7 @@ export function toEventLabel(urn, entities) {
 
   if (urn.includes('anomaly')) {
     const [, id] = urn.split(':anomaly:');
-    label = `#${id} ${label}`;
+    label = `Anomaly #${id} ${label}`;
```
aaronucsd: The tooltip used the same?

PR #3624, opened 2018-12-22
[PINOT-7370] Return number of bytes read from the reader interfaces/implementations

sunithabeeram:
This change stems from feedback in https://github.com/apache/incubator-pinot/pull/3511. We would like to track the number of bytes read with minimal overhead. For most data-types the overhead is minimal, but for strings/byte types there is a linear time penalty. With this change, we update all batch read APIs to return the number of bytes read so its easier for higher layers to work with the data across both dictionary based and no-dictionary based readers.

### DIFF
--- a/pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvalsets/SingleValueSet.java
+++ b/pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvalsets/SingleValueSet.java
@@ -46,7 +46,7 @@ public final class SingleValueSet extends BaseBlockValSet {
   }
 
   @Override
-  public void getIntValues(int[] inDocIds, int inStartPos, int inDocIdsSize, int[] outValues, int outStartPos) {
+  public long getIntValues(int[] inDocIds, int inStartPos, int inDocIdsSize, int[] outValues, int outStartPos) {
     int inEndPos = inStartPos + inDocIdsSize;
     ReaderContext context = _reader.createContext();
     if (_dataType == DataType.INT) {
@@ -56,10 +56,11 @@ public final class SingleValueSet extends BaseBlockValSet {
     } else {
       throw new UnsupportedOperationException();
     }
+    return inDocIdsSize * Integer.BYTES;
   }
 
   @Override
-  public void getLongValues(int[] inDocIds, int inStartPos, int inDocIdsSize, long[] outValues, int outStartPos) {
+  public long getLongValues(int[] inDocIds, int inStartPos, int inDocIdsSize, long[] outValues, int outStartPos) {
     int inEndPos = inStartPos + inDocIdsSize;
     ReaderContext context = _reader.createContext();
     switch (_dataType) {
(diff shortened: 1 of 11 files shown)


pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvalsets/SingleValueSet.java
```diff
@@ -56,104 +56,114 @@ public void getIntValues(int[] inDocIds, int inStartPos, int inDocIdsSize, int[]
...
+
+    long bytesRead = 0;
     int inEndPos = inStartPos + inDocIdsSize;
     ReaderContext context = _reader.createContext();
     if (_dataType == DataType.STRING) {
       for (int i = inStartPos; i < inEndPos; i++) {
-        outValues[outStartPos++] = _reader.getString(inDocIds[i], context);
+        String val =  _reader.getString(inDocIds[i], context);
+        outValues[outStartPos++] = val;
+        bytesRead += val.length();
```
mcvsubbu: This is not really the bytes read. It is the string length. The number of bytes read depends on the encoding.
I am not sure if it is worth (performance and garbage-wise) to convert string to bytes just to get the length here.

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/readers/BaseDictionary.java
```diff
@@ -54,50 +54,62 @@ public String getStringValue(int dictId) {
...
   }
 
   @Override
-  public void readStringValues(int[] dictIds, int inStartPos, int length, String[] outValues, int outStartPos) {
+  public long readStringValues(int[] dictIds, int inStartPos, int length, String[] outValues, int outStartPos) {
+    long bytesRead = 0;
     int inEndPos = inStartPos + length;
     for (int i = inStartPos; i < inEndPos; i++) {
-      outValues[outStartPos++] = getStringValue(dictIds[i]);
+      String str = getStringValue(dictIds[i]);
```
mcvsubbu: Same comment as before

pinot-core/src/main/java/com/linkedin/pinot/core/segment/index/readers/StringDictionary.java
```diff
@@ -51,11 +51,15 @@ public String getStringValue(int dictId) {
...
 
   @Override
-  public void readStringValues(int[] dictIds, int inStartPos, int length, String[] outValues, int outStartPos) {
+  public long readStringValues(int[] dictIds, int inStartPos, int length, String[] outValues, int outStartPos) {
+    long bytesRead = 0;
     byte[] buffer = getBuffer();
     int inEndPos = inStartPos + length;
     for (int i = inStartPos; i < inEndPos; i++) {
-      outValues[outStartPos++] = getUnpaddedString(dictIds[i], buffer);
+      String val = getUnpaddedString(dictIds[i], buffer);
```
mcvsubbu: same comment as before

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvalsets/SingleValueSet.java
sunithabeeram: Good catch! I've updated the code to use bytes based method to read and decode once we have picked up the bytes read values.

pinot-core/src/main/java/com/linkedin/pinot/core/operator/docvalsets/SingleValueSet.java
```diff
@@ -56,104 +57,116 @@ public void getIntValues(int[] inDocIds, int inStartPos, int inDocIdsSize, int[]
...
+  public long getStringValues(int[] inDocIds, int inStartPos, int inDocIdsSize, String[] outValues, int outStartPos) {
+
+    long bytesRead = 0;
     int inEndPos = inStartPos + inDocIdsSize;
     ReaderContext context = _reader.createContext();
     if (_dataType == DataType.STRING) {
       for (int i = inStartPos; i < inEndPos; i++) {
-        outValues[outStartPos++] = _reader.getString(inDocIds[i], context);
+        // read as bytes and then decode to string - allows correct estimation of bytes-read
+        byte[] bytes = _reader.getBytes(inDocIds[i], context);
```
mcvsubbu: Calling getBytes here ends up allocating a new byte array VarByteChunkSingleValueReader, right? Whereas caliing getString() reuses from the thread-local buffer. Presumably this optimization was made to avoid too much garbage during queries. Should we introduce a different interface that returns both the string as well as the number of bytes?

npawar:

PR #3632, opened 2019-01-02
[TE] frontend - harleyjj/home - make email share available even when …

harleyjj:
…there are no anomalies

The route.js changes are to clean up eslint warnings.


harleyjj:
Yes, it will just display the placeholder where the anomalies would typically be. 
screen shot 2019-01-02 at 4 39 17 pm

PR #3634, opened 2019-01-03
[TE] task - randomize execution order to increase parallel throughput

apucher:


thirdeye/thirdeye-pinot/src/main/java/com/linkedin/thirdeye/anomaly/task/TaskDriver.java
```diff
@@ -147,6 +149,9 @@ private TaskDTO acquireTask() {
       if (CollectionUtils.isNotEmpty(anomalyTasks)) {
         LOG.info("Thread {} : Found {} tasks in waiting state", Thread.currentThread().getId(), anomalyTasks.size());
 
+        // shuffle candidate tasks to avoid synchronized patterns across threads (and hosts)
+        Collections.shuffle(anomalyTasks);
```
cyy0714: will it change the order of replay tasks? If yes, then this part causes the incorrectness of replay. The replay task ordering affects the final replay result a lot.

apucher: the order already isn't deterministic (multiple threads, head-tail fetching at random). So this doesn't change the guarantees or semantics of task execution

cyy0714: I see. I also figure out it doesn't affects the execution order. But if the execution order isn't deterministic, how can we promise the replay result is the same as the result they see via scheduled tasks?

PR #3636, opened 2019-01-03
[TE] frontend - aaronucsd/Basic yaml editor with alert and subscripti…

aaronucsd:
…on group editors

1. Add new component for yaml
2. Add creation and edit features

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/yaml-editor/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/yaml-editor/component.js
@@ -1,0 +1,323 @@
+/**
+ * Component to render pre-set time range selection pills and a 'custom' one using date-range-picker.
+ * @module components/range-pill-selectors
+ * @property {Object} timeRangeOptions - object containing our range options
+ * @property {Number} timePickerIncrement - determines selectable time increment in date-range-picker
+ * @property {Date} activeRangeStart - default start date for range picker
+ * @property {Date} activeRangeEnd - default end date for range picker
+ * @property {String} uiDateFormat - date format specified by parent route (often specific to metric granularity)
+ * @property {Action} selectAction - closure action from parent
+ * @example
+  {{yaml-editor
+    alertTitle="New Editor Title"
+    isEditMode=true
+    showSettings=true
+    onYMLSelector=(action "onYMLSelector")
+    saveAlertYaml=(action "saveAlertYaml")
+    cancelAlertYaml=(action "cancelAlertYaml")
+  }}
+ * @author
+ */
+
+import Component from '@ember/component';
+import { computed, set, get, getProperties } from '@ember/object';
+import { checkStatus } from 'thirdeye-frontend/utils/utils';
+import { yamlAlertProps, yamlAlertSettings, yamIt } from 'thirdeye-frontend/utils/constants';
+import yamljs from 'yamljs';
+import RSVP from "rsvp";
+import fetch from 'fetch';
+import {
+  selfServeApiGraph,
+  selfServeApiCommon
+} from 'thirdeye-frontend/utils/api/self-serve';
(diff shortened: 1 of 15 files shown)


thirdeye/thirdeye-frontend/app/pods/manage/yaml/template.hbs
```diff
@@ -0,0 +1,41 @@
+<section class="te-page__top te-search-results {{if isEditModeActive "te-search-results--slim"}}">
+  <div class="container">
+    {{#self-serve-alert-yaml-details
+      detectionYaml=model.detectionYaml
+      isLoadError=isLoadError
+      displayMode="single"
+    }}
+
+      {{!-- {{#if (not isReplayPending)}}
```
jihaozh: Still need this?

thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/controller.js
```diff
@@ -880,89 +766,18 @@ export default Controller.extend({
...
-
-    },
 
     /**
      * Clears YAML content, disables 'save changes' button, and moves to form
      */
     cancelAlertYaml() {
-      set(this, 'disableYamlSave', true);
-      set(this, 'alertYamlContent', null);
+      //set(this, 'disableYamlSave', true);
```
jihaozh: clean these lines?

thirdeye/thirdeye-frontend/app/utils/constants.js
```diff
@@ -139,3 +139,24 @@ rules:
...
+
+detectionConfigIds:
+ - 5773069
+
+fromAddress: thirdeye-dev@linkedin.com
+recipients:
+ to:
+  - "akrai@linkedin.com"
+ cc:
+  - "thirdeye-dev@linkedin.com"
```
akshayrai: This has and will potentially have a lot of LinkedIn specific things. I think we should move this out.

Maybe call an endpoint and let the backend return this value. LinkedIn stuff can be easily wrapped in the backend.

thirdeye/thirdeye-frontend/app/pods/components/yaml-editor/component.js
```diff
@@ -0,0 +1,326 @@
...
+  isYamlParseable: true,
+  alertTitle: 'Define anomaly detection in YAML',
+  alertSettingsTitle: 'Define notification settings',
+  isEditMode: false,
+  showSettings: true,
+  disableYamlSave: true,
+  errorMsg: '',
+  alertYaml: null,           // The YAML for the anomaly alert detection
+  detectionSettingsYaml:  null,   // The YAML for the subscription group
+ //currentYamlSettings: null,// The YAML content for the notification settings
```
akshayrai: I'd prefer removing commented code if not required.

thirdeye/thirdeye-frontend/app/utils/constants.js
jihaozh: this is the example template that shown as default in the alert Yaml edit page. To keep this simple, I think a yaml with dummy values (w/o linkedin specific information) would be enough for now.

jihaozh: this is the example template that shown as default in the alert Yaml edit page. To keep this simple, I think a yaml with dummy values (w/o linkedin specific information) would be enough for now.

jihaozh: this is the example template that shown as default in the alert Yaml edit page. To keep this simple, I think a yaml with dummy values (w/o linkedin specific information) would be enough for now.

aaronucsd: Yep. Agreed. TOo much work to make it api driven for something so simple as a default. I can make it generic.

PR #3637, opened 2019-01-03
[TE] detection - fix yaml translator

xiaohui-sun:
1. Minor fix for yaml translator.
2. Parse yaml first before storing into database.


xiaohui-sun:
This pr is not needed since we should not change the component map.

PR #3644, opened 2019-01-05
Split ValidationManager duties into separate ControllerPeriodicTasks

npawar:
As discussed in issue https://github.com/apache/incubator-pinot/issues/3642 we want to be able to run realtime part of the validation manager much faster than all the other activities  in validation manager. 
Splitting all the tasks of the ValidationManager into separate ControllerPeriodicTasks, so that we can better control the frequency of each one:
1) OfflineSegmentIntervalChecker - to check for missing offline segments, can run daily by default
2) RealtimeSegmentValidationManager - to fix partitions which have stopped consuming, can run hourly by default. Can be configured at the controller level to match realtime SLA of the cluster
3) BrokerResourceValidationManager - to update broker on instance set changes

### DIFF
--- a/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
@@ -51,15 +51,40 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final String CONSOLE_WEBAPP_ROOT_PATH = "controller.query.console";
   private static final String CONSOLE_WEBAPP_USE_HTTPS = "controller.query.console.useHttps";
   private static final String EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT = "controller.upload.onlineToOfflineTimeout";
-  private static final String RETENTION_MANAGER_FREQUENCY_IN_SECONDS = "controller.retention.frequencyInSeconds";
-  private static final String VALIDATION_MANAGER_FREQUENCY_IN_SECONDS = "controller.validation.frequencyInSeconds";
-  private static final String STATUS_CHECKER_FREQUENCY_IN_SECONDS = "controller.statuschecker.frequencyInSeconds";
-  private static final String REALTIME_SEGMENT_RELOCATOR_FREQUENCY = "controller.realtime.segment.relocator.frequency";
-  private static final String STATUS_CHECKER_WAIT_FOR_PUSH_TIME_IN_SECONDS = "controller.statuschecker.waitForPushTimeInSeconds";
+
+  public static class ControllerPeriodicTasksConf {
+    private static final String RETENTION_MANAGER_FREQUENCY_IN_SECONDS = "controller.retention.frequencyInSeconds";
+    private static final String OFFLINE_SEGMENT_INTERVAL_CHECKER_FREQUENCY_IN_SECONDS =
+        "controller.offline.segment.interval.checker.frequencyInSeconds";
+    private static final String REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS =
+        "controller.realtime.segment.validation.frequencyInSeconds";
+    private static final String BROKER_RESOURCE_VALIDATION_FREQUENCY_IN_SECONDS =
+        "controller.broker.resource.validation.frequencyInSeconds";
+    private static final String STATUS_CHECKER_FREQUENCY_IN_SECONDS = "controller.statuschecker.frequencyInSeconds";
+    private static final String STATUS_CHECKER_WAIT_FOR_PUSH_TIME_IN_SECONDS =
+        "controller.statuschecker.waitForPushTimeInSeconds";
+    private static final String TASK_MANAGER_FREQUENCY_IN_SECONDS = "controller.task.frequencyInSeconds";
+    private static final String REALTIME_SEGMENT_RELOCATOR_FREQUENCY =
+        "controller.realtime.segment.relocator.frequency";
+    // Because segment level validation is expensive and requires heavy ZK access, we run segment level validation with a
+    // separate interval
+    private static final String SEGMENT_LEVEL_VALIDATION_INTERVAL_IN_SECONDS =
+        "controller.segment.level.validation.intervalInSeconds";
+
+    private static final int DEFAULT_RETENTION_CONTROLLER_FREQUENCY_IN_SECONDS = 6 * 60 * 60; // 6 Hours.
+    private static final int DEFAULT_OFFLINE_SEGMENT_INTERVAL_CHECKER_FREQUENCY_IN_SECONDS = 6 * 60 * 60; // 6 Hours.
+    private static final int DEFAULT_REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS = 60 * 60; // 1 Hour.
+    private static final int DEFAULT_BROKER_RESOURCE_VALIDATION_FREQUENCY_IN_SECONDS = 60 * 60; // 1 Hour.
+    private static final int DEFAULT_STATUS_CONTROLLER_FREQUENCY_IN_SECONDS = 5 * 60; // 5 minutes
+    private static final int DEFAULT_STATUS_CONTROLLER_WAIT_FOR_PUSH_TIME_IN_SECONDS = 10 * 60; // 10 minutes
+    private static final int DEFAULT_TASK_MANAGER_FREQUENCY_IN_SECONDS = -1; // Disabled
+    private static final String DEFAULT_REALTIME_SEGMENT_RELOCATOR_FREQUENCY = "1h"; // 1 hour
+    private static final int DEFAULT_SEGMENT_LEVEL_VALIDATION_INTERVAL_IN_SECONDS = 24 * 60 * 60;
+  }
+
   private static final String SERVER_ADMIN_REQUEST_TIMEOUT_SECONDS = "server.request.timeoutSeconds";
   private static final String SEGMENT_COMMIT_TIMEOUT_SECONDS = "controller.realtime.segment.commit.timeoutSeconds";
   private static final String DELETED_SEGMENTS_RETENTION_IN_DAYS = "controller.deleted.segments.retentionInDays";
-  private static final String TASK_MANAGER_FREQUENCY_IN_SECONDS = "controller.task.frequencyInSeconds";
   private static final String TABLE_MIN_REPLICAS = "table.minReplicas";
   private static final String ENABLE_SPLIT_COMMIT = "controller.enable.split.commit";
   private static final String JERSEY_ADMIN_API_PORT = "jersey.admin.api.port";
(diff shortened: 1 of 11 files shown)


npawar:
Thanks for pointing that out @mcvsubbu 
I went over the reads that happen in all 3 tasks.

1. OfflineSegmentIntervalChecker
 reads ideal state of offline table
reads all segment metadata for the offline table

2. RealtimeSegmentValidationManager
reads ideal state of realtime table
reads only names of all segment metadata for the realtime table
reads all instance configs inside StreamPartitionAssignmentGenerator for the table

3. BrokerResourceValidationManager
reads ideal state for the table
reads instance configs once in preprocess

It doesn't look like any reads will be repeated _because_ of splitting out these tasks. 1 and 2 read different ideal state/segment metadata. 2 and 3 both read instance configs, which they both did even before splitting, so it's not a new addition. The instance configs read will be better anyway after https://github.com/apache/incubator-pinot/pull/3633

Let me know what you think

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/RealtimeSegmentValidationManager.java
```diff
@@ -0,0 +1,162 @@
...
+        return;
+      }
+
+      CommonConstants.Helix.TableType tableType = TableNameBuilder.getTableTypeFromTableName(tableNameWithType);
+      if (tableType == CommonConstants.Helix.TableType.REALTIME) {
+        if (_updateRealtimeDocumentCount) {
+          updateRealtimeDocumentCount(tableConfig);
+        }
+      }
+      Map<String, String> streamConfigMap = tableConfig.getIndexingConfig().getStreamConfigs();
```
jackjlli: Should this logic be put to the check of realtime table, ie. the brackets between Line 86 and Line 90?

npawar: Good catch! Moved it inside.
Also moved the code to fetch table config inside the if, so that we avoid fetching table config from zk unnecessarily

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/BrokerResourceValidationManager.java
```diff
@@ -0,0 +1,80 @@
...
+  private List<InstanceConfig> _instanceConfigs;
+
+  public BrokerResourceValidationManager(ControllerConf config, PinotHelixResourceManager pinotHelixResourceManager) {
+    super("BrokerResourceValidationManager", config.getBrokerResourceValidationFrequencyInSeconds(),
+        pinotHelixResourceManager);
+  }
+
+  @Override
+  protected void preprocess() {
+    // Cache instance configs to reduce ZK access
```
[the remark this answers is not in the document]
npawar: I think it was put there by original author to say that we are fetching instance configs before processing all tables. 
Removed it, as it is confusing and probably irrelevant now

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/OfflineSegmentIntervalChecker.java
```diff
@@ -49,83 +40,41 @@
...
-  private final int _segmentLevelValidationIntervalInSeconds;
   private final PinotLLCRealtimeSegmentManager _llcRealtimeSegmentManager;
   private final ValidationMetrics _validationMetrics;
 
-  private long _lastSegmentLevelValidationTimeMs = 0L;
-  private boolean _runSegmentLevelValidation;
-  private List<InstanceConfig> _instanceConfigs;
-
-  public ValidationManager(ControllerConf config, PinotHelixResourceManager pinotHelixResourceManager,
+  public OfflineSegmentIntervalChecker(ControllerConf config, PinotHelixResourceManager pinotHelixResourceManager,
```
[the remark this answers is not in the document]
npawar: Yes, left behind by mistake. Removed it.

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -51,15 +51,40 @@
...
-  private static final String VALIDATION_MANAGER_FREQUENCY_IN_SECONDS = "controller.validation.frequencyInSeconds";
-  private static final String STATUS_CHECKER_FREQUENCY_IN_SECONDS = "controller.statuschecker.frequencyInSeconds";
-  private static final String REALTIME_SEGMENT_RELOCATOR_FREQUENCY = "controller.realtime.segment.relocator.frequency";
-  private static final String STATUS_CHECKER_WAIT_FOR_PUSH_TIME_IN_SECONDS = "controller.statuschecker.waitForPushTimeInSeconds";
+
+  public static class ControllerPeriodicTasksConf {
+    private static final String RETENTION_MANAGER_FREQUENCY_IN_SECONDS = "controller.retention.frequencyInSeconds";
+    private static final String OFFLINE_SEGMENT_INTERVAL_CHECKER_FREQUENCY_IN_SECONDS =
+        "controller.offline.segment.interval.checker.frequencyInSeconds";
+    private static final String REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS =
```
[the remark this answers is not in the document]
npawar: That can be done for brokerResourceValidation. Changed it accordingly.

But for offlineSegment, we need to look at new default. The default for offline should be 6 hours. In the previous design of validation manager, the configured frequency was 1h, and the 6h was being controlled by internal flags and time variables. The OfflineSegmentIntervalChecker thus cannot rely on the older validation frequency.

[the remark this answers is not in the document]
mcvsubbu: What if there was an existing configuration controller.segment.level.validation.intervalInSeconds=1hr?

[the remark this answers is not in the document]
mcvsubbu: Maybe we should keep "controller.segment.level.validation.intervalInSeconds"  and use this for offline segment checker as well as realtime completed segment checker (to update numdocs).

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -321,58 +339,118 @@ public String getControllerVipProtocol() {
...
+        Integer.toString(validationFrequencyInSeconds));
+  }
+
+  public int getRealtimeSegmentValidationFrequencyInSeconds() {
+
+    if (containsKey(ControllerPeriodicTasksConf.REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS)) {
+      return Integer.parseInt(
+          (String) getProperty(ControllerPeriodicTasksConf.REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS));
+    }
+    return ControllerPeriodicTasksConf.DEFAULT_REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS;
```
mcvsubbu: if controller.validation.frequencyInSeconds is configured, return it here.

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -51,15 +51,42 @@
   private static final String CONSOLE_WEBAPP_ROOT_PATH = "controller.query.console";
   private static final String CONSOLE_WEBAPP_USE_HTTPS = "controller.query.console.useHttps";
   private static final String EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT = "controller.upload.onlineToOfflineTimeout";
-  private static final String RETENTION_MANAGER_FREQUENCY_IN_SECONDS = "controller.retention.frequencyInSeconds";
-  private static final String VALIDATION_MANAGER_FREQUENCY_IN_SECONDS = "controller.validation.frequencyInSeconds";
```
mcvsubbu: rename to DEPRECATED_VALIDATION_MANAGER_FREQUENCY_IN_SECONDS

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -321,58 +339,118 @@ public String getControllerVipProtocol() {
...
+   * Return broker resource validation frequency if present, else return the validation manager frequency
+   * This is so that we can rollout with no config changes to the frequency of this task
+   * @return
+   */
+  public int getBrokerResourceValidationFrequencyInSeconds() {
+    if (containsKey(ControllerPeriodicTasksConf.BROKER_RESOURCE_VALIDATION_FREQUENCY_IN_SECONDS)) {
+      return Integer.parseInt(
+          (String) getProperty(ControllerPeriodicTasksConf.BROKER_RESOURCE_VALIDATION_FREQUENCY_IN_SECONDS));
+    }
+    return getValidationControllerFrequencyInSeconds();
```
mcvsubbu: if controller.validation.frequencyInSeconds is configured, return it here.

pinot-controller/src/main/java/com/linkedin/pinot/controller/validation/RealtimeSegmentValidationManager.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        if (_updateRealtimeDocumentCount) {
+          updateRealtimeDocumentCount(tableConfig);
+        }
+
+        Map<String, String> streamConfigMap = tableConfig.getIndexingConfig().getStreamConfigs();
+        StreamConfig streamConfig = new StreamConfig(streamConfigMap);
+        if (streamConfig.hasLowLevelConsumerType()) {
+          _llcRealtimeSegmentManager.validateLLCSegments(tableConfig);
```
mcvsubbu: maybe rename the method in llcRealtimeSegmentManager to be ensureConsumingSegments() or some such?

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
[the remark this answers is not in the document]
npawar: Changed this.
`OfflineSegmentIntervalChecker`  defaults to `getSegmentLevelValidationIntervalInSeconds()`
`RealtimeSegmentValidationManager` defaults to `getValidationControllerFrequencyInSeconds()`
`BrokerResourceValidationManager` defaults to `getValidationControllerFrequencyInSeconds()`

This retains all behaviour, wherein broker and realtime validations happen at validation manager freq and offline checks happen at segment level interval

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -321,58 +342,116 @@ public String getControllerVipProtocol() {
...
+   * If it doesn't exist, returns the validation controller frequency. This is done in order to retain the current behavior,
+   * wherein the realtime validation tasks were done at validation controller frequency
+   * @return
+   */
+  public int getRealtimeSegmentValidationFrequencyInSeconds() {
+    if (containsKey(ControllerPeriodicTasksConf.REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS)) {
+      return Integer.parseInt(
+          (String) getProperty(ControllerPeriodicTasksConf.REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS));
+    }
+    return getValidationControllerFrequencyInSeconds();
```
mcvsubbu: If VALIDATION_CONTROLLER_FREQUENCY is not set, you want to return REALTIME_SEGMENT_VALIDATION_FREQUENCY here, right? So, calling getValidationControllerFrequency method will return the wrong value, right?

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -321,58 +342,116 @@ public String getControllerVipProtocol() {
...
+   * wherin the broker resource validation tasks were done at validation controller frequency
+   * @return
+   */
+  public int getBrokerResourceValidationFrequencyInSeconds() {
+    if (containsKey(ControllerPeriodicTasksConf.BROKER_RESOURCE_VALIDATION_FREQUENCY_IN_SECONDS)) {
+      return Integer.parseInt(
+          (String) getProperty(ControllerPeriodicTasksConf.BROKER_RESOURCE_VALIDATION_FREQUENCY_IN_SECONDS));
     }
-    return DEFAULT_VALIDATION_CONTROLLER_FREQUENCY_IN_SECONDS;
+    return getValidationControllerFrequencyInSeconds();
```
mcvsubbu: same as above

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
npawar: That was done to totally preserve older behavior. 
Changed this to reflect new defaults after our discussion offline

PR #3661, opened 2019-01-08
Add documentation for the new config format

jfim:


docs/table_configuration.rst
```diff
@@ -0,0 +1,220 @@
...
+    schema {
+      // ...
+    }
+    "streamConfigs.stream.kafka.broker.list": "${config.kafkaHost}"
+  }
+
+Configuration profiles
+----------------------
+
+Configuration profiles can also be used to simplify configuration management across different Pinot deployments. The Pinot
```
mcvsubbu: I suggest: "To simplify configuration management, Pinot provides a way to set up configuration profiles."

docs/table_configuration.rst
```diff
@@ -0,0 +1,220 @@
...
+  profiles=[prod, prod-us-west-4, test, test-us-east-2]
+  table {
+    // ...
+    "streamConfigs.stream.kafka.broker.list": "${config.kafkaHost}"
+  }
+
+If the Pinot tools are used to deploy this table configuration with the profiles ``prod`` and ``prod-us-west-4``, this configuration
+will get deployed to that particular Pinot deployment. However, the table would be skipped if the configuration profiles ``prod``
+and ``prod-tokyo-1`` are passed, due to the ``prod-tokyo-1`` configuration profile not being part of the ``profiles`` section.
+
```
mcvsubbu: will be good to add an explicit note that _all_ profiles in a table configuration must be included in the Pinot deployment command in order for a table configuration to be deployed.

docs/table_configuration.rst
```diff
@@ -0,0 +1,220 @@
...
+For batch automatic deployments, the table configuration file parameter can be a directory, which will recursively process all the table
+definition files in that directory.
+
+We recommend managing table configurations using a source control system and automating table deployments by running the
+``ApplyTableConfig`` command on source control changes or periodically using a scheduler.
+
+Backing up table configurations
+-------------------------------
+
+Table configurations can be backed up using the ``BackupTableConfigs`` command:
```
mcvsubbu: Can you please clarify that "backing up" here means downloading configurations from the Pinot controller (to the current directory?)

docs/table_configuration.rst
```diff
@@ -0,0 +1,220 @@
...
+
+  pinot/bin/pinot-admin.sh BackupTableConfigs -tableName myTable \
+      -controllerUrl http://pinot-controller-proxy.us-west-4.cloudprovider.com:23456/
+
+Migrating from the legacy configuration format
+----------------------------------------------
+
+For single cluster Pinot deployments, migration is done by backing up all of the table configurations as shown above.
+
+For multiple cluster Pinot deployments, an additional merge step needs to be done to minimize the number of table files.
```
mcvsubbu: It will be useful to expand this a bit. I suppose the merge step is really not needed, but is useful (as you have pointed out, to minimize the number of files).
I would add a step between 1 and 2 that says something like "Create a folder structure depending on your deployments. For example, you can have folders configBackup/prod/prod-us-west-4, configBackup/prod/prod-us-east-3"
Expanding step 2 a bit is also useful. As in "Now change directory to each of the folders for specific deployment and enter the BackupTableConfigs command"

Some clarifications:
Is there a requirement that the 'profiles' directory be under configBackup/ some place? Can you document it either way?

docs/table_configuration.rst
```diff
@@ -0,0 +1,220 @@
...
+    //...
+    replication___prod = 3
+    replication___testing = 1
+  }
+
+In the example above, the replication for the table will be 3 if the ``prod`` profile is enabled, 1 if the ``testing``
+profile is enabled, and unset otherwise.
+
+There is no conflict resolution between conflicting conditional values, so a configuration containing simultaneous
+different values for profile conditionals is invalid.
```
mcvsubbu: So, the result will be that the ApplyTableConfig command will throw an error? If so, please mention that. 
Also, I suppose there is no implicit assumption that 'prod-us-west-4' is an extension of 'prod' profile, right? In the previous example, if somehow 'prod' and 'testing' profiles are enabled simultaneously it would result in the same error?

npawar:

PR #3664, opened 2019-01-08
Update license header

sunithabeeram:
This change updates the license header on the files in preparation for Apache release.

Changes done:
- Include a new license.header file in the top level directory instead of relying on the apache2.0 template (which by default includes copyright year and owner).
- Updated the existing license headers using mvn license:format command.


pom.xml
```diff
@@ -1102,11 +1105,7 @@
         <artifactId>license-maven-plugin</artifactId>
         <version>2.8</version>
         <configuration>
-          <header>com/mycila/maven/plugin/license/templates/APACHE-2.txt</header>
-          <properties>
-            <owner>LinkedIn Corp.</owner>
-            <email>pinot-core@linkedin.com</email>
-          </properties>
+          <header>license.header</header>
           <excludes>
```
Jackie-Jiang: Please check all the excludes as well. Some of them might not make sense, some of them might need to be updated.

PR #3667, opened 2019-01-08
Try fixing flaky tests by adding 0.5s sleep

jackjlli:
Try fixing flaky Test `testInstanceListingAndCreation` by adding 0.5s sleep.

### DIFF
--- a/pinot-controller/src/test/java/com/linkedin/pinot/controller/api/resources/PinotInstanceRestletResourceTest.java
+++ b/pinot-controller/src/test/java/com/linkedin/pinot/controller/api/resources/PinotInstanceRestletResourceTest.java
@@ -63,6 +63,9 @@ public class PinotInstanceRestletResourceTest extends ControllerTest {
     serverInstance.put("host", "2.3.4.5");
     sendPostRequest(_controllerRequestURLBuilder.forInstanceCreate(), serverInstance.toString());
 
+    // Sleep 0.5s for cache accessor to update cache.
+    Thread.sleep(500L);
+
     // Check that there are four instances
     instanceList = new JSONObject(sendGetRequest(_controllerRequestURLBuilder.forInstanceList()));
     assertEquals(instanceList.getJSONArray("instances").length(), 4,


pinot-controller/src/test/java/com/linkedin/pinot/controller/api/resources/PinotInstanceRestletResourceTest.java
```diff
@@ -63,6 +63,9 @@ public void testInstanceListingAndCreation() throws Exception {
     serverInstance.put("host", "2.3.4.5");
     sendPostRequest(_controllerRequestURLBuilder.forInstanceCreate(), serverInstance.toString());
 
+    // Sleep 0.5s for cache accessor to update cache.
+    Thread.sleep(500L);
```
sunithabeeram: Is the sleep required only here?

Also, a better pattern might be to use TestUtils.waitForCondition instead of a simple sleep if we are to wait on results of asynchronous operations to get reflected.

sunithabeeram: See example here: https://github.com/apache/incubator-pinot/blob/master/pinot-integration-tests/src/test/java/com/linkedin/pinot/integration/tests/OfflineClusterIntegrationTest.java#L162

jackjlli: Done. Thanks for the tip!

pinot-controller/pom.xml
```diff
@@ -65,6 +65,12 @@
       <groupId>com.linkedin.pinot</groupId>
       <artifactId>pinot-core</artifactId>
     </dependency>
+    <dependency>
```
mcvsubbu: Controller should not pull in test jars.

mcvsubbu: Ah no, my bad, sorry. I believe this is fine.

PR #3671, opened 2019-01-10
 Make different PinotFS concrete classes have the same behaviors

jackjlli:
Currently different PinotFS behaves differently. 
E.g., as for `listFile(URI, true)` LocalPinotFS lists all the files and directories, while HadoopPinotFS only list non-dir files.
This PR:
* makes sure that these two `FS` have the same behaviors.
* follows up from https://github.com/apache/incubator-pinot/pull/3589
* unit tests added.

### DIFF
--- a/pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
+++ b/pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
@@ -56,13 +56,13 @@ public abstract class PinotFS implements Closeable {
 
   /**
    * Moves the file from the src to dst. Does not keep the original file. If the dst has parent directories
-   * that haven't been created, this method will create all the necessary parent directories. If dst already exists,
-   * it will overwrite. Will work either for moving a directory or a file. Currently assumes that both the srcUri
-   * and the dstUri are of the same type (both files or both directories). If srcUri is a file, it will move the file to
-   * dstUri's location. If srcUri is a directory, it will move the directory to dstUri. Does not support moving a file under a
-   * directory.
-   *
-   * For example, if a/b/c is moved to x/y/z, in the case of overwrite, a/b/c will be renamed to x/y/z.
+   * that haven't been created, this method will create all the necessary parent directories.
+   * If both src and dst are files, dst will be overwritten.
+   * If src is a file and dst is a directory, src file will get copied under dst directory.
+   * If both src and dst are directories, src folder will get copied under dst directory.
+   * If src is a directory and dst is a file, operation will fail.
+   * For example, if a file a/b/c is moved to x/y/z, in the case of overwrite, a/b/c will be renamed to x/y/z;
+   * if a file x is copied to x/y, all the original files under x/y will be kept.
    * @param srcUri URI of the original file
    * @param dstUri URI of the final file location
    * @param overwrite true if we want to overwrite the dstURI, false otherwise
@@ -74,8 +74,8 @@ public abstract class PinotFS implements Closeable {
       throws IOException;
 
   /**
-   * Same as move except the srcUri is retained. For example, if x/y/z is copied to a/b/c, x/y/z will be retained
-   * and x/y/z will also be present as a/b/c; if x is copied to x/y, all the original files under x/y will be kept.
+   * Same as move except the srcUri is retained. For example, if a file x/y/z is copied to a/b/c, x/y/z will be retained
+   * and x/y/z will also be present as a/b/c; if a file x is copied to x/y, all the original files under x/y will be kept.
    * @param srcUri URI of the original file
    * @param dstUri URI of the final file location
    * @return true if copy is successful
@@ -95,7 +95,7 @@ public abstract class PinotFS implements Closeable {
       throws IOException;
 
   /**
-   * Returns the length of the file at the provided location. Will throw exception if a directory
+   * Returns the length of the file at the provided location. Will throw exception if a directory.
    * @param fileUri location of file
    * @return the number of bytes
    * @throws IOException on IO Failure
--- a/pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
+++ b/pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
@@ -74,8 +74,8 @@ public abstract class PinotFS implements Closeable {
       throws IOException;
 
   /**
-   * Same as move except the srcUri is not retained. For example, if x/y/z is copied to a/b/c, x/y/z will be retained
-   * and x/y/z will also be present as a/b/c
+   * Same as move except the srcUri is retained. For example, if x/y/z is copied to a/b/c, x/y/z will be retained
+   * and x/y/z will also be present as a/b/c; if x is copied to x/y, all the original files under x/y will be kept.
    * @param srcUri URI of the original file
    * @param dstUri URI of the final file location
    * @return true if copy is successful
@@ -105,7 +105,7 @@ public abstract class PinotFS implements Closeable {
       throws IOException;
 
   /**
-   * Lists all the files at the location provided. Lists recursively.
+   * Lists all the files and directories at the location provided. Lists recursively.
    * Throws exception if this abstract pathname is not valid, or if
    * an I/O error occurs.
    * @param fileUri location of file
@@ -143,7 +143,7 @@ public abstract class PinotFS implements Closeable {
    * @return true if uri is a directory, false otherwise.
    * @throws Exception if uri is not valid or present
    */
-  public abstract boolean isDirectory(URI uri);
+  public abstract boolean isDirectory(URI uri) throws IOException;
 
   /**
    * Returns the age of the file
--- a/pinot-filesystem/src/test/java/org/apache/pinot/filesystem/LocalPinotFSTest.java
+++ b/pinot-filesystem/src/test/java/org/apache/pinot/filesystem/LocalPinotFSTest.java
@@ -141,13 +141,36 @@ public class LocalPinotFSTest {
     Assert.assertEquals(_newTmpDir.listFiles().length, files);
     Assert.assertFalse(dstFile.exists());
 
-    // Check that a moving a file a non-existent destination folder will work
+    // Check that copying a file to a non-existent destination folder will work
     FileUtils.deleteQuietly(_nonExistentTmpFolder);
     Assert.assertFalse(_nonExistentTmpFolder.exists());
     File srcFile = new File(_absoluteTmpDirPath, "srcFile");
     localPinotFS.mkdir(_absoluteTmpDirPath.toURI());
     Assert.assertTrue(srcFile.createNewFile());
     dstFile = new File(_nonExistentTmpFolder.getPath() + "/newFile");
+    Assert.assertFalse(dstFile.exists());
+    Assert.assertTrue(localPinotFS.copy(srcFile.toURI(), dstFile.toURI()));
+    Assert.assertTrue(srcFile.exists());
+    Assert.assertTrue(dstFile.exists());
+
+    //Check that copying a folder to a non-existent destination folder works
+    FileUtils.deleteQuietly(_nonExistentTmpFolder);
+    Assert.assertFalse(_nonExistentTmpFolder.exists());
+    localPinotFS.mkdir(_absoluteTmpDirPath.toURI());
+    dstFile = new File(_nonExistentTmpFolder.getPath() + "/srcFile" );
+    Assert.assertFalse(dstFile.exists());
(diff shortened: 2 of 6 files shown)


pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
```diff
@@ -97,7 +97,8 @@ public boolean move(URI srcUri, URI dstUri, boolean overwrite) throws IOExceptio
   public boolean copy(URI srcUri, URI dstUri) throws IOException {
     File srcFile = new File(decodeURI(srcUri.getRawPath()));
     File dstFile = new File(decodeURI(dstUri.getRawPath()));
-    if (dstFile.exists()) {
+    // delete dst only if dst isn't under src.
+    if (!dstFile.getCanonicalPath().startsWith(srcFile.getCanonicalPath()) && dstFile.exists()) {
```
sunithabeeram: Is it valid to delete dst if its a directory?

If someone attempts to copy name1 to dst/name1, where the source name1 is a directory and dst/name1 is a file, shouldn't the copy fail?

pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/HadoopPinotFS.java
```diff
@@ -97,39 +109,93 @@ public boolean move(URI srcUri, URI dstUri, boolean overwrite) throws IOExceptio
...
-    if (sourceFiles != null) {
-      while (sourceFiles.hasNext()) {
-        boolean succeeded = FileUtil.copy(_hadoopFS, sourceFiles.next().getPath(), _hadoopFS, target, true, _hadoopConf);
-        if (!succeeded) {
-          return false;
+    if (!exists(srcUri)) {
+      LOGGER.warn("Source {} does not exist", srcUri);
+      return false;
+    }
+    if (srcUri.equals(dstUri)) {
```
sunithabeeram: What if srcUri is a directory and has files that actually need to copied over?

pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/HadoopPinotFS.java
```diff
@@ -97,39 +109,93 @@ public boolean move(URI srcUri, URI dstUri, boolean overwrite) throws IOExceptio
...
+      LOGGER.info("Source {} and destination {} are the same.", srcUri, dstUri);
+      return true;
+    }
+    if (!dstUri.getRawPath().startsWith(srcUri.getRawPath()) && exists(dstUri)) {
```
sunithabeeram: Same concern as above with deleting dst if its a folder.

pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/HadoopPinotFS.java
```diff
@@ -139,6 +205,20 @@ public long length(URI fileUri) throws IOException {
...
+    List<FileStatus> fileStatuses = new ArrayList<>();
+    FileStatus[] files = _hadoopFS.listStatus(path);
+    for (FileStatus file : files) {
+      fileStatuses.add(file);
+      if (file.isDirectory() && recursive) {
+        List<FileStatus> subFiles = Arrays.asList(listStatus(file.getPath(), true));
+        fileStatuses.addAll(subFiles);
+      }
+    }
+    FileStatus[] fileStatusesArr = new FileStatus[fileStatuses.size()];
```
sunithabeeram: Why can't we return the array-list here?

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/PinotFS.java
```diff
@@ -71,7 +71,7 @@
 
   /**
    * Same as move except the srcUri is not retained. For example, if x/y/z is copied to a/b/c, x/y/z will be retained
```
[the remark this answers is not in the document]
jackjlli: Good catch! 👍

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
jackjlli: In your case, src is name1 and dst is dst/name1.
So the condition `!dstFile.getCanonicalPath().startsWith(srcFile.getCanonicalPath())` is satisfied and so is`dstFile.exists()`. Thus, the previous dst will be deleted.

pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/HadoopPinotFS.java
jackjlli: Do you mean there's an edge case that Line 116 can't handle? 
If src and dst are the same, regardless of src's type, nothing should be done.

pinot-hadoop-filesystem/src/main/java/com/linkedin/pinot/filesystem/HadoopPinotFS.java
jackjlli: The reason is that I'm not sure whether this API will be exposed in the future. Right now all the `listStatus` from `Hadoop FileSystem` return an FileStatus array. So I think it's best to follow the convention.

pinot-filesystem/src/main/java/com/linkedin/pinot/filesystem/LocalPinotFS.java
sunithabeeram: I don't think previous dst/name1 should be deleted. Will be good to review this and see if we should honor this contract:
If name1 is a folder and dst/name1 is a file, operation should fail.
If name1 is a folder and dst/name1 is a folder, src folder will get copied under dst/name1.
If name1 is a file and dst/name1 is a file, dst/name1 will get overwritten.
If name1 is a file and dst/name1 is a folder, src file will get copied under dst/name1.

jackjlli: That's a good question. @jenniferdai do you have any comments on it?
I've modified the code to keep the original destination dir if it exists.

jenniferdai: You just need to make sure that all code references to these methods have the same contract as well.

jackjlli: Discussed with @jenniferdai offline. She's happy with the changes @sunithabeeram proposed.

jackjlli: @jenniferdai I've verified that all code references to these methods have the same contract.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -74,8 +74,8 @@ public abstract boolean move(URI srcUri, URI dstUri, boolean overwrite)
       throws IOException;
 
   /**
-   * Same as move except the srcUri is not retained. For example, if x/y/z is copied to a/b/c, x/y/z will be retained
-   * and x/y/z will also be present as a/b/c
+   * Same as move except the srcUri is retained. For example, if a file x/y/z is copied to a/b/c, x/y/z will be retained
```
mcvsubbu: Instead of "Same as move" can you mention each case explicitly and describe the behavior? thanks.

pinot-hadoop-filesystem/src/main/java/org/apache/pinot/filesystem/HadoopPinotFS.java
```diff
@@ -189,9 +272,8 @@ public void copyFromLocalFile(File srcFile, URI dstUri)
   }
 
   @Override
-  public boolean isDirectory(URI uri) {
-    FileStatus fileStatus = new FileStatus();
-    fileStatus.setPath(new Path(uri));
+  public boolean isDirectory(URI uri) throws IOException {
```
mcvsubbu: can we just invoke _hadoopFS.isDirectory()

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -95,7 +95,7 @@ public abstract boolean exists(URI fileUri)
       throws IOException;
 
   /**
-   * Returns the length of the file at the provided location. Will throw exception if a directory
+   * Returns the length of the file at the provided location. Will throw exception if a directory.
```
mcvsubbu: Looks like this API was added to this class only for testing? Pinot code does not seem access it.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -105,7 +105,7 @@ public abstract long length(URI fileUri)
       throws IOException;
 
   /**
-   * Lists all the files at the location provided. Lists recursively.
+   * Lists all the files and directories at the location provided. Lists recursively.
```
mcvsubbu: Lists recursively if "recursive" is set to true.

Also, if the URI provided is not a directory, are we expected to throw exception no matter what the value of 'recursive' is? i think the localfs will not throw exception if recursive is set to false

pinot-filesystem/src/test/java/org/apache/pinot/filesystem/LocalPinotFSTest.java
```diff
@@ -117,15 +117,24 @@ public void testFS()
...
+    Assert.assertTrue(localPinotFS.exists(new File(testDirUnderNewAbsoluteTempDirPath3, "testFile").toURI()));
 
     // Check file copy to location where something already exists still works
     localPinotFS.copy(testFileUri, thirdTestFile.toURI());
     // Check length of file
     Assert.assertEquals(0, localPinotFS.length(secondTestFileUri));
     Assert.assertTrue(localPinotFS.exists(thirdTestFile.toURI()));
 
+    // Check that destination being directory within the source directory still works
+    File anotherFileUnderAbsoluteThreeDir = new File(newAbsoluteTempDirPath3, "anotherFile");
```
mcvsubbu: thanks for adding these test cases

pinot-hadoop-filesystem/src/main/java/org/apache/pinot/filesystem/HadoopPinotFS.java
```diff
@@ -92,9 +96,18 @@ public boolean delete(URI segmentUri, boolean forceDelete)
...
+    if (!exists(srcUri)) {
+      LOGGER.warn("Source {} does not exist", srcUri);
       return false;
     }
+    if (exists(dstUri)) {
+      if (!overwrite) {
+        return false;
+      } else {
+        delete(dstUri, true);
+        mkdir(dstUri);
```
mcvsubbu: not sure why mkdir here. shouldn't we just be doing a move and let the file system move either a file or a dir?

pinot-hadoop-filesystem/src/main/java/org/apache/pinot/filesystem/HadoopPinotFS.java
```diff
@@ -105,31 +118,87 @@ public boolean move(URI srcUri, URI dstUri, boolean overwrite)
...
-    Path source = new Path(srcUri);
-    Path target = new Path(dstUri);
-    RemoteIterator<LocatedFileStatus> sourceFiles = _hadoopFS.listFiles(source, true);
-    if (sourceFiles != null) {
-      while (sourceFiles.hasNext()) {
-        boolean succeeded =
-            FileUtil.copy(_hadoopFS, sourceFiles.next().getPath(), _hadoopFS, target, true, _hadoopConf);
-        if (!succeeded) {
-          return false;
+    if (!exists(srcUri)) {
```
mcvsubbu: Maybe use a base class and move these there? You can add a TODO

mcvsubbu: One thing that occurs to me as I review something like this, is that for some (slow) FS, the latency of making each call may be so bad, that we are better off calling some native file system method and you get what you get. Going this route is relatively easy, and we just need to make sure that Pinot's requirements are met in all file system (i.e. if Pinot uses the case of copying from a dir to a file and that case works fine across everything, then that is alll we need).

Let us discuss a bit more before we put in too much work to unify file systems.

jackjlli:
I'm closing this PR since it's been split to multiple PRs.

PR #3674, opened 2019-01-10
[TE] license - update to apache license headers

apucher:


apucher:
hacking my contributor stats!

PR #3675, opened 2019-01-10
Sslbranch

fredxfred:
I have been working on identifying components in the Pinot code base where data (aside from metadata e.g. segment name) can be transmitted over the network. This PR is to enable SSL configurations to be passed into controllers and brokers, and then to configure grizzly to support https. I can provide examples of the full configs of an example https controller/broker + a script to generate self-signed certs for testing if desired.

Please let me know if there is anything missing that needs to be added. I am still testing these changes (and others, see below). I wasn't sure how to perform in-code tests of whether the data is being properly encrypted but I am definitely open to suggestions.

Currently, this code does not allow a component to be configured with both http and https, which is a feature that has also been requested. Also, this does not include code I have written to configure Netty Servers, only Grizzly. Fortunately, when we URI push Azure data, that data is encrypted by default, so currently we have no plans to work on encryption during the segmentfetcher/upload process. I am putting this code up now to get a head start on review, and I do plan on adding Netty encryption and the ability to allow both http and https.

### DIFF
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerAdminApiApplication.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerAdminApiApplication.java
@@ -33,6 +33,8 @@ import org.glassfish.grizzly.http.server.HttpServer;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
 import org.glassfish.jersey.grizzly2.httpserver.GrizzlyHttpServerFactory;
 import org.glassfish.jersey.server.ResourceConfig;
+import org.glassfish.grizzly.ssl.SSLContextConfigurator;
+import org.glassfish.grizzly.ssl.SSLEngineConfigurator;
 
 
 public class BrokerAdminApiApplication extends ResourceConfig {
@@ -40,6 +42,12 @@ public class BrokerAdminApiApplication extends ResourceConfig {
 
   private URI _baseUri;
   private HttpServer _httpServer;
+  
+  private String _keyStoreFile; // contains server keypair
+  private String _keyStorePass;
+  private String _trustStoreFile; // contains client certificate
+  private String _trustStorePass;
+  private boolean _useHTTPS = false;
 
   public BrokerAdminApiApplication(BrokerServerBuilder brokerServerBuilder) {
     packages(RESOURCE_PACKAGE);
@@ -57,10 +65,30 @@ public class BrokerAdminApiApplication extends ResourceConfig {
     registerClasses(io.swagger.jaxrs.listing.SwaggerSerializers.class);
   }
 
+  public void setSSLConfigs(String keyStoreFile, String keyStorePass, String trustStoreFile, String trustStorePass) {
+	  _useHTTPS = true;
+      _keyStoreFile = keyStoreFile;
+	  _keyStorePass = keyStorePass;
+	  _trustStoreFile = trustStoreFile;
+	  _trustStorePass = trustStorePass;
+  }
+
   public void start(int httpPort) {
     Preconditions.checkArgument(httpPort > 0);
-    _baseUri = URI.create("http://0.0.0.0:" + httpPort + "/");
-    _httpServer = GrizzlyHttpServerFactory.createHttpServer(_baseUri, this);
+
+    if (_useHTTPS) {
+      _baseUri = URI.create("https://0.0.0.0:" + httpPort + "/");
+      SSLContextConfigurator sslContext = new SSLContextConfigurator();
+	  sslContext.setKeyStoreFile(_keyStoreFile);
+	  sslContext.setKeyStorePass(_keyStorePass);
+	  sslContext.setTrustStoreFile(_trustStoreFile);
+	  sslContext.setTrustStorePass(_trustStorePass);
+	  _httpServer = GrizzlyHttpServerFactory.createHttpServer(_baseUri, this, true, 
+        new SSLEngineConfigurator(sslContext).setClientMode(false).setNeedClientAuth(false));
+    } else {
+      _baseUri = URI.create("http://0.0.0.0:" + httpPort + "/");
+      _httpServer = GrizzlyHttpServerFactory.createHttpServer(_baseUri, this);
+    }
     setupSwagger();
   }
 
@@ -70,7 +98,12 @@ public class BrokerAdminApiApplication extends ResourceConfig {
     beanConfig.setDescription("APIs for accessing Pinot broker information");
     beanConfig.setContact("https://github.com/linkedin/pinot");
     beanConfig.setVersion("1.0");
-    beanConfig.setSchemes(new String[]{"http"});
+    if (_useHTTPS) {
+      beanConfig.setSchemes(new String[]{"https"});
+    }
+    else {
+      beanConfig.setSchemes(new String[]{"http"});
+    }
     beanConfig.setBasePath(_baseUri.getPath());
     beanConfig.setResourcePackage(RESOURCE_PACKAGE);
     beanConfig.setScan(true);
--- a/pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerAdminApiApplication.java
+++ b/pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerAdminApiApplication.java
@@ -79,7 +79,7 @@ public class BrokerAdminApiApplication extends ResourceConfig {
     if (_useHTTPS) {
       _baseUri = URI.create("https://0.0.0.0:" + httpPort + "/");
       SSLContextConfigurator sslContext = new SSLContextConfigurator();
-	  sslContext.setKeyStoreFile(_keyStoreFile);
+          sslContext.setKeyStoreFile(_keyStoreFile);
 	  sslContext.setKeyStorePass(_keyStorePass);
 	  sslContext.setTrustStoreFile(_trustStoreFile);
 	  sslContext.setTrustStorePass(_trustStorePass);
--- a/pinot-broker/src/test/java/com/linkedin/pinot/broker/broker/HelixBrokerStarterTest.java
+++ b/pinot-broker/src/test/java/com/linkedin/pinot/broker/broker/HelixBrokerStarterTest.java
@@ -76,7 +76,7 @@ public class HelixBrokerStarterTest {
     _zkClient = new ZkClient(ZkStarter.DEFAULT_ZK_STR);
     final String instanceId = "localhost_helixController";
     _pinotResourceManager =
-        new PinotHelixResourceManager(ZkStarter.DEFAULT_ZK_STR, HELIX_CLUSTER_NAME, instanceId, null, 10000L, true, /*isUpdateStateModel=*/false, true);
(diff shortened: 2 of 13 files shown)


fredxfred:
Just added support for using both http and https. Hopefully this works.

We ended up not going forward with encrypting the netty servers since currently ours aren't exposed to the public internet. Unlike with controllers and brokers, we can't put the servers' apis behind load balancers as easily, so it makes certs too much of a hastle.

To keep the controller query portal working after the brokers' apis were encrypted, I added a configuration that allows the user to change the way query portal queries are executed by passing all queries directly to a load balancer.

kishoreg:
@mcvsubbu Can you take a look at this PR and see if there is anything pending? @fredxfred, can you please rebase this. I think the conflicts are mostly because of reformatting the code when we moved to Apache.

mcvsubbu:
@fredxfred I missed this again. I will resolve this today.
cc: @kishoreg

pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerAdminApiApplication.java
```diff
@@ -57,10 +65,30 @@ protected void configure() {
     registerClasses(io.swagger.jaxrs.listing.SwaggerSerializers.class);
   }
 
+  public void setSSLConfigs(String keyStoreFile, String keyStorePass, String trustStoreFile, String trustStorePass) {
+	  _useHTTPS = true;
```
mcvsubbu: please fix indentation

pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerAdminApiApplication.java
```diff
@@ -57,10 +65,30 @@ protected void configure() {
...
+	  _trustStoreFile = trustStoreFile;
+	  _trustStorePass = trustStorePass;
+  }
+
   public void start(int httpPort) {
     Preconditions.checkArgument(httpPort > 0);
-    _baseUri = URI.create("http://0.0.0.0:" + httpPort + "/");
-    _httpServer = GrizzlyHttpServerFactory.createHttpServer(_baseUri, this);
+
+    if (_useHTTPS) {
```
mcvsubbu: If I am reading this right, we start one server that serves  _either_ https _or_ http. How will existing installations migrate to the new broker? Should we not be starting _both_ http and https servers, so that the PQL clients can first move to https, and then the brokers can be configured to serve https only? So we need an http as well as https port.

The start() method must be maintained backward compatibile, so one way to do this is to introduce a new method that provides two ports.  We may deprecate the existing method or just retain it, just not starting https if it is called.

Correct me if I am wrong.

pinot-broker/src/main/java/com/linkedin/pinot/broker/broker/BrokerAdminApiApplication.java
```diff
@@ -70,7 +98,12 @@ private void setupSwagger() {
     beanConfig.setDescription("APIs for accessing Pinot broker information");
     beanConfig.setContact("https://github.com/linkedin/pinot");
     beanConfig.setVersion("1.0");
-    beanConfig.setSchemes(new String[]{"http"});
+    if (_useHTTPS) {
+      beanConfig.setSchemes(new String[]{"https"});
+    }
```
mcvsubbu: please move up the else to this line, as per coding conventions.  See https://pinot.readthedocs.io/en/latest/dev_guide.html on how to set up your dev guide so that coding conventions are automatic. thanks.

pinot-broker/src/test/java/com/linkedin/pinot/broker/broker/HelixBrokerStarterTest.java
```diff
@@ -76,7 +76,7 @@ public void setUp() throws Exception {
     _zkClient = new ZkClient(ZkStarter.DEFAULT_ZK_STR);
     final String instanceId = "localhost_helixController";
     _pinotResourceManager =
-        new PinotHelixResourceManager(ZkStarter.DEFAULT_ZK_STR, HELIX_CLUSTER_NAME, instanceId, null, 10000L, true, /*isUpdateStateModel=*/false, true);
+        new PinotHelixResourceManager(ZkStarter.DEFAULT_ZK_STR, HELIX_CLUSTER_NAME, instanceId, null, 10000L, true, /*isUpdateStateModel=*/false, true, null);
```
mcvsubbu: this has been re-factored, so you won't need this I think

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -200,12 +207,64 @@ public String getQueryConsoleWebappPath() {
...
+    setProperty(USE_HTTP, useHTTP);
   }
 
-  public boolean getQueryConsoleUseHttps() {
-    return containsKey(CONSOLE_WEBAPP_USE_HTTPS) && getBoolean(CONSOLE_WEBAPP_USE_HTTPS);
+  public boolean getUseHttp() {
+    if (!containsKey(USE_HTTP)) {
+      return true;
+    } else {
+      return (boolean) getBoolean(USE_HTTP);
```
mcvsubbu: Do we need this cast?

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -270,6 +333,14 @@ public String getControllerPort() {
     return (String) getProperty(CONTROLLER_PORT);
   }
 
+  public String getControllerHttpsPort() {
+    if (!containsKey(CONTROLLER_HTTPS_PORT)) {
+      return "443";
```
mcvsubbu: Can you please add a static final for the default port like the other configurations?

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerStarter.java
```diff
@@ -97,7 +98,24 @@
...
+    }
+    else {
+      _encryptedAdminApp = null;
+    }
+
+    if (_config.getUseHttp()) {
+      LOGGER.info("Initializing http server");
+      _adminApp = new ControllerAdminApiApplication(_config.getQueryConsoleWebappPath(), false);
+    }
+    else {
```
mcvsubbu: coding conv

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerStarter.java
```diff
@@ -244,15 +264,39 @@ protected void configure() {
...
+        bind(connectionManager).to(HttpConnectionManager.class);
+        bind(_executorService).to(Executor.class);
+        bind(_controllerMetrics).to(ControllerMetrics.class);
+        bind(accessControlFactory).to(AccessControlFactory.class);
+        bind(metadataEventNotifierFactory).to(MetadataEventNotifierFactory.class);
+        }
+      });
+      _encryptedAdminApp.start(httpsJerseyPort);
+      LOGGER.info("Started Jersey API for https server on port {}", httpsJerseyPort);
+      LOGGER.info("Pinot controller ready and listening on port {} for API requests", httpsJerseyPort);
```
mcvsubbu: s/port/https port/
thanks

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PqlQueryResource.java
```diff
@@ -108,32 +109,39 @@ public String getQueryResponse(String pqlQuery, String traceEnabled, HttpHeaders
...
     String tableName = TableNameBuilder.extractRawTableName(brokerRequest.getQuerySource().getTableName());
 
-    // Validate data access
-    AccessControl accessControl = _accessControlFactory.create();
-    if (!accessControl.hasDataAccess(httpHeaders, tableName)) {
-      return QueryException.ACCESS_DENIED_ERROR.toString();
-    }
+    String brokerLBurl = _pinotHelixResourceManager.getBrokerLoadBalancerAddress();
+    if (brokerLBurl != null) {
+      LOGGER.info("using load balancer address: {}", brokerLBurl);
```
mcvsubbu: This log in info level is too verbose.

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PqlQueryResource.java
```diff
@@ -108,32 +109,39 @@ public String getQueryResponse(String pqlQuery, String traceEnabled, HttpHeaders
...
+      String url = "https://" + brokerLBurl + "/query";
+      return sendPQLRaw(url, pqlQuery, traceEnabled);
+    } else {
+      // Validate data access
+      AccessControl accessControl = _accessControlFactory.create();
```
mcvsubbu: We need to validate access control irrespective of https or http, right?

pinot-controller/src/main/java/com/linkedin/pinot/controller/ControllerConf.java
```diff
@@ -200,12 +207,64 @@ public String getQueryConsoleWebappPath() {
...
+  
+  public boolean getUseHttps() {
+	return containsKey(USE_HTTPS) && getBoolean(USE_HTTPS);
+  }
+  
+  public void setKeyStoreFile(String keyStoreFile) {
+	setProperty(KEYSTORE_FILE, keyStoreFile);
+  }
+
+  public String getBrokerLoadbalancerAddress() {
```
mcvsubbu: There can be multiple brokers in the cluster, each serving some subset of the tables. Not sure what the admin is supposed to set for the broker load balancer here

pinot-controller/src/main/java/com/linkedin/pinot/controller/api/resources/PqlQueryResource.java
```diff
@@ -108,32 +109,39 @@ public String getQueryResponse(String pqlQuery, String traceEnabled, HttpHeaders
     }
     String tableName = TableNameBuilder.extractRawTableName(brokerRequest.getQuerySource().getTableName());
 
-    // Validate data access
-    AccessControl accessControl = _accessControlFactory.create();
-    if (!accessControl.hasDataAccess(httpHeaders, tableName)) {
-      return QueryException.ACCESS_DENIED_ERROR.toString();
-    }
+    String brokerLBurl = _pinotHelixResourceManager.getBrokerLoadBalancerAddress();
```
mcvsubbu: We should get the list of broker instances for the table from pinot helix resource manager, and then use https on them if configured to do so. So, perhaps that calls for another config to use https when querying brokers.

mcvsubbu: Alternative is for brokers to set their own configs in Helix instance config. We have gone that route with server http ports for example.

mcvsubbu:
Can you split this into two PRs? The broker one should be relatively simple, so let us get that out first.

fredxfred:
Hi, yes I will work on splitting the PRs and addressing the code review soon.

npawar:

PR #3677, opened 2019-01-11
Unify all JSON usage to fasterxml.jackson package

Jackie-Jiang:
Remove the usage of json package from org.json, org.codehaus.jackson, com.google.code.gson, com.alibaba
Add JsonUtils class to reuse ObjectMapper and provide util methods

### DIFF
--- a/pinot-api/pom.xml
+++ b/pinot-api/pom.xml
@@ -57,8 +57,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.json</groupId>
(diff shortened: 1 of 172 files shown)


kishoreg:
Awesome!  How can we make sure that it can parse all the existing configurations in Zookeeper?

Jackie-Jiang:
@kishoreg If all the tests passed, then it should be fine. fasterxml.jackson is the newer version of codehaus.jackson, which we are using right now.

snleee:
Thank you for working ont his!

PR #3679, opened 2019-01-11
[TE] frontend - harleyjj/share-dashboard - maintains columns and dash…

harleyjj:
…board-level comments in share link

This moves extra comparison columns and checkmarks into tree, which is ultimately the 'shared state' put into the blob, for consistent settings after following share link.

The email toggle buttons do not line up with the shown columns after following share link - I'm working on that next.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/home/share-dashboard/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/home/share-dashboard/route.js
@@ -11,6 +11,9 @@ import {
   set,
   setProperties
 } from '@ember/object';
+import { appendFilters } from 'thirdeye-frontend/utils/rca-utils';
+import { humanizeFloat, humanizeChange, checkStatus } from 'thirdeye-frontend/utils/utils';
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-frontend/app/pods/home/share-dashboard/route.js
```diff
@@ -123,6 +126,69 @@ export default Route.extend(AuthenticatedRouteMixin, {
...
+  _fetchOffsets: task (function * (anomalyMapping) {
+    if (!anomalyMapping) { return; }
+
+    let map = {};
+    let index = 1;
+    // Iterate through each anomaly
+    yield Object.keys(anomalyMapping).some(function(metric) {
+      Object.keys(anomalyMapping[metric].items).some(function(alert) {
+        anomalyMapping[metric].items[alert].items.forEach(async (item) => {
+          const metricName = get(item.anomaly, 'metricName');
```
aaronucsd: We can use `getProperties` here.

PR #3680, opened 2019-01-11
Move all files in com.linkedin to org.apache

npawar:
For building an apache release, we need to change all packages from com.linkedin to org.apache.
This is step 1 of the move, where we are simply moving the files to the new directory structure. This will be followed up with another PR which changes all imports and references.


mcvsubbu:
Please note that this change will break the build. We are following this method so that we can retain the history of the files.

PR #3681, opened 2019-01-11
Rename all package references from com.linkedin to org.apache

npawar:
For building an apache release, we need to change all packages from com.linkedin to org.apache.
This is step 2 of the move, where we renaming all imports and references to com.linkedin to org.apache


mcvsubbu:
Pls go ahead and merge it so that we have minimal time that build is broken. thanks

PR #3683, opened 2019-01-11
Optimize imports for all files

Jackie-Jiang:
The import optimization is based on the LinkedIn Style which will be checked in in a separate PR.


sunithabeeram:
What does "optimize" mean in this case? And how was this change made?

Jackie-Jiang:
@sunithabeeram This is done by IDE automatically, which orders all imports alphabetically, removes redundant imports, and merge imports if possible.

sunithabeeram:
> @sunithabeeram This is done by IDE automatically, which orders all imports alphabetically, removes redundant imports, and merge imports if possible.

Can we pull these rules into pinot-style.xml so everyone applies them? Are there other ways to better enforce formatting so we don't diverge going forward?

Jackie-Jiang:
> > @sunithabeeram This is done by IDE automatically, which orders all imports alphabetically, removes redundant imports, and merge imports if possible.
> 
> Can we pull these rules into pinot-style.xml so everyone applies them? Are there other ways to better enforce formatting so we don't diverge going forward?

I tried to apply the pinot-style.xml and get the same rule for imports. This can be easily done by using the IDE optimize imports feature before submitting any changes. There is no way to enforce formatting though.

sunithabeeram:
We should use the maven check-style plugin with a set of rules that work for us; The google check-style rules include ones to group/order imports: https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml

Jackie-Jiang:
> We should use the maven check-style plugin with a set of rules that work for us; The google check-style rules include ones to group/order imports: https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml

Found the LinkedIn style, will address in a separate PR.

PR #3685, opened 2019-01-12
Change the realtime stats file to a new name

npawar:
The stats files generated in realtime servers will no longer be deserializable. This is because when serializing, the older objects were used, the ones with "com.linkedin". After the package name change, the newer objects will be used to deserialize, the ones with "org.apache", and the deserialization will fail.
Changing the stats file name from "stats.ser" to "statsV2.ser". We will start populating stats afresh in all realtime tables.
Please note: stats from stats.ser will not be considered once the new server is started.


pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -65,7 +65,7 @@
   private RealtimeSegmentStatsHistory _statsHistory;
   private final Semaphore _segmentBuildSemaphore;
 
-  private static final String STATS_FILE_NAME = "stats.ser";
+  private static final String STATS_FILE_NAME = "statsV2.ser";
```
mcvsubbu: Well, we are really not changing the version of the file (in the sense that the formatting, the fields, etc, are the same), So , I would call it something else altogether. "segment-stats.ser" seems ok.

Also, please add a comment here that says that the old name used to stats.ser and we had to change it when we move the package to be org.apache

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/RealtimeSegmentStatsHistoryTest.java
```diff
@@ -222,9 +222,9 @@ public void testMultiThreadedUse() throws Exception {
   // This test attempts to ensure that future modifications to RealtimeSegmentStatsHistory does not prevent the software
   // from reading data serialized by earlier versions. The serialized data has one segment, with 2 columns -- "v1col1" and
   // "v1col2".
-  //@Test
+  @Test
   public void testVersion1() throws Exception {
-    final String fileName = "realtime-segment-stats-history-v1.ser";
+    final String fileName = "realtime-segment-stats-history.ser";
```
mcvsubbu: I would retain the old name, because the _contents_ of this file is the actual V1 contents,

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
npawar: done

npawar:
If resetting stats is not desirable, the old stats will have to be copied over to the new file. 
I wrote a tool to help with this translation. This tool was a java project which had both versions of the file `RealtimeSegmentStatsHistory.java` in the same project (`com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory` and  `org.apache.pinot.core.realtime.impl.RealtimeSegmentStatsHistory`). It also had a main method which created a new stats file and copied the old stats to this file.

Here's snippets from a tool I wrote.

Main.java
```
import com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory;
import java.io.File;
import java.io.IOException;
import java.nio.file.Files;


public class Main {

    public static void main(String[] args) throws Exception {

        String oldStatsFilePath = args[0];
        RealtimeSegmentStatsHistory
            oldHistory = RealtimeSegmentStatsHistory.deserialzeFrom(new File(oldStatsFilePath));

        String newStatsFilePath = args[1];
        File newStatsFile = new File(newStatsFilePath);
        Files.deleteIfExists(newStatsFile.toPath());

        org.apache.pinot.core.realtime.impl.RealtimeSegmentStatsHistory newHistory;
        try {
            newHistory = org.apache.pinot.core.realtime.impl.RealtimeSegmentStatsHistory.deserialzeFrom(newStatsFile);
        } catch (IOException | ClassNotFoundException e) {
            throw new RuntimeException(
                "Exception when deserializing stats history from stats file " + newStatsFile.getAbsolutePath(), e);
        }

        // copy entries and cursor
        newHistory.setEntries(oldHistory.getEntries());
        newHistory.setCursor(oldHistory.getCursor());

        System.out.println(oldHistory.toString());
        System.out.println(newHistory.toString());

    }
}
```

Added these methods in the `org.apache.pinot.core.realtime.impl.RealtimeSegmentStatsHistory` file:

setEntries
```
public void setEntries(com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory.SegmentStats[] linkedinEntries) {
    for (com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory.SegmentStats linkedinEntry : linkedinEntries) {
      if (linkedinEntry == null) {
        continue;
      }
      SegmentStats newSegmentStats = convertFromLinkedinToApache(linkedinEntry);
      addSegmentStats(newSegmentStats);
    }
  }
```


convertFromLinkedinToApache for SegmentStats:
```
public SegmentStats convertFromLinkedinToApache(
      com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory.SegmentStats linkedinSegmentStats) {
    SegmentStats apacheSegmentStats = new SegmentStats();
    apacheSegmentStats.setNumRowsConsumed(linkedinSegmentStats.getNumRowsConsumed());
    apacheSegmentStats.setNumRowsIndexed(linkedinSegmentStats.getNumRowsIndexed());
    apacheSegmentStats.setNumSeconds(linkedinSegmentStats.getNumSeconds());
    apacheSegmentStats.setMemUsedBytes(linkedinSegmentStats.getMemUsedBytes());
    Map<String, com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory.ColumnStats> colNameToStats =
        linkedinSegmentStats.getColNameToStats();
    for (Map.Entry<String, com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory.ColumnStats> entry : colNameToStats.entrySet()) {
      ColumnStats newColumnStats = convertFromLinkedinToApache(entry.getValue());
      String column = entry.getKey();
      apacheSegmentStats.setColumnStats(column, newColumnStats);
    }
    return apacheSegmentStats;
  }
```


convertFromLinkedinToApache for ColumnStats:
```
public ColumnStats convertFromLinkedinToApache(
      com.linkedin.pinot.core.realtime.impl.RealtimeSegmentStatsHistory.ColumnStats linkedinColumnStats) {
    ColumnStats apacheColumnStats = new ColumnStats();
    apacheColumnStats.setAvgColumnSize(linkedinColumnStats.getAvgColumnSize());
    apacheColumnStats.setCardinality(linkedinColumnStats.getCardinality());
    return apacheColumnStats;
  }
```

Some getters and setters will need to be added to both classes (cursor, entries, segmentStats map etc.)

Ran this tool to convert all `stats.ser` files on realtime servers to `segment-stats.ser`

PR #3687, opened 2019-01-14
exclude single char folders from check

apucher:
this is a half-solution to ignore single-character directories - including the "?" directory


apucher:
@Jackie-Jiang

PR #3689, opened 2019-01-15
[TE] detection - update tunable interface

jihaozh:
update the tunable interface to include metric urn.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/components/Tunable.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/components/Tunable.java
@@ -32,8 +32,9 @@ public interface Tunable<T extends AbstractSpec> extends BaseComponent<T> {
    * Returns the new spec for the component it's tuning for a given metric urn.
    * @param currentSpec current spec for the component. empty if not exist.
    * @param tuningWindow the tuning window
-   * @param metricUrn the metric anomaly detection to run on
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/components/Tunable.java
```diff
@@ -33,7 +31,9 @@
   /**
    * Returns the new spec for the component it's tuning
    * @param currentSpec current spec for the component. empty if not exist
+   * @param tuningWindow the tuning window
+   * @param metricUrn the metric detection runs on
```
cyy0714: should be the metric tuning runs on.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
```diff
@@ -144,6 +144,7 @@
   private static final String PROP_PARAMS = "params";
   private static final String PROP_METRIC_URN = "metricUrn";
   private static final String PROP_DIMENSION_FILTER_METRIC = "dimensionFilterMetric";
+  private static final String PROP_DIMENSIONS = "dimensions";
```
cyy0714: Where it is used?

jihaozh: removed

PR #3691, opened 2019-01-15
Update license-maven-plugin setting to correctly exclude files

Jackie-Jiang:
Also remove stale license headers from excluded files


Jackie-Jiang:
@apucher Can you try this branch and see if it can build?

pinot-common/src/main/java/org/apache/pinot/common/metadata/segment/SegmentPartitionMetadata.java
```diff
@@ -18,22 +18,6 @@
  */
 package org.apache.pinot.common.metadata.segment;
 
-/**
```
sunithabeeram: Good catch :)

PR #3695, opened 2019-01-16
Cleaning up the pom.xml file

snleee:
1. Updated information on apache
2. Removed dependecies that are not used.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,11 @@
   <artifactId>pinot</artifactId>
   <version>0.016</version>
   <packaging>pom</packaging>
-  <name>pinot</name>
+
+  <name>Pinot</name>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -45,60 +49,49 @@
     <module>pinot-hadoop-filesystem</module>
     <module>pinot-azure-filesystem</module>
   </modules>
+
   <licenses>
     <license>
       <name>Apache License, Version 2.0</name>
-      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+      <url>http://www.apache.org/licenses/LICENSE-2.0</url>
```
Jackie-Jiang: I think we should use the **.txt** file here

PR #3701, opened 2019-01-17
Adding Apache Rat plugin for maven build

snleee:
1. Added Apache Rat plugin which is an audit tool for checking licenses for apache project.
2. Added header for some files based on the rat-plugin result

In order to run rat plugin, one can run the following command.
$ mvn apache-rat:check


snleee:
After discussion with @Jackie-Jiang offline, we have decided to stick with the current `license-maven-plugin`.

PR #3703, opened 2019-01-17
Fix the json deserialization issue in ColumnPartitionMetadata

Jackie-Jiang:
For old jackson version (2.8.3), JsonDeserialize annotation on fields are not picked up correctly when we use parameterized constructor
Moving the annotation from field into the constructor fixes the issue

Also change jackson version from 2.9.6 to 2.8.3 to be compatible with jersey


Jackie-Jiang:
@snleee Updated

PR #3704, opened 2019-01-17
Fix SegmentStatusChecker exception

npawar:
`SegmentStatusChecker` is instantiated in the `ControllerStarter` constructor, right after `PinotHelixResourceManager` is instantiated. The `SegmentStatusChecker` (and all other periodic tasks) receive the `_pinotHelixResourceManager` object as a parameter in their constructor. However, the `_pinotHelixResourceManager` is not supposed to be used until `PinotHelixResourceManager::start()` is invoked. Only after invoking start() does the `_pinotHelixResourceManager` properly get initialized i.e. the _helixAdmin, _helixClusterName, _propertyStore get set. Until start() is invoked, the _helixAdmin, _helixClusterName and _propertyStore will be `null` in the `_pinotHelixResourceManager`.
As a result, even though we pass `_pinotHelixResourceManager` to all periodic tasks in their constructor, we cannot be accessing these objects, until after the invocation of start().
This isn't obviously stated anywhere in the objects receiving this `_pinotHelixResourceManager`. 

As a followup to this PR, we should bring in changes that will not allow this to happen. 
Either:
1) invoke start() on _pinotHelixResourceManager when we instantiate it i.e in the constructor (before passing it to any other object)
2) Move instantiation of other objects that depend on _pinotHelixResourceManager, to some place which is after the invocation of start(). But this would again be susceptible to erroneous design, as there's no way to stop someone using it earlier.
3) Disallow direct usage of _helixAdmin, _helixClusterName and _propertyStore, and create methods in PinotHelixResourceManager to manage these accesses. But again, this will not stop someone from calling these methods prematurely.
Option 1 seems the best to me

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -1928,8 +1928,8 @@ public class PinotHelixResourceManager {
 
   /**
    * Gets the ideal state of the table
-   * @param tableNameWithType
-   * @return
+   * @param tableNameWithType Table name with suffix
+   * @return IdealState of tableNameWithType
    */
   public IdealState getTableIdealState(String tableNameWithType) {
     return _helixAdmin.getResourceIdealState(_helixClusterName, tableNameWithType);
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1926,6 +1926,24 @@ public boolean hasOfflineTable(String tableName) {
     return getAllTables().contains(actualTableName);
   }
 
+  /**
+   * Gets the ideal state of the table
+   * @param tableNameWithType
```
Jackie-Jiang: [Minor] If we are not going to add doc for param and return, remove these 2 lines.

sunithabeeram:
Could we add a test to cover this path?

npawar:
I'm not sure a test case can be added to stop this. Imo it has to be a slight refactoring of when we allow to start passing the _pinotHelixResourceManager as arguments. I've update the PR description with more details

PR #3707, opened 2019-01-17
Fix the issue of realtime data manager calling wrong API to load segment

Jackie-Jiang:
Currently realtime data manager does not have machanism to download a new
segment copy from controller and re-load the segment when encountering
exception (HLC maintains segment on server side only), so realtime data
manager should always call ImmutableSegmentLoader.load() without Schema.

Also ignore virtual columns when loading segment metadata from metadata
file. Ideally virtual columns should never been stored into metadata file,
but we add this as an extra protection.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -758,7 +758,7 @@ public class LLRealtimeSegmentDataManager extends RealtimeSegmentDataManager {
       return false;
     }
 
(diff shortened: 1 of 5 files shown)


sunithabeeram:
Looks like the code was using the non-schema based API and that was changed with this commit: https://github.com/apache/incubator-pinot/pull/2000/files

We should consider better usability of the APIs involved and/or add checks in code that asserts when the incorrect/unexpected API is used + add tests to ensure coverage so if there are regressions we can catch them better.

Jackie-Jiang:
@sunithabeeram There is no easy way to add table type check in the loader layer (table name is not available). Also, I don't think calling the wrong API without reading the API doc should be considered bad usability of the APIs. This is a quite urgent fix, and I don't want this to involve big changes for now.

PR #3710, opened 2019-01-17
[TE] aaronucsd/fix for the live alerts not showing bug

aaronucsd:


harleyjj:
elegant fix

apucher:
> > elegant fix
> 
> technically we should not need to do a `toLowerCase` on an empty string. :)

this here is the definition of pre-mature optimization :-)

PR #3718, opened 2019-01-18
Remove dependencies from libraries under category X licenses.

snleee:
For an ASF project, its distribution should not contain any of work under
category X license. After running "mvn project-info-reports:dependencies",
three libraries are under category X libraries for our distribution package.
This PR resolves this issue.

1. Removed "pinot-perf" module from distribution package because pinot-perf
   module is pulling "jmh-core" based on GPL 2 license, which is a category X.
2. Bumped up the version of org.testng from 6.0.1 to 6.11 because the old
   version was pulling in "org.beanshell:bsh:2.0b4", which is also based on
   LGPL license. The newer version no longer pulls beanshell.
3. Bumped up the swagger version from 1.5.10 to 1.5.15. The old version was
   pulling "org.reflections:reflections:0.9.10", which pulls
   "com.google.code.findbugs:annotations", which was based on LGPL.

For the list of category X licenses, please refer
https://www.apache.org/legal/resolved.html


felixcheung:
is there replacement for "pinot-perf" module?

snleee:
@felixcheung `pinot-perf` module includes some benchmarking codes that are mainly used for measuring performances (e.g. dictionary, index reader, query engine..etc). Given that this is mainly for testing, I don't think that we need to include this for distribution package.

snleee:
@felixcheung As you mentioned, `pinot-perf` will still be included in the source code distribution while it won't be a part of binary distribution. I have added the documentation on how to build & run `pinot-perf` module from the source code and noted that this module will pull `jmh` which is under GPL 2 license.

PR #3723, opened 2019-01-19
Remove auto-gen class files that were checked-in

kishoreg:
Removed the auto-generated classes from request.thrift and response.thrift. The generated files show up under target/generated-sources/thrift and are automatically added to the source folder.  There is one downside with this. We need thrift installed to compile and build Pinot code.


pinot-common/src/main/java/org/apache/pinot/common/config/Deserializer.java
```diff
@@ -398,8 +398,9 @@ public ConfigIncluder withFallback(ConfigIncluder fallback) {
     config = config.resolve();
 
     try {
-      return deserialize(clazz, io.vavr.collection.HashSet.ofAll(config.entrySet())
-          .toMap(entry -> Tuple.of(entry.getKey(), entry.getValue().unwrapped())), "");
+      Map<String, Object> map = io.vavr.collection.HashSet.ofAll(config.entrySet())
```
Jackie-Jiang: Anything changed here?

pinot-common/pom.xml
```diff
@@ -72,6 +72,25 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-enforcer-plugin</artifactId>
       </plugin>
+       <plugin>
```
Jackie-Jiang: Can you reformat this file?

snleee:
> This is great! @snleee do we need to add extra license for the new dependency?

@Jackie-Jiang `thrift` is an apache project so there is no issue. We will clean up `LICENSE` when we do the next release.

kishoreg:
There are a couple of things we need to consider here

1. we will have to modify Travis script to install thrift before triggering the build (this is not hard, we can use thrift docker images)
2. Adds the additional step of installing thrift in order to build Pinot locally. Currently, since the generated code is checked. This is a one-time thing

The only way to avoid the above steps is to migrate to Avro or Protobuf but that will be a bigger change.

If there are no concerns, we can add the steps to install thrift to Travis script.

.travis.yml
```diff
@@ -20,6 +20,13 @@ cache:
 addons:
   firefox: latest
 
+before_install:
+  - sudo apt-get update -qq
```
kishoreg: Will this be part of build instructions as well. If so, should we move these into install_thrift.sh script?

fx19880617:
This is for just for Ubuntu travis image. We should point people to thrift page(https://thrift.apache.org/docs/install/) for build instruction based on their OS.

.travis.yml
fx19880617: updated readme, this should still be here.

fx19880617:
Upgraded libthrift to 0.12.0 so we could compile the java code generated by thrift 0.12.0.
Tested backward-compatibility that new libthrift could compile the java code generated by thrift 0.9.x.

.travis.yml
```diff
@@ -20,6 +20,13 @@ cache:
...
+  - sudo apt-get install automake bison flex g++ git libboost-all-dev libevent-dev libssl-dev libtool make pkg-config
+  - wget http://archive.apache.org/dist/thrift/0.9.3/thrift-0.9.3.tar.gz -O /tmp/thrift-0.9.3.tar.gz
```
snleee: Can we update this to `0.12.0` to make versions in sync?

.travis.yml
```diff
@@ -20,6 +20,13 @@ cache:
...
+  - tar xfz /tmp/thrift-0.9.3.tar.gz --directory  /tmp
```
Jackie-Jiang: (nit) extra space before `/tmp`

.travis_test.sh
```diff
@@ -58,12 +58,12 @@ fi
 passed=0
 # Only run integration tests if needed
 if [ "$RUN_INTEGRATION_TESTS" != 'false' ]; then
-  mvn test -B -P travis,travis-integration-tests-only
+  mvn test -B -P travis,travis-integration-tests-only,build-thrift
```
Jackie-Jiang: We don't need build-thrift profile in test right?

.travis_test.sh
```diff
@@ -58,12 +58,12 @@ fi
...
   if [ $? -eq 0 ]; then
     passed=1
   fi
 else
-  mvn test -B -P travis,travis-no-integration-tests
+  mvn test -B -P travis,travis-no-integration-tests,build-thrift
```
Jackie-Jiang: Same here

snleee:
@fx19880617 
Does `mvn clean` removes java files that are created and copied? 
Also, please wait until we make sure that our internal tool supports installing thrift (I'm not sure if we can use `sudo` there). I will get back to you.

fx19880617:
`mvn clean` will remove the generated java files, which means that it won't compile with `mvn clean install`. One option is to move the thrift generation from `profile` to `build`. So it will run without specifying `-P build-thrift`

fx19880617:
Updated pom file to let thrift generate java code during maven compile time all the time.

snleee:
@fx19880617 Is it possible to pass the following as a parameter and use the following value as a default value? Moving thrift executable binary file to `/usr/local/bin` requires `sudo` access and making thrift available there for all machines that are running our internal release pipeline will take some time. If this can be configured, I can look into a workaround.
```
              <thriftExecutable>/usr/local/bin/thrift</thriftExecutable>
```

fx19880617:
Updated PR.

You can run `mvn install package -DskipTests -Pbin-dist -Dthrift.executable=$(which thrift)`, if thrift is in your classpath.

Or just specify the path of thrift executable, e.g. `mvn clean install -DskipTests -Dthrift.executable=/usr/local/Cellar/thrift/0.12.0/bin/thrift`

PR #3724, opened 2019-01-19
[TE] Migration endpoints for anomaly function and application

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionMigrationResource.java
```diff
@@ -288,12 +376,112 @@ private String getBucketPeriod(AnomalyFunctionDTO functionDTO) {
...
+        // Build the new alert object
+        Map<String, Object> detectionAlertYaml = translateAlertToYaml(alertConfigDTO);
+        DetectionAlertConfigDTO alertConfig = new YamlDetectionAlertConfigTranslator(detectionConfigDAO).translate(detectionAlertYaml);
+
+        // Migrate all the subscribed anomaly functions
+        long currentTimestamp = System.currentTimeMillis();
+        List<Long> newDetectionIds = new ArrayList<>();
+        Map<Long, Long> vectorClocks = new HashMap<>();
+        List<Long> detectionIds = ConfigUtils.getLongs(alertConfig.getProperties().get(PROP_DETECTION_CONFIG_IDS));
+        for (long detectionId : detectionIds) {
```
jihaozh: can we do `migrateLegacyAnomalyFunction(anomalyFunctionId)` first for each legacy function. and then let
`YamlDetectionAlertConfigTranslator(detectionConfigDAO).translate(detectionAlertYaml)` to handle the vector clocks and detection config id list? So that we don't need extra logic to plug in vector clocks.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlDetectionAlertConfigTranslator.java
```diff
@@ -192,8 +195,17 @@ public DetectionAlertConfigDTO translate(Map<String,Object> yamlAlertConfig) {
...
-      detectionConfigIds.addAll(detectionNames.stream().map(detectionName ->  this.detectionConfigDAO.findByPredicate(
-          Predicate.EQ("name", detectionName)).get(0).getId()).collect(Collectors.toList()));
+      detectionConfigIds.addAll(detectionNames.stream().map(detectionName -> {
+        List<DetectionConfigDTO> detectionConfigDTOs = DAORegistry.getInstance().getDetectionConfigManager()
+            .findByPredicate(Predicate.EQ("name", detectionName));
+
+        if (!detectionConfigDTOs.isEmpty()) {
+          return detectionConfigDTOs.get(0).getId();
+        } else {
+          return DAORegistry.getInstance().getAnomalyFunctionDAO()
```
jihaozh: Is it possible that we don't access the anomaly function DAO here? This translator will be kept after the migration, i.e. after the anomaly function DAO deprecated. I'd prefer not to refer it in the new code.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionMigrationResource.java
akshayrai: This is a good idea! The issue is that the vector clocks in the translator set the time watermark to 0, while here, since we do not want to resend the anomalies we need to override the clock to the current timestamp. Let's sync up offline to see if we really need to set the timestamp to 0 in the first place.

akshayrai: I have set the vector to current timestamp and updated the PR.

akshayrai:
Thanks for the review @jihaozh. Brace yourself! The migration is coming.

PR #3725, opened 2019-01-19
Replace partition ranges with partitions

Jackie-Jiang:
Currently for partition based routing and pruning, we store a list
of ranges (IntRange), but does not support any operations on ranges.
Instead, storing a set of partitions (int) should be enough to serve
all use cases.

New behavior:
- For json format ColumnPartitionMetadata, replace "partitionRanges"
  with "partitions" key.
  E.g. "partitionRanges":"[0 0],[1 1]" -> "partitions":[0,1]
- For partition in segment metadata, replace strings with integers
  E.g. partitionValues = [0 0],[1 1] -> partitionValues = 0,1

Backward-compatibility:
- Add custom deserializer for ColumnPartitionMetadata to handle both
  new and legacy json format
- Add ColumnPartitionMetadata.extractPartitions() to handle both new
  and legacy config format
- Add ColumnPartitionMetadataTest


kishoreg:
What’s the benefit of changing this?  What happens if minion merges two segments with different ranges?

Jackie-Jiang:
> What’s the benefit of changing this? What happens if minion merges two segments with different ranges?

We treat range as single value right now, which is very confusing. When minion merges two segments, the partitions will also be merged. For example, segment 1 has partition 1 and 3, segment 2 has partition 3 and 5, then merged segment will have partition 1, 3 and 5.

kishoreg:
Good example. With your change, how can we store partition 1,3,5?

Jackie-Jiang:
@kishoreg Currently we are storing a list of ranges, which is quite redundant (and not supported when range covers more than 1 value). A list of integer partitions should be good enough, clear and simple. We should not have large number of partitions, so ranges do not buy us much. This change is mostly for simplicity and readability, which is not a must have, but a good to have IMO. With this, it's easier to add support for multi-partitions in the future.

Jackie-Jiang:
Also, if we use range 1-5 to replace [1,3,5], then when trying to purge segment with partition 2, we will get false negative (segment should be purged but not).

pinot-common/src/main/java/org/apache/pinot/common/metadata/segment/ColumnPartitionMetadata.java
```diff
@@ -41,77 +36,120 @@
...
   }
 
   /**
-   * Custom Json serializer for list of IntRange's.
+   * Helper method to extract partitions from configuration.
+   * <p>
+   * There are two format of partition strings:
+   * <ul>
+   *   <li>Integer format: e.g. {@code "0"}</li>
+   *   <li>Range format (legacy): e.g. {@code "[0 0]"}</li>
```
snleee: Can you add `TODO` comment to remove range format in the future?

pinot-common/src/main/java/org/apache/pinot/common/metadata/segment/ColumnPartitionMetadata.java
```diff
@@ -41,77 +36,120 @@
...
-        throws IOException {
-      jsonGenerator.writeString(ColumnPartitionConfig.rangesToString(value));
+  public static List<Integer> extractPartitions(List partitionList) {
+    int numPartitions = partitionList.size();
+    List<Integer> partitions = new ArrayList<>(numPartitions);
+    for (Object o : partitionList) {
+      String partitionString = o.toString();
+      if (partitionString.charAt(0) == '[') {
+        // Range format
+        partitions.add(Integer.parseInt(partitionString.substring(1, partitionString.indexOf(' '))));
```
snleee: I think that we need to check both numbers in the range. When a segment includes partition 0-9, I have seen something like `partitionRanges: [0 9]`, not `[0 0], [1 1] ... [9 9]`

pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
```diff
@@ -179,7 +180,8 @@ public synchronized void computeOnExternalViewChange(String tableName, ExternalV
   }
 
   /**
-   * Get partition id from segment Zk metadata. This helper function assumes there is only one column partition metadata.
+   * Get partition id from segment Zk metadata.
+   * TODO: This helper function assumes the segment is partitioned on at most one column, and contains only 1 partition.
```
snleee: I guess that removing the above assumption is `TODO` task? Can you update the comment what needs to be resolved?

pinot-common/src/main/java/org/apache/pinot/common/metadata/segment/ColumnPartitionMetadata.java
```diff
@@ -41,77 +36,120 @@
...
+        for (int i = 0; i < numPartitions; i++) {
+          partitions.add(jsonPartitions.get(i).asInt());
+        }
+      } else {
+        // Legacy format: "partitionRanges":"[0 0],[1 1]"
+        String partitionRanges = jsonMetadata.get(LEGACY_PARTITIONS_KEY).asText();
+        String[] partitionStrings = StringUtils.split(partitionRanges, LEGACY_PARTITION_DELIMITER);
+        partitions = new ArrayList<>(partitionStrings.length);
+        for (String partitionString : partitionStrings) {
+          partitions.add(Integer.parseInt(partitionString.substring(1, partitionString.indexOf(' '))));
```
snleee: Same as the above

pinot-core/src/main/java/org/apache/pinot/core/query/pruner/PartitionSegmentPruner.java
```diff
@@ -95,25 +94,19 @@ public boolean pruneSegment(FilterQueryTree filterQueryTree, Map<String, ColumnM
 
     // Leaf node
     String column = filterQueryTree.getColumn();
-    ColumnMetadata metadata = columnMetadataMap.get(column);
-    if (metadata == null) {
-      return false;
+    ColumnMetadata columnMetadata = columnMetadataMap.get(column);
+    // NOTE: should have already been pruned in DataSchemaSegmentPruner
```
snleee: This will change the behavior. I have checked that `DataSchemaSegmentPruner` is checking if all columns from the query exists in the query request. Checking column is enough for checking column metadata existence?

pinot-common/src/main/java/org/apache/pinot/common/metadata/segment/ColumnPartitionMetadata.java
Jackie-Jiang: Added

pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/PartitionAwareOfflineRoutingTableBuilder.java
Jackie-Jiang: Done.

pinot-core/src/main/java/org/apache/pinot/core/query/pruner/PartitionSegmentPruner.java
Jackie-Jiang: That is correct. But if we use this pruner alone (not with `DataSchemaSegmentPruner`), we should check whether ColumnMetadata exists.

PR #3727, opened 2019-01-19
Matches support in pql

kishoreg:
The first step towards supporting search analytics functionality in Pinot. 
syntax
 <colName> MATCHES ($queryString, $queryMatches)

Sample query
if col is a map 
- Select count(*) from T where col MATCHES ('k1=v1');

if col is a text 
- Select count(*) from T where col MATCHES ('fox AND brown', '{queryFormat:lucene}');

if col is a nested json
- Select count(*) from T where col MATCHES ('foo.bar=v1', '');

The second argument 'query options' is a json string and allows us to add more options later on. 

Looking for feedback on the syntax. 

For e.g. MySQL has the following syntax
`MATCH (col1,col2,...) AGAINST (expr [search_modifier])`

Amazon Athena
`WHERE json_array_contains(json_extract(users, '$.projects'), 'project2')`


kishoreg:
Closing since we have the full-text search now

PR #3728, opened 2019-01-21
Support for specifying Object Type in Schema

kishoreg:
Since Pinot now has the support to store byte[], it will be good to specify the ObjectType in the schema. This allows us to know how to parse the byte[] and provide additional functionality. Note, we already do this for HyperLogLog, TDigest,RawString etc. But this is spread all over the code and mostly used in Aggregation. Making this a first class citizen in schema allows us to support Map, JSON, etc.

This is part two of supporting search analytics.


pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
```diff
@@ -86,6 +86,10 @@
 
   @ConfigKey("virtualColumnProvider")
   protected String _virtualColumnProvider;
+  
+  //Complex type that can be constructed from raw bytes stored e.g. map, json, text
+  @ConfigKey("objectType")
+  protected String _objectType;
```
mayankshriv: Can this be an enum instead of a String?

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
```diff
@@ -98,15 +102,21 @@ public FieldSpec(String name, DataType dataType, boolean isSingleValueField) {
...
     this(name, dataType, isSingleValueField, DEFAULT_MAX_LENGTH, defaultNullValue);
   }
-
+  
   public FieldSpec(String name, DataType dataType, boolean isSingleValueField, int maxLength,
       @Nullable Object defaultNullValue) {
+    this(name, dataType, isSingleValueField, maxLength, defaultNullValue, null);
+  }
+  
+  public FieldSpec(String name, DataType dataType, boolean isSingleValueField, int maxLength,
```
mayankshriv: What's the issue with extending DataType to capture the object type?

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.apache.pinot.common.data.PinotObject;
+
+import com.beust.jcommander.internal.Lists;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+
+public class JSONObject implements PinotObject {
```
mayankshriv: The JSON library we use should already have a JSONObject? If so, we could use a facade?

pinot-common/src/main/java/org/apache/pinot/common/data/objects/MapObject.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.pinot.common.data.PinotObject;
+
+import com.beust.jcommander.internal.Lists;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class MapObject implements PinotObject {
+
+  private static ObjectMapper _MAPPER = new ObjectMapper();
+  Map<String, Object> _stringMap;
```
mayankshriv: Would using generics help to support typed maps?

pinot-common/src/main/java/org/apache/pinot/common/data/objects/TextObject.java
```diff
@@ -0,0 +1,35 @@
+package org.apache.pinot.common.data.objects;
+
+import java.util.List;
+
+import org.apache.pinot.common.data.PinotObject;
+
+import com.google.common.collect.Lists;
+
+public class TextObject implements PinotObject {
```
mayankshriv: What's the difference between Text and String?

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
kishoreg: it can be, was thinking of keeping it flexible so that new types can be added without changing Pinot code

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
kishoreg: yes, that was the initial solution. But wanted to keep the io layer (reader and writers) separate from the logical object layer. In other words, get<DataType> is supported by all the readers and writers.

Also, there are too many switch(dataType) in the code which needs to handle various these DataTypes and only thing we can do there is handle them as string or bytes without much value.

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
kishoreg: yes, we can change this to something that is fast and easy. This is just to show the concept. There are too many json libraries and found this link useful https://github.com/fabienrenaud/java-json-benchmark

pinot-common/src/main/java/org/apache/pinot/common/data/objects/TextObject.java
kishoreg: Text is a long string. It can give us some hints such as don't do dictionary encoding. More like TEXT vs VARCHAR in MySQL

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
```diff
@@ -98,15 +102,21 @@ public FieldSpec(String name, DataType dataType, boolean isSingleValueField) {
...
+      @Nullable Object defaultNullValue, @Nullable String objectType) {
```
sunithabeeram: Shouldn't this method actually set _objectType?

pinot-common/src/main/java/org/apache/pinot/common/data/PinotObject.java
```diff
@@ -0,0 +1,43 @@
...
+ * Common interface for complex Object types such as HyperLogLog, Map, JSON etc
+ */
+public interface PinotObject {
+
+  void init(byte[] bytes);
+
+  byte[] toBytes();
+
+  /**
+   * @return
```
sunithabeeram: Please add appropriate javadocs or remove the comment. (Same for the method below)

pinot-common/src/main/java/org/apache/pinot/common/data/PinotObject.java
```diff
@@ -0,0 +1,43 @@
...
+   */
+  List<String> getPropertyNames();
+  /**
+   * Asks for the 
+   * @param fieldName
+   * @return
+   */
+  Object getProperty(String fieldName);
```
sunithabeeram: Might be better to call the param propertyName.

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
```diff
@@ -0,0 +1,83 @@
...
+    try {
+      return _MAPPER.writeValueAsBytes(_jsonNode);
+    } catch (JsonProcessingException e) {
+      return "{}".getBytes();
+    }
+  }
+
+  @Override
+  public List<String> getPropertyNames() {
+    List<String> fields = Lists.newArrayList();
```
sunithabeeram: Might be good to materialize/cache the properties.

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
```diff
@@ -0,0 +1,83 @@
...
+      String fieldName = (String) iterator.next();
+      fields.add(fieldName);
+    }
+    return fields;
+  }
+
+  @Override
+  public Object getProperty(String fieldName) {
+    JsonNode jsonNode = _jsonNode.get(fieldName);
+    if (jsonNode.isArray()) {
```
sunithabeeram: The special casing here isn't clear to me.

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
```diff
@@ -0,0 +1,83 @@
...
+      Iterator<JsonNode> iterator = jsonNode.iterator();
+      List<String> list = new ArrayList<String>();
+      while (iterator.hasNext()) {
+        list.add(iterator.next().asText());
+      }
+      return list;
+    } else {
+      return jsonNode.asText();
```
sunithabeeram: Why are we returning this as text/string?

pinot-common/src/main/java/org/apache/pinot/common/request/AggregationInfo.java
```diff
@@ -24,29 +24,40 @@
  */
```
sunithabeeram: A lot of files below have just formatting changes. Would be great to exclude them from the PR - the changes also seem to be undoing some of the imports organization that was done in a recent PR.

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
sunithabeeram: Is there a plan to add a PinotObjectFactory that constructs the appropriate objects based on this type?

Its not clear how String vs enum really helps; enum gives some enforcement/bounds on the type.

pinot-common/src/main/antlr4/org/apache/pinot/pql/parsers/PQL2.g4
```diff
@@ -77,6 +77,7 @@ predicate:
   | betweenClause                         # BetweenPredicate
   | isClause                              # IsPredicate
   | regexpLikeClause                      # RegexpLikePredicate
+  | matchesClause                         # MatchesPredicate
```
sunithabeeram: <nit> Would be good to keep this change to just object type changes and bring in the pql changes in a different PR</nit>

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
kishoreg: PinotObject itself has those methods, init and toBytes. The only downside with Enum is every time users (external to Pinot Dev) want to add a new Object type, we need to update the enum. We can definitely start with enum and change this later.

pinot-common/src/main/java/org/apache/pinot/common/data/PinotObject.java
kishoreg: will do.

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
kishoreg: yes, will fix it.

pinot-common/src/main/java/org/apache/pinot/common/data/PinotObject.java
kishoreg: will do

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
kishoreg: yes. there are lots of optimizations we can do in the initial set of ObjectType. I am not too happy about the json library as well. I will leave those optimizations for another PR.

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
kishoreg: Its similar to single value vs multi value. If its array type, we need to know each value in the array.

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
kishoreg: could not find a method to get the underlying object in this json library. As I mentioned in another comment, we have to do another iteration on the Pinot Object implementation.

pinot-common/src/main/java/org/apache/pinot/common/request/AggregationInfo.java
kishoreg: thats coming from thrift auto-gen files, not sure how to control them. See the other PR to remove the auto-gen files from source.

kishoreg:
Thanks @sunithabeeram  and @mayankshriv for quick review. I should have mentioned this earlier, this PR is to get high-level overview of the concept. Once we get agree on that, we can start looking into the actual implementation of PinotObject. I will push another PR that uses PinotObject, that will give a better idea on the benefits of PinotObject Type.

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
mcvsubbu: I am not sure how/where the serialization (and deserialization) contract is defined between Pinot and the entity that ingests data into Pinot (and the one that reads it, and specifies queries against it).

I agree with using enums, so that we can start with some small number of supported object types, I suppose

pinot-common/src/main/java/org/apache/pinot/common/data/PinotObject.java
```diff
@@ -0,0 +1,43 @@
...
+public interface PinotObject {
+
+  void init(byte[] bytes);
+
+  byte[] toBytes();
+
+  /**
+   * @return
+   */
+  List<String> getPropertyNames();
```
mcvsubbu: I guess property names and values are specific to the object type? Some objects may have properties specific to the instance (e.g. for json objects it will be "keys", I suppose) whereas others may have properties specific to the object definition?

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
mcvsubbu: May be useful to rename this class as PinotJSONObject. Every JSON library (and there are several) probably has its own JSONObject :)

pinot-common/src/main/java/org/apache/pinot/common/data/objects/MapObject.java
mcvsubbu: PinotMapObject?

pinot-common/src/main/java/org/apache/pinot/common/data/objects/MapObject.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public void init(byte[] bytes) {
+    try {
+      _stringMap = _MAPPER.readValue(bytes, Map.class);
+    } catch (IOException e) {
+      _stringMap = Collections.emptyMap();
```
mcvsubbu: Not sure how this is used, but I am noting that we may not want to swallow exceptions all the time. If (say) hadoop job is trying to ingest objects. we may want to return an exception right away instead of replacing the value with our own choice. On the other hand, while processing a query, perhaps returning an empty string is ok (but ideally we should just discard the docid?)

I assume that the answer to my question about serialization contract is that the ingesting entity has to use Pinot library (which we shall preserve backward compatibility, with appropriate versioning etc.). In that case should these definitions move to pinot-api package?

pinot-common/src/main/java/org/apache/pinot/common/data/objects/TextObject.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  public List<String> getPropertyNames() {
+    return _FIELDS;
+  }
+
+  @Override
+  public Object getProperty(String field) {
+    // TODO Auto-generated method stub
```
mcvsubbu: I guess we choose utf-8 as the encoding scheme, or do we define the encoding scheme as a property?

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
kishoreg: I didn't get your question, the data coming into Pinot will always be byte[]. Looking at the fieldspec.objectType in schema , Pinot can construct the PinotObject. Now, this can be used in multiple places.

- Index Creation - we can now index nested fields within this complex object. thats why I added getProperties() and getPropertyValue(). 
- querying - we can now support various operations on these objects. For e.g. select extract(user, 'user.firstName') can extract a property from a PinotObject. Another example would be to aggregate objects.

Hope that makes sense. 

It might help if we try to see that we already support HyperLogLog, QDigest & TDigest but without having a concept of PinotObject. This PR is trying to standardize on how we create a custom object from a byte[]. Having an interface and enforcing certain methods make extending the functionality easily.

I will make it enum.

pinot-common/src/main/java/org/apache/pinot/common/data/objects/JSONObject.java
kishoreg: will do

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
mcvsubbu: Let us say the object was a map (as per your example). Presumably, the entity that ingests data into Pinot has serialized a map into bytes using some serde. That serde method better be the same that Pinot uses to deserialize while fielding a query, or while returning result of a query (may have to deserialize, and serialize back to deserialize again). Did I understand it right?

pinot-common/src/main/java/org/apache/pinot/common/data/objects/MapObject.java
kishoreg: I think we should handle it gracefully and have an error counter. This is what we do currently in real time ingestion. If we cannot convert or parse, we set the default value and move on.

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
kishoreg: Yes, that's right. Note that this is no different from HyperLogLog and TDigest.

PR #3730, opened 2019-01-22
Update versions of packages to move to jersey 2.28

sunithabeeram:
Pick up the latest version of Jersey so that we can move to latest version of Jackson.

Other changes were required (esp around javax.validation) as swagger and jersey required different versions - I've pinned the higher version in the top level pom and that seems to let the compilation go through.

maven-shade-plugin bump up was required as the MojoExecutor had IllegalArgumentException in pinot-common package that was failing the build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,11 +111,10 @@
     <!-- jfim: for Kafka 0.9.0.0, use zkclient 0.7 -->
     <kafka.version>0.9.0.1</kafka.version>
     <zkclient.version>0.7</zkclient.version>
-    <!-- TODO: bump up the jackson version once we update jersey version to 2.28 or later -->
-    <jackson.version>2.8.3</jackson.version>
+    <jackson.version>2.9.8</jackson.version>
     <async-http-client.version>1.9.21</async-http-client.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -111,11 +111,10 @@
     <!-- jfim: for Kafka 0.9.0.0, use zkclient 0.7 -->
     <kafka.version>0.9.0.1</kafka.version>
     <zkclient.version>0.7</zkclient.version>
-    <!-- TODO: bump up the jackson version once we update jersey version to 2.28 or later -->
-    <jackson.version>2.8.3</jackson.version>
+    <jackson.version>2.9.8</jackson.version>
     <async-http-client.version>1.9.21</async-http-client.version>
-    <jersey.version>2.23</jersey.version>
-    <swagger.version>1.5.10</swagger.version>
+    <jersey.version>2.28-RC4</jersey.version>
```
snleee: `RC` versions are release candidates, not the official releases. If `2.27` version contained the fix that we need, I would use that.

mcvsubbu: Our best bet is with 2.28 now. Hopefully we will get a release soon that we can incorporate.

PR #3731, opened 2019-01-22
Ensures that pinot-broker.log logged queries are on single lines

hufman:


sunithabeeram:
It might be good to add/update a test where the query/where-clause does have newlines - that way, we can ensure the code paths are rightly handled. (We can't add assertions for logging)

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -120,7 +120,8 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
     requestStatistics.setRequestArrivalTimeMillis(System.currentTimeMillis());
 
     String query = request.get(PQL).asText();
-    LOGGER.debug("Query string for request {}: {}", requestId, query);
+    String queryLoggable = StringUtils.substring(query.replace("\n", "\\n"), 0, _queryLogLength);
```
kishoreg: extract this to a method and invoke only when needed. Looks like we log it only in debug / error cases. 

Also, better to trim first and then replace.

npawar:

PR #3734, opened 2019-01-23
Make delete table API async

jackjlli:
Currently the method deleteTable is a synchronized API. And there's no log to measure how long it takes to complete. This PR makes delete table API async so that the response can be returned fast. 

Modified the code as Subbu suggests:
Sync steps:
* Removing resource from broker idealstate
* Removing table resource from helix

Async steps:
* Removing segments
* Removing table config
* Removing replica group partition assigement


jackjlli:
Right now there's no need to make it changes. I'm closing this PR now.

PR #3741, opened 2019-01-24
Add integration test for SegmentStatusChecker periodic task

npawar:
There are no comprehensive integration tests for the controller periodic tasks. This PR is the first of many such integration tests.

### DIFF
--- a/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/controller/periodic/tasks/SegmentStatusCheckerIntegrationTest.java
+++ b/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/controller/periodic/tasks/SegmentStatusCheckerIntegrationTest.java
@@ -58,7 +58,7 @@ public class SegmentStatusCheckerIntegrationTest extends BaseClusterIntegrationT
   private String errorOfflineTable = "table4_OFFLINE";
   private String realtimeTableErrorState = "table5_REALTIME";
 
-  private static final int SEGMENT_STATUS_CHECKER_INITIAL_DELAY_SECONDS = 40;
+  private static final int SEGMENT_STATUS_CHECKER_INITIAL_DELAY_SECONDS = 60;
 
   @BeforeClass
   public void setUp() throws Exception {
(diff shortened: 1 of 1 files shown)


pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/controller/periodic/tasks/SegmentStatusCheckerIntegrationTest.java
```diff
@@ -0,0 +1,259 @@
...
+ */
+public class SegmentStatusCheckerIntegrationTest extends BaseClusterIntegrationTestSet {
+
+  private String emptyTable = "table1_OFFLINE";
+  private String disabledOfflineTable = "table2_OFFLINE";
+  private String basicOfflineTable = "table3_OFFLINE";
+  private String errorOfflineTable = "table4_OFFLINE";
+  private String realtimeTableErrorState = "table5_REALTIME";
+
+  private static final int SEGMENT_STATUS_CHECKER_INITIAL_DELAY_SECONDS = 60;
```
jackjlli: Could we shorten the initial delay time here? Waiting for 60 seconds is a bit long for a test and will affect the total time to run.

npawar: I had started with 40s, but that was not enough time on travis infrastructure to setup the brokers and servers. That was making the tests fail on travis, even though locally they succeeded. Hence bumped it up.
Locally it takes about 35-40s for setup to be ready. I could try making it 50, but I'm afraid the tests might fail unpredictably

sunithabeeram: If starting up the servers is an issue, then may be we can update the code so that they are done once (or just a few times for the tests) and the state cleared between tests. We don't have to do this  now, but something we should consider/work towards.

npawar: Yes, that's a good idea. We could also maybe have common zk, controllers, brokers as well? The only thing changing across tests is the kind of tables created, the data pushed and the check/queries run

mcvsubbu:
Instead of 10s wait for the setup, can we wait until the first run is completed? It may (and hopefully it will) complete in less than 10s?

PR #3746, opened 2019-01-25
Fix a bug in PinotSegmentRestletResource.getInstanceToSegmentsMap()

Jackie-Jiang:
The value for key "segments" should be a serialized map string instead of a json object.
This is for backward-compatible.


sunithabeeram:
Please consider adding a unit test so we don't regress on this in the future.

PR #3748, opened 2019-01-25
Use the PinotFS instead of local file system to open a segment file t…

chenboat:
…o extract metadata because the segment may not reside on local disk. @jenniferdai @jamesyfshao

https://github.com/apache/incubator-pinot/issues/2539

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -369,10 +369,9 @@ public class PinotLLCRealtimeSegmentManager {
     String segmentName = committingSegmentDescriptor.getSegmentName();
     String segmentLocation = committingSegmentDescriptor.getSegmentLocation();
     URI segmentFileURI = ControllerConf.getUriFromPath(segmentLocation);
-    URI baseDirURI = ControllerConf.getUriFromPath(_controllerConf.getDataDir());
     URI tableDirURI = ControllerConf.getUriFromPath(StringUtil.join("/", _controllerConf.getDataDir(), tableName));
     URI uriToMoveTo = ControllerConf.getUriFromPath(StringUtil.join("/", tableDirURI.toString(), segmentName));
-    PinotFS pinotFS = PinotFSFactory.create(baseDirURI.getScheme());
+    PinotFS pinotFS = getPinotFS();
 
     if (!isConnected() || !isLeader()) {
       // We can potentially log a different value than what we saw ....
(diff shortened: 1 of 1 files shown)


chenboat:
> I would prefer that we avoid getting the segment file altogether.
> 
> The server uploads the segment metadata file as a part of the COMMIT message, and that gets passed around in the CommittingSegmentDescriptor.
> 
> We should avoid pulling the segment metadata file out of the tarball.

Thank you for the suggestion. That will simplify the workflow. I am working on that.

npawar:
I did some evaluation. You can introduce the new metadata upload
**Step 1**
Add a new endpoint in LLCSegmentCompletionHandler as
```
  @POST
  @Path(SegmentCompletionProtocol.MSG_TYPE_COMMIT_END_METADATA)
  @Consumes(MediaType.MULTIPART_FORM_DATA)
  public String segmentCommitEndWithMetadata(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE_ID) String instanceId,
      @QueryParam(SegmentCompletionProtocol.PARAM_SEGMENT_NAME) String segmentName,
      @QueryParam(SegmentCompletionProtocol.PARAM_SEGMENT_LOCATION) String segmentLocation,
      @QueryParam(SegmentCompletionProtocol.PARAM_OFFSET) long offset,
      @QueryParam(SegmentCompletionProtocol.PARAM_MEMORY_USED_BYTES) long memoryUsedBytes,
      @QueryParam(SegmentCompletionProtocol.PARAM_BUILD_TIME_MILLIS) long buildTimeMillis,
      @QueryParam(SegmentCompletionProtocol.PARAM_WAIT_TIME_MILLIS) long waitTimeMillis,
      @QueryParam(SegmentCompletionProtocol.PARAM_ROW_COUNT) int numRows,
      @QueryParam(SegmentCompletionProtocol.PARAM_SEGMENT_SIZE_BYTES) long segmentSizeBytes,
      FormDataMultiPart metadataFiles) {
```
In the `FormDataMultiPart metadataFiles` multipart, we expect `metadata.properties` and `column.meta`. Extract those and copy them to a local temp folder. You can use that folder to create a new SegmentMetadataImpl (the way it's done in `PinotLLCRealtimeSegmentManager::extractSegmentMetadata`



**Step 2**
In `ServerSegmentCompletionProtocolHandler`, introduce a method `segmentCommitEndWithMetadata(SegmentCompletionProtocol.Request.Params params,
      File metadataPropertiesFile, File creationMetaFile)`. The url that gets constructed needs to change to POST. These 2 files need to be added as entities (similar to the way segment tar file is added in `getUploadFileRequest`)

**Step 3**
In `LLRealtimeSegmentDataManager::doSplitCommit`, change call from `segmentCommitEnd` to `segmentCommitEndWithMetadata`. The metadata files for this call can be extracted from the segmentTarFile.

**Step 4**
Conditions can be added, which check if the metadata was extracted successfully. If not, download the metadata from the segment store. Ensure path coming from non-split commit upload also works

@chenboat @mcvsubbu

mcvsubbu:
Thanks, @npawar for evaluating this. Here are things I will add:

step 2 should be made configurable on the server side (i.e. only if a config is enabled, we should upload metadata)

step 1a should be to modify LLCSegmentCompletionHandlers.commit to include checking whether two more parts are included, and if so, use them as metadataPropertiesFile and creationMetaFile)

In Step 2 we need to handle both segmentCommit as well as segmentCommitEndWithMetadata under the same config mentioned before

In step 3 call method depending on config

chenboat:
Thanks @npawar and @mcvsubbu for the detailed evaluation. So the server sends the extracted meta data files to the controller -- that is a good idea. I was thinking of servers sending structured data fields to the controller which seemed involved too much data transformation. I will send a revised PR alone the line you described.

mcvsubbu:
In that case, please discard this PR. 
You can send multiple PRs for the implementation, so that it is easier to review.
We believe it should be straight-forward, but please reach out on the apache slack channel if you need any help.

PR #3755, opened 2019-01-28
[TE] detection - report missing anomaly endpoint for pipeline v2

jihaozh:
The endpoint to report a missing anomaly in the new detection pipeline.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
@@ -440,7 +440,7 @@ public class DetectionResource {
   @POST
   @Path(value = "/report-anomaly/{detectionConfigId}")
   @ApiOperation("Report a missing anomaly for a detection config")
-  public void createUserAnomaly(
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
```diff
@@ -427,4 +423,54 @@ public Response detectionReplay(
...
+   * @param metricUrn the metric urn of the anomaly
+   * @param feedbackType anomaly feedback type
+   * @param comment anomaly feedback comment (optional)
+   * @throws IllegalArgumentException if the anomaly function id cannot be found
+   * @throws IllegalArgumentException if the anomaly cannot be stored
+   */
+  @POST
+  @Path(value = "/report-anomaly/{detectionConfigId}")
+  @ApiOperation("Report a missing anomaly for a detection config")
+  public void createUserAnomaly(
```
akshayrai: Maybe return a response with the created anomaly ID so that the caller knows the status of the api call.

jihaozh: good idea. changed.

PR #3767, opened 2019-01-30
[TE] detection - yaml translation/migration changes

jihaozh:
- Do not  trigger tune for migrated anomaly functions
- Default tuning window

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
@@ -24,24 +24,6 @@ import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
-import org.apache.pinot.thirdeye.api.TimeGranularity;
-import org.apache.pinot.thirdeye.datalayer.dto.DatasetConfigDTO;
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
```diff
@@ -158,10 +159,23 @@
...
   private static final String PROP_NAME = "name";
+  private static final String DEFAULT_TIMEZONE = "America/Los_Angeles";
   private static final String DEFAULT_BASELINE_PROVIDER_YAML_TYPE = "RULE_BASELINE";
 
   private static final DetectionRegistry DETECTION_REGISTRY = DetectionRegistry.getInstance();
-  private static final Map<String, String> DETECTOR_TO_BASELINE = ImmutableMap.of("ALGORITHM", "ALGORITHM_BASELINE");
+  static {
+    // do not tune for alerts migrated from legacy anomaly function.
+    DetectionRegistry.registerComponent("com.linkedin.thirdeye.detection.components.AdLibAlertFilter",
+        "MIGRATED_ALGORITHM_FILTER");
```
akshayrai: 1. Why should tuning be disabled for migrated functions? Can you provide some context?
2. Do we really need a new MIGRATED tag? When will we update them later?

jihaozh: Yes. Otherwise, all previously manually set parameters will be lost (override by tunable module) during migration and impacts detection performance.

PR #3770, opened 2019-01-31
Remove unnecessary init from crypter

jenniferdai:
* Crypter is initialized in controller/server starter, so we do not need to init it here. The problem with initing here is that there can be a possible race condition where we are trying to decrypt another segment but the keys will have disappeared since we have now introduced concurrent segment downloads


pinot-server/src/main/java/org/apache/pinot/server/starter/helix/SegmentFetcherAndLoader.java
```diff
@@ -200,7 +200,6 @@ private String downloadSegmentToLocal(@Nonnull String uri, PinotCrypter crypter,
       SegmentFetcherFactory.getInstance().getSegmentFetcherBasedOnURI(uri).fetchSegmentToLocal(uri, tempDownloadFile);
       if (crypter != null) {
         // TODO: We should not need to initialize crypter each time, instead Factory should have an initialized version ready.
```
snleee: Is this comment still valid?

PR #3783, opened 2019-02-02
Add methods toPrettyJsonString() and toSingleLineJsonString() to Schema

Jackie-Jiang:
- For human-read API, use pretty string
- For other purpose, use compact single-line string

Fix the issue of storing pretty string in ZK (consume more space and less readable)


Jackie-Jiang:
@sunithabeeram We support both format, and it's mainly for saving space & readability.

PR #3786, opened 2019-02-03
[PINOT-5] make BitmapInvertedIndexTest Windows compatible

xueyumusic:
BitmapInvertedIndexTest failed on windows caused by two places, one is that `resourceUrl` of `AVRO_FILE_PATH` contains '%5c' which was '\\' originally, the other is `Files.setPosixFilePermissions` which throws exception on windows


kishoreg:
Thanks for the patch. Are these the only tests that fail on windows?

xueyumusic:
Hi, @kishoreg , thanks for review. In pinot-core there are a batch of non-Windows-compatible tests. I tried to fix part of them in another PR. There are still several tests left, e.g. `PinotDataBufferTest(I suspect about xerial mmap lib)`, `CountStarTreeV2Test(which uses '*' in file name)`, and `FixedByteWidthRowColDataFileWriterTest(which test special chars)` etc..

PR #3788, opened 2019-02-04
Adding explicit message to alert users to empty input files during segment generation

jenniferdai:


pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java
```diff
@@ -119,7 +119,7 @@ public void init(SegmentGeneratorConfig config, SegmentCreationDataSource dataSo
     this.config = config;
     this.createStarTree = config.isEnableStarTreeIndex();
     recordReader = dataSource.getRecordReader();
-    Preconditions.checkState(recordReader.hasNext(), "No record in data source");
+    Preconditions.checkState(recordReader.hasNext(), "No records found: You must have at least one record in each of your input files");
```
mcvsubbu: Can you remove the pre-conditions check here, and add an explicit error message with as much details as possible (input file name, full path, whatever). And then throw the exception with the same message?

mcvsubbu: Alternatively, check in hadoop job before we call this method, and error-log + exception like i described before. Then the pre-condition check here is totally valid

PR #3793, opened 2019-02-05
Fixing jackson dependency issues with swagger

jenniferdai:
* Swagger UI pulls in an old version of Jackson, failing our current build. 
* Checked snapshot build to make sure dependencies work

### DIFF
--- a/pinot-server/pom.xml
+++ b/pinot-server/pom.xml
@@ -216,6 +216,12 @@
     <dependency>
         <groupId>org.webjars</groupId>
         <artifactId>swagger-ui</artifactId>
+      <exclusions>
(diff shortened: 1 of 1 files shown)


sunithabeeram:
Lets wait for tests to pass.

mcvsubbu:
Did u test swagger? Our tests do not cover access via swagger

sunithabeeram:
> Did u test swagger? Our tests do not cover access via swagger

@mcvsubbu I did add basic tests for console in https://github.com/apache/incubator-pinot/pull/3740. While it won't test rendering, we'll know if something is horribly broken.

PR #3798, opened 2019-02-06
[TE] Include a detection description field to describe the alert

akshayrai:


akshayrai:
@jihaozh, the bean and translators are already updated with the description field. I removed the references from the tests and rca

PR #3805, opened 2019-02-08
[TE] dashboard link point to yaml if the anomaly is detected by new pipeline

jihaozh:
dashboard link point to yaml if the anomaly is detected by the new pipeline


harleyjj:
Thank you for fixing this!

PR #3812, opened 2019-02-09
[TE] Refactor detection & preview API for better debugging

akshayrai:


jihaozh:
Looks like there are some issues on detection yaml translation caused by this PR. I'll submit a follow up PR.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/validators/DetectionConfigValidator.java
```diff
@@ -0,0 +1,111 @@
...
+
+
+public class DetectionConfigValidator extends ConfigValidator {
+
+  private static DetectionConfigValidator INSTANCE;
+
+  private final DataProvider provider;
+  private final DetectionPipelineLoader loader;
+
+  public static DetectionConfigValidator getInstance() {
```
jihaozh: Let's not overuse Singleton pattern.

Since there is no access to the resource will be requested from multiple, disparate parts of the system and the initialization of the object is not heavy.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/validators/ConfigValidator.java
```diff
@@ -55,7 +55,7 @@ public boolean validateYAMLConfig(String yamlConfig) throws ValidationException
     try {
       Map<String, Object> yamlConfigMap = (Map<String, Object>) YAML.load(yamlConfig);
```
jihaozh: Looks like this yamlConfig is loaded and thrown away immediately. It should be fine to let the client code to do validate yaml syntax, cause it has to load the yaml anyways.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -260,37 +237,51 @@ public Response createYamlAlert(@ApiParam(value =  "a json contains both notific
...
-    } catch (Exception e){
-      return Response.status(Response.Status.BAD_REQUEST).entity(ImmutableMap.of("message", "detection yaml parsing error, " + e.getMessage())).build();
+      detectionConfigId = createDetectionPipeline(payload, startTime, endTime);
+    } catch (ValidationException e) {
+      LOG.warn("Validation error while creating detection pipeline with payload " + payload, e);
+      responseMessage.put("message", "Validation Error! " + e.getMessage());
+      return Response.serverError().entity(responseMessage).build();
+    } catch (Exception e) {
+      LOG.error("Error creating subscription group with payload " + payload, e);
+      responseMessage.put("message", "Failed to create the subscription group. Reach out to the ThirdEye team.");
```
jihaozh: Should be detection pipeline, not subscription group. And this error checking logic is missing in the /create-alert endpoint.

PR #3813, opened 2019-02-09
Refactor Hadoop Jobs

Jackie-Jiang:
- Make Hadoop Jobs easier to extend
- Add interface ControllerRestApi to provide APIs for config fetching and segment pushing
- Add BaseSegmentJob abstract class for common segment related jobs
- In SegmentCreationJob, allow plugable mapper class and additional job properties
- In SegmentCreationMapper, allow additional segment generator configs
- Tested with Hadoop integration test


pinot-hadoop/src/main/java/org/apache/pinot/hadoop/utils/PushLocation.java
```diff
@@ -18,13 +18,25 @@
...
+
+
 public class PushLocation {
   private final String _host;
   private final int _port;
 
-  private PushLocation(PushLocationBuilder pushLocationBuilder) {
-    _host = pushLocationBuilder._host;
-    _port = pushLocationBuilder._port;
+  public PushLocation(String host, int port) {
```
snleee: This change may break hadoop jobs for people out side if they do something similar to PBNJ (importing pinot-hadoop jar for their hadoop job project). For now, I think that this is fine; however, it's going to be hard for us to change the constructor in very near future. Once we publish our jar files to public maven repo, this kind of change should happen in multiple steps. (Add constructor, add deprecated annotation, remove them with the major version bump).

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/HadoopSegmentBuildPushOfflineClusterIntegrationTest.java
```diff
@@ -155,13 +155,13 @@ private void generateAndPushSegmentsFromHadoop()
     properties.setProperty(JobConfigConstants.PUSH_TO_PORT, getDefaultControllerConfiguration().getControllerPort());
 
     // Run segment creation job
-    SegmentCreationJob creationJob = new SegmentCreationJob("TestSegmentCreation", properties);
+    SegmentCreationJob creationJob = new SegmentCreationJob(properties);
```
snleee: I think that it's better to expose `JobName` to be configurable. It's fine to include that config within `properties` and read it back along with other config instead of having it as an input variable for constructor.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/BaseSegmentJob.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.pinot.common.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public abstract class BaseSegmentJob extends Configured {
+  protected final Logger _logger = LoggerFactory.getLogger(getClass());
```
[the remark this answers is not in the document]
Jackie-Jiang: LOGGER is static final, while _logger is non-static. The reason why this one is non-static is because we want the name shows as the actual class instead of the abstract class.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -35,328 +37,261 @@
...
+    _outputDir = Preconditions.checkNotNull(getPathFromProperty(JobConfigConstants.PATH_TO_OUTPUT));
+    _stagingDir = new Path(_outputDir, UUID.randomUUID().toString());
+    _rawTableName = Preconditions.checkNotNull(_properties.getProperty(JobConfigConstants.SEGMENT_TABLE_NAME));
 
-  private String[] _hosts;
-  private int _port;
+    // Optional
+    _depsJarDir = getPathFromProperty(JobConfigConstants.PATH_TO_DEPS_JAR);
+    _schemaFile = getPathFromProperty(JobConfigConstants.PATH_TO_SCHEMA);
+    _defaultPermissionsMask = _properties.getProperty(JobConfigConstants.DEFAULT_PERMISSIONS_MASK);
```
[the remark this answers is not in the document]
Jackie-Jiang: Yes, correct. We'll do null check before applying it.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/utils/PushLocation.java
Jackie-Jiang: Good point. Once we go with official release, these kind of changes need to be out in multiple steps.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/HadoopSegmentBuildPushOfflineClusterIntegrationTest.java
Jackie-Jiang: Job name is inferred in the job class. Other properties should be passed through `properties`.

PR #3815, opened 2019-02-11
[TE] Move wrapper classes under api to common

akshayrai:
Today, the _api_ and _common_ packages share utilities across the thirdeye modules and there is not much distinction between them. This PR tries to merge the two packages into one.

Going forward, we will start to use the _api_ folder/package to document all the internal and external facing apis which will help us bring some code structure into the endpoints supported by ThirdEye.


jihaozh:
Thanks for the effort. But afaik, anomaly detection libs have heavy dependencies on thirdeye.api package.Those references will be broken by this change.

akshayrai:
I will update their libraries.

PR #3819, opened 2019-02-12
Refactor periodic task

Jackie-Jiang:
Refactor periodic task to fix the following issues：
- PinotTaskManager.scheduleTasks() has no effect on non-leader controller, but returns result from previous run
  The reason for this is that, we store states for each run as member variables in the periodic task, which mixed the concept of task and run
- After calling stop() (lose leadership), when re-gaining leadership, no start() method is called to set up the environment
- Potential race condition

Replace init() with start() and call it whenever the leadership is gained
Move the basic control methods for periodic task (start(), run(), stop()) into BasePeriodicTask
Keep table level methods in ControllerPeriodicTask
For per-run states, add context generic type to pass them between processing multiple tables

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/SegmentStatusChecker.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/SegmentStatusChecker.java
@@ -39,7 +39,7 @@ import org.slf4j.LoggerFactory;
  * Manages the segment status metrics, regarding tables with fewer replicas than requested
  * and segments in error state.
  */
-public class SegmentStatusChecker extends ControllerPeriodicTask {
+public class SegmentStatusChecker extends ControllerPeriodicTask<SegmentStatusChecker.Context> {
   private static final Logger LOGGER = LoggerFactory.getLogger(SegmentStatusChecker.class);
   private static final int MaxOfflineSegmentsToLog = 5;
   public static final String ONLINE = "ONLINE";
@@ -50,10 +50,6 @@ public class SegmentStatusChecker extends ControllerPeriodicTask {
   // log messages about disabled tables atmost once a day
   private static final long DISABLED_TABLE_LOG_INTERVAL_MS = TimeUnit.DAYS.toMillis(1);
   private long _lastDisabledTableLogTimestamp = 0;
-  private boolean _logDisabledTables;
-  private int _realTimeTableCount;
-  private int _offlineTableCount;
-  private int _disabledTableCount;
 
   /**
    * Constructs the segment status checker.
@@ -69,58 +65,50 @@ public class SegmentStatusChecker extends ControllerPeriodicTask {
   }
 
   @Override
-  public void initTask() {
+  protected void setUpTask() {
     LOGGER.info("Initializing table metrics for all the tables.");
     setStatusToDefault();
   }
 
   @Override
-  protected void preprocess() {
-    _realTimeTableCount = 0;
-    _offlineTableCount = 0;
-    _disabledTableCount = 0;
-
+  protected Context preprocess() {
+    Context context = new Context();
     // check if we need to log disabled tables log messages
     long now = System.currentTimeMillis();
     if (now - _lastDisabledTableLogTimestamp >= DISABLED_TABLE_LOG_INTERVAL_MS) {
-      _logDisabledTables = true;
+      context._logDisabledTables = true;
       _lastDisabledTableLogTimestamp = now;
-    } else {
-      _logDisabledTables = false;
+    }
+    return context;
+  }
+
+  @Override
+  protected void processTable(String tableNameWithType, Context context) {
+    try {
+      updateSegmentMetrics(tableNameWithType, context);
+    } catch (Exception e) {
+      LOGGER.error("Caught exception while updating segment status for table {}", tableNameWithType, e);
+      // Remove the metric for this table
+      resetTableMetrics(tableNameWithType);
     }
   }
 
   @Override
-  protected void processTable(String tableNameWithType) {
-    updateSegmentMetrics(tableNameWithType);
-  }
-
-  @Override
-  protected void exceptionHandler(String tableNameWithType, Exception e) {
-    LOGGER.error("Caught exception while updating segment status for table {}", tableNameWithType, e);
-    // Remove the metric for this table
-    resetTableMetrics(tableNameWithType);
-  }
-
-  @Override
-  protected void postprocess() {
-    _metricsRegistry.setValueOfGlobalGauge(ControllerGauge.REALTIME_TABLE_COUNT, _realTimeTableCount);
-    _metricsRegistry.setValueOfGlobalGauge(ControllerGauge.OFFLINE_TABLE_COUNT, _offlineTableCount);
-    _metricsRegistry.setValueOfGlobalGauge(ControllerGauge.DISABLED_TABLE_COUNT, _disabledTableCount);
+  protected void postprocess(Context context) {
+    _controllerMetrics.setValueOfGlobalGauge(ControllerGauge.REALTIME_TABLE_COUNT, context._realTimeTableCount);
+    _controllerMetrics.setValueOfGlobalGauge(ControllerGauge.OFFLINE_TABLE_COUNT, context._offlineTableCount);
+    _controllerMetrics.setValueOfGlobalGauge(ControllerGauge.DISABLED_TABLE_COUNT, context._disabledTableCount);
   }
 
   /**
    * Runs a segment status pass over the given table.
    * TODO: revisit the logic and reduce the ZK access
-   *
-   * @param tableNameWithType
    */
-  private void updateSegmentMetrics(String tableNameWithType) {
-
+  private void updateSegmentMetrics(String tableNameWithType, Context context) {
     if (TableNameBuilder.getTableTypeFromTableName(tableNameWithType) == TableType.OFFLINE) {
-      _offlineTableCount++;
+      context._offlineTableCount++;
     } else {
-      _realTimeTableCount++;
+      context._realTimeTableCount++;
     }
 
     IdealState idealState = _pinotHelixResourceManager.getTableIdealState(tableNameWithType);
@@ -132,11 +120,11 @@ public class SegmentStatusChecker extends ControllerPeriodicTask {
     }
 
     if (!idealState.isEnabled()) {
-      if (_logDisabledTables) {
+      if (context._logDisabledTables) {
         LOGGER.warn("Table {} is disabled. Skipping segment status checks", tableNameWithType);
       }
       resetTableMetrics(tableNameWithType);
-      _disabledTableCount++;
+      context._disabledTableCount++;
       return;
     }
 
@@ -147,16 +135,16 @@ public class SegmentStatusChecker extends ControllerPeriodicTask {
       } catch (NumberFormatException e) {
         // Ignore
       }
-      _metricsRegistry
+      _controllerMetrics
           .setValueOfTableGauge(tableNameWithType, ControllerGauge.NUMBER_OF_REPLICAS, nReplicasFromIdealState);
-      _metricsRegistry.setValueOfTableGauge(tableNameWithType, ControllerGauge.PERCENT_OF_REPLICAS, 100);
-      _metricsRegistry.setValueOfTableGauge(tableNameWithType, ControllerGauge.PERCENT_SEGMENTS_AVAILABLE, 100);
+      _controllerMetrics.setValueOfTableGauge(tableNameWithType, ControllerGauge.PERCENT_OF_REPLICAS, 100);
+      _controllerMetrics.setValueOfTableGauge(tableNameWithType, ControllerGauge.PERCENT_SEGMENTS_AVAILABLE, 100);
       return;
     }
 
-    _metricsRegistry
+    _controllerMetrics
         .setValueOfTableGauge(tableNameWithType, ControllerGauge.IDEALSTATE_ZNODE_SIZE, idealState.toString().length());
-    _metricsRegistry.setValueOfTableGauge(tableNameWithType, ControllerGauge.SEGMENT_COUNT,
+    _controllerMetrics.setValueOfTableGauge(tableNameWithType, ControllerGauge.SEGMENT_COUNT,
         (long) (idealState.getPartitionSet().size()));
     ExternalView externalView = _pinotHelixResourceManager.getTableExternalView(tableNameWithType);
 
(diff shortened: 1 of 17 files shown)


pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -42,9 +61,125 @@ public long getInitialDelayInSeconds() {
...
+      LOGGER.info("Finish running task: {} in {}ms", _taskName, System.currentTimeMillis() - startTime);
+    } else {
+      LOGGER.warn("Task: {} is skipped because it is not started or already stopped", _taskName);
+    }
+
+    _running = false;
+  }
+
+  /**
+   * Executes the task. This method should early terminate if {@link}
```
jackjlli: Please add the condition when it should early terminate.

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -42,9 +61,125 @@ public long getInitialDelayInSeconds() {
...
+      stopTask();
+    } catch (Exception e) {
+      LOGGER.error("Caught exception while cleaning up task: {}", _taskName, e);
+    }
+  }
+
+  /**
+   * Can be override for extra task cleanups.
+   */
+  protected void stopTask() {
```
jackjlli: CleanupTask might be better?

pinot-core/src/main/java/org/apache/pinot/core/periodictask/PeriodicTaskScheduler.java
```diff
@@ -37,16 +37,14 @@
   private List<PeriodicTask> _tasksWithValidInterval;
 
   /**
-   * Initialize the PeriodicTaskScheduler with list of PeriodicTasks
-   * @param periodicTasks
+   * Initialize the PeriodicTaskScheduler with list of PeriodicTasks.
```
jackjlli: Initializes

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -42,9 +61,125 @@ public long getInitialDelayInSeconds() {
...
+
+  /**
+   * {@inheritDoc}
+   * <p>
+   * This method sets {@code started} flag to true.
+   */
   @Override
-  public String getTaskName() {
-    return _taskName;
+  public final synchronized void start() {
```
jackjlli: So the purpose of adding synchronized here is to avoid different periodic tasks running at the same time? Then is it the same as running multiple tasks in one thread?

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -42,9 +61,125 @@ public long getInitialDelayInSeconds() {
...
+      return;
+    }
+    _started = false;
+
+    if (_running) {
+      long startTimeMs = System.currentTimeMillis();
+      long remainingTimeMs = MAX_PERIODIC_TASK_STOP_TIME_MILLIS;
+      LOGGER.info("Task: {} is running, wait for at most {}ms for it to finish", _taskName, remainingTimeMs);
+      while (_running && remainingTimeMs > 0L) {
+        long sleepTimeMs = Long.max(remainingTimeMs, 1000L);
```
jackjlli: I'm not quite sure the logic here. The variable `remainingTimeMs` is initially set to 30_000L. Then `sleepTimeMs` is set to 30_000L, since it's assigned from the larger number between 30_000L and 1_000L. Then `remainingTimeMs` is set to 0L. In this case, the thread will be slept only once. Could it be `min()` instead of `max()`?

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/retention/RetentionManagerTest.java
```diff
@@ -92,7 +92,7 @@ private void testDifferentTimeUnits(long pastTimeStamp, TimeUnit timeUnit, long
     conf.setRetentionControllerFrequencyInSeconds(0);
     conf.setDeletedSegmentsRetentionInDays(0);
     RetentionManager retentionManager = new RetentionManager(pinotHelixResourceManager, conf, controllerMetrics);
-    retentionManager.init();
+    retentionManager.start();
```
jackjlli: I'm thinking whether the method `stop` be added at the end but it doesn't seem matter.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/relocation/RealtimeSegmentRelocator.java
```diff
@@ -62,32 +61,13 @@ public RealtimeSegmentRelocator(PinotHelixResourceManager pinotHelixResourceMana
...
-
-  @Override
-  protected void preprocess() {
-  }
-
   @Override
   protected void processTable(String tableNameWithType) {
-    CommonConstants.Helix.TableType tableType = TableNameBuilder.getTableTypeFromTableName(tableNameWithType);
-    if (tableType == CommonConstants.Helix.TableType.REALTIME) {
+    if (TableNameBuilder.REALTIME.tableHasTypeSuffix(tableNameWithType)) {
```
jackjlli: 👍

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
Jackie-Jiang: Good catch. Seems like where I took a break and ran for dinner haha

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
Jackie-Jiang: Done.

pinot-core/src/main/java/org/apache/pinot/core/periodictask/PeriodicTaskScheduler.java
Jackie-Jiang: Done.

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
Jackie-Jiang: No, the purpose of this synchronized is to avoid calling start() and stop() at the same time. Basically another level of protection. (start() and stop() are called in scheduler, not in the task thread)

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
Jackie-Jiang: Good catch! It should be min()

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/retention/RetentionManagerTest.java
Jackie-Jiang: I have the same feeling, but I'm just lazy lol

sunithabeeram:
@Jackie-Jiang what race condition was addressed in this PR and how?
Was any manual testing done to make sure leader and non-leader controllers behave as expected?

mcvsubbu:
@Jackie-Jiang could you also please check with @npawar ? I think we had similar questions (should there be a stop/start, etc.) in her PR and had made some choices because of the way things were at the time. Thanks.

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -18,20 +18,39 @@
...
+
+
 /**
  * A base class to implement periodic task interface.
  */
+@ThreadSafe
 public abstract class BasePeriodicTask implements PeriodicTask {
+  private static final Logger LOGGER = LoggerFactory.getLogger(BasePeriodicTask.class);
+
+  // Wait for at most 5 minutes while calling stop() for task to terminate
```
npawar: comment says 5 minutes, value says 30s?

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -42,9 +61,126 @@ public long getInitialDelayInSeconds() {
...
+  public final void run() {
+    _running = true;
+
+    if (_started) {
+      long startTime = System.currentTimeMillis();
+      LOGGER.info("Start running task: {}", _taskName);
+      try {
+        runTask();
+      } catch (Exception e) {
+        LOGGER.error("Caught exception while running task: {}", _taskName, e);
```
npawar: Seems like a good place to introduce a metric to count failures in PeriodicTasks, so that we know if a task is erroring out

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
```diff
@@ -42,9 +61,126 @@ public long getInitialDelayInSeconds() {
...
+        try {
+          Thread.sleep(sleepTimeMs);
+        } catch (InterruptedException e) {
+          LOGGER.error("Caught InterruptedException while waiting for task: {} to finish", _taskName);
+          Thread.currentThread().interrupt();
+          break;
+        }
+      }
+      long waitTimeMs = System.currentTimeMillis() - startTimeMs;
+      if (_running) {
```
npawar: the LOG in if and else look like they're switched

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -31,152 +31,90 @@
...
     }
+    postprocess(context);
+    _controllerMetrics
+        .setValueOfGlobalGauge(ControllerGauge.PERIODIC_TASK_NUM_TABLES_PROCESSED, _taskName, numTablesProcessed);
+    LOGGER.info("Finish processing {}/{} tables in task: {}", numTablesProcessed, numTables, _taskName);
   }
 
   /**
-   * This method runs before processing all tables
+   * Can be override to provide context before processing the tables.
```
npawar: *can be overridden

pinot-controller/src/main/java/org/apache/pinot/controller/helix/SegmentStatusChecker.java
```diff
@@ -69,58 +65,50 @@ public SegmentStatusChecker(PinotHelixResourceManager pinotHelixResourceManager,
...
+    return context;
   }
 
   @Override
-  protected void processTable(String tableNameWithType) {
-    updateSegmentMetrics(tableNameWithType);
-  }
-
-  @Override
-  protected void exceptionHandler(String tableNameWithType, Exception e) {
```
npawar: Why remove exceptionHandler method? It was consciously added so that we avoid such try catch in each ControllerPeriodicTask.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/minion/PinotTaskManager.java
```diff
@@ -44,134 +43,126 @@
...
   }
 
-  /**
-   * Performs necessary cleanups (e.g. remove metrics) when the controller leadership changes.
-   */
-  @Override
-  public void stopTask() {
-    LOGGER.info("Perform task cleanups.");
-    // Performs necessary cleanups for each task type.
+  private void nonLeaderCleanUp() {
```
npawar: Can some comments be added, either here or above cleanupTask explaining why non leader cleanup is needed in this task? Since these tasks run on controller leadership gain, it's confusing to see this method here

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -31,152 +31,90 @@
 /**
  * The base periodic task for pinot controller only. It uses <code>PinotHelixResourceManager</code> to determine
  * which table resources should be managed by this Pinot controller.
+ *
+ * @param <C> the context type
```
npawar: Nicely done with the Context

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
```diff
@@ -31,152 +31,90 @@
...
-        try {
-          processTable(tableNameWithType);
-          numTablesProcessed++;
-        } catch (Exception e) {
-          exceptionHandler(tableNameWithType, e);
-        }
+      try {
+        processTable(tableNameWithType, context);
+      } catch (Exception e) {
+        LOGGER.error("Caught exception while processing table: {} in task: {}", tableNamesWithType, _taskName, e);
```
npawar: why not keep exceptionHandler and call it here?

Jackie-Jiang:
@sunithabeeram The race condition between start() and stop(), and is addressed by adding synchronization.
As for the leader and non-leader change, ControllerPeriodicTaskTest covers the test of start() and stop(), which are called in ControllerPeriodicTaskScheduler onBecomingLeader() and onBecomingNonLeader() (implementing LeadershipChangeSubscriber).

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
Jackie-Jiang: Missed a zero. Change comment to 30s. Thanks

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
Jackie-Jiang: Correct, good catch

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
Jackie-Jiang: Fixed

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
Jackie-Jiang: Added metrics for CONTROLLER_PERIODIC_TASK_RUN and CONTROLLER_PERIODIC_TASK_ERROR to track both number of runs and errors

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
Jackie-Jiang: Added back processTableExceptionHandler and pass in Context

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/minion/PinotTaskManager.java
Jackie-Jiang: Added comments in BasePeriodicTask

pinot-controller/src/main/java/org/apache/pinot/controller/helix/SegmentStatusChecker.java
Jackie-Jiang: After consideration, I still think this exception handling logic should be wrapped inside the processTable(). It might need some variables from the processTable(), which are not accessible with this callback handler.

Jackie-Jiang:
@npawar Addressed the comments. Periodic task logic is covered by unit test: ControllerPeriodicTask and PeriodicTaskSchedulerTest. For each individual periodic task, it's out of the scope of this PR, and we can add tests later.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/periodictask/ControllerPeriodicTask.java
npawar: I think you missed it again :) Here and also in 2 other places in this file

pinot-core/src/main/java/org/apache/pinot/core/periodictask/BasePeriodicTask.java
npawar: Not setting them here? Or planning to handle it later? If so, please add a TODO

PR #3822, opened 2019-02-12
[TE] Return yaml errors after grouping and use consistent convention …

akshayrai:
…(subscription over notification)

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
@@ -191,62 +191,80 @@ public class YamlResource {
   public Response createYamlAlert(@ApiParam(value =  "a json contains both subscription and detection yaml as string")  String payload,
       @ApiParam("tuning window start time for tunable components") @QueryParam("startTime") long startTime,
       @ApiParam("tuning window end time for tunable components") @QueryParam("endTime") long endTime) throws Exception {
-    Map<String, String> yamls = OBJECT_MAPPER.readValue(payload, Map.class);
-    if (StringUtils.isBlank(payload)){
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -405,10 +424,10 @@ long createSubscriptionGroup(String yamlAlertConfig) throws ValidationException
   }
 
   @POST
-  @Path("/notification")
+  @Path("/subscription")
```
jihaozh: change this path requires frontend changes too.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -483,10 +502,10 @@ void updateSubscriptionGroup(long oldAlertConfigID, String yamlAlertConfig) thro
   }
 
   @PUT
-  @Path("/notification/{id}")
+  @Path("/subscription/{id}")
```
jihaozh: same as above.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
akshayrai: Yes, I have already communicated this with @aaronucsd

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
akshayrai: as above

PR #3832, opened 2019-02-13
[TE] Extend anomalies endpoint to fetch by metric/dataset and true an…

akshayrai:
…omalies

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/api/user/dashboard/UserDashboardResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/api/user/dashboard/UserDashboardResource.java
@@ -17,32 +17,10 @@
  * under the License.
  */
 
-package org.apache.pinot.thirdeye.dashboard.resources.v2;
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/api/user/dashboard/UserDashboardResource.java
```diff
@@ -345,7 +370,17 @@ public int compare(MergedAnomalyResultDTO o1, MergedAnomalyResultDTO o2) {
     return output;
   }
 
+  private Collection<? extends MergedAnomalyResultDTO> fetchFrameworkAnomaliesByMetricDataset(
+      List<Predicate> predicates) {
+    predicates.add(Predicate.NEQ("detectionConfigId", 0));
```
jihaozh: don't need to add this predicate

akshayrai: Anomalies is a list. This is a safety condition that avoids potential duplicates.

PR #3833, opened 2019-02-13
Update maven config files for preparing release

snleee:
1. Added apache parent pom
2. Make 'apache-release' profile to build source tarbells,
   sha512 hash, and sign
3. Override configs for Apache rat to make the check pass
4. Turning off doclint for javadoc
5. Fix "LoaderUtilsTest" to use randomly generated tmp directory
   instead of using "tmp" dir within a project
6. Fix bin-dist to include base directory for its tarbell
7. Add Apache RAT exclude configs


pom.xml
```diff
@@ -827,6 +851,9 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-javadoc-plugin</artifactId>
           <version>2.10.1</version>
+          <configuration>
+            <additionalparam>-Xdoclint:none</additionalparam>
```
Jackie-Jiang: Add a comment to remove this after fixing all the java doc issues.

snleee: done

PR #3834, opened 2019-02-13
Unify move method in PinotFS

jackjlli:
This PR unifies move method in different implementations of PinotFS.

In AzureFS, a file cannot move to a non-empty directory. Since the only usage of this API in Pinot is to move a file to a file, we add a note in PinotFS interface to suggest always specifying the names of both src and dst to make less confusion.

### DIFF
--- a/pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
+++ b/pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
@@ -103,12 +103,8 @@ public class AzurePinotFS extends PinotFS {
    * then the call fails rather than overwrite the directory. Thus, it's best to specify
    * the full paths of both src and dst when using this method in AzurePinotFS.
    */
-  public boolean doMove(URI srcUri, URI dstUri)
+  protected boolean doMove(URI srcUri, URI dstUri)
       throws IOException {
-    // ensures the parent path of dst exists.
-    URI parentUri = Paths.get(dstUri).toUri();
-    _adlStoreClient.createDirectory(parentUri.getPath());
-    // renames the source file/directory to destination.
     return _adlStoreClient.rename(srcUri.getPath(), dstUri.getPath());
   }
 
--- a/pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
+++ b/pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
@@ -98,24 +98,17 @@ public class AzurePinotFS extends PinotFS {
   }
 
   @Override
-  public boolean move(URI srcUri, URI dstUri, boolean overwrite)
+  /**
+   * Note: One corner scenario for this method in AzureFS is if the destination is a non-empty directory,
+   * then the call fails rather than overwrite the directory. Thus, it's best to specify
+   * the full paths of both src and dst when using this method in AzurePinotFS.
+   */
+  public boolean doMove(URI srcUri, URI dstUri)
       throws IOException {
-    if (exists(dstUri)) {
-      if (overwrite) {
-        delete(dstUri, true);
-      } else {
-        // dst file exists, returning
-        return false;
-      }
-    } else {
-      URI parentUri = Paths.get(dstUri).toUri();
-      // ensure the dst path exists
-      _adlStoreClient.createDirectory(parentUri.getPath());
-    }
-    if (exists(dstUri) && !overwrite) {
-      return false;
-    }
-    //rename the file
+    // ensures the parent path of dst exists.
+    URI parentUri = Paths.get(dstUri).toUri();
+    _adlStoreClient.createDirectory(parentUri.getPath());
+    // renames the source file/directory to destination.
     return _adlStoreClient.rename(srcUri.getPath(), dstUri.getPath());
   }
 
--- a/pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
+++ b/pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
@@ -100,6 +100,18 @@ public class AzurePinotFS extends PinotFS {
   @Override
   public boolean move(URI srcUri, URI dstUri, boolean overwrite)
       throws IOException {
+    if (exists(dstUri)) {
+      if (overwrite) {
+        delete(dstUri, true);
+      } else {
+        // dst file exists, returning
+        return false;
+      }
+    } else {
+      URI parentUri = Paths.get(dstUri).toUri();
+      // ensure the dst path exists
+      _adlStoreClient.createDirectory(parentUri.getPath());
+    }
     if (exists(dstUri) && !overwrite) {
       return false;
     }
--- a/pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
+++ b/pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
@@ -98,11 +98,6 @@ public class AzurePinotFS extends PinotFS {
   }
 
   @Override
-  /**
-   * Note: One corner scenario for this method in AzureFS is if the destination is a non-empty directory,
-   * then the call fails rather than overwrite the directory. Thus, it's best to specify
-   * the full paths of both src and dst when using this method in AzurePinotFS.
-   */
   protected boolean doMove(URI srcUri, URI dstUri)
       throws IOException {
     return _adlStoreClient.rename(srcUri.getPath(), dstUri.getPath());
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
@@ -170,8 +170,8 @@ public class ControllerConf extends PropertiesConfiguration {
 
   public static URI constructSegmentLocation(String baseDataDir, String tableName, String segmentName) {
     try {
-      return new URI(StringUtil.join(File.separator, baseDataDir, tableName, URLEncoder.encode(segmentName, "UTF-8")));
-    } catch (UnsupportedEncodingException | URISyntaxException e) {
+      return getUriFromPath(StringUtil.join(File.separator, baseDataDir, tableName, URLEncoder.encode(segmentName, "UTF-8")));
+    } catch (UnsupportedEncodingException e) {
       LOGGER
           .error("Could not construct segment location with baseDataDir {}, tableName {}, segmentName {}", baseDataDir,
               tableName, segmentName);
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/SegmentDeletionManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/SegmentDeletionManager.java
@@ -188,12 +188,16 @@ public class SegmentDeletionManager {
       try {
         if (pinotFS.exists(fileToMoveURI)) {
           // Overwrites the file if it already exists in the target directory.
-          pinotFS.move(fileToMoveURI, deletedSegmentDestURI, true);
-          // Updates last modified.
-          // Touch is needed here so that removeAgedDeletedSegments() works correctly.
-          pinotFS.touch(deletedSegmentDestURI);
-          LOGGER.info("Moved segment {} from {} to {}", segmentId, fileToMoveURI.toString(),
-              deletedSegmentDestURI.toString());
+          if (pinotFS.move(fileToMoveURI, deletedSegmentDestURI, true)) {
+            // Updates last modified.
+            // Touch is needed here so that removeAgedDeletedSegments() works correctly.
+            pinotFS.touch(deletedSegmentDestURI);
+            LOGGER.info("Moved segment {} from {} to {}", segmentId, fileToMoveURI.toString(),
+                deletedSegmentDestURI.toString());
+          } else {
+            LOGGER.warn("Failed to move segment {} from {} to {}", segmentId, fileToMoveURI.toString(),
+                deletedSegmentDestURI.toString());
+          }
         } else {
           if (!SegmentName.isHighLevelConsumerSegmentName(segmentId)) {
             LOGGER.warn("Not found local segment file for segment {}" + fileToMoveURI.toString());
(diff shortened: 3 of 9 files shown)


pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
```diff
@@ -100,6 +100,18 @@ public boolean delete(URI segmentUri, boolean forceDelete)
...
       throws IOException {
+    if (exists(dstUri)) {
+      if (overwrite) {
+        delete(dstUri, true);
+      } else {
+        // dst file exists, returning
+        return false;
+      }
+    } else {
+      URI parentUri = Paths.get(dstUri).toUri();
```
mcvsubbu: Am i reading this right that we are essentially creating dstUri as a directory but asking the user to pass the input param a a file?

jackjlli: Correct the method according to the latest API doc we made in PinotFS interface.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/SegmentDeletionManager.java
```diff
@@ -188,12 +188,13 @@ protected void removeSegmentFromStore(String tableNameWithType, String segmentId
       try {
         if (pinotFS.exists(fileToMoveURI)) {
           // Overwrites the file if it already exists in the target directory.
-          pinotFS.move(fileToMoveURI, deletedSegmentDestURI, true);
-          // Updates last modified.
-          // Touch is needed here so that removeAgedDeletedSegments() works correctly.
-          pinotFS.touch(deletedSegmentDestURI);
-          LOGGER.info("Moved segment {} from {} to {}", segmentId, fileToMoveURI.toString(),
-              deletedSegmentDestURI.toString());
+          if (pinotFS.move(fileToMoveURI, deletedSegmentDestURI, true)) {
```
sunithabeeram: Should an message be logged if move returns false? If we can rely on move to always log, then you can remove the additional log in PinotLLCRealtimeSegmentManager.java

pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
```diff
@@ -100,9 +100,28 @@ public boolean delete(URI segmentUri, boolean forceDelete)
...
+        } else {
+          // dst file exists, returning
+          return false;
+        }
+      }
+    }
+    // ensures the parent path of dst exists.
+    URI parentUri = Paths.get(dstUri).toUri();
+    _adlStoreClient.createDirectory(parentUri.getPath());
     //rename the file
```
sunithabeeram: This comment is misleading as its assuming a 'file' whereas all of the above code seems more flexible.

pinot-hadoop-filesystem/src/main/java/org/apache/pinot/filesystem/HadoopPinotFS.java
```diff
@@ -92,9 +92,25 @@ public boolean delete(URI segmentUri, boolean forceDelete)
   @Override
   public boolean move(URI srcUri, URI dstUri, boolean overwrite)
       throws IOException {
-    if (exists(dstUri) && !overwrite) {
+    if (!exists(srcUri)) {
```
sunithabeeram: This part of the code exists in all implementations. Can we add this to PinotFS, call super.move() and then handle FS specific logic in the impl classes?

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -59,12 +59,15 @@ public abstract boolean delete(URI segmentUri, boolean forceDelete)
...
-   * If src is a file and dst is a directory, src file will get moved under dst directory.
+   * If src is a file and dst is an empty directory, src file will get moved under dst directory.
    * If both src and dst are directories, src directory will get moved under dst directory.
    * If src is a directory and dst is a file, operation will fail.
    * For example, if a file /a/b/c is moved to a file /x/y/z, in the case of overwrite, the directory /a/b still exists,
    * but will not contain the file 'c'. Instead, /x/y/z will contain the contents of 'c'.
    * If a file /a is moved to a directory /x/y, all the original files under /x/y will be kept.
+   *
+   * As of now, the usage of this API in Pinot is just to move a file to a file. So please specify the full names of
+   * both src and dst to make less confusion.
```
sunithabeeram: Remove "to make less confusion"

mcvsubbu:
I think we should insist that the interfaces work only if full paths are specified in both src and dest. We can change the usage to be that way, and the implementations will get simpler across file systems.

So, for example it is illegal to move(src, dst)  unless the last part of src and dst match. Same with copy.

@sunithabeeram  what if we approach it this way? We can still support all combinations of moving from file to dir, dir to dir, etc. just restricting the usage.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/SegmentDeletionManager.java
jackjlli: I think it's essential to log when returning false because the false result may be returned from the deep inner logic of different FS.

pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
jackjlli: Adjusted the comment.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
jackjlli: Done.

sunithabeeram:
@mcvsubbu since the common logic is moved to the abstract class now, its a bit more easier to enforce the contracts across FS impls. We have noted exceptions in case of AzureFS - didn't want to twist the top level interface because one FS behaved differently.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -22,14 +22,19 @@
...
+import java.nio.file.Paths;
 import org.apache.commons.configuration.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
  * The PinotFS is intended to be a thin wrapper on top of different filesystems. This interface is intended for internal
  * Pinot use only. This class will be implemented for each pluggable storage type.
  */
```
sunithabeeram: Lets add notes on some of the core discussions we have around this so its clear to other why the interface is the way it is:
"PinotFS is a restricted FS API that exposes functionality that is required for Pinot to use different FS implementations. The restrictions are in place due to 2 driving factors:
- Prevent unexpected performance hit when a broader API is implemented - especially, we would like to reduce calls to remote filesystems that might be needed for a broader API, but not necessarily required by Pinot (see the documentation for move() method below).
- Provide an interface that is simple to be implemented across different FS types. The contract that developers have to adhere to will be simpler.

Please read the method level docs carefully to note the exceptions while using the APIs."

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -58,31 +63,53 @@ public abstract boolean delete(URI segmentUri, boolean forceDelete)
...
    * that haven't been created, this method will create all the necessary parent directories.
-   * If both src and dst are files, dst will be overwritten.
-   * If src is a file and dst is a directory, src file will get moved under dst directory.
-   * If both src and dst are directories, src directory will get moved under dst directory.
-   * If src is a directory and dst is a file, operation will fail.
+   * Note: In Pinot we recommend the full paths of both src and dst be specified.
    * For example, if a file /a/b/c is moved to a file /x/y/z, in the case of overwrite, the directory /a/b still exists,
    * but will not contain the file 'c'. Instead, /x/y/z will contain the contents of 'c'.
-   * If a file /a is moved to a directory /x/y, all the original files under /x/y will be kept.
+   * If a directory /a/b/ is renamed to another directory /x/y/, the directory /x/y/ will contains the content of /a/b/.
```
sunithabeeram: Lets not mix terminology here: Give examples in terms of move API so its clear to readers. i.e., if srcUri = /a/b; an the contents of /a/b are: /a/b/c, /a/b/d; and dstUri=/x/y, the result will be /a (directory b removed), and dst will be /x/y/c, /x/y/d.

And then provide the exception case: if you would like to move b under /x/y, invoke move with srcUri=/a/b and dstUri=/x/y/b.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -58,31 +63,53 @@ public abstract boolean delete(URI segmentUri, boolean forceDelete)
...
    * that haven't been created, this method will create all the necessary parent directories.
-   * If both src and dst are files, dst will be overwritten.
-   * If src is a file and dst is a directory, src file will get copied under dst directory.
-   * If both src and dst are directories, src directory will get copied under dst directory.
-   * If src is a directory and dst is a file, operation will fail.
-   * For example, if a file /x/y/z is copied to /a/b/c, /x/y/z will be retained and /x/y/z will also be present as /a/b/c;
-   * if a file /a is copied to a directory /x/y, all the original files under /x/y will be kept.
+   * Note: In Pinot we recommend the full paths of both src and dst be specified.
+   * For example, if a file /a/b/c is copied to a file /x/y/z, in the case of overwrite, the directory /a/b still exists,
+   * but will not contain the file 'c'. Instead, /x/y/z will contain the contents of 'c'.
```
sunithabeeram: That seems wrong. Source will still have c. Its just that if /x/y/z had a c, c will be overwritten.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -58,31 +63,53 @@ public abstract boolean delete(URI segmentUri, boolean forceDelete)
...
+   * If a directory /a/b/ is copied to another directory /x/y/, the directory /x/y/ will contains the content of /a/b/.
```
sunithabeeram: Might be good to rephrase this as in the move example with src and dst arguments used for the calls.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -170,8 +170,8 @@ public static URI getUriFromPath(String path) {
 
   public static URI constructSegmentLocation(String baseDataDir, String tableName, String segmentName) {
     try {
-      return new URI(StringUtil.join(File.separator, baseDataDir, tableName, URLEncoder.encode(segmentName, "UTF-8")));
-    } catch (UnsupportedEncodingException | URISyntaxException e) {
+      return getUriFromPath(StringUtil.join(File.separator, baseDataDir, tableName, URLEncoder.encode(segmentName, "UTF-8")));
```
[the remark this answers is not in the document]
jackjlli: Yes, this change is needed as `creating parent directory` logic is put to base method level in PinotFS class.

pinot-hadoop-filesystem/src/main/java/org/apache/pinot/filesystem/HadoopPinotFS.java
jackjlli: Common logic has been moved to base class.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
jackjlli: Thanks for the detail! I'll add your comments to the API doc.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
jackjlli: Good catch! 👍 Addressed.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
jackjlli: Tests added for this two scenario.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
jackjlli: Comments updated.

pinot-azure-filesystem/src/main/java/org/apache/pinot/filesystem/AzurePinotFS.java
```diff
@@ -98,12 +98,13 @@ public boolean delete(URI segmentUri, boolean forceDelete)
   }
 
   @Override
-  public boolean move(URI srcUri, URI dstUri, boolean overwrite)
+  /**
```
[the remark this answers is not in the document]
jackjlli: Removed.

pinot-filesystem/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -58,31 +69,54 @@ public abstract boolean delete(URI segmentUri, boolean forceDelete)
...
    * Copies the file or directory from the src to dst. The original file is retained. If the dst has parent directories
    * that haven't been created, this method will create all the necessary parent directories.
-   * If both src and dst are files, dst will be overwritten.
-   * If src is a file and dst is a directory, src file will get copied under dst directory.
-   * If both src and dst are directories, src directory will get copied under dst directory.
-   * If src is a directory and dst is a file, operation will fail.
-   * For example, if a file /x/y/z is copied to /a/b/c, /x/y/z will be retained and /x/y/z will also be present as /a/b/c;
-   * if a file /a is copied to a directory /x/y, all the original files under /x/y will be kept.
+   * Note: In Pinot we recommend the full paths of both src and dst be specified.
+   * For example, if a file /a/b/c is copied to a file /x/y/z, in the case of overwrite, the directory /a/b still exists,
```
sunithabeeram: Didn't catch this earlier - I don't see an overwrite for copy! We can move addressing/reviewing copy related changes out of this PR if you prefer.

jackjlli: There's no code changes for copy in this PR. So no harm to leave the doc change here. :)

sunithabeeram: Lets not have incorrect documentation though, please :)

jackjlli: The doc has been updated without mentioning `overwrite`.

PR #3838, opened 2019-02-14
Fix listFiles method in HadoopPinotFS

jackjlli:
Currently the existing behavior of `listFiles()` in HadoopPinotFS only returns pure files instead of both files and directories, which makes SegmentDeletionManager doesn't work in HDFS.

This PR changes the behavior so that both files and directories can be returned by calling `listFiles()`.


mcvsubbu:
If you can please add a TODO to move the tmp file logic into one file (perhaps SegmentCompletionUtils) that will be nice.

PR #3840, opened 2019-02-15
Add headers for docs

snleee:
According to the apache header policy, documentation files (md, rst)
also needs the headers just like source codes. This pr addresses
the issue.


snleee:
https://www.apache.org/legal/src-headers.html

PR #3842, opened 2019-02-15
[TE] detection - time series provider loading cache

jihaozh:
- Set the maximum weight of the loading cache
- Improve onboarding performance


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DefaultDataProvider.java
```diff
@@ -84,27 +87,38 @@ public DefaultDataProvider(MetricConfigManager metricDAO, DatasetConfigManager d
...
-            return fetchTimeseries(Collections.singleton(slice)).get(slice);
-          }
-        });
-  }
 
+    if (DETECTION_TIME_SERIES_CACHE == null) {
+      LOG.info("initializing detection timeseries cache");
+      DETECTION_TIME_SERIES_CACHE = CacheBuilder.newBuilder()
+          // don't use more than one third of memory for detection time series
+          .maximumWeight(Runtime.getRuntime().freeMemory() / 3)
```
akshayrai: I think it would be better to use a hard-coded size than determining the size at run-time! Probably easier to debug issues in the future.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DefaultDataProvider.java
```diff
@@ -84,27 +87,38 @@ public DefaultDataProvider(MetricConfigManager metricDAO, DatasetConfigManager d
...
+          .expireAfterWrite(15, TimeUnit.MINUTES)
```
akshayrai: Can you include a comment stating that these values are chosen randomly and needs to be tuned.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DefaultDataProvider.java
```diff
@@ -136,13 +150,17 @@ public DataFrame load(MetricSlice slice) {
...
-        this.timeseriesCache.get(slice);
-      } catch (Exception e){
-        LOG.warn("cache time series for slice {} failed", slice.toString());
+  public Map<MetricSlice, DataFrame> fetchTimeseries(Collection<MetricSlice> slices) {
+    try {
+      Map<MetricSlice, DataFrame> cacheResult = DETECTION_TIME_SERIES_CACHE.getAll(slices);
+      Map<MetricSlice, DataFrame> timeseriesResult = new HashMap<>();
+      for (Map.Entry<MetricSlice, DataFrame> entry : cacheResult.entrySet()){
+        // make a copy of the result so that cache won't be contaminated by client code
+        timeseriesResult.put(entry.getKey(), entry.getValue().copy());
```
akshayrai: Does this mean we will make multiple copies and store them in-memory for each sliding window? That might be a lot of data.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DefaultDataProvider.java
jihaozh: It's hard to estimate how much would be enough. (Depends on how many detection jobs is running, etc). If set too low, the performance suffers, if set too high, the cache would eat too much memory. To cap it to one-third of memory is reasonable I think. added a log there to show the cache size.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DefaultDataProvider.java
jihaozh: added

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DefaultDataProvider.java
jihaozh: yup. but even the cache disabled, it creates a copy of time series anyways.

Issue #3847, opened 2019-02-15
Controller downloadSegment() api not working on Low Level Kafka consumer on deep storage.


chenboat:
When using low level realtime data consumer, a slow server now may download the segments from controller. The controller's segment download api uses FileUploadPathProvider. Its getBaseDataDir() method now assume all segments are stored locally and requires the data dir path scheme being "file" -- thus deep storage scheme can not work.

==================================
2019-02-15 22:38:14 INFO  FileUploadPathProvider:55 - Data directory: viewfs://ns-default/app/upinot/llc_staging
2019-02-15 22:38:14 INFO  FileUploadPathProvider:57 - Schema temporary directory: viewfs://ns-default/app/upinot/llc_staging/schemasTemp
2019-02-15 22:38:14 INFO  FileUploadPathProvider:71 - Local temporary directory: file:/var/upinot/stream-pinot-controller/data
2019-02-15 22:38:14 INFO  FileUploadPathProvider:76 - File upload temporary directory: file:/var/upinot/stream-pinot-controller/data/fileUploadTemp
2019-02-15 22:38:14 INFO  FileUploadPathProvider:78 - Untarred file temporary directory: file:/var/upinot/stream-pinot-controller/data/fileUploadTemp/untarred
2019-02-15 22:38:14 ERROR WebApplicationExceptionMapper:39 - Server error: 
java.lang.IllegalArgumentException: URI scheme is not "file"
        at java.io.File.<init>(File.java:421)
        at com.linkedin.pinot.controller.api.resources.FileUploadPathProvider.getBaseDataDir(FileUploadPathProvider.java:129)
        at com.linkedin.pinot.controller.api.resources.PinotSegmentUploadRestletResource.downloadSegment(PinotSegmentUploadRestletResource.java:184)
..........
2019-02-15 22:38:14 INFO  ControllerResponseFilter:47 - Handled request from 10.76.133.45 GET http://streampinot-sandbox01-sjc1:5984/segments/llc_test_split_commit/llc_test_split_commit__1__1__20190215T2224Z, content-type null status code 500 Internal Server Error

PR #3849, opened 2019-02-16
Using PinotFS interface instead of local file to download a segment f…

chenboat:
#3847 

When using low level realtime data consumer, a slow server now may download the segments from controller. The controller's segment download api uses FileUploadPathProvider. Its getBaseDataDir() method now assume all segments are stored locally and requires the data dir path scheme being "file" -- thus deep storage scheme can not work.

1. This refactoring uses PinotFS interface instead. 
2. I am not sure the reason why the original test created segment files with suffix ' %'. This kind of filenames are good for local filesystem but not valid as an URI path used by the PinotFS. I deleted such segment file names.

@mcvsubbu @npawar @jamesyfshao


pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -182,19 +186,34 @@ public Response downloadSegment(
...
       throw new ControllerApplicationException(LOGGER,
-          "Segment " + segmentName + " or table " + tableName + " not found", Response.Status.NOT_FOUND);
+          "Segment " + segmentName + " or table " + tableName + " not found in " + segmentFileURI.toString(), Response.Status.NOT_FOUND);
     }
-    Response.ResponseBuilder builder = Response.ok(dataFile);
-    builder.header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=" + dataFile.getName());
-    builder.header(HttpHeaders.CONTENT_LENGTH, dataFile.length());
+    File tmpSegmentFile = new File(StringUtil.join("/", _controllerConf.getLocalTempDir(), tableName,
+        StringUtil.join("_", segmentName, String.valueOf(System.nanoTime()))));
+    pinotFS.copyToLocalFile(segmentFileURI, tmpSegmentFile);
```
mcvsubbu: This means an extra copy for local file system. Is there any way to avoid it?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -182,19 +186,34 @@ public Response downloadSegment(
...
-      segmentName = URLDecoder.decode(segmentName, "UTF-8");
+      segmentName = URLDecoder.decode(segmentName, URL_ENCODING_SCHEME);
     } catch (UnsupportedEncodingException e) {
       String errStr = "Could not decode segment name '" + segmentName + "'";
       throw new ControllerApplicationException(LOGGER, errStr, Response.Status.BAD_REQUEST);
     }
-    final File dataFile = new File(provider.getBaseDataDir(), StringUtil.join("/", tableName, segmentName));
-    if (!dataFile.exists()) {
+
+    final java.net.URI segmentFileURI =
```
npawar: we already have `import java.net.URI` in the imports, so we shouldn't need the fully qualified name here

mcvsubbu:
Also, the servers should pull from the URI that is there in the metadata. The way we should solve this is to change the URI in the metadata to point to a remote store for completed realtime segments

chenboat:
> Also, the servers should pull from the URI that is there in the metadata. The way we should solve this is to change the URI in the metadata to point to a remote store for completed realtime segments

Are you suggesting in case of non-local storage, Pinot servers by-pass Pinot controllers to download the segment from remote directly. I think that is a good idea (and we in fact talked out this during last meeting) but it is out of scope of this PR in my opinion because:
(1) PinotSegmentUploadRestletResource right now does not work with Pinot FS -- so it is a bug we should fix unless we want to delete this resource.
(2) Currently in LLC, in the common path the commit servers anyway need to pass segments to Pinot controller during commit. The download segment path is in fact a less common code path. So I do not see the point of doing the less common path optimization before we solve the whole problem of by-passing Pinot controller for segment data transfer.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
chenboat: There will be an extra copy operation. But the tmp file will be deleted before this method exits though. One optimization I can think of is to optimize for the local FS case: there is no need for the copy op. For deep storage, I do not think of an easy way except the structural you suggested below (which i have some pushback too).

npawar:
Let's avoid the extra copy in the localFS case. Like you suggested, adding an isLocal check should be an okay approach for now.
One thing we could do is  push the entire logic of fetching file and building Response into the PinotFS interface. But that wouldn't be very clean either, as it is too specific a task for an FS to perform.

chenboat:
> Let's avoid the extra copy in the localFS case. Like you suggested, adding an isLocal check should be an okay approach for now.
> One thing we could do is push the entire logic of fetching file and building Response into the PinotFS interface. But that wouldn't be very clean either, as it is too specific a task for an FS to perform.

Ok. I filed a new PR #4757 to replace this one which is too old to update.

chenboat:
No longer need as fixed by #4808

PR #3850, opened 2019-02-17
Aggregation of population standard variance support

xueyumusic:
Population standard variance is a common aggregation in statistic and data analysis, and also supported in various data software, such as hive, spark, flink etc. It maybe an improvement to support it.

The updates are:
* add VarPopAggregationFunction and VarPopMVAggregationFunction
* add VarianceTuple similar as AvgPair to hold VarianceData
* update related aggregation classes
* add new tests
* add doc


xueyumusic:
Hi, @kishoreg , thanks for review, I investigated a bit more and found that the algo adopted by spark to compute approx variance and stddev maybe more practical. From [this wiki](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance) there are several different algos, and comparing them it looks Welford's online algorithm and its related algo(parallel algo and high-order algo) is more practical. They are also algos adopted by spark at [here](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CentralMomentAgg.scala#L189). So I think it maybe more appropriate to be used here. What do you think?

npawar:

MrNeocore:
Hello
Any update or related issues / PR on the topic ?

PR #3852, opened 2019-02-18
add support for parquet reader

WangGuangxin:
Parquet is widely used in hadoop echo, such as spark, hive. Many companies use it as ETL result format. So I think it is necessary to support generating pinot segment from parquet.

This PR add a ParquetRecordReader, which supports parquet as datasource.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
@@ -32,10 +32,11 @@ import org.apache.parquet.hadoop.ParquetWriter;
 import org.apache.parquet.hadoop.metadata.ParquetMetadata;
 
 import java.io.IOException;
-import java.util.List;
 
 
 public class ParquetUtils {
+  public static final String DEFAULT_FS = "file:///";
+
   /**
    * Get a ParquetReader with the given file.
    * @param fileName the parquet file to read
@@ -74,26 +75,24 @@ public class ParquetUtils {
   }
 
   /**
-   * write records as parquet file
+   * Get a ParquetWriter with the given file
+   * @param fileName
+   * @param schema
+   * @return a ParquetWriter
+   * @throws IOException
    */
-  public static void writeParquetRecord(String fileName, Schema schema, List<GenericRecord> records)
+  public static ParquetWriter<GenericRecord> getParquetWriter(String fileName, Schema schema)
       throws IOException {
-    ParquetWriter<GenericRecord> writer =
-        AvroParquetWriter.<GenericRecord>builder(new Path(fileName)).withSchema(schema).withConf(getConfiguration())
-            .build();
-
-    try {
-      for (GenericRecord r : records) {
-        writer.write(r);
-      }
-    } finally {
-      writer.close();
-    }
+    Path dataFsPath = new Path(fileName);
+    return AvroParquetWriter.<GenericRecord>builder(dataFsPath).withSchema(schema).withConf(getConfiguration()).build();
   }
 
   private static Configuration getConfiguration() {
+    // The file path used in ParquetRecordReader is a local file path without prefix 'file:///',
+    // so we have to make sure that the configuration item 'fs.defaultFS' is set to 'file:///'
+    // in case that user's hadoop conf overwrite this item
     Configuration conf = new Configuration();
-    conf.set("fs.defaultFS", "file:///");
+    conf.set("fs.defaultFS", DEFAULT_FS);
     conf.set("fs.file.impl", org.apache.hadoop.fs.LocalFileSystem.class.getName());
     return conf;
   }
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
@@ -22,7 +22,6 @@ import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.avro.AvroParquetReader;
 import org.apache.parquet.avro.AvroParquetWriter;
@@ -33,68 +32,33 @@ import org.apache.parquet.hadoop.ParquetWriter;
 import org.apache.parquet.hadoop.metadata.ParquetMetadata;
 
 import java.io.IOException;
-import java.util.Iterator;
 import java.util.List;
-import java.util.NoSuchElementException;
 
 
 public class ParquetUtils {
   /**
-   * Get the parquet record iterator for the given file
+   * Get a ParquetReader with the given file.
+   * @param fileName the parquet file to read
+   * @return a ParquetReader
+   * @throws IOException
    */
-  public static Iterator<GenericRecord> getParquetReader(String fileName)
+  public static ParquetReader<GenericRecord> getParquetReader(String fileName)
       throws IOException {
     Path dataFsPath = new Path(fileName);
-    Configuration conf = new Configuration();
-    ParquetReader<GenericRecord> parquet =
-        AvroParquetReader.<GenericRecord>builder(dataFsPath).disableCompatibility().withDataModel(GenericData.get())
-            .withConf(conf).build();
-
-    return new Iterator<GenericRecord>() {
-      private boolean hasNext = false;
-      private GenericRecord next = advance();
-
-      @Override
-      public boolean hasNext() {
-        return hasNext;
-      }
-
-      @Override
-      public GenericRecord next() {
-        if (!hasNext) {
-          throw new NoSuchElementException();
-        }
-
-        GenericRecord toReturn = next;
-        next = advance();
-        return toReturn;
-      }
-
-      private GenericRecord advance() {
-        try {
-          GenericRecord next = parquet.read();
-          hasNext = (next != null);
-
-          if (hasNext == false) {
-            parquet.close();
-          }
-
-          return next;
-        } catch (IOException e) {
-          throw new RuntimeException("Failed while reading parquet file: " + fileName, e);
-        }
-      }
-    };
+    return AvroParquetReader.<GenericRecord>builder(dataFsPath).disableCompatibility().withDataModel(GenericData.get())
+        .withConf(getConfiguration()).build();
   }
 
   /**
-   * Get parquet schema for the given file
+   * Read the parquet file schema
+   * @param fileName
+   * @return the parquet file schema
+   * @throws IOException
    */
   public static Schema getParquetSchema(String fileName)
       throws IOException {
     Path dataFsPath = new Path(fileName);
-    FileSystem fs = dataFsPath.getFileSystem(new Configuration());
-    ParquetMetadata footer = ParquetFileReader.readFooter(fs.getConf(), dataFsPath);
+    ParquetMetadata footer = ParquetFileReader.readFooter(getConfiguration(), dataFsPath);
 
     String schemaString = footer.getFileMetaData().getKeyValueMetaData().get("parquet.avro.schema");
     if (schemaString == null) {
@@ -115,7 +79,8 @@ public class ParquetUtils {
   public static void writeParquetRecord(String fileName, Schema schema, List<GenericRecord> records)
       throws IOException {
     ParquetWriter<GenericRecord> writer =
-        AvroParquetWriter.<GenericRecord>builder(new Path(fileName)).withSchema(schema).build();
+        AvroParquetWriter.<GenericRecord>builder(new Path(fileName)).withSchema(schema).withConf(getConfiguration())
+            .build();
 
     try {
       for (GenericRecord r : records) {
@@ -124,5 +89,12 @@ public class ParquetUtils {
     } finally {
       writer.close();
     }
+  }
+
+  private static Configuration getConfiguration() {
+    Configuration conf = new Configuration();
+    conf.set("fs.defaultFS", "file:///");
+    conf.set("fs.file.impl", org.apache.hadoop.fs.LocalFileSystem.class.getName());
+    return conf;
   }
 }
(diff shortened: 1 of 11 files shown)


pinot-core/pom.xml
```diff
@@ -213,5 +213,32 @@
...
       <artifactId>commons-math</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-avro</artifactId>
+      <version>${parquet.version}</version>
+    </dependency>
+
+    <dependency>
```
kishoreg: Hadoop brings a lot of dependencies. Its better to avoid pulling in those dependencies  Is there a way to get the schema without depending on Hadoop libs?

If we have to depend on Hadoop then its better to add code into pinot-hadoop module?

WangGuangxin: Ok. It's I'll move it to pinot-hadoop module.

pinot-core/src/main/java/org/apache/pinot/core/data/readers/ParquetRecordReader.java
```diff
@@ -0,0 +1,137 @@
+/**
```
snleee: Can you download pinot codestyle file and apply it to this file? It's located in `config/codestyle-eclipse.xml` or `config/intellij.xml`.

pinot-core/src/main/java/org/apache/pinot/core/data/readers/ParquetRecordReader.java
```diff
@@ -0,0 +1,137 @@
...
+    private final Schema _schema;
+
+    private ParquetReader<GenericRecord> _reader;
+    private GenericRecord _next;
+    private boolean _hasNext;
+
+    public ParquetRecordReader(File dataFile, Schema schema) throws IOException {
+        _dataFile = dataFile;
+        _schema = schema;
+
```
snleee: remove empty line

pinot-core/src/main/java/org/apache/pinot/core/data/readers/ParquetRecordReader.java
```diff
@@ -0,0 +1,137 @@
...
+    @Override
+    public GenericRow next() throws IOException {
+        return next(new GenericRow());
+    }
+
+    @Override
+    public GenericRow next(GenericRow reuse) throws IOException {
+        AvroUtils.fillGenericRow(_next, reuse, _schema);
+        advance();
+
```
snleee: remove line

pinot-core/pom.xml
WangGuangxin: > Hadoop brings a lot of dependencies. Its better to avoid pulling in those dependencies Is there a way to get the schema without depending on Hadoop libs?
> 
> If we have to depend on Hadoop then its better to add code into pinot-hadoop module?

I'v refactored code, put parquet related codes in `ParquetUtils.java`, which locates in `pinot-common`, since it already has `hadoop-client` dependency. 
I didn't move it to `pinot-hadoop` because there'll be cross reference between `pinot-core` and `pinot-hadoop`.

Now, `pinot-core` needs `hadoop-client` only in test scope. Is that ok?

pinot-core/src/main/java/org/apache/pinot/core/data/readers/ParquetRecordReader.java
WangGuangxin: Thanks for your review. I'v reformatted all my changes.

pinot-core/pom.xml
kishoreg: It's probably ok for now. I cant think of any other work around for now.
 
the hadoop-client dependency in common was added because of hdfs support I think. We plan to remove it once we have the support for deep storage and Uber folks stop using that feature.  Adding more code that depends on this will make it harder for us to remove the dependency on hadoop client.

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
```diff
@@ -0,0 +1,128 @@
...
+
+public class ParquetUtils {
+  /**
+   * Get the parquet record iterator for the given file
+   */
+  public static Iterator<GenericRecord> getParquetReader(String fileName)
+      throws IOException {
+    Path dataFsPath = new Path(fileName);
+    Configuration conf = new Configuration();
+    ParquetReader<GenericRecord> parquet =
```
snleee: I think that you can keep `ParquetReader<GenericRecord>` in the `ParquestRecordReader` and put the following logic in there. In that way, we don't have to go back and forth to check the actual logic. 

If this part is required for the issue of pulling dependency, can you add the comment why you need this 2 layer of record reader?

pinot-core/src/main/java/org/apache/pinot/core/data/readers/ParquetRecordReader.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class ParquetRecordReader implements RecordReader {
+  private final String _dataFilePath;
+  private final Schema _schema;
+
+  private Iterator<GenericRecord> _reader;
+
+  public ParquetRecordReader(File dataFile, Schema schema)
+      throws IOException {
+    _dataFilePath = "file://" + dataFile.getAbsolutePath();
```
snleee: Is `file://` necessary?

pinot-core/src/test/java/org/apache/pinot/core/data/readers/ParquetRecordReaderTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.commons.io.FileUtils;
+import org.apache.pinot.common.utils.ParquetUtils;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class ParquetRecordReaderTest extends RecordReaderTest {
+  private static final File TEMP_DIR = new File(FileUtils.getTempDirectory(), "ParquetRecordReaderTest");
+  private static final File DATA_FILE = new File(TEMP_DIR, "data.parquet");
+  private static final String DATA_FILE_PATH = "file://" + DATA_FILE.getAbsolutePath();
```
snleee: same here. Is `file://` required?

pinot-core/src/main/java/org/apache/pinot/core/data/readers/ParquetRecordReader.java
sunithabeeram: +1

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
WangGuangxin: Done.

pinot-core/src/main/java/org/apache/pinot/core/data/readers/ParquetRecordReader.java
WangGuangxin: It is necessary to tell hadoop file system that this is a local file. 
I deleted it here and setting the default file URI schema to 'file://' in hadoop conf, which may looks more elegant now.

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
```diff
@@ -0,0 +1,100 @@
...
+      return new Schema.Parser().parse(schemaString);
+    } else {
+      return new AvroSchemaConverter().convert(footer.getFileMetaData().getSchema());
+    }
+  }
+
+  /**
+   * write records as parquet file
+   */
+  public static void writeParquetRecord(String fileName, Schema schema, List<GenericRecord> records)
```
snleee: This function is only applicable to testing code. Can you move this part to either Util class in test module or within `ParquetRecordReaderTest`?

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
```diff
@@ -0,0 +1,100 @@
...
+    try {
+      for (GenericRecord r : records) {
+        writer.write(r);
+      }
+    } finally {
+      writer.close();
+    }
+  }
+
+  private static Configuration getConfiguration() {
```
snleee: Same here.

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
WangGuangxin: > This function is only applicable to testing code. Can you move this part to either Util class in test module or within `ParquetRecordReaderTest`?

I moved unnecessary code to `ParquetRecordReaderTest`, leaving a `getParquetWriter` in `ParquetUtils`, because it needs hadoop dependency when init a parquet writer.

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
WangGuangxin: `getConfiguration` is not used for test only. It is used by all the three functions in `ParquetUtils`

snleee: Will `fs.defaultFS` and `fs.file.impl` automatically get overridden when we run this on Hadoop?

WangGuangxin: > Will `fs.defaultFS` and `fs.file.impl` automatically get overridden when we run this on Hadoop?

yes, it will automatically get overridden.  But sometimes, users may change `fs.defaultFS` in their hadoop conf to `hdfs://`. In that case, this job will fail because the file path passed to `*RecordReader` is a local file. 
So we have to either add `file://` prefix to file path to denote that this is a local file, or make sure `fs.defaultFS` is `file://`.  How about add a comment to `getConfiguration`

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
```diff
@@ -0,0 +1,94 @@
...
+   */
+  public static ParquetWriter<GenericRecord> getParquetWriter(String fileName, Schema schema)
+      throws IOException {
+    Path dataFsPath = new Path(fileName);
+    return AvroParquetWriter.<GenericRecord>builder(dataFsPath).withSchema(schema).withConf(getConfiguration()).build();
+  }
+
+  private static Configuration getConfiguration() {
+    Configuration conf = new Configuration();
+    conf.set("fs.defaultFS", "file:///");
```
snleee: Can you add public static final variable for `DEFAULT_FILE_SYSTEM = "file:///"`?

WangGuangxin: Done.

WangGuangxin:
rebuild

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
snleee: Can you add the comment as I mentioned above?

snleee:
There's an ongoing work on modifying record reader's validation logic for time column in https://github.com/apache/incubator-pinot/pull/3895

Depending on that, you may need to add some extra validation logic for time column. Let's wait until that's resolved.

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
WangGuangxin: > Can you add the comment as I mentioned above?

Done

WangGuangxin:
> validation



> There's an ongoing work on modifying record reader's validation logic for time column in #3895
> 
> Depending on that, you may need to add some extra validation logic for time column. Let's wait until that's resolved.

ok

pinot-common/src/main/java/org/apache/pinot/common/utils/ParquetUtils.java
```diff
@@ -0,0 +1,101 @@
...
+   * @throws IOException
+   */
+  public static ParquetWriter<GenericRecord> getParquetWriter(String fileName, Schema schema)
+      throws IOException {
+    Path dataFsPath = new Path(fileName);
+    return AvroParquetWriter.<GenericRecord>builder(dataFsPath).withSchema(schema).withConf(getConfiguration()).build();
+  }
+
+  private static Configuration getConfiguration() {
+    /**
```
snleee: I'm sorry to be picky but we generally use `//` for comments inside of the function and `/** ... */` style for the comments above the function

WangGuangxin: done

snleee:
#3895 has been merged. Can you make the required changes after looking at the change happened in that pr? The main issue was to handle time column when the column name for incoming & outgoing time spec are different.

WangGuangxin:
> #3895 has been merged. Can you make the required changes after looking at the change happened in that pr? The main issue was to handle time column when the column name for incoming & outgoing time spec are different.

yes, I have merged master and make according changes. You can  check it by commit https://github.com/apache/incubator-pinot/pull/3852/commits/53753d8feaaed89b5c91b1ce0aa45e862c459cf8

snleee:
@WangGuangxin Travis test is failing. Can you double check on this? Once that passes, I will merge the pr. Thank you for working on this!

snleee:
@WangGuangxin Any update on this?

WangGuangxin:
> @WangGuangxin Travis test is failing. Can you double check on this? Once that passes, I will merge the pr. Thank you for working on this!

Sorry for the slow response. Checks passed by rebase with master branch. Thanks

snleee:
Merged. Thanks for making the change.

Jackie-Jiang:
@WangGuangxin We should not add parquet dependency into **pinot-common**. I'll move the parquet related classes into a new package **pinot-parquet**, similar to **pinot-orc**. In order to use it, pass in the record reader class path into the Hadoop configuration. Will share the pr with you.

WangGuangxin:
> @WangGuangxin We should not add parquet dependency into **pinot-common**. I'll move the parquet related classes into a new package **pinot-parquet**, similar to **pinot-orc**. In order to use it, pass in the record reader class path into the Hadoop configuration. Will share the pr with you.

ok, thanks

Jackie-Jiang:
> > @WangGuangxin We should not add parquet dependency into **pinot-common**. I'll move the parquet related classes into a new package **pinot-parquet**, similar to **pinot-orc**. In order to use it, pass in the record reader class path into the Hadoop configuration. Will share the pr with you.
> 
> ok, thanks

#4073

PR #3856, opened 2019-02-20
Check assigned instances for uploading new segment

jackjlli:
This PR improves segment validator to check assigned instances for uploading new segment.

Refactor the code in `SegmentValidator` so that the results of validation can be re-used.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
@@ -57,6 +57,7 @@ import javax.ws.rs.core.Response;
 import org.apache.commons.httpclient.HttpConnectionManager;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.helix.ZNRecord;
 import org.apache.helix.model.IdealState;
 import org.apache.pinot.common.config.TableNameBuilder;
 import org.apache.pinot.common.metrics.ControllerMeter;
@@ -72,6 +73,7 @@ import org.apache.pinot.controller.ControllerConf;
 import org.apache.pinot.controller.api.access.AccessControl;
 import org.apache.pinot.controller.api.access.AccessControlFactory;
 import org.apache.pinot.controller.api.upload.SegmentValidator;
+import org.apache.pinot.controller.api.upload.SegmentValidatorResponse;
 import org.apache.pinot.controller.api.upload.ZKOperator;
 import org.apache.pinot.controller.helix.core.PinotHelixResourceManager;
 import org.apache.pinot.controller.helix.core.PinotHelixSegmentOnlineOfflineStateModelGenerator;
@@ -315,12 +317,13 @@ public class PinotSegmentUploadRestletResource {
               clientAddress);
 
       // Validate segment
-      new SegmentValidator(_pinotHelixResourceManager, _controllerConf, _executor, _connectionManager,
-          _controllerMetrics).validateSegment(segmentMetadata, tempSegmentDir);
+      SegmentValidatorResponse segmentValidatorResponse =
+          new SegmentValidator(_pinotHelixResourceManager, _controllerConf, _executor, _connectionManager,
+              _controllerMetrics).validateSegment(segmentMetadata, tempSegmentDir);
 
       // Zk operations
       completeZkOperations(enableParallelPushProtection, headers, tempEncryptedFile, provider, segmentMetadata,
-          segmentName, zkDownloadUri, moveSegmentToFinalLocation);
+          segmentName, zkDownloadUri, moveSegmentToFinalLocation, segmentValidatorResponse);
 
       return new SuccessResponse(
           "Successfully uploaded segment: " + segmentMetadata.getName() + " of table: " + segmentMetadata
@@ -380,7 +383,7 @@ public class PinotSegmentUploadRestletResource {
 
   private void completeZkOperations(boolean enableParallelPushProtection, HttpHeaders headers, File tempDecryptedFile,
       FileUploadPathProvider provider, SegmentMetadata segmentMetadata, String segmentName, String zkDownloadURI,
-      boolean moveSegmentToFinalLocation)
+      boolean moveSegmentToFinalLocation, SegmentValidatorResponse segmentValidatorResponse)
       throws Exception {
     String finalSegmentPath = StringUtil
         .join("/", provider.getBaseDataDirURI().toString(), segmentMetadata.getTableName(),
@@ -388,7 +391,7 @@ public class PinotSegmentUploadRestletResource {
     URI finalSegmentLocationURI = new URI(finalSegmentPath);
     ZKOperator zkOperator = new ZKOperator(_pinotHelixResourceManager, _controllerConf, _controllerMetrics);
     zkOperator.completeSegmentOperations(segmentMetadata, finalSegmentLocationURI, tempDecryptedFile,
-        enableParallelPushProtection, headers, zkDownloadURI, moveSegmentToFinalLocation);
+        enableParallelPushProtection, headers, zkDownloadURI, moveSegmentToFinalLocation, segmentValidatorResponse);
   }
 
   private void decryptFile(String crypterClassHeader, File tempEncryptedFile, File tempDecryptedFile) {
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/upload/SegmentValidator.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/upload/SegmentValidator.java
@@ -20,10 +20,12 @@ package org.apache.pinot.controller.api.upload;
 
 import java.io.File;
 import java.util.Date;
+import java.util.List;
 import java.util.concurrent.Executor;
 import javax.annotation.Nonnull;
 import javax.ws.rs.core.Response;
 import org.apache.commons.httpclient.HttpConnectionManager;
+import org.apache.helix.ZNRecord;
 import org.apache.pinot.common.config.TableConfig;
 import org.apache.pinot.common.config.TableNameBuilder;
 import org.apache.pinot.common.exception.InvalidConfigException;
@@ -62,7 +64,7 @@ public class SegmentValidator {
     _controllerMetrics = controllerMetrics;
   }
 
-  public void validateSegment(SegmentMetadata segmentMetadata, File tempSegmentDir) {
+  public SegmentValidatorResponse validateSegment(SegmentMetadata segmentMetadata, File tempSegmentDir) {
     String rawTableName = segmentMetadata.getTableName();
     String offlineTableName = TableNameBuilder.OFFLINE.tableNameWithType(rawTableName);
     String segmentName = segmentMetadata.getName();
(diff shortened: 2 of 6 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/api/upload/SegmentValidatorResponse.java
```diff
@@ -0,0 +1,55 @@
...
+package org.apache.pinot.controller.api.upload;
+
+import java.util.List;
+import org.apache.helix.ZNRecord;
+import org.apache.pinot.common.config.TableConfig;
+import org.apache.pinot.controller.validation.StorageQuotaChecker;
+
+
+public class SegmentValidatorResponse {
+  private TableConfig _offlineTableConfig;
```
Jackie-Jiang: Make member variables final

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1697,21 +1717,14 @@ private void sendSegmentRefreshMessage(OfflineSegmentZKMetadata segmentZKMetadat
    * @param segmentMetadata Meta-data for the segment, used to access segmentName and tableName.
    */
   // NOTE: method should be thread-safe
-  private void addNewOfflineSegment(SegmentMetadata segmentMetadata) {
+  private void addNewOfflineSegment(SegmentMetadata segmentMetadata, List<String> assignedInstances) {
```
Jackie-Jiang: Should we allow null assignedInstances here? That might cause the problem we are trying to solve now

pinot-controller/src/main/java/org/apache/pinot/controller/api/upload/SegmentValidator.java
```diff
@@ -62,7 +67,7 @@ public SegmentValidator(PinotHelixResourceManager pinotHelixResourceManager, Con
     _controllerMetrics = controllerMetrics;
   }
 
-  public void validateSegment(SegmentMetadata segmentMetadata, File tempSegmentDir) {
+  public SegmentValidatorResponse validateSegment(SegmentMetadata segmentMetadata, File tempSegmentDir) {
```
Jackie-Jiang: Well done!

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: Yes, the variable `assignedInstances` can be null, but it can't be **empty**. 
If it's empty, it's guaranteed that the method `getAssignedInstancesForSegment` gets called but there's no assigned instance. In this case it has been handled in `SegmentValidator` before hitting this part, i.e. throwing an exception in validator.
If it's null, the caller might be from `PerfBenchmarkDriver`. And it can also be that this is an existing segment, which shouldn't be in Method `addNewOfflineSegment`. So we're good here.

pinot-controller/src/main/java/org/apache/pinot/controller/api/upload/SegmentValidatorResponse.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1686,6 +1687,25 @@ private void sendSegmentRefreshMessage(OfflineSegmentZKMetadata segmentZKMetadat
     }
   }
 
+  /**
+   * Gets assigned instances for uploading new segment. This method can be used to detect
```
sunithabeeram: Please update the comment. The intent is not to just to discover misconfigurations. You can remove the second sentence.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1461,10 +1461,11 @@ public PinotResourceManagerResponse dropTable(String tableName) {
   }
 
   public void addNewSegment(@Nonnull SegmentMetadata segmentMetadata, @Nonnull String downloadUrl) {
-    addNewSegment(segmentMetadata, downloadUrl, null);
+    addNewSegment(segmentMetadata, downloadUrl, null, null);
```
mcvsubbu: get the assigned instances here before calling the new API?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1461,10 +1461,11 @@ public PinotResourceManagerResponse dropTable(String tableName) {
...
   }
 
-  public void addNewSegment(@Nonnull SegmentMetadata segmentMetadata, @Nonnull String downloadUrl, String crypter) {
+  public void addNewSegment(@Nonnull SegmentMetadata segmentMetadata, @Nonnull String downloadUrl, String crypter,
+      List<String> assignedInstances) {
```
mcvsubbu: mark assigned instances as @ NonNull?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: This will only be called when it's a new segment. You can call getAssignedInstancesForSegment() when assignedInstances are not provided in addNewSegment(), but it should not be null here. I would recommend add a Preconditions.checkNotNull() because when we call this, the segment ZK metadata has been updated.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: Added.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: Added.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: `@Nonnull` has been added in front of the param.
And `getAssignedInstancesForSegment()` is called when `assignedInstances` not provided in Method `addNewSegment(@Nonnull SegmentMetadata segmentMetadata, @Nonnull String downloadUrl)`

jackjlli:
@sunithabeeram Unit test added.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: @Jackie-Jiang A Precondition check on `assignedInstances` is added before adding metadata to zk.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -315,12 +317,13 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
               clientAddress);
 
       // Validate segment
-      new SegmentValidator(_pinotHelixResourceManager, _controllerConf, _executor, _connectionManager,
-          _controllerMetrics).validateSegment(segmentMetadata, tempSegmentDir);
+      SegmentValidatorResponse segmentValidatorResponse =
+          new SegmentValidator(_pinotHelixResourceManager, _controllerConf, _executor, _connectionManager,
```
sunithabeeram: Just curious if we need to allocate a segmentvalidator everytime - or if we can just expose a static method on a util like class.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/PinotURIUploadIntegrationTest.java
```diff
@@ -186,6 +187,34 @@ private void verifyNRows(int currentNrows, int finalNrows)
...
+    } catch (Exception e) {
+      Assert.assertNotNull(e);
+      Assert.assertTrue(e.getMessage().contains("No assigned Instances for Segment"));
+    }
+
+    // Re-enable the server instance.
+    _helixAdmin.enableInstance(getHelixClusterName(), serverInstanceId, true);
+
+    try {
+      generateAndUploadRandomSegment(segment6, nRows1);
```
sunithabeeram: Can we split generate and upload so we can reuse the same segment here?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jackjlli: SegmentValidator also depends on other variables like `pinotHelixResourceManager`, `executor`, `connectionManager`, etc, which are not related to the segment itself. And the logic is in Resource class. I think it's best to instantiate one, since the latency of uploading an offline segment is acceptable.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/PinotURIUploadIntegrationTest.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: I agree that this should be a normal class instead of a util class. This way in the future, it will be easier to add more functionalities, or split the current one into multiple steps (e.g. checkTimeStamp(), checkResourceAllocation() etc.).

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/PinotURIUploadIntegrationTest.java
```diff
@@ -186,6 +186,38 @@ private void verifyNRows(int currentNrows, int finalNrows)
...
+    _helixAdmin.enableInstance(getHelixClusterName(), serverInstanceId, false);
+
+    final String segment6 = "segmentToBeRefreshed_6";
+    final int nRows1 = 69;
+    File segmentTarDir = generateRandomSegment(segment6, nRows1);
+    try {
+      uploadSegmentsDirectly(segmentTarDir);
+      Assert.fail("Uploading segments should fail.");
+    } catch (Exception e) {
+      Assert.assertNotNull(e);
```
Jackie-Jiang: ???

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/PinotURIUploadIntegrationTest.java
```diff
@@ -186,6 +186,38 @@ private void verifyNRows(int currentNrows, int finalNrows)
...
+      Assert.assertTrue(e.getMessage().contains("No assigned Instances for Segment"));
```
Jackie-Jiang: Don't check the exception message content. We might change the exception message in the future and that should not fail the test. The fail() in try block should be enough

jackjlli: Done.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/PinotURIUploadIntegrationTest.java
jackjlli: Done.

PR #3857, opened 2019-02-20
Fix compile failure issue when docs/_build is present

snleee:


pom.xml
```diff
@@ -1140,6 +1143,9 @@
             <!-- Sphynx auto generated makefile -->
             <exclude>docs/Makefile</exclude>
 
+            <!-- Exclude build files for docs -->
+            <exclude>docs/_build/**</exclude>
```
Jackie-Jiang: Should we ignore .rst files in rat?

snleee: From some of the voting mails that i checked, they do check if documentation files (.md or .rst) have the proper headers. So, I would prefer to keep them from the rat check.

Jackie-Jiang: But we get the following warning when we build the project:

snleee: Got it. For now, I removed the check from license-maven-plugin. But, during the release process, RAT plugin will check (RAT checks the valid license but not exact text given in `LICENSE` file.) 

`mycila/license-maven-plugin` does not natively support rst file nor support `..` comment style. http://code.mycila.com/license-maven-plugin/

PR #3865, opened 2019-02-22
only show inconsistent alert when it is daily metrics

xiaohui-sun:


harleyjj:
I wonder if this will fail integration tests.  I changed the test case to have granularity of '1_DAYS' in the PR that this adjusts.  If it fails, this is probably why.

harleyjj:
https://github.com/apache/incubator-pinot/pull/3861/files

PR #3866, opened 2019-02-22
[TE] detection - task scheduler backoff fix

jihaozh:
Fix the task scheduler backoff mechanism to prevent detection tasks pile up.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/DetectionAlertJob.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/DetectionAlertJob.java
@@ -21,11 +21,14 @@ package org.apache.pinot.thirdeye.detection.alert;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/DetectionAlertJob.java
```diff
@@ -61,6 +63,21 @@ public void execute(JobExecutionContext jobExecutionContext) throws JobExecution
 
     DetectionAlertTaskInfo taskInfo = new DetectionAlertTaskInfo(detectionAlertConfigId);
 
+    // check if a task for this detection alerter is already scheduled
+    String jobName = String.format("%s_%d", TaskConstants.TaskType.DETECTION_ALERT, detectionAlertConfigId);
+    List<TaskDTO> scheduledTasks = taskDAO.findByPredicate(Predicate.AND(
```
akshayrai: Do we know why the tasks are piling up? Is it because of one server?

Adding this condition might stop the tasks from piling up but we could delay the detection and increase the MTTD without even realizing it, right!

jihaozh: Yes. When it tries to catch up, the task takes longer to execute. Without this backoff logic, lots of duplicated tasks will pill up, which exacerbate the overloading issue.

PR #3868, opened 2019-02-22
Disable realtime aggregate metrics if multi value columns present

npawar:
Consumption will stop in the case of aggregateMetrics=true in combination with multi-value dimensions or metrics. More details in https://github.com/apache/incubator-pinot/issues/3867

This will be fixed in a separate PR. Until then, disabling the realtime aggregation in this scenario

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
@@ -668,6 +668,7 @@ public class MutableSegmentImpl implements MutableSegment {
     int i = 0;
     int[] dictIds = new int[_numKeyColumns]; // dimensions + time column.
 
+    // TODO: this for loop breaks for multi value columns. https://github.com/apache/incubator-pinot/issues/3867
     for (String column : _schema.getDimensionNames()) {
       dictIds[i++] = (Integer) dictIdMap.get(column);
     }
@@ -715,10 +716,18 @@ public class MutableSegmentImpl implements MutableSegment {
     }
 
     // All dimension columns should be dictionary encoded.
+    // All dimensions must be single value
     for (String dimension : schema.getDimensionNames()) {
       if (noDictionaryColumns.contains(dimension)) {
         _logger
             .warn("Metrics aggregation cannot be turned ON in presence of no-dictionary dimensions, eg: {}", dimension);
+        _aggregateMetrics = false;
+        break;
+      }
+      // https://github.com/apache/incubator-pinot/issues/3867
+      if (!schema.getDimensionSpec(dimension).isSingleValueField()) {
+        _logger
+            .warn("Metrics aggregation cannot be turned ON in presence of multi-value dimension columns, eg: {}", dimension);
         _aggregateMetrics = false;
         break;
       }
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -715,13 +716,21 @@ private int getOrCreateDocId(Map<String, Object> dictIdMap) {
...
+    // All dimensions must be single value
     for (String dimension : schema.getDimensionNames()) {
       if (noDictionaryColumns.contains(dimension)) {
         _logger
             .warn("Metrics aggregation cannot be turned ON in presence of no-dictionary dimensions, eg: {}", dimension);
         _aggregateMetrics = false;
         break;
       }
+      // https://github.com/apache/incubator-pinot/issues/3867
+      if (!schema.getDimensionSpec(dimension).isSingleValueField()) {
```
mayankshriv: This limitation should apply to multi-valued metrics as well, right?

npawar: i don't think so. The line where it's breaking right now is MutableSegmentImpl::672
```
for (String column : _schema.getDimensionNames()) {
      dictIds[i++] = (Integer) dictIdMap.get(column);
    }
```
It's checked only for dimensions and time. Unless I'm missing some other place?

mayankshriv: Well the aggregation code also assumes the metric is single valued, so it will fail, if I am not wrong. For example, the line below is casting the value to Integer, and would fail if the value was an int[].
indexReaderWriter.setInt(docId, (Integer) value + indexReaderWriter.getInt(docId));

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -668,6 +668,7 @@ private int getOrCreateDocId(Map<String, Object> dictIdMap) {
     int i = 0;
     int[] dictIds = new int[_numKeyColumns]; // dimensions + time column.
 
+    // TODO: this for loop breaks for multi value columns. https://github.com/apache/incubator-pinot/issues/3867
```
mcvsubbu: we dont need this TODO anymore right?

mayankshriv: No, this TODO is to make the actual support for MV dimensions in RT aggregation of metrics. This PR is just disabling it in presence of MV columns.

mcvsubbu: well, as the code stands it will never break since we disable if we have MV dimensions...but that is my ocd i guess :)

npawar: i changed TODO to FIXME

npawar:
Issue: https://github.com/apache/incubator-pinot/issues/3867

PR #3869, opened 2019-02-23
add config to control kafka fetcher size and increase default

jamesyfshao:
1. currently we hardcode the fetcher size to 500k. If this consumer fetched a larger message we will fail simple consumer silently and kafka consumer will get stuck. Our previous attempts try to fix this issue (https://github.com/apache/incubator-pinot/pull/3584) turns out to be not completed and we verify that we need one more fix

2. we increase the default from 500k to 5m to handle more use-cases and pinot consumer should handle it fine for larger memory.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaPartitionLevelConsumer.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaPartitionLevelConsumer.java
@@ -38,14 +38,21 @@ import org.slf4j.LoggerFactory;
 public class KafkaPartitionLevelConsumer extends KafkaConnectionHandler implements PartitionLevelConsumer {
   private static final Logger LOGGER = LoggerFactory.getLogger(KafkaPartitionLevelConsumer.class);
 
+  private final int _fetchRequestMinBytes;
+  private final int _fetchRequestSize;
+
   public KafkaPartitionLevelConsumer(String clientId, StreamConfig streamConfig, int partition) {
     super(clientId, streamConfig, partition, new KafkaSimpleConsumerFactoryImpl());
+    this._fetchRequestSize = getkafkaLowLevelStreamConfig().getKafkaFetcherSizeBytes();
+    this._fetchRequestMinBytes = getkafkaLowLevelStreamConfig().getKafkaFetcherMinBytes();
   }
 
   @VisibleForTesting
   public KafkaPartitionLevelConsumer(String clientId, StreamConfig streamConfig, int partition,
       KafkaSimpleConsumerFactory kafkaSimpleConsumerFactory) {
     super(clientId, streamConfig, partition, kafkaSimpleConsumerFactory);
+    this._fetchRequestSize = getkafkaLowLevelStreamConfig().getKafkaFetcherSizeBytes();
+    this._fetchRequestMinBytes = getkafkaLowLevelStreamConfig().getKafkaFetcherMinBytes();
   }
 
   /**
@@ -75,8 +82,8 @@ public class KafkaPartitionLevelConsumer extends KafkaConnectionHandler implemen
     }
 
     FetchResponse fetchResponse = _simpleConsumer.fetch(
-        new FetchRequestBuilder().minBytes(100000).maxWait(timeoutMillis)
-            .addFetch(_topic, _partition, startOffset, 500000).build());
+        new FetchRequestBuilder().minBytes(_fetchRequestMinBytes).maxWait(timeoutMillis)
+            .addFetch(_topic, _partition, startOffset, _fetchRequestSize).build());
 
     if (!fetchResponse.hasError()) {
       final Iterable<MessageAndOffset> messageAndOffsetIterable =
@@ -106,6 +113,16 @@ public class KafkaPartitionLevelConsumer extends KafkaConnectionHandler implemen
 
       return true;
     });
+  }
+
+  @VisibleForTesting
+  public int getFetchRequestSize() {
+    return _fetchRequestSize;
+  }
+
+  @VisibleForTesting
+  public int getFetchRequestMinBytes() {
+    return _fetchRequestMinBytes;
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


jamesyfshao:
@snleee @mcvsubbu could you guys help me to review this diff for fixing bug in low-level consumer?

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaPartitionLevelConsumer.java
```diff
@@ -38,14 +38,21 @@
 public class KafkaPartitionLevelConsumer extends KafkaConnectionHandler implements PartitionLevelConsumer {
   private static final Logger LOGGER = LoggerFactory.getLogger(KafkaPartitionLevelConsumer.class);
 
+  private final int _fetchRequestMinBytes;
+  private final int _fetchRequestSize;
+
   public KafkaPartitionLevelConsumer(String clientId, StreamConfig streamConfig, int partition) {
     super(clientId, streamConfig, partition, new KafkaSimpleConsumerFactoryImpl());
+    this._fetchRequestSize = getkafkaLowLevelStreamConfig().getKafkaFetcherSizeBytes();
```
snleee: Can you remove `this.`?

snleee: Can you make the variable name consistent? `fetchRequestSize vs fetcherSizeBytes`

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaPartitionLevelConsumer.java
```diff
@@ -38,14 +38,21 @@
...
+    this._fetchRequestMinBytes = getkafkaLowLevelStreamConfig().getKafkaFetcherMinBytes();
```
snleee: Same here `fetchRequestMinBytes vs fetcherMinBytes`

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaStreamConfigProperties.java
```diff
@@ -43,9 +43,12 @@
   public static class LowLevelConsumer {
     public static final String KAFKA_BROKER_LIST = "kafka.broker.list";
     public static final String KAFKA_BUFFER_SIZE = "kafka.buffer.size";
-    public static final int KAFKA_BUFFER_SIZE_DEFAULT = 512000;
+    public static final int KAFKA_BUFFER_SIZE_DEFAULT = 5120000;
```
snleee: I can see that increasing buffer size will enable to fetch larger message. Other than that, what is the impact of increasing buffer? I guess that it will increase the throughput but may impact on latency?

jamesyfshao: So in kafka consumer, there are 3 toggles to control the [consumer fetch behavior](https://kafka.apache.org/090/documentation.html#consumerconfigs):
1. fetch.min.bytes: this control when the kafka broker returns the data. This will impact the server latency (but usually minimal, like sub-seconds level) because kafka broker will try to ensure it has that many data before it return to the consumer. We are still using old config for this value.
2. fetch.message.max.bytes: this control how much data kafka broker supposes to return when it has more data. Usually, higher value can increase throughput when we have large lag (more efficient batching) and might have slight impact on consumer fetch latency (sub-second level) for sending more data if there are.
3. there is also a timeout config in stream config (fetch.timeout.millis) which allow us to control how much latency we will wait at most for kafka consumer. This will allow us to put an upper bound to consumer fetch latency if it is a concern.

Overall increasing this default value will increase latency at a minimal level (as it is bound by fetch.timeout.millis and kafka supposed to return data when reaching minBytes) and if this value is lower than kafka message size the consumer will fail silently at that message and get stuck. It would seem to be a small price to pay for getting pinot more flexible at handling large message and better catch-up performance

snleee:
@jamesyfshao Does Kafka pass any exception when the incoming message size is larger than the buffer size? This still does not fix the issue of silent failure. Logging error or throwing exception when it happens would be a good start. Ideally, we should be able to recover from the such situation and let users to pick on the behavior when such failure happens.

1. throw exception and stop consume message until a user restarts server with the new config if they don't want to skip the message.
2. If their use case can tolerate skipping messages, we need to continue to consume from the next offset.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaStreamConfigProperties.java
jamesyfshao: @mcvsubbu I have reverted the default back to the old value.

jamesyfshao:
> @jamesyfshao Does Kafka pass any exception when the incoming message size is larger than the buffer size? This still does not fix the issue of silent failure. Logging error or throwing exception when it happens would be a good start. Ideally, we should be able to recover from the such situation and let users to pick on the behavior when such failure happens.
> 
> 1. throw exception and stop consume message until a user restarts server with the new config if they don't want to skip the message.
> 2. If their use case can tolerate skipping messages, we need to continue to consume from the next offset.

Unfortunately. kafka simple consumer doesn't throw any exception while it encountered this issue. I have tested multiple version of kafka simple consumer library (from 0.9 to 1.1.1), they just simply stuck in a loop trying to read from network channel but failed due to there was no buffer left to host the next message. Eventually, there might be a socketTimeoutException after the consumer timed out from the configured value. However, this exception is fairly generic and it is very hard to tell it apart from just losing the connection from kafka broker (which is a retriable exception and reconnect to kafka should solve it). I feel we should be looking at kafka standard consumer API as simpleConsumer is marked as deprecated in later versions of Kafka (1.0+) and it seems to not getting proper bug-fix like this.

mcvsubbu:
+1 to moving to later versions of Kafka consumers.

mcvsubbu:
Thanks for reverting the default. If you can please make sure travis passes, we are good.

jamesyfshao:
> Thanks for reverting the default. If you can please make sure travis passes, we are good.

the previous fails due to jenkins not able to fetch earlier commit, I rebased and it is passing now. Thanks for the review and it would be great if you or @snleee could see if anything else we should change for this diff.

PR #3872, opened 2019-02-23
Add metrics to track work done in group-by phase

sunithabeeram:
Add code to log and emit metrics related to group-by execution:
- Number of groups ignored when generating group-by keys per segment (IgnoredPreCombine)
- Number of groups ignored during the combine phase (IgnoredInCombine)
- Number of groups aggregated during combine phase (AggrInCombine)
- Number of groups ignored/trimmed after aggregation (PostCombine)


kishoreg:
Closing this since we have groupByLimit reached flag in the response

PR #3875, opened 2019-02-25
[TE] detection add upper/lower bounds to detection output

xiaohui-sun:
1. Added "DetectionOutput" class which contains list of anomalies and "TimeSeries".
2. Add upper/lower bounds to "TimeSeries".


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
```diff
@@ -31,6 +32,7 @@
 
   Map<String, Object> diagnostics;
   List<MergedAnomalyResultDTO> anomalies;
+  TimeSeries predictions;
```
jihaozh: Need a Map from string to TimeSeries here. We need a predicted time series per detection rule and dimension value.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -143,12 +145,15 @@ public DetectionPipelineResult run() throws Exception {
...
+    TimeSeries predictions = new TimeSeries();
     for (Interval window : monitoringWindows) {
       List<MergedAnomalyResultDTO> anomaliesForOneWindow = new ArrayList<>();
       try {
         LOG.info("[New Pipeline] running detection for config {} metricUrn {}. start time {}, end time {}", config.getId(), metricUrn, window.getStart(), window.getEnd());
         long ts = System.currentTimeMillis();
-        anomaliesForOneWindow = anomalyDetector.runDetection(window, this.metricUrn);
+        DetectionOutput output = anomalyDetector.runDetection(window, this.metricUrn);
+        anomaliesForOneWindow = output.getAnomalies();
+        predictions.merge(output.getPredictions());
```
jihaozh: The windows in `monitoringWindows` are overlapped with each other. doing a merge/append may not get the right result. Unless the moving window is moved into the library, then we can simply get the predictions for the whole detection range.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -164,7 +169,7 @@ public DetectionPipelineResult run() throws Exception {
       anomaly.setDimensions(DetectionUtils.toFilterMap(this.metricEntity.getFilters()));
       anomaly.getProperties().put(PROP_DETECTOR_COMPONENT_NAME, this.detectorName);
     }
-    return new DetectionPipelineResult(anomalies, this.getLastTimeStamp());
+    return new DetectionPipelineResult(anomalies, predictions, this.getLastTimeStamp());
```
jihaozh: In the merger, dimension explorer, and anomaly filter wrapper need to handle this predicted time series too.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
xiaohui-sun: Why the windows in monitoringWindows are overlapped?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
xiaohui-sun: Why merger needs to handle this? Merger's input is anomalies only.

xiaohui-sun:
> Will the upper and lower bounds apply to all output coming from BaselineProviders? If so, when the baseline is stored in anomaly results, it is easily to be overflowed in DB. I think we should think of when we need the confidence intervals and when to store them.

xiaohui-sun:
@cyy0714 The upper/lower bound is optional which means you don't need to store it in Anomaly.
The main purpose is to use in preview and RCA.

cyy0714:
> @cyy0714 The upper/lower bound is optional which means you don't need to store it in Anomaly.
> The main purpose is to use in preview and RCA.

For RCA, how is the baseline calculated? If we don't use the sliding window fashion as preview, we will lose the accuracy.

xiaohui-sun:
> > @cyy0714 The upper/lower bound is optional which means you don't need to store it in Anomaly.
> > The main purpose is to use in preview and RCA.
> 
> For RCA, how is the baseline calculated? If we don't use the sliding window fashion as preview, we will lose the accuracy.

I believe it is not sliding window but that is another issue we should fix.

PR #3877, opened 2019-02-25
Pinot controller side change to enhance LLC segment metadata upload.

chenboat:
As designed by @npawar in an earlier [PR](https://github.com/apache/incubator-pinot/pull/3748), add a new POST segmentCommitEndWithMetadata endpoint to Pinot controller. This endpoint allows the Pinot controller to upload segment metadata directly in the commitEnd step. This enhancement allows the controller to skip extract metadata again from the segment files.

**Manual tests performed:** Combined with the https://github.com/apache/incubator-pinot/pull/3941 to test the new segmentCommitEndWithMetadata() endpoint.


pinot-common/src/main/java/org/apache/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -200,6 +203,7 @@ public String getUrl(String hostPort, String protocol) {
       private String _segmentLocation;
       private long _memoryUsedBytes;
       private long _segmentSizeBytes;
+      private SegmentMetadata _segmentMetadata;
```
mcvsubbu: You should not be sending segment metadata as a query parameter.

mcvsubbu:
Also, can you modify the existing SegmentCommitEnd to take on an optional 2-part form? https://github.com/apache/incubator-pinot/pull/3748#issuecomment-461229243

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +280,79 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+      FileUtils.deleteQuietly(tempMetadataDir);
+    }
+  }
+
+  private void extractMetadataFromInputForm(FormDataMultiPart metadataFiles, String tempMetadataDirStr,
+                                            String metaFileName) throws IOException {
+    FormDataBodyPart metadataFilesField = metadataFiles.getField(metaFileName);
+    Preconditions.checkNotNull(metadataFilesField, "The metadata input field %s does not exist.",
+            metaFileName);
+    InputStream metadataPropertiesInputStream = metadataFilesField.getValueAs(InputStream.class);
```
npawar: Perhaps put the InputStream creation in a try-with-resource block so that it gets closed ?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +280,79 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+            .withMemoryUsedBytes(memoryUsedBytes).withSegmentMetadata(segmentMetadata);
+    LOGGER.info("Processing segmentCommitEnd:{}", requestParams.toString());
+
+
+    final boolean isSuccess = true;
+    final boolean isSplitCommit = true;
+
+    SegmentCompletionProtocol.Response response =
+            SegmentCompletionManager.getInstance().segmentCommitEnd(requestParams, isSuccess, isSplitCommit);
+    final String responseStr = response.toJsonString();
```
npawar: How about adding it as a parameter to SegmentCompletionManager::segmentCommitEnd? That way we don't have to put SegmentMetadata in the SegmentCompletionProtocol.Request.Params . I see that the way you've used it is for an internal call, and not really a query param. However, putting it here could make someone accidently use it in non internal calls as a query param.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +280,79 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+                                            String metaFileName) throws IOException {
+    FormDataBodyPart metadataFilesField = metadataFiles.getField(metaFileName);
+    Preconditions.checkNotNull(metadataFilesField, "The metadata input field %s does not exist.",
+            metaFileName);
+    InputStream metadataPropertiesInputStream = metadataFilesField.getValueAs(InputStream.class);
+    Preconditions.checkNotNull(metadataPropertiesInputStream, "Unable to parse %s from input.",
+            metaFileName);
+    java.nio.file.Path metadataPropertiesPath =
+            FileSystems.getDefault().getPath(tempMetadataDirStr, metaFileName);
+    Files.copy(metadataPropertiesInputStream, metadataPropertiesPath);
```
npawar: We should have some exception handling in this method as well, so that we know which file is having problems

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
npawar: We could also create CommittingSegmentDescriptor right here, and pass that down. Just discussed this approach with @mcvsubbu  as well and we think it'll be the better way to go ahead

npawar:
+1 to @mcvsubbu's comment about adding it in the existing call instead of having a new segmentCommitEndWithMetadata. It would much simpler to migrate, and we'll avoid a lot of code duplication

chenboat:
> +1 to @mcvsubbu's comment about adding it in the existing call instead of having a new segmentCommitEndWithMetadata. It would much simpler to migrate, and we'll avoid a lot of code duplication

I think @mcvsubbu was asking to add an optional form to the existing segmentCommitEnd() and add a new POST endpoint segmentCommitEndWithMetadata() ?

The current segmentCommitEnd() is a GET.  If I make it a POST (which seems the right type to me) and add the form to it, would it be backward compatible?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-common/src/main/java/org/apache/pinot/common/protocols/SegmentCompletionProtocol.java
chenboat: Done.

chenboat:
> Also, can you modify the existing SegmentCommitEnd to take on an optional 2-part form? [#3748 (comment)](https://github.com/apache/incubator-pinot/pull/3748#issuecomment-461229243)

To clarify. The current SegmentCommitEnd() is a GET method. Can a GET method take a form? It looks to me abuse of GET. If we change to POST, would that be a back compatibility issue?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -544,8 +547,18 @@ protected boolean updateOldSegmentMetadataZNRecord(String realtimeTableName, LLC
...
-    SegmentMetadataImpl segmentMetadata = extractSegmentMetadata(rawTableName, committingSegmentNameStr);
+    // Pull segment metadata from committingSegmentDescriptor if present. Otherwise extract it from the incoming segment
+    // and set it in zk segment metadata.
+    // Note: ideally we should use the interface SegmentMetadata here for the class of segmentMetadata. However,
+    // SegmentMetadataImpl has member like ColumnMetadata which is not in the interface. Adding these members to the
+    // interface will introduce circular dependency between Pinot-Core and Pinot-common.
+    SegmentMetadataImpl segmentMetadata;
+    if (committingSegmentDescriptor.getSegmentMetadata() != null &&
+        committingSegmentDescriptor.getSegmentMetadata() instanceof SegmentMetadataImpl) {
+      segmentMetadata = (SegmentMetadataImpl)committingSegmentDescriptor.getSegmentMetadata();
```
mcvsubbu: Do we need the cast here? Maybe we can just define the field in commiting segment descriptor as SegmentMetadataImpl

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -544,8 +547,18 @@ protected boolean updateOldSegmentMetadataZNRecord(String realtimeTableName, LLC
...
+    } else {
+      segmentMetadata = extractSegmentMetadata(rawTableName, committingSegmentNameStr);
```
mcvsubbu: Can you look to moving this out so that segment metadata is extracted in LLCSegmentCompletionHandlers? We can unitfy the segmentCommit path so that if metadata is not present, segmentCommit() downloads the segment and extract it.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +282,91 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+  }
+
+  private SegmentMetadataImpl extractSegmentMetadata(FormDataMultiPart metadataFiles, String segmentNameStr) {
+    String tempMetadataDirStr = StringUtil.join("/", _controllerConf.getLocalTempDir(), segmentNameStr + METADATA_TEMP_DIR_SUFFIX);
+    File tempMetadataDir = new File(tempMetadataDirStr);
+
+    try {
+      Preconditions.checkState(tempMetadataDir.mkdirs(), "Failed to create directory: %s", tempMetadataDirStr);
+      // Extract metadata.properties from the metadataFiles.
+      if (extractMetadataFromInputForm(metadataFiles, tempMetadataDirStr, V1Constants.MetadataKeys.METADATA_FILE_NAME)) {
```
npawar: extractMetadataFromInputForm returns true on success and false on failure, is that correct? Then we shouldn't be returning null if true right? Here and also for the call below

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +282,91 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+      // Load segment metadata
+      return new SegmentMetadataImpl(tempMetadataDir);
+    } catch (Exception e) {
+      throw new RuntimeException("Exception extracting and reading segment metadata for " + segmentNameStr, e);
+    } finally {
+      FileUtils.deleteQuietly(tempMetadataDir);
+    }
+  }
+
+  private boolean extractMetadataFromInputForm(FormDataMultiPart metadataFiles, String tempMetadataDirStr,
```
npawar: Please add some javadoc, specifically to explain what is returned in which scenario

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +282,91 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+        return null;
+      }
+      // Extract creation.meta from the metadataFiles.
+      if (extractMetadataFromInputForm(metadataFiles, tempMetadataDirStr, V1Constants.SEGMENT_CREATION_META)) {
+        return null;
+      }
+      // Load segment metadata
+      return new SegmentMetadataImpl(tempMetadataDir);
+    } catch (Exception e) {
+      throw new RuntimeException("Exception extracting and reading segment metadata for " + segmentNameStr, e);
```
npawar: If we encounter this exception, what happens to the server who is waiting for a response? We should return FAILED response to the server.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +282,91 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+    File tempMetadataDir = new File(tempMetadataDirStr);
+
+    try {
+      Preconditions.checkState(tempMetadataDir.mkdirs(), "Failed to create directory: %s", tempMetadataDirStr);
+      // Extract metadata.properties from the metadataFiles.
+      if (extractMetadataFromInputForm(metadataFiles, tempMetadataDirStr, V1Constants.MetadataKeys.METADATA_FILE_NAME)) {
+        return null;
+      }
+      // Extract creation.meta from the metadataFiles.
+      if (extractMetadataFromInputForm(metadataFiles, tempMetadataDirStr, V1Constants.SEGMENT_CREATION_META)) {
```
npawar: There are 3 scenarios here
1) Happy case - returned the metadata
2) error extracting one of the 2 files - returning null
3) error creating SegmentMetadata - throwing exception
Maybe throwing exception isn't what we want? We should log the error and proceed with the metadata download, like case 2 would have done

pinot-common/src/main/java/org/apache/pinot/common/protocols/SegmentCompletionProtocol.java
```diff
@@ -23,6 +23,8 @@
 import com.fasterxml.jackson.core.JsonProcessingException;
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
+
+import org.apache.pinot.common.segment.SegmentMetadata;
```
npawar: unused import?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
chenboat: Move segment metadata extraction to the LLCSegmentCompletionHandler.

npawar:
> > Also, can you modify the existing SegmentCommitEnd to take on an optional 2-part form? [#3748 (comment)](https://github.com/apache/incubator-pinot/pull/3748#issuecomment-461229243)
> 
> To clarify. The current SegmentCommitEnd() is a GET method. Can a GET method take a form? It looks to me abuse of GET. If we change to POST, would that be a back compatibility issue?

Ah yes, existing method is GET. In that case I think what you have is the best way

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: You are right. Flip the boolean expr.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Right. I made 2 and 3 both return null.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Resolved by returning null as pointed in the comment below.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +296,152 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+
+    try (InputStream metadataPropertiesInputStream = metadataFilesField.getValueAs(InputStream.class)) {
+      Preconditions.checkNotNull(metadataPropertiesInputStream, "Unable to parse %s from input.",
+              metaFileName);
+      java.nio.file.Path metadataPropertiesPath =
+              FileSystems.getDefault().getPath(tempMetadataDirStr, metaFileName);
+      Files.copy(metadataPropertiesInputStream, metadataPropertiesPath);
+      return true;
+    } catch (IOException e) {
+      LOGGER.error("Failed to copy metadata property file: {}", metaFileName);
```
npawar: Can we also add the exception e to the log?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,6 +296,152 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+            .withMemoryUsedBytes(memoryUsedBytes);
+    LOGGER.info("Processing segmentCommitEnd:{}", requestParams.toString());
+
+
+    final boolean isSuccess = true;
+    final boolean isSplitCommit = true;
+    SegmentMetadataImpl segmentMetadata = extractMetadataFromInput(metadataFiles, segmentName);
+    // If it fails to extract metadata from the input form, try to download the segment and extract it from the segment.
+    if (segmentMetadata == null) {
+      segmentMetadata = extractMetadataFromSegmentFile(segmentName);
```
npawar: minor: can we have log statements in both these if blocks?

chenboat:
> Minor comments. LGTM otherwise

Thanks. Additional logs added.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -236,9 +251,23 @@ public String segmentCommit(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
     if (response.equals(SegmentCompletionProtocol.RESP_COMMIT_CONTINUE)) {
       // Get the segment and put it in the right place.
-      boolean success = uploadSegment(multiPart, instanceId, segmentName, false) != null;
-
-      response = segmentCompletionManager.segmentCommitEnd(requestParams, success, false);
+      URI segmentURI = uploadSegment(multiPart, instanceId, segmentName, false);
+      boolean success =  (segmentURI != null);
+      SegmentMetadataImpl segmentMetadata;
+      try {
+        segmentMetadata = extractMetadataFromLocalSegmentFile(segmentName, segmentURI.getPath());
```
mcvsubbu: We should attempt to get metadata from the multi-part in this method as well. That way, we can add the server code to send multi-part form with metadata and everything will work. If metadata is not present in multi-part then we need to extract from local.
You can add this as a separate PR before moving on to other parts, or add it now, your choice.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,8 +305,158 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+   * table.
+   * <p>Segment tar-zipped file path: segmentLocation.
+   * <p>We extract the metadata.properties and creation.meta into a temporary metadata directory:
+   * DATADIR/rawTableName/segmentName.metadata.tmp, and load metadata from there.
+   *
+   * @param segmentNameStr Name of the segment
+   * @param segmentLocation the location of the segment file.
+   * @return SegmentMetadataImpl if it is able to extract the metadata file from the tar-zipped segment file.
+   */
+  private SegmentMetadataImpl extractMetadataFromLocalSegmentFile(final String segmentNameStr,
```
mcvsubbu: pls rename to extractMetadataFromSegmentFile. The segment file is in deep store, right?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
mcvsubbu: Actually, disregard this comment. If the segment is there, we don't need to push metadata, so you are good here.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: No. The segment file in this function is on local mounted FS. This should work on the current clients using local FS. I will modify this function so that it also works for clients on deep storage.

chenboat:
> Please note any manual testing you have done in the PR.
> Here is the test I would suggest. Run LLRealtimeClusterIntegration test in debug mode on your IDE, and issue manual curl calls to start/upload-data and end commit with metadata. Also, if you modify segmentCommit end point, then test that as well the same way.
> 
> These tests will make sure that before we add the server side, the code we have on the controller work right.
> 
> When we add the server side, we should add integration tests for the whole package.
> 
> thanks

I rely on two integration tests: **LLCRealtimeClusterIntegrationTest** and **LLCRealtimeClusterSplitCommitIntegrationTest** which covers segmentCommit and segmentCommitEnd http rest points. For extra safety I used debug modes to check the metadata extraction and it looks good.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Change done to make the method work for Deep storage. Note this function still download the segment file but it is not primary means to extract metadata now.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -204,8 +204,8 @@ public String segmentCommitEnd(@QueryParam(SegmentCompletionProtocol.PARAM_INSTA
       return SegmentCompletionProtocol.RESP_FAILED.toJsonString();
     }
 
-    SegmentMetadataImpl segmentMetadata = extractMetadataFromLocalSegmentFile(segmentName,
-            ControllerConf.getUriFromPath(segmentLocation).getPath());
+    SegmentMetadataImpl segmentMetadata = extractMetadataFromSegmentFile(segmentName,
+            ControllerConf.getUriFromPath(segmentLocation));
```
mcvsubbu: segmentLocation should already be a well-formed URI.  See the segmentUpload endpoint.

chenboat: I still prefer to use a URI to pass into 
extractMetadataFromSegmentFile(.,.) because PinotFS expects an URI in general.  Construct a URI directly instead of calling 
ControllerConf.getUriFromPath (which is essentially a thin wrapper).

mcvsubbu: This is a URI returned by the controller in the segmentUpload call. The servers cannot send some random URIs to commit. So, to PinotFS, you should be able to pass the URI from new URI(segmentLocation)

chenboat: @mcvsubbu My [latest commit ](https://github.com/apache/incubator-pinot/pull/3877/commits/abba120a37dd8c06fb76bcf885a7e7ee45b7f27b) already passed in 
new URI(segmentLocation) as suggested. Please take a look.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -236,9 +257,19 @@ public String segmentCommit(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
     SegmentCompletionProtocol.Response response = segmentCompletionManager.segmentCommitStart(requestParams);
     if (response.equals(SegmentCompletionProtocol.RESP_COMMIT_CONTINUE)) {
       // Get the segment and put it in the right place.
-      boolean success = uploadSegment(multiPart, instanceId, segmentName, false) != null;
+      URI segmentURI = uploadSegment(multiPart, instanceId, segmentName, false);
```
mcvsubbu: Please correct me if I am wrong, but it seems that we get the segment to a tmp dir locally, store it in the file system, download it again, and retrieve the metadata. We should be retrieving metadata before we upload,.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -197,6 +205,16 @@ public String segmentCommitEnd(@QueryParam(SegmentCompletionProtocol.PARAM_INSTA
       return SegmentCompletionProtocol.RESP_FAILED.toJsonString();
     }
 
+    SegmentMetadataImpl segmentMetadata;
+    try {
+      segmentMetadata = extractMetadataFromSegmentFile(segmentName, new java.net.URI(segmentLocation));
```
mcvsubbu: Can you please update the rest of the file to use java.net.URI as well? (and therefore import the right class and not qualify it everywhere)? It should be ok since the server only treats this as a string, so no compat issues.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done. Now in uploadSegment(), the metadata will be extracted directly for the uploaded file if it is not a split commit.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Use java.net.URI uniformly now. In fact, my test shows that the original apache uri usage may not work for URI with % char in it. Please see the two added tests in LLCRealtimeClusterSplitCommitIntegrationTest() for details. They can be deleted when this PR is submitted.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -19,11 +19,12 @@
 package org.apache.pinot.controller.api.resources;
 
 import com.google.common.annotations.VisibleForTesting;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
+
+import java.io.*;
```
mcvsubbu: Can we have individual imports as needed instead of wild-card? thanks

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -58,8 +65,10 @@
 @Path("/")
 public class LLCSegmentCompletionHandlers {
 
+  public static final String SEGMENT_TMP_DIR = "segment.tmp";
```
mcvsubbu: private is fine, right? We don't use it outside this file

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,10 +307,193 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+              FileSystems.getDefault().getPath(tempMetadataDirStr, V1Constants.MetadataKeys.METADATA_FILE_NAME);
+      Files.copy(metadataPropertiesInputStream, metadataPropertiesPath);
+
+      Preconditions.checkNotNull(creationMetaInputStream, "%s does not exist", V1Constants.SEGMENT_CREATION_META);
+      java.nio.file.Path creationMetaPath = FileSystems.getDefault().getPath(tempMetadataDirStr, V1Constants.SEGMENT_CREATION_META);
+      Files.copy(creationMetaInputStream, creationMetaPath);
+      // Load segment metadata
+      return new SegmentMetadataImpl(tempMetadataDir);
+    } catch (Exception e) {
+      LOGGER.error("Exception extracting and reading segment metadata for " + segmentName.getSegmentName(), e);
```
mcvsubbu: Please use parameters for logging not concatenation

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,10 +307,193 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+    File tempSegmentDataDir = new File(tempSegmentDataDirStr);
+    File segDstFile = new File(StringUtil.join("/", tempSegmentDataDirStr, segmentNameStr));
+    // Use PinotFS to copy the segment file to local fs for metadata extraction.
+    PinotFS pinotFS = PinotFSFactory.create(ControllerConf.getUriFromPath(_controllerConf.getDataDir()).getScheme());
+    try {
+      Preconditions.checkState(tempSegmentDataDir.mkdirs(), "Failed to create directory: %s", tempSegmentDataDir);
+      pinotFS.copyToLocalFile(segmentLocation, segDstFile);
+      return getSegmentMetadataFromLocalFile(segmentName, segDstFile);
+    } catch (Exception e) {
+      LOGGER.error("Exception copy segment file to local " + segmentNameStr, e);
```
mcvsubbu: please use {} to log segmentNameStr

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,10 +307,193 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+        return null;
+      }
+      // Extract creation.meta from the metadataFiles.
+      if (!extractMetadataFromInputField(metadataFiles, tempMetadataDirStr, V1Constants.SEGMENT_CREATION_META)) {
+        return null;
+      }
+      // Load segment metadata
+      return new SegmentMetadataImpl(tempMetadataDir);
+    } catch (Exception e) {
+      LOGGER.error("Exception extracting and reading segment metadata for " + segmentNameStr, e);
```
mcvsubbu: Please use {} for logging segmentNamemStr

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,10 +307,193 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+    final boolean isSplitCommit = true;
+    SegmentMetadataImpl segmentMetadata = extractMetadataFromInput(metadataFiles, segmentName);
+    // If it fails to extract metadata from the input form, try to download the segment and extract it from the segment.
+    if (segmentMetadata == null) {
+      LOGGER.info("Failed to extract segment metadata for {} from input form, fallback to use the segment file.",
+              segmentName);
+      try {
+        segmentMetadata = extractMetadataFromSegmentFile(segmentName, new URI(segmentLocation));
+      } catch (URISyntaxException e) {
+        LOGGER.error("Invalid segment location: ", segmentLocation);
```
mcvsubbu: Missing {}

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -236,9 +258,18 @@ public String segmentCommit(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
     SegmentCompletionProtocol.Response response = segmentCompletionManager.segmentCommitStart(requestParams);
     if (response.equals(SegmentCompletionProtocol.RESP_COMMIT_CONTINUE)) {
       // Get the segment and put it in the right place.
-      boolean success = uploadSegment(multiPart, instanceId, segmentName, false) != null;
+      ImmutablePair<URI, SegmentMetadataImpl>  uploadResult = uploadSegment(multiPart, instanceId, segmentName, false);
+      boolean success =  (uploadResult != null && uploadResult.getLeft() != null && uploadResult.getRight() != null);
+      SegmentMetadataImpl segmentMetadata = null;
```
mcvsubbu: we can fail the commit if segment metadata is null.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -260,13 +291,13 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
     requestParams.withInstanceId(instanceId).withSegmentName(segmentName).withOffset(offset);
     LOGGER.info("Processing segmentUpload:{}", requestParams.toString());
 
-    final String segmentLocation = uploadSegment(multiPart, instanceId, segmentName, true);
-    if (segmentLocation == null) {
+    final ImmutablePair<URI, SegmentMetadataImpl> uploadResults = uploadSegment(multiPart, instanceId, segmentName, true);
```
mcvsubbu: we should not be extracting metadata in this handler, since we are doing that when the commit end happens.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,10 +307,193 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+
+    final boolean isSuccess = true;
+    final boolean isSplitCommit = true;
+    SegmentMetadataImpl segmentMetadata = extractMetadataFromInput(metadataFiles, segmentName);
+    // If it fails to extract metadata from the input form, try to download the segment and extract it from the segment.
+    if (segmentMetadata == null) {
+      LOGGER.info("Failed to extract segment metadata for {} from input form, fallback to use the segment file.",
+              segmentName);
+      try {
+        segmentMetadata = extractMetadataFromSegmentFile(segmentName, new URI(segmentLocation));
```
mcvsubbu: we should fail the commit if we cannot extract metadata. If a server posted to this URI it better include metadata in the request.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: No. I am not doing metadata extracting in this handler. There is a control flag isSplitCommit in calling the uploadSegment(). If it is true, there will be no metadata extraction. Note that we pass true at this line.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Agree. I suppose the code is already doing what you suggested? In the lines follows:

    // Return failure to server if both extraction efforts fail.
    if (segmentMetadata == null) {
      LOGGER.warn("Segment metadata extraction failure for segment {}", segmentName);
      return SegmentCompletionProtocol.RESP_FAILED.toJsonString();
    }

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: I got what you were saying. Simplify the logic and return failure if there is no files in the input.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/LLCRealtimeClusterSplitCommitIntegrationTest.java
```diff
@@ -41,4 +48,57 @@ public void startServer() {
     serverConfig.setProperty(CommonConstants.Server.CONFIG_OF_ENABLE_SPLIT_COMMIT, true);
     startServer(serverConfig);
   }
+  // The goals of the following tests to show the diff between org.apache.commons.httpclient.URI and java.net.URI.
```
mcvsubbu: Thanks so much for demonstrating this in an automated fashion. Yes, you can remove these tests in your next iteration.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -276,10 +317,182 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+   *
+   * @param multiPart input uploaded files.
+   * @param instanceId
+   * @param segmentName
+   * @param isSplitCommit
+   * @return (1) the URI for the uploaded segment file and; (2) for a non-split commit, the corresponding segment metadata.
+   */
   @Nullable
-  private String uploadSegment(FormDataMultiPart multiPart, String instanceId, String segmentName,
+  private ImmutablePair<URI, SegmentMetadataImpl> uploadSegment(FormDataMultiPart multiPart, String instanceId, String segmentName,
```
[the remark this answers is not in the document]
chenboat: Done refactoring.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -19,15 +19,21 @@
 package org.apache.pinot.controller.api.resources;
 
 import com.google.common.annotations.VisibleForTesting;
+
+
```
mcvsubbu: remove extra newine

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -254,31 +311,212 @@ public String segmentCommit(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+      Files.copy(metadataPropertiesInputStream, metadataPropertiesPath);
+      return true;
+    } catch (IOException e) {
+      LOGGER.error("Failed to copy metadata property file: {}", metaFileName, e);
+    }
+    return false;
+  }
+
+  /**
+   * Extract the segment metadata files from the tar-zipped segment file
```
mcvsubbu: wrong comment. we are extracting metadata from a segment located in PinotFS

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -219,30 +245,61 @@ public String segmentCommitEnd(@QueryParam(SegmentCompletionProtocol.PARAM_INSTA
...
   @Produces(MediaType.APPLICATION_JSON)
   public String segmentCommit(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE_ID) String instanceId,
-      @QueryParam(SegmentCompletionProtocol.PARAM_SEGMENT_NAME) String segmentName,
-      @QueryParam(SegmentCompletionProtocol.PARAM_OFFSET) long offset,
-      @QueryParam(SegmentCompletionProtocol.PARAM_MEMORY_USED_BYTES) long memoryUsedBytes,
-      @QueryParam(SegmentCompletionProtocol.PARAM_BUILD_TIME_MILLIS) long buildTimeMillis,
-      @QueryParam(SegmentCompletionProtocol.PARAM_WAIT_TIME_MILLIS) long waitTimeMillis,
-      @QueryParam(SegmentCompletionProtocol.PARAM_SEGMENT_SIZE_BYTES) long segmentSizeBytes,
-      @QueryParam(SegmentCompletionProtocol.PARAM_ROW_COUNT) int numRows, FormDataMultiPart multiPart) {
+                              @QueryParam(SegmentCompletionProtocol.PARAM_SEGMENT_NAME) String segmentName,
```
mcvsubbu: please use the coding guidelines as in https://pinot.readthedocs.io/en/latest/dev_env.html

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -182,33 +195,46 @@ public String segmentCommitStart(@QueryParam(SegmentCompletionProtocol.PARAM_INS
...
+              instanceId, segmentLocation);
       // TODO: memoryUsedInBytes = 0 if not present in params. Add validation when we start using it
       return SegmentCompletionProtocol.RESP_FAILED.toJsonString();
     }
 
+    SegmentMetadataImpl segmentMetadata;
+    try {
+      segmentMetadata = extractMetadataFromSegmentFile(segmentName, new URI(segmentLocation));
+    } catch (URISyntaxException e) {
+      LOGGER.error("Invalid segment location: ", segmentLocation);
```
mcvsubbu: Please include segment name in this log. The idea is that we should be able to grep a segment name in the logs and find out what happened to it.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -360,4 +546,47 @@ private String uploadSegment(FormDataMultiPart multiPart, String instanceId, Str
...
+      }
+      pinotFS.copyFromLocalFile(localTmpFile, segmentFileURI);
+    }
+  }
+
+  private URI localSegementFileToPinotFsTmpLocation(FileUploadPathProvider provider, File localTmpFile,
+                                                    String segmentName) throws Exception {
+    final String rawTableName = new LLCSegmentName(segmentName).getTableName();
+    // We only clean up tmp segment file under table dir, so don't create any sub-dir under table dir.
+    // See PinotLLCRealtimeSegmentManager.commitSegmentFile().
```
mcvsubbu: Please include the TODO here regarding moving the tmpfile logic to SegmentCompletionUtils

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -254,31 +311,212 @@ public String segmentCommit(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+
+    try (InputStream metadataPropertiesInputStream = metadataFilesField.getValueAs(InputStream.class)) {
+      Preconditions.checkNotNull(metadataPropertiesInputStream, "Unable to parse %s from input.",
+              metaFileName);
+      java.nio.file.Path metadataPropertiesPath =
+              FileSystems.getDefault().getPath(tempMetadataDirStr, metaFileName);
+      Files.copy(metadataPropertiesInputStream, metadataPropertiesPath);
+      return true;
+    } catch (IOException e) {
+      LOGGER.error("Failed to copy metadata property file: {}", metaFileName, e);
```
mcvsubbu: Include segment name here. You may have to pass it as parameter, but worth it.
Also, change 'copy' to 'extract'

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -254,31 +311,212 @@ public String segmentCommit(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+    File tempSegmentDataDir = new File(tempSegmentDataDirStr);
+    File segDstFile = new File(StringUtil.join("/", tempSegmentDataDirStr, segmentNameStr));
+    // Use PinotFS to copy the segment file to local fs for metadata extraction.
+    PinotFS pinotFS = PinotFSFactory.create(ControllerConf.getUriFromPath(_controllerConf.getDataDir()).getScheme());
+    try {
+      Preconditions.checkState(tempSegmentDataDir.mkdirs(), "Failed to create directory: %s", tempSegmentDataDir);
+      pinotFS.copyToLocalFile(segmentLocation, segDstFile);
+      return getSegmentMetadataFromLocalFile(segmentName, segDstFile);
+    } catch (Exception e) {
+      LOGGER.error("Exception copy segment file to local {}",  segmentNameStr, e);
```
mcvsubbu: Include segment name in the log, change 'copy' to 'extract'

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
chenboat: Reformat with guideline applied.

chenboat:
> Can you make sure that you have the segment name logged in all the new log messages you added? I tried to note it, but I may have missed some.

ACK.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/LLCRealtimeClusterSplitCommitIntegrationTest.java
```diff
@@ -20,7 +20,14 @@
 
 import org.apache.commons.configuration.Configuration;
 import org.apache.pinot.common.utils.CommonConstants;
+import org.apache.pinot.common.utils.StringUtil;
```
mcvsubbu: Looks like there are no change to this file. Can we revert this file to the original please?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -234,13 +257,43 @@ public String segmentCommit(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
 
-      response = segmentCompletionManager.segmentCommitEnd(requestParams, success, false);
+    CommittingSegmentDescriptor committingSegmentDescriptor =
+        CommittingSegmentDescriptor.fromSegmentCompletionReqParams(requestParams);
+    boolean success = false;
+
+    if (response.equals(SegmentCompletionProtocol.RESP_COMMIT_CONTINUE)) {
+      File localTmpFile = null;
+      try {
+        // Get the segment from the form input and put it in the right place.
```
mcvsubbu: Wrong comment. Get the segment and put it in a tmp area in the local file system

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -260,25 +313,203 @@ public String segmentUpload(@QueryParam(SegmentCompletionProtocol.PARAM_INSTANCE
...
+            CommittingSegmentDescriptor.fromSegmentCompletionReqParamsAndMetadata(requestParams, segmentMetadata));
+    final String responseStr = response.toJsonString();
+    LOGGER.info("Response to segmentCommitEnd:{}", responseStr);
+    return responseStr;
+  }
+
+  /**
+   * Extract and return the segment metadata from the two input form data files (metadata file and creation meta).
+   * Return null if any of the two files is missing or there is exception during parsing and extraction.
+   *
```
mcvsubbu: extra line

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -360,4 +539,50 @@ private String uploadSegment(FormDataMultiPart multiPart, String instanceId, Str
...
+    // The server that was asked to commit did so very slowly (due to network speeds). Meanwhile the FSM in
+    // SegmentCompletionManager timed out, and allowed another server to commit, which did so very quickly (somehow
+    // the network speeds changed). The second server made it through the FSM and reached this point.
+    // The synchronization below takes care that exactly one file gets moved in place.
+    // There are still corner conditions that are not handled correctly. For example,
+    // 1. What if the offset of the faster server was different?
+    // 2. We know that only the faster server will get to complete the COMMIT call successfully. But it is possible
+    //    that the race to this statement is won by the slower server, and so the real segment that is in there is that
+    //    of the slower server.
+    // In order to overcome controller restarts after the segment is renamed, but before it is committed, we DO need to
```
mcvsubbu: s/renamed/moved to pinotFS/

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/LLCSegmentCompletionHandlers.java
```diff
@@ -360,4 +539,50 @@ private String uploadSegment(FormDataMultiPart multiPart, String instanceId, Str
...
+    // check for existing segment file and remove it. So, the block cannot be removed altogether.
+    // For now, we live with these corner cases. Once we have split-commit enabled and working, this code will no longer
+    // be used.
+    synchronized (SegmentCompletionManager.getInstance()) {
```
mcvsubbu: Can you move the synchronized block around the caller of this method, along with all of the comments above? It wil make better sense there than here. thanks.

mcvsubbu: Better yet, let us just get rid of this method and move the whole code to the caller. That way, the comment and synchronized block reside where they can be most associated with

PR #3885, opened 2019-02-26
Actively check cluster changes if there is no callback for a long time

Jackie-Jiang:
We encountered a site issue recently, and is suspecting that Helix callback for cluster changes is not working, probably because ZK re-connection.
This PR enabled the proactive cluster change check if there is no callback for 1 hour.

Changes include:
1. Rewrite ClusterChangeMediator to proactively perform cluster change check
2. Disable the Helix batch-mode and perform deduplication in ClusterChangeMediator
3. Disable the Helix pre-fetch to reduce the ZK accesses
4. Add interface ClusterChangeHandler for general cluster change handle
5. Add ClusterChangeHandler implementation: ExternalViewChangeHandler, InstanceConfigChangeHandler, LiveInstanceChangeHandler
6. Add metrics to track CLUSTER_CHANGE_QUEUE_TIME and PROACTIVE_CLUSTER_CHANGE_CHECK count

Testing done:
1. Run ClusterChangeMediator without starting the cluster change handling thread, the backup inline change processing is working properly
2. Set proactive change check interval to 1 second, both callback and proactive change check work properly

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
@@ -18,21 +18,21 @@
  */
 package org.apache.pinot.broker.broker.helix;
 
-import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
-import org.apache.commons.lang3.tuple.ImmutablePair;
-import org.apache.commons.lang3.tuple.Pair;
+import org.apache.helix.HelixConstants.ChangeType;
 import org.apache.helix.NotificationContext;
+import org.apache.helix.api.listeners.BatchMode;
 import org.apache.helix.api.listeners.ExternalViewChangeListener;
 import org.apache.helix.api.listeners.InstanceConfigChangeListener;
 import org.apache.helix.api.listeners.LiveInstanceChangeListener;
+import org.apache.helix.api.listeners.PreFetch;
 import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.LiveInstance;
-import org.apache.pinot.broker.queryquota.TableQueryQuotaManager;
-import org.apache.pinot.broker.routing.HelixExternalViewBasedRouting;
+import org.apache.pinot.common.metrics.BrokerMeter;
 import org.apache.pinot.common.metrics.BrokerMetrics;
 import org.apache.pinot.common.metrics.BrokerTimer;
 import org.slf4j.Logger;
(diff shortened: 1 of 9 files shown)


sunithabeeram:
@Jackie-Jiang , is it possible to decouple the changes in this PR? Especially, 
* Disable the Helix batch-mode and perform deduplication in ClusterChangeMediator
* Disable the Helix pre-fetch to reduce the ZK accesses

Jackie-Jiang:
@sunithabeeram Helix batch-mode is not enabled before, this PR only makes it explicit, and adds comments for that. All the improvement will come from disabling Helix pre-fetch.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,136 @@
...
+              ClusterChangeHandler changeHandler = entry.getValue();
+              long currentTime = System.currentTimeMillis();
+              Long lastChangeTime = _lastChangeTimeMap.remove(changeType);
+              if (lastChangeTime != null) {
+                brokerMetrics.addTimedValue(BrokerTimer.CLUSTER_CHANGE_QUEUE_TIME, currentTime - lastChangeTime,
+                    TimeUnit.MILLISECONDS);
+                processClusterChange(changeType, changeHandler);
+              } else {
+                long lastProcessTime = _lastProcessTimeMap.get(changeType);
+                if (currentTime - lastProcessTime > PROACTIVE_CHANGE_CHECK_INTERVAL_MS) {
```
mcvsubbu: Am I right in reading this as the first update will not be processed until 1 hour after start? Maybe we should set the map to time 0 in line 70?

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,136 @@
...
-                _tableQueryQuotaManager.processQueryQuotaChange();
-              } catch (Exception e) {
-                LOGGER.warn("Caught exception while updating external view", e);
+            for (Map.Entry<ChangeType, ClusterChangeHandler> entry : _changeHandlerMap.entrySet()) {
+              ChangeType changeType = entry.getKey();
+              ClusterChangeHandler changeHandler = entry.getValue();
+              long currentTime = System.currentTimeMillis();
+              Long lastChangeTime = _lastChangeTimeMap.remove(changeType);
+              if (lastChangeTime != null) {
+                brokerMetrics.addTimedValue(BrokerTimer.CLUSTER_CHANGE_QUEUE_TIME, currentTime - lastChangeTime,
```
mcvsubbu: Do you want to bump different metrics for each change type? That may be useful. We see more EV changes than any other, right?

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,136 @@
...
 
-            if (instanceConfigUpdated) {
-              try {
-                _helixExternalViewBasedRouting.processInstanceConfigChange();
-              } catch (Exception e) {
-                LOGGER.warn("Caught exception while processing instance config", e);
-              }
-            }
+            // Sleep 1 second after each round of checks
+            Thread.sleep(1000L);
```
mcvsubbu: please make this a static final int 1000L.

So, we have changed a conditional wait into busy wait.  I don't see the benefits of this

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,136 @@
...
-      } catch (InterruptedException e) {
-        LOGGER.warn("Was interrupted while trying to add external view change to queue", e);
-      }
-    } else {
-      LOGGER.warn(
-          "Deferred cluster updater thread is null or stopped, not deferring external view routing table rebuild");
-      _helixExternalViewBasedRouting.processExternalViewChange();
-      _tableQueryQuotaManager.processQueryQuotaChange();
-    }
+    Preconditions.checkState(externalViewList.isEmpty(), "Helix pre-fetch should be disabled");
```
mcvsubbu: If I understand right, the pre-fetch is an annotation on this method. Why not add the annotation and set it to be false? Another way of putting this question: What is an admin supposed to do if they see this pre-condition fail?

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,136 @@
...
+    Preconditions.checkState(_clusterChangeHandlingThread.isAlive(), "Cluster change handling thread is not alive");
```
mcvsubbu: This can happen during a race condition where the thread takes longer to start up? maybe you should add a condition to wait until thread is alive before adding these listeners

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,136 @@
...
+          } catch (Exception e) {
+            LOGGER.error("Caught exception while handling changes", e);
           }
         }
-
-        LOGGER.warn("Stopping deferred cluster state update thread");
-        _deferredClusterUpdater = null;
       }
     };
+    _clusterChangeHandlingThread.start();
```
mcvsubbu: add  code here to wait  until thread is alive. Otherwise it is possible that we hit one of the precondition checks

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: This only acts as a fall back mechanism. We should get an INIT callback when we start the cluster.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: This metric only tracks the queue time, not the actual processing time. The processing time is tracked separately. IMO, for queue time, one metric for all types should be good enough.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: Thread is always alive after calling start() unless it's died, in which case we want the precondition checks to fail

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: The pre-fetch is an annotation on the class (line 52). The purpose of this preconditions is to prevent the extra ZK accesses if the Helix behavior changes in the future and ignores the annotation. In such case, our test should be able to catch the preconditions check failure.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: > A thread is alive if it has been started and has not yet died.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: Added static final CHANGE_CHECK_INTERVAL_MS for this, and also add comments for the constants.

The reason for the sleep instead of condition wait is because using sleep we can achieve deduplication of multiple changes of the same type. All the callbacks inside this 1 second will be processed only once.

pinot-common/src/main/java/org/apache/pinot/common/metrics/BrokerMeter.java
```diff
@@ -127,5 +129,4 @@ public String getUnit() {
   @Override
   public boolean isGlobal() {
     return global;
-  }
-}
+  }}
```
jackjlli: Split these two brackets.

pinot-common/src/main/java/org/apache/pinot/common/metrics/BrokerTimer.java
```diff
@@ -48,5 +50,4 @@ public String getTimerName() {
   @Override
   public boolean isGlobal() {
     return global;
-  }
-}
+  }}
```
jackjlli: Same here.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,139 @@
...
-        LOGGER.warn("Was interrupted while trying to add external view change to queue", e);
-      }
-    } else {
-      LOGGER.warn(
-          "Deferred cluster updater thread is null or stopped, not deferring instance config change notification");
-      _helixExternalViewBasedRouting.processInstanceConfigChange();
-    }
+  public void onInstanceConfigChange(List<InstanceConfig> instanceConfigs, NotificationContext changeContext) {
+    Preconditions.checkState(instanceConfigs.isEmpty(), "Helix pre-fetch should be disabled");
+    Preconditions.checkState(_clusterChangeHandlingThread.isAlive(), "Cluster change handling thread is not alive");
```
sunithabeeram: Its not clear who handles this runtime exception and what the recovery path is. The earlier code processed the change inline.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,139 @@
...
+              if (lastChangeTime != null) {
+                brokerMetrics.addTimedValue(BrokerTimer.CLUSTER_CHANGE_QUEUE_TIME, currentTime - lastChangeTime,
+                    TimeUnit.MILLISECONDS);
+                processClusterChange(changeType, changeHandler);
+              } else {
+                long lastProcessTime = _lastProcessTimeMap.get(changeType);
+                if (currentTime - lastProcessTime > PROACTIVE_CHANGE_CHECK_INTERVAL_MS) {
+                  LOGGER.info("Proactive check {} change", changeType);
+                  brokerMetrics.addMeteredGlobalValue(BrokerMeter.PROACTIVE_CLUSTER_CHANGE_CHECK, 1L);
+                  processClusterChange(changeType, changeHandler);
```
sunithabeeram: If we get into the proactive check mode once, does it mean that the proactive check is the only one that drives further updates? In other words, does anything guarantee that the updates will be event based after a proactive check?

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
mcvsubbu: If I recall correctly, the merge of the callbacks was added later as an optimization. Not sure how much things have moved around, but hopefully you can check some history.

pinot-common/src/main/java/org/apache/pinot/common/metrics/BrokerMeter.java
Jackie-Jiang: Good catch! Stupid auto-format lol...

Jackie-Jiang:
@kishoreg By doing deduplication on our side, we get more control on the interval between change checks (no interval for Helix batch-mode, so we might end up doing more fetches), and we can track the cluster change queue time (also, the logic is much simpler than the Helix batch-mode).
In order to add proactive check for all type of changes (not only external view), the change handler must implement the same interface, thus making the change big.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: Changed the behavior to be: if thread is dead, log an error but process inline.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: No, event based will still work. Proactive check only kicks in if there is no event based change for 1 hour.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -18,142 +18,170 @@
...
-              }
-            }
-
-            if (externalViewUpdated) {
-              try {
-                _helixExternalViewBasedRouting.processExternalViewChange();
-                _tableQueryQuotaManager.processQueryQuotaChange();
-              } catch (Exception e) {
-                LOGGER.warn("Caught exception while updating external view", e);
+            for (Map.Entry<ChangeType, ClusterChangeHandler> entry : _changeHandlerMap.entrySet()) {
```
mcvsubbu: You should use while (!shuttingDown) as the loop condition instead of while (true). Pass the shutting down flag down from BrokerStarter

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
sunithabeeram: Nevermind, looks like Helix Spectator behavior should guarantee that watch is set after a read (didn't double check the code though). I was wondering what guarantees that the watch will be reset.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
sunithabeeram: Thanks. If this had gone uncaught, then even with the proactive mode enabled we wouldn't have been able to process the changes, right? Could you please note the testing that was done to validate the changes? If any reliable unit tests can be added to test the new code, that would be really good!

Jackie-Jiang: That's true. And that's why the error logs and metrics are added. When we encounter issues next time, we can figure out the root cause by analyzing the logs and metrics.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: Added stop() to stop the thread, and is called in HelixBrokerStarter. Currently HelixBrokerStarter and BrokerServerBuilder seems not interact correctly. Will address that in a separate PR because that's out of the scope of this one.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
sunithabeeram: The intention was definitely not captured in the description. Regardless, given how critical processing EV changes is, we can always have failsafe mechanism with metrics and alerts helping us followup on when the issue happens.

Regardless, please include testing done (even if it is manual) so we can review the coverage the change has received.

Jackie-Jiang: I am a little bit confused... This 2 lines have been changed to the old behavior if you check the updated code.

mcvsubbu:
@Jackie-Jiang  can you please run a manual test to ensure that the cluster change processor runs after an hour without any problems? thanks

Jackie-Jiang:
@mcvsubbu @sunithabeeram Updated testing done:
1. Run ClusterChangeMediator without starting the cluster change handling thread, the backup inline change processing is working properly
2. Set proactive change check interval to 1 second, both callback and proactive change check work properly

kishoreg:
> we can track the cluster change queue time (also, the logic is much simpler than the Helix batch-mode).
> In order to add proactive check for all type of changes (not only external view), the change handler must implement the same interface, thus making the change big.

Which part of the new code handles the deduping part. Is this already merged?

Jackie-Jiang:
@kishoreg The PR has been merged. For the deduplication, we keep only one last change time for each change type, and reset it whenever we start processing the change. Next call back for that change type can set it again, so it will be picked up in next round. Multiple call backs in the same round will only be processed once.

PR #3886, opened 2019-02-26
Move images files to img directory for docs

snleee:


snleee:
@mcvsubbu I checked it locally and it shows the images correctly.

PR #3894, opened 2019-02-28
Replace deprecated ZkClient APIs

narendly:
It was discovered that for some reason the compiler is not able to locate deleteRecursively() in zkClient. For TableQueryQuotaManagerTest and TimeBoundaryServiceTest, their zkClient initializations refer to the same ZkClient import, yet the former doesn't recognize it as a valid API. This probably has to do with the ambiguity of the imports and multiple ZkClient classes in different paths, and it is just better to replace them with non-deprecated APIs.


narendly:
My analysis in the log above was incorrect. There's no change to the commit itself, but I've identified the exact cause for this. See the following:

The underlying interface for ZkClient in HelixManager has been changed to HelixZkClient. Using ZkClient directly is no longer recommended by Helix because it might cause ZkConnection leakage. However, the way Pinot uses ZkClient here is for testing purposes only, so we need to just replace the deprecated APIs with updated ones.

PR #3898, opened 2019-03-01
In ClusterChangeMediator, remove sleep and make it notify based

Jackie-Jiang:
When we get the Helix cluster change callbacks, we want to process them ASAP
Replace sleep() with wait() and notify() so that the new changes are processed immediately

Testing done:
- Check that the queue time matches with the processing time
- Run tests without starting the cluster change handling thread
- Reduce proactive change check interval to 1 second and check it interacts correctly with Helix callbacks


Jackie-Jiang:
@kishoreg Manually tested. Also, if the thread is not notified, all the integration tests will fail due to the 6 minutes wait()

PR #3899, opened 2019-03-01
[Parsing][PQL] Allow more than one dot in the identifier

agrawaldevesh:
Summary: Allow column names to contain more than one dot in them.

Pinot allows queries of the form: "select a.b from foo limit 1". But it
forbids "select a.b.c from foo limit 1".

Context: At Uber, we flatten out our deeply nested kafka schemas into
columns containing multiple dots. So we have columns like this:
msg.data, msg.data.head, msg.data.tail, msg.app.author.name and so on.

Without this diff, we could not select these columns. But now with this
diff we can. (earlier we could have selected msg.data only but not the
other multiple-dot column names)

Test Plan: Ran this on Uber's Pinot clusters with such production kafka
topics and verified.

Also tested locally by modifying the baseBall stats CSV file to contain
"dots" in the column names header. And correspondingly edited the
baseBall schema json to contain "dot" column names. And ran queries
locally.

### DIFF
--- a/pinot-common/src/main/antlr4/org/apache/pinot/pql/parsers/PQL2.g4
+++ b/pinot-common/src/main/antlr4/org/apache/pinot/pql/parsers/PQL2.g4
@@ -40,8 +40,7 @@ outputColumnProjection:
   expression (AS (IDENTIFIER | STRING_LITERAL))?    # OutputColumn;
 
 expression:
-  IDENTIFIER                                    # Identifier
-  | IDENTIFIER '.' IDENTIFIER                   # Identifier
+  IDENTIFIER ('.' IDENTIFIER)*                  # Identifier
   | literal                                     # Constant
   | '(' expression ')'                          # ExpressionParenthesisGroup
   | function '(' expressions? ')'               # FunctionCall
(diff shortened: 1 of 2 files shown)


agrawaldevesh:
Without this diff, 
"select `a.b.playerID` from baseballStats limit 10" clearly works
but "select `a.b.playerID`, count(*) from baseballStats group by `a.b.playerID`" returns no results. 

OTOH, with this diff: both of these above queries work and the group by returns non empty results. It also obviously works if I remove the backticks. 

I understand that this makes us further deviate from SQL. 

Having said that, going to SQL isn't going to be easy and would anyway require more changes. I know that some of you have started thinking of Calcite integration etc, which would wholesale replace the current the compilation subsystem. 

So I think this change is worth putting in for the short term until we decide to take bigger work towards SQL. (Which is totally something Uber wants !).

kishoreg:
@agrawaldevesh what do you want to do with this PR?

jitendratheta:
@agrawaldevesh @kishoreg is there any plans in roadmap for extending query patterns with Calcite or something else?

pckushan:
@agrawaldevesh Can we use dot notation for the columnName of the table schema config we provide ?
ex: 

> "dimensionFieldSpecs":[
>       {
>          "name":"eventDetails.createdTime",
>          "dataType":"LONG",
> 
>       }
> ]

PR #3900, opened 2019-03-01
Refactor in preparation for the ORCRecordReader

jenniferdai:
Refactoring record reader to allow for ORC configuration during segment creation.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderFactory.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderFactory.java
@@ -35,6 +35,7 @@ public class RecordReaderFactory {
 
     Schema schema = segmentGeneratorConfig.getSchema();
     FileFormat fileFormat = segmentGeneratorConfig.getFormat();
+    String recordReaderPath = segmentGeneratorConfig.getRecordReaderPath();
     switch (fileFormat) {
       case AVRO:
       case GZIPPED_AVRO:
@@ -49,6 +50,14 @@ public class RecordReaderFactory {
       // NOTE: PinotSegmentRecordReader does not support time conversion (field spec must match)
       case PINOT:
         return new PinotSegmentRecordReader(dataFile, schema, segmentGeneratorConfig.getColumnSortOrder());
+      case ORC:
+        // The ORC reader currently uses hive, we don't want to bring this dependency into pinot-core
+        if (recordReaderPath == null) {
+          throw new RuntimeException("Record reader path must be set for ORC");
+        }
+        RecordReader recordReader = (RecordReader) Class.forName(recordReaderPath).newInstance();
+        recordReader.init(segmentGeneratorConfig);
+        return recordReader;
       default:
         throw new UnsupportedOperationException("Unsupported input file format: " + fileFormat);
     }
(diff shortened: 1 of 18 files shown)


sunithabeeram:
Couple of suggestions:
- We can have init() present at all times - I don't see an issue with an empty implementation for most and overridable by the readers that need it.
- We can make the reflection based support to be more general: If the users specify a record-reader class in the segment generation config, you can instantiate it and return. If only the input format is specified, we will instantiate the default readers for specific formats. Do capture the order in which these will be handled and what happens when both are specified (we probably just need one of them to be specified to avoid confusion).
- In the recordreader factory, allocate the record reader and before returning call init() on it; ie, move the return to the end of the getReader method.

jenniferdai:
* We can have init() present at all times - I don't see an issue with an empty implementation for most and overridable by the readers that need it.

> I like this idea, but I am concerned because some of private "init" methods there already, and a lot of code (perhaps other open source users too?) directly instantiates these readers through the constructor rather than the init method. I also don't want people to be confused when they add a new record reader. 

* We can make the reflection based support to be more general: If the users specify a record-reader class in the segment generation config, you can instantiate it and return. If only the input format is specified, we will instantiate the default readers for specific formats. Do capture the order in which these will be handled and what happens when both are specified (we probably just need one of them to be specified to avoid confusion).

> I made the reflection based support more general, I like that. However, for the order in which they are handled, ideally, it would be nice to throw an exception so users know they can't set both, however, in segment generator config, we default have avro format currently, and I don't want to break our users' behavior

* In the recordreader factory, allocate the record reader and before returning call init() on it; ie, move the return to the end of the getReader method.
> Since I wanted the record reader path because of the reason above to override, I moved it in front instead. Let me know if you have thoughts on this.

kishoreg:
Does it need the entire segmentGenerationConfig? How can we make these record readers work for both realtime and offline. 

Also, we should probably move these record reader implementations out of pinot-core

jenniferdai:
Does it need the entire segmentGenerationConfig? 
> Different record readers need different params, so thought segment generation config would be the best/most general. eg: csv record reader needs an extra reader config, json record reader needs a column sort order, etc)

How can we make these record readers work for both realtime and offline.
> Not sure, will look into this.

Also, we should probably move these record reader implementations out of pinot-core 
> Sure, I can put them in their own module although my orc reader brings in hive, so I still would want the reflection instantiation option.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mapper/SegmentCreationMapper.java
```diff
@@ -235,6 +238,12 @@ protected void map(LongWritable key, Text value, Context context)
   }
 
   protected FileFormat getFileFormat(String fileName) {
+    // ORC files do not necessarily have the .orc file extension nor is there a reliable way to decide
```
sunithabeeram: We don't need this anymore, right?

jenniferdai: changed the file formatting to bring in a general option

pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderFactory.java
```diff
@@ -35,6 +40,22 @@ public static RecordReader getRecordReader(SegmentGeneratorConfig segmentGenerat
...
     FileFormat fileFormat = segmentGeneratorConfig.getFormat();
+    String recordReaderPath = segmentGeneratorConfig.getRecordReaderPath();
+
+    // Allow for instantiation general record readers from a record reader path passed into segment generator config
+    // If this is set, this will override the file format
+    if (recordReaderPath != null) {
+      if (fileFormat != null) {
+        // We currently have default file format set to AVRO inside segment generator config,
+        // do not want to break this behavior for clients.
+        LOGGER.warn("Using recordReaderPath {} to read segment, ignoring fileformat {}",
```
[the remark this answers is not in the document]
sunithabeeram: As discussed offline, you can make this change in a followup PR.

PR #3902, opened 2019-03-01
[TE] migration - bucket period conversion 

jihaozh:
Bucket period converts to 1 day instead of  24 hours to cover daylight saving time.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionMigrationResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionMigrationResource.java
@@ -271,6 +271,9 @@ public class DetectionMigrationResource {
   }
 
   private String getBucketPeriod(AnomalyFunctionDTO functionDTO) {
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionMigrationResource.java
```diff
@@ -271,6 +271,9 @@ private String getTimezone(AnomalyFunctionDTO functionDTO) {
   }
 
   private String getBucketPeriod(AnomalyFunctionDTO functionDTO) {
+    if (functionDTO.getBucketSize() == 1 && functionDTO.getBucketUnit().equals(TimeUnit.DAYS)){
+      return "P1D";
```
cyy0714: Can be more general to Period.days(functionDTO.getBucketSize()).toString()

jihaozh: updated

PR #3910, opened 2019-03-05
Update the display for sidebar navigation to include section names

sunithabeeram:
Update the sidebar display as shown in the screenshot. Main change required was to update index.rst to have the :caption: keyword. ~~However, with the referenced sphinx theme, the display of highlevel sections (for ex, "Introduction" in the sidebar) was in a light color (not quite readable). Following the instructions on https://sphinx-rtd-theme.readthedocs.io/en/latest/installing.html, I downloaded the theme and included it in the docs folder and that has made it work better.~~

screenshot from 2019-03-05 15-03-00


sunithabeeram:
@mcvsubbu it looks like the local build isn't displaying the Edit on Github link. Will create an issue if it still is an issue on the main rtd link.

snleee:
As we discussed offline, we should do the following:
1. remove license check for `_themes`
2. remove license header for external code (css, html, js..etc) and keep them the same as the original code.
3. update `LICENSE` file

sunithabeeram:
@snleee updated.

sunithabeeram:
@snleee I've removed the includes for now based on our discussion. Locally, the sphinx-build version I have is v1.6.3. Its perhaps an older one and the one thats on rtd is newer (I also saw other differences).  Will push this code and build it out to ensure things work correctly.

sunithabeeram:
readthedocs build uses sphinx v1.7.9 and that renders the captions correctly.

PR #3912, opened 2019-03-06
Fix SegmentGeneratorConfig where the time format in schema is not picked up

Jackie-Jiang:
Fix the issue where the time related specs in schema are not picked up by SegmentGeneratorConfig
Also fix the bug in SegmentGeneratorConfig.loadConfigFiles() where incoming time unit is picked (should use outgoing time unit)

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
@@ -46,6 +46,7 @@ import org.apache.pinot.common.data.FieldSpec.FieldType;
 import org.apache.pinot.common.data.Schema;
 import org.apache.pinot.common.data.StarTreeIndexSpec;
 import org.apache.pinot.common.data.TimeFieldSpec;
+import org.apache.pinot.common.data.TimeGranularitySpec;
 import org.apache.pinot.common.utils.JsonUtils;
 import org.apache.pinot.core.data.readers.CSVRecordReaderConfig;
 import org.apache.pinot.core.data.readers.FileFormat;
@@ -160,12 +161,10 @@ public class SegmentGeneratorConfig {
   /**
    * This constructor is used during offline data generation. Note that it has an option that will generate inverted
    * index.
-   * @param tableConfig
-   * @param schema
    */
   public SegmentGeneratorConfig(@Nullable TableConfig tableConfig, Schema schema) {
     Preconditions.checkNotNull(schema);
-    _schema = schema;
+    setSchema(schema);
 
     if (tableConfig == null) {
       return;
@@ -213,7 +212,7 @@ public class SegmentGeneratorConfig {
   }
 
   public SegmentGeneratorConfig(Schema schema) {
-    _schema = schema;
+    setSchema(schema);
   }
 
   public Map<String, String> getCustomProperties() {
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
```diff
@@ -469,6 +459,28 @@ public Schema getSchema() {
   public void setSchema(Schema schema) {
     Preconditions.checkNotNull(schema);
     _schema = schema;
+
+    // Set time related fields
```
mcvsubbu: RealtimeSegmentConverter uses SegmentGeneratorConfig and sets time unit and time column name independently after constructing it with schema. Perhaps you can remove those lines form there? And move the TODOs here?

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfigTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.pinot.common.data.FieldSpec;
+import org.apache.pinot.common.data.Schema;
+import org.apache.pinot.common.data.TimeGranularitySpec;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNull;
+
+
+public class SegmentGeneratorConfigTest {
```
mcvsubbu: thanks for adding this test. We need more tests here, so if you can add a TODO we can grow it over time. The SegmentGeneratorConfig class has too much logic in it (rather than just holding the config)

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
Jackie-Jiang: Done

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfigTest.java
Jackie-Jiang: Done. Also created a ticket to clean up SegmentGeneratorConfig to use builder.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -335,9 +336,10 @@ void writeMetadata()
     ColumnIndexCreationInfo timeColumnIndexCreationInfo = indexCreationInfoMap.get(timeColumn);
     if (timeColumnIndexCreationInfo != null) {
       // Use start/end time in config if defined
```
jackjlli: Please update the comment here if logic changes.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
```diff
@@ -382,11 +382,7 @@ public void setSegmentNamePostfix(String postfix) {
   }
 
   public String getTimeColumnName() {
-    if (_segmentTimeColumnName != null) {
-      return _segmentTimeColumnName;
-    }
-    // TODO: if segmentTimeColumnName is null, getQualifyingFields DATETIME. If multiple found, throw exception "must specify primary timeColumnName"
```
mcvsubbu: @npawar  can you confirm that it is ok to remove this and keep backward comaptibility?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
Jackie-Jiang: The logic is not changed, but a bug fix. Updated the comments to make that more clear

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
Jackie-Jiang: Here getting TIME field is an no-op. Moved the TODO for DATETIME field support into `setSchema()`

npawar: this looks fine to me.

mcvsubbu:
@Jackie-Jiang if you can please have travis clear it

otherwise lgtm

PR #3913, opened 2019-03-06
Adding the support for sampling logs

snleee:
For high throughput use cases, logging becomes the bottleneck
on both server and broker. "Category.callAppenders()" is using
a synchronized block that all the worker threads can be stalled
due to logging. Long term solution should be migration towards
log4j2. As a short term fix, this pr adds the configuration for
sampling logs.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.annotation.Nullable;
@@ -65,6 +66,7 @@ import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.TRACE
 public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   private static final Logger LOGGER = LoggerFactory.getLogger(BaseBrokerRequestHandler.class);
   private static final Pql2Compiler REQUEST_COMPILER = new Pql2Compiler();
+  private static final Random RANDOM = new Random();
 
   protected final Configuration _config;
   protected final RoutingTable _routingTable;
@@ -81,6 +83,7 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   protected final long _brokerTimeoutMs;
   protected final int _queryResponseLimit;
   protected final int _queryLogLength;
+  protected final float _queryLogSamplingRate;
 
   public BaseBrokerRequestHandler(Configuration config, RoutingTable routingTable,
       TimeBoundaryService timeBoundaryService, AccessControlFactory accessControlFactory,
@@ -96,9 +99,12 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     _brokerTimeoutMs = config.getLong(CONFIG_OF_BROKER_TIMEOUT_MS, DEFAULT_BROKER_TIMEOUT_MS);
     _queryResponseLimit = config.getInt(CONFIG_OF_BROKER_QUERY_RESPONSE_LIMIT, DEFAULT_BROKER_QUERY_RESPONSE_LIMIT);
     _queryLogLength = config.getInt(CONFIG_OF_BROKER_QUERY_LOG_LENGTH, DEFAULT_BROKER_QUERY_LOG_LENGTH);
+    _queryLogSamplingRate =
+        config.getFloat(CONFIG_OF_BROKER_QUERY_LOG_SAMPLING_RATE, DEFAULT_BROKER_QUERY_LOG_SAMPLING_RATE);
 
-    LOGGER.info("Broker Id: {}, timeout: {}ms, query response limit: {}, query log length: {}", _brokerId,
-        _brokerTimeoutMs, _queryResponseLimit, _queryLogLength);
+    LOGGER.info(
+        "Broker Id: {}, timeout: {}ms, query response limit: {}, query log length: {}, query log sampling rate: {}",
+        _brokerId, _brokerTimeoutMs, _queryResponseLimit, _queryLogLength, _queryLogSamplingRate);
   }
 
   private String getDefaultBrokerId() {
@@ -292,17 +298,21 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
 
     LOGGER.debug("Broker Response: {}", brokerResponse);
 
-    // Table name might have been changed (with suffix _OFFLINE/_REALTIME appended)
-    LOGGER.info(
-        "RequestId:{}, table:{}, timeMs:{}, docs:{}/{}, entries:{}/{}, segments(queried/processed/matched):{}/{}/{} "
-            + "servers:{}/{}, groupLimitReached:{}, exceptions:{}, serverStats:{}, query:{}", requestId,
-        brokerRequest.getQuerySource().getTableName(), totalTimeMs, brokerResponse.getNumDocsScanned(),
-        brokerResponse.getTotalDocs(), brokerResponse.getNumEntriesScannedInFilter(),
-        brokerResponse.getNumEntriesScannedPostFilter(), brokerResponse.getNumSegmentsQueried(),
-        brokerResponse.getNumSegmentsProcessed(), brokerResponse.getNumSegmentsMatched(),
-        brokerResponse.getNumServersResponded(), brokerResponse.getNumServersQueried(),
-        brokerResponse.isNumGroupsLimitReached(), brokerResponse.getExceptionsSize(), serverStats.getServerStats(),
-        StringUtils.substring(query, 0, _queryLogLength));
+    // Sampling the query log based on the sampling rate configuration
+    if(_queryLogSamplingRate > RANDOM.nextFloat()) {
+      // Table name might have been changed (with suffix _OFFLINE/_REALTIME appended)
+      LOGGER.info(
+          "RequestId:{}, table:{}, timeMs:{}, docs:{}/{}, entries:{}/{}, segments(queried/processed/matched):{}/{}/{} "
+              + "servers:{}/{}, groupLimitReached:{}, exceptions:{}, serverStats:{}, query:{}", requestId,
+          brokerRequest.getQuerySource().getTableName(), totalTimeMs, brokerResponse.getNumDocsScanned(),
+          brokerResponse.getTotalDocs(), brokerResponse.getNumEntriesScannedInFilter(),
+          brokerResponse.getNumEntriesScannedPostFilter(), brokerResponse.getNumSegmentsQueried(),
+          brokerResponse.getNumSegmentsProcessed(), brokerResponse.getNumSegmentsMatched(),
+          brokerResponse.getNumServersResponded(), brokerResponse.getNumServersQueried(),
+          brokerResponse.isNumGroupsLimitReached(), brokerResponse.getExceptionsSize(), serverStats.getServerStats(),
+          StringUtils.substring(query, 0, _queryLogLength));
+    }
+
 
     return brokerResponse;
   }
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -20,6 +20,7 @@ package org.apache.pinot.broker.requesthandler;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.base.Splitter;
+import com.google.common.util.concurrent.RateLimiter;
 import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -66,7 +67,6 @@ import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.TRACE
 public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   private static final Logger LOGGER = LoggerFactory.getLogger(BaseBrokerRequestHandler.class);
   private static final Pql2Compiler REQUEST_COMPILER = new Pql2Compiler();
-  private static final Random RANDOM = new Random();
 
   protected final Configuration _config;
   protected final RoutingTable _routingTable;
@@ -83,7 +83,7 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   protected final long _brokerTimeoutMs;
   protected final int _queryResponseLimit;
   protected final int _queryLogLength;
-  protected final float _queryLogSamplingRate;
+  protected final RateLimiter _queryLogRateLimiter;
 
   public BaseBrokerRequestHandler(Configuration config, RoutingTable routingTable,
       TimeBoundaryService timeBoundaryService, AccessControlFactory accessControlFactory,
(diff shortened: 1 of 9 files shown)


kishoreg:
can we set a maxRate instead of setting samplingRate? we can use the RateLimiter from Gauva to implement this for now. This will make the rate adaptive and at low qps, we will still get all the logs unlike sampling where it will always filter.

We can then move to BurstFilter (https://logging.apache.org/log4j/2.0/manual/filters.html#BurstFilter) when we move to log4j2

snleee:
@kishoreg We currently log within `QueryScheduler` in server and `BaseBrokerHandler` in broker. Unless we log in the separate thread, we should not use rate limiter since it will block threads in the critical path. Even if we write logs in a separate thread, task queue will grow indefinitely if incoming logs are queueing much faster than `maxRate`. How do we address this?

snleee:
@kishoreg Oh I think i misunderstood your comment. So, I think that you meant dropping logs once we reach `maxRate`. That makes more sense. I will make the change.

kishoreg:
thats right. tryAcquire is non-blocking (please verify)

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -65,6 +66,7 @@
 public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   private static final Logger LOGGER = LoggerFactory.getLogger(BaseBrokerRequestHandler.class);
   private static final Pql2Compiler REQUEST_COMPILER = new Pql2Compiler();
+  private static final Random RANDOM = new Random();
```
mcvsubbu: Good to seed it with some value.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -292,17 +298,21 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
...
-            + "servers:{}/{}, groupLimitReached:{}, exceptions:{}, serverStats:{}, query:{}", requestId,
-        brokerRequest.getQuerySource().getTableName(), totalTimeMs, brokerResponse.getNumDocsScanned(),
-        brokerResponse.getTotalDocs(), brokerResponse.getNumEntriesScannedInFilter(),
-        brokerResponse.getNumEntriesScannedPostFilter(), brokerResponse.getNumSegmentsQueried(),
-        brokerResponse.getNumSegmentsProcessed(), brokerResponse.getNumSegmentsMatched(),
-        brokerResponse.getNumServersResponded(), brokerResponse.getNumServersQueried(),
-        brokerResponse.isNumGroupsLimitReached(), brokerResponse.getExceptionsSize(), serverStats.getServerStats(),
-        StringUtils.substring(query, 0, _queryLogLength));
+    // Sampling the query log based on the sampling rate configuration
+    if(_queryLogSamplingRate > RANDOM.nextFloat()) {
```
mcvsubbu: maybe have other criteria which will force a log:
* if number of exceptions > 0
* timeMs > 100 (say)
* entries scanned post filter is high
* num groups limit is reached

I suggest adding a method shouldLog()

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
```diff
@@ -48,9 +50,14 @@
  */
 public abstract class QueryScheduler {
   private static final Logger LOGGER = LoggerFactory.getLogger(QueryScheduler.class);
+  private static final Random RANDOM = new Random();
```
mcvsubbu: Seed it with time

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
```diff
@@ -170,13 +181,17 @@ public void stop() {
...
-    LOGGER.info(
-        "Processed requestId={},table={},segments(queried/processed/matched)={}/{}/{},schedulerWaitMs={},totalExecMs={},totalTimeMs={},broker={},numDocsScanned={},scanInFilter={},scanPostFilter={},sched={}",
-        requestId, tableNameWithType, numSegmentsQueried, numSegmentsProcessed, numSegmentsMatched,
-        timerContext.getPhaseDurationMs(ServerQueryPhase.SCHEDULER_WAIT),
-        timerContext.getPhaseDurationMs(ServerQueryPhase.QUERY_PROCESSING),
-        timerContext.getPhaseDurationMs(ServerQueryPhase.TOTAL_QUERY_TIME), queryRequest.getBrokerId(), numDocsScanned,
-        numEntriesScannedInFilter, numEntriesScannedPostFilter, name());
+
+    // Sampling the query log based on the sampling rate configuration
+    if (queryLogSamplingRate > RANDOM.nextFloat()) {
```
mcvsubbu: Same as broker. Maybe we should log under certain conditions even if random criteria is not met.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
snleee: @mcvsubbu Some numbers would be use case specific. For now, I have put some lenient numbers. We can tune this as we needed.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
snleee: changed to `maxRate` approach. This is no longer needed.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -111,6 +111,8 @@ public ServerType getServerType() {
     public static final int DEFAULT_BROKER_QUERY_RESPONSE_LIMIT = Integer.MAX_VALUE;
     public static final String CONFIG_OF_BROKER_QUERY_LOG_LENGTH = "pinot.broker.query.log.length";
     public static final int DEFAULT_BROKER_QUERY_LOG_LENGTH = Integer.MAX_VALUE;
+    public static final String CONFIG_OF_BROKER_QUERY_LOG_MAX_RATE = "pinot.broker.query.log.maxRate";
+    public static final double DEFAULT_BROKER_QUERY_LOG_MAX_RATE = 10_000d;
```
sunithabeeram: Could you indicate the time unit for this? Is it max rate per second, minute?

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
```diff
@@ -48,9 +51,13 @@
  */
 public abstract class QueryScheduler {
   private static final Logger LOGGER = LoggerFactory.getLogger(QueryScheduler.class);
+
   private static final String INVALID_NUM_SCANNED = "-1";
   private static final String INVALID_SEGMENTS_COUNT = "-1";
+  private static final String QUERY_LOG_MAX_RATE_KEY = "query_log_max_rate";
```
sunithabeeram: Indicate timeunit. Should the key be "query.log.max.rate" (to conform with other keys we have)?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -292,21 +298,45 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
...
+          "RequestId:{}, table:{}, timeMs:{}, docs:{}/{}, entries:{}/{}, segments(queried/processed/matched):{}/{}/{} "
+              + "servers:{}/{}, groupLimitReached:{}, exceptions:{}, serverStats:{}, query:{}", requestId,
+          brokerRequest.getQuerySource().getTableName(), totalTimeMs, brokerResponse.getNumDocsScanned(),
+          brokerResponse.getTotalDocs(), brokerResponse.getNumEntriesScannedInFilter(),
+          brokerResponse.getNumEntriesScannedPostFilter(), brokerResponse.getNumSegmentsQueried(),
+          brokerResponse.getNumSegmentsProcessed(), brokerResponse.getNumSegmentsMatched(),
+          brokerResponse.getNumServersResponded(), brokerResponse.getNumServersQueried(),
+          brokerResponse.isNumGroupsLimitReached(), brokerResponse.getExceptionsSize(), serverStats.getServerStats(),
+          StringUtils.substring(query, 0, _queryLogLength));
+    }
```
sunithabeeram: Could you track how many logs were throttled and log the count (if non-zero) the next time we are able to acquire a permit for logging? Else, it might be quite confusing to users if they expect to see some logs and find no trace of them.

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
snleee: I was following the existing convention of schedulerConfig. Referred `ResourceManager` and it has `query_runner_threads` and `query_worker_threads`. I can do `query.log.max.rate`. Full config name will be `pinot.server.scheduler.query.log.max.rate` since `QueryScheduler` will only get configs with prefix of `pinot.server.scheduler.*`. How do you think?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
snleee: fixed

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
snleee: added

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
```diff
@@ -170,13 +181,16 @@ public void stop() {
...
-        numEntriesScannedInFilter, numEntriesScannedPostFilter, name());
+    long schedulerWaitMs = timerContext.getPhaseDurationMs(ServerQueryPhase.SCHEDULER_WAIT);
+
+    if (queryLogRateLimiter.tryAcquire() || forceLog(schedulerWaitMs, numDocsScanned)) {
+      LOGGER.info(
+          "Processed requestId={},table={},segments(queried/processed/matched)={}/{}/{},schedulerWaitMs={},totalExecMs={},totalTimeMs={},broker={},numDocsScanned={},scanInFilter={},scanPostFilter={},sched={}",
+          requestId, tableNameWithType, numSegmentsQueried, numSegmentsProcessed, numSegmentsMatched, schedulerWaitMs,
+          timerContext.getPhaseDurationMs(ServerQueryPhase.QUERY_PROCESSING),
+          timerContext.getPhaseDurationMs(ServerQueryPhase.TOTAL_QUERY_TIME), queryRequest.getBrokerId(),
+          numDocsScanned, numEntriesScannedInFilter, numEntriesScannedPostFilter, name());
```
mcvsubbu: Do you want to log the dropped messages in the server  as well?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -292,21 +304,60 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
...
+          brokerRequest.getQuerySource().getTableName(), totalTimeMs, brokerResponse.getNumDocsScanned(),
+          brokerResponse.getTotalDocs(), brokerResponse.getNumEntriesScannedInFilter(),
+          brokerResponse.getNumEntriesScannedPostFilter(), brokerResponse.getNumSegmentsQueried(),
+          brokerResponse.getNumSegmentsProcessed(), brokerResponse.getNumSegmentsMatched(),
+          brokerResponse.getNumServersResponded(), brokerResponse.getNumServersQueried(),
+          brokerResponse.isNumGroupsLimitReached(), brokerResponse.getExceptionsSize(), serverStats.getServerStats(),
+          StringUtils.substring(query, 0, _queryLogLength));
+
+      // Limit the dropping log message at most once per second.
+      if (_numDroppedLogRateLimiter.tryAcquire()) {
```
sunithabeeram: Not sure we need a another ratelimiter. Here is how I thought about it using the existing query log ratelimiter: if tryAcquire() fails once, it will fail for the rest of the one second duration, so we'll not enter this block and just increment in the else block. Once we enter this block, we check if numDropped > 0, then log, reset and move on. The if check for numDropped > 0 should not be satisfied until it gets rate limited and the duration passes. Let me know if this won't work.

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
sunithabeeram: You can keep this for now; we should have a followup to fix it to be consistent.

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
sunithabeeram: +1

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
```diff
@@ -185,6 +199,25 @@ public void stop() {
...
+  /**
+   * Helper function to decide whether to force the log
+   *
+   * TODO: come up with other criteria for forcing a log and come up with better numbers
+   *
+   */
+  private boolean forceLog(long schedulerWaitMs, long numDocsScanned) {
+    // If scheduler wait time is larger than 100ms, force the log
+    if (schedulerWaitMs > 100L) {
+      return false;
```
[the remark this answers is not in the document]
snleee: oops. I will make a change. Thanks for the catch.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
snleee: This won't log once per sec. Instead, it will log dropped messages whenever `tryAcquire()` succeeds.

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
snleee: added

pinot-core/src/main/java/org/apache/pinot/core/query/scheduler/QueryScheduler.java
snleee: I changed it to `query.log.maxRatePerSecond` to make it consistent with broker. For `query_runner_threads` and `query_worker_threads`, I will address it with a separate pr along with checking other configs.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
sunithabeeram: Ok, the permits are scatter across the second, hence my assumption was incorrect. Thanks for handling this.

snleee:
Perf stress test result: 

Experiment setup: 1 broker, 1 server, 1.5k qps, internal production data

original code (no throttling)
broker side avg 99th percentile latency - 83ms
server side avg 99th percentile latency - 8.3ms
server side avg 999th percentile latency - 85ms

throttling log at maximum rate of 100qps
broker side avg 99th percentile latency - 44ms
server side avg 99th percentile latency - 12.63ms
server side avg 999th percentile latency - 31.8ms

sunithabeeram:
Thanks for debugging this @snleee!

PR #3917, opened 2019-03-06
[TE] detection - add default values for duration filter

jihaozh:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/DurationAnomalyFilterSpec.java
```diff
@@ -21,8 +21,8 @@
 package org.apache.pinot.thirdeye.detection.spec;
 
 public class DurationAnomalyFilterSpec extends AbstractSpec {
-  private String minDuration;
-  private String maxDuration;
+  private String minDuration = "PT0S";
```
akshayrai: Can you include a comment on what these values represent?

jihaozh: sure

PR #3918, opened 2019-03-07
[TE] Include description and last detection time in alert header, rem…

akshayrai:
…ove application


harleyjj:
Nice - I think these variable names make things clearer

PR #3919, opened 2019-03-07
 readthedocs: Fill in dev-guide sections

sunithabeeram:
Add more info to the dev guide. Contribution_guidelines doc is WIP.

NOTE: In dev env setup, we are recommend users to create a fork to limit the number of branches on the apache repo. This is a slight deviation from the current process, but shouldn't impact the workflow much.
The code_modules documentation is intended to orient the contributor at a highlevel to the code structure. Contribution_guidelines will reference this further.

I've attached a sample of the code-modules section.
screenshot from 2019-03-06 17-22-00


sunithabeeram:
@snleee the version shows only for the local build. On rtd, it shows latest.

PR #3921, opened 2019-03-07
[TE] frontend - harleyjj/alert-details - fix timeseries fetch

harleyjj:
Removes granularity and timezone from timeseries fetch in Alert Overview and Preview

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/alert-details/component.js
+++ b/thirdeye/thirdeye-frontend/app/pods/components/alert-details/component.js
@@ -597,12 +597,11 @@ export default Component.extend({
       analysisRange,
       selectedBaseline
     } = this.getProperties('metricUrn', 'analysisRange', 'selectedBaseline');
-    const granularity = '15_MINUTES';
-    const timezone = moment.tz.guess();
 
     set(this, 'errorTimeseries', null);
 
-    const urlCurrent = `/rootcause/metric/timeseries?urn=${metricUrn}&start=${analysisRange[0]}&end=${analysisRange[1]}&offset=current&granularity=${granularity}&timezone=${timezone}`;
+    const timeZone = 'America/Los_Angeles';
+    const urlCurrent = `/rootcause/metric/timeseries?urn=${metricUrn}&start=${analysisRange[0]}&end=${analysisRange[1]}&offset=current&timezone=${timeZone}`;
     fetch(urlCurrent)
       .then(checkStatus)
       .then(res => {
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-frontend/app/pods/components/alert-details/component.js
```diff
@@ -597,12 +597,10 @@ export default Component.extend({
       analysisRange,
       selectedBaseline
     } = this.getProperties('metricUrn', 'analysisRange', 'selectedBaseline');
-    const granularity = '15_MINUTES';
-    const timezone = moment.tz.guess();
 
     set(this, 'errorTimeseries', null);
 
-    const urlCurrent = `/rootcause/metric/timeseries?urn=${metricUrn}&start=${analysisRange[0]}&end=${analysisRange[1]}&offset=current&granularity=${granularity}&timezone=${timezone}`;
+    const urlCurrent = `/rootcause/metric/timeseries?urn=${metricUrn}&start=${analysisRange[0]}&end=${analysisRange[1]}&offset=current`;
```
xiaohui-sun: Should get the granularity and timezone from detect function.
E.g., the underline data may have hourly granularity, but we may want to monitor it in daily granularity.

jihaozh: The time zone should be the frontend's timezone. 

Technically, the granularity of the graph should be the granularity for detection. However, with yaml, we allow the user to monitor in different granularity in different detection rules. Therefore, multiple granularities could be used in the in one detection, we don't know which one to show. I suggest using default time granularity for the metric in preview graph for now.

thirdeye/thirdeye-frontend/app/pods/components/alert-details/component.js
```diff
@@ -597,12 +597,10 @@ export default Component.extend({
       analysisRange,
       selectedBaseline
     } = this.getProperties('metricUrn', 'analysisRange', 'selectedBaseline');
-    const granularity = '15_MINUTES';
-    const timezone = moment.tz.guess();
```
jihaozh: Should we keep the timezone? @harleyjj could you check how RCA makes this call, it should use the same timezone. 

The granularity can be just left empty.

jihaozh: Just checked this. 

The timezone should be pinned to "America/Los_Angeles".
Granularity should be left empty.

Thanks.

harleyjj: I'll go ahead and change that asap

PR #3923, opened 2019-03-07
readthedocs: Fix formatting issues and figures used

sunithabeeram:


docs/README.md
```diff
@@ -1,23 +1,28 @@
-..
```
snleee: Is this added by the license plugin or added manually? If it's from license plugin, I think that we should fix the issue.

sunithabeeram: I don't think so. Its likely that I copied and edited the file manually. Let me check the history.

PR #3932, opened 2019-03-07
Editing pluggable storage docs

jenniferdai:
Moving configuration doc from github link to readthedocs


docs/pluggable_storage.rst
```diff
@@ -35,6 +35,26 @@ New Storage Type implementation
...
+
+In your controller and server configs, please set the FS class you would like to support. pinot.controller.storage.factory.class.${YOUR_URI_SCHEME} to the full path of the FS class you would like to include
+
+You also need to configure pinot.controller.local.temp.dir for the local dir on the controller machine.
+
+For filesystem specific configs, you can pass in the following with either the pinot.controller prefix or the pinot.server prefix.
+
+All the following configs need to be prefixed with storage.factory. AzurePinotFS requires the following configs:
+
+adl.accountId, adl.authEndpoint, adl.clientId, adl.clientSecret segment.fetcher.protocols : adl
```
sunithabeeram: What does the trailing ": adl" imply here? Same with ":  hdfs" below.

jenniferdai: That's what users should set if they want to use adl or hdfs, essentially lets us know which segment fetchers to initialize

PR #3934, opened 2019-03-08
support abs,ceil,exp,floor,ln,log,sqrt math transform function

xueyumusic:
This PR adds abs, ceil, exp, floor, ln, log, sqrt math transform function to support queries like "select SUM(ABS(c1)) from foo" etc..


pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,248 @@
...
+    int length = projectionBlock.getNumDocs();
+
+    if (_firstTransformFunction == null) {
+      Arrays.fill(_result, 0, length, _firstLiteral);
+    } else {
+      switch (_firstTransformFunction.getResultMetadata().getDataType()) {
+        case INT:
+          int[] intValues = _firstTransformFunction.transformToIntValuesSV(projectionBlock);
+          for (int i = 0; i < length; i++) {
+            _result[i] = transformInt(intValues[i]);
```
kishoreg: Not sure if all JVM will inline these methods automatically. better to pass the entire array to avoid the overhead of function call.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.Map;
+import javax.annotation.Nonnull;
+import org.apache.pinot.core.common.DataSource;
+import org.apache.pinot.core.operator.blocks.ProjectionBlock;
+import org.apache.pinot.core.operator.transform.TransformResultMetadata;
+import org.apache.pinot.core.plan.DocIdSetPlanNode;
+import org.apache.pinot.core.util.ArrayCopyUtils;
+
+
+public abstract class SingleParamMathTransformFunction extends BaseTransformFunction {
```
kishoreg: add java doc

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,248 @@
...
+      return FUNCTION_NAME;
+    }
+
+    @Override
+    protected double transformDouble(double value) {
+      return Math.log(value);
+    }
+  }
+
+  public static class Log10TransformFunction extends SingleParamMathTransformFunction {
```
mayankshriv: Instead of Log10 and ln, can we have log(number, base)?

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
xueyumusic: Yes, that makes sense. updated the codes and java doc. Thanks @kishoreg

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
xueyumusic: @mayankshriv , yes `log(number, base)` is more generic and better to have. This pr only related with kinds of single parameter function which are have common base class. Will try to add this in the following. Thank you~

mayankshriv: @xueyumusic In that case, could you please take out the ln and log10 from this PR? Otherwise we will end up with three different log functions.

xueyumusic: Done, @mayankshriv ,updated, thank you.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,343 @@
...
+import javax.annotation.Nonnull;
+import org.apache.pinot.core.common.DataSource;
+import org.apache.pinot.core.operator.blocks.ProjectionBlock;
+import org.apache.pinot.core.operator.transform.TransformResultMetadata;
+import org.apache.pinot.core.plan.DocIdSetPlanNode;
+
+/**
+ * A group of commonly used math transformation which has only one single parameter,
+ * including abs, exp, ceil, floor, ln, log10, sqrt.
+ *
```
Jackie-Jiang: (nit) remove the empty line

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,343 @@
...
+ */
+public abstract class SingleParamMathTransformFunction extends BaseTransformFunction {
+  private double _firstLiteral;
```
Jackie-Jiang: I don't think we need to handle the case for literal. Also, the current implementation cannot handle literal properly since the math operator is not applied to it.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,343 @@
...
+  private TransformFunction _firstTransformFunction;
```
Jackie-Jiang: We only have one argument, so let's just call it _transformFunction.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,343 @@
...
+  private double[] _result;
```
Jackie-Jiang: _results?

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,343 @@
...
+
+  @Override
+  public void init(@Nonnull List<TransformFunction> arguments, @Nonnull Map<String, DataSource> dataSourceMap) {
+    // Check that there are exactly 1 arguments
+    if (arguments.size() != 1) {
+      throw new IllegalArgumentException("Exactly 1 arguments are required for " + getName() + " transform function");
+    }
+
+    TransformFunction firstArgument = arguments.get(0);
+    if (firstArgument instanceof LiteralTransformFunction) {
```
Jackie-Jiang: If we are not going to handle literal, add an argument check here to reject literal.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,343 @@
...
+    if (_result == null) {
+      _result = new double[DocIdSetPlanNode.MAX_DOC_PER_CALL];
+    }
+
+    int length = projectionBlock.getNumDocs();
+
+    if (_firstTransformFunction == null) {
+      Arrays.fill(_result, 0, length, _firstLiteral);
+    } else {
+      switch (_firstTransformFunction.getResultMetadata().getDataType()) {
```
Jackie-Jiang: Since the results are always with type DOUBLE, directly get double[] (**_firstTransformFunction.transformToDoubleValuesSV(projectionBlock)**) as input and only implement the transform for double type. No need to duplicate the type conversion logic.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,343 @@
...
+    return _result;
+  }
+
+  abstract protected void transformInt(double[] result, int[] values, int length);
+
+  abstract protected void transformLong(double[] result, long[] values, int length);
+
+  abstract protected void transformFloat(double[] result, float[] values, int length);
+
+  abstract protected void transformDouble(double[] result, double[] values, int length);
```
Jackie-Jiang: Only this one is required. Recommend changing it to applyMathOperator()

pinot-core/src/test/java/org/apache/pinot/core/operator/transform/function/BaseTransformFunctionTest.java
```diff
@@ -65,6 +65,7 @@
   protected static final String STRING_SV_COLUMN = "stringSV";
   protected static final String INT_MV_COLUMN = "intMV";
   protected static final String TIME_COLUMN = "time";
+  protected static final String DOUBLE2_SV_COLUMN = "double2SV";
```
Jackie-Jiang: Please reuse the DOUBLE_SV_COLUMN, while you can change its value (line 92) if you need value larger than 1 for tests.

pinot-core/src/test/java/org/apache/pinot/core/operator/transform/function/BaseTransformFunctionTest.java
```diff
@@ -87,7 +87,7 @@ public void setUp()
       _intSVValues[i] = RANDOM.nextInt();
       _longSVValues[i] = RANDOM.nextLong();
       _floatSVValues[i] = RANDOM.nextFloat();
-      _doubleSVValues[i] = RANDOM.nextDouble();
+      _doubleSVValues[i] = _intSVValues[i] * RANDOM.nextDouble();
```
Jackie-Jiang: Can you also change floatSV and stringSV so they are consistent?

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,175 @@
...
+    if (_results == null) {
+      _results = new double[DocIdSetPlanNode.MAX_DOC_PER_CALL];
+    }
+
+    int length = projectionBlock.getNumDocs();
+
+    if (_transformFunction == null) {
+      Arrays.fill(_results, 0, length, 0);
+    } else {
+      switch (_transformFunction.getResultMetadata().getDataType()) {
```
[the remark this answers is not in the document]
xueyumusic: Thanks for review, @Jackie-Jiang , It looks this test case with STRING_SV_COLUMN will failed.
```
    expression =
        TransformExpressionTree.compileToExpressionTree(String.format("abs(%s)", STRING_SV_COLUMN));
    transformFunction = TransformFunctionFactory.get(expression, _dataSourceMap);
    Assert.assertTrue(transformFunction instanceof AbsTransformFunction);
    for (int i = 0; i < NUM_ROWS; i++) {
      expectedValues[i] = Math.abs(Double.parseDouble(_stringSVValues[i]));
    }
    testTransformFunction(transformFunction, expectedValues);
```
Do you think this test case is reasonable? Thank you.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
```diff
@@ -0,0 +1,175 @@
...
+
+import javax.annotation.Nonnull;
+import org.apache.pinot.core.common.DataSource;
+import org.apache.pinot.core.operator.blocks.ProjectionBlock;
+import org.apache.pinot.core.operator.transform.TransformResultMetadata;
+import org.apache.pinot.core.plan.DocIdSetPlanNode;
+
+/**
+ * A group of commonly used math transformation which has only one single parameter,
+ * including abs, exp, ceil, floor, sqrt.
```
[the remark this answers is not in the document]
mayankshriv: I had looked at standard SQL and found only one version. I don't see a lot of value in having three version for log function, it does not add enough convenience to users to warrant polluting the function space.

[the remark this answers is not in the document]
mayankshriv: I stand corrected, SQL does seem to have all three. Also, @Jackie-Jiang pointed out that we cannot specify 'e' in log(number, base).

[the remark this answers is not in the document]
Jackie-Jiang: MySQL support LN(), LOG(), LOG10(), LOG2(): https://www.w3schools.com/sql/sql_ref_mysql.asp
We can add later if needed.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/SingleParamMathTransformFunction.java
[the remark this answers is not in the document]
Jackie-Jiang: What is the failure message? Have you removed the data type switch? I think the test is reasonable.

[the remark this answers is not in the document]
xueyumusic: Hi, @Jackie-Jiang , actually it failed whether removing the switch or not. The exception stacktrace for this case is 
```
java.lang.UnsupportedOperationException
        at org.apache.pinot.core.segment.index.readers.BaseDictionary.getDoubleValue(BaseDictionary.java:45)
        at org.apache.pinot.core.segment.index.readers.BaseDictionary.readDoubleValues(BaseDictionary.java:86)
        at org.apache.pinot.core.common.DataFetcher.fetchDoubleValues(DataFetcher.java:166)
        at org.apache.pinot.core.common.DataBlockCache.getDoubleValuesForSVColumn(DataBlockCache.java:166)
        at org.apache.pinot.core.operator.docvalsets.ProjectionBlockValSet.getDoubleValuesSV(ProjectionBlockValSet.java:89)
        at org.apache.pinot.core.operator.transform.function.IdentifierTransformFunction.transformToDoubleValuesSV(IdentifierTransformFunction.java:95)
        at org.apache.pinot.core.operator.transform.function.SingleParamMathTransformFunction.transformToDoubleValuesSV(SingleParamMathTransformFunction.java:77)
        at org.apache.pinot.core.operator.transform.function.BaseTransformFunction.transformToIntValuesSV(BaseTransformFunction.java:83)
        at org.apache.pinot.core.operator.transform.function.BaseTransformFunctionTest.testTransformFunction(BaseTransformFunctionTest.java:146)
        at org.apache.pinot.core.operator.transform.function.SingleParamMathTransformFunctionTest.testAbsTransformFunction(SingleParamMathTransformFunctionTest.java:78)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
```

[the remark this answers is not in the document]
Jackie-Jiang: You are correct. The type conversion need to be pushed down to the dictionary level, and that involves bigger change. For now, you can fetch the doubleValues based on the type (cast values if needed), then applyMathOperator on double values.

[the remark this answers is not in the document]
xueyumusic: Thanks a lot for review, @Jackie-Jiang , I updated the codes to address your comments.

PR #3937, opened 2019-03-08
readthedocs: Update instructions for using fork

sunithabeeram:


docs/dev_env.rst
```diff
@@ -37,16 +37,21 @@ Getting the Source Code
...
 Clone the repository locally
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: none
 
-  $mkdir workspace
-  $cd workspace
-  $git clone https://github.com/<username>/incubator-pinot.git
-  $cd incubator-pinot
+  $ mkdir workspace
```
mcvsubbu: Can we move the section on how to create a fork into the contribution guidelines? We seem to say here that creating a branch is not recommended, yet that is what we write about in contribution guidelines. Maybe we should remove the text on creating branches, and just retain one on creating forks? Not sure what the limitations are in terms of branches created. Does Apache have a limit? Or, are we worried that people may not delete the branch after commit?

sunithabeeram: A branch on the Apache repo is not recommended. We almost always need branches.

sunithabeeram: On Apache repo, we would like to see branches that are used for releases (and not be cluttered with everything else that the community is working on)

mcvsubbu: In that case, isn't it better to remove the other section on creating branches?

docs/contribution_guidelines.rst
```diff
@@ -35,7 +35,15 @@ Create a branch for your change
 
 .. code-block:: none
 
-  $ cd incubator-pinot
+  $ cd pinot
```
mcvsubbu: Suggest: In line 23, last part of the sentence, add " and created your own fork of the pinot source code".

PR #3941, opened 2019-03-08
Pinot server side change to optimize LLC segment completion with direct metadata upload. 

chenboat:
As designed by @npawar and @mcvsubbu , this is the second part of the code change to optimize LLC segment completion with direct metadata upload. A server config is added to control the code path to upload metadata files or not. By default, it is true.

The code here has been tested on top of #3877. It will be committed after that.


pinot-core/src/main/java/org/apache/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
```diff
@@ -50,22 +51,30 @@
   private static SSLContext _sslContext;
   private static Integer _controllerHttpsPort;
 
+  private static boolean _commitWithMetadata;
+
   private final FileUploadDownloadClient _fileUploadDownloadClient;
   private final ServerMetrics _serverMetrics;
 
-  public static void init(Configuration uploaderConfig) {
+  public static void init(Configuration uploaderConfig, boolean commitWithMetadata) {
```
mcvsubbu: Can you move the config of commitWithMetadata to LLRealtimeSegmentDataManager? It is being checked in the other class, so makes sense to move it there. You may have a bit more config plumbing to do, but it is worth it

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/LLCRealtimeClusterSplitCommitEndWithMetadataTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.apache.pinot.common.utils.CommonConstants;
+import org.apache.pinot.controller.ControllerConf;
+
+/**
+ * Integration test that extends LLCRealtimeClusterIntegrationTest but with split commit enabled and segment commit end
+ * with metadata file upload enabled.
+ */
+public class LLCRealtimeClusterSplitCommitEndWithMetadataTest extends LLCRealtimeClusterIntegrationTest {
+
+    @Override
```
mcvsubbu: Can you please follow the same indentation level, etc? thanks

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/LLCRealtimeClusterSplitCommitEndWithMetadataTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.apache.commons.configuration.Configuration;
+import org.apache.pinot.common.utils.CommonConstants;
+import org.apache.pinot.controller.ControllerConf;
+
+/**
+ * Integration test that extends LLCRealtimeClusterIntegrationTest but with split commit enabled and segment commit end
+ * with metadata file upload enabled.
+ */
+public class LLCRealtimeClusterSplitCommitEndWithMetadataTest extends LLCRealtimeClusterIntegrationTest {
```
mcvsubbu: Instead of adding a new test (that increases test time),  we can just change the split commit test to just use the new controller endpoint to commit with metadata. We should deprecate the one that commits without metadata, since this one is far superior. The only reason to use the other one is for backward compat for one release.

pinot-core/src/main/java/org/apache/pinot/server/realtime/ServerSegmentCompletionProtocolHandler.java
chenboat: Done. Put the config into IndexLoadingConfig just like splitCommit.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -735,14 +742,45 @@ protected SegmentBuildDescriptor buildSegmentInternal(boolean forCommit) {
...
       params.withMemoryUsedBytes(_memoryManager.getTotalAllocatedBytes());
     }
-    SegmentCompletionProtocol.Response commitEndResponse = _protocolHandler.segmentCommitEnd(params);
+    SegmentCompletionProtocol.Response commitEndResponse;
+    if (!_indexLoadingConfig.isEnableSplitCommitEndWithMetadata()) {
+      commitEndResponse = _protocolHandler.segmentCommitEnd(params);
+    } else {
+      Map<String, File> metadataFiles = new HashMap<>();
+      try {
+        metadataFiles.put(V1Constants.MetadataKeys.METADATA_FILE_NAME,
```
mcvsubbu: Can you find a way to avoid extracting from the tar file again? CPU and disk IO are expensive on the server, and we would like to avoid it if we can. 

The only way I could think of was to store the metadata files in another directory and keep it along with the segment descriptor. Once the commit is completed, we need to take care to remove the directory, as we don't want to have cruft left-over. It has to be done a little carefully. @npawar  do you have other ideas?

chenboat: Let me explore a bit your idea to pass in the metadata files via the Segment Descriptor.

mcvsubbu: I think we keep the raw files and build the tar.gz in addition. The segment needs to be in raw format for serving anyways.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixInstanceDataManagerConfig.java
```diff
@@ -163,6 +165,11 @@ public boolean isEnableSplitCommit() {
     return _instanceDataManagerConfiguration.getBoolean(ENABLE_SPLIT_COMMIT, false);
   }
 
+  @Override
+  public boolean isEnableSplitCommitEndWithMetadata() {
+    return _instanceDataManagerConfiguration.getBoolean(ENABLE_SPLIT_COMMIT_END_WITH_METADATA, false);
```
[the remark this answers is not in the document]
chenboat: Done.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
chenboat: I see. Then we can just use the existing metadata files on the server -- and there is no need to delete any temp files. I tested it and it works. Please check the latest commits.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -681,11 +690,23 @@ protected SegmentBuildDescriptor buildSegmentInternal(boolean forCommit) {
           TimeUnit.MILLISECONDS.toSeconds(waitTimeMillis));
 
       if (forCommit) {
+        Preconditions.checkState(destDir.listFiles() != null && destDir.listFiles().length > 0,
+            "The index dir is empty: ", destDir);
+        File metadataFileName = new File(destDir.listFiles()[0], V1Constants.MetadataKeys.METADATA_FILE_NAME);
```
npawar: destDir.listFiles() is being called 4 times. Might be cleaner if we store it in a local variable

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixInstanceDataManagerConfig.java
```diff
@@ -163,6 +165,11 @@ public boolean isEnableSplitCommit() {
     return _instanceDataManagerConfiguration.getBoolean(ENABLE_SPLIT_COMMIT, false);
   }
 
+  @Override
+  public boolean isEnableSplitCommitEndWithMetadata() {
+    return _instanceDataManagerConfiguration.getBoolean(ENABLE_SPLIT_COMMIT_END_WITH_METADATA, true);
+  }
```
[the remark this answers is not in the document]
npawar: I see Subbu's comments about making default true. But the PR description says it will be default false. Please update the description in that case

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
chenboat: Done.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixInstanceDataManagerConfig.java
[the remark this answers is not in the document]
chenboat: Done.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -681,11 +690,23 @@ protected SegmentBuildDescriptor buildSegmentInternal(boolean forCommit) {
           TimeUnit.MILLISECONDS.toSeconds(waitTimeMillis));
 
       if (forCommit) {
+        Preconditions.checkState(destDir.listFiles() != null && destDir.listFiles().length > 0,
```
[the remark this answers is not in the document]
chenboat: Done.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -735,7 +765,13 @@ protected SegmentBuildDescriptor buildSegmentInternal(boolean forCommit) {
     if (_isOffHeap) {
       params.withMemoryUsedBytes(_memoryManager.getTotalAllocatedBytes());
     }
-    SegmentCompletionProtocol.Response commitEndResponse = _protocolHandler.segmentCommitEnd(params);
+    SegmentCompletionProtocol.Response commitEndResponse;
+    if (!_indexLoadingConfig.isEnableSplitCommitEndWithMetadata()) {
```
mcvsubbu: nit I would prefer the condition inverted, given that both if and else portions are one line long.

PR #3942, opened 2019-03-08
Set processingException when all queried segments cannot be acquired

sunithabeeram:
**Motivation**: We have seen cases where due to bugs or delays wrt EV processing on the broker, broker sends a list of segments to the servers that are not hosted by it. This is technically a partial-response, but we currently don't indicate it as such. We would like to track this to the most extent possible.

**Details**: There are cases where segments can be legitimately missing on the server; the known reason for this is when segments are deleted (either through retention or manually) and the broker's EV isn't updated yet. To not count partial responses in this scenario, we maintain an in-memory cache of deleted segment names. If a recently deleted segment is queried, we check our cache and not count it as an error if the segment-name exists in the cache. 
NOTE: This can still miss cases where the server is restarted in the window between deletion and broker EV update. However, the chances of this are deemed low (and cost of catching this error deemed high).

**Testing done**: Added unit tests to cover the new paths.

**Impact**: Once this change is rolled out, num-segments-missing server metric should report lower numbers; we should see partial-responses broker metric go up if we hit missing-segments issue on the server.

Please see https://github.com/apache/incubator-pinot/issues/4007 for details of the broader effort for this PR.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
@@ -168,6 +168,7 @@ public abstract class BaseTableDataManager implements TableDataManager {
   /**
    * Called when a segment is deleted. The actual handling of segment delete is outside of this method.
    * This method provides book-keeping around deleted segments.
+   * @param segmentName name of the segment to track.
    */
   public void trackDeletedSegment(@Nonnull String segmentName) {
     // add segment to the cache
@@ -176,6 +177,9 @@ public abstract class BaseTableDataManager implements TableDataManager {
 
   /**
    * Check if a segment is recently deleted.
+   *
+   * @param segmentName name of the segment to check.
+   * @return true if segment is in the cache, false otherwise
    */
   public boolean isRecentlyDeleted(@Nonnull String segmentName) {
     return _deletedSegmentsCache.getIfPresent(segmentName) != null;
@@ -183,6 +187,8 @@ public abstract class BaseTableDataManager implements TableDataManager {
 
   /**
    * Remove a segment from the deleted cache if it is being added back.
+   *
+   * @param segmentName name of the segment that needs to removed from the cache (if needed)
    */
   private void untrackIfDeleted(@Nonnull String segmentName) {
     _deletedSegmentsCache.invalidate(segmentName);
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
@@ -19,6 +19,8 @@
 package org.apache.pinot.core.data.manager;
 
 import com.google.common.base.Preconditions;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
@@ -42,8 +44,12 @@ import org.slf4j.LoggerFactory;
 @ThreadSafe
 public abstract class BaseTableDataManager implements TableDataManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(BaseTableDataManager.class);
+  // cache atmost a given max of deleted segments
+  private static final int MAX_DELETED_SEGMENT_CACHE_SIZE = 1000;
 
   protected final ConcurrentHashMap<String, SegmentDataManager> _segmentDataManagerMap = new ConcurrentHashMap<>();
+
+  protected Cache<String, Boolean> _deletedSegmentsCache;
 
   protected TableDataManagerConfig _tableDataManagerConfig;
   protected String _instanceId;
@@ -59,6 +65,7 @@ public abstract class BaseTableDataManager implements TableDataManager {
       @Nonnull ZkHelixPropertyStore<ZNRecord> propertyStore, @Nonnull ServerMetrics serverMetrics) {
     LOGGER.info("Initializing table data manager for table: {}", tableDataManagerConfig.getTableName());
 
+    _deletedSegmentsCache = CacheBuilder.newBuilder().maximumSize(MAX_DELETED_SEGMENT_CACHE_SIZE).build();
     _tableDataManagerConfig = tableDataManagerConfig;
     _instanceId = instanceId;
     _propertyStore = propertyStore;
@@ -156,6 +163,23 @@ public abstract class BaseTableDataManager implements TableDataManager {
     }
   }
 
+  /**
+   * Called when a segment is deleted. The actual handling of semgent delete is outside of this method.
+   * This method provides book-keeping around deleted segments.
+   */
+  public void deleteSegment(@Nonnull String segmentName) {
+
+    // add segment to the cache
+    _deletedSegmentsCache.put(segmentName, true);
+  }
+
+  /**
+   * Check if a segment is recently deleted.
+   */
+  public boolean isDeleted(@Nonnull String segmentName) {
+    return _deletedSegmentsCache.getIfPresent(segmentName) != null;
+  }
+
   @Nonnull
   @Override
   public List<SegmentDataManager> acquireAllSegments() {
@@ -176,8 +200,6 @@ public abstract class BaseTableDataManager implements TableDataManager {
       SegmentDataManager segmentDataManager = _segmentDataManagerMap.get(segmentName);
       if (segmentDataManager != null && segmentDataManager.increaseReferenceCount()) {
         segmentDataManagers.add(segmentDataManager);
-      } else {
-        handleMissingSegment(segmentName);
       }
     }
     return segmentDataManagers;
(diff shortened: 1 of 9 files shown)


pinot-core/src/main/java/org/apache/pinot/core/data/manager/TableDataManager.java
```diff
@@ -79,6 +79,16 @@ void addSegment(@Nonnull String segmentName, @Nonnull TableConfig tableConfig,
    */
   void removeSegment(@Nonnull String segmentName);
 
+  /**
+   * Track a deleted segment.
+   */
+  void deleteSegment(@Nonnull String segmentName);
```
snleee: Can we rename this function? `deleteSegment` sounds like we are deleting segments instead of book-keeping.  It's confusing since we have another method called `removeSegment`.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
```diff
@@ -42,9 +44,13 @@
 @ThreadSafe
 public abstract class BaseTableDataManager implements TableDataManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(BaseTableDataManager.class);
+  // cache atmost a given max of deleted segments
+  private static final int MAX_DELETED_SEGMENT_CACHE_SIZE = 1000;
```
snleee: Discussed offline. We can probably add the time based eviction and increase `MAX_DELETED_SEGMENT_CACHE_SIZE`.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/TableDataManager.java
mcvsubbu: +1

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -124,7 +125,22 @@ public DataTable processQuery(ServerQueryRequest queryRequest, ExecutorService e
...
+    // acquire the segments
+    int missingSegments = 0;
+    List<String> segmentsToQuery = queryRequest.getSegmentsToQuery();
+    List<SegmentDataManager> segmentDataManagers = new ArrayList<>();
+    for (String segmentName : segmentsToQuery) {
+      SegmentDataManager segmentDataManager = tableDataManager.acquireSegment(segmentName);
+      if (segmentDataManager != null) {
+        segmentDataManagers.add(segmentDataManager);
+      } else {
+        if (!tableDataManager.isDeleted(segmentName)) {
```
mcvsubbu: isRecentlyDeleted()?

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -195,6 +211,13 @@ public DataTable processQuery(ServerQueryRequest queryRequest, ExecutorService e
     long queryProcessingTime = queryProcessingTimer.getDurationMs();
     dataTable.getMetadata().put(DataTable.NUM_SEGMENTS_QUERIED, Integer.toString(numSegmentsQueried));
     dataTable.getMetadata().put(DataTable.TIME_USED_MS_METADATA_KEY, Long.toString(queryProcessingTime));
+
+    if (missingSegments > 0) {
+      dataTable.addException(QueryException.getException(QueryException.SEGMENTS_MISSING_ERROR,
```
mcvsubbu: It may be worth adding a log message here with the name of the segments missing, Can help us debug

pinot-core/src/main/java/org/apache/pinot/core/data/manager/InstanceDataManager.java
```diff
@@ -78,6 +78,13 @@ void addRealtimeSegment(@Nonnull String realtimeTableName, @Nonnull String segme
   void removeSegment(@Nonnull String tableNameWithType, @Nonnull String segmentName)
       throws Exception;
 
+  /**
+   * Handles deletion of a segment from the table.
+   *
+   * This method performs book keeping of deleted segments.
+   */
+  void deleteSegment(@Nonnull String tableNameWithType, @Nonnull String segmentName);
```
mcvsubbu: same with this method name. It is not really deleting a segment, is it?

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixInstanceDataManager.java
```diff
@@ -151,6 +151,15 @@ public void removeSegment(@Nonnull String tableNameWithType, @Nonnull String seg
     }
   }
 
+  @Override
+  public void deleteSegment(@Nonnull String tableNameWithType, @Nonnull String segmentName) {
+    LOGGER.info("Tracking deleted segment: {} from table: {}", segmentName, tableNameWithType);
```
mcvsubbu: I think we have other logs for dropping a segment, do we need this one as well? If we really do, then it should go inside the 'if' statement

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/SegmentOnlineOfflineStateModelFactory.java
```diff
@@ -196,6 +196,8 @@ public void onBecomeDroppedFromOffline(Message message, NotificationContext cont
       String tableNameWithType = message.getResourceName();
       String segmentName = message.getPartitionName();
 
+      _instanceDataManager.deleteSegment(tableNameWithType, segmentName);
```
mcvsubbu: This line should be inserted after line 184, not here. The segment is not available for any new queries sometime after line 184 starts execution

mcvsubbu: Also, can we not get the table data manager directly from here? Do we need to go through the instanceDataManager? Dependency issues?

kishoreg:
Agree with comments. Looking at the existing code, I felt it could be handled within BaseTableDataManager - have the cache (you have it), just update it in removeSegment and use that in handleMissingSegment (don't increment the metric). Am I missing something?

pinot-core/src/main/java/org/apache/pinot/core/data/manager/InstanceDataManager.java
sunithabeeram: Will change the method name.

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
sunithabeeram: Good point; will add it around line 138

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/SegmentOnlineOfflineStateModelFactory.java
sunithabeeram: After line 184, the segment is unavailable correct; and if it is due to a move/rebalance, we should count it as an error. We only want to suppress it when the segment has been 'deleted'.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/TableDataManager.java
sunithabeeram: Done. Renamed to trackDeletedSegment

pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
sunithabeeram: Updated.

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
sunithabeeram: Renamed.

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
sunithabeeram: Done.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixInstanceDataManager.java
sunithabeeram: Removed log.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
```diff
@@ -156,6 +165,35 @@ public void removeSegment(@Nonnull String segmentName) {
...
+  public boolean isRecentlyDeleted(@Nonnull String segmentName) {
+    return _deletedSegmentsCache.getIfPresent(segmentName) != null;
+  }
+
+  /**
+   * Remove a segment from the deleted cache if it is being added back.
+   *
+   * @param segmentName name of the segment that needs to removed from the cache (if needed)
+   */
+  private void untrackIfDeleted(@Nonnull String segmentName) {
```
snleee: In my opinion, `untrackDeletedSegment()` seems to be better.

sunithabeeram: I am ambivalent about this. If you feel strongly, I can change it. Will see if Subbu has any input.

mcvsubbu: Or simply, notifySegmentDeleted() and notifySegmentAdded()?

pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
```diff
@@ -110,6 +118,7 @@ public void shutDown() {
   @Override
   public void addSegment(@Nonnull ImmutableSegment immutableSegment) {
     String segmentName = immutableSegment.getSegmentName();
+    untrackIfDeleted(segmentName);
```
mcvsubbu: This line should move down to be after line 128 (when we populate the _segmentDataManagerMap). The segment is available for query only after that point.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
sunithabeeram: I think thats a better a choice. Thanks for the suggestion!

pinot-core/src/main/java/org/apache/pinot/core/data/manager/BaseTableDataManager.java
sunithabeeram: Done.

PR #3943, opened 2019-03-08
Segment creation with flexible file paths

jenniferdai:
* Currently, we enforce segments to have certain file extensions. If record reader path is there, we will allow any extensions.


pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -104,6 +104,10 @@ public SegmentCreationJob(Properties properties) {
 
   @Override
   protected boolean isDataFile(String fileName) {
+    // Other files may have different extensions, eg: orc can have no extension
```
sunithabeeram: Other files =>Might be better to reword: "When user specified record-readers are used, files can have different/unknown extensions. e.g., orc can have no extension. Skip the check in this case."

PR #3946, opened 2019-03-11
Make RetentionManager and OfflineSegmentIntervalChecker initial delays configurable

sunithabeeram:
*Motivation*: In one of our clusters that hosts 1000's of tables with 1000's of segments each, we are noticing frequent controller leadership changes due to long GC pauses. Analysis shows that the GC pressure might be due to heavy hitting tasks running concurrently. RetentionManager and OfflineSegmentIntervalChecker tasks hold tables segment metadata in memory while working with tables and are likely to cause high GC pressure when run concurrently.  While the issue can be addressed in many ways, this change attempts a simpler/short-term fix to stagger the start time by making them configurable.

*Testing done*: Existing tests pass.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
@@ -84,6 +84,10 @@ public class ControllerConf extends PropertiesConfiguration {
     // Initial delays
     private static final String STATUS_CHECKER_INITIAL_DELAY_IN_SECONDS =
         "controller.statusChecker.initialDelayInSeconds";
+    private static final String RETENTION_MANAGER_INITIAL_DELAY_IN_SECONDS =
+        "controller.retentionManager.initialDelayInSeconds";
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -84,6 +84,10 @@
     // Initial delays
     private static final String STATUS_CHECKER_INITIAL_DELAY_IN_SECONDS =
         "controller.statusChecker.initialDelayInSeconds";
+    private static final String RETENTION_MANAGER_INITIAL_DELAY_IN_SECONDS =
```
jackjlli: It would be good to leave some comments to stagger these two tasks here.

sunithabeeram: For the test I was pursuing, staggering didn't make a difference. I will however check this in to have the flexibility in the future.

PR #3966, opened 2019-03-13
[TE] detection - merger max duration validation

jihaozh:
Max duration must be a positive number and more than 15 minutes.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/MergeWrapper.java
```diff
@@ -87,7 +87,7 @@ public MergeWrapper(DataProvider provider, DetectionConfigDTO config, long start
 
     this.maxGap = MapUtils.getLongValue(config.getProperties(), "maxGap", TimeUnit.HOURS.toMillis(2));
     this.maxDuration = MapUtils.getLongValue(config.getProperties(), "maxDuration", Long.MAX_VALUE);
-    Preconditions.checkArgument(this.maxDuration > 0, "Max duration must be a positive number.");
+    Preconditions.checkArgument(this.maxDuration > 0 && this.maxDuration <= TimeUnit.MINUTES.toMillis(15), "Max duration must be a positive number and more than 15 minutes");
```
akshayrai: shouldn't it be `maxDuration >= 15 mins`?

jihaozh: Yeah. I'll just check the number of anomalies instead.

PR #3967, opened 2019-03-13
[TE] Yaml validator for maxDuration

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/validators/DetectionConfigValidator.java
```diff
@@ -90,6 +93,16 @@ public void validateConfig(DetectionConfigDTO detectionConfig) throws Validation
       throw new ValidationException("The detection cron specified is incorrect. Please verify your cron expression"
           + " using online cron makers.");
     }
+    // Validate parameters under properties
+    if (detectionConfig.getProperties() != null) {
+      Long maxDuration = MapUtils.getLong(detectionConfig.getProperties(), PROP_MAX_DURATION);
```
jihaozh: The max duration actually not in the properties map. It's in the nested merge wrapper properties.

PR #3979, opened 2019-03-18
Track "freshness" timestamp across consuming segments

sunithabeeram:
*Motivation*: We would like to track freshness of the realtime data being consumed and available for queries. The true indication of this could be a timestamp for when the message was accepted into the stream-provider's (such as Kafka's) queue. We build on the change in #4025 to expose this information when available - and fall back to the last time any message was indexed in the segment as a proxy if that information is not available.
Exposing this through the query path helps comparison against client defined SLAs and provides ability to reflect freshness as part of Pinot's availability.

*Changes*: The PR exposes lastIndexedTimestamp and latestIngestionTimestamp from consuming segments for a query. The min timestamp across consuming segments is obtained by the server query-executor before the segments are pruned out. This along with the number of consuming segments queried is returned to the broker in the response.

*Testing done*: Unit and integration tests added.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
@@ -16,28 +16,21 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.pinot.core.indexsegment.mutable;
+package org.apache.pinot.common.metadata;
 
-import org.apache.pinot.common.metadata.RowMetadata;
-import org.apache.pinot.core.data.GenericRow;
-import org.apache.pinot.core.indexsegment.IndexSegment;
+import org.apache.pinot.annotations.InterfaceAudience;
 
 
-public interface MutableSegment extends IndexSegment {
-
-  /**
-   * Indexes a record into the segment with optionally provided metadata.
-   *
-   * @param row Record represented as a {@link GenericRow}
-   * @param rowMetadata the metadata associated with the message
-   * @return Whether the segment is full (i.e. cannot index more record into it)
-   */
-   boolean index(GenericRow row, RowMetadata rowMetadata);
+/**
+ * A class that provides relevant row-level metadata for rows ingested into a segment.
+ *
+ * Currently this is relevant for rows ingested into a mutable segment.
+ */
+@InterfaceAudience.Public
+public interface RowMetadata {
 
   /**
-   * Returns the number of records already indexed into the segment.
-   *
-   * @return The number of records indexed
+   * Return the ingestion timestamp of the row.
    */
-  int getNumDocsIndexed();
+  long getIngestionTimestamp();
 }
--- a/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
@@ -34,5 +34,5 @@ public interface RowMetadata {
   /**
    * Return the ingestion timestamp of the row.
    */
-  long getIngestionTimestamp();
+  long getIngestionTimesMs();
 }
--- a/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
@@ -34,5 +34,5 @@ public interface RowMetadata {
   /**
    * Return the ingestion timestamp of the row.
    */
-  long getIngestionTimesMs();
+  long getIngestionTimeMs();
 }
--- a/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
@@ -19,6 +19,7 @@
 package org.apache.pinot.common.metadata;
 
 import org.apache.pinot.annotations.InterfaceAudience;
+import org.apache.pinot.annotations.InterfaceStability;
 
 
 /**
@@ -27,6 +28,7 @@ import org.apache.pinot.annotations.InterfaceAudience;
  * Currently this is relevant for rows ingested into a mutable segment.
  */
 @InterfaceAudience.Public
+@InterfaceStability.Evolving
 public interface RowMetadata {
 
   /**
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
@@ -39,7 +39,7 @@ import org.apache.pinot.common.utils.JsonUtils;
  *
  * Supports serialization via JSON.
  */
-@JsonPropertyOrder({"selectionResults", "aggregationResults", "exceptions", "numServersQueried", "numServersResponded", "numSegmentsQueried", "numSegmentsProcessed", "numSegmentsMatched", "numDocsScanned", "numEntriesScannedInFilter", "numEntriesScannedPostFilter", "numGroupsLimitReached", "totalDocs", "timeUsedMs", "segmentStatistics", "traceInfo"})
+@JsonPropertyOrder({"selectionResults", "aggregationResults", "exceptions", "numServersQueried", "numServersResponded", "numSegmentsQueried", "numSegmentsProcessed", "numSegmentsMatched", "numConsumingSegmentsQueried", "numDocsScanned", "numEntriesScannedInFilter", "numEntriesScannedPostFilter", "numGroupsLimitReached", "totalDocs", "timeUsedMs", "segmentStatistics", "traceInfo"})
 public class BrokerResponseNative implements BrokerResponse {
   public static final BrokerResponseNative EMPTY_RESULT = BrokerResponseNative.empty();
   public static final BrokerResponseNative NO_TABLE_RESULT =
@@ -54,7 +54,9 @@ public class BrokerResponseNative implements BrokerResponse {
   private long _numSegmentsProcessed = 0L;
   private long _numSegmentsMatched = 0L;
   private long _numConsumingSegmentsQueried = 0L;
-  private long _minLastIngestionTimeMs = 0L;
+  // the timestamp indicating the freshness of the data queried in consuming segments.
+  // This can be ingestion timestamp if provided by the stream, or the last index time
+  private long _minConsumingFreshnessTimeMs = 0L;
 
   private long _totalDocs = 0L;
   private boolean _numGroupsLimitReached = false;
@@ -205,24 +207,24 @@ public class BrokerResponseNative implements BrokerResponse {
     _numSegmentsMatched = numSegmentsMatched;
   }
 
-  @JsonProperty("numConsumingQueried")
+  @JsonProperty("numConsumingSegmentsQueried")
   public long getNumConsumingSegmentsQueried() {
     return _numConsumingSegmentsQueried;
   }
 
-  @JsonProperty("numConsumingQueried")
+  @JsonProperty("numConsumingSegmentsQueried")
   public void setNumConsumingSegmentsQueried(long queried) {
     _numConsumingSegmentsQueried = queried;
   }
 
   @JsonProperty("minConsumingFreshnessTimeMs")
-  public long getMinLastIngestionTimeMs() {
-    return _minLastIngestionTimeMs;
+  public long getMinConsumingFreshnessTimeMs() {
+    return _minConsumingFreshnessTimeMs;
   }
 
   @JsonProperty("minConsumingFreshnessTimeMs")
-  public void setMinLastIngestionTimeMs(long ts) {
-    _minLastIngestionTimeMs = ts;
+  public void setMinConsumingFreshnessTimeMs(long ts) {
+    _minConsumingFreshnessTimeMs = ts;
   }
 
 
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
@@ -54,7 +54,7 @@ public class BrokerResponseNative implements BrokerResponse {
   private long _numSegmentsProcessed = 0L;
   private long _numSegmentsMatched = 0L;
   private long _numConsumingQueried = 0L;
-  private long _minConsumingIndexTs = 0L;
+  private long _minConsumingFreshnessTimeMs = 0L;
 
   private long _totalDocs = 0L;
   private boolean _numGroupsLimitReached = false;
@@ -215,14 +215,14 @@ public class BrokerResponseNative implements BrokerResponse {
     _numConsumingQueried = queried;
   }
 
-  @JsonProperty("minConsumingIndexTs")
-  public long getMinConsumingIndexTs() {
-    return _minConsumingIndexTs;
+  @JsonProperty("minConsumingFreshnessTimeMs")
+  public long getMinConsumingFreshnessTimeMs() {
+    return _minConsumingFreshnessTimeMs;
   }
 
-  @JsonProperty("minConsumingIndexTs")
-  public void setMinConsumingIndexTs(long ts) {
-    _minConsumingIndexTs = ts;
+  @JsonProperty("minConsumingFreshnessTimeMs")
+  public void setMinConsumingFreshnessTimeMs(long ts) {
+    _minConsumingFreshnessTimeMs = ts;
   }
 
 
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
@@ -53,6 +53,8 @@ public class BrokerResponseNative implements BrokerResponse {
   private long _numSegmentsQueried = 0L;
   private long _numSegmentsProcessed = 0L;
   private long _numSegmentsMatched = 0L;
+  private long _numConsumingQueried = 0L;
+  private long _minConsumingIndexTs = 0L;
 
   private long _totalDocs = 0L;
   private boolean _numGroupsLimitReached = false;
@@ -202,6 +204,27 @@ public class BrokerResponseNative implements BrokerResponse {
   public void setNumSegmentsMatched(long numSegmentsMatched) {
     _numSegmentsMatched = numSegmentsMatched;
   }
+
+  @JsonProperty("numConsumingQueried")
+  public long getNumConsumingQueried() {
+    return _numConsumingQueried;
+  }
+
+  @JsonProperty("numConsumingQueried")
+  public void setNumConsumingQueried(long queried) {
+    _numConsumingQueried = queried;
+  }
+
+  @JsonProperty("minConsumingIndexTs")
+  public long getMinConsumingIndexTs() {
+    return _minConsumingIndexTs;
+  }
+
+  @JsonProperty("minConsumingIndexTs")
+  public void setMinConsumingIndexTs(long ts) {
+    _minConsumingIndexTs = ts;
+  }
+
 
   @JsonProperty("totalDocs")
   @Override
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
@@ -53,8 +53,8 @@ public class BrokerResponseNative implements BrokerResponse {
   private long _numSegmentsQueried = 0L;
   private long _numSegmentsProcessed = 0L;
   private long _numSegmentsMatched = 0L;
-  private long _numConsumingQueried = 0L;
-  private long _minConsumingFreshnessTimeMs = 0L;
+  private long _numConsumingSegmentsQueried = 0L;
+  private long _minLastIngestionTimeMs = 0L;
 
   private long _totalDocs = 0L;
   private boolean _numGroupsLimitReached = false;
@@ -206,23 +206,23 @@ public class BrokerResponseNative implements BrokerResponse {
   }
 
   @JsonProperty("numConsumingQueried")
-  public long getNumConsumingQueried() {
-    return _numConsumingQueried;
+  public long getNumConsumingSegmentsQueried() {
+    return _numConsumingSegmentsQueried;
   }
 
   @JsonProperty("numConsumingQueried")
-  public void setNumConsumingQueried(long queried) {
-    _numConsumingQueried = queried;
+  public void setNumConsumingSegmentsQueried(long queried) {
+    _numConsumingSegmentsQueried = queried;
   }
 
   @JsonProperty("minConsumingFreshnessTimeMs")
-  public long getMinConsumingFreshnessTimeMs() {
-    return _minConsumingFreshnessTimeMs;
+  public long getMinLastIngestionTimeMs() {
+    return _minLastIngestionTimeMs;
   }
 
   @JsonProperty("minConsumingFreshnessTimeMs")
-  public void setMinConsumingFreshnessTimeMs(long ts) {
-    _minConsumingFreshnessTimeMs = ts;
+  public void setMinLastIngestionTimeMs(long ts) {
+    _minLastIngestionTimeMs = ts;
   }
 
 
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
@@ -213,8 +213,8 @@ public class BrokerResponseNative implements BrokerResponse {
   }
 
   @JsonProperty("numConsumingSegmentsQueried")
-  public void setNumConsumingSegmentsQueried(long queried) {
-    _numConsumingSegmentsQueried = queried;
+  public void setNumConsumingSegmentsQueried(long numConsumingSegmentsQueried) {
+    _numConsumingSegmentsQueried = numConsumingSegmentsQueried;
   }
 
   @JsonProperty("minConsumingFreshnessTimeMs")
@@ -223,8 +223,8 @@ public class BrokerResponseNative implements BrokerResponse {
   }
 
   @JsonProperty("minConsumingFreshnessTimeMs")
-  public void setMinConsumingFreshnessTimeMs(long ts) {
-    _minConsumingFreshnessTimeMs = ts;
+  public void setMinConsumingFreshnessTimeMs(long minConsumingFreshnessTimeMs) {
+    _minConsumingFreshnessTimeMs = minConsumingFreshnessTimeMs;
   }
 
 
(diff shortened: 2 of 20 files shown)


pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
```diff
@@ -203,6 +205,27 @@ public void setNumSegmentsMatched(long numSegmentsMatched) {
...
+  public long getNumConsumingQueried() {
+    return _numConsumingQueried;
+  }
+
+  @JsonProperty("numConsumingQueried")
+  public void setNumConsumingQueried(long queried) {
+    _numConsumingQueried = queried;
+  }
+
+  @JsonProperty("minConsumingIndexTs")
```
mcvsubbu: Instead of 'Ts' could we suffix it with Millis (or Seconds, whatever the units may be)

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -168,6 +180,14 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
     brokerResponseNative.setNumSegmentsMatched(numSegmentsMatched);
     brokerResponseNative.setTotalDocs(numTotalRawDocs);
     brokerResponseNative.setNumGroupsLimitReached(numGroupsLimitReached);
+    if (numConsumingQueried > 0) {
+      if (minConsumingIndexTs == Long.MAX_VALUE) {
+        LOGGER.error("Invalid lastIndexedTimestamp across {} consuming segments", numConsumingQueried);
```
mcvsubbu: Log the query ID here so that we can use it to correlate query logs. Also, please ensure that if this condition happens we override the throttle in the broker logs.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -204,20 +205,22 @@ public boolean index(GenericRow row) {
...
     } else {
-      Preconditions
-          .checkState(_aggregateMetrics, "Invalid document-id during indexing: " + docId + " expected: " + numDocs);
+      Preconditions.checkState(_aggregateMetrics, "Invalid document-id during indexing: " + docId + " expected: " + numDocs);
       // Update metrics for existing document.
-      return aggregateMetrics(row, docId);
+      canTakeMore = aggregateMetrics(row, docId);
     }
+    // update indexing time
+    _lastIndexedTimestamp = System.currentTimeMillis();
```
mcvsubbu: If there is a delay in the stream, that will show up as a pinot delay. Another choice is to take the last _polled_ timestamp. That will ensure that as long as Pinot polls the stream, the available data is considered fresh. Not sure why we are not doing that.  A stream might (and probably should) have its own staleness metric being tracked and monitored so that an application does not need to track a staleness metric via pinot and root cause it to the stream if that is indeed the case.

If you want to continue with this approach, then I suggest that you extend the interface to retrieve rows from a stream to include the highest available timestamp of a record _in_ the stream. That way, streams that provide the staleness metric can translate (via Pinot) to the staleness monitoring application.

sunithabeeram: The user defined SLA will account for any expected delays in the incoming stream itself. Just going by the poll time can miss cases where we have issues with indexing (there have been a few in the past; for ex, issues with aggregatable-metrics for multi-value columns etc).

The timestamp in the record can have a very coarse granularity, for ex, days - which isn't ideal for the granularity we would like to track freshness at. Also, see motivation notes about the lack of other usable timestamps.

mcvsubbu: Issues with indexing the row should be dealt with as a bug and fixed, since that is a loss of data. 

Also, I was proposing to extend the interface so that a stream that keeps track of timestamps of incoming messages (for its own SLA) can use the mechanism. I was not referring to the time column of the schema

We can chat a little to see if any of these ideas can be explored further before discarding

pinot-common/src/main/java/org/apache/pinot/common/utils/DataTable.java
```diff
@@ -35,6 +35,8 @@
   String NUM_SEGMENTS_QUERIED = "numSegmentsQueried";
   String NUM_SEGMENTS_PROCESSED = "numSegmentsProcessed";
   String NUM_SEGMENTS_MATCHED = "numSegmentsMatched";
+  String NUM_CONSUMING_SEGMENTS_QUERIED = "numConsumingQueried";
+  String MIN_CONSUMING_TIMESTAMP = "minConsumingTs";
```
sunithabeeram: I need a better name for this - its meant to indicate the min latest ingestion timestamp across segments.

sunithabeeram:
@Jackie-Jiang - thanks for raising the concerns. I have captured some responses here. We can discuss further in person if needed.
> We need to have a design doc for all the interfaces and do a team discussion on this
We did have an internal discussion exactly on this today and its unfortunate that while you were present during that discussion the concerns weren't brought up there. We have an internal doc discussing the details - I will create an external version of the doc soon.

> StreamMessageMetadata is not general enough, and is a concrete class which is introduced to the general interface of MutableSegment
I am not sure its a good idea to make everything as an interface simply because its getting exposed through an interface. I can see that the name StreamMessageMetadata does tie it to streams a bit more than we want associated with MutableSegment - I can change this to RowMetadata.

> 2. The ingestion time and index time do not make sense for offline (immutable) segment but got added to the SegmentMetadata interface.
I definitely went down the path of splitting SegmentMetadata interface into mutable and immutable. Currently, SegmentMetadata exposes methods such as refreshTime(), pushTime() which don't quite apply to mutable data. I finally backed out of that path because of the way MutableSegmentImpl currently manages the segment-metadata (using inline variables instead of explicit setters/getters) - and given that the metadata gets update for every row ingested, I decided to keep with the current approach than introduce additional overhead by making explicit method calls everytime.

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -149,6 +150,33 @@ public DataTable processQuery(ServerQueryRequest queryRequest, ExecutorService e
       TraceContext.register(requestId);
     }
 
+    int numConsuming = 0;
+    long minIndexTime = Long.MAX_VALUE;
+    long minIngestionTime = Long.MAX_VALUE;
+    // gather stats for realtime consuming segments
+    for (SegmentDataManager segmentMgr : segmentDataManagers) {
```
mcvsubbu: should we be doing this after pruning segments?

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
mcvsubbu: Can you add the query ID here?

pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
```diff
@@ -53,6 +53,8 @@
   private long _numSegmentsQueried = 0L;
   private long _numSegmentsProcessed = 0L;
   private long _numSegmentsMatched = 0L;
+  private long _numConsumingQueried = 0L;
```
Jackie-Jiang: Can we change this to **_numConsumingSegmentsQueried**? The current name is not clear to me without reading the code.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/IndexSegment.java
```diff
@@ -29,6 +29,15 @@
 @InterfaceAudience.Private
 public interface IndexSegment {
 
+  public enum IndexSegmentType{
```
Jackie-Jiang: Why do we need this? On this level we should not have concept of IMMUTABLE or MUTABLE. We can do `if (indexSegment instanceOf MutableSegment)` if necessary.

pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
```diff
@@ -53,6 +53,8 @@
...
+  private long _minConsumingFreshnessTimeMs = 0L;
```
Jackie-Jiang: _minLastIngestionTimeMs?

pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
```diff
@@ -0,0 +1,38 @@
...
+ * Currently this is relevant for rows ingested into a mutable segment.
+ */
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
+public interface RowMetadata {
+
+  /**
+   * Return the ingestion timestamp of the row.
+   */
+  long getIngestionTimestamp();
```
Jackie-Jiang: getIngestionTimeMs()?

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageMetadata.java
```diff
@@ -18,24 +18,22 @@
...
 
 
 /**
  * A class that provides metadata associated with the message of a stream, for e.g.,
  * ingestion-timestamp of the message.
  */
-@InterfaceAudience.Public
-@InterfaceStability.Evolving
-public class StreamMessageMetadata {
+public class StreamMessageMetadata implements RowMetadata {
```
Jackie-Jiang: Let's rename this to DefaultRowMetadata or StreamRowMetadata? In the future we might want to make this class support more types of metadata

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageMetadata.java
```diff
@@ -18,24 +18,22 @@
...
 
   private long _ingestionTimestamp = Long.MIN_VALUE;
```
Jackie-Jiang: Make this final and set it in constructor? I don't think we need to recycle this long value wrapper.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/MessageBatch.java
```diff
@@ -62,9 +63,9 @@
    * when the message was ingested by the upstream stream-provider and other relevant metadata.
    *
    */
-  default void getMetadataAtIndex(int index, StreamMessageMetadata metadata) {
+  default RowMetadata getMetadataAtIndex(int index) {
     // update metadata as required - default implementation does nothing
```
Jackie-Jiang: Please update the comments

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaStreamLevelConsumer.java
```diff
@@ -87,11 +88,12 @@ public GenericRow next(GenericRow destination) {
     return next(destination, null);
   }
 
+  // NOTE: currently metadata is not updated as the simple consumer does not expose the relevant information
   @Override
-  public GenericRow next(GenericRow destination, StreamMessageMetadata metadata) {
+  public GenericRow next(GenericRow destination, RowMetadata metadata) {
```
Jackie-Jiang: Please remove this API and keep the old one (without passing in RowMetadata). If we have to return metadata from this method in the future, we can put metadata into GenericRow.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamLevelConsumer.java
```diff
@@ -47,10 +48,9 @@ void start()
   /**
    * Get next row from the stream and decode it into a generic row
    * @param destination the decoded generic row
-   * @param metadata the metadata associated with the decoded stream if any
    * @return the decoded generic row
    */
-  default GenericRow next(GenericRow destination, StreamMessageMetadata metadata) {
+  default GenericRow next(GenericRow destination, RowMetadata metadata) {
```
Jackie-Jiang: Please remove this API since it's not used right now. This interface is public and stable, so let's be careful about it.

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
sunithabeeram: The idea is to get the stats for queried segments, so will keep it the same.

sunithabeeram: For the question on ReduceService below, we currently don't get the queryId passed through to the BrokerReduceService. Will address that in a followup.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/IndexSegment.java
sunithabeeram: I don't think I agree with this. Exposing a type is a much cleaner way of handling this than doing instanceOf.

pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
sunithabeeram: Updated.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageMetadata.java
sunithabeeram: I think StreamMessageMetadata works better (we use messages in the context of streams).

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamLevelConsumer.java
sunithabeeram: I wanted plumb-in the implementation for both HLC and LLC - so its less confusing in the future. Given there is no specific use case for this now, I will drop it - once things look good with LLC, I might update the interface to make both support similar functionality.

pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
```diff
@@ -203,6 +207,27 @@ public void setNumSegmentsMatched(long numSegmentsMatched) {
     _numSegmentsMatched = numSegmentsMatched;
   }
 
+  @JsonProperty("numConsumingSegmentsQueried")
+  public long getNumConsumingSegmentsQueried() {
+    return _numConsumingSegmentsQueried;
+  }
+
+  @JsonProperty("numConsumingSegmentsQueried")
+  public void setNumConsumingSegmentsQueried(long queried) {
```
Jackie-Jiang: (Optional) change the parameter to be **numConsumingSegmentsQueried** (follow the convention of other methods)

pinot-common/src/main/java/org/apache/pinot/common/response/broker/BrokerResponseNative.java
```diff
@@ -203,6 +207,27 @@ public void setNumSegmentsMatched(long numSegmentsMatched) {
...
+    _numConsumingSegmentsQueried = queried;
+  }
+
+  @JsonProperty("minConsumingFreshnessTimeMs")
+  public long getMinConsumingFreshnessTimeMs() {
+    return _minConsumingFreshnessTimeMs;
+  }
+
+  @JsonProperty("minConsumingFreshnessTimeMs")
+  public void setMinConsumingFreshnessTimeMs(long ts) {
```
Jackie-Jiang: (Optional) change the parameter to be **minConsumingFreshnessTimeMs** (follow the convention of other methods)

pinot-common/src/main/java/org/apache/pinot/common/utils/DataTable.java
```diff
@@ -35,6 +35,8 @@
   String NUM_SEGMENTS_QUERIED = "numSegmentsQueried";
   String NUM_SEGMENTS_PROCESSED = "numSegmentsProcessed";
   String NUM_SEGMENTS_MATCHED = "numSegmentsMatched";
+  String NUM_CONSUMING_SEGMENTS_QUERIED = "numConsumingQueried";
```
Jackie-Jiang: **"numConsumingSegmentsQueried"**?

pinot-common/src/main/java/org/apache/pinot/common/utils/DataTable.java
```diff
@@ -35,6 +35,8 @@
...
+  String MIN_CONSUMING_FRESHNESS_TIMESTAMP = "minConsumingFreshnessTs";
```
Jackie-Jiang: **MIN_CONSUMING_FRESHNESS_MS**

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -96,9 +97,9 @@
   private final int _numKeyColumns;
 
   // default message metadata
-  private static final StreamMessageMetadata _defaultMetadata = new StreamMessageMetadata();
-  private long _lastIndexedTimestamp = Long.MIN_VALUE;
-  private long _latestIngestionTimestamp = Long.MIN_VALUE;
+  private static final StreamMessageMetadata _defaultMetadata = new StreamMessageMetadata(System.currentTimeMillis());
+  private volatile long _lastIndexedTimestamp = Long.MIN_VALUE;
+  private volatile long _latestIngestionTimestamp = Long.MIN_VALUE;
```
Jackie-Jiang: (Optional) **_lastIndexedTimeMs** and **_latestIngestionTimeMs** to be more clear?

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -149,6 +151,33 @@ public DataTable processQuery(ServerQueryRequest queryRequest, ExecutorService e
       TraceContext.register(requestId);
     }
 
+    int numConsuming = 0;
+    long minIndexTime = Long.MAX_VALUE;
+    long minIngestionTime = Long.MAX_VALUE;
```
Jackie-Jiang: (Optional) **minIndexTimeMs** and **minIngestionTimeMs** to be more clear?

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -149,6 +151,33 @@ public DataTable processQuery(ServerQueryRequest queryRequest, ExecutorService e
       TraceContext.register(requestId);
     }
 
+    int numConsuming = 0;
```
Jackie-Jiang: (Optional) **numConsumingSegmentsQueried** to be more clear?

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -82,6 +82,8 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
     long numSegmentsQueried = 0L;
     long numSegmentsProcessed = 0L;
     long numSegmentsMatched = 0L;
+    long numConsumingQueried = 0L;
```
Jackie-Jiang: **numConsumingSegmentsQueried**

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -82,6 +82,8 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
...
+    long minConsumingFreshnessTs = Long.MAX_VALUE;
```
Jackie-Jiang: **minConsumingFreshnessTimeMs**

pinot-core/src/main/java/org/apache/pinot/core/query/executor/ServerQueryExecutorV1Impl.java
```diff
@@ -149,6 +151,33 @@ public DataTable processQuery(ServerQueryRequest queryRequest, ExecutorService e
...
+        if (ingestionTime != Long.MIN_VALUE && ingestionTime < minIngestionTime) {
+          minIngestionTime = ingestionTime;
+        }
+      }
+    }
+
+    if (numConsuming > 0) {
+      if (minIngestionTime == Long.MAX_VALUE) {
+        LOGGER.debug("Did not find valid ingestionTimestamp across consuming segments! Using indexTime instead");
+        minIngestionTime = minIndexTime;
```
Jackie-Jiang: (Optional) Add an extra variable **minConsumingFreshnessTimeMs** to be more clear?

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageMetadata.java
```diff
@@ -18,33 +18,24 @@
...
  */
-@InterfaceAudience.Public
-@InterfaceStability.Evolving
-public class StreamMessageMetadata {
+public class StreamMessageMetadata implements RowMetadata {
 
-  private long _ingestionTimestamp = Long.MIN_VALUE;
-
-  public StreamMessageMetadata() {
+  private final long _ingestionTimestamp;
```
Jackie-Jiang: **_ingestionTimeMs**?

pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
```diff
@@ -0,0 +1,38 @@
...
+ * Currently this is relevant for rows ingested into a mutable segment.
+ */
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
+public interface RowMetadata {
+
+  /**
+   * Return the ingestion timestamp of the row.
+   */
+  long getIngestionTimesMs();
```
Jackie-Jiang: **getIngestionTimeMs()** (remove the extra **s**)

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -168,6 +180,14 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
     brokerResponseNative.setNumSegmentsMatched(numSegmentsMatched);
     brokerResponseNative.setTotalDocs(numTotalRawDocs);
     brokerResponseNative.setNumGroupsLimitReached(numGroupsLimitReached);
+    if (numConsumingQueried > 0) {
+      if (minConsumingFreshnessTs == Long.MAX_VALUE) {
+        LOGGER.error("Invalid lastIndexedTimestamp across {} consuming segments", numConsumingQueried);
+        minConsumingFreshnessTs = 0L;
```
Jackie-Jiang: Do we want 0 or Long.MAX_VALUE as the invalid value?

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
Jackie-Jiang: @mcvsubbu Seems we already lost the requestId info in this layer. I think it should be fine because it should not happen, or any query can hit this error message.

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplTest.java
```diff
@@ -89,6 +93,11 @@ public void testMetadata() {
     SegmentMetadata expectedSegmentMetadata = _immutableSegment.getSegmentMetadata();
     Assert.assertEquals(actualSegmentMetadata.getTotalDocs(), expectedSegmentMetadata.getTotalDocs());
 
+    // assert that the last indexed timestamp is close to what we expect
+    long actualTs = _mutableSegmentImpl.getSegmentMetadata().getLastIndexedTimestamp();
```
Jackie-Jiang: Do you also want to test if last ingestion time matches the default metadata time?

pinot-tools/src/main/java/org/apache/pinot/tools/realtime/provisioning/MemoryEstimator.java
```diff
@@ -129,13 +129,11 @@ public File initializeStatsHistory() {
     // create mutable segment impl
     MutableSegmentImpl mutableSegmentImpl = new MutableSegmentImpl(realtimeSegmentConfigBuilder.build());
 
-    StreamMessageMetadata messageMetadata = new StreamMessageMetadata();
+    StreamMessageMetadata messageMetadata = new StreamMessageMetadata(System.currentTimeMillis());
```
Jackie-Jiang: We can simply pass in **null** here right?

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
```diff
@@ -212,17 +212,16 @@ public void run() {
...
-            messageMetadata.reset();
-            consumedRow = _streamLevelConsumer.next(consumedRow, messageMetadata);
+            consumedRow = _streamLevelConsumer.next(consumedRow);
 
             if (consumedRow != null) {
               GenericRow transformedRow = _recordTransformer.transform(consumedRow);
               if (transformedRow != null) {
-                notFull = realtimeSegment.index(transformedRow, messageMetadata);
+                // we currently do not get ingestion data through stream-consumer
+                notFull = realtimeSegment.index(transformedRow, new StreamMessageMetadata(System.currentTimeMillis()));
```
Jackie-Jiang: Pass in **null** as StreamMessageMetadata?

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegment.java
```diff
@@ -29,10 +29,10 @@
    * Indexes a record into the segment with optionally provided metadata.
    *
    * @param row Record represented as a {@link GenericRow}
-   * @param msgMetadata the metadata associated with the message
+   * @param rowMetadata the metadata associated with the message
    * @return Whether the segment is full (i.e. cannot index more record into it)
    */
-   boolean index(GenericRow row, StreamMessageMetadata msgMetadata);
+   boolean index(GenericRow row, RowMetadata rowMetadata);
```
Jackie-Jiang: Can we annotate the RowMetadata as **Nullable**? Also all the implementations

pinot-common/src/main/java/org/apache/pinot/common/utils/DataTable.java
```diff
@@ -35,6 +35,8 @@
   String NUM_SEGMENTS_QUERIED = "numSegmentsQueried";
   String NUM_SEGMENTS_PROCESSED = "numSegmentsProcessed";
   String NUM_SEGMENTS_MATCHED = "numSegmentsMatched";
+  String NUM_CONSUMING_SEGMENTS_QUERIED = "numConsumingSegmentsQueried";
+  String MIN_CONSUMING_FRESHNESS_MS = "minConsumingFreshnessMs";
```
Jackie-Jiang: MIN_CONSUMING_FRESHNESS_TIME_MS :-P

PR #3992, opened 2019-03-20
Fixing type casting issue for BYTES type values during realtime segment persistence

fx19880617:


pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/stats/RealtimeColumnStatistics.java
```diff
@@ -139,6 +140,11 @@ public boolean isSorted() {
       return false;
     }
 
+    // Values in BYTES are not comparable
+    if(_block.getMetadata().getDataType().getStoredType() == DataType.BYTES) {
```
mayankshriv: There's a wrapper around byte[] that is comparable: org.apache.pinot.common.utils.primitiveorg.apache.pinot.common.utils.primitive.ByteArray

xiangfu1: Update with ByteArray comparison.

kishoreg: +1 on using this wrapper comparable. Can we have byte[] column that is sortable?

PR #3993, opened 2019-03-20
In TableConfig, add checks for mandatory fields

Jackie-Jiang:
Add explicit checks for mandatory fields
Without the explicit checks, it will throw NPE, which is not clear and hard to debug

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
@@ -27,7 +27,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.apache.helix.ZNRecord;
 import org.apache.pinot.common.data.StarTreeIndexSpec;
@@ -50,6 +49,8 @@ public class TableConfig {
   public static final String QUOTA_CONFIG_KEY = "quota";
   public static final String TASK_CONFIG_KEY = "task";
   public static final String ROUTING_CONFIG_KEY = "routing";
+
+  private static final String FIELD_MISSING_MESSAGE_TEMPLATE = "Mandatory field '%s' is missing";
 
   @ConfigKey("name")
   @ConfigDoc(value = "The name for the table.", mandatory = true, exampleValue = "myTable")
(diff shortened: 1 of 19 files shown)


pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
```diff
@@ -118,14 +118,24 @@ public static TableConfig fromJsonString(String jsonString)
...
     String tableName = TableNameBuilder.forType(tableType).tableNameWithType(jsonConfig.get(TABLE_NAME_KEY).asText());
-
+    Preconditions
+        .checkState(jsonConfig.has(VALIDATION_CONFIG_KEY), "Mandatory config '%s' is missing", VALIDATION_CONFIG_KEY);
     SegmentsValidationAndRetentionConfig validationConfig =
         extractChildConfig(jsonConfig, VALIDATION_CONFIG_KEY, SegmentsValidationAndRetentionConfig.class);
+    Preconditions.checkState(jsonConfig.has(TENANT_CONFIG_KEY), "Mandatory config '%s' is missing", TENANT_CONFIG_KEY);
     TenantConfig tenantConfig = extractChildConfig(jsonConfig, TENANT_CONFIG_KEY, TenantConfig.class);
+    Preconditions
+        .checkState(jsonConfig.has(INDEXING_CONFIG_KEY), "Mandatory config '%s' is missing", INDEXING_CONFIG_KEY);
```
sunithabeeram: Are INDEXING and CUSTOM configs really mandatory?

Jackie-Jiang: Currently they are mandatory, and all the callers assume they are non-null.

pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
```diff
@@ -118,14 +118,24 @@ public static TableConfig fromJsonString(String jsonString)
...
   public static TableConfig fromJSONConfig(@Nonnull JsonNode jsonConfig)
       throws IOException {
+    // Mandatory fields
+    Preconditions.checkState(jsonConfig.has(TABLE_TYPE_KEY), "Table type is missing");
     TableType tableType = TableType.valueOf(jsonConfig.get(TABLE_TYPE_KEY).asText().toUpperCase());
+    Preconditions.checkState(jsonConfig.has(TABLE_NAME_KEY), "Table name is missing");
     String tableName = TableNameBuilder.forType(tableType).tableNameWithType(jsonConfig.get(TABLE_NAME_KEY).asText());
-
+    Preconditions
+        .checkState(jsonConfig.has(VALIDATION_CONFIG_KEY), "Mandatory config '%s' is missing", VALIDATION_CONFIG_KEY);
```
jackjlli: Since `"Mandatory config '%s' is missing"` is commonly used here, can we put it as a final static variable in the same class?

Jackie-Jiang: Done

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/OfflineClusterIntegrationTest.java
```diff
@@ -150,6 +152,22 @@ public void testInstancesStarted() {
...
+        new TableConfig.Builder(CommonConstants.Helix.TableType.OFFLINE).setTableName("badTable").build();
+    ObjectNode jsonConfig = tableConfig.toJsonConfig();
+    // Remove a mandatory field
+    jsonConfig.remove(TableConfig.VALIDATION_CONFIG_KEY);
+    try {
+      sendPostRequest(_controllerRequestURLBuilder.forTableCreate(), jsonConfig.toString());
+      fail();
+    } catch (IOException e) {
+      // Should get response code 400 (BAD_REQUEST)
+      assertTrue(e.getMessage().startsWith("Server returned HTTP response code: 400"));
```
sunithabeeram: Can we assert on the rest of the message? (If we don't have access to the message at this level, ignore)

The reason is, currently the client does get a 400, but the message is "null" which makes it hard to debug.

Jackie-Jiang: I understand your concern, but on this level we don't have access to the message. I added checks on the message in the unit test.

PR #3994, opened 2019-03-20
Adding ORC reader

jenniferdai:
* Performance generating the same segment with ORC and Avro
ORC: 3 mins 41 seconds
Avro: 3 mins 35 seconds
* Accuracy is verified, checked the metadata.properties of both and they were identical
* Next steps - add multivalue columns and test/add byte arrays


pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.orc.mapred.OrcMapredRecordReader;
+import org.apache.pinot.common.data.Schema;
+import org.apache.pinot.core.data.GenericRow;
+import org.apache.pinot.core.data.readers.RecordReader;
+import org.apache.pinot.core.indexsegment.generator.SegmentGeneratorConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class ORCRecordReader implements RecordReader {
```
sunithabeeram: Add class level java docs.

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -0,0 +1,187 @@
...
+      throws IOException {
+    _recordReader.nextBatch(_reusableVectorizedRowBatch);
+    fillGenericRow(reuse, _reusableVectorizedRowBatch);
+    return reuse;
+  }
+
+  private void fillGenericRow(GenericRow genericRow, VectorizedRowBatch rowBatch) throws IOException {
+    // Read the row data
+    TypeDescription schema = _reader.getSchema();
+    // Create a row batch with max size 1
```
sunithabeeram: Move this comment to line 81 (where you are allocating the row-batch)

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -0,0 +1,187 @@
...
+
+    if (schema.getCategory().equals(TypeDescription.Category.STRUCT)) {
+      for (int i = 0; i < schema.getChildren().size(); i++) {
+        // Get current column in schema
+        TypeDescription currColumn = schema.getChildren().get(i);
+        String currColumnName = currColumn.getFieldNames().get(0);
+        int currColRowIndex = currColumn.getId();
+        ColumnVector vector = rowBatch.cols[currColRowIndex];
+        // Previous value set to null, not used except to save allocation memory in OrcMapredRecordReader
+        WritableComparable writableComparable = OrcMapredRecordReader.nextValue(vector, currColRowIndex, _orcSchema, null);
```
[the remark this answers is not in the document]
jenniferdai: good catch, yes we do that for other record readers, added

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -58,6 +67,24 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ORCRecordReader.class);
 
+  @VisibleForTesting
+  public ORCRecordReader(String inputPath) {
+    Configuration conf = new Configuration();
+    try {
```
sunithabeeram: Add a init() that takes a path instead of segment-generator config (it can be a private method). Have this constructor and the public init(SegmentGeneratorConfig) use that.

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -0,0 +1,218 @@
...
+  @Override
+  public GenericRow next(GenericRow reuse)
+      throws IOException {
+    _recordReader.nextBatch(_reusableVectorizedRowBatch);
+    fillGenericRow(reuse, _reusableVectorizedRowBatch);
+    return reuse;
+  }
+
+  private void fillGenericRow(GenericRow genericRow, VectorizedRowBatch rowBatch) throws IOException {
+    // ORC's TypeDescription is the equivalent of a schena. The way we will support ORC in Pinot
```
sunithabeeram: schena -> schema

jenniferdai:
I don't believe we can use that particular next method.

That next method keeps the row batch (of unknown number) within it, which means that we can't tell whether there is a next row. We can only tell if there is a next batch. Additionally and more importantly, records come back in random order. Problems arise for sorted columns, other specialized pinot features that depend on the order of record to be consistent.

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -0,0 +1,207 @@
...
+        TypeDescription currColumn = _orcSchema.getChildren().get(i);
+        String currColumnName = _orcSchema.getFieldNames().get(i);
+        if (!_pinotSchema.getColumnNames().contains(currColumnName)) {
+          LOGGER.warn("Skipping column {} because it is not in pinot schema", currColumnName);
+          continue;
+        }
+        int currColRowIndex = currColumn.getId();
+        // Struct is top level, so the id of the struct is 0. However, the children start from 1+, etc, so we need to
+        // subtract one since the row batch we get has only children column vectors
+        ColumnVector vector = rowBatch.cols[currColRowIndex - 1];
```
wagnermarkd: You don't need to use `getId()` for indexing into the returned batch. The ids are an internal detail of ORC and reference individual streams in the serialized file. The VectorizedRowBatch that's returned to you will have exactly the fields that you specified when you set the schema in your ReaderOptions (or the file schema if you don't set it) and they will be in the same order as that schema. Therefore, you can just use `rowBatch.cols[i]` to get what you want.

wagnermarkd: That `currColRowIndex - 1` happens to work is only because you've been using flat schemas.

jenniferdai: Thank you so much! Changed

antumbde:
I suggest we highlight difference in the semantics of array data types for users of this tool. This is important if a user wants to run a query in Presto on ORC dataset generated using tool. Arrays are treated differently by traditional ORC/DB users from Pinot.

antumbde:
Adding a note about what is different (based on memory; please double check)

Arrays in Pinot are like collapsed set of rows.  If a table has array column, acol, with value [1,2,3] then that row will be selected by predicate condition 'where acol = 3'.  This is different from Presto (and other DBs). For same result, users have to write their query as 'where 3 in acol'. Only value  "[1,2,3]" will match a row "[1,2,3]" in presto.

PR #4003, opened 2019-03-21
[TE] Fix ArrayIndexOutOfBoundsException; Wrap replay endpoint and ret…

akshayrai:
…urn meaningful message

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/PercentageChangeRuleAnomalyFilter.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/PercentageChangeRuleAnomalyFilter.java
@@ -70,15 +70,25 @@ public class PercentageChangeRuleAnomalyFilter implements AnomalyFilter<Percenta
 
     Map<MetricSlice, DataFrame> aggregates =
         this.dataFetcher.fetchData(new InputDataSpec().withAggregateSlices(slices)).getAggregates();
-    if (aggregates.get(currentSlice).isEmpty() || aggregates.get(baselineSlice).isEmpty()) {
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/PercentageChangeRuleAnomalyFilter.java
```diff
@@ -67,6 +70,11 @@ public boolean isQualified(MergedAnomalyResultDTO anomaly) {
 
     Map<MetricSlice, DataFrame> aggregates =
         this.dataFetcher.fetchData(new InputDataSpec().withAggregateSlices(slices)).getAggregates();
+    if (aggregates.get(currentSlice).isEmpty() || aggregates.get(baselineSlice).isEmpty()) {
+      LOG.warn("Unable to fetch data for current or baseline slice for anomaly {}. start = {} end = {} filters = {}",
+          anomaly.getId(), anomaly.getStartTime(), anomaly.getEndTime(), me.getFilters());
+      return false;
```
jihaozh: should it return true? otherwise, the anomaly is suppressed.

akshayrai: Ideally, we shouldn't get this error. If we do get it, then displaying the anomaly can mislead the users. Returning true would mean that the anomaly passed the filter when in fact it didn't, right?

jihaozh: right, now I think of it, in this case, should we use the value stored in the anomaly instead?

PR #4004, opened 2019-03-21
[TE] frontend - harleyjj/yaml-editor - increases toast timeout from 4…

harleyjj:
… seconds to 10 seconds


thirdeye/thirdeye-frontend/app/pods/components/alert-details/component.js
```diff
@@ -496,7 +496,9 @@ export default Component.extend({
       isPreviewMode,
       alertId
     } = this.getProperties('analysisRange', 'notifications', 'isPreviewMode', 'alertId');
-
+    const toastOptions = {
+      timeOut: 10000
```
akshayrai: Would it be possible to define this in a single place?

harleyjj: definitely.  I'll add it to an appropriate utils file

harleyjj: Done

PR #4009, opened 2019-03-22
Adding ORC Multi-value Column Support

jenniferdai:


pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -186,6 +192,20 @@ private Object getBaseObject(WritableComparable w) {
     return obj;
   }
 
+  private List<Object> translateList(OrcList<? extends WritableComparable> l) {
+    if (l == null || l.size() < 1) {
```
sunithabeeram: If an isEmpty() method is exposed, it might be better to use that instead of size() < 1.

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -83,6 +87,8 @@ private void init(String inputPath, Schema schema) {
         LOGGER.warn("Pinot schema is not set in segment generator config");
       }
       _recordReader = _reader.rows(_reader.options().schema(_orcSchema));
+
+      _orcMapredRecordReader = new OrcMapredRecordReader<>(_recordReader, _orcSchema);
```
sunithabeeram: It doesn't look like this is used.

pinot-orc/src/test/java/org/apache/pinot/orc/data/readers/ORCRecordReaderTest.java
```diff
@@ -77,12 +83,33 @@ public void setUp()
...
+    struct.setFieldValue("emails", emails);
+
+    Writer mvWriter = OrcFile.createWriter(new Path(MULTIVALUE_ORC_FILE.getAbsolutePath()),
+        OrcFile.writerOptions(new Configuration())
+            .setSchema(orcTypeDesc));
+
+    OrcMapredRecordWriter mrRecordWriter = new OrcMapredRecordWriter(mvWriter);
+    mrRecordWriter.write(null, struct);
+    mrRecordWriter.close(null);
+    System.out.println();
```
sunithabeeram: Remove.

pinot-orc/src/test/java/org/apache/pinot/orc/data/readers/ORCRecordReaderTest.java
```diff
@@ -47,11 +51,13 @@
 public class ORCRecordReaderTest {
   private static final File TEMP_DIR = FileUtils.getTempDirectory();
   private static final File ORC_FILE = new File(TEMP_DIR.getAbsolutePath(), "my-file.orc");
+  private static final File MULTIVALUE_ORC_FILE = new File(TEMP_DIR.getAbsolutePath(), "mv-my-file.orc");
```
sunithabeeram: Can you include this to be part of the same table/file as the other values?

sunithabeeram: Also, would be good to add an explicit test for bytes.

sunithabeeram:
Can you please check on travis failure? Create a flaky test tracker/issue if needed.

pinot-orc/src/test/java/org/apache/pinot/orc/data/readers/ORCRecordReaderTest.java
jenniferdai: I believe I added bytes if this is what you are referring to: 
    FieldSpec yFieldSpec = new DimensionFieldSpec("y", FieldSpec.DataType.BYTES, true);

jenniferdai: I have not been able to set the ListColumnVector correctly, so I am using builder APIs to do so. While I can put my other values in my MV test, no apis are exposed to add multiple rows, so I think I should keep them separate.

sunithabeeram: ok; we can check on options for that. For now, please do add more than one row in the MV test.

PR #4010, opened 2019-03-23
Add docs for record reader

Jackie-Jiang:


docs/record_reader.rst
```diff
@@ -0,0 +1,133 @@
...
+Pinot package provides the following record readers out of the box:
+
+- Avro record reader: record reader for Avro format files
+- CSV record reader: record reader for CSV format files
+- JSON record reader: record reader for JSON format files
+- Thrift record reader: record reader for Thrift format files
+- Pinot segment record reader: record reader for Pinot segment
+
+For other file formats, we provide a general interface for record reader - `RecordReader`. To index the file into Pinot
+segment, implement the interface and plug it into the index engine - `SegmentCreationDriverImpl`.
```
sunithabeeram: Provide a pointer/example to ORCReader that was recently added.

docs/record_reader.rst
```diff
@@ -0,0 +1,133 @@
...
+Contracts for Record Reader
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are several contracts for record readers that developers should follow when implementing their own record readers:
+
+- The output GenericRow should follow the table schema provided, in the sense that:
+
+  - All the columns in the schema should be preserved (if column does not exist in the original record, put default
+    value instead)
+  - Columns not in the schema should be escaped
```
sunithabeeram: escaped? Is removed (not included) a better choice?

docs/record_reader.rst
```diff
@@ -0,0 +1,133 @@
...
+  - Values for the column should follow the field spec from the schema (data type, single-valued/multi-valued)
+
+- For the time column, record reader should be able to read both incoming and outgoing time:
```
sunithabeeram: It might be better to explain what the incoming/outgoing specs are for (or link to the docs that explain it)

docs/record_reader.rst
```diff
@@ -0,0 +1,133 @@
...
+Thrift Record Reader Config
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The Thrift record reader config is mandatory. It contains the Thrift class name for the record reader to de-serialize
+the Thrift objects.
+
+Implement Your Own Record Reader
+--------------------------------
+
+You can implement your own record reader for file formats that is not supported natively. The following methods need to
```
sunithabeeram: I would prefer if we just point to the record reader interface in the code; else we'll have to update this doc everytime there is *any* change to the interface.

- I would put explanations for GenericRow etc before pointing to the interface.
- It will also be good to high light that we use a 2 pass algorithm, hence "rewind" is necessary. This will inform the readers about additional restrictions on the interface.

sunithabeeram:
Given this is a doc change, travis shouldn't have failed. Can you please follow the failure and create a Flaky test issue/tracker if required?

docs/record_reader.rst
Jackie-Jiang: Done

docs/record_reader.rst
Jackie-Jiang: Done

docs/record_reader.rst
Jackie-Jiang: Done

docs/record_reader.rst
Jackie-Jiang: Done

Jackie-Jiang:
@sunithabeeram Addressed the comments. The travis failure is from a timeout which is caused by too less CPU cores running the tests.

docs/record_reader.rst
```diff
@@ -0,0 +1,104 @@
...
+
+The Thrift record reader config is mandatory. It contains the Thrift class name for the record reader to de-serialize
+the Thrift objects.
+
+Implement Your Own Record Reader
+--------------------------------
+
+For other file formats, we provide a general interface for record reader - `RecordReader <https://github.com/apache/incubator-pinot/blob/master/pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReader.java>`_.
+To index the file into Pinot segment, simply implement the interface and plug it into the index engine - `SegmentCreationDriverImpl <https://github.com/apache/incubator-pinot/blob/master/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentIndexCreationDriverImpl.java>`_.
+We use a 2-passes algorithm to index the file into Pinot segment, hence the *rewind()* method is required for the record
```
sunithabeeram: 2-passes => 2-pass. 

Nit: Might be better to call out that the reader should be able to read the data twice (for e.g., relying on a streaming reader would not work).

docs/record_reader.rst
```diff
@@ -0,0 +1,104 @@
...
+- The output GenericRow should follow the table schema provided, in the sense that:
+
+  - All the columns in the schema should be preserved (if column does not exist in the original record, put default
+    value instead)
+  - Columns not in the schema should not be included
+  - Values for the column should follow the field spec from the schema (data type, single-valued/multi-valued)
+
+- For the time column (refer to `TimeFieldSpec <https://github.com/apache/incubator-pinot/blob/master/pinot-common/src/main/java/org/apache/pinot/common/data/TimeFieldSpec.java>`_),
+  record reader should be able to read both incoming and outgoing time (we allow *incoming time - time value from the
+  original data* to *outgoing time - time value stored in Pinot* conversion during index creation).
```
sunithabeeram: Might be better to phrase it as "incoming-time i.e., time value from the original data to outgoing-time i.e., time value stored in Pinot"

PR #4011, opened 2019-03-24
Added a percent threshold to consider startup of services

mcvsubbu:
In installations where each server may host 1000s of tabls, and 10s of
thousands of segments, it is useful in practice to declare a server
as STARTED when "most" segments of "most" of the tables are ready (i.e.
externalview and idealstate match, as do currentstate and idealstate).

This PR makes an attempt at defining "most" (on an experimental basis).

For now, we disregard the number of (helix) partitions within each resource.
Instead, we count the number of resources that have converged state
and declare success if a (configured) percent are ready.

This is still experimental, and it is suggested to set the percent
to 100 all the time (or, not set it at all, and let the default ride).

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
@@ -137,7 +137,8 @@ public class ServiceStatus {
     private final Set<String> _resourcesToMonitor;
     private final int _numTotalResourcesToMonitor;
     private Iterator<String> _resourceIterator = null;
-    private final double _minResourcesStartPercent;
+    // Minimum number of resources to be in converged state before we declare the service state as STARTED
+    private final int _minResourcesStartCount;
 
     private String _statusDescription = STATUS_DESCRIPTION_INIT;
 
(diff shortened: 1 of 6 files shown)


pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -243,11 +278,18 @@ public synchronized Status getServiceStatus() {
         }
 
         // Resource is done starting up, remove it from the set
-        iterator.remove();
```
sunithabeeram: There is a return in line 275 - should that be a continue (as we want to track the % of resources online and not get stuck on just one, right?)

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -189,20 +194,50 @@ public IdealStateMatchServiceStatusCallback(HelixManager helixManager, String cl
 
     protected abstract String getMatchName();
 
+    private boolean isDone() {
+      if (_resourcesToMonitor.isEmpty() || _numTotalResourcesToMonitor <= 0) {
+        return true;
+      }
+      if ((double)(_numTotalResourcesToMonitor - _resourcesToMonitor.size())*100/_numTotalResourcesToMonitor
```
sunithabeeram: We can estimate the number of resources we need to be "done" _once_ and compare against it.

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
mcvsubbu: Nope. the return should be as it is, because we want to stop when we reach a bad resource. We already check the threshold at the top of the loop, so if we are inside the loop the threshold is not met yet. We check only one resource each time in the loop

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
mcvsubbu: done

PR #4025, opened 2019-03-28
Extend stream interfaces to expose message metadata like ingestion timestamp

sunithabeeram:
This PR provides the basic structure to expose freshness metrics as captured in https://github.com/apache/incubator-pinot/issues/4007. The stream interfaces are extended to provide an optional StreamMessageMetadata - this class currently exposes the ingestion timestamp of the message and can be extended to add other relevant metadata as needed. The ingestion timestamp is intended to reflect the time that the message was ingested into a streaming service such as Kafka and is used to track the freshness/lag of realtime message consumption in Pinot.

*Backward compatibility considerations*
The newly added methods to the interfaces are default methods and call the existing methods for backward compatibility. Stream extensions that would like to take advantage of freshness tracking, should override the default methods. However, just upgrading to the new version will not result in compilation failures.

*Testing done*
None. Once this change is reviewed, I will be updating https://github.com/apache/incubator-pinot/pull/3979 to expose both ingestion and indexed timestamps in a meaningful way and update the tests there.


pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegment.java
```diff
@@ -32,10 +33,40 @@
...
   boolean index(GenericRow row);
 
+  /**
+   * Indexes a record into the segment with optionally provided metadata.
+   *
+   * @param row Record represented as a {@link GenericRow}
+   * @param msgMetadata the metadata associated with the message
+   * @return Whether the segment is full (i.e. cannot index more record into it)
+   */
+  default boolean index(GenericRow row, StreamMessageMetadata msgMetadata) {
```
mcvsubbu: This interface is not meant to be used by any Stream implementation (or other injected implementations). We can go ahead and modify the index() method in line 34.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegment.java
```diff
@@ -32,10 +33,40 @@
...
    * @return The number of records indexed
    */
   int getNumDocsIndexed();
+
+  /**
+   * Returns the latest indexed timestamp for the segment.
+   *
+   * @return the timestamp of last message indexed
+   */
+  default Long getLastIndexedTimestamp() {
```
mcvsubbu: same

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegment.java
```diff
@@ -32,10 +33,40 @@
...
+    return null;
+  }
+
+  /**
+   * Returns the ingestion timestamp for the latest message in the segment.
+   *
+   * @return the ingestion timestamp of last message indexed
+   */
+  default Long getLatestIngestionTimestamp() {
```
mcvsubbu: same

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -194,8 +200,14 @@ public long getMaxTime() {
     return _maxTime;
   }
 
+
   @Override
   public boolean index(GenericRow row) {
+    return index(row, _defaultMetadata);
```
mcvsubbu: this method should not be needed

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -210,6 +222,11 @@ public boolean index(GenericRow row) {
       // Add forward and inverted indices for new document.
       addForwardIndex(row, docId, dictIdMap);
       addInvertedIndex(docId, dictIdMap);
+
+      _lastIndexedTimestamp = System.currentTimeMillis();
```
mcvsubbu: You also need to update it before the return in line 236 since we are aggregating the metrics there. Not sure if aggregateMetrics() silently skips aggregation, please verify.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageDecoder.java
```diff
@@ -54,4 +54,40 @@ void init(Map<String, String> props, Schema indexingSchema, String topicName)
    * @return A new row decoded from the buffer
    */
   GenericRow decode(T payload, int offset, int length, GenericRow destination);
+
+  /**
+   * Decodes the payload received into a generic row and optionally returns any metadata associated
```
mcvsubbu: suggest s/returns/populates/

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageDecoder.java
```diff
@@ -54,4 +54,40 @@ void init(Map<String, String> props, Schema indexingSchema, String topicName)
...
+   * with the message.
+   *
+   * This is an interface default method is provided for backward compatibility and should be
+   * overridden by implementations that can provide meaningful metadata.
```
mcvsubbu: add a ref here to see the StreamMessageMetadata class

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageDecoder.java
```diff
@@ -54,4 +54,40 @@ void init(Map<String, String> props, Schema indexingSchema, String topicName)
...
+   * @param metadata The {@link StreamMessageMetadata} associated with the message
+   * @return A new row decoded from the buffer
+   */
+  default GenericRow decode(T payload, GenericRow destination, StreamMessageMetadata metadata) {
+    // by default, invoke the non-metadata based method for backwards compatibility
+    return decode(payload, destination);
+  }
+
+  /**
+   * Decodes the payload received into a generic row and optionally returns any metadata associated
```
mcvsubbu: same

kishoreg:
Why do we have to change the interface of IndexSegment to achieve this? This is creating a circular dependency between the base core and real-time. Conceptually, we have a spi layer (io and segment interfaces) and an offline and real-time module. This change means the base layer depends on an implementation (realtime) layer.

Is there no other option to achieve this without making changes to the interface?

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageMetadata.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package org.apache.pinot.core.realtime.stream;
+
+/**
+ * A class that provides metadata associated with the message of a stream, for e.g.,
+ * ingestion-timestamp of the message.
+ */
+public class StreamMessageMetadata {
+
+  private Long _ingestionTimestamp;
```
mcvsubbu: Any reason why we are not using primitives here? Consider high ingestion use cases, a single server serving multiple partitions. Will it generate more unnecessary garbage? We could use -1 for an invalid value?

mcvsubbu:
> Why do we have to change the interface of IndexSegment to achieve this? This is creating a circular dependency between the base core and real-time. Conceptually, we have a spi layer (io and segment interfaces) and an offline and real-time module. This change means the base layer depends on an implementation (realtime) layer.
> 
> Is there no other option to achieve this without making changes to the interface?

@kishoreg  I don;t see a circular dependency. Perhaps I am missing something? The new class is in the generic streams area which is where all the other generic streams classes are.

kishoreg:
@mcvsubbu it’s about changing the indexsegment interface which now depends on stream package and stream package depends on core Segment package

sunithabeeram:
>@mcvsubbu it’s about changing the indexsegment interface which now depends on stream package and stream package depends on core Segment package

@kishoreg, @mcvsubbu : I can update the index() method to just take in the values we need (in this case ingestion timestamp, instead of stream-metadata. I had it at that object level to allow for easy future extensions, but we can always pass in specific parameters.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageMetadata.java
sunithabeeram: Sure, I think I can use -1 to indicate the same that I was going for with null.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegment.java
sunithabeeram: Agreed. I was being defensive. 

I think we should annotate the interfaces as "public" and "internal" so when we are updating or reviewing updates we know when to look for backward compatibility.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/MessageBatch.java
```diff
@@ -51,6 +51,17 @@
...
   int getMessageLengthAtIndex(int index);
 
+  /**
+   * Returns the timestamp associated with the message at a particular index. This is typicall the timestamp
+   * when the message was ingested by the upstream stream-provider.
+   *
+   * @return the timestamp associated with the message
+   *         Long.MIN_VALUE if its not supported by the stream
+   */
+  default long getTimestampAtIndex(int index) {
```
mcvsubbu: Suggest change it to getMetadataAtIndex()

pinot-common/src/main/java/org/apache/pinot/common/segment/SegmentMetadata.java
```diff
@@ -73,6 +73,23 @@
    */
   long getRefreshTime();
 
+  // methods specific to realtime segments
```
mcvsubbu: I would prefer adding this comment to each of the comment block below. Otherwise, it seems as if the comment is valid for all methods below this line.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaStreamLevelConsumer.java
```diff
@@ -83,9 +84,14 @@ public void start()
 
   @Override
   public GenericRow next(GenericRow destination) {
+    return next(destination, null);
```
mcvsubbu: should we mark this method deprecated in the interface?

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageDecoder.java
```diff
@@ -58,4 +58,40 @@ void init(Map<String, String> props, Schema indexingSchema, String topicName)
    * @return A new row decoded from the buffer
    */
   GenericRow decode(T payload, int offset, int length, GenericRow destination);
+
```
mcvsubbu: should we deprecate the methods in lines 53 and  60?

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageMetadata.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * A class that provides metadata associated with the message of a stream, for e.g.,
+ * ingestion-timestamp of the message.
+ */
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
+public class StreamMessageMetadata {
+
+  private long _ingestionTimestamp;
```
mcvsubbu: default to Long.MIN_VALUE?

pinot-common/src/main/java/org/apache/pinot/common/segment/SegmentMetadata.java
sunithabeeram: Removed the comment as each method already indicates the same.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageDecoder.java
sunithabeeram: I will add deprecations in a followup as it needs to be done across the interfaces we have touched here.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageMetadata.java
sunithabeeram: Updated.

sunithabeeram:
@mayankshriv: 
1. The interface modification has been kept generic enough to allow other metadata to be included from stream->mutable segment.
2. #3979  and other changes will wire this up. The specific classes that implement the MessageBatch or MessageDecoder will eventually provide the values needed.
3. We definitely considered this option - but accessing the information through segment metadata is a lot more cleaner than other options. I have an updated to the other PR - will push it now. You can review that.

PR #4032, opened 2019-03-29
Fail pipeline all windows failed

xiaohui-sun:
1. Fail pipeline if all detection windows failed.
2. Set default cache warm up to 0 to fix the epoch time granularity issue.
3. Add more logs for merger if the anomalyCountAfterSplit exceeds NUMBER_OF_SPLITED_ANOMALIES_LIMIT.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -73,7 +75,7 @@
   private static final String PROP_TIMEZONE = "timezone";
   private static final String PROP_BUCKET_PERIOD = "bucketPeriod";
   private static final String PROP_CACHE_PERIOD_LOOKBACK = "cachingPeriodLookback";
-  private static final long DEFAULT_CACHING_PERIOD_LOOKBACK = TimeUnit.DAYS.toMillis(30);
+  private static final long DEFAULT_CACHING_PERIOD_LOOKBACK = TimeUnit.DAYS.toMillis(0);
```
jihaozh: Use -1 if we would like to disable cache warm up by default?

xiaohui-sun: There is no difference between 0 and -1.

PR #4037, opened 2019-03-29
Add more logs to ServiceStatus

mcvsubbu:
When we declare a service good due to its reaching the percentage threshold, it is useful to
log information about at least one partition in the resources that are not started up
correctly.

Also included a max number of resources that we log so that we don't get extra long log lines

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
@@ -24,6 +24,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.apache.commons.math3.util.Pair;
 import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixManager;
@@ -50,6 +51,8 @@ public class ServiceStatus {
 
   public static String STATUS_DESCRIPTION_NONE = "None";
   public static String STATUS_DESCRIPTION_INIT = "Init";
+
+  private static int MAX_RESOURCE_NAMES_TO_LOG = 5;
 
   /**
    * Callback that returns the status of the service.
@@ -171,7 +174,7 @@ public class ServiceStatus {
       _minResourcesStartCount = (int)Math.round(Math.ceil(minResourcesStartPercent * _numTotalResourcesToMonitor/100));
 
       LOGGER.info("Monitoring {} resources: {} for start up of instance {}", _numTotalResourcesToMonitor,
-          _resourcesToMonitor, _instanceName);
+          getResourceListAsString(), _instanceName);
     }
 
     public IdealStateMatchServiceStatusCallback(HelixManager helixManager, String clusterName, String instanceName,
@@ -186,7 +189,7 @@ public class ServiceStatus {
 
       _minResourcesStartCount = (int)Math.round(Math.ceil(minResourcesStartPercent * _numTotalResourcesToMonitor/100));
       LOGGER.info("Monitoring {} resources: {} for start up of instance {}", _numTotalResourcesToMonitor,
-          _resourcesToMonitor, _instanceName);
+          getResourceListAsString(), _instanceName);
     }
 
     protected abstract T getState(String resourceName);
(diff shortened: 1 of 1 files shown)


pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -227,70 +230,98 @@ private boolean isDone() {
...
-        IdealState idealState = getResourceIdealState(resourceName);
+        descriptionSuffix = String
+            .format("waitingFor=%s, resource=%s, numResourcesLeft=%d, numTotalResources=%d, minStartCount=%d,", getMatchName(),
+                resourceName, _resourcesToMonitor.size(), _numTotalResourcesToMonitor, _minResourcesStartCount);
 
-        // If the resource has been removed or disabled, ignore it
-        if (idealState == null || !idealState.isEnabled()) {
+        Pair<Status, String> statusPair = evaluateResourceStatus(resourceName, descriptionSuffix);
+
+        // Resource is done starting up, remove it from the set
```
sunithabeeram: This should move below 249?

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -227,70 +230,98 @@ private boolean isDone() {
...
-        _statusDescription = String.format("waitingFor=%s, numResourcesLeft=%d, numTotalResources=%d, minStartCount=%d",
-            getMatchName(), _resourcesToMonitor.size(), _numTotalResourcesToMonitor, _minResourcesStartCount);
-        LOGGER.info("Instance {} returning GOOD because {}", _statusDescription);
+    private String getResourceListAsString() {
+      if (_resourcesToMonitor.size() <= MAX_RESOURCE_NAMES_TO_LOG) {
+        return _resourcesToMonitor.toString();
       }
-
-      return Status.GOOD;
+      return "[" + _resourcesToMonitor.iterator().next() + ",...]";
```
sunithabeeram: I am not clear about this method - why is this moving the resource iterator? It sounds its mostly a toString that restricts it to less than 5 resources logged...

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -227,70 +230,98 @@ private boolean isDone() {
...
+        descriptionSuffix = String
+            .format("waitingFor=%s, resource=%s, numResourcesLeft=%d, numTotalResources=%d, minStartCount=%d,", getMatchName(),
+                resourceName, _resourcesToMonitor.size(), _numTotalResourcesToMonitor, _minResourcesStartCount);
 
-        // If the resource has been removed or disabled, ignore it
-        if (idealState == null || !idealState.isEnabled()) {
+        Pair<Status, String> statusPair = evaluateResourceStatus(resourceName, descriptionSuffix);
+
+        // Resource is done starting up, remove it from the set
+        _statusDescription = statusPair.getSecond();
```
sunithabeeram: Is this being used? I see it getting reassinged in 257 and 260...

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -227,70 +230,98 @@ private boolean isDone() {
...
         }
         resourceName = _resourceIterator.next();
-        IdealState idealState = getResourceIdealState(resourceName);
+        descriptionSuffix = String
+            .format("waitingFor=%s, resource=%s, numResourcesLeft=%d, numTotalResources=%d, minStartCount=%d,", getMatchName(),
+                resourceName, _resourcesToMonitor.size(), _numTotalResourcesToMonitor, _minResourcesStartCount);
 
-        // If the resource has been removed or disabled, ignore it
-        if (idealState == null || !idealState.isEnabled()) {
+        Pair<Status, String> statusPair = evaluateResourceStatus(resourceName, descriptionSuffix);
```
sunithabeeram: <nit> might be better to append the suffix after the method is called instead of passing the suffix to it; just seems cleaner </nit>

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -24,6 +24,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.apache.commons.math3.util.Pair;
```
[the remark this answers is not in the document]
mcvsubbu: I thought about that, but didn't want to make this PR bigger. I don't see anything wrong with apache commons though. Do you have a concern?

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
mcvsubbu: It prints the name of one table the first one and then ...
If the number of tavles are small, it just returns the entire list in string form.

PR #4041, opened 2019-03-30
Use hex string as the representation of byte array for queries

fx19880617:
Use hex string to represent byte[] value in query.
Refer to https://github.com/apache/incubator-pinot/issues/4040

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOffHeapMutableDictionary.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOffHeapMutableDictionary.java
@@ -21,6 +21,9 @@ package org.apache.pinot.core.realtime.impl.dictionary;
 import java.io.IOException;
 import java.util.Arrays;
 import javax.annotation.Nonnull;
+import org.apache.commons.codec.DecoderException;
+import org.apache.commons.codec.binary.Hex;
+import org.apache.pinot.common.Utils;
 import org.apache.pinot.common.utils.primitive.ByteArray;
 import org.apache.pinot.core.io.readerwriter.PinotDataBufferMemoryManager;
 import org.apache.pinot.core.io.writer.impl.MutableOffHeapByteArrayStore;
@@ -30,6 +33,7 @@ import org.apache.pinot.core.io.writer.impl.MutableOffHeapByteArrayStore;
  * OffHeap mutable dictionary for Bytes data type.
  */
 public class BytesOffHeapMutableDictionary extends BaseOffHeapMutableDictionary {
+
   private final MutableOffHeapByteArrayStore _byteStore;
 
   private ByteArray _min = null;
@@ -47,13 +51,24 @@ public class BytesOffHeapMutableDictionary extends BaseOffHeapMutableDictionary 
   public BytesOffHeapMutableDictionary(int estimatedCardinality, int maxOverflowHashSize,
       PinotDataBufferMemoryManager memoryManager, String allocationContext, int avgLength) {
     super(estimatedCardinality, maxOverflowHashSize, memoryManager, allocationContext);
-    _byteStore = new MutableOffHeapByteArrayStore(memoryManager, allocationContext, estimatedCardinality, avgLength);
+    _byteStore = new MutableOffHeapByteArrayStore(memoryManager, allocationContext,
+        estimatedCardinality, avgLength);
   }
 
   @Override
   public int indexOf(Object rawValue) {
     assert rawValue instanceof byte[];
-    byte[] bytes = (byte[]) rawValue;
+    byte[] bytes = null;
+    // Convert hex string to byte[].
+    if (rawValue instanceof String) {
+      try {
+        bytes = Hex.decodeHex(((String) rawValue).toCharArray());
+      } catch (DecoderException e) {
+        Utils.rethrowException(e);
+      }
+    } else {
+      bytes = (byte[]) rawValue;
+    }
     return getDictId(new ByteArray(bytes), bytes);
   }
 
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOnHeapMutableDictionary.java
```diff
@@ -20,20 +20,29 @@
...
   private ByteArray _max = null;
 
   @Override
   public int indexOf(Object rawValue) {
     assert rawValue instanceof byte[];
-    return getDictId(new ByteArray((byte[]) rawValue));
+    byte[] bytes;
+    // Convert hex string to byte[].
+    if (rawValue instanceof String) {
+      bytes = DatatypeConverter.parseHexBinary((String) rawValue);
```
mayankshriv: Other parts of the code use org.apache.commons.codec.binary.Hex for encoding/decoding byte-array to String. Either one is fine, but all code should use the same encoding/decoding.

xiangfu1: sounds good, updated with org.apache.commons.codec.binary.Hex api.

fx19880617: sounds good, updated with org.apache.commons.codec.binary.Hex api.

kishoreg:
can we add a test case?

xiangfu1:
> can we add a test case?

Modified existing dictionary test to let BYTES dictionary index values contain half `byte[]` and half hex `String`.

fx19880617:
> can we add a test case?

Modified existing dictionary test to let BYTES dictionary index values contain half `byte[]` and half hex `String`.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOffHeapMutableDictionary.java
```diff
@@ -47,13 +51,24 @@
   public BytesOffHeapMutableDictionary(int estimatedCardinality, int maxOverflowHashSize,
       PinotDataBufferMemoryManager memoryManager, String allocationContext, int avgLength) {
     super(estimatedCardinality, maxOverflowHashSize, memoryManager, allocationContext);
-    _byteStore = new MutableOffHeapByteArrayStore(memoryManager, allocationContext, estimatedCardinality, avgLength);
+    _byteStore = new MutableOffHeapByteArrayStore(memoryManager, allocationContext,
```
mayankshriv: Minor nit: This seems like a codestyle diff? Please ensure to use the recommend code style. https://pinot.readthedocs.io/en/latest/dev_env.html#intellij

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOffHeapMutableDictionary.java
```diff
@@ -47,13 +51,24 @@
...
+        estimatedCardinality, avgLength);
   }
 
   @Override
   public int indexOf(Object rawValue) {
-    assert rawValue instanceof byte[];
-    byte[] bytes = (byte[]) rawValue;
+    byte[] bytes = null;
+    // Convert hex string to byte[].
+    if (rawValue instanceof String) {
```
mayankshriv: Will the common case be of String of byte[]? Common case should be the first 'if' condition (same in other api's as well).

mcvsubbu:
Can you add some motivation behind the change? I am guessing (but not sure) that we want to ingest string data that are known to be hexbytes. Is that right? Is this a part of a bigger design document? If so, can you point me to one (or, add one under https://cwiki.apache.org/confluence/display/PINOT/Design+Documents)
thanks

fx19880617:
> Can you add some motivation behind the change? I am guessing (but not sure) that we want to ingest string data that are known to be hexbytes. Is that right? Is this a part of a bigger design document? If so, can you point me to one (or, add one under https://cwiki.apache.org/confluence/display/PINOT/Design+Documents)
> thanks

My major motivation is related to this issue: https://github.com/apache/incubator-pinot/issues/4040.
Basically I want to ingest data like `uid` as bytes format and write a lookup query on top of it. Since hex string is used to display bytes type in query results, I extend this concept and make hex string/bytes both accepted for this column.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOffHeapMutableDictionary.java
fx19880617: will do

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOffHeapMutableDictionary.java
fx19880617: common case is still byte[], just wanna shorten code to still keep assert for other data types, so put String type first. I will change the if logic below to move the common case up.
```
if (rawValue instanceof byte[]) {
...
} if (rawValue instanceof String) {
...
} else {
assert rawValue instanceof byte[];
}

mcvsubbu:
> > Can you add some motivation behind the change? I am guessing (but not sure) that we want to ingest string data that are known to be hexbytes. Is that right? Is this a part of a bigger design document? If so, can you point me to one (or, add one under https://cwiki.apache.org/confluence/display/PINOT/Design+Documents)
> > thanks
> 
> My major motivation is related to this issue: #4040.
> Basically I want to ingest data like `uid` as bytes format and write a lookup query on top of it. Since hex string is used to display bytes type in query results, I extend this concept and make hex string/bytes both accepted for this column.

Is it not possible to use a string datatype for this? Assuming that the string is byte-array encoded does not seem right to me. Perhaps I am missing something here

fx19880617:
> > > Can you add some motivation behind the change? I am guessing (but not sure) that we want to ingest string data that are known to be hexbytes. Is that right? Is this a part of a bigger design document? If so, can you point me to one (or, add one under https://cwiki.apache.org/confluence/display/PINOT/Design+Documents)
> > > thanks
> > 
> > 
> > My major motivation is related to this issue: #4040.
> > Basically I want to ingest data like `uid` as bytes format and write a lookup query on top of it. Since hex string is used to display bytes type in query results, I extend this concept and make hex string/bytes both accepted for this column.
> 
> Is it not possible to use a string datatype for this? Assuming that the string is byte-array encoded does not seem right to me. Perhaps I am missing something here

This is to match the data in select query. Basically in `select uid from myTable limit 10` query, I got a random row with `uid` value shown as `"c8b3bce0b378fc5ce8067fc271a34892"` which is a hex string. And if I want to fetch this row with predicate, I should use the same hex string in my query, say 
`select * from myTable where uid="c8b3bce0b378fc5ce8067fc271a34892"`

xiangfu1:
will do

PR #4045, opened 2019-04-01
Add ORC docs

jenniferdai:


mcvsubbu:
just a note that we need to organize the doc per release, and make this for the next release.

s-sahay:
@jenniferdai : Let's merge this now.

PR #4047, opened 2019-04-01
Create leadControllerResource in helix cluster

jackjlli:
This PR creates lead controller resource in Helix cluster (if applicable) when starting a helix controller.

This new resource will be used for getting lead controller of each participant in this resource, so that these lead controllers can do the **periodic tasks** and **realtime segment completion** for their own partition.

The logic of handling this new resource will be added in the next following PR (controller and server separately).

Design doc: https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot
Test plan:
https://github.com/apache/incubator-pinot/issues/3957

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
@@ -37,10 +37,6 @@ public class TagNameUtils {
     return tenantName + "_" + TenantRole.BROKER.toString();
   }
 
-  public static boolean hasValidServerTagSuffix(String tagName) {
-    return tagName.endsWith(ServerType.REALTIME.toString()) || tagName.endsWith(ServerType.OFFLINE.toString());
-  }
-
   public static boolean isServerTag(String tagName) {
     return isOfflineServerTag(tagName) || isRealtimeServerTag(tagName);
   }
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
@@ -58,7 +58,7 @@ public class TagNameUtils {
     return tagName.endsWith(ServerType.REALTIME.toString());
   }
 
-  public static boolean isBrokerTags(String tagName) {
+  public static boolean isBrokerTag(String tagName) {
     return tagName.endsWith(TenantRole.BROKER.toString());
   }
 
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
@@ -18,8 +18,6 @@
  */
 package org.apache.pinot.common.config;
 
-import org.apache.pinot.common.exception.InvalidConfigException;
-import org.apache.pinot.common.utils.CommonConstants;
 import org.apache.pinot.common.utils.ServerType;
 import org.apache.pinot.common.utils.TenantRole;
 
@@ -40,10 +38,7 @@ public class TagNameUtils {
   }
 
   public static boolean hasValidServerTagSuffix(String tagName) {
-    if (tagName.endsWith(ServerType.REALTIME.toString()) || tagName.endsWith(ServerType.OFFLINE.toString())) {
-      return true;
-    }
-    return false;
+    return tagName.endsWith(ServerType.REALTIME.toString()) || tagName.endsWith(ServerType.OFFLINE.toString());
   }
 
   public static boolean isServerTag(String tagName) {
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
@@ -19,6 +19,7 @@
 package org.apache.pinot.common.config;
 
 import org.apache.pinot.common.exception.InvalidConfigException;
+import org.apache.pinot.common.utils.CommonConstants;
 import org.apache.pinot.common.utils.ServerType;
 import org.apache.pinot.common.utils.TenantRole;
 
@@ -45,8 +46,18 @@ public class TagNameUtils {
     return false;
   }
 
-  public static TenantRole getTenantRoleFromTag(String tagName)
+  public static boolean isServerTag(String tagName)
       throws InvalidConfigException {
+    return TenantRole.SERVER == getTenantRoleFromTag(tagName);
+  }
+
+  public static boolean isBrokerTags(String tagName)
+      throws InvalidConfigException {
+    return TenantRole.BROKER == getTenantRoleFromTag(tagName);
+  }
+
+  // Make this method private to avoid exposing null out of this class.
+  private static TenantRole getTenantRoleFromTag(String tagName) throws InvalidConfigException {
     if (tagName.endsWith(ServerType.REALTIME.toString())) {
       return TenantRole.SERVER;
     }
@@ -55,6 +66,11 @@ public class TagNameUtils {
     }
     if (tagName.endsWith(TenantRole.BROKER.toString())) {
       return TenantRole.BROKER;
+    }
+    // Helix uses this tag to support full-auto.
+    // Return null if the tag is controller, which isn't a type of tenant in Pinot.
+    if (tagName.equalsIgnoreCase(CommonConstants.Helix.CONTROLLER_INSTANCE_TYPE)) {
+      return null;
     }
     throw new InvalidConfigException("Cannot identify tenant type from tag name : " + tagName);
   }
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
@@ -46,33 +46,20 @@ public class TagNameUtils {
     return false;
   }
 
-  public static boolean isServerTag(String tagName)
-      throws InvalidConfigException {
-    return TenantRole.SERVER == getTenantRoleFromTag(tagName);
+  public static boolean isServerTag(String tagName) {
+    return isOfflineServerTag(tagName) || isRealtimeServerTag(tagName);
   }
 
-  public static boolean isBrokerTags(String tagName)
-      throws InvalidConfigException {
-    return TenantRole.BROKER == getTenantRoleFromTag(tagName);
+  public static boolean isOfflineServerTag(String tagName) {
+    return tagName.endsWith(ServerType.OFFLINE.toString());
   }
 
-  // Make this method private to avoid exposing null out of this class.
-  private static TenantRole getTenantRoleFromTag(String tagName) throws InvalidConfigException {
-    if (tagName.endsWith(ServerType.REALTIME.toString())) {
-      return TenantRole.SERVER;
-    }
-    if (tagName.endsWith(ServerType.OFFLINE.toString())) {
-      return TenantRole.SERVER;
-    }
-    if (tagName.endsWith(TenantRole.BROKER.toString())) {
-      return TenantRole.BROKER;
-    }
-    // Helix uses this tag to support full-auto.
-    // Return null if the tag is controller, which isn't a type of tenant in Pinot.
-    if (tagName.equalsIgnoreCase(CommonConstants.Helix.CONTROLLER_INSTANCE_TYPE)) {
-      return null;
-    }
-    throw new InvalidConfigException("Cannot identify tenant type from tag name : " + tagName);
+  public static boolean isRealtimeServerTag(String tagName) {
+    return tagName.endsWith(ServerType.REALTIME.toString());
+  }
+
+  public static boolean isBrokerTags(String tagName) {
+    return tagName.endsWith(TenantRole.BROKER.toString());
   }
 
   public static String getTagFromTenantAndServerType(String tenantName, ServerType type) {
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
@@ -44,8 +44,8 @@ public class CommonConstants {
 
     // More information on why these numbers are set can be found in the following doc:
     // https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot
-    public static final int NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 17;
-    public static final int NUMBER_OF_CONTROLLER_REPLICAS = 1;
+    public static final int NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 24;
+    public static final int LEAD_CONTROLLER_RESOURCE_REPLICA_COUNT = 1;
 
     public static final String UNTAGGED_SERVER_INSTANCE = "server_untagged";
     public static final String UNTAGGED_BROKER_INSTANCE = "broker_untagged";
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
@@ -46,6 +46,9 @@ public class CommonConstants {
     // https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot
     public static final int NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 24;
     public static final int LEAD_CONTROLLER_RESOURCE_REPLICA_COUNT = 1;
+    public static final boolean ENABLE_DELAY_REBALANCE = true;
+    public static final int MIN_ACTIVE_REPLICAS = 0;
+    public static final long REBALANCE_DELAY_MS = 300_000L; // 5 minutes.
 
     public static final String UNTAGGED_SERVER_INSTANCE = "server_untagged";
     public static final String UNTAGGED_BROKER_INSTANCE = "broker_untagged";
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
@@ -37,8 +37,16 @@ public class CommonConstants {
 
     public static final String SERVER_INSTANCE_TYPE = "server";
     public static final String BROKER_INSTANCE_TYPE = "broker";
+    public static final String CONTROLLER_INSTANCE_TYPE = "controller";
 
     public static final String BROKER_RESOURCE_INSTANCE = "brokerResource";
+    public static final String LEAD_CONTROLLER_RESOURCE_NAME = "leadControllerResource";
+
+    // This is for specifying the number of partitions in leadControllerResource.
+    // In most of the cases there won't be more than 17 controllers in one single cluster.
+    // Hard code the number of partitions to 17 since it's rarely changed and for less changes of hash collision.
+    public static final int DEFAULT_NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 17;
+    public static final int DEFAULT_NUMBER_OF_CONTROLLER_REPLICAS = 1;
 
     public static final String UNTAGGED_SERVER_INSTANCE = "server_untagged";
     public static final String UNTAGGED_BROKER_INSTANCE = "broker_untagged";
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
@@ -42,11 +42,10 @@ public class CommonConstants {
     public static final String BROKER_RESOURCE_INSTANCE = "brokerResource";
     public static final String LEAD_CONTROLLER_RESOURCE_NAME = "leadControllerResource";
 
-    // This is for specifying the number of partitions in leadControllerResource.
-    // In most of the cases there won't be more than 17 controllers in one single cluster.
-    // Hard code the number of partitions to 17 since it's rarely changed and for less changes of hash collision.
-    public static final int DEFAULT_NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 17;
-    public static final int DEFAULT_NUMBER_OF_CONTROLLER_REPLICAS = 1;
+    // More information on why these numbers are set can be found in the following doc:
+    // https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot
+    public static final int NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 17;
+    public static final int NUMBER_OF_CONTROLLER_REPLICAS = 1;
 
     public static final String UNTAGGED_SERVER_INSTANCE = "server_untagged";
     public static final String UNTAGGED_BROKER_INSTANCE = "broker_untagged";
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
@@ -57,8 +57,6 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final String CONSOLE_WEBAPP_USE_HTTPS = "controller.query.console.useHttps";
   private static final String EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT = "controller.upload.onlineToOfflineTimeout";
   private static final String CONTROLLER_MODE = "controller.mode";
-  private static final String NUMBER_OF_CONTROLLER_REPLICAS = "controller.number.replicas";
-  private static final String NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = "controller.number.partitions";
 
   public enum ControllerMode {
     DUAL,
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
@@ -57,6 +57,8 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final String CONSOLE_WEBAPP_USE_HTTPS = "controller.query.console.useHttps";
   private static final String EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT = "controller.upload.onlineToOfflineTimeout";
   private static final String CONTROLLER_MODE = "controller.mode";
+  private static final String NUMBER_OF_CONTROLLER_REPLICAS = "controller.number.replicas";
+  private static final String NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = "controller.number.partitions";
 
   public enum ControllerMode {
     DUAL,
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
@@ -220,7 +220,7 @@ public class ControllerStarter {
     // Register and connect instance as Helix controller.
     LOGGER.info("Starting Helix controller");
     _helixControllerManager = HelixSetupUtils
-        .setup(_helixClusterName, _helixZkURL, _instanceId, _isUpdateStateModel, _enableBatchMessageMode);
+        .setup(_helixClusterName, _helixZkURL, _instanceId);
 
     // Emit helix controller metrics
     _controllerMetrics.addCallbackGauge(CommonConstants.Helix.INSTANCE_CONNECTED_METRIC_NAME,
@@ -243,6 +243,9 @@ public class ControllerStarter {
     if (_controllerMode == ControllerConf.ControllerMode.PINOT_ONLY && !isPinotOnlyModeSupported()) {
       throw new RuntimeException("Pinot only controller currently isn't supported in production yet.");
     }
+
+    // Set up Pinot cluster in Helix
+    HelixSetupUtils.setupPinotCluster(_helixClusterName, _helixZkURL, _isUpdateStateModel, _enableBatchMessageMode);
 
     // Start all components
     initPinotFSFactory();
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
@@ -219,8 +219,7 @@ public class ControllerStarter {
   private void setUpHelixController() {
     // Register and connect instance as Helix controller.
     LOGGER.info("Starting Helix controller");
-    _helixControllerManager = HelixSetupUtils
-        .setup(_helixClusterName, _helixZkURL, _instanceId);
+    _helixControllerManager = HelixSetupUtils.setup(_helixClusterName, _helixZkURL, _instanceId);
 
     // Emit helix controller metrics
     _controllerMetrics.addCallbackGauge(CommonConstants.Helix.INSTANCE_CONNECTED_METRIC_NAME,
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -1170,7 +1170,7 @@ public class PinotHelixResourceManager {
     if (tagOverrideConfig != null) {
       String realtimeConsumingTag = tagOverrideConfig.getRealtimeConsuming();
       if (realtimeConsumingTag != null) {
-        if (!TagNameUtils.hasValidServerTagSuffix(realtimeConsumingTag)) {
+        if (!TagNameUtils.isServerTag(realtimeConsumingTag)) {
           throw new InvalidTableConfigException(
               "Invalid realtime consuming tag: " + realtimeConsumingTag + " for table " + tableNameWithType
                   + ". Must have suffix _REALTIME or _OFFLINE");
@@ -1184,7 +1184,7 @@ public class PinotHelixResourceManager {
 
       String realtimeCompletedTag = tagOverrideConfig.getRealtimeCompleted();
       if (realtimeCompletedTag != null) {
-        if (!TagNameUtils.hasValidServerTagSuffix(realtimeCompletedTag)) {
+        if (!TagNameUtils.isServerTag(realtimeCompletedTag)) {
           throw new InvalidTableConfigException(
               "Invalid realtime completed tag: " + realtimeCompletedTag + " for table " + tableNameWithType
                   + ". Must have suffix _REALTIME or _OFFLINE");
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -53,6 +53,7 @@ import org.apache.helix.PropertyKey;
 import org.apache.helix.PropertyKey.Builder;
 import org.apache.helix.PropertyPathBuilder;
 import org.apache.helix.ZNRecord;
+import org.apache.helix.examples.MasterSlaveStateModelFactory;
 import org.apache.helix.manager.zk.ZkBaseDataAccessor;
 import org.apache.helix.manager.zk.ZkCacheBaseDataAccessor;
 import org.apache.helix.model.CurrentState;
@@ -60,6 +61,8 @@ import org.apache.helix.model.ExternalView;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.LiveInstance;
+import org.apache.helix.model.MasterSlaveSMD;
+import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.store.zk.ZkHelixPropertyStore;
 import org.apache.pinot.common.config.IndexingConfig;
 import org.apache.pinot.common.config.OfflineTagConfig;
@@ -92,7 +95,6 @@ import org.apache.pinot.common.utils.CommonConstants.Helix.StateModel.BrokerOnli
 import org.apache.pinot.common.utils.CommonConstants.Helix.StateModel.SegmentOnlineOfflineStateModel;
 import org.apache.pinot.common.utils.CommonConstants.Helix.TableType;
 import org.apache.pinot.common.utils.SchemaUtils;
-import org.apache.pinot.common.utils.TenantRole;
 import org.apache.pinot.common.utils.helix.HelixHelper;
 import org.apache.pinot.common.utils.helix.PinotHelixPropertyStoreZnRecordProvider;
 import org.apache.pinot.common.utils.retry.RetryPolicies;
@@ -127,6 +129,7 @@ public class PinotHelixResourceManager {
   private final String _helixZkURL;
   private final String _helixClusterName;
   private final String _instanceId;
+  private final String _controllerParticipantInstanceId;
   private final String _dataDir;
   private final long _externalViewOnlineToOfflineTimeoutMillis;
   private final boolean _isSingleTenantCluster;
@@ -150,6 +153,7 @@ public class PinotHelixResourceManager {
     _helixZkURL = HelixConfig.getAbsoluteZkPathForHelix(zkURL);
     _helixClusterName = helixClusterName;
     _instanceId = controllerInstanceId;
+    _controllerParticipantInstanceId = CommonConstants.Helix.PREFIX_OF_CONTROLLER_INSTANCE + _instanceId;
     _dataDir = dataDir;
     _externalViewOnlineToOfflineTimeoutMillis = externalViewOnlineToOfflineTimeoutMillis;
     _isSingleTenantCluster = isSingleTenantCluster;
@@ -185,6 +189,10 @@ public class PinotHelixResourceManager {
     _cacheInstanceConfigsDataAccessor =
         new ZkCacheBaseDataAccessor<>((ZkBaseDataAccessor<ZNRecord>) baseDataAccessor, instanceConfigs, null,
             Collections.singletonList(instanceConfigs));
+
+    // Add instance group tag for controller
+    addInstanceGroupTag();
+
     _keyBuilder = _helixDataAccessor.keyBuilder();
     _segmentDeletionManager = new SegmentDeletionManager(_dataDir, _helixAdmin, _helixClusterName, _propertyStore);
     ZKMetadataProvider.setClusterTenantIsolationEnabled(_propertyStore, _isSingleTenantCluster);
@@ -260,17 +268,31 @@ public class PinotHelixResourceManager {
    */
   private HelixManager registerAndConnectAsHelixParticipant() {
     HelixManager helixManager = HelixManagerFactory
-        .getZKHelixManager(_helixClusterName, CommonConstants.Helix.PREFIX_OF_CONTROLLER_INSTANCE + _instanceId,
-            InstanceType.PARTICIPANT, _helixZkURL);
+        .getZKHelixManager(_helixClusterName, _controllerParticipantInstanceId, InstanceType.PARTICIPANT, _helixZkURL);
+
+    StateMachineEngine stateMach = helixManager.getStateMachineEngine();
+    MasterSlaveStateModelFactory factory = new MasterSlaveStateModelFactory();
+    stateMach.registerStateModelFactory(MasterSlaveSMD.name, factory);
     try {
       helixManager.connect();
       return helixManager;
     } catch (Exception e) {
-      String errorMsg =
-          String.format("Exception when connecting the instance %s as Participant to Helix.", _instanceId);
+      String errorMsg = String.format("Exception when connecting the instance %s as Participant to Helix.",
+          _controllerParticipantInstanceId);
       LOGGER.error(errorMsg, e);
       throw new RuntimeException(errorMsg);
     }
+  }
+
+  /**
+   * Add instance group tag for controller so that pinot controller can be assigned to lead controller resource.
+   */
+  private void addInstanceGroupTag() {
+    _helixZkManager.getClusterManagmentTool().enableInstance(_helixClusterName, _controllerParticipantInstanceId, true);
+    InstanceConfig instanceConfig = getHelixInstanceConfig(_controllerParticipantInstanceId);
+    instanceConfig.addTag(CommonConstants.Helix.CONTROLLER_INSTANCE_TYPE);
+    HelixDataAccessor accessor = _helixZkManager.getHelixDataAccessor();
+    accessor.setProperty(accessor.keyBuilder().instanceConfig(_controllerParticipantInstanceId), instanceConfig);
   }
 
   /**
@@ -840,15 +862,12 @@ public class PinotHelixResourceManager {
             .equals(CommonConstants.Minion.UNTAGGED_INSTANCE)) {
           continue;
         }
-        TenantRole tenantRole;
         try {
-          tenantRole = TagNameUtils.getTenantRoleFromTag(tag);
-          if (tenantRole == TenantRole.BROKER) {
+          if (TagNameUtils.isBrokerTags(tag)) {
             tenantSet.add(TagNameUtils.getTenantNameFromTag(tag));
           }
         } catch (InvalidConfigException e) {
           LOGGER.warn("Instance {} contains an invalid tag: {}", instanceName, tag);
-          continue;
         }
       }
     }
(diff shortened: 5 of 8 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -58,6 +57,8 @@
   private static final String CONSOLE_WEBAPP_USE_HTTPS = "controller.query.console.useHttps";
   private static final String EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT = "controller.upload.onlineToOfflineTimeout";
   private static final String CONTROLLER_MODE = "controller.mode";
+  private static final String NUMBER_OF_CONTROLLER_REPLICAS = "controller.number.replicas";
```
mcvsubbu: Is this number of pinot controller hosts? Not sure why we need to configure this in. Would be best if we did not have to configure number of controller replicas. We should be able to roll out additional controllers as needed.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -64,10 +65,12 @@
   private static final Logger LOGGER = LoggerFactory.getLogger(HelixSetupUtils.class);
 
   public static synchronized HelixManager setup(String helixClusterName, String zkPath,
-      String pinotControllerInstanceId, boolean isUpdateStateModel, boolean enableBatchMessageMode) {
+      String pinotControllerInstanceId, boolean isUpdateStateModel, boolean enableBatchMessageMode,
+      LeadControllerResourceConfig leadControllerResourceConfig) {
 
     try {
-      createHelixClusterIfNeeded(helixClusterName, zkPath, isUpdateStateModel, enableBatchMessageMode);
+      createHelixClusterIfNeeded(helixClusterName, zkPath, isUpdateStateModel, enableBatchMessageMode,
```
mcvsubbu: can u add .a TODO that we should not be creating helix cluster in pinot-only mode? thanks

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -110,6 +113,8 @@ public static void createHelixClusterIfNeeded(String helixClusterName, String zk
           zkClient.close();
         }
       }
+      createLeadControllerResourceIfNeeded(helixClusterName, admin, enableBatchMessageMode,
```
mcvsubbu: When we already have a cluster and everything is good, we will return from line 100, right? So, we will never create controller resource

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -154,6 +159,8 @@ public static void createHelixClusterIfNeeded(String helixClusterName, String zk
         .buildEmptyIdealStateForBrokerResource(admin, helixClusterName, enableBatchMessageMode);
     admin.setResourceIdealState(helixClusterName, CommonConstants.Helix.BROKER_RESOURCE_INSTANCE, idealState);
     initPropertyStorePath(helixClusterName, zkPath);
+
+    createLeadControllerResourceIfNeeded(helixClusterName, admin, enableBatchMessageMode, leadControllerResourceConfig);
```
mcvsubbu: can we do this outside this method and have the caller call createLeadControllerResourceIfNeeded?

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
jackjlli: Sure, I'll make it non-configurable for now.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: `HelixSetupUtils .setup()` method will only be called by either Helix only mode or dual mode controller (in `setUpHelixController` method of `ControllerStarter`).

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Good catch! I'll add that after Line 100.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: The reason why I put it inside `createHelixClusterIfNeeded` is that `createHelixClusterIfNeeded` method can only focus on setting up cluster steps, whereas `startHelixControllerInStandadloneMode` method can only focus on initializing controller.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: I created a new method called `setupHelixCluster()` which will call `createHelixClusterIfNeeded` and `createLeadControllerResourceIfNeeded` sequentially.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -134,6 +135,7 @@ private static long getRandomInitialDelayInSeconds() {
   private static final String ENABLE_STORAGE_QUOTA_CHECK = "controller.enable.storage.quota.check";
 
   private static final String ENABLE_BATCH_MESSAGE_MODE = "controller.enable.batch.message.mode";
+  private static final String ENABLE_LEAD_CONTROLLER_RESOURCE = "controller.enable.lead.controller.resource";
```
mcvsubbu: Will be easier to review  if you put up a design doc in cwiki. thanks,

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
jackjlli: Updated.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -35,8 +35,12 @@
 
     public static final String SERVER_INSTANCE_TYPE = "server";
     public static final String BROKER_INSTANCE_TYPE = "broker";
+    public static final String CONTROLLER_INSTANCE_TYPE = "controller";
 
     public static final String BROKER_RESOURCE_INSTANCE = "brokerResource";
+    public static final String LEAD_CONTROLLER_RESOURCE_INSTANCE = "leadControllerResource";
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -58,6 +57,8 @@
   private static final String CONSOLE_WEBAPP_USE_HTTPS = "controller.query.console.useHttps";
   private static final String EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT = "controller.upload.onlineToOfflineTimeout";
   private static final String CONTROLLER_MODE = "controller.mode";
+  private static final String NUMBER_OF_CONTROLLER_REPLICAS = "controller.number.replicas";
+  private static final String NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = "controller.number.partitions";
```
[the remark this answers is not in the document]
jackjlli: Done. Moved under CommonConstant.Helix.

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
```diff
@@ -55,6 +56,11 @@ public static TenantRole getTenantRoleFromTag(String tagName) throws InvalidConf
     if (tagName.endsWith(TenantRole.BROKER.toString())) {
       return TenantRole.BROKER;
     }
+    // Helix uses this tag to support full-auto.
+    // Return null if the tag is controller, which isn't a type of tenant in Pinot.
+    if (tagName.equalsIgnoreCase(CommonConstants.Helix.CONTROLLER_INSTANCE_TYPE)) {
+      return null;
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -175,4 +185,40 @@ private static HelixManager startHelixControllerInStandadloneMode(String helixCl
     return HelixControllerMain
         .startHelixController(zkUrl, helixClusterName, pinotControllerInstanceId, HelixControllerMain.STANDALONE);
   }
+
+  private static void createLeadControllerResourceIfNeeded(String helixClusterName, HelixAdmin admin,
+      boolean enableBatchMessageMode, boolean enableLeadControllerResource) {
+    List<String> resources = admin.getResourcesInCluster(helixClusterName);
```
mcvsubbu: Instead of getting all resources (which can number in 1000s) can we just get the idealstate of the lead controller resource to check if it is there?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -175,4 +185,40 @@ private static HelixManager startHelixControllerInStandadloneMode(String helixCl
...
+          admin.getResourceIdealState(helixClusterName, CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME);
+      leadControllerIdealState.setInstanceGroupTag(CommonConstants.Helix.CONTROLLER_INSTANCE_TYPE);
+      leadControllerIdealState.setBatchMessageMode(enableBatchMessageMode);
+      admin.setResourceIdealState(helixClusterName, CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME,
+          leadControllerIdealState);
+    }
+
+    LOGGER.info("Re-balance lead controller resource with replicas: {}",
+        CommonConstants.Helix.DEFAULT_NUMBER_OF_CONTROLLER_REPLICAS);
+    admin.rebalance(helixClusterName, CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME,
```
mcvsubbu: Do we need to do this each time helix (or dual) controller is restarted? Only the first time right?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -175,4 +185,40 @@ private static HelixManager startHelixControllerInStandadloneMode(String helixCl
...
+        CommonConstants.Helix.DEFAULT_NUMBER_OF_CONTROLLER_REPLICAS);
+
+    // Enable or disable lead controller resource.
+    LOGGER.info("{} lead controller resource.", enableLeadControllerResource ? "Enabling" : "Disabling");
+    admin.enableResource(helixClusterName, CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME,
```
mcvsubbu: remove this line. We need to enable/disable resource directly via helix-admin

jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Moved it inside of the bracket.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: That's a good idea! Addressed.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -37,8 +37,16 @@
...
     public static final String SERVER_INSTANCE_TYPE = "server";
     public static final String BROKER_INSTANCE_TYPE = "broker";
+    public static final String CONTROLLER_INSTANCE_TYPE = "controller";
 
     public static final String BROKER_RESOURCE_INSTANCE = "brokerResource";
+    public static final String LEAD_CONTROLLER_RESOURCE_NAME = "leadControllerResource";
+
+    public static final String MASTER_SLAVE_STATE_MODEL_DEFINITION = "MasterSlave";
+
+    // Hard code the number of partitions to 17 since it's rarely changed.
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -65,25 +69,32 @@
...
       return null;
     }
   }
 
-  public static void createHelixClusterIfNeeded(String helixClusterName, String zkPath, boolean isUpdateStateModel,
+  public static void setupHelixCluster(String helixClusterName, String zkPath, boolean isUpdateStateModel,
       boolean enableBatchMessageMode) {
     final HelixAdmin admin = new ZKHelixAdmin(zkPath);
+
+    createHelixClusterIfNeeded(helixClusterName, zkPath, admin, isUpdateStateModel, enableBatchMessageMode);
```
mcvsubbu: in t his method we should create helix cluster only. Anything else like adding state models or creating broker/lead-controller resources, etc should be done by pinot-mode (or dual mode) controller.

jackjlli: PR updated.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -838,15 +860,12 @@ public boolean isServerTenantDeletable(String tenantName) {
...
-        TenantRole tenantRole;
         try {
-          tenantRole = TagNameUtils.getTenantRoleFromTag(tag);
-          if (tenantRole == TenantRole.BROKER) {
+          if (TagNameUtils.isBrokerTags(tag)) {
             tenantSet.add(TagNameUtils.getTenantNameFromTag(tag));
           }
         } catch (InvalidConfigException e) {
           LOGGER.warn("Instance {} contains an invalid tag: {}", instanceName, tag);
-          continue;
```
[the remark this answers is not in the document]

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -37,8 +37,16 @@
 
     public static final String SERVER_INSTANCE_TYPE = "server";
     public static final String BROKER_INSTANCE_TYPE = "broker";
+    public static final String CONTROLLER_INSTANCE_TYPE = "controller";
 
     public static final String BROKER_RESOURCE_INSTANCE = "brokerResource";
+    public static final String LEAD_CONTROLLER_RESOURCE_NAME = "leadControllerResource";
+
+    // This is for specifying the number of partitions in leadControllerResource.
```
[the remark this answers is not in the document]
jackjlli: Doc added.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -864,15 +883,12 @@ public boolean isServerTenantDeletable(String tenantName) {
...
-        TenantRole tenantRole;
         try {
-          tenantRole = TagNameUtils.getTenantRoleFromTag(tag);
-          if (tenantRole == TenantRole.SERVER) {
+          if (TagNameUtils.isServerTag(tag)) {
             tenantSet.add(TagNameUtils.getTenantNameFromTag(tag));
           }
         } catch (InvalidConfigException e) {
           LOGGER.warn("Instance {} contains an invalid tag: {}", instanceName, tag);
-          continue;
```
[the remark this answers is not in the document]

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -2380,6 +2398,16 @@ public boolean isSingleTenantCluster() {
     return endpointToInstance;
   }
 
+  public IdealState getLeadControllerResourceIdealState() {
```
Jackie-Jiang: Do we need this two methods in the resource manager?

jackjlli: The method `getLeadControllerResourceIdealState` is to check whether this resource is enabled or not. `getLeadControllerExternalView` is to fetch the master controller of partition, which will be used for realtime segment completion.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -37,8 +37,15 @@
...
     public static final String BROKER_INSTANCE_TYPE = "broker";
+    public static final String CONTROLLER_INSTANCE_TYPE = "controller";
 
     public static final String BROKER_RESOURCE_INSTANCE = "brokerResource";
+    public static final String LEAD_CONTROLLER_RESOURCE_NAME = "leadControllerResource";
+
+    // More information on why these numbers are set can be found in the following doc:
+    // https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot
+    public static final int DEFAULT_NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 17;
+    public static final int DEFAULT_NUMBER_OF_CONTROLLER_REPLICAS = 1;
```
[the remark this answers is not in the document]
jackjlli: The reason is in the ciwiki link above. There's no huge difference between 1 or more, except there'll be SLAVEs shown in the resource if the number is > 1. For Pinot, it takes minimum efforts to transit from OFFLINE to SLAVE and from SLAVE to MASTER. Thus, we don't need SLAVE to be shown in the resource.

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
```diff
@@ -45,7 +46,18 @@ public static boolean hasValidServerTagSuffix(String tagName) {
...
+    return TenantRole.SERVER == getTenantRoleFromTag(tagName);
+  }
+
+  public static boolean isBrokerTags(String tagName)
+      throws InvalidConfigException {
+    return TenantRole.BROKER == getTenantRoleFromTag(tagName);
+  }
+
+  // Make this method private to avoid exposing null out of this class.
+  private static TenantRole getTenantRoleFromTag(String tagName) throws InvalidConfigException {
```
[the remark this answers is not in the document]
jackjlli: I think one of the reasons of keeping that is `tagName` can be misconfigured. This method `isServerTag` is to check whether it's a wrong tag name. Otherwise it will just return `false` and provides no extra information, which makes it harder to figure out.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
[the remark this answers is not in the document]
mcvsubbu: Also, Pinot does not do anything in slaves. Master partitions do all the work.

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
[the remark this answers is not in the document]
jackjlli: Discussed with @Jackie-Jiang. The exception should have been caught when table config was added/updated. So there's no need to keep the validation check of tag name here.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -37,8 +37,15 @@
 
     public static final String SERVER_INSTANCE_TYPE = "server";
     public static final String BROKER_INSTANCE_TYPE = "broker";
+    public static final String CONTROLLER_INSTANCE_TYPE = "controller";
 
     public static final String BROKER_RESOURCE_INSTANCE = "brokerResource";
+    public static final String LEAD_CONTROLLER_RESOURCE_NAME = "leadControllerResource";
```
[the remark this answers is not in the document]
jackjlli: This resource only focuses on lead controller assignment, so the name should be fine. :)

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -37,8 +37,15 @@
...
+
+    // More information on why these numbers are set can be found in the following doc:
+    // https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot
+    public static final int NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 17;
+    public static final int NUMBER_OF_CONTROLLER_REPLICAS = 1;
```
mcvsubbu: Please rename this to LEAD_CONTROLLER_RESOURCE_REPLICA_COUNT

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
```diff
@@ -67,6 +72,25 @@ public void testDualModeController()
...
+    controllerConfig.setControllerPort(Integer.toString(Integer.parseInt(this.config.getControllerPort()) + controllerPortOffset++));
+
+    ControllerStarter secondDualModeController = new TestOnlyControllerStarter(controllerConfig);
+    secondDualModeController.start();
+    TestUtils.waitForCondition(
+        aVoid -> secondDualModeController.getHelixResourceManager().getHelixZkManager().isConnected(), TIMEOUT_IN_MS,
+        "Failed to start " + config.getControllerMode() + " controller in " + TIMEOUT_IN_MS + "ms.");
+    Assert.assertEquals(secondDualModeController.getControllerMode(), ControllerConf.ControllerMode.DUAL);
+
+    Thread.sleep(100000_000L);
```
mcvsubbu: clean up

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
```diff
@@ -67,6 +72,25 @@ public void testDualModeController()
         "Failed to start " + config.getControllerMode() + " controller in " + TIMEOUT_IN_MS + "ms.");
     Assert.assertEquals(_controllerStarter.getControllerMode(), ControllerConf.ControllerMode.DUAL);
 
+    // Enable the lead controller resource.
```
mcvsubbu: You also need a method to wipe zk before each test case

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -37,8 +37,15 @@
...
     public static final String SERVER_INSTANCE_TYPE = "server";
     public static final String BROKER_INSTANCE_TYPE = "broker";
+    public static final String CONTROLLER_INSTANCE_TYPE = "controller";
 
     public static final String BROKER_RESOURCE_INSTANCE = "brokerResource";
+    public static final String LEAD_CONTROLLER_RESOURCE_NAME = "leadControllerResource";
+
+    // More information on why these numbers are set can be found in the following doc:
+    // https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot
+    public static final int DEFAULT_NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 17;
```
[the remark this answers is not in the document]
jackjlli: A test `testLeadControllerAssignment` was added in PR to verify that the difference of the number of masters for all controllers is at most 1. And it shouldn't be that bad to have 1 more/ 1 less partition assignment.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -260,19 +268,33 @@ public HelixAdmin getHelixAdmin() {
...
       LOGGER.error(errorMsg, e);
       throw new RuntimeException(errorMsg);
     }
   }
 
+  /**
+   * Add instance group tag for controller so that pinot controller can be assigned to lead controller resource.
+   */
+  private void addInstanceGroupTag() {
+    _helixZkManager.getClusterManagmentTool().enableInstance(_helixClusterName, _controllerParticipantInstanceId, true);
```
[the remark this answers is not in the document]
jackjlli: Good point. Removed.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
jackjlli: Check the test code. There's a method called `protected void startController(ControllerConf config, boolean deleteCluster)`. If  `deleteCluster` is true, it'll wipe out the existing cluster info. So no need to implement a wipe out method any more.

jackjlli:
> Add notes about manual tests done, thanks

Test plan added in this issue: https://github.com/apache/incubator-pinot/issues/3957

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -220,7 +220,7 @@ private void setUpHelixController() {
     // Register and connect instance as Helix controller.
     LOGGER.info("Starting Helix controller");
     _helixControllerManager = HelixSetupUtils
-        .setup(_helixClusterName, _helixZkURL, _instanceId, _isUpdateStateModel, _enableBatchMessageMode);
+        .setup(_helixClusterName, _helixZkURL, _instanceId);
```
Jackie-Jiang: (nit) reformat this line

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -866,15 +880,8 @@ public boolean isServerTenantDeletable(String tenantName) {
...
-        TenantRole tenantRole;
-        try {
-          tenantRole = TagNameUtils.getTenantRoleFromTag(tag);
-          if (tenantRole == TenantRole.SERVER) {
-            tenantSet.add(TagNameUtils.getTenantNameFromTag(tag));
-          }
-        } catch (InvalidConfigException e) {
-          LOGGER.warn("Instance {} contains an invalid tag: {}", instanceName, tag);
-          continue;
+        if (TagNameUtils.isServerTag(tag)) {
```
Jackie-Jiang: Remove line 878-882

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -840,15 +861,8 @@ public boolean isServerTenantDeletable(String tenantName) {
...
-        TenantRole tenantRole;
-        try {
-          tenantRole = TagNameUtils.getTenantRoleFromTag(tag);
-          if (tenantRole == TenantRole.BROKER) {
-            tenantSet.add(TagNameUtils.getTenantNameFromTag(tag));
-          }
-        } catch (InvalidConfigException e) {
-          LOGGER.warn("Instance {} contains an invalid tag: {}", instanceName, tag);
-          continue;
+        if (TagNameUtils.isBrokerTags(tag)) {
```
Jackie-Jiang: Remove line 859-863

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -127,6 +129,7 @@
   private final String _helixZkURL;
   private final String _helixClusterName;
   private final String _instanceId;
+  private final String _controllerParticipantInstanceId;
```
Jackie-Jiang: I'll remove this field and directly use _instanceId. In the constructor, directly pass in `Controller_xxx`

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -260,19 +268,32 @@ public HelixAdmin getHelixAdmin() {
...
+          _controllerParticipantInstanceId);
       LOGGER.error(errorMsg, e);
       throw new RuntimeException(errorMsg);
     }
   }
 
+  /**
+   * Add instance group tag for controller so that pinot controller can be assigned to lead controller resource.
+   */
+  private void addInstanceGroupTag() {
```
Jackie-Jiang: Change this to `addTagIfNeeded` and only set instanceConfig if the tag does not exist already.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -260,19 +268,32 @@ public HelixAdmin getHelixAdmin() {
    */
   private HelixManager registerAndConnectAsHelixParticipant() {
     HelixManager helixManager = HelixManagerFactory
-        .getZKHelixManager(_helixClusterName, CommonConstants.Helix.PREFIX_OF_CONTROLLER_INSTANCE + _instanceId,
-            InstanceType.PARTICIPANT, _helixZkURL);
+        .getZKHelixManager(_helixClusterName, _controllerParticipantInstanceId, InstanceType.PARTICIPANT, _helixZkURL);
+
+    StateMachineEngine stateMach = helixManager.getStateMachineEngine();
```
Jackie-Jiang: Line 273-275 can be merged into one line, and please add some comments for this.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: I understand. My question is, should we keep these two methods in the resource manager (IMO they are not very related to the resource manager), or directly use helix admin to fetch them (same as table resource and broker resource.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -53,126 +55,217 @@
...
  *
  */
 public class HelixSetupUtils {
-
   private static final Logger LOGGER = LoggerFactory.getLogger(HelixSetupUtils.class);
 
   public static synchronized HelixManager setup(String helixClusterName, String zkPath,
-      String pinotControllerInstanceId, boolean isUpdateStateModel, boolean enableBatchMessageMode) {
-
+      String pinotControllerInstanceId) {
```
Jackie-Jiang: helixControllerInstanceId?

Jackie-Jiang: We should throw the exception out instead of logging error and returning null.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -53,126 +55,217 @@
...
+
+  /**
+   * Customizes existing Helix cluster to run Pinot components.
+   */
+  public static void setupPinotCluster(String helixClusterName, String zkPath, boolean isUpdateStateModel,
+      boolean enableBatchMessageMode) {
+    final HelixAdmin admin = new ZKHelixAdmin(zkPath);
+    if (!admin.getClusters().contains(helixClusterName)) {
+      LOGGER.error("Helix cluster: {} hasn't been set up", helixClusterName);
+      throw new RuntimeException();
```
Jackie-Jiang: Use Preconditions.checkState() and add exception message. No need to log error here.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -53,126 +55,217 @@
...
+    }
+
+    // Ensure auto join.
+    ensureAutoJoin(helixClusterName, admin);
+
+    // Add segment state model definition if needed
+    addSegmentStateModelDefinitionIfNeeded(helixClusterName, admin, zkPath, isUpdateStateModel);
+
+    // Add broker resource online offline state model definition if needed
+    addBrokerResourceOnlineOfflineStateModelDefinitionIfNeeded(helixClusterName, admin);
```
Jackie-Jiang: addBrokerResourceOnlineOfflineStateModelDefinitionIfNeeded() and createBrokerResourceIfNeeded() can be combined into one right? (Same as the leadControllerResource)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -53,126 +55,217 @@
...
-
-    LOGGER.info("Adding empty ideal state for Broker!");
-    HelixHelper.updateResourceConfigsFor(new HashMap<String, String>(), CommonConstants.Helix.BROKER_RESOURCE_INSTANCE,
-        helixClusterName, admin);
-    IdealState idealState = PinotTableIdealStateBuilder
-        .buildEmptyIdealStateForBrokerResource(admin, helixClusterName, enableBatchMessageMode);
-    admin.setResourceIdealState(helixClusterName, CommonConstants.Helix.BROKER_RESOURCE_INSTANCE, idealState);
-    initPropertyStorePath(helixClusterName, zkPath);
-    LOGGER.info("New Cluster setup completed... ********************************************** ");
+  private static void addSegmentStateModelDefinitionIfNeeded(String helixClusterName, HelixAdmin admin, String zkPath,
```
Jackie-Jiang: Most of the part for this method is for LLC migration. Since it's done, remove the redundant part (only keep line 156-168, and remove redundant arguments)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -53,126 +55,217 @@
...
+      admin.addResource(helixClusterName, LEAD_CONTROLLER_RESOURCE_NAME,
+          CommonConstants.Helix.NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE, MasterSlaveSMD.name,
+          IdealState.RebalanceMode.FULL_AUTO.toString(), CrushEdRebalanceStrategy.class.getName());
+
+      // Set instance group tag for lead controller resource.
+      IdealState leadControllerIdealState =
+          admin.getResourceIdealState(helixClusterName, LEAD_CONTROLLER_RESOURCE_NAME);
+      leadControllerIdealState.setInstanceGroupTag(CommonConstants.Helix.CONTROLLER_INSTANCE_TYPE);
+      leadControllerIdealState.setBatchMessageMode(enableBatchMessageMode);
+      leadControllerIdealState.setMinActiveReplicas(0);
```
Jackie-Jiang: What does this 0 mean? Please add comments

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: Right now `helix admin` isn't exposed to controller starter yet. If we choose the 2nd plan, it needs to expose to controller starter. Or always call `pinotHelixResourceManager.getHelixAdmin()` to get it.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Updated.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: We've thought about it. Most of the time that part of code won't be touched. But it's good to leave this code there just for backward compatibility.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Added.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
mcvsubbu: It is not clear that everyone using open source pinot has migrated from HLC to LLC. So, please keep this code.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Merged.

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
```diff
@@ -45,18 +46,20 @@ public static boolean hasValidServerTagSuffix(String tagName) {
...
+
+  public static boolean isOfflineServerTag(String tagName) {
+    return tagName.endsWith(ServerType.OFFLINE.toString());
+  }
+
+  public static boolean isRealtimeServerTag(String tagName) {
+    return tagName.endsWith(ServerType.REALTIME.toString());
+  }
+
+  public static boolean isBrokerTags(String tagName) {
```
Jackie-Jiang: isBrokerTag

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -159,13 +161,14 @@ public PinotHelixResourceManager(@Nonnull String zkURL, @Nonnull String helixClu
 
   public PinotHelixResourceManager(@Nonnull String zkURL, @Nonnull String helixClusterName,
       @Nonnull String controllerInstanceId, @Nonnull String dataDir) {
-    this(zkURL, helixClusterName, controllerInstanceId, dataDir, DEFAULT_EXTERNAL_VIEW_UPDATE_TIMEOUT_MILLIS, false,
-        true, true);
+    this(zkURL, helixClusterName, CommonConstants.Helix.PREFIX_OF_CONTROLLER_INSTANCE + controllerInstanceId, dataDir,
```
Jackie-Jiang: controllerInstanceId should already have the prefix. Also, maybe we can remove this constructor?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -259,9 +266,12 @@ public HelixAdmin getHelixAdmin() {
...
   private HelixManager registerAndConnectAsHelixParticipant() {
-    HelixManager helixManager = HelixManagerFactory
-        .getZKHelixManager(_helixClusterName, CommonConstants.Helix.PREFIX_OF_CONTROLLER_INSTANCE + _instanceId,
-            InstanceType.PARTICIPANT, _helixZkURL);
+    HelixManager helixManager =
+        HelixManagerFactory.getZKHelixManager(_helixClusterName, _instanceId, InstanceType.PARTICIPANT, _helixZkURL);
+
+    // Registers Master-Slave state model to state machine engine, which is for calculating participant assignment in lead controller resource.
+    helixManager.getStateMachineEngine()
+        .registerStateModelFactory(MasterSlaveSMD.name, new MasterSlaveStateModelFactory());
```
Jackie-Jiang: (nit) add an empty line after this

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: I prefer calling `pinotHelixResourceManager.getHelixAdmin()`, but you can decide

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
Jackie-Jiang: But since LLC is done, we can (and we should) always update state model right? Also, the comments need to be updated.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
Jackie-Jiang: I'm still confused. Why setting to 0 instead of CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_REPLICA_COUNT?

jackjlli: This config guarantees if active number of replicas is no less than minimum active replica, there will not be partition movements happened.
If min active replica is set to `LEAD_CONTROLLER_RESOURCE_REPLICA_COUNT`, it forces Helix controller to keep calculating the assignment all the time.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: It seems it's unused. Removed it.

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
jackjlli: Updated.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
mcvsubbu: It is updated under a config. We can ignore the config and update it all the time, if that is what you are thinking about. 
I suggest we keep the config, and use it only to update state model and can be used in future as well in case we do so (unlikely, i admit)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -53,126 +56,214 @@
...
  */
 public class HelixSetupUtils {
-
   private static final Logger LOGGER = LoggerFactory.getLogger(HelixSetupUtils.class);
 
   public static synchronized HelixManager setup(String helixClusterName, String zkPath,
-      String pinotControllerInstanceId, boolean isUpdateStateModel, boolean enableBatchMessageMode) {
-
+      String helixControllerInstanceId) {
     try {
```
Jackie-Jiang: No need for try-catch. The caller is responsible for logging the exception

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -53,126 +56,214 @@
...
-      createHelixClusterIfNeeded(helixClusterName, zkPath, isUpdateStateModel, enableBatchMessageMode);
+      setupHelixCluster(helixClusterName, zkPath);
     } catch (final Exception e) {
-      LOGGER.error("Caught exception", e);
-      return null;
+      LOGGER.error("Caught exception when setting up Helix cluster: {}", helixClusterName, e);
+      throw e;
     }
 
     try {
```
Jackie-Jiang: Same here

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
```diff
@@ -41,8 +46,7 @@ public void setUp() {
   }
 
   @Test
-  public void testHelixOnlyController()
-      throws Exception {
+  public void testHelixOnlyController() throws Exception {
```
Jackie-Jiang: Please update IDE to use Pinot Style, and reformat all the changes

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
```diff
@@ -57,25 +61,42 @@ public void testHelixOnlyController()
...
     config.setControllerMode(ControllerConf.ControllerMode.DUAL);
     config.setControllerPort(Integer.toString(Integer.parseInt(config.getControllerPort()) + controllerPortOffset++));
 
+    // Helix cluster will be set up when starting the first controller.
     startController(config);
     TestUtils.waitForCondition(aVoid -> _helixManager.isConnected(), TIMEOUT_IN_MS,
         "Failed to start " + config.getControllerMode() + " controller in " + TIMEOUT_IN_MS + "ms.");
     Assert.assertEquals(_controllerStarter.getControllerMode(), ControllerConf.ControllerMode.DUAL);
 
+    // Enable the lead controller resource.
```
Jackie-Jiang: Do we need to explicitly enable the resource?

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
```diff
@@ -19,6 +19,7 @@
 package org.apache.pinot.common.config;
 
 import org.apache.pinot.common.exception.InvalidConfigException;
+import org.apache.pinot.common.utils.CommonConstants;
```
Jackie-Jiang: Not used import? Please reformat all the changed files

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
jackjlli: By default it's disabled when creating a Helix cluster. In tests we'd like to explicitly enable it. That's why we need this line in the tests.

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
jackjlli: Deleted.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
jackjlli: Reformatted.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Removed.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
jackjlli: Removed.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
Jackie-Jiang: **(CRITICAL)** Please double check this parameter. I think by setting this, when a controller is down, no new master will be selected.

jackjlli: @Jackie-Jiang Added code to explicitly set delay to 5 mins. And added comments on how this works in the latest push.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -46,6 +46,8 @@
     // https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot
     public static final int NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE = 24;
     public static final int LEAD_CONTROLLER_RESOURCE_REPLICA_COUNT = 1;
+    public static final int MINIMUM_NUMBER_OFACTIVE_REPLICAS = 0;
```
Jackie-Jiang: MIN_ACTIVE_REPLICAS and REBALANCE_DELAY_MS is better?

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
```diff
@@ -39,24 +38,23 @@ private static String buildBrokerTenantTagFromTenantName(String tenantName) {
   }
 
   public static boolean hasValidServerTagSuffix(String tagName) {
```
Jackie-Jiang: This can be replaced with isServerTag()

PR #4051, opened 2019-04-02
Fix the race condition for real-time inverted index reader

Jackie-Jiang:
When getting doc Ids from RealtimeInvertedIndexReader, the given
dict Id might not be added to the inverted index yet. We first add
the value to the dictionary. Before the value is added to the
inverted index, the query might have a predicate which matches the
newly added value. In that case, the given dictionary Id does not
exist in the inverted index, and we return an empty bitmap.

Simplify BitmapBasedFilterOperator because getDocIds() never
returns null.
Add a test for the change.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/filter/BitmapBasedFilterOperator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/filter/BitmapBasedFilterOperator.java
@@ -25,12 +25,9 @@ import org.apache.pinot.core.operator.docidsets.BitmapDocIdSet;
 import org.apache.pinot.core.operator.filter.predicate.PredicateEvaluator;
 import org.apache.pinot.core.segment.index.readers.InvertedIndexReader;
 import org.roaringbitmap.buffer.ImmutableRoaringBitmap;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 
 public class BitmapBasedFilterOperator extends BaseFilterOperator {
-  private static final Logger LOGGER = LoggerFactory.getLogger(BitmapBasedFilterOperator.class);
   private static final String OPERATOR_NAME = "BitmapBasedFilterOperator";
 
   private final PredicateEvaluator _predicateEvaluator;
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/filter/BitmapBasedFilterOperator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/filter/BitmapBasedFilterOperator.java
@@ -19,8 +19,6 @@
 package org.apache.pinot.core.operator.filter;
 
 import com.google.common.base.Preconditions;
-import java.util.ArrayList;
-import java.util.List;
 import org.apache.pinot.core.common.DataSource;
 import org.apache.pinot.core.operator.blocks.FilterBlock;
 import org.apache.pinot.core.operator.docidsets.BitmapDocIdSet;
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/org/apache/pinot/core/operator/filter/BitmapBasedFilterOperator.java
```diff
@@ -78,27 +76,14 @@ protected FilterBlock getNextBlock() {
...
-    // filter out null bitmaps
     InvertedIndexReader invertedIndex = _dataSource.getInvertedIndex();
     int length = dictIds.length;
-    List<ImmutableRoaringBitmap> bitmaps = new ArrayList<>(length);
-    for (int dictId : dictIds) {
-      ImmutableRoaringBitmap bitmap = (ImmutableRoaringBitmap) invertedIndex.getDocIds(dictId);
-      if (bitmap != null) {
-        bitmaps.add(bitmap);
-      }
+    ImmutableRoaringBitmap[] bitmaps = new ImmutableRoaringBitmap[length];
```
snleee: +1 for this change..

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/invertedindex/RealtimeInvertedIndexReader.java
```diff
@@ -60,6 +60,13 @@ public MutableRoaringBitmap getDocIds(int dictId) {
     ThreadSafeMutableRoaringBitmap bitmap;
     try {
       _readLock.lock();
+      // NOTE: the given dict Id might not be added to the inverted index yet. We first add the value to the dictionary.
+      // Before the value is added to the inverted index, the query might have a predicate which matches the newly added
+      // value. In that case, the given dictionary Id does not exist in the inverted index, and we return an empty
+      // bitmap.
+      if (_bitmaps.size() == dictId) {
```
kishoreg: I don’t understand this check. This is still prone to error rt. Imagine two new dict ids added

Also just thinking out loud. the previous behavior was probably better than returning empty bitmap?  What happens in Not In case.

snleee: Previous behavior caused `IndexOutOfBounds` error when the race condition happened (the query hits when the value got added to dictionary but not yet in the inverted index). I thought that it's only possible to meet the case where `bitmap.size() == dictId` since the consumption is happening in a single thread. @Jackie-Jiang can explain it in more detail and also comment on `not in` case.

Jackie-Jiang: @kishoreg Actually you are right for multi-valued column. I'll change this to `_bitmaps.size() <= dictId` to handle all cases.
In the ancient code, we might return null, but then we refactored the code and uniformed the interface. Right now the interface does not expect null return value, and this is corner case which happens rarely, so I would rather handle this inside the implementation by returning an empty bitmap instead of null.

kishoreg: cool.

Another thought I had was, should we even check for the _bitmaps.size() <= dictId. What if we return empty one if the bitmap is null.

kishoreg: Ignore the last comment. I forgot to refresh the PR and was looking at old code.

PR #4053, opened 2019-04-02
[TE] Speed up minute level detection

xiaohui-sun:
1. Speed up minute level detection
If detection window >= 1 day, set the bucketPeriod, windowSize, windowUnit to 1 day.

2. Change EARLY_TERMINATE_WINDOW from 10 to 5 to fail early.

3. Remove pre-cache for detection window for two reasons:
a. We need to load the data anyway so pre-cache doesn't help.
b. If the detection consistently fail and triggered early-terminate then pre-cache everything is a waste of time.


xiaohui-sun:
> I do not see any code removing pre-cache as mentioned in the description!

This part
        // pre cache the time series for the whole detection time period instead of fetching for each window	
        if (this.cachingPeriodLookback >= 0) {	
          MetricSlice cacheSlice =	
              MetricSlice.from(this.metricEntity.getId(), startTime - cachingPeriodLookback, endTime,	
                  this.metricEntity.getFilters(), toTimeGranularity(this.bucketPeriod));	
          this.provider.fetchTimeseries(Collections.singleton(cacheSlice));	
        }

PR #4055, opened 2019-04-02
 Handle exceptions during teardown

sunithabeeram:
We see random integration test failures in SegmentStatusCheckerIntegrationTest.tearDown
with the trace:  
java.lang.NullPointerException
at org.apache.helix.participant.HelixStateMachineEngine.reset(HelixStateMachineEngine.java:156)
at org.apache.helix.messaging.handling.HelixTaskExecutor.reset(HelixTaskExecutor.java:602)
at org.apache.helix.messaging.handling.HelixTaskExecutor.shutdown(HelixTaskExecutor.java:1128)
at org.apache.helix.manager.zk.ZKHelixManager.disconnect(ZKHelixManager.java:701)
at org.apache.pinot.server.starter.helix.HelixServerStarter.stop(HelixServerStarter.java:362)
at org.apache.pinot.integration.tests.ClusterTest.stopServer(ClusterTest.java:227)
at org.apache.pinot.integration.tests.controller.periodic.tasks.SegmentStatusCheckerIntegrationTest.tearDown(SegmentStatusCheckerIntegrationTest.java:282)

Its unclear if this is due to an issue in Helix's task framework that has been fixed in subsequent releases. However, this does point to the case where the teardown might abort due to an exception (we attempt to stop broker, server and minion) and an unclean shutdown *may* cause other issues.

This change just  handles exceptions so we can still proceed with the rest of the cleanup.

### DIFF
--- a/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterTest.java
+++ b/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterTest.java
@@ -243,7 +243,7 @@ public abstract class ClusterTest extends ControllerTest {
       try {
         minionStarter.stop();
       } catch (Exception e) {
-        LOGGER.error("Encountered exception while stopping server {}", e.getMessage());
+        LOGGER.error("Encountered exception while stopping minion {}", e.getMessage());
       }
     }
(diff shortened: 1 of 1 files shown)


Jackie-Jiang:
Can you please create an issue/ticket tracking this? Currently we simply ignore the exception during the tear down, which is not the purpose of the test.

PR #4059, opened 2019-04-03
[TE] Improve logging messages; put default ref links only when empty

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineScheduler.java
```diff
@@ -82,7 +82,7 @@ public void run() {
 
         try {
           if (scheduler.checkExists(key)) {
-            LOG.info("Detection config  " + key + " is already scheduled.");
+            LOG.info("Detection config  " + key.getName() + " is already scheduled");
```
xiaohui-sun: Can you keep both? Having key is easier to debug. E.g, you can quickly look up key in database.

akshayrai: This is what is being logged today.
`Detection config  DETECTION.DETECTION_107451840 is already scheduled`
After this change, you will see,
`Detection config  DETECTION_107451840 is already scheduled`

PR #4063, opened 2019-04-03
Introduce interface stability and audience annotations

sunithabeeram:
Implementation for https://github.com/apache/incubator-pinot/issues/4031. I have annotated couple of interfaces as an example. Once this is merged, will review rest of the interfaces and mark them. The main thing I would like to accomplish is to call out "external" APIs with InterfaceAudience.public (may not annotate all interfaces as Private should be assumed where the annotation is missing).

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
+++ b/pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
@@ -27,13 +27,13 @@ import java.lang.annotation.Documented;
  * Currently the audience can be {@link org.apache.pinot.annotations.InterfaceAudience.Public},
  * {@link org.apache.pinot.annotations.InterfaceAudience.LimitedPrivate} or
  * {@link org.apache.pinot.annotations.InterfaceAudience.Private}. <br>
- * All public classes must have InterfaceAudience annotation. <br>
+ * It is recommended that public classes are annotated with InterfaceAudience annotation. <br>
  * <ul>
  * <li>Public classes that are not marked with this annotation must be
  * considered by default as {@link org.apache.pinot.annotations.InterfaceAudience.Private}.</li>
  *
  * <li>External applications must only use classes that are marked
- * {@link org.apache.pinot.annotations.InterfaceAudience.Public}. Avoid using non public classes as these classes
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Public}. Do not depend on non-public classes as these classes
  * could be removed or change in incompatible ways.</li>
  *
  * <li> Methods may have a different annotation that it is more restrictive
--- a/pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
+++ b/pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
@@ -27,13 +27,14 @@ import java.lang.annotation.Documented;
  * Currently the audience can be {@link org.apache.pinot.annotations.InterfaceAudience.Public},
  * {@link org.apache.pinot.annotations.InterfaceAudience.LimitedPrivate} or
  * {@link org.apache.pinot.annotations.InterfaceAudience.Private}. <br>
- * It is recommended that public classes are annotated with InterfaceAudience annotation. <br>
+ *
  * <ul>
  * <li>Public classes that are not marked with this annotation must be
  * considered by default as {@link org.apache.pinot.annotations.InterfaceAudience.Private}.</li>
  *
- * <li>External applications must only use classes that are marked
- * {@link org.apache.pinot.annotations.InterfaceAudience.Public}. Do not depend on non-public classes as these classes
+ * <li>External application developers depending on Pinot must only use classes that are marked
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Public}. Do not depend on classes without an
+ * explicit InterfaceAudience.Public annotation as these classes
  * could be removed or change in incompatible ways.</li>
  *
  * <li> Methods may have a different annotation that it is more restrictive
--- a/pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
+++ b/pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
@@ -56,7 +56,7 @@ public class InterfaceAudience {
 
   /**
    * Intended only for the project(s) specified in the annotation.
-   * No specific uses for this exist currently.
+   * Reserved for future use.
    */
   @Documented
   public @interface LimitedPrivate {
--- a/pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
+++ b/pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
@@ -1,0 +1,72 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.annotations;
+
+import java.lang.annotation.Documented;
+
+
+/**
+ * Annotation to inform users of a package, class or method's intended audience.
+ *
+ * Currently the audience can be {@link org.apache.pinot.annotations.InterfaceAudience.Public},
+ * {@link org.apache.pinot.annotations.InterfaceAudience.LimitedPrivate} or
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Private}. <br>
+ * All public classes must have InterfaceAudience annotation. <br>
+ * <ul>
+ * <li>Public classes that are not marked with this annotation must be
+ * considered by default as {@link org.apache.pinot.annotations.InterfaceAudience.Private}.</li>
+ *
+ * <li>External applications must only use classes that are marked
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Public}. Avoid using non public classes as these classes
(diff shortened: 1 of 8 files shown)


pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
```diff
@@ -0,0 +1,72 @@
...
+import java.lang.annotation.Documented;
+
+
+/**
+ * Annotation to inform users of a package, class or method's intended audience.
+ *
+ * Currently the audience can be {@link org.apache.pinot.annotations.InterfaceAudience.Public},
+ * {@link org.apache.pinot.annotations.InterfaceAudience.LimitedPrivate} or
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Private}. <br>
+ * All public classes must have InterfaceAudience annotation. <br>
```
mcvsubbu: why "must"? That will mean we go add this annotation to pretty much every class in Pinot?

pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
```diff
@@ -0,0 +1,72 @@
...
+ * <ul>
+ * <li>Public classes that are not marked with this annotation must be
+ * considered by default as {@link org.apache.pinot.annotations.InterfaceAudience.Private}.</li>
+ *
+ * <li>External applications must only use classes that are marked
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Public}. Avoid using non public classes as these classes
```
mcvsubbu: s/Avoid/Do NOT/
You can also add "If you see a need to use a class/interface/method that is not marked as Public, submit a PR to mark it so (or re-factor the code in other ways to get the information you want)"

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/MessageBatch.java
```diff
@@ -18,11 +18,17 @@
...
+import org.apache.pinot.annotations.InterfaceStability;
+
+
 /**
  * Interface wrapping stream consumer. Throws IndexOutOfBoundsException when trying to access a message at an
  * invalid index.
  * @param <T>
  */
+@InterfaceAudience.Public
+@InterfaceStability.Stable
```
mcvsubbu: I would call this evolving at this time, until we get another stream in place.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/PartitionLevelConsumer.java
```diff
@@ -19,11 +19,15 @@
...
 import java.io.Closeable;
+import org.apache.pinot.annotations.InterfaceAudience;
+import org.apache.pinot.annotations.InterfaceStability;
 
 
 /**
  * Interface for a consumer which fetches messages at the partition level of a stream, for given offsets
  */
+@InterfaceAudience.Public
+@InterfaceStability.Stable
```
mcvsubbu: Same

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamLevelConsumer.java
```diff
@@ -18,12 +18,16 @@
...
+import org.apache.pinot.annotations.InterfaceAudience;
+import org.apache.pinot.annotations.InterfaceStability;
 import org.apache.pinot.core.data.GenericRow;
 
 
 /**
  * Interface for a consumer that consumes at stream level and is unaware of any partitions of the stream
  */
+@InterfaceAudience.Public
+@InterfaceStability.Stable
```
mcvsubbu: Same

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMessageDecoder.java
```diff
@@ -27,6 +29,8 @@
  * Interface for a decoder of messages fetched from the stream
  * @param <T>
  */
+@InterfaceAudience.Public
+@InterfaceStability.Stable
```
mcvsubbu: Same

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamMetadataProvider.java
```diff
@@ -20,11 +20,15 @@
...
 import javax.annotation.Nonnull;
+import org.apache.pinot.annotations.InterfaceAudience;
+import org.apache.pinot.annotations.InterfaceStability;
 
 
 /**
  * Interface for provider of stream metadata such as partition count, partition offsets
  */
+@InterfaceAudience.Public
+@InterfaceStability.Stable
```
mcvsubbu: Same

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/MessageBatch.java
sunithabeeram: Based on this definition of stable:

"Can evolve while retaining compatibility for minor release boundaries.;
   * can break compatibility only at major release (ie. at m.0)."

and the fact that we expect external users to extend this interface, I think we can call it stable. Evolving annotation would be used for anything we are still figuring out - for ex, a first draft of an interface.

pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
sunithabeeram: Updated.

pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
sunithabeeram: Updated.

sunithabeeram:
Travis failure is unrelated. Following up with @snleee to see if the newly introduced travis wait is able to provide more info.

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/MessageBatch.java
mcvsubbu: agreed. This is a note to ourselves that since there may be users who have implemented this interface but are not able to contribute the code here for whatever reason, and we don't want to break their code as  best as we can

pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
```diff
@@ -0,0 +1,72 @@
...
+ * Currently the audience can be {@link org.apache.pinot.annotations.InterfaceAudience.Public},
+ * {@link org.apache.pinot.annotations.InterfaceAudience.LimitedPrivate} or
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Private}. <br>
+ * It is recommended that public classes are annotated with InterfaceAudience annotation. <br>
+ * <ul>
+ * <li>Public classes that are not marked with this annotation must be
+ * considered by default as {@link org.apache.pinot.annotations.InterfaceAudience.Private}.</li>
+ *
+ * <li>External applications must only use classes that are marked
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Public}. Do not depend on non-public classes as these classes
```
mcvsubbu: I may change "non-public classes" to "classes not marked thus" (the word public here is too generic)

pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceAudience.java
```diff
@@ -0,0 +1,72 @@
...
+import java.lang.annotation.Documented;
+
+
+/**
+ * Annotation to inform users of a package, class or method's intended audience.
+ *
+ * Currently the audience can be {@link org.apache.pinot.annotations.InterfaceAudience.Public},
+ * {@link org.apache.pinot.annotations.InterfaceAudience.LimitedPrivate} or
+ * {@link org.apache.pinot.annotations.InterfaceAudience.Private}. <br>
+ * It is recommended that public classes are annotated with InterfaceAudience annotation. <br>
```
mcvsubbu: I recommend removing this sentence altogether. If the interfaceaudience notation is missing, do not use it. End of story. A recommendation here means ideally we would like to have _all_ classes marked with this annotation. I wold stay away from that

pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceStability.java
```diff
@@ -0,0 +1,60 @@
...
+ * class or method not changing over time. Currently the stability can be
+ * {@link org.apache.pinot.annotations.InterfaceStability.Stable},
+ * {@link org.apache.pinot.annotations.InterfaceStability.Evolving} or
+ * {@link org.apache.pinot.annotations.InterfaceStability.Unstable}. <br>
+ *
+ * <ul><li>All classes that are annotated with {@link org.apache.pinot.annotations.InterfaceAudience.Public} or
+ * {@link org.apache.pinot.annotations.InterfaceAudience.LimitedPrivate} must have InterfaceStability annotation. </li>
+ * <li>Classes that are {@link org.apache.pinot.annotations.InterfaceAudience.Private} are to be considered unstable unless
+ * a different InterfaceStability annotation states otherwise.</li>
+ * <li>Incompatible changes must not be made to classes marked as stable.</li>
```
mcvsubbu: Suggest in active voice: "Developers must not make incompatible changes to classes marked as stable"

pinot-common/src/main/java/org/apache/pinot/annotations/InterfaceStability.java
```diff
@@ -0,0 +1,60 @@
...
+ * Annotation to inform users of how much to rely on a particular package,
+ * class or method not changing over time. Currently the stability can be
+ * {@link org.apache.pinot.annotations.InterfaceStability.Stable},
+ * {@link org.apache.pinot.annotations.InterfaceStability.Evolving} or
+ * {@link org.apache.pinot.annotations.InterfaceStability.Unstable}. <br>
+ *
+ * <ul><li>All classes that are annotated with {@link org.apache.pinot.annotations.InterfaceAudience.Public} or
+ * {@link org.apache.pinot.annotations.InterfaceAudience.LimitedPrivate} must have InterfaceStability annotation. </li>
+ * <li>Classes that are {@link org.apache.pinot.annotations.InterfaceAudience.Private} are to be considered unstable unless
+ * a different InterfaceStability annotation states otherwise.</li>
```
mcvsubbu: Can you think of a use case where we may have a private interface and mark it 'stable'?

sunithabeeram: This wouldn't be typical, but may be possible where internal to pinot code, if a contributor wants to indicate the status of the stability of a class they would have an option to annotate it so.

PR #4064, opened 2019-04-03
[TE] Deprecate and remove thirdeye-hadoop module

akshayrai:


xiaohui-sun:
Can you hold the deletion for now? I am looking into how this is used in UMP. We need a plan to retire them completely.

PR #4065, opened 2019-04-03
Exit 1 when caught exception in Pinot Admin command.

fx19880617:
For better integration with scripting languages.

### DIFF
--- a/pinot-tools/src/main/java/org/apache/pinot/tools/admin/PinotAdministrator.java
+++ b/pinot-tools/src/main/java/org/apache/pinot/tools/admin/PinotAdministrator.java
@@ -81,8 +81,7 @@ public class PinotAdministrator {
     return _status;
   }
 
-  public void execute(String[] args)
-      throws Exception {
+  public void execute(String[] args) {
     try {
       CmdLineParser parser = new CmdLineParser(this);
       parser.parseArgument(args);
@@ -102,10 +101,10 @@ public class PinotAdministrator {
     }
   }
 
-  public static void main(String[] args)
-      throws Exception {
+  public static void main(String[] args) {
     PinotAdministrator pinotAdministrator = new PinotAdministrator();
     pinotAdministrator.execute(args);
+    System.exit(pinotAdministrator.getStatus() ? 0 : 1);
   }
 
   public void printUsage() {
--- a/pinot-tools/src/main/java/org/apache/pinot/tools/admin/PinotAdministrator.java
+++ b/pinot-tools/src/main/java/org/apache/pinot/tools/admin/PinotAdministrator.java
@@ -104,7 +104,9 @@ public class PinotAdministrator {
   public static void main(String[] args) {
     PinotAdministrator pinotAdministrator = new PinotAdministrator();
     pinotAdministrator.execute(args);
-    System.exit(pinotAdministrator.getStatus() ? 0 : 1);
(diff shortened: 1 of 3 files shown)


pinot-tools/src/main/java/org/apache/pinot/tools/admin/PinotAdministrator.java
```diff
@@ -102,10 +102,13 @@ public void execute(String[] args)
...
 
-  public static void main(String[] args)
-      throws Exception {
+  public static void main(String[] args) {
     PinotAdministrator pinotAdministrator = new PinotAdministrator();
-    pinotAdministrator.execute(args);
+    try {
+      pinotAdministrator.execute(args);
+    } catch (Exception e) {
+      System.exit(1);
```
mayankshriv: May be log the message/exception?

fx19880617: Actually the exception will be caught in execute() method. Will let this method do logging, and let main() function check status and do exit.

mcvsubbu:
Last time I added an exit to this class, all sorts of things broke. It was fixed in this commit
a5ea0bd75fa07cb59de6d0ea5b36bb33bab56aad

fx19880617:
> Last time I added an exit to this class, all sorts of things broke. It was fixed in this commit
> [a5ea0bd](https://github.com/apache/incubator-pinot/commit/a5ea0bd75fa07cb59de6d0ea5b36bb33bab56aad)

lol.
Found that PinotAdmin cmds are directly called in ChaosMonkeyIntegrationTest. Let me think how to bypass this.

mayankshriv:
One way to address this could be how PerfBenchmarkDriver starts the cluster without the PinotAdmin.

fx19880617:
I thought it’s intentional to have ChaosMonkey test here to start processes for sending signal? 

I added a check for system properties, so by default it will always trigger system.exit. But not if we explicitly set it in tests.

jackjlli:
@fx19880617 Running zookeeper doesn't work after committing this changes.

This fix only works for **single thread** use case though.

You can re-produce the failure by running `./pinot-admin.sh StartZookeeper`.

kishoreg:
let's revert this change?. We need to think through this. @mcvsubbu had a good suggestion to have a -daemon flag for pinot-admin.sh

fx19880617:
sure, let's revert this, I will fix this later.

fx19880617:
Another way is that we still keep this env variable, but only do System.exit() for tooling scripts, like AddSchema/AddTable, for them, the exit code is important to script those tooling.

kishoreg:
How about setting the default to false and set the env variable for the cases where we know it needs to exit with proper code?

mayankshriv:
We don't typically use env variables for the rest of the system. Why not just use a boolean field (with a desired default value), and callers (admin tools, or integ tests) can set the field as needed?

mcvsubbu:
Not sure what prompted the checkin. What was the use case that needed exit?

In my case, when I tried this last time, I was just trying to be a correct UNIX command that exits with 0 if successful, 1 otherwise. This would help (I thought) write scripts, so that (for example) we could start up multiple components in independent jvms. bind them to differnt ports, and have a cluster set up.

I suggest that we decide this on a per-command basis. e.g. StarZookeeper should jsut run as a daemon. No choice. So, if we set an appropriate variable while coming in from pinot-admin.sh we should be good? Otherwise, a -daemon flag will suffice, but it will funny to add a flag for the obvious  (e.g. startzk)

PR #4069, opened 2019-04-03
Revert "add support for parquet reader"

snleee:
Reverts apache/incubator-pinot#3852


sunithabeeram:
Can you note why this is being reverted?

PR #4072, opened 2019-04-04
[TE] detection - handle day light savings time in moving windows

jihaozh:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -151,7 +152,7 @@ public DetectionPipelineResult run() throws Exception {
     // 1. get the last time stamp for the time series.
     // 2. to calculate current values and  baseline values for the anomalies detected
     // 3. anomaly detection current and baseline time series value
-    if( this.cachingPeriodLookback > 0) {
+    if (this.cachingPeriodLookback >= 0) {
```
xiaohui-sun: Why >= 0? If it is zero then there is no need to cache.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
```diff
@@ -309,30 +309,30 @@ private void fillInDetectorWrapperProperties(Map<String, Object> properties, Map
...
+    // override from yaml
+    if (yamlConfig.containsKey(PROP_WINDOW_SIZE)) {
+      properties.put(PROP_MOVING_WINDOW_DETECTION, true);
+      properties.put(PROP_WINDOW_SIZE, MapUtils.getString(yamlConfig, PROP_WINDOW_SIZE));
+    }
+    if (yamlConfig.containsKey(PROP_WINDOW_UNIT)) {
+      properties.put(PROP_MOVING_WINDOW_DETECTION, true);
+      properties.put(PROP_WINDOW_UNIT, MapUtils.getString(yamlConfig, PROP_WINDOW_UNIT));
+    }
+    if (yamlConfig.containsKey(PROP_WINDOW_DELAY)) {
```
xiaohui-sun: Currently there is a mix of three time units in YAML now and is confusing.
1. Milliseconds. E.g, max duration time.
2. Unit and value. E.g, window_deplay and window_delay_unit.
3. Period. E.g, bucket period.
Do we have plan to unify them?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionUtils.java
```diff
@@ -219,4 +221,20 @@ public static Period getMonitoringGranularityPeriod(String monitoringGranularity
     return TimeGranularity.fromString(monitoringGranularity).toPeriod();
   }
 
+  public static Period periodFromTimeUnit(int size, TimeUnit unit) {
+    switch (unit) {
+      case DAYS:
+        return Period.days(size);
+      case HOURS:
+        return Period.hours(size);
+      case MILLISECONDS:
```
xiaohui-sun: Can you put MILLISECONDS to the end. Then the case is ordered by size.

xiaohui-sun: Or remove MILLISECONDS since it is default value.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -48,12 +48,14 @@
...
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+import org.joda.time.Duration;
 import org.joda.time.Interval;
 import org.joda.time.Period;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static org.apache.pinot.thirdeye.dataframe.util.DataFrameUtils.*;
+import static org.apache.pinot.thirdeye.detection.DetectionUtils.*;
```
xiaohui-sun: This is not LinkedIn coding convention.
https://iwww.corp.linkedin.com/wiki/cf/display/ENGS/Java+Coding+Conventions#JavaCodingConventions-2.5.1AvoidStarImports

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
jihaozh: Not necessarily. If it's 0, i.e. the cache look back is 0, which means it only caches current time series. If it's negative, then it means the cache is turned off.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
jihaozh: Agree. It's confusing now. The correct approach is to use Period. However, with the legacy notations already stored in config files, it's hard to eliminate them. We can come up with a way to use Period only in the future.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionUtils.java
jihaozh: Sure.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
jihaozh: Thanks for pointing out. But this is static utility classes, so I assume it's ok as said in the doc?

jihaozh: Cleaned up

PR #4074, opened 2019-04-04
Always enable default column feature, remove the configuration

Jackie-Jiang:
The configuration is introduced to test the default column feature.
Since it has been tested thoroughly for years, remove the
configuration and always enable it.


pinot-server/src/test/java/org/apache/pinot/server/integration/realtime/RealtimeTableDataManagerTest.java
```diff
@@ -59,6 +59,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
```
jackjlli: This import never gets used though. 
But since you're gonna delete this test, it should be good here.

PR #4075, opened 2019-04-04
Remove stale test RealtimeTableDataManagerTest

Jackie-Jiang:
RealtimeTableDataManagerTest is a stale test class and does not have any test enabled
The test itself does not reflect to the data file, and the functionality is covered by the RealtimeClusterIntegrationTest


Jackie-Jiang:
> If I understand right, this test is covering mutable segment's ability to add new rows and serving queries out of it. That aspedct is not covered by RealtimeClusterIntegrationTest

As long as I see, this test does not even add new rows into the data manager, and does not have any assertions. It's just a piece of legacy code left over and does not reflect the data set. I don't think it is worth to be fixed.

PR #4078, opened 2019-04-04
Reduce logging in controller

npawar:
Reducing some long/redundant logs
1) Removing log from retention manager which prints all segments, as it is done again in PinotHelixResourceManager::deleteSegments
2) Reducing the number of segments printed in OfflineSegmentIntervalChecker
3) Reducing the number of segments printed in TableSizeReader

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/util/TableSizeReader.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/util/TableSizeReader.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executor;
+import java.util.stream.Collectors;
 import javax.annotation.Nonnegative;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -47,6 +48,8 @@ import org.slf4j.LoggerFactory;
  */
 public class TableSizeReader {
   private static final Logger LOGGER = LoggerFactory.getLogger(TableSizeReader.class);
+
+  private static final int MAX_MISSING_SEGMENTS_TO_LIST = 5;
   private final Executor _executor;
   private final HttpConnectionManager _connectionManager;
   private final PinotHelixResourceManager _helixResourceManager;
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/util/TableSizeReader.java
```diff
@@ -225,14 +228,13 @@ public TableSubTypeSizeDetails getTableSubtypeSize(String tableNameWithType, int
...
-        LOGGER.warn("Failed to get size report for all {} segments: {} for table: {}", numSegments, missingSegments,
-            tableNameWithType);
+        LOGGER.warn("Failed to get size report for all {} segments of table: {}", numSegments, tableNameWithType);
         subTypeSizeDetails.reportedSizeInBytes = -1;
         subTypeSizeDetails.estimatedSizeInBytes = -1;
       } else {
-        LOGGER
-            .warn("Missing size report for {} out of {} segments: {} for table {}", subTypeSizeDetails.missingSegments,
-                numSegments, missingSegments, tableNameWithType);
+        LOGGER.warn("Missing size report for {} out of {} segments for table {}. Listing up to {} segments: {}",
```
sunithabeeram: I am not sure if we really care much about the specific segment names. We can just list how many were missed.

pinot-controller/src/main/java/org/apache/pinot/controller/validation/OfflineSegmentIntervalChecker.java
```diff
@@ -81,18 +83,23 @@ private void validateOfflineSegmentPush(TableConfig tableConfig) {
...
-          segmentsWithInvalidInterval.add(offlineSegmentZKMetadata.getSegmentName());
+          numSegmentsWithInvalidIntervals ++;
+          if (segmentsWithInvalidInterval.size() < MAX_SEGMENTS_WITH_INVALID_INTERVALS_TO_LOG) {
+            segmentsWithInvalidInterval.add(offlineSegmentZKMetadata.getSegmentName());
+          }
         }
       }
       if (!segmentsWithInvalidInterval.isEmpty()) {
-        LOGGER.warn("Table: {} has segments with invalid interval: {}", offlineTableName, segmentsWithInvalidInterval);
+        LOGGER.warn("Table: {} has {} segments with invalid interval. Listing up to {}: {}", offlineTableName,
```
sunithabeeram: If there is an admin tool one can use to figure out the segments that have invalid interval, we can use that to find out the ones that are bad and just log the segment count here. (If one doesn't exist we can build one).

npawar: hmm, that's a good idea. Periodically printing this is not really helping anyway. Will add that as a followup.

pinot-controller/src/main/java/org/apache/pinot/controller/util/TableSizeReader.java
npawar: Changed.

pinot-controller/src/main/java/org/apache/pinot/controller/validation/OfflineSegmentIntervalChecker.java
```diff
@@ -81,18 +81,19 @@ private void validateOfflineSegmentPush(TableConfig tableConfig) {
...
             .timeValueInValidRange(timeInterval.getEndMillis())) {
           segmentIntervals.add(timeInterval);
         } else {
-          segmentsWithInvalidInterval.add(offlineSegmentZKMetadata.getSegmentName());
+          numSegmentsWithInvalidIntervals ++;
         }
       }
-      if (!segmentsWithInvalidInterval.isEmpty()) {
-        LOGGER.warn("Table: {} has segments with invalid interval: {}", offlineTableName, segmentsWithInvalidInterval);
+      if (numSegmentsWithInvalidIntervals > 0) {
```
mcvsubbu: I would suggest to log the list here if the list is small, otherwise just the number of segments that have invalid intervals. A pinot-admin utility to list segments with invalid intervals will be a useful debug tool in case pinot admins  get questions from table owners.

PR #4081, opened 2019-04-05
Support legacy time type in SegmentValidationAndRetentionConfig

Jackie-Jiang:
Support legacy time type 'daysSinceEpoch', 'hoursSinceEpoch', 'minutesSinceEpoch' and 'secondsSinceEpoch'
Change SegmentsValidationAndRetentionConfig.getTimeType() to directly return TimeUnit instead of String

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/SegmentsValidationAndRetentionConfig.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/SegmentsValidationAndRetentionConfig.java
@@ -21,7 +21,9 @@ package org.apache.pinot.common.config;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import java.lang.reflect.Field;
+import java.util.concurrent.TimeUnit;
 import org.apache.pinot.common.utils.EqualityUtils;
+import org.apache.pinot.common.utils.time.TimeUtils;
 import org.apache.pinot.startree.hll.HllConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -55,7 +57,7 @@ public class SegmentsValidationAndRetentionConfig {
   private String timeColumnName;
 
   @ConfigKey(value = "timeType")
-  private String timeType;
+  private TimeUnit _timeType;
 
   @ConfigKey(value = "segmentAssignmentStrategy")
   private String segmentAssignmentStrategy;
@@ -86,12 +88,12 @@ public class SegmentsValidationAndRetentionConfig {
     this.timeColumnName = timeColumnName;
   }
 
-  public String getTimeType() {
-    return timeType;
+  public TimeUnit getTimeType() {
+    return _timeType;
   }
 
   public void setTimeType(String timeType) {
-    this.timeType = timeType;
+    _timeType = timeType != null ? TimeUtils.timeUnitFromString(timeType) : null;
   }
 
   public String getRetentionTimeUnit() {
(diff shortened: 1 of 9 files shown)


pinot-common/src/main/java/org/apache/pinot/common/config/SegmentsValidationAndRetentionConfig.java
```diff
@@ -86,12 +88,17 @@ public void setTimeColumnName(String timeColumnName) {
...
   }
 
-  public String getTimeType() {
-    return timeType;
+  public TimeUnit getTimeType() {
+    return _timeType;
+  }
+
+  @JsonIgnore
+  public void setTimeType(TimeUnit timeType) {
```
snleee: I guess this is for completeness? (seems to be not used)

pinot-common/src/test/java/org/apache/pinot/common/utils/UtilsTest.java
```diff
@@ -24,48 +24,53 @@
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+import static org.testng.Assert.assertEquals;
```
snleee: Is there any benefit on this except we can use `assertEquals` instead of `Assert.assertEquals`?

snleee:
Pls check
```
Failed tests: 
  HelixBrokerStarterTest.testTimeBoundaryUpdate:256 NullPointer
  TimeBoundaryServiceTest.testExternalViewBasedTimeBoundaryService:79 NullPointer
```

pinot-common/src/main/java/org/apache/pinot/common/config/SegmentsValidationAndRetentionConfig.java
Jackie-Jiang: Actually this is the cause of NPE... Removed

pinot-common/src/test/java/org/apache/pinot/common/utils/UtilsTest.java
Jackie-Jiang: No. I just prefer importing static Assert methods because IDE does so when creating test automatically.

pinot-common/src/main/java/org/apache/pinot/common/config/SegmentsValidationAndRetentionConfig.java
```diff
@@ -55,7 +57,7 @@
   private String timeColumnName;
 
   @ConfigKey(value = "timeType")
-  private String timeType;
+  private TimeUnit _timeType;
```
mcvsubbu: Is this compatible with the zookeeper config we store? Will reads work right? Please test manually that the reads of old configs can work correctly.

pinot-core/src/main/java/org/apache/pinot/core/segment/name/NormalizedDateSegmentNameGenerator.java
```diff
@@ -44,11 +44,11 @@
   private final SimpleDateFormat _inputSDF;
 
   public NormalizedDateSegmentNameGenerator(String tableName, @Nullable String segmentNamePrefix,
-      @Nullable String excludeSequenceId, @Nullable String pushType, @Nullable String pushFrequency,
-      @Nullable String timeType, @Nullable String timeFormat) {
+      boolean excludeSequenceId, @Nullable String pushType, @Nullable String pushFrequency, @Nullable TimeUnit timeType,
```
mcvsubbu: This constructor could be used by outside applications? If so, please use a new constructor as needed and translate

pinot-common/src/main/java/org/apache/pinot/common/config/SegmentsValidationAndRetentionConfig.java
Jackie-Jiang: Tested (we have this covered in TableConfigTest)

pinot-core/src/main/java/org/apache/pinot/core/segment/name/NormalizedDateSegmentNameGenerator.java
Jackie-Jiang: I don't believe so, it should only be called by the segment generator to generate the segment name

PR #4087, opened 2019-04-07
Upgrade RoaringBitmap to 0.8.0

xueyumusic:
Current RoaringBitmap is 0.5.10, this PR upgrade RoaringBitmap to 0.8.0 to pick up bug-fixes, new feature and optimize between these new releases.
For `BitmapDocIdSetTest.java`, it minor refactors to direct serialize to ByteBuffer.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/docidsets/BitmapDocIdSet.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/docidsets/BitmapDocIdSet.java
@@ -40,10 +40,7 @@ public class BitmapDocIdSet implements FilterBlockDocIdSet {
       _bitmap = orBitmap;
     } else if (numBitmaps == 1) {
       if (exclusive) {
-        // NOTE: cannot use ImmutableRoaringBitmap.flip() because the library has a bug in that method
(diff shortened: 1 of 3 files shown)


kishoreg:
Is this backward compatible? Can we load an index generated by older version.

richardstartin:
@kishoreg the serialisation format did not change in 0.8.0, it's just a technical optimisation.

xueyumusic:
Thanks a lot for explain, @richardstartin , :+1: . Actually this pr is inspired by your pr of upgrading RoaringBitmap in druid.

kishoreg:
LGTM. others, any concerns?

sunithabeeram:
I didn't check the backward compatibility part (I am assuming we have generated indexes in atleast one of our tests), however, I was recently looking into the performance of a query against one of our tables; It had a huge IN clause and also had a couple of NOT IN clauses following this format:

SELECT count(col) from table where col1 IN "<100+ values>" and col2 NOT IN "<2 values>" and col3 NOT IN "<10+values>" 

The base latency of this query was ~450ms. With only updating to roaringbitmap 0.8.0, I see that the latency drops to ~70ms. Thats pretty neat.

The mileage across other queries might vary.

PR #4089, opened 2019-04-08
In HelixBrokerStarter, allow custom cluster change handlers

Jackie-Jiang:
Motivation: all cluster change listeners should be added through the ClusterChangeMediater in order to perform dedup and proactive check

Add ClusterChangeHandler interface
Add methods in HelixBrokerStarter to allow custom cluster change handlers to be plugged in
Split the responsibility of helix managers on broker side
- Spectator: handles the custom change listeners, properties read/write
- Participant: handles Helix functionality such as state transitions and messages


pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -116,11 +117,18 @@ public void run() {
...
 
-  private void processClusterChange(ChangeType changeType, ClusterChangeHandler changeHandler) {
+  private void processClusterChange(ChangeType changeType, List<ClusterChangeHandler> changeHandlers) {
     long startTime = System.currentTimeMillis();
     LOGGER.info("Start processing {} change", changeType);
-    changeHandler.processClusterChange();
-    long endTime = System.currentTimeMillis();
+    long handlerStartTime = startTime;
+    for (ClusterChangeHandler changeHandler : changeHandlers) {
+      changeHandler.processClusterChange(changeType);
```
sunithabeeram: It might be good to handle exceptions from this handler here so we don't skip others if one has an error.

Jackie-Jiang: Done

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -116,10 +117,20 @@ public void run() {
...
-  private void processClusterChange(ChangeType changeType, ClusterChangeHandler changeHandler) {
+  private void processClusterChange(ChangeType changeType, List<ClusterChangeHandler> changeHandlers) {
     long startTime = System.currentTimeMillis();
     LOGGER.info("Start processing {} change", changeType);
-    changeHandler.processClusterChange();
+    for (ClusterChangeHandler changeHandler : changeHandlers) {
+      try {
+        long handlerStartTime = System.currentTimeMillis();
+        changeHandler.processClusterChange(changeType);
+        LOGGER.info("Finish handling {} change for handler: {} in {}ms", changeType, changeHandler.getClass().getName(),
```
mcvsubbu: I would introduce a method in ClusterChangeHandler to get the name.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeHandler.java
```diff
@@ -18,10 +18,16 @@
...
+import org.apache.helix.HelixConstants;
+
+
 /**
  * Handles cluster changes such as external view changes, instance config changes, live instance changes etc.
  */
 public interface ClusterChangeHandler {
 
-  void processClusterChange();
+  /**
```
mcvsubbu: please annotate this interface as an evolving application usable interface.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -226,6 +232,45 @@ public void run() {
...
+  /**
+   * To be overridden to plug in custom external view change handlers.
+   * <p>NOTE: all change handlers will be run in a single thread, so any slow change handler can block other change
+   * handlers from running. For slow change handler, make it asynchronous.
+   *
+   * @param spectatorHelixManager Spectator Helix manager
+   * @return List of custom external view change handlers to plug in
+   */
+  @SuppressWarnings("unused")
+  protected List<ClusterChangeHandler> getCustomExternalViewChangeHandlers(HelixManager spectatorHelixManager) {
```
mcvsubbu: Why not inject configuration for class names and instantiate them through a factory like we do for other such classes?

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeHandler.java
Jackie-Jiang: Will do

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
Jackie-Jiang: I think the class name itself should be enough, no need to complicate the interface

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
Jackie-Jiang: I'm not a big fan of using Class.newInstance() because of the reasons here: https://stackoverflow.com/questions/4612386/what-is-the-difference-between-the-new-operator-and-class-newinstance
Also, it might be hard to setup handlers with zero-argument constructor if user want to pass in some properties not in open source code.

Currently we have both flavors of code in the code base. We can discuss this offline.

sunithabeeram: I did think along the same lines as @mcvsubbu - however it appeared that the HelixBrokerStarter is "expected" to be extended and hence its cleaner to let it override the method than introduce yet another config and instantiate through reflection.

mcvsubbu: How will someone running pinot-broker.sh use this feature?

PR #4092, opened 2019-04-09
Add support for '<millis/milliseconds/micros/microseconds/nanos/nanoseconds>SinceEpcoh' time unit string

Jackie-Jiang:


Jackie-Jiang:
@snleee For completeness in case user wants to use it.

PR #4094, opened 2019-04-09
[TE] onboard dataset with epoch time stamps

jihaozh:
For datasets with epoch timestamps, the real dataset granularity and time unit granularity is usually different. By default, TE onboard it as a five-minute granularity metric. This can be configured in dataset configs.


xiaohui-sun:
How to configure it as other granularity like daily or hourly? Does preview work?

PR #4095, opened 2019-04-09
Increasing wait time before beginning tests, to account for delays introduced by upload segments in setup

npawar:
This test is flaky. The realtime table setup sometimes takes long due to the interaction with Kafka and the setup of multiple nodes with errors introduced for test. The periodic task sometimes begins before the setup is completed, resulting in prematurely declaring the test to fail.

### DIFF
--- a/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/controller/periodic/tasks/SegmentStatusCheckerIntegrationTest.java
+++ b/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/controller/periodic/tasks/SegmentStatusCheckerIntegrationTest.java
@@ -243,6 +243,16 @@ public class SegmentStatusCheckerIntegrationTest extends BaseClusterIntegrationT
     Assert.assertEquals(
         controllerMetrics.getValueOfTableGauge(errorOfflineTable, ControllerGauge.PERCENT_SEGMENTS_AVAILABLE), 100);
 
+    // additional 30 seconds wait, to ensure realtime table setup has also completed
+    millisToWait = TimeUnit.MILLISECONDS.convert(30, TimeUnit.SECONDS);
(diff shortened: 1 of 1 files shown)


pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/controller/periodic/tasks/SegmentStatusCheckerIntegrationTest.java
```diff
@@ -243,6 +243,16 @@ public void testSegmentStatusChecker() {
     Assert.assertEquals(
         controllerMetrics.getValueOfTableGauge(errorOfflineTable, ControllerGauge.PERCENT_SEGMENTS_AVAILABLE), 100);
 
+    // additional 30 seconds wait, to ensure realtime table setup has also completed
```
sunithabeeram: Can this be made conditional? How do we know 30s is enough?

npawar: By conditional, do you mean not have the 30 seconds timeout at all? Or take a different path and not do the tests if 30s is up?

npawar:
My earlier assumption about the realtime test table setup taking long is false. The offline tables take up time in the calls to upload segments. On local tests, they can take upto 20s per table. We have 2 such uploads. Based on this, the initial delay for the checker was set to 2 minutes. But this could take longer depending on the setup.
Removing the 30s delay for realtime tests. Instead, increasing the time for beginning tests, as this will affect all tables setup after the uploads. Again, there's no guarantee that the setup will be done in 3 minutes, but practically, it does finish.
@sunithabeeram

npawar:
This code got very refactored as part of this: https://github.com/apache/incubator-pinot/pull/4130
Will discard this change.

PR #4100, opened 2019-04-10
Refactor HelixBrokerStarter to separate constructor and start()

Jackie-Jiang:
For the starter class, constructor should only set the properies,
but not start the service, which should be done by start() instead.

- Refactor HelixBrokerStarter to separate constructor and start()
- Remove redundant configs, always use constant for configs
- Modify shutdown() to shut done all components properly

NOTE: THIS IS A BACKWARD-INCOMPATIBLIE CHANGE
- Need to call start() separately in order to start the broker
- Order of arguments changed in constructor (move nullable to the
  last. This is intentional so that user can find the imcompatible
  easier during compilation, thus add the start())

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
@@ -223,7 +223,8 @@ public class HelixBrokerStarter {
     _participantHelixManager.getMessagingService()
         .registerMessageHandlerFactory(Message.MessageType.USER_DEFINE_MSG.toString(), _tbiMessageHandler);
     addInstanceTagIfNeeded();
-    brokerMetrics.addCallbackGauge("helix.connected", () -> _participantHelixManager.isConnected() ? 1L : 0L);
+    brokerMetrics.addCallbackGauge(CommonConstants.Helix.INSTANCE_CONNECTED_METRIC_NAME,
+        () -> _participantHelixManager.isConnected() ? 1L : 0L);
     _participantHelixManager
         .addPreConnectCallback(() -> brokerMetrics.addMeteredGlobalValue(BrokerMeter.HELIX_ZOOKEEPER_RECONNECTS, 1L));
 
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
@@ -21,7 +21,6 @@ package org.apache.pinot.broker.broker.helix;
 import com.google.common.collect.ImmutableList;
 import com.yammer.metrics.core.MetricsRegistry;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -48,6 +47,7 @@ import org.apache.pinot.broker.routing.HelixExternalViewBasedRouting;
 import org.apache.pinot.common.config.TagNameUtils;
 import org.apache.pinot.common.metadata.ZKMetadataProvider;
 import org.apache.pinot.common.metrics.BrokerMeter;
+import org.apache.pinot.common.metrics.BrokerMetrics;
 import org.apache.pinot.common.utils.CommonConstants;
 import org.apache.pinot.common.utils.NetUtil;
 import org.apache.pinot.common.utils.ServiceStatus;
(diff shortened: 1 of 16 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -47,189 +48,196 @@
...
+
+    // Set up the broker server builder
+    LOGGER.info("Setting up broker server builder");
+    _brokerServerBuilder = new BrokerServerBuilder(_properties, _helixExternalViewBasedRouting,
+        _helixExternalViewBasedRouting.getTimeBoundaryService(), liveInstanceChangeHandler, tableQueryQuotaManager);
+    _accessControlFactory = _brokerServerBuilder.getAccessControlFactory();
     _metricsRegistry = _brokerServerBuilder.getMetricsRegistry();
+    BrokerMetrics brokerMetrics = _brokerServerBuilder.getBrokerMetrics();
+    _helixExternalViewBasedRouting.setBrokerMetrics(brokerMetrics);
+    tableQueryQuotaManager.setBrokerMetrics(brokerMetrics);
```
jackjlli: Can you elaborate why tableQueryQuotaManager is converted to local variable whereas _helixExternalViewBasedRouting remains the same way?

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterTest.java
```diff
@@ -96,33 +96,38 @@
...
+  protected void startBrokers(int numBrokers, int basePort, String zkStr)
+      throws Exception {
     _brokerBaseApiUrl = "http://localhost:" + basePort;
     for (int i = 0; i < numBrokers; i++) {
-      Configuration configuration = BrokerTestUtils.getDefaultBrokerConfiguration();
-      configuration.setProperty("pinot.broker.timeoutMs", 100 * 1000L);
-      configuration.setProperty("pinot.broker.client.queryPort", Integer.toString(basePort + i));
-      configuration.setProperty("pinot.broker.routing.table.builder.class", "random");
-      configuration.setProperty(BrokerServerBuilder.DELAY_SHUTDOWN_TIME_MS_CONFIG, 0);
+      Configuration properties = new PropertiesConfiguration();
```
jackjlli: Can we rename it to `brokerConf`?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/StartBrokerCommand.java
```diff
@@ -107,23 +110,22 @@ public StartBrokerCommand setConfigFileName(String configFileName) {
   public boolean execute()
       throws Exception {
     try {
-      Configuration configuration = readConfigFromFile(_configFileName);
-      if (configuration == null) {
+      Configuration properties = readConfigFromFile(_configFileName);
```
jackjlli: Same as here and other places?

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
Jackie-Jiang: We need to access HelixExternalViewBasedRouting but not TableQueryQuotaManager in other methods (solve IDE warnings lol).
Actually I think it would be better to make all of them protected so the class extends this can access them

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterTest.java
Jackie-Jiang: Done.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/StartBrokerCommand.java
Jackie-Jiang: Done.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/DefaultHelixBrokerConfig.java
```diff
@@ -1,59 +0,0 @@
...
- * under the License.
- */
-package org.apache.pinot.broker.broker.helix;
-
-import java.util.Iterator;
-import org.apache.commons.configuration.Configuration;
-import org.apache.commons.configuration.PropertiesConfiguration;
-
-
-public class DefaultHelixBrokerConfig {
```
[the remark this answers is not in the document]
Jackie-Jiang: Other properties are for config based routing which has been removed long time ago. Didn't clean up the config during that time. As for the default value, because when we read property, we use default value when the property does not exist, so no need to explicitly put a magic number there. Will split this into a separate PR.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -47,189 +48,196 @@
...
-        _helixExternalViewBasedRouting.getTimeBoundaryService(), _liveInstanceChangeHandler, _tableQueryQuotaManager);
-    _accessControlFactory = brokerServerBuilder.getAccessControlFactory();
-    _helixExternalViewBasedRouting.setBrokerMetrics(brokerServerBuilder.getBrokerMetrics());
-    _tableQueryQuotaManager.setBrokerMetrics(brokerServerBuilder.getBrokerMetrics());
-    brokerServerBuilder.start();
-
-    LOGGER.info("Pinot broker ready and listening on port {} for API requests",
-        config.getProperty("pinot.broker.client.queryPort"));
-
-    Runtime.getRuntime().addShutdownHook(new Thread() {
```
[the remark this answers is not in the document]
Jackie-Jiang: It is called in shutdown(), the same as all other services.

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -161,110 +222,68 @@ public HelixBrokerStarter(@Nullable String brokerHost, String helixClusterName,
             CommonConstants.Broker.DEFAULT_BROKER_REFRESH_TIMEBOUNDARY_INFO_SLEEP_INTERVAL_MS));
     _participantHelixManager.getMessagingService()
         .registerMessageHandlerFactory(Message.MessageType.USER_DEFINE_MSG.toString(), _tbiMessageHandler);
-
-    addInstanceTagIfNeeded(helixClusterName, brokerId);
+    addInstanceTagIfNeeded();
+    brokerMetrics.addCallbackGauge("helix.connected", () -> _participantHelixManager.isConnected() ? 1L : 0L);
```
jackjlli: This metric is quite often used. Can we put it to `CommonConstant`?

PR #4102, opened 2019-04-10
Cleanup singletons for 3 classes in ControllerStarter

jackjlli:
This PR cleans up singletons for `PinotLLCRealtimeSegmentManager`, `SegmentCompletionManager` and `RebalanceSegmentStrategyFactory`.

The benefits are: 
* `SegmentCompletionManager` should be injected instead of using a singleton pattern.
* `PinotLLCRealtimeSegmentManager` should be put inside HelixResourceManager (right now these two code are mixed together).
* `RebalanceSegmentStrategyFactory` is only used in `HelixResourceManager`, which shouldn't be a global object.
* Basically singleton shouldn't be created and explicitly freed.
* Singleton isn't the only way to identify there's only 1 object. Using instanceId to differentiate controller processes should be good enough.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
@@ -56,6 +56,7 @@ import org.apache.pinot.controller.helix.core.minion.PinotTaskManager;
 import org.apache.pinot.controller.helix.core.periodictask.ControllerPeriodicTaskScheduler;
 import org.apache.pinot.controller.helix.core.realtime.PinotLLCRealtimeSegmentManager;
 import org.apache.pinot.controller.helix.core.realtime.PinotRealtimeSegmentManager;
+import org.apache.pinot.controller.helix.core.realtime.SegmentCompletionManager;
 import org.apache.pinot.controller.helix.core.rebalance.RebalanceSegmentStrategyFactory;
 import org.apache.pinot.controller.helix.core.relocation.RealtimeSegmentRelocator;
 import org.apache.pinot.controller.helix.core.retention.RetentionManager;
@@ -108,6 +109,8 @@ public class ControllerStarter {
   private ControllerPeriodicTaskScheduler _controllerPeriodicTaskScheduler;
   private PinotHelixTaskResourceManager _helixTaskResourceManager;
   private PinotRealtimeSegmentManager _realtimeSegmentsManager;
+  private PinotLLCRealtimeSegmentManager _pinotLLCRealtimeSegmentManager;
+  private SegmentCompletionManager _segmentCompletionManager;
   private ControllerLeadershipManager _controllerLeadershipManager;
   private List<ServiceStatus.ServiceStatusCallback> _serviceStatusCallbackList;
 
@@ -259,8 +262,16 @@ public class ControllerStarter {
 
     // Helix resource manager must be started in order to create PinotLLCRealtimeSegmentManager
     LOGGER.info("Starting realtime segment manager");
-    PinotLLCRealtimeSegmentManager
-        .create(_helixResourceManager, _config, _controllerMetrics, _controllerLeadershipManager);
+
+    _pinotLLCRealtimeSegmentManager =
+        new PinotLLCRealtimeSegmentManager(_helixResourceManager, _config, _controllerMetrics,
+            _controllerLeadershipManager);
+    // TODO: Need to put this inside HelixResourceManager when ControllerLeadershipManager is removed.
+    _helixResourceManager.registerPinotLLCRealtimeSegmentManager(_pinotLLCRealtimeSegmentManager);
+    _segmentCompletionManager =
+        new SegmentCompletionManager(helixParticipantManager, _pinotLLCRealtimeSegmentManager, _controllerMetrics,
+            _controllerLeadershipManager, _config.getSegmentCommitTimeoutSeconds());
+
     _realtimeSegmentsManager = new PinotRealtimeSegmentManager(_helixResourceManager, _controllerLeadershipManager);
     _realtimeSegmentsManager.start(_controllerMetrics);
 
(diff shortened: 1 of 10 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -211,8 +214,8 @@ public void start() {
   private void setUpHelixController() {
     // Register and connect instance as Helix controller.
     LOGGER.info("Starting Helix controller");
-    _helixControllerManager = HelixSetupUtils
-        .setup(_helixClusterName, _helixZkURL, _instanceId, _isUpdateStateModel, _enableBatchMessageMode);
+    _helixControllerManager = HelixSetupUtils.setup(_helixClusterName, _helixZkURL, _instanceId, _isUpdateStateModel,
```
Jackie-Jiang: Have you changed your setting for the code format? Seems this part is reformatted back and forth.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -254,8 +257,15 @@ private void setUpPinotController() {
...
     // Helix resource manager must be started in order to create PinotLLCRealtimeSegmentManager
     LOGGER.info("Starting realtime segment manager");
-    PinotLLCRealtimeSegmentManager
-        .create(_helixResourceManager, _config, _controllerMetrics, _controllerLeadershipManager);
+
+    _pinotLLCRealtimeSegmentManager =
+        new PinotLLCRealtimeSegmentManager(_helixResourceManager, _config, _controllerMetrics,
+            _controllerLeadershipManager);
+    // TODO: Need to put this inside HelixResourceManager when ControllerLeadershipManager is removed.
+    _helixResourceManager.registerPinotLLCRealtimeSegmentManager(_pinotLLCRealtimeSegmentManager);
```
Jackie-Jiang: This line can be moved into the constructor of PinotLLCRealtimeSegmentManager?

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -265,8 +275,9 @@ private void setUpPinotController() {
     _controllerPeriodicTaskScheduler = new ControllerPeriodicTaskScheduler();
     _controllerPeriodicTaskScheduler.init(controllerPeriodicTasks, _controllerLeadershipManager);
 
-    LOGGER.info("Creating rebalance segments factory");
-    RebalanceSegmentStrategyFactory.createInstance(helixParticipantManager);
+    LOGGER.info("Registering rebalance segments factory");
+    _helixResourceManager.registerRebalanceSegmentStrategyFactory(
```
Jackie-Jiang: This line can be moved into the constructor of RebalanceSegmentStrategyFactory and pass in _helixResourceManager

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -139,25 +136,13 @@ public String getControllerVipUrl() {
     return _controllerConf.generateVipUrl();
   }
 
-  public static synchronized void create(PinotHelixResourceManager helixResourceManager, ControllerConf controllerConf,
+  public PinotLLCRealtimeSegmentManager(PinotHelixResourceManager helixResourceManager, ControllerConf controllerConf,
```
Jackie-Jiang: I think you can remove the old constructor and only keep this one. (For the test, use mock)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
```diff
@@ -84,12 +81,15 @@ public static int getMaxCommitTimeForAllSegmentsSeconds() {
 
   // TODO keep some history of past committed segments so that we can avoid looking up PROPERTYSTORE if some server comes in late.
 
-  protected SegmentCompletionManager(HelixManager helixManager, PinotLLCRealtimeSegmentManager segmentManager,
-      ControllerMetrics controllerMetrics, ControllerLeadershipManager controllerLeadershipManager) {
+  public SegmentCompletionManager(HelixManager helixManager, PinotLLCRealtimeSegmentManager segmentManager,
```
Jackie-Jiang: Re-order the arguments probably?

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
jackjlli: I've tried that but since the object isn't fully initialized, it will only assign a null reference to it.
Plus, the logic of these 2 objects are mixed together: 
* `Resource manager` has to be instantiated before `PinotLLCRealtimeSegmentManager`.
* `Resource manager` needs to call the method in `PinotLLCRealtimeSegmentManager`, which is created after `Resource manager` got created.

It's a circular reference dependency between these two objects.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
jackjlli: Same as this one.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
Jackie-Jiang: Make sense lol

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
jackjlli: Done.

PR #4103, opened 2019-04-10
Add validations to table config before updating a table config

npawar:
We did not enforce any tenant level validations to update requests (like we do in add table) that come from update rest api. This can result in incorrect tenant configs being added to live tables.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -302,7 +302,7 @@ public class PinotTableRestletResource {
 
       ensureMinReplicas(tableConfig);
       verifyTableConfigs(tableConfig);
-      _pinotHelixResourceManager.setExistingTableConfig(tableConfig, tableNameWithType, tableType);
+      _pinotHelixResourceManager.updateTableConfig(tableConfig, tableNameWithType, tableType);
     } catch (PinotHelixResourceManager.InvalidTableConfigException e) {
       String errStr = String.format("Failed to update configuration for %s due to: %s", tableName, e.getMessage());
       _controllerMetrics.addMeteredGlobalValue(ControllerMeter.CONTROLLER_TABLE_UPDATE_ERROR, 1L);
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -102,6 +102,11 @@ public class PinotTableRestletResource {
   @Inject
   ExecutorService _executorService;
 
+  /**
+   * API to create a table. Before adding, validations will be done (min number of replicas,
+   * checking offline and realtime table configs match, checking for tenants existing)
+   * @param tableConfigStr
+   */
   @POST
   @Produces(MediaType.APPLICATION_JSON)
   @Path("/tables")
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -1145,7 +1145,8 @@ public class PinotHelixResourceManager {
       if (realtimeConsumingTag != null) {
         if (!TagNameUtils.hasValidServerTagSuffix(realtimeConsumingTag)) {
           throw new PinotHelixResourceManager.InvalidTableConfigException(
-              "Invalid realtime consuming tag: " + realtimeConsumingTag + ". Must have suffix _REALTIME or _OFFLINE");
+              "Invalid realtime consuming tag: " + realtimeConsumingTag + " for table " + tableNameWithType
+                  + ". Must have suffix _REALTIME or _OFFLINE");
         }
         if (HelixHelper.getInstancesWithTag(_helixZkManager, realtimeConsumingTag).isEmpty()) {
           throw new PinotHelixResourceManager.InvalidTableConfigException(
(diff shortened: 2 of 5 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1160,7 +1113,60 @@ public void addTable(@Nonnull TableConfig tableConfig)
...
+      throw new PinotHelixResourceManager.InvalidTableConfigException(
+          "Server tenant: " + serverTenantName + " does not exist for table: " + tableNameWithType);
+    }
+    TagOverrideConfig tagOverrideConfig = tenantConfig.getTagOverrideConfig();
+    if (tagOverrideConfig != null) {
+      String realtimeConsumingTag = tagOverrideConfig.getRealtimeConsuming();
+      if (realtimeConsumingTag != null) {
+        if (!TagNameUtils.hasValidServerTagSuffix(realtimeConsumingTag)) {
+          throw new PinotHelixResourceManager.InvalidTableConfigException(
+              "Invalid realtime consuming tag: " + realtimeConsumingTag + ". Must have suffix _REALTIME or _OFFLINE");
```
mcvsubbu: include table name in the exception

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1160,7 +1113,60 @@ public void addTable(@Nonnull TableConfig tableConfig)
...
+              "No instances found with overridden realtime consuming tag: " + realtimeConsumingTag + " for table: "
+                  + tableNameWithType);
+        }
+      }
+
+      String realtimeCompletedTag = tagOverrideConfig.getRealtimeCompleted();
+      if (realtimeCompletedTag != null) {
+        if (!TagNameUtils.hasValidServerTagSuffix(realtimeCompletedTag)) {
+          throw new PinotHelixResourceManager.InvalidTableConfigException(
+              "Invalid realtime completed tag: " + realtimeCompletedTag + ". Must have suffix _REALTIME or _OFFLINE");
```
mcvsubbu: include table name in the exception

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1160,7 +1113,60 @@ public void addTable(@Nonnull TableConfig tableConfig)
...
 
-    handleBrokerResource(tableNameWithType, brokersForTenant);
+    String brokerTenantName = TagNameUtils.getBrokerTagForTenant(tableConfig.getTenantConfig().getBroker());
+    handleBrokerResource(tableNameWithType, HelixHelper.getInstancesWithTag(_helixZkManager, brokerTenantName));
+  }
+
+  /**
+   * Validates the tenant config for the table
+   */
+  private void validateTableTenantConfig(TableConfig tableConfig, String tableNameWithType, TableType tableType) {
```
mcvsubbu: Will be great if you can add a unit test for this. It will mean mocking the helix get tagged instances call, I suppose.

npawar: This whole method is not something I've added. I've just wrapped the code into a method. And I can see addTable is used in multiple tests. I'll check what all those cover and add if specific cases are missing

npawar:
> Is the table config modification command covered in an integration test? If not, can you add it to one of the existing integration tests? It may also be useful to do some basic sanity checks (manual) on table config changes. thanks

It is covered in a few of them. Which is why I had to correct the usage of the update in one of the tests.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -302,7 +302,7 @@ public SuccessResponse updateTableConfig(
 
       ensureMinReplicas(tableConfig);
       verifyTableConfigs(tableConfig);
-      _pinotHelixResourceManager.setExistingTableConfig(tableConfig, tableNameWithType, tableType);
+      _pinotHelixResourceManager.updateTableConfig(tableConfig, tableNameWithType, tableType);
```
jackjlli: We've already have a method called `verifyTableConfigs` in Line 304 and there's a new method `validateTableTenantConfig ` in `updateTableConfig`. Can they be merged together?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1160,7 +1113,62 @@ public void addTable(@Nonnull TableConfig tableConfig)
...
+    TenantConfig tenantConfig = tableConfig.getTenantConfig();
+    if (tenantConfig == null || tenantConfig.getBroker() == null || tenantConfig.getServer() == null) {
+      throw new PinotHelixResourceManager.InvalidTableConfigException(
```
jackjlli: Why do we need to explicitly specify the class as `PinotHelixResourceManager.InvalidTableConfigException`? Is it ok to just say `InvalidTableConfigException` since the exception is just in Class `PinotHelixResourceManager `?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
npawar: verifyTableCOnfigs compares the offline and realtime table configs for hybrid tables.
validateTableTenantCOnfig is specifically for validating tenant configs.
I intentionally pushed the validation inside the update method, because update and add are called from multiple files (v1 and v2 configs). And it was better to have the call to validate in 1 place instead of in every endpoint.
Maybe the ensureMinReplicas and verify can also be pushed down. But that would need more evaluation, because all tests use the add and update and setExisting methods, which do not want all the validations

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
npawar: yes it should be okay. I havent written this method, moved the code into a method. But sure, I can clean that up too

npawar:
> LGTM. One more thing we should do before committing this PR is to add some comments on `addTable` API, saying broker and server hosts have to be ready with tag before calling this API.
> Thanks for working on that!

Thanks for the suggestion! Added that comment to the api

npawar:
Linking to https://github.com/apache/incubator-pinot/issues/4076

PR #4104, opened 2019-04-11
[TE] frontend - harleyjj/alert-overview - remove word Average in anom…

harleyjj:
…aly table

1) Removes word "Average," since values could be aggregate or average depending on the case


xiaohui-sun:
good job!

PR #4105, opened 2019-04-12
Set Helix flapping window for all components

Jackie-Jiang:
Helix will disconnect the manager and disable the instance if it
detects flapping (too frequent disconnect from ZooKeeper). Setting
flapping time window to a small value can avoid this from
happening. Helix ignores the non-positive value, so set the default
value as 1.

Move the config into CommonConstants
- For server, keep the old behavior
- For broker, change default value from 0 (not supported) to 1
- For controller and minion, add this system property


Jackie-Jiang:
@snleee Good point, added.

PR #4113, opened 2019-04-13
[TE] Fix MySQL and H2 timestamp automatic timezone conversion issues

tangdian:
Context: For MySQL and H2, the "timestamp" field will be automatically converted to connection timezone (thirdeye backend timezone), which may cause inconsistency issue in local and prod. So I am giving user the option to set server timezone.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/pinot/resultset/ThirdEyeResultSetUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/pinot/resultset/ThirdEyeResultSetUtils.java
@@ -45,9 +45,12 @@ import org.slf4j.LoggerFactory;
 
 public class ThirdEyeResultSetUtils {
   private static final Logger LOG = LoggerFactory.getLogger(ThirdEyeResultSetUtils.class);
+  private static final String MYSQL = "MySQL";
+  private static final String H2 = "H2";
 
   public static List<String[]> parseResultSets(ThirdEyeRequest request,
-      Map<MetricFunction, List<ThirdEyeResultSet>> metricFunctionToResultSetList, String sourceName) throws ExecutionException {
+      Map<MetricFunction, List<ThirdEyeResultSet>> metricFunctionToResultSetList,
+      String sourceName) throws ExecutionException {
 
     int numGroupByKeys = 0;
     boolean hasGroupBy = false;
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/pinot/resultset/ThirdEyeResultSetUtils.java
```diff
@@ -85,11 +88,15 @@
...
       boolean isISOFormat = false;
       DateTimeFormatter inputDataDateTimeFormatter = null;
+      DateTimeFormatter serverDataDateTimeFormatter = null;
       String timeFormat = dataTimeSpec.getFormat();
       if (timeFormat != null && !timeFormat.equals(TimeSpec.SINCE_EPOCH_FORMAT)) {
         isISOFormat = true;
         inputDataDateTimeFormatter = DateTimeFormat.forPattern(timeFormat).withZone(dateTimeZone);
       }
+      if (sourceName.equals(MYSQL) || sourceName.equals(H2) ) {
+        serverDataDateTimeFormatter = DateTimeFormat.forPattern(timeFormat).withZone(DateTimeZone.forID(timezone));
```
jihaozh: Add some comments about `timezone` vs `dateTimeZone`

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/sql/SqlThirdEyeDataSource.java
```diff
@@ -45,11 +45,13 @@
   private static final ThirdEyeCacheRegistry CACHE_REGISTRY_INSTANCE = ThirdEyeCacheRegistry.getInstance();
   protected LoadingCache<RelationalQuery, ThirdEyeResultSetGroup> sqlResponseCache;
   private SqlResponseCacheLoader sqlResponseCacheLoader;
-  public static final String DATA_SOURCE_NAME = SqlThirdEyeDataSource.class.getSimpleName();
-
+  private String serverTimezone = "UTC";
 
 
   public SqlThirdEyeDataSource(Map<String, Object> properties) throws Exception {
+    if (properties.containsKey("serverTimezone")) {
```
jihaozh: So the `serverTimeZone` have to be configured? What about using 'TimeZone tz = TimeZone.getDefault();' It will get the local time zone automatically. 
See https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html

tangdian: Good to know! I didn't realize there is a way to get local time zone. I will be using DateTimeZone.getDefault().

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/pinot/resultset/ThirdEyeResultSetUtils.java
tangdian: I will be using DateTimeZone.getDefault() now.

PR #4117, opened 2019-04-15
Admin tool for listing segments with invalid intervals for offline tables

npawar:
As per discussion in https://github.com/apache/incubator-pinot/pull/4078 we reduced the loggin in OfflineSegmentIntervalChecker periodic task, and decided to add admin option to be able to list segments with invalid intervals


pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
```diff
@@ -0,0 +1,149 @@
...
+  private boolean _help = false;
+
+  @Override
+  public boolean getHelp() {
+    return _help;
+  }
+
+  @Override
+  public String toString() {
+    return ("OfflineSegmentIntervalChecker");
```
jackjlli: Nit: why do we need the brackets?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
```diff
@@ -0,0 +1,149 @@
...
+      for (String tableName : resourcesInCluster) {
+        if (TableNameBuilder.isOfflineTableResource(tableName)) {
+          offlineTables.add(tableName);
+        }
+      }
+    } else {
+      for (String tableName : _tableNames.split(",")) {
+        if (ZKMetadataProvider.getOfflineTableConfig(_propertyStore, tableName) != null) {
+          offlineTables.add(tableName);
+        }
```
jackjlli: If it returns null znode, can we log a message here(i.e. sth needs to be cleaned up)?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
```diff
@@ -0,0 +1,149 @@
...
+      }
+    }
+
+    LOGGER.info("Tables to check: {}", offlineTables);
```
jackjlli: Does it print out the content of the array list or just the physical address? Or should we just print the number of offline tables to check?

pinot-controller/src/main/java/org/apache/pinot/controller/util/SegmentIntervalUtils.java
```diff
@@ -0,0 +1,74 @@
...
+    return timeInterval != null && TimeUtils.timeValueInValidRange(timeInterval.getStartMillis()) && TimeUtils
+        .timeValueInValidRange(timeInterval.getEndMillis());
+  }
+
+  /**
+   * We only want to check missing segments if the table has at least 2 segments and a time column
+   */
+  public static boolean eligibleForMissingSegmentCheck(int numSegments,
+      SegmentsValidationAndRetentionConfig validationConfig) {
+    return numSegments >= 2 && StringUtils.isNotEmpty(validationConfig.getTimeColumnName());
```
jackjlli: You can just call `eligibleForSegmentIntervalCheck` inside `eligibleForMissingSegmentCheck` to reduce duplicated logic.

pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
```diff
@@ -122,4 +122,11 @@ public static String extractRawTableName(@Nonnull String tableName) {
   public static boolean isTableResource(@Nonnull String resourceName) {
     return OFFLINE.tableHasTypeSuffix(resourceName) || REALTIME.tableHasTypeSuffix(resourceName);
   }
+
+  /**
+   * Return whether the given resource name represents an offline table resource.
+   */
+  public static boolean isOfflineTableResource(@Nonnull String resourceName) {
```
sunithabeeram: Why not use getTableTypeFromTableName?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.pinot.controller.util.SegmentIntervalUtils;
+import org.apache.pinot.tools.Command;
+import org.joda.time.Interval;
+import org.kohsuke.args4j.Option;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Pinot admin command to list all offline segments with invalid intervals, group by table name
```
sunithabeeram: Could this also take a table-name and check for just that one?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
npawar: It prints contents of the array. We want the tables to be listed. This is an adhoc tool, so shouldn't matter that much

pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
npawar: Wanted to make one similar to the method above (isTableResource), but only for OFFLINE

pinot-controller/src/main/java/org/apache/pinot/controller/util/SegmentIntervalUtils.java
npawar: they are checks which are independent of each other, and I didn't want missingSegmentCheck to know about segmentIntervalCheck..

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
npawar: Yes, -tableNames is an option

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
npawar: Done

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
npawar: Fixed. It was a result of copying code from other AdminCommands :)

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
jackjlli: Yeah, but the table names will be printed in line 118 though. Not sure that's duplicated or not..

npawar: My intention was to print out the list upfront here, because some of the tables could get skipped because of the conditions just above. How about I let it stick for now, and I'll remove it later if it becomes a nuisance during analysis? :)

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/OfflineSegmentIntervalCheckerCommand.java
```diff
@@ -0,0 +1,151 @@
...
+      for (OfflineSegmentZKMetadata offlineSegmentZKMetadata : offlineSegmentZKMetadataList) {
+        Interval timeInterval = offlineSegmentZKMetadata.getTimeInterval();
+        if (!SegmentIntervalUtils.isValidInterval(timeInterval)) {
+          segmentsWithInvalidIntervals.add(offlineSegmentZKMetadata.getSegmentName());
+        }
+      }
+    }
+    return segmentsWithInvalidIntervals;
+  }
+
```
snleee: remove line

npawar: done

PR #4118, opened 2019-04-16
[TE] frontend - harleyjj/home - get all anomalies by subscription group

harleyjj:
1) Implements get all subscription groups in Ember Data with cache, which is accessible from all routes in the Ember app
2) On home, makes it possible to get all anomalies by subscription group instead of application

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/home/index/template.hbs
+++ b/thirdeye/thirdeye-frontend/app/pods/home/index/template.hbs
@@ -19,19 +19,46 @@
 
   <article class="dashboard-container__body">
     <section class="dashboard-container__application-header">
-      <h2 class="dashboard-container__title">Application:<span>{{appName}}</span></h2>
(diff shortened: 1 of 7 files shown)


thirdeye/thirdeye-frontend/app/pods/home/index/template.hbs
```diff
@@ -19,19 +19,46 @@
...
-          selected=appNameSelected
-          searchField="application"
-          searchEnabled=true
-          placeholder="Please pick an application"
-          onchange=(action "selectApplication")
-          as |app|}}
-          {{app.application}}
-        {{/power-select}}
-      </div>
+      <h3 class="dashboard-container__title">Anomalies by: </h3>
```
akshayrai: Would name it as "Filter by"

harleyjj: How about "Anomalies Filtered by" ?

PR #4125, opened 2019-04-16
[TE] Fix the dimension values that contain dot to show up correctly in ThirdEye

jihaozh:
- If there is a dot in a dimension value, it won't be able to show up correctly in the RCA page. This also affects the time series fetching cause it's querying a wrong dimension name. This is caused by a hack to parse the timestamps that contain a dot in the sql result. This PR removes this hack and resolves the issue. The timestamp parsing should be configured in the data set time stamp format instead.
- Add static final constants to decrease memory footprint.


xiaohui-sun:
Did you forget to include some files?
I can only see the static final changes.

xiaohui-sun:
Oh I missed that. Approved.

PR #4127, opened 2019-04-17
Declare byte array size when initializing ByteArrayOutputStream

jackjlli:
We've seen some use cases with high latency on `dataTable.toBytes()` method.
This PR declares byte array size when initializing `ByteArrayOutputStream`, which reduces the time of resizing the object. This helps especially when the object time is large.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/common/datatable/DataTableImplV2.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/common/datatable/DataTableImplV2.java
@@ -50,6 +50,11 @@ public class DataTableImplV2 implements DataTable {
   // VARIABLE_SIZE_DATA (START|SIZE)
   private static final int HEADER_SIZE = Integer.BYTES * 13;
 
+  private static final int OFFSET_BYTE_SIZE = 4;
+  private static final int SIZE_OF_BYTE_ARRAY_BYTE_SIZE = 4;
+  // denotes version, number of rows and columns.
+  private static final int MANDATORY_INFO_BYTE_SIZE = 3 * 4;
+
   private final int _numRows;
   private final int _numColumns;
   private final DataSchema _dataSchema;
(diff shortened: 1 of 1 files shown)


sunithabeeram:
Can you provide more context here? How much of a slow down are you seeing? How much is this change improving things by?

jackjlli:
@sunithabeeram In one of the high qps use cases, we've seen around 40k requests which latencies are > 400ms. There're 60% of them that have < 10ms scheduler wait time, < 10ms execution time, but higher total time. The extra time is taken in `dataTable.toBytes()` method. Thus, optimizing this method would definitely help to reduce the time of latency.

jackjlli:
@Jackie-Jiang The byte array `buf` is encapsulated in `byteArrayOutputStream` and cannot directly be accessed though. Plus, `ByteBuffer.wrap()` only returns byte buffer, which isn't what we want to return in the end.

sunithabeeram:
Thanks for the info @jackjlli. A few more questions:
- How big of a response is being sent back? Resize typically increases exponentially, so we shouldn't see too many resizes. I have chased some of these resize based optimizations in the past - although the theory is sound for runtime as well as garbage generated, the gains typically didn't materialize - which is why I am bit skeptical about this. If you have numbers to share for after this fix, that will help. Also, if possible, attaching the CPU profile snapshot showing the hotspot would be nice.
- Should metadata serialization take the same approach? I am wondering if we are sending a lot more metadata back; that part of the code still doesn't initialize the size.
- I didn't see a comment on the testing for this. Do we have existing tests covering all the paths for this method?

pinot-core/src/main/java/org/apache/pinot/core/common/datatable/DataTableImplV2.java
```diff
@@ -346,6 +354,17 @@ public void addException(@Nonnull ProcessingException processingException) {
...
 
+  private int getTotalByteArraySize(byte[] metadataBytes, byte[] dictionaryMapBytes, byte[] dataSchemaBytes) {
+    int metadataSize = metadataBytes.length + 1;
+    int dictionarySize = dictionaryMapBytes != null ? dictionaryMapBytes.length + 1 : 2;
+    int dataSchemaSize = dataSchemaBytes != null ? dataSchemaBytes.length + 1 : 2;
+    int fixedSizeDataSize = _fixedSizeDataBytes != null ? _fixedSizeDataBytes.length + 1 : 2;
+    int variableSizeDataSize = _variableSizeDataBytes != null ? _variableSizeDataBytes.length + 1 : 2;
+
+    // total size that byte array needs. 3 denotes version, number of rows and columns.
+    return 3 + metadataSize + dictionarySize + dataSchemaSize + fixedSizeDataSize + variableSizeDataSize;
```
sunithabeeram: Could we use static constants instead of magic numbers above? Its not immediately clear why something is a 2 vs 1 etc. Looking through the actual serialization gives a clue, but thats not evident here.

I am assuming that even if the estimate here is off, the buffer will still get resized and we'll not encounter issues. If so, do the estimates need to be precise? Also, would it be an option to use a thread-local BAOS so the object is not allocated everytime?

Jackie-Jiang:
> @Jackie-Jiang The byte array `buf` is encapsulated in `byteArrayOutputStream` and cannot directly be accessed though. Plus, `ByteBuffer.wrap()` only returns byte buffer, which isn't what we want to return in the end.

You need to pass in a byte array into `ByteBuffer.wrap()`, which you can directly return.
While, after consideration, maybe we should move it off-heap. But let's first make a tool to test the performance and gather some numbers. You can find examples under pinot-perf package.

jackjlli:
@sunithabeeram 
* Right now we don't have any metric of tracking the response size yet. It'd be good to add one.
* We can do the same for serializing metadata, dictionary and data schema. But there may be several `for` loops in those methods, which need to create multiple hashmaps to track the byte[] beforehand.
* We've already had unit tests on DataTable, so we're good here.

jackjlli:
> > @Jackie-Jiang The byte array `buf` is encapsulated in `byteArrayOutputStream` and cannot directly be accessed though. Plus, `ByteBuffer.wrap()` only returns byte buffer, which isn't what we want to return in the end.
> 
> You need to pass in a byte array into `ByteBuffer.wrap()`, which you can directly return.
> While, after consideration, maybe we should move it off-heap. But let's first make a tool to test the performance and gather some numbers. You can find examples under pinot-perf package.

Done in the latest update.

sunithabeeram:
> Right now we don't have any metric of tracking the response size yet. It'd be good to add one.
The broker side logs will have the response size.

jackjlli:
I did the benchmark on one of the high qps use cases, finding out 720k out of 1.18m responses have sizes > 1000 bytes.

jackjlli:
In another high qps case where the percentage of response sizes > 1000 bytes is only 10% doesn't seem to have the latency symptom though.

jackjlli:
This PR was to solve some use cases where the heap size is big. But after finding out that splitting the data into two or more servers mitigates the issue, I don't think we need this PR any more. Closing this PR now.

PR #4128, opened 2019-04-17
fix non-engish field value gibberish

sunny209868:
Using Java client API, met with the value of the non-english garbled words phenomenon, found in the httpResponse.GetResponseBody(), no unified character encoding, submit code fixes and unit test


pinot-api/src/test/java/org/apache/pinot/client/QueryClientTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+
+/**
+ *  Tests for resp body java client api coding utf-8
+ */
+public class QueryClientTest {
+  @Test
+  public void ClientRespCoding() {
+    Connection connection =
+        ConnectionFactory.fromHostList("10.85.126.103:8099", "10.85.126.104:8099", "10.85.126.105:8099");
```
Jackie-Jiang: This is not a unit test. You need to setup the cluster in these ip addresses in order to make it work...

sunny209868: yes, If  want to do unit tests,I need to launch zookeeper, controller, broker, and server first, then create schema, table, upload data, and finally do query tests. It is a little complicated, so I follow ConnectionFactoryTest and encapsulate the tests into the main function

sunny209868:
> I think you can just remove the test

ok, I remove the test.

PR #4133, opened 2019-04-18
 add more types parsing for string coerceValueIntoField

katewang0710:
Hi,

change code to make coerceValueIntoField support more type parsing for string type.


Jackie-Jiang:
Can you handle the primitive types as well?
Also, I think the correct way to check is `if (fieldType.isAssgnableFrom(<type>.class))`

katewang0710:
Hi @Jackie-Jiang,

Do you mean support  boolean , byte , char , short , int , long , float and double for string type value?

Thanks,
Kate

Jackie-Jiang:
@katewang0710 Yes, similar to how we handle primitive int and boolean

katewang0710:
@kishoreg @Jackie-Jiang Looks like all the tests have passed. Can we merge this change? If anything needs to change please let me know.

Thanks

pinot-common/src/main/java/org/apache/pinot/common/config/Deserializer.java
```diff
@@ -349,17 +349,26 @@
...
-            } else {
-              // TODO Add float/double/short/char
+            } else if (fieldType.isAssignableFrom(Float.class) || fieldType.isAssignableFrom(float.class)){
+              destinationValue = Float.parseFloat(stringValue);
+            } else if (fieldType.isAssignableFrom(Double.class) || fieldType.isAssignableFrom(double.class)){
+              destinationValue = Double.parseDouble(stringValue);
+            } else if (fieldType.isAssignableFrom(Short.class) || fieldType.isAssignableFrom(short.class)){
+              destinationValue = Short.parseShort(stringValue);
+            } else if (fieldType.isAssignableFrom(Byte.class) || fieldType.isAssignableFrom(byte.class)){
+              destinationValue = stringValue.getBytes()[0];
```
Jackie-Jiang: Byte.parseByte(stringValue)?

katewang0710: Exactly! Thanks

Jackie-Jiang: Can you fix the format issue? Then I'll merge it.

PR #4147, opened 2019-04-19
Add upper lower bounds

xiaohui-sun:
Add upper/lower bounds to TimeSeries.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/TimeSeries.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/TimeSeries.java
@@ -37,49 +37,47 @@ public class TimeSeries {
     this.df = new DataFrame();
   }
 
-  /**
-   * Add the time stamps into the timeseries
-   * @param timestamps
-   */
-  public void addTimeStamps(LongSeries timestamps) {
+  public TimeSeries(LongSeries timestamps, DoubleSeries baselineValues) {
     this.df.addSeries(COL_TIME, timestamps).setIndex(COL_TIME);
-  }
-
-  /**
-   * Add the predicted baseline into the timeseries
-   * @param baselineValues predicted baseline values
-   */
-  public void addPredictedBaseline(DoubleSeries baselineValues) {
     this.df.addSeries(DataFrameUtils.COL_VALUE, baselineValues);
   }
 
-  /**
-   * Add the predicted upper bound into the timeseries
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/TimeSeries.java
```diff
@@ -53,17 +53,47 @@ public void addPredictedBaseline(DoubleSeries baselineValues) {
     this.df.addSeries(DataFrameUtils.COL_VALUE, baselineValues);
   }
 
+  /**
```
jihaozh: Shouldn't the result include current time series too?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/TimeSeries.java
```diff
@@ -53,17 +53,47 @@ public void addPredictedBaseline(DoubleSeries baselineValues) {
...
+   * Add the predicted upper bound into the timeseries
+   * @param upperBoundValues predicted upper bound values
+   */
+  public void addPredictedUpperBound(DoubleSeries upperBoundValues) {
```
jihaozh: I'd prefer we have a time series builder class to construct the TimeSeries class. Something like:

` Public static class Builder() { `
        `  public Builder(LongSeries timeStamp, DoubleSeries currentTimeSeries, DoubleSeries predictedBaseline); `
       `   public Builder upperBound(DoubleSeries predictedUpperBound);`
       `   public Builder lowerBound(DoubleSeries predictedLowerBound);`
       ` .  public TimeSeries Build();`
`   }
`
Then the DataFrame class won't be exposed to client and we have a explicit contract to build TimeSeries instance.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/TimeSeries.java
xiaohui-sun: Good point! We should include current as well.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/TimeSeries.java
xiaohui-sun: As discussed. Changed the constructors to handle it more efficiently.

PR #4148, opened 2019-04-19
[TE] detection - two side threshold filter

jihaozh:
- This PR makes it possible to set different threshold in different anomaly pattern in percentage change anomaly filter.
- Fix baseline values gathering in rule-based anomaly filters. Previously, the calculation of the baseline values in some anomaly filters only considers the first metric slice's value, which might result in wrong values in some cases.
- Unit tests.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/PercentageChangeRuleAnomalyFilterSpec.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/PercentageChangeRuleAnomalyFilterSpec.java
@@ -21,9 +21,11 @@ package org.apache.pinot.thirdeye.detection.spec;
 
 public class PercentageChangeRuleAnomalyFilterSpec extends AbstractSpec {
   private String timezone = "UTC";
-  private double threshold = Double.NaN;
   private String offset;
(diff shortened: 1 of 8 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/PercentageChangeRuleAnomalyFilterSpec.java
```diff
@@ -21,9 +21,11 @@
 
 public class PercentageChangeRuleAnomalyFilterSpec extends AbstractSpec {
   private String timezone = "UTC";
-  private double threshold = Double.NaN;
   private String offset;
   private String pattern= "UP_OR_DOWN";
+  private double threshold = 0.0; // by default set threshold to 0 to pass all anomalies
+  private double upThreshold = Double.NaN;
```
xiaohui-sun: Why default value is NaN?

jihaozh: Most of the filters are one side unless the threshold is overridden by upThreshold or downThreshold. If they are not NaN, then we know it's overridden.

PR #4149, opened 2019-04-19
[TE] support monitoring granularity for Holt Winters

tangdian:
This will enable 7_DAYS, 1_MONTHS monitoring granularity and beyond.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/HoltWintersDetector.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/HoltWintersDetector.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.math3.analysis.MultivariateFunction;
 import org.apache.commons.math3.optim.PointValuePair;
+import org.apache.pinot.thirdeye.common.time.TimeGranularity;
 import org.apache.pinot.thirdeye.dataframe.BooleanSeries;
 import org.apache.pinot.thirdeye.dataframe.DataFrame;
 import org.apache.pinot.thirdeye.dataframe.DoubleSeries;
@@ -32,7 +33,6 @@ import org.apache.pinot.thirdeye.dataframe.Series;
 import org.apache.pinot.thirdeye.dataframe.util.MetricSlice;
 import org.apache.pinot.thirdeye.datalayer.dto.DatasetConfigDTO;
 import org.apache.pinot.thirdeye.datalayer.dto.MergedAnomalyResultDTO;
-import org.apache.pinot.thirdeye.detection.ConfigUtils;
 import org.apache.pinot.thirdeye.detection.DetectionUtils;
 import org.apache.pinot.thirdeye.detection.InputDataFetcher;
 import org.apache.pinot.thirdeye.detection.Pattern;
@@ -50,7 +50,6 @@ import org.apache.pinot.thirdeye.rootcause.impl.MetricEntity;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.Interval;
-import org.joda.time.Period;
 import org.apache.commons.math3.optim.MaxIter;
 import org.apache.commons.math3.optim.nonlinear.scalar.ObjectiveFunction;
 import org.apache.commons.math3.optim.MaxEval;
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/HoltWintersDetector.java
```diff
@@ -112,13 +112,25 @@ public void init(HoltWintersDetectorSpec spec, InputDataFetcher dataFetcher) {
     this.pattern = spec.getPattern();
     this.smoothing = spec.getSmoothing();
     this.sensitivity = spec.getSensitivity();
+    this.monitoringGranularity = spec.getMonitoringGranularity();
+
+    if (this.monitoringGranularity.equals("1_MONTHS")) {
+      this.timeGranularity = MetricSlice.NATIVE_GRANULARITY;
```
xiaohui-sun: What is the rationale behind this logic?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/HoltWintersDetector.java
```diff
@@ -137,15 +149,20 @@ public TimeSeries computePredictedTimeSeries(MetricSlice slice) {
   @Override
   public List<MergedAnomalyResultDTO> runDetection(Interval window, String metricUrn) {
     MetricEntity metricEntity = MetricEntity.fromURN(metricUrn);
-    DateTime trainStart = window.getStart().minus(lookbackPeriod);
+    DateTime trainStart;
+    if (!timeGranularity.equals(MetricSlice.NATIVE_GRANULARITY) && timeGranularity.getUnit() == TimeUnit.DAYS) {
```
jihaozh: Suggest putting this condition into a helper function. For example:

```
private boolean isMultiDayGranularity(){
        return !timeGranularity.equals(MetricSlice.NATIVE_GRANULARITY) && timeGranularity.getUnit() == TimeUnit.DAYS;
}
```

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/HoltWintersDetector.java
jihaozh: Pinot doesn't support monthly aggregation from the query. Monthly detection is only doable if only the dataset itself is marked as daily granularity but only have one data point per month. So for monthly detection, when fetching data should still use native granularity, but the algorithm needs to be aware that it's a monthly granularity data and fetches the right training data to construct the correct baseline. 

@tangdian I'm not sure this will work for the monthly dataset. looks like you are still retrieving 60 days' training data. Then it's likely that there only 1 data point in there. Since this is not required, you can throw a NotImpelemented exception in this case if it's hard to make it work.

tangdian: According to Jihao, monthly data from pinot will only have monthly data points. i.e. It will not do aggregation. So we will set timeGranularity to NATIVE_GRANULARITY.

tangdian: > Pinot doesn't support monthly aggregation from the query. Monthly detection is only doable if only the dataset itself is marked as daily granularity but only have one data point per month. So for monthly detection, when fetching data should still use native granularity, but the algorithm needs to be aware that it's a monthly granularity data and fetches the right training data to construct the correct baseline.
> 
> @tangdian I'm not sure this will work for the monthly dataset. looks like you are still retrieving 60 days' training data. Then it's likely that there only 1 data point in there. Since this is not required, you can throw a NotImpelemented exception in this case if it's hard to make it work.

I have updated the code to support 1_MONTHS data

tangdian:
I have updated the code to support 1_MONTHS data

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/HoltWintersDetector.java
```diff
@@ -177,10 +199,12 @@ public TimeSeries computePredictedTimeSeries(MetricSlice slice) {
     df.addSeries(COL_DIFF_VIOLATION, df.getDoubles(COL_DIFF).abs().gte(df.getDoubles(COL_ERROR)));
     df.mapInPlace(BooleanSeries.ALL_TRUE, COL_ANOMALY, COL_PATTERN, COL_DIFF_VIOLATION);
 
+    System.out.println(df);
```
jihaozh: please remove

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/HoltWintersDetector.java
```diff
@@ -514,6 +561,10 @@ private static double sensitivityToZscore(double sensitivity) {
     return z;
   }
 
+  private boolean isMultiDayGranularity() {
```
jihaozh: maybe add a comment here. for example. check if the granularity is daily.

PR #4151, opened 2019-04-20
[TE] Add upper/lower/current to HoltWinters baseline

xiaohui-sun:
Add upper/lower/current to HoltWinters baseline.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/HoltWintersDetector.java
```diff
@@ -416,14 +423,26 @@ public boolean apply(long... values) {
...
+      double predicted = result.getPredictedValue();
+      double error = result.getErrorBound();
 
-      resultArray[k] = result.getPredictedValue();
-      errorArray[k] = result.getErrorBound();
+      // if current value doesn't have data then impute with 0
+      if (k < y.length) {
+        currentArray[k] = y[k];
+      } else {
+        currentArray[k] = 0;
```
jihaozh: Should this be null instead? Impute with 0 might cause the current value to be wrong because it's computed from the imputed data.

xiaohui-sun: good point. will change it to null.

PR #4152, opened 2019-04-20
Fix the issue where ZkCacheBaseDataAccessor.getChildren() can return list with null znRecords

Jackie-Jiang:
This can happen if the record gets removed while calling this method.
Need to handle it to prevent NPE.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/metadata/ZKMetadataProvider.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/metadata/ZKMetadataProvider.java
@@ -297,21 +297,28 @@ public class ZKMetadataProvider {
    */
   public static List<OfflineSegmentZKMetadata> getOfflineSegmentZKMetadataListForTable(
       ZkHelixPropertyStore<ZNRecord> propertyStore, String tableName) {
-    List<OfflineSegmentZKMetadata> resultList = new ArrayList<>();
-    if (propertyStore == null) {
-      return resultList;
-    }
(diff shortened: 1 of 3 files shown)


pinot-common/src/main/java/org/apache/pinot/common/metadata/ZKMetadataProvider.java
```diff
@@ -297,70 +297,88 @@ public static Schema getTableSchema(@Nonnull ZkHelixPropertyStore<ZNRecord> prop
    */
   public static List<OfflineSegmentZKMetadata> getOfflineSegmentZKMetadataListForTable(
```
sunithabeeram: Should the method signature (here and below) be updated to reflect that propertyStore is @NotNullable?

pinot-common/src/main/java/org/apache/pinot/common/metadata/ZKMetadataProvider.java
```diff
@@ -297,70 +297,88 @@ public static Schema getTableSchema(@Nonnull ZkHelixPropertyStore<ZNRecord> prop
...
       ZkHelixPropertyStore<ZNRecord> propertyStore, String tableName) {
```
sunithabeeram: Should the signature include @Nonnull for propertyStore?

Jackie-Jiang: There was a discussion before about how to use the **@Nullable** and **@NonNull** annotation. Because almost all arguments should be non-null, we decide to only use **@Nullable** annotation to mark arguments or return value that could be null.

PR #4154, opened 2019-04-22
[TE] avoid tuning the detector and baseline provider separately if both interfaces are implemented in the same class

jihaozh:
Previously, the baseline provider and detector are decoupled into separated class and tunning need to be triggered for both detector and baseline provider. This PR makes the following optimization: If both interfaces are implemented by the same detector class, the tunable will be only triggered once, which improves the performance by 50%.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/annotation/registry/DetectionRegistry.java
```diff
@@ -51,6 +54,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(DetectionRegistry.class);
   private static final String KEY_CLASS_NAME = "className";
   private static final String KEY_ANNOTATION = "annotation";
+  private static final String KEY_IS_BASELINE_PROVIDER = "isBaseline";
```
xiaohui-sun: "isBaselineProvider"

PR #4156, opened 2019-04-23
Refactor HelixExternalViewBasedTimeBoundaryService to support all time units

Jackie-Jiang:
Currently we pick the segment end time as the time boundary, and
append filter 'timeColumn < boundary' to offline table and filter
'timeColumn >= boundary' to realtime table to achieve the hybrid
table federation. The problem with this is that, if the time unit
is not DAYS (for example, MILLISECONDS), and the offline table has
multiple daily segments to push, then we might get incomplete
result before all offline segments are pushed.

The solution is: always use (end time - 1 HOUR/DAY) as the time
boundary, append filter 'timeColumn <= boundary' to offline table
and 'timeColumn > boundary' to realtime table. This can ensure
all HOURLY and DAILY pushed segments be covered regardless of the
time unit, but also use as much offline side data as possible.

For HOURLY push table with time unit other than DAYS, use
(maxTimeValue - 1 HOUR) as the time boundary; otherwise, use
(maxTimeValue - 1 DAY)

Also, add checks for the time consistency between table config
and schema. We should use the time spec in schema as the source of
truth for time column because data is generated based on the
schema. In the future we might remove the timeType field from the
SegmentsValidationAndRetentionConfig.


pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
```diff
@@ -48,77 +48,84 @@ public HelixExternalViewBasedTimeBoundaryService(ZkHelixPropertyStore<ZNRecord>
...
 
-    TableConfig offlineTableConfig = ZKMetadataProvider.getOfflineTableConfig(_propertyStore, tableName);
-    assert offlineTableConfig != null;
-    TimeUnit tableTimeUnit = offlineTableConfig.getValidationConfig().getTimeType();
-    if (tableTimeUnit == null) {
-      LOGGER.info("Skipping updating time boundary service for table '{}' because time unit is not set", tableName);
+    Schema schema = ZKMetadataProvider.getTableSchema(_propertyStore, tableNameWithType);
+    assert schema != null;
+    // TODO: support SDF
+    String timeColumn = schema.getTimeColumnName();
```
mcvsubbu: I thought we were tending towards the time column definition in the tableConfig rather than schema (and all derived time columns in the schema). Please check with @npawar

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -428,8 +428,7 @@ private BrokerRequest getRealtimeBrokerRequest(BrokerRequest hybridBrokerRequest
   private void attachTimeBoundary(String rawTableName, BrokerRequest brokerRequest, boolean isOfflineRequest) {
     TimeBoundaryService.TimeBoundaryInfo timeBoundaryInfo =
         _timeBoundaryService.getTimeBoundaryInfoFor(TableNameBuilder.OFFLINE.tableNameWithType(rawTableName));
-    if (timeBoundaryInfo == null || timeBoundaryInfo.getTimeColumn() == null
-        || timeBoundaryInfo.getTimeValue() == null) {
+    if (timeBoundaryInfo == null) {
```
mcvsubbu: if we skipped setting the time boundary as in line 110 of HelixExternalviewBasedTimeBoundaryService.java, then is it still guaranteed that timeBoundaryInfo.getTimeValue() is non-null?

Jackie-Jiang: If we skip setting time boundary, then timeBoundaryInfo will be null. It is guaranteed that time column and time value are non-null in timeBoundaryInfo.

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
```diff
@@ -48,77 +48,84 @@ public HelixExternalViewBasedTimeBoundaryService(ZkHelixPropertyStore<ZNRecord>
...
+
+    // Skip real-time table, only use offline table to update the time boundary
+    if (TableNameBuilder.getTableTypeFromTableName(tableNameWithType) == TableType.REALTIME) {
       return;
     }
 
     Set<String> offlineSegmentsServing = externalView.getPartitionSet();
     if (offlineSegmentsServing.isEmpty()) {
-      LOGGER.info("Skipping updating time boundary service for table '{}' with no offline segments.", tableName);
+      LOGGER.warn("Skip updating time boundary for table: '{}' with no offline segment", tableNameWithType);
```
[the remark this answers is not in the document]
Jackie-Jiang: Done

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
Jackie-Jiang: Discussed with @npawar. When we start using dateTime, pick the time column from table config, and use the DateTimeFieldSpec from the schema to determine the time unit. For now, added a TODO.

Jackie-Jiang:
@mcvsubbu Let's have a team discussion on the retention and make a decision.

pinot-broker/src/test/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryServiceTest.java
```diff
@@ -0,0 +1,164 @@
...
+        new HelixExternalViewBasedTimeBoundaryService(_propertyStore);
+
+    int tableIndex = 0;
+    for (TimeUnit timeUnit : TimeUnit.values()) {
+      String rawTableName = "table" + tableIndex;
+      String offlineTableName = TableNameBuilder.OFFLINE.tableNameWithType(rawTableName);
+      String realtimeTableName = TableNameBuilder.REALTIME.tableNameWithType(rawTableName);
+      addTableConfig(rawTableName, timeUnit, "daily");
+      addSchema(rawTableName, timeUnit);
+      addSegmentZKMetadata(rawTableName, 10 + tableIndex, timeUnit);
```
mcvsubbu: If you can please add a final int 10 and final int 9 that would also explain where these came from. It will help if the test fails for any reason. It will also help if we want to modify the test. thanks,

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
```diff
@@ -48,77 +50,106 @@ public HelixExternalViewBasedTimeBoundaryService(ZkHelixPropertyStore<ZNRecord>
...
+    // TODO: when we start using dateTime, pick the time column from the retention config, and use the DateTimeFieldSpec
+    //       from the schema to determine the time unit
+    // TODO: support SDF
+    TableConfig tableConfig = ZKMetadataProvider.getTableConfig(_propertyStore, tableNameWithType);
+    assert tableConfig != null;
+    SegmentsValidationAndRetentionConfig retentionConfig = tableConfig.getValidationConfig();
+    String timeColumn = retentionConfig.getTimeColumnName();
+    TimeUnit tableTimeUnit = retentionConfig.getTimeType();
+    if (timeColumn == null || tableTimeUnit == null) {
+      LOGGER.error("Skipping updating time boundary for table: '{}' because time column/unit is not set",
```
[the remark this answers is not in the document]
Jackie-Jiang: We'll reach here only if the table is a hybrid table. IMO the check should be on the caller side, but not in this method (we check that in the caller already). In order to check if both offline and realtime table exist, we need another ZK fetch, which seems a little bit redundant to me.

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
```diff
@@ -48,77 +50,106 @@ public HelixExternalViewBasedTimeBoundaryService(ZkHelixPropertyStore<ZNRecord>
...
+          tableNameWithType);
       return;
     }
 
-    // Bulk reading all segment zk-metadata at once is more efficient than reading one at a time.
+    Schema schema = ZKMetadataProvider.getTableSchema(_propertyStore, tableNameWithType);
```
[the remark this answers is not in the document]
Jackie-Jiang: This is for the TODO in line 66. Because segment is generated based on the schema, we should always pick the time column and unit from schema instead of the table config. In the future if we have multiple time columns (DateTimeField) in schema, we can use the time column from table config to determine which one to use as the main time column, but the time unit should still from schema.
Because we want to follow the time field spec in schema in the future, I'll change it to use time column/unit from schema, and keep the checks so that we can find the inconsistency.

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
```diff
@@ -48,77 +50,106 @@ public HelixExternalViewBasedTimeBoundaryService(ZkHelixPropertyStore<ZNRecord>
...
+      // Only consider segments in the external view
+      if (!offlineSegmentsServing.contains(segmentName)) {
+        LOGGER.warn("Skipping processing segment: '{}' for table: '{}' because it does not exist in the external view",
+            segmentName, tableNameWithType);
+        continue;
+      }
 
-    LOGGER.info("Updated time boundary service for table '{}', maxTime: {}", tableName, maxTimeValue);
-  }
+      // Check if time unit in segment ZK metadata matches table time unit
```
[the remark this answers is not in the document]
Jackie-Jiang: Wanted to catch the inconsistency between ZK metadata and schema, but seems not necessary, removed.

pinot-broker/src/test/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryServiceTest.java
Jackie-Jiang: Done

Jackie-Jiang:
@mcvsubbu Addressed all the comments.

PR #4158, opened 2019-04-23
[TE] fix detection alert filter recipients class casting issue

jihaozh:
Previously, a runtime casting exception will be thrown when running PerUserDimensionAlertFilter. This PR fixes the issue.


jihaozh:
> What is the casting error? Can we not cast and pass the values?

When running the code, the constructor is called with to/cc/bcc as a list instead of set. I think it makes sense to accept any collection.

PR #4159, opened 2019-04-24
Add multiple locks which can be obtained based on segment name in lookupOrCreateFSM

npawar:
Adding segment name based locks instead of synchronizing the method lookupOrCreateFSM. The latter prevents even other segments from entering the method and accessing their own fsm. A stall in a single call (could be on zk access, gc) can create a ripple effect delay in all segments waiting on that method

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
@@ -19,12 +19,15 @@
 package org.apache.pinot.controller.helix.core.realtime;
 
 import com.google.common.annotations.VisibleForTesting;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import org.apache.helix.HelixManager;
 import org.apache.helix.ZNRecord;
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
```diff
@@ -90,6 +95,8 @@ public SegmentCompletionManager(HelixManager helixManager, PinotLLCRealtimeSegme
     _controllerLeadershipManager = controllerLeadershipManager;
     SegmentCompletionProtocol
         .setMaxSegmentCommitTimeMs(TimeUnit.MILLISECONDS.convert(segmentCommitTimeoutSeconds, TimeUnit.SECONDS));
+    _fsmLocks = new Lock[NUM_FSM_LOCKS];
+    Arrays.fill(_fsmLocks, new ReentrantLock());
```
Jackie-Jiang: This will fill all locks with the same lock object

npawar:
> I might be wrong here. Cant parse the indentation properly on phone.
> 
> There is possibility that fsm can be null in the end rt? I thought it’s better to lookup the fsmmap in the end.

The idea is to lookup in the fsmMap, and use that if not null. If it is null, then create a new instance of fsm and put it in the map. The fsm cannot be null after the if-else if-else block. Which place are you referring to?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
npawar: Thanks for catching that! changed

kishoreg:
My bad, I thought there were two if blocks (outer and inner).  In other words, we dont need a lock to check the existence in fsmMap we need the lock if the fsm does not exists.

PR #4166, opened 2019-04-25
[TE] add threshold for dimension explore

xiaohui-sun:
1. Add threshold for dimension explore.
2. Expose the inner exception to wrapper.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
@@ -70,6 +70,11 @@ public class DimensionWrapper extends DetectionPipeline {
   private static final String PROP_NESTED_METRIC_URNS = "nestedMetricUrns";
 
   private static final String PROP_CLASS_NAME = "className";
+
+  // Max number of dimension combinations we can handle.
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -197,12 +206,29 @@ public DetectionPipelineResult run() throws Exception {
...
+  }
+
+  @Override
+  public DetectionPipelineResult run() throws Exception {
+    List<MetricEntity> nestedMetrics = dimensionExplore();
+    if (nestedMetrics.isEmpty()) {
+      return new DetectionPipelineResult(Collections.<MergedAnomalyResultDTO>emptyList(), -1);
+    }
+    if (nestedMetrics.size() > MAX_DIMENSION_COMBINATIONS) {
+      throw new DetectionPipelineException(String.format(
```
jihaozh: Should we cap the max number of dimensions instead of stopping the pipeline?

xiaohui-sun: I would prefer to fail it explicitly.

xiaohui-sun: If we want to cap it should be configured in Yaml.

PR #4167, opened 2019-04-25
[TE] Show more debug information in preview

xiaohui-sun:
1. Print up to 5 stack messages in preview.
2. Change the fake detection id to 0.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
@@ -576,7 +576,7 @@ public class YamlResource {
       Map<String, Object> newDetectionConfigMap = new HashMap<>(ConfigUtils.getMap(this.yaml.load(payload)));
       DetectionConfigDTO detectionConfig = buildDetectionConfigFromYaml(tuningStart, tuningEnd, newDetectionConfigMap, null);
       Preconditions.checkNotNull(detectionConfig);
-      detectionConfig.setId(Long.MAX_VALUE);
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -576,7 +576,7 @@ public Response yamlPreviewApi(
       Map<String, Object> newDetectionConfigMap = new HashMap<>(ConfigUtils.getMap(this.yaml.load(payload)));
       DetectionConfigDTO detectionConfig = buildDetectionConfigFromYaml(tuningStart, tuningEnd, newDetectionConfigMap, null);
       Preconditions.checkNotNull(detectionConfig);
-      detectionConfig.setId(Long.MAX_VALUE);
+      detectionConfig.setId(0L);
```
jihaozh: I remember we had some issues before to set this to 0. Could we keep this unchanged?

xiaohui-sun: Do you know what issue?
Long.MAX_VALUE is very long in alert message. How about setting it to -1?

xiaohui-sun: ok, I reverted first.

jihaozh: Setting to 0 will cause fetching all the anomalies from the database in preview. we can keep it for now.

PR #4168, opened 2019-04-26
[TE] Clean up anomalyTimelinesView for anomalies order than 90 days

xiaohui-sun:
1. Remove anomalyTimelinesView if anomaly is older than 90 days.
2. Remove legacy code which cleans up raw anomalies.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/monitor/MonitorTaskRunner.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/monitor/MonitorTaskRunner.java
@@ -111,9 +111,10 @@ public class MonitorTaskRunner implements TaskRunner {
     }
   }
 
-  private int cleanUpMergedAnomalies(int cleanUpDays) {
-    DateTime cleanupStartDate = new DateTime().withTimeAtStartOfDay().minusDays(cleanUpDays);
-    DateTime cleanupEndDate = cleanupStartDate.plusDays(1);
+  private int cleanupMergedAnomalies(int cleanupDays, int cleanupWindow) {
+    // will clean up anomalies with (cleanupDays - cleanupWindow, cleanupDays)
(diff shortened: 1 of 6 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/monitor/MonitorTaskRunner.java
```diff
@@ -108,6 +111,28 @@ private void executeMonitorUpdate(MonitorTaskInfo monitorTaskInfo) {
     }
   }
 
+  private int cleanUpMergedAnomalies(int cleanUpDays) {
+    DateTime cleanupStartDate = new DateTime().withTimeAtStartOfDay().minusDays(cleanUpDays);
+    DateTime cleanupEndDate = cleanupStartDate.plusDays(1);
```
jihaozh: Looks like we are cleaning up the anomalies in the 90th-day before today. What about the anomalies before that?

jihaozh: Also, how often the monitoring job runs? If it's running daily and failing on someday, there might still be some anomalies left to be not cleaned. 

Can we consider the following options?
- Instead of plus 1 day, we can add some additional lookback to make it more robust.
- Get the last successful monitoring job, and clean up from that time till the 90 days before.

xiaohui-sun: The clean window is 1 day and in a rolling window fashion. The anomalies before that should already been cleaned.

xiaohui-sun: Good points. 
By looking at the last successful monitoring job is also not robust since this clean may fail as well.
Will add more buffer but still want to make it lightweight.

npawar:

PR #4169, opened 2019-04-26
Update Audience and Stability annotations for basic classes

sunithabeeram:
First pass on adding annotations for the most basic classes.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BrokerRequestHandler.java
@@ -21,11 +21,15 @@ package org.apache.pinot.broker.requesthandler;
 import com.fasterxml.jackson.databind.JsonNode;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
+import org.apache.pinot.annotations.InterfaceAudience;
+import org.apache.pinot.annotations.InterfaceStability;
 import org.apache.pinot.broker.api.RequestStatistics;
 import org.apache.pinot.broker.api.RequesterIdentity;
 import org.apache.pinot.common.response.BrokerResponse;
 
 
+@InterfaceAudience.Public
+@InterfaceStability.Stable
 @ThreadSafe
 public interface BrokerRequestHandler {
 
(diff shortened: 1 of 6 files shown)


pinot-common/src/main/java/org/apache/pinot/filesystem/PinotFSFactory.java
```diff
@@ -23,10 +23,14 @@
 import java.util.Iterator;
 import java.util.Map;
 import org.apache.commons.configuration.Configuration;
+import org.apache.pinot.annotations.InterfaceAudience;
+import org.apache.pinot.annotations.InterfaceStability;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
+@InterfaceAudience.Public
```
Jackie-Jiang: I won't count this class an interface that we should apply the annotations.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BrokerRequestHandler.java
```diff
@@ -21,11 +21,15 @@
...
 import javax.annotation.concurrent.ThreadSafe;
+import org.apache.pinot.annotations.InterfaceAudience;
+import org.apache.pinot.annotations.InterfaceStability;
 import org.apache.pinot.broker.api.RequestStatistics;
 import org.apache.pinot.broker.api.RequesterIdentity;
 import org.apache.pinot.common.response.BrokerResponse;
 
 
+@InterfaceAudience.Public
+@InterfaceStability.Stable
```
Jackie-Jiang: Let's not annotate this class for now. I don't know if start() and shutDown() should be in the interface

sunithabeeram: I can mark it evolving - but it is something that is being depended on outside of the project, so we should atleast make that aspect clear.

Jackie-Jiang: I don't know if that is true because so far we don't allow user to plug in their own request handler.

sunithabeeram: Right, its not pluggable currently. I will remove the annotation.

PR #4171, opened 2019-04-26
[TE] UI: Add predicted baseline to preview

tangdian:
@harleyjj I am going to hand this off to you. I have made the graph show upper bound and lower bound series, but it hasn't fill in  the space between upper and lower bound as the design does. It needs some hack to realize.


xiaohui-sun:
@tangdian  can you add a screenshot of current graph?

PR #4173, opened 2019-04-29
Reduce ZK access

jackjlli:
Currently `SegmentDeletionManager` fetches IdealState and ExternalView two times. 
This PR reduces the ZK access.


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/SegmentDeletionManager.java
```diff
@@ -101,8 +101,9 @@ public void run() {
   protected synchronized void deleteSegmentFromPropertyStoreAndLocal(String tableName, Collection<String> segmentIds,
       long deletionDelay) {
     // Check if segment got removed from ExternalView and IdealStates
```
Jackie-Jiang: Can you replace 'and' with 'or' to reflect the code correctly?

jackjlli: Done

PR #4175, opened 2019-04-29
Set forceDelete to false for deleteTaskQueue

jackjlli:
Setting force delete to true may cause a race condition where the Controller might write a deleted workflow back to ZK because it's still caching it.

This PR sets this variable to false.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/minion/PinotHelixTaskResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/minion/PinotHelixTaskResourceManager.java
@@ -46,7 +46,8 @@ import org.slf4j.LoggerFactory;
  */
 public class PinotHelixTaskResourceManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(PinotHelixTaskResourceManager.class);
-
+  // 3 minutes, which is the same as the one for monitoring workflow or job state in Helix code.
+  private final static int DEFAULT_TIMEOUT = 3 * 60 * 1000;
   // Do not change this because Helix uses the same separator
   public static final String TASK_NAME_SEPARATOR = "_";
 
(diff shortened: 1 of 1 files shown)


Jackie-Jiang:
This is changing the behavior of deleting the task queue. Task queue might stuck in bad state and never recover before (might be fixed in new Helix version), so we want to remove it regardless of whether the tasks are finished or not. Let's make sure we can get the expected behavior here.

jackjlli:
Discussed with Helix committer, this is an expected race condition. This happens because the force delete causes inconsistency between the in-memory cache and ZK and shouldn't generally be used. They've agreed to update the JavaDoc.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/minion/PinotHelixTaskResourceManager.java
```diff
@@ -139,11 +140,19 @@ public synchronized void resumeTaskQueue(@Nonnull String taskType) {
...
    * @param taskType Task type
    */
-  public synchronized void deleteTaskQueue(@Nonnull String taskType) {
+  public synchronized void deleteTaskQueue(@Nonnull String taskType)
+      throws InterruptedException {
     String helixJobQueueName = getHelixJobQueueName(taskType);
     LOGGER.info("Deleting task queue: {} for task type: {}", helixJobQueueName, taskType);
-    // NOTE: set force delete to true to remove the task queue from ZooKeeper immediately
-    _taskDriver.delete(helixJobQueueName, true);
+    // NOTE: set force delete to true to remove the task queue from ZooKeeper immediately.
```
sunithabeeram: The comments here don't see to make sense. Can you remove them?

jackjlli: Done.

jackjlli:
Closing this PR since the flaky API has been commented in integration test.

Jackie-Jiang:
Don't close this one. We want to fix the API and test it in the integration test once Helix side issue is resolved.

jackjlli:
This issue is to track whether the API will be fixed once new Helix release is bumped up in Pinot:
https://github.com/apache/incubator-pinot/issues/4268

After new Helix release gets bumped, we need to uncomment this test:
https://github.com/apache/incubator-pinot/pull/4207

PR #4176, opened 2019-04-29
[TE] anomaly detector interface change and implementation for rule-based detection

jihaozh:
- Update the anomaly detector interface to return a detection result which includes anomalies and time series (containing time stamps, predicted baseline, current value, upper/lower bounds)
- Implement the new detector interface in Threshold, Percentage Change, Absolute Change, and Holt-Winters detector.
- Implement the baseline provider interfaces in Threshold, Percentage Change, Absolute Change classes.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/AbsoluteChangeRuleDetector.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/AbsoluteChangeRuleDetector.java
@@ -26,8 +26,10 @@ import org.apache.pinot.thirdeye.common.time.TimeGranularity;
 import org.apache.pinot.thirdeye.dashboard.resources.v2.BaselineParsingUtils;
 import org.apache.pinot.thirdeye.dataframe.BooleanSeries;
 import org.apache.pinot.thirdeye.dataframe.DataFrame;
+import org.apache.pinot.thirdeye.dataframe.DoubleSeries;
 import org.apache.pinot.thirdeye.dataframe.util.MetricSlice;
 import org.apache.pinot.thirdeye.datalayer.dto.DatasetConfigDTO;
+import org.apache.pinot.thirdeye.datalayer.dto.MergedAnomalyResultDTO;
 import org.apache.pinot.thirdeye.detection.DetectionUtils;
 import org.apache.pinot.thirdeye.detection.InputDataFetcher;
 import org.apache.pinot.thirdeye.detection.Pattern;
@@ -46,6 +48,7 @@ import org.apache.pinot.thirdeye.rootcause.impl.MetricEntity;
 import org.apache.pinot.thirdeye.rootcause.timeseries.Baseline;
 import org.joda.time.Interval;
 
+import static org.apache.pinot.thirdeye.dataframe.DoubleSeries.*;
 import static org.apache.pinot.thirdeye.dataframe.util.DataFrameUtils.*;
 
 
@@ -70,7 +73,6 @@ public class AbsoluteChangeRuleDetector implements AnomalyDetector<AbsoluteChang
   private TimeGranularity timeGranularity;
 
   private static final String COL_CURR = "current";
-  private static final String COL_BASE = "baseline";
   private static final String COL_ANOMALY = "anomaly";
   private static final String COL_DIFF = "diff";
   private static final String COL_PATTERN = "pattern";
@@ -85,11 +87,11 @@ public class AbsoluteChangeRuleDetector implements AnomalyDetector<AbsoluteChang
     InputData data = this.dataFetcher.fetchData(new InputDataSpec().withTimeseriesSlices(slices).withMetricIdsForDataset(
         Collections.singletonList(slice.getMetricId())));
     DataFrame dfCurr = data.getTimeseries().get(slice).renameSeries(COL_VALUE, COL_CURR);
-    DataFrame dfBase = this.baseline.gather(slice, data.getTimeseries()).renameSeries(COL_VALUE, COL_BASE);
+    DataFrame dfBase = this.baseline.gather(slice, data.getTimeseries());
 
     DataFrame df = new DataFrame(dfCurr).addSeries(dfBase);
     // calculate absolute change
-    df.addSeries(COL_DIFF, df.getDoubles(COL_CURR).subtract(df.get(COL_BASE)));
+    df.addSeries(COL_DIFF, df.getDoubles(COL_CURR).subtract(df.get(COL_VALUE)));
 
     // defaults
     df.addSeries(COL_ANOMALY, BooleanSeries.fillValues(df.size(), false));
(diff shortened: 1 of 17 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/DetectionResult.java
```diff
@@ -0,0 +1,59 @@
+package org.apache.pinot.thirdeye.detection.spi.model;
```
xiaohui-sun: Add headers.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/DetectionResult.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import org.apache.pinot.thirdeye.datalayer.dto.MergedAnomalyResultDTO;
+
+
+/**
+ * The detection result
```
xiaohui-sun: Better to add more description.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/AbsoluteChangeRuleDetector.java
```diff
@@ -105,11 +109,44 @@
 
     // make anomalies
     DatasetConfigDTO datasetConfig = data.getDatasetForMetricId().get(me.getId());
-    return DetectionUtils.makeAnomalies(slice, df, COL_ANOMALY, window.getEndMillis(),
+    List<MergedAnomalyResultDTO> anomalies = DetectionUtils.makeAnomalies(slice, df, COL_ANOMALY, window.getEndMillis(),
         DetectionUtils.getMonitoringGranularityPeriod(monitoringGranularity, datasetConfig), datasetConfig);
+    return DetectionResult.from(anomalies, TimeSeries.fromDataFrame(constructAbsoluteChangeBoundaries(df)));
```
xiaohui-sun: I would suggest to put baseline calculation in a separate line so it is easier to debug.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/AbsoluteChangeRuleDetector.java
```diff
@@ -105,11 +109,44 @@
...
 
   @Override
+  public TimeSeries computePredictedTimeSeries(MetricSlice slice) {
+    InputData data = this.dataFetcher.fetchData(new InputDataSpec().withTimeseriesSlices(this.baseline.scatter(slice)));
+    DataFrame dfBase = this.baseline.gather(slice, data.getTimeseries());
+    return TimeSeries.fromDataFrame(constructAbsoluteChangeBoundaries(dfBase));
+  }
+
+  private DataFrame constructAbsoluteChangeBoundaries(DataFrame dfBase) {
+    if (!Double.isNaN(this.absoluteChange)) {
```
xiaohui-sun: Why we have the check here? When will absoluteChange would be NaN?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/ThresholdRuleDetector.java
```diff
@@ -87,8 +91,38 @@
...
+  @Override
+  public TimeSeries computePredictedTimeSeries(MetricSlice slice) {
+    InputData data =
+        this.dataFetcher.fetchData(new InputDataSpec().withTimeseriesSlices(Collections.singletonList(slice)));
+    DataFrame df = data.getTimeseries().get(slice);
+    return TimeSeries.fromDataFrame(constructThresholdBoundaries(df));
+  }
+
+  private DataFrame constructThresholdBoundaries(DataFrame df) {
+    if (!Double.isNaN(this.min)) {
```
xiaohui-sun: When will this.min be NaN?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/ThresholdRuleDetector.java
```diff
@@ -87,8 +91,38 @@
...
+      df.addSeries(COL_LOWER_BOUND, DoubleSeries.fillValues(df.size(), this.min));
+    }
+    if (!Double.isNaN(this.max)) {
+      df.addSeries(COL_UPPER_BOUND, DoubleSeries.fillValues(df.size(), this.max));
+    }
+    df.addSeries(COL_VALUE, DoubleSeries.fillValues(df.size(), predictedValue()));
+    return df;
+  }
+
+  private double predictedValue() {
```
xiaohui-sun: There should be no predicted value (or the predicted value is the current value) for threshold based detector.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/TimeSeries.java
```diff
@@ -62,17 +71,34 @@ private static void addSeries(TimeSeries ts, DataFrame df, String name) {
...
+    return ts;
+  }
+
   /**
    * Add DataFrame into TimeSeries.
    * @param df The source DataFrame.
    * @return TimeSeries that contains the predicted values.
    */
   public static TimeSeries fromDataFrame(DataFrame df) {
+    Preconditions.checkArgument(df.contains(COL_TIME));
```
xiaohui-sun: Good job to have this precondition check!

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -187,7 +187,7 @@ public DetectionPipelineResult run() throws Exception {
         LOG.warn("[DetectionConfigID{}] detecting anomalies for window {} to {} failed.", this.config.getId(), window.getStart(), window.getEnd(), e);
         lastException = e;
       }
-      anomalies.addAll(anomaliesForOneWindow);
+      anomalies.addAll(detectionResult.getAnomalies());
```
xiaohui-sun: I am confused here only the anomalies are merged.
What about the baselines in detectionResult?

xiaohui-sun:
Thanks for adding this. Seems this is not a complete solution. Do you plan to add more PRs later to expose DetectionResult out?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/AbsoluteChangeRuleDetector.java
jihaozh: By default, it's NaN which means not set. If it's NaN, no anomaly will be generated.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/ThresholdRuleDetector.java
jihaozh: By default, it's NaN which means not set. It's possible only one side threshold is set

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/ThresholdRuleDetector.java
jihaozh: Good point. will change to the same value as the current.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
jihaozh: Those will be handled in a follow-up PR.

jihaozh:
> Thanks for adding this. Seems this is not a complete solution. Do you plan to add more PRs later to expose DetectionResult out?

Yes. This PR focus on the interface change and implementation for the rule-based detectors.
I will submit follow up PRs to do the following:
- Implement the interface in ALGORITHM detectors.
- Change the detection pipeline to handle the baseline time series.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/AbsoluteChangeRuleDetector.java
xiaohui-sun: I see. It is a little weird to use NaN as default value.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/ThresholdRuleDetector.java
jihaozh: One thing is that, if we set the predicted value the same as current value, the percentage change will be 0 in UI.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/ThresholdRuleDetector.java
xiaohui-sun: Got it. Thanks.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/ThresholdRuleDetector.java
xiaohui-sun: Why we show percentage change for threshold based rules?

xiaohui-sun: Let's discuss that offline and maybe submit another PR to fix this. @jihaozh

PR #4182, opened 2019-05-02
Generate inverted index in purge task if it exists

Jackie-Jiang:
If purge task does not generate inverted index, then server needs
to generate it while loading the segment, which can potentailly
cause performance issue.

TODO: uniform the behavior of Pinot Hadoop, segment converter and
purger

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/minion/SegmentPurger.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/minion/SegmentPurger.java
@@ -21,12 +21,14 @@ package org.apache.pinot.core.minion;
 import com.google.common.base.Preconditions;
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/org/apache/pinot/core/minion/SegmentPurger.java
```diff
@@ -100,14 +105,28 @@ public File purgeSegment()
         config.setSegmentTimeUnit(segmentMetadata.getTimeUnit());
       }
 
+      // Generate inverted index if it exists in the original segment
+      // TODO: once the column metadata correctly reflects whether inverted index exists for the column, use that
```
sunithabeeram: +1

mcvsubbu: In v3 segments we shiould be able to read the index file to see if inverted index exists for the column?

Jackie-Jiang: SegmentDirectory is an abstraction layer on top of files, which can handle all segment format properly.

PR #4190, opened 2019-05-07
[TE] pass predicted time series thought out the detection pipeline

jihaozh:
We've changed the detector interface to pass predicted time series from detector to detection pipeline. This PR updates each stage of the detection pipeline to pass the time series. This makes the time series available in preview endpoints. Also, this makes the evaluation metrics calculation possible.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
@@ -19,6 +19,7 @@
 
 package org.apache.pinot.thirdeye.detection;
 
+import java.util.Collections;
 import org.apache.pinot.thirdeye.datalayer.dto.MergedAnomalyResultDTO;
 import java.util.HashMap;
 import java.util.List;
@@ -29,38 +30,42 @@ public class DetectionPipelineResult {
   public static String DIAGNOSTICS_DATA = "data";
   public static String DIAGNOSTICS_CHANGE_POINTS = "changepoints";
 
-  Map<String, Object> diagnostics;
-  List<MergedAnomalyResultDTO> anomalies;
-  long lastTimestamp;
+  private final List<MergedAnomalyResultDTO> anomalies;
+  private final long lastTimestamp;
+  private final List<PredictionResult> predictedTimeSeries;
+  private Map<String, Object> diagnostics;
 
   public DetectionPipelineResult(List<MergedAnomalyResultDTO> anomalies) {
     this.anomalies = anomalies;
     this.lastTimestamp = getMaxTime(anomalies);
     this.diagnostics = new HashMap<>();
+    this.predictedTimeSeries = Collections.emptyList();
   }
 
   public DetectionPipelineResult(List<MergedAnomalyResultDTO> anomalies, long lastTimestamp) {
     this.anomalies = anomalies;
     this.lastTimestamp = lastTimestamp;
+    this.predictedTimeSeries = Collections.emptyList();
+    this.diagnostics = new HashMap<>();
+  }
+
+  public DetectionPipelineResult(List<MergedAnomalyResultDTO> anomalies, long lastTimestamp,
+      List<PredictionResult> predictedTimeSeries) {
+    this.anomalies = anomalies;
+    this.lastTimestamp = lastTimestamp;
+    this.predictedTimeSeries = predictedTimeSeries;
+    this.diagnostics = new HashMap<>();
   }
 
   public List<MergedAnomalyResultDTO> getAnomalies() {
     return anomalies;
   }
 
-  public DetectionPipelineResult setAnomalies(List<MergedAnomalyResultDTO> anomalies) {
-    this.anomalies = anomalies;
-    return this;
-  }
 
   public long getLastTimestamp() {
     return lastTimestamp;
   }
 
-  public DetectionPipelineResult setLastTimestamp(long lastTimestamp) {
-    this.lastTimestamp = lastTimestamp;
-    return this;
-  }
 
   public Map<String, Object> getDiagnostics() {
     return diagnostics;
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
@@ -30,10 +30,10 @@ public class DetectionPipelineResult {
   public static String DIAGNOSTICS_DATA = "data";
   public static String DIAGNOSTICS_CHANGE_POINTS = "changepoints";
 
+  private Map<String, Object> diagnostics;
   private final List<MergedAnomalyResultDTO> anomalies;
   private final long lastTimestamp;
   private final List<PredictionResult> predictions;
-  private Map<String, Object> diagnostics;
 
   public DetectionPipelineResult(List<MergedAnomalyResultDTO> anomalies) {
     this.anomalies = anomalies;
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
@@ -31,21 +31,20 @@ public class DetectionPipelineResult {
   public static String DIAGNOSTICS_CHANGE_POINTS = "changepoints";
 
   private Map<String, Object> diagnostics;
+  // detected anomalies
   private final List<MergedAnomalyResultDTO> anomalies;
+  // last time stamp, all data point before this time stamp has been inspected
   private final long lastTimestamp;
+  // predicted baselines result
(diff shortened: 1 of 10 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -201,8 +206,58 @@ public DetectionPipelineResult run() throws Exception {
...
+    return new DetectionPipelineResult(anomalyResults, lastTimeStamp, Collections.singletonList(predictedTimeSeries));
+  }
+
+  /**
+   * Join two time series, including current, baseline, lower bound and upper bound. If two time series have overlapped region, take the average
+   * @param ts1 timeseries 1
+   * @param ts2 timeseries 2
+   * @return the consolidated time series
+   */
+  static TimeSeries consolidateTimeSeries(TimeSeries ts1, TimeSeries ts2) {
```
xiaohui-sun: Is it better to put this join into TimeSeries object?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -201,8 +206,58 @@ public DetectionPipelineResult run() throws Exception {
...
+    }
+  }
+
+  /**
+   * Calculate the average of two double time series. If the value in either one is missing, take the available value as the result
+   * @param s1 series 1
+   * @param s2 series 2
+   * @return the averaged time series
+   */
+  private static DoubleSeries robustAverage(DoubleSeries s1, DoubleSeries s2) {
```
xiaohui-sun: It would be hard to debug if using Average.
How about always use the latest window to  overwrite?

E.g, if we want to merge two windows
1. t1 to t2
2. t3 to t4, and t1 < t3 < t2
Then use the values t3 to t4 to overwrite the value for t2  to t3.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
jihaozh: For now, this is logic is only used here and seems very specific for this purpose. What about we put it here for now, and refactor this to be a util function if there is another use case.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
jihaozh: Sure. I can make it that way, as long as the prediction result makes sense.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
```diff
@@ -29,38 +30,46 @@
...
 
   public DetectionPipelineResult(List<MergedAnomalyResultDTO> anomalies) {
     this.anomalies = anomalies;
     this.lastTimestamp = getMaxTime(anomalies);
     this.diagnostics = new HashMap<>();
+    this.predictions = Collections.emptyList();
   }
 
   public DetectionPipelineResult(List<MergedAnomalyResultDTO> anomalies, long lastTimestamp) {
     this.anomalies = anomalies;
```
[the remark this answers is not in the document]
jihaozh: good point. changed.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineResult.java
```diff
@@ -29,38 +30,46 @@
   public static String DIAGNOSTICS_DATA = "data";
   public static String DIAGNOSTICS_CHANGE_POINTS = "changepoints";
 
-  Map<String, Object> diagnostics;
-  List<MergedAnomalyResultDTO> anomalies;
-  long lastTimestamp;
+  private Map<String, Object> diagnostics;
+  private final List<MergedAnomalyResultDTO> anomalies;
+  private final long lastTimestamp;
```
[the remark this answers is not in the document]
jihaozh: added.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
xiaohui-sun: Makes sense. Let's keep it for now.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
xiaohui-sun: Normally we don't need to merge the windows. It is only used when the detection windows overlap. Also better to check with AI team on what is the best way.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
```diff
@@ -201,8 +206,58 @@ public DetectionPipelineResult run() throws Exception {
...
+   * Join two time series, including current, baseline, lower bound and upper bound. If two time series have overlapped region, take the value in the right time series
+   * @param leftTimeSeries timeseries 1
+   * @param rightTimeSeries timeseries 2
+   * @return the consolidated time series
+   */
+  static TimeSeries consolidateTimeSeries(TimeSeries leftTimeSeries, TimeSeries rightTimeSeries) {
+    DataFrame df1 = leftTimeSeries.getDataFrame();
+    DataFrame df2 = rightTimeSeries.getDataFrame();
+    DataFrame joinedDf = df1.joinOuter(df2, COL_TIME);
+    consolidateJoinedDf(joinedDf, COL_VALUE);
```
xiaohui-sun: Can we simply use joinOuter?

jihaozh: joinOuter will generate two double series. For example, "current_left" and "current_right". Then we need to consolidate the result into one series "current".

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
jihaozh: Sure. Last time I check with them, they say we can either take the average or take the lasted values. I can double check this.

xiaohui-sun: ok let's just take the latest values. Technically they should be the same.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
xiaohui-sun: I see. So we don't have the native support in dataframe to merge the series?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
jihaozh: yes. they should be the same.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
jihaozh: yes.

xiaohui-sun: Since we simplified the join logic (not using the average), let's try to use DataFrame APIs to do this.

jihaozh: updated to utilize built-in data frame groupBy aggregation for this purpose.

xiaohui-sun: That's great. Thanks.

PR #4197, opened 2019-05-09
Add SimpleAvroMessageDecoder which allows passing in Avro schema directly

Jackie-Jiang:
For cases where users have avro encoded msg in kafka but not yet setup a schema service to fetch schema on the fly.
Contributor: fx19880617

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/SimpleAvroMessageDecoder.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/SimpleAvroMessageDecoder.java
@@ -1,0 +1,84 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/SimpleAvroMessageDecoder.java
```diff
@@ -0,0 +1,77 @@
...
+    _avroRecordConverter = new AvroRecordToPinotRowGenerator(indexingSchema);
+  }
+
+  @Override
+  public GenericRow decode(byte[] payload, GenericRow destination) {
+    return decode(payload, 0, payload.length, destination);
+  }
+
+  @Override
+  public GenericRow decode(byte[] payload, int offset, int length, GenericRow destination) {
```
sunithabeeram: Referencing code from KafkaAvroMessageDecoder, it looks like we are not skipping over the HEADER. If the user of this API is expected to provide the right offset, it might be good to call that out.

sunithabeeram:
A few thoughts on this:
- Have we considered updating KafkaAvroMessageDecoder to work with this case? I do see that class being tied up quite a bit with schema registry, but I don't see why that can't be made cleaner and able to handle the case this class tries to.
- This is more of a comment on KafkaAvroMessageDecoder, but I guess we could reuse the GenericDatumReader there unless the schema has changed (which should be rare and we can avoid the extra object allocation on each decode)
- Depending on how important this particular implementation is, it will be good to add a test to guard against regressions.

Jackie-Jiang:
@sunithabeeram This is a significantly simplified version of Avro decoder where schema is directly provided (not tight to kafka as well), so I would rather put it separate from the heavy used KafkaAvroMessageDecoder with all the complex schema registry logic.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/kafka/SimpleAvroMessageDecoder.java
Jackie-Jiang: Added notes in the javadoc

PR #4198, opened 2019-05-10
Enhance the support for BYTES data type

Jackie-Jiang:
1. Add BytesUtils to handle encode/decode for Hex string
2. Add FieldSpec.DataType.convert() to convert string value to data type
3. Support Hex string for index(), indexOf(), inRange() in all BYTES dictionaries
4. Add MutableDictionary.compare(dictId1, dictId2) to help the sort process and support sorting byte[]
5. Support returning Hex string in group-by for byte[] column
6. In MutableSegmentImpl, uniform the sort for all types, reduce the memory usage especially for STRING and BYTES columns

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
@@ -22,11 +22,9 @@ import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import javax.annotation.Nullable;
 import org.apache.avro.Schema.Type;
-import org.apache.commons.codec.DecoderException;
-import org.apache.commons.codec.binary.Hex;
-import org.apache.pinot.common.Utils;
 import org.apache.pinot.common.config.ConfigKey;
 import org.apache.pinot.common.config.ConfigNodeLifecycleAware;
+import org.apache.pinot.common.utils.BytesUtils;
 import org.apache.pinot.common.utils.EqualityUtils;
 import org.apache.pinot.common.utils.JsonUtils;
 
@@ -168,7 +166,7 @@ public abstract class FieldSpec implements Comparable<FieldSpec>, ConfigNodeLife
    */
   protected static String getStringValue(Object value) {
     if (value instanceof byte[]) {
-      return Hex.encodeHexString((byte[]) value);
+      return BytesUtils.toHexString((byte[]) value);
     } else {
       return value.toString();
     }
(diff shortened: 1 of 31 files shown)


pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
```diff
@@ -439,6 +415,28 @@ public int size() {
           throw new IllegalStateException("Cannot get number of bytes for: " + this);
       }
     }
+
+    /**
+     * Converts the given string value to the data type.
+     */
+    public Object convert(String value) {
```
kishoreg: probably not part of this PR but why is this method part of the FieldSpec?

pinot-common/src/main/java/org/apache/pinot/common/utils/BytesUtils.java
```diff
@@ -0,0 +1,56 @@
...
+   */
+  public static byte[] toBytes(Object rawValue) {
+    if (rawValue instanceof String) {
+      try {
+        return Hex.decodeHex(((String) rawValue).toCharArray());
+      } catch (DecoderException e) {
+        throw new IllegalArgumentException("Value: " + rawValue + " is not Hex encoded", e);
+      }
+    } else {
+      return (byte[]) rawValue;
```
kishoreg: might need additional safety check if it can be cast to byte[]

pinot-common/src/main/java/org/apache/pinot/common/utils/primitive/ByteArray.java
```diff
@@ -45,19 +45,13 @@ public int length() {
...
-  /**
-   * Static utility function to convert a byte[] to Hex string.
-   *
-   * @param bytes byte[] to convert
-   * @return Equivalent Hex String.
-   */
-  public static String toHexString(byte[] bytes) {
-    return Hex.encodeHexString(bytes);
+  public String toHexString() {
+    return BytesUtils.toHexString(_bytes);
```
kishoreg: why did we move this to BytesUtils class?

pinot-core/src/main/java/org/apache/pinot/core/query/pruner/AbstractSegmentPruner.java
```diff
@@ -87,26 +89,17 @@ protected boolean pruneNonLeaf(@Nonnull FilterQueryTree filterQueryTree,
...
-          return Long.valueOf(input);
-        case FLOAT:
-          return Float.valueOf(input);
-        case DOUBLE:
-          return Double.valueOf(input);
-        case STRING:
-          return input;
-        default:
-          throw new IllegalStateException("Unsupported data type: " + dataType);
+      if (dataType != FieldSpec.DataType.BYTES) {
```
kishoreg: Is there a way we can eliminate this special check for BYTES?

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
Jackie-Jiang: This is part of the DataType, which is inside FieldSpec.

pinot-common/src/main/java/org/apache/pinot/common/utils/BytesUtils.java
Jackie-Jiang: If it cannot be cast, it will throw ClassCastException, which is the expected behavior.

pinot-common/src/main/java/org/apache/pinot/common/utils/primitive/ByteArray.java
Jackie-Jiang: BytesUtils is the util class for all byte[] related methods. ByteArray is the comparable wrapper for byte[], which should not be used as util class.

pinot-core/src/main/java/org/apache/pinot/core/query/pruner/AbstractSegmentPruner.java
Jackie-Jiang: If we only keep ByteArray as the value inside Pinot, we can achieve that. Right now we have mixed of byte[] and ByteArray, thus hard to uniform all behaviors. We might want to change the code to use only ByteArray internal in a separate PR.

Jackie-Jiang: Internally we mix the use of byte[] and ByteArray, which causes this special handling. IMO we should uniform the behavior to always use ByteArray (at least for segment generation, and filters where comparison applies). That will involve much bigger changes, which can be submitted in a separate PR if we decide to do so.

PR #4203, opened 2019-05-10
[TE] Threshold filter on current value of an anomaly

jihaozh:
This PR make the threshold filter to filter on the current value of the anomaly.


akshayrai:
Can you update the doc?

jihaozh:
> Can you update the doc?

Sure

PR #4205, opened 2019-05-13
use conf start pinot

sunny209868:
use conf start pinot controller,  broker, server

### DIFF
--- a/docs/in_production.rst
+++ b/docs/in_production.rst
@@ -39,8 +39,31 @@ In general, when deploying Pinot services, it is best to adhere to a specific or
 The ordering is as follows:
 
 #. pinot-controller
+vim conf/default_controller.properties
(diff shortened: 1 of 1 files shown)


Jackie-Jiang:
I don't think the changes are following the rst format. Also, IMO these should be part of the quick start instead of the running Pinot in production.

mcvsubbu:
> I don't think the changes are following the rst format. Also, IMO these should be part of the quick start instead of the running Pinot in production.

+1

sunny209868:
> I don't think the changes are following the rst format. Also, IMO these should be part of the quick start instead of the running Pinot in production.

yes, the format can be modified. Currently, from the pinot document, we use this kind of startup in the production environment. How do you start it? @Jackie-Jiang @mcvsubbu

PR #4206, opened 2019-05-14
[TE] frontend - harleyjj/create-alert - break yaml editor into two co…

harleyjj:
…mponents

1) breaks yaml-editor into two separate components: detection-yaml and subscription-yaml
2) Installs new components on create-alert
3) Moves preview directly onto create-alert and migrates shared props to controller

These actions will help prepare the create-alert route for forms (full ui).
The next step is to install new components on edit alert.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/route.js
@@ -4,7 +4,7 @@
  * @exports alert create model
  */
 import fetch from 'fetch';
-import RSVP from 'rsvp';
+import { hash } from 'rsvp';
 import moment from 'moment';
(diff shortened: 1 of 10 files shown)


thirdeye/thirdeye-frontend/app/pods/self-serve/create-alert/route.js
```diff
@@ -19,18 +19,22 @@ import { get } from '@ember/object';
...
+  anomaliesApiService: service('services/api/anomalies'),
   session: service(),
 
   /**
    * Model hook for the create alert route.
    * @method model
    * @return {Object}
    */
-  model(params, transition) {
+  async model(params, transition) {
```
jihaozh: Nice refactor to use the async/await key words!

PR #4208, opened 2019-05-14
Increase the limit of groups in combine operator (inter-segments)

Jackie-Jiang:
Right now we limit number of groups stored both inner-segment and
inter-segments with the same limit. Once the segment results reach
the inner-segment limit, it will also reach the inter-segments
limit after combining the results in the combine operator. This
will cause the new groups from the other segment being dropped.
This might cause inconsistent result depending on the order of
segment execution.
The solution would be, increase the limit (multiply by factor 2) of
groups across segments to get consistent result regardless of the
order of execution. For most cases, most groups from each segment
should be the same, thus the total number of groups across segments
should be equal or slightly higher than the number of groups in
each segment. We still put a limit across segments to protect cases
where data is very skewed across different segments.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/CombineGroupByOperator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/CombineGroupByOperator.java
@@ -54,10 +54,17 @@ public class CombineGroupByOperator extends BaseOperator<IntermediateResultsBloc
   private static final Logger LOGGER = LoggerFactory.getLogger(CombineGroupByOperator.class);
   private static final String OPERATOR_NAME = "CombineGroupByOperator";
 
+  // Use a higher limit for groups stored across segments. For most cases, most groups from each segment should be the
+  // same, thus the total number of groups across segments should be equal or slightly higher than the number of groups
+  // in each segment. We still put a limit across segments to protect cases where data is very skewed across different
+  // segments.
+  private static final int INTER_SEGMENT_NUM_GROUPS_LIMIT_FACTOR = 2;
+
   private final List<Operator> _operators;
   private final BrokerRequest _brokerRequest;
   private final ExecutorService _executorService;
   private final long _timeOutMs;
+  // Limit on number of groups stored for each segment, beyond which no new group will be created
   private final int _numGroupsLimit;
 
   public CombineGroupByOperator(List<Operator> operators, BrokerRequest brokerRequest, ExecutorService executorService,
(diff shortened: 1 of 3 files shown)


mayankshriv:
What about some heuristic like dropping x% of bottom groups each time the limit is reached?
A better alternative is to use penalize groups each time a new group is seen, and then drop the groups that reach zero. This would be costly to implement, and also is known to work mostly on frequency, not sure about aggregation functions such as sum.

Jackie-Jiang:
@mayankshriv There is no easy way to drop groups since we should drop groups based on the value of the groups. Once we support the normal sql semantic of group-by (keep the same groups for all aggregations and use same key to order them), we can implement what you suggested.

kishoreg:
Removing the limit is definitely a concern. We don't want one bad query to kill the server with OOM.

I like @mayankshriv  suggestion of dropping bottom x% every time we reach the limit.  Not sure why this would be expensive.

 100k limit for group by is already a  big number. is there a use case that is hitting this limit?

Another naive solution would be to set the Intersegment limit to be double the inner segment limit.

Jackie-Jiang:
@kishoreg For most of the queries, removing the inter-segments limit won't add new keys because all the keys are common across segments. If keys are not common across segments, then the data is very skewed, thus limiting the key can give very wrong result.
Actually there are several use cases hitting this limit. With this change, at least we can guarantee consistent result running the same query.
The problem with @mayankshriv suggestion is that, as I mentioned, with the current semantic, we don't know how to define the bottom x% if there are multiple aggregations.

kishoreg:
I dont understand why someone would be fetching 100k group by values with Pinot. It's definitely not the right use of Pinot.

How about making it configurable and override it as needed. Let's leave the default as it is at 100k for inner and maybe 200k for inter?

sunithabeeram:
@kishoreg I don't believe that users are fetching >100k groups. But given the current implementation, its possible that incorrect group-by results are returned because many groups are not considered.

kishoreg:
@sunithabeeram I didn't follow that comment, do you have a concrete example. Let's say user does group by x,y,z. the max cardinality of the group by keys is cardinality(x) * cardinality(y) * cardinality(z).
So the only time we reach this limit is the unique group by keys cross 100k within the segment and 200k across segments.

 If our goal is to consider all groups then the only solution is to remove all limits but then we risk one bad query crashing the system.

Jackie-Jiang:
@kishoreg It's not about fetching 100k groups, it's about there are more than 100k unique groups inside the segment, and we only keep the first 100k unique groups in the memory. The heuristic here is that if the group is not in the first 100k groups, then it's not important and we can drop it. GROUP BY some id can easily reach this limit.
Adding inter-segment limit is unnecessary for most cases, but I can use 2*inner-segment limit for extreme corner cases.

Jackie-Jiang:
@kishoreg Can you have another look on the PR?

PR #4209, opened 2019-05-14
[TE] Model evaluator interface & Implementation

jihaozh:
- The boilerplate code to retrieve the evaluations in detection components, including dataProvider, InputDataFetcher, etc.
- Define the model evaluator interface, which read the evaluation metrics and evaluate the detection model status.
- Implement the MAPE percentage change evaluator, which compare the mean MAPE over past 7 days and past 30 days, if the up percentage change is more than a certain threshold, returns a bad detection status.

Will submit the following PR to call the model evaluators and automatically tunes the detection model based on the evaluation result.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MapePercentageChangeModelEvaluator.java
```diff
@@ -0,0 +1,102 @@
...
+ *  auto configuration.
+ */
+public class MapePercentageChangeModelEvaluator implements ModelEvaluator<MapePercentageChangeModelEvaluatorSpec> {
+  private InputDataFetcher dataFetcher;
+  private double threshold;
+
+  @Override
+  public ModelEvaluationResult evaluateModel(Instant evaluationTimeStamp) {
+    EvaluationSlice evaluationSlice =
+        new EvaluationSlice().withStartTime(evaluationTimeStamp.toDateTime().minusDays(30).getMillis())
```
xiaohui-sun: Please put 30 and 7 as private final values.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MapePercentageChangeModelEvaluator.java
```diff
@@ -0,0 +1,102 @@
...
+    Collection<EvaluationDTO> evaluations =
+        this.dataFetcher.fetchData(new InputDataSpec().withEvaluationSlices(Collections.singleton(evaluationSlice)))
+            .getEvaluations()
+            .get(evaluationSlice);
+    // calculate past 7 day mean MAPE for each metric urn
+    Map<String, Double> sevenDaysMeanMapeForMetricUrns =
+        getMeanMapeForEachMetricUrn(evaluations, evaluationTimeStamp, 7);
+
+    // calculate past 30 day mean MAPE for each metric urn
+    Map<String, Double> thirtyDaysMeanMapeForMetricUrns =
```
xiaohui-sun: What if we don't have enough data for evaluation?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/ModelEvaluationResult.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ *
+ */
+
+package org.apache.pinot.thirdeye.detection.spi.model;
+
+/**
+ * The model valuation result
+ */
+public class ModelEvaluationResult {
```
xiaohui-sun: Can you put more information like 7 days MAPE, 30 days MAPE in this result? Currently it only returns GOOD/BAD which is too simple.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/ModelStatus.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+
+package org.apache.pinot.thirdeye.detection.spi.model;
+
+/**
+ * The detection model status.
+ */
+public enum ModelStatus {
+  GOOD, BAD
```
xiaohui-sun: Should we have an unknown status? Several cases for unknown: 
1. Data is missing
2. Too new to evaluate

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MapePercentageChangeModelEvaluator.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.pinot.thirdeye.detection.spi.model.ModelStatus;
+import org.joda.time.Instant;
+
+
+/**
+ *  Monitor the recent mean MAPE in last 7 days, and compare that with the mean MAPE for the last 30 days.
+ *  If the percentage change dropped to a certain threshold for a metric urn, return a bad model status to trigger
+ *  auto configuration.
+ */
+public class MapePercentageChangeModelEvaluator implements ModelEvaluator<MapePercentageChangeModelEvaluatorSpec> {
```
xiaohui-sun: MeanAveragePercentageChangeModelEvaluator

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/MapePercentageChangeModelEvaluatorSpec.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ */
+
+package org.apache.pinot.thirdeye.detection.spec;
+
+/**
+ * The spec class for MAPE change evaluator
+ */
+public class MapePercentageChangeModelEvaluatorSpec extends AbstractSpec {
+  private double threshold;
```
xiaohui-sun: Do we have a default value for threshold?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MapePercentageChangeModelEvaluator.java
```diff
@@ -0,0 +1,102 @@
...
+
+    // evaluate for each metric urn
+    Map<String, Boolean> evaluationResultForMetricUrns = sevenDaysMeanMapeForMetricUrns.entrySet()
+        .stream()
+        .collect(Collectors.toMap(Map.Entry::getKey,
+            // compare the MAPE percentage change to threshold
+            sevenDayMeanMape -> sevenDayMeanMape.getValue() / thirtyDaysMeanMapeForMetricUrns.get(sevenDayMeanMape.getKey()) - 1 <= threshold));
+
+    if (evaluationResultForMetricUrns.values().stream().allMatch(result -> result)) {
+      // if all metric urn's status is good, return overall good status
```
xiaohui-sun: This condition seems too restrict for multidimensional metrics. 
Let's keep an eye on it.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/EvaluationSlice.java
```diff
@@ -0,0 +1,68 @@
...
+  private final long end;
+
+  private EvaluationSlice(long start, long end) {
+    this.start = start;
+    this.end = end;
+  }
+
+  public EvaluationSlice() {
+    // -1 means match any
+    this(-1, -1);
```
xiaohui-sun: How about:
EvaluationSlice(Long.MIN_VALUE, Long.MAX_VALUE)

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/EvaluationSlice.java
```diff
@@ -0,0 +1,68 @@
...
+  public long getStart() {
+    return start;
+  }
+
+  public long getEnd() {
+    return end;
+  }
+
+  public boolean match(EvaluationDTO evaluationDTO) {
+    if (this.start >= 0 && evaluationDTO.getEndTime() <= this.start)
```
xiaohui-sun: If we use max/min value for default then this could be simplified as
return evaluationDTO.getStartTime() >= this.start && evaluationDTO.getStartTime() <= this.end

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DataProvider.java
```diff
@@ -136,4 +139,15 @@
...
+  /**
+   * Returns a multimap of evaluations (keyed by the evaluations slice) for a given set of evaluations slices.
+   *
+   * @see Evaluation
+   *
+   * @param evaluationSlices the evaluation slices
+   * @param configId configId
+   * @return a multimap of evaluations (keyed by the evaluations slice)
+   */
+  Multimap<EvaluationSlice, EvaluationDTO> fetchEvaluations(Collection<EvaluationSlice> evaluationSlices, long configId);
```
xiaohui-sun: What if there are multiple rules in one configId? How do we handle this?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MapePercentageChangeModelEvaluator.java
jihaozh: Good point. I'll add an UNKNOWN status and return it in that case.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/MapePercentageChangeModelEvaluatorSpec.java
jihaozh: I can set the default threshold here. What would be a good threshold? Maybe 10%?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MapePercentageChangeModelEvaluator.java
jihaozh: Yeah. In the next PR, I'll add a filed to record the last tuning time.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/EvaluationSlice.java
jihaozh: Good idea. However, the existing convention for Anomaly and Events are using -1 as select all. Maybe it's better to keep the existing convention because it's been used in many codes?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/MapePercentageChangeModelEvaluatorSpec.java
xiaohui-sun: Sounds good. Let's use 10% first.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DataProvider.java
jihaozh: Good point. This data fetcher will fetch the evaluations for all the rules. Maybe when evaluating models, we should group by the rule names as well.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MapePercentageChangeModelEvaluator.java
jihaozh: renamed

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/EvaluationSlice.java
xiaohui-sun: Sounds good.

PR #4211, opened 2019-05-16
StreamPartitionAssignmentStrategy interface to plug in various partition assignment strategies (uniform, replica group based)

npawar:
Phase 1 - Step 2 of https://github.com/apache/incubator-pinot/issues/4192

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/partition/ReplicaGroupBasedStreamPartitionAssignmentStrategy.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/partition/ReplicaGroupBasedStreamPartitionAssignmentStrategy.java
@@ -1,3 +1,21 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 package org.apache.pinot.common.partition;
 
 import com.google.common.annotations.VisibleForTesting;
--- a/pinot-common/src/main/java/org/apache/pinot/common/partition/ReplicaGroupBasedStreamPartitionAssignmentStrategy.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/partition/ReplicaGroupBasedStreamPartitionAssignmentStrategy.java
@@ -46,21 +46,18 @@ public class ReplicaGroupBasedStreamPartitionAssignmentStrategy implements Strea
       throw new InvalidConfigException("ReplicaGroupPartitionAssignment is null for table:" + tableNameWithType);
     }
     int numReplicaGroups = replicaGroupPartitionAssignment.getNumReplicaGroups();
-    if (numReplicaGroups != numReplicas) {
-      throw new InvalidConfigException(
-          "numReplicas:" + numReplicas + " is not equal to numReplicaGroups:" + numReplicaGroups
-              + " from znode for table:" + tableNameWithType);
-    }
+    // TODO: we might move to a model of not having the same numInstancesPerReplicaGroup.
+    // We would have to handle numInstancesInReplicaGroup on a replica group by replica group basis, and uniformly assign withing each replica group
     int numInstancesPerReplicaGroup = replicaGroupPartitionAssignment.getInstancesFromReplicaGroup(0, 0).size();
 
     PartitionAssignment streamPartitionAssignment = new PartitionAssignment(tableNameWithType);
 
     List<List<String>> verticalSlices = new ArrayList<>(numInstancesPerReplicaGroup);
     for (int i = 0; i < numInstancesPerReplicaGroup; i++) {
-      verticalSlices.add(new ArrayList<>(numReplicas));
+      verticalSlices.add(new ArrayList<>(numReplicaGroups));
     }
 
-    for (int replicaGroupNumber = 0; replicaGroupNumber < numReplicas; replicaGroupNumber++) {
+    for (int replicaGroupNumber = 0; replicaGroupNumber < numReplicaGroups; replicaGroupNumber++) {
       List<String> instancesFromReplicaGroup =
           replicaGroupPartitionAssignment.getInstancesFromReplicaGroup(0, replicaGroupNumber);
       for (int serverIndex = 0; serverIndex < numInstancesPerReplicaGroup; serverIndex++) {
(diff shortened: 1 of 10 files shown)


pinot-common/src/main/java/org/apache/pinot/common/partition/StreamPartitionAssignmentStrategyFactory.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class StreamPartitionAssignmentStrategyFactory {
+
+  /**
+   * Given a table config, get the {@link StreamPartitionAssignmentStrategy}
+   */
+  static StreamPartitionAssignmentStrategy getStreamPartitionAssignmentStrategy(TableConfig tableConfig) {
+    SegmentsValidationAndRetentionConfig validationConfig = tableConfig.getValidationConfig();
+    if (validationConfig != null) {
+      String segmentAssignmentStrategy = validationConfig.getSegmentAssignmentStrategy();
```
kishoreg: why is this part of validationConfig?

pinot-common/src/main/java/org/apache/pinot/common/partition/StreamPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import javax.annotation.Nonnull;
+import org.apache.helix.HelixManager;
+import org.apache.pinot.common.exception.InvalidConfigException;
+
+
+/**
+ * Creates a partition assignment for the partitions of a realtime table
+ */
+// TODO: Unify the interfaces for {@link org.apache.pinot.controller.helix.core.sharding.SegmentAssignmentStrategy} and {@link StreamPartitionAssignmentStrategy}
```
kishoreg: thanks for this comment, I was wondering why do we need another interface.

pinot-common/src/main/java/org/apache/pinot/common/partition/StreamPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+public interface StreamPartitionAssignmentStrategy {
+
+  /**
+   * Given the list of partitions and replicas, come up with a {@link PartitionAssignment}
+   */
+  PartitionAssignment getStreamPartitionAssignment(HelixManager helixManager, @Nonnull String tableNameWithType,
```
kishoreg: it might be useful to pass the current partition assignment. We might add some smarts to minimize the shuffle

pinot-common/src/main/java/org/apache/pinot/common/partition/StreamPartitionAssignmentStrategyFactory.java
snleee: `segmentAssignmentStrategy` got added to `SegmentValidationAndRetentionConfig` from the beginning.

pinot-common/src/main/java/org/apache/pinot/common/partition/StreamPartitionAssignmentStrategy.java
mcvsubbu: While it is nice to combine the interfaces, we need to be clear whether it will allow us any re-use. It will be a pity if we have code  like "if (partition assignment) then do this else do that" inside. Here is my reasoning:

Segment assignment strategy is given 1 or more segments, and returns the servers on which they should be hosted. It is invoked each time new segments shows up. It does NOT change any assignment of existing segments. If the segment is a CONSUMING segment (a new one comes up each time previous one in a stream partition completes), then our goal should be to assign it to a set of servers that have previously hosted consuming segments from the table. We currently do the simple strategy of assigning it to the same server that the previous segment completed. So, essentially:

segmentAssignmentStrategy.assign(Collection segments) returns a collection of servers.

On the other hand, (stream) partition assignment strategy should be invoked at the table creation time, each time we get new partitions (not very often). Furthermore, the partition assignment strategy MAY change the existing assignment for partitions previously known. Therefore:

partitionAssignmentStrategy.assign(numPartitions) returns a map of partition to a collection of servers.

@kishoreg  and @Jackie-Jiang let me know if I am off here.

kishoreg: Good point. I am not really looking for merging the interfaces and have only one method. I feel its better to understand the entire lifecyle of partition assignment and segment assignment through one interface and not differentiate between real-time and offline.

To me, having partition assignment should really be optional and really up to the segment assignment strategy if wants to place all segments of the same partition on a particular set of nodes or spread them to different nodes. Technically, we can just assign a new real-time segment to any of the nodes without relying on partition assignment (similar to the way we assign offline segments). 

Replica Group Assignment is also another case of partition assignment strategy and we had to implement them separately for real-time and offline. 

At a high level, we can view partition, replica, tenant, table name, state(consuming/offline/online) etc as just a bunch of attributes on a segment. And the segment assignment strategy is to just pick a bunch of servers given these attributes. What it does internally is up to the strategy for e.g. it can choose to keep partition affinity or replica groups, tenant matching etc.

In other words, I am more towards avoiding if(real-time) than if(partition assignment) which means our first goal should be to have one common interface across offline and real-time. 

```
segmentAssignmentStrategy.init(tableConfig)
List<Instance> segmentAssignmentStrategy.assign(segment, [tenant name, tablename, partition, replicaId, state ...] )
```

pinot-common/src/main/java/org/apache/pinot/common/partition/ReplicaGroupBasedStreamPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,85 @@
...
+    if (replicaGroupPartitionAssignment == null) {
+      throw new InvalidConfigException("ReplicaGroupPartitionAssignment is null for table:" + tableNameWithType);
+    }
+    int numReplicaGroups = replicaGroupPartitionAssignment.getNumReplicaGroups();
+    if (numReplicaGroups != numReplicas) {
+      throw new InvalidConfigException(
+          "numReplicas:" + numReplicas + " is not equal to numReplicaGroups:" + numReplicaGroups
+              + " from znode for table:" + tableNameWithType);
+    }
+    int numInstancesPerReplicaGroup = replicaGroupPartitionAssignment.getInstancesFromReplicaGroup(0, 0).size();
```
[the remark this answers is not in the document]
npawar: That would be a good extension. Will add a TODO indicating that we might want to do this if we want to remove strict requirement that numInstancesPerReplicaGroup is same across all replica groups

pinot-common/src/main/java/org/apache/pinot/common/partition/ReplicaGroupBasedStreamPartitionAssignmentStrategy.java
```diff
@@ -0,0 +1,85 @@
...
+      @Nonnull List<String> partitions, int numReplicas, List<String> instances) throws InvalidConfigException {
+
+    ReplicaGroupPartitionAssignment replicaGroupPartitionAssignment =
+        getReplicaGroupPartitionAssignment(helixManager, tableNameWithType);
+    if (replicaGroupPartitionAssignment == null) {
+      throw new InvalidConfigException("ReplicaGroupPartitionAssignment is null for table:" + tableNameWithType);
+    }
+    int numReplicaGroups = replicaGroupPartitionAssignment.getNumReplicaGroups();
+    if (numReplicaGroups != numReplicas) {
+      throw new InvalidConfigException(
```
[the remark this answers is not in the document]
npawar: Good catch, this check was not the right behavior. Consumption would've broken between updating replicas in table configs and rebalance.
Changed this to not look at numReplicas from table config at all, but instead rely only on the replica group znode. That way the assignment does not know about new servers/replicas, until a rebalance is done (just like offline)

pinot-common/src/main/java/org/apache/pinot/common/partition/StreamPartitionAssignmentStrategy.java
npawar: Added it as a TODO for now, to be done when we decide to use it. Should be fairly easy to bring it in. Did not add it right now, because it would be fair bit of computation each time to compute from ideal state, and better to avoid it until we will actually want to use it

PR #4213, opened 2019-05-16
[TE] Refactor cube algorithm for adding ratio cube in the future.

cyenjung:
__Major changes__:
1. Move cube algorithm related code to package `cube`.
2. Extract multiple classes to interface and implementation classes:
    * `ThirdEyeSummaryClient` --> `CubeDBClient` (interface) and `AdditiveCubeDBClient` (Impl)
    * `Row` --> `Row` (Interface) and `AdditiveRow` (Impl)
    * `HierarchyNode` --> `CubeNode` (Interface) and `AdditiveCubeNode` (Impl)
3. Update related classes for future support of ratio metrics.
4. Update unit tests for the new changes.

__Tests__:
1. Passed the updated unit tests.
2. The integration test is done through ThirdEye API.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/additive/AdditiveDBClient.java
```diff
@@ -0,0 +1,97 @@
...
+        rowTable.put(dimensionValues, row);
+      }
+      switch (tag) {
+        case Baseline:
+          row.setBaselineValue(value);
+          break;
+        case Current:
+          row.setCurrentValue(value);
+          break;
+        default:
```
jihaozh: should it throw an exception in this case?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/BaseCubeNode.java
```diff
@@ -0,0 +1,144 @@
...
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.apache.pinot.thirdeye.cube.data.dbrow.DimensionValues;
+import org.apache.pinot.thirdeye.cube.data.dbrow.Dimensions;
+import org.apache.pinot.thirdeye.cube.data.dbrow.Row;
+
+
+public abstract class BaseCubeNode<N extends BaseCubeNode, R extends Row> implements CubeNode<N> {
```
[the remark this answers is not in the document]
cyenjung: `BaseCubeNode` is a base class it won't be able to be instantiated. Therefore, only some classes that inheritances `BaseCubeNode` uses N. Thus, <N extends BaseCubeNode>.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNode.java
```diff
@@ -0,0 +1,211 @@
...
+   * Resets all values (e.g., baseline and current value) of this node.
+   */
+  void resetValues();
+
+  /**
+   * Updates all values when the child node is extracted from this node.
+   *
+   * @param node the child node to be extracted.
+   */
+  void removeNodeValues(N node);
```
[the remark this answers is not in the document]
cyenjung: Explained in the above comment.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/additive/AdditiveDBClient.java
cyenjung: The default is no-op because tags are shared across additive and ratio metrics. If I throw an exception, then I won't be able to support both cases.

jihaozh: So the `AdditiveDBClient` will also work with ratio metrics?

cyenjung: No, it won't. There will be another `DBClient` called `RatioDBClient`. I see your point now; will amend it.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/BaseCubeNode.java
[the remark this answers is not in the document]
xiaohui-sun: It should be fine as long as you don't instantiate it.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/dbrow/Row.java
```diff
@@ -0,0 +1,78 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.pinot.thirdeye.cube.data.dbrow;
+
+import org.apache.pinot.thirdeye.cube.data.node.CubeNode;
+
+
+public interface Row {
```
xiaohui-sun: Can you add more comments on this  interface?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNode.java
```diff
@@ -0,0 +1,211 @@
...
+
+package org.apache.pinot.thirdeye.cube.data.node;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import java.util.List;
+import org.apache.pinot.thirdeye.cube.data.dbrow.DimensionValues;
+import org.apache.pinot.thirdeye.cube.data.dbrow.Dimensions;
+
+
+public interface CubeNode<N extends CubeNode> {
```
xiaohui-sun: Can you add more comments on this interface?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNode.java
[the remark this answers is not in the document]
xiaohui-sun: Can you double check? @cyenjung

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNodeUtils.java
```diff
@@ -0,0 +1,72 @@
...
+   *
+   * @return true if both hierarchical tree are the same.
+   */
+  public static boolean equalHierarchy(CubeNode node1, CubeNode node2) {
+    return equalHierarchy(node1, null, node2, null);
+  }
+
+  public static boolean equalHierarchy(CubeNode node1, CubeNode node1Parent, CubeNode node2, CubeNode node2Parent) {
+    boolean sameData = ObjectUtils.equals(node1, node2);
+    if (sameData) {
```
xiaohui-sun: It would be more clear if using:
if (!sameData) return false;
// other logics  .....

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNode.java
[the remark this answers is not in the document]
cyenjung: If we use `CubeNode` here, we need to cast it to the same type as the current node. Because removing a ratio cube node from an additive cube node doesn't make sense and vice versa.

To prevent from such mistake happens during runtime (we want to report such error during compile time), it's a better idea to use an explicit class `N`. However, I want to bound it to `BaseCubeNode` instead of any arbitrary class. Hence `N extends BaseCubeNode`.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/BaseCubeNode.java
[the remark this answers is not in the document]
cyenjung: See the comment below, which is a more concrete use case for explanation. Thanks.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/dbrow/Row.java
cyenjung: Sure, I will add it along with the Ratio Cube code. Thanks for noting this.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNode.java
cyenjung: Sure, I will add it along with the Ratio Cube code. Thanks for noting this issue.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNodeUtils.java
cyenjung: Awesome! Will flip it when I am working on the Ratio Cube code.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/dbrow/Row.java
cyenjung: Addressed in the PR for ratio cube algorithm: https://github.com/apache/incubator-pinot/pull/4246

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNode.java
cyenjung: Addressed in the PR for ratio cube algorithm: https://github.com/apache/incubator-pinot/pull/4246

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/data/node/CubeNodeUtils.java
cyenjung: Addressed in the PR for ratio cube algorithm: https://github.com/apache/incubator-pinot/pull/4246

PR #4214, opened 2019-05-16
Support default BYTES (zero-length byte array) in aggregation function and aggregator

Jackie-Jiang:
Treat zero-length byte array as null, and skip processing it
Add DefaultBytesQueriesTest for test coverage

Motivation:
Without this change, in order to create default columns for serialized Object such as HyperLogLog, TDigest etc., user needs to serialize the default Object (empty HyperLogLog or TDigest) to get the default value and put it into the schema. The default values are quite long for these Objects (HyperLogLog: "00000008000000ac00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"; QDigest: "3fa999999999999a0000000000000000000000000004b9467fffffffffffffff800000000000000000000000"; TDigest: "000000017ff0000000000000fff0000000000000405900000000000000000000").
With this change, we directly treat empty byte array as null, so user doesn't need to specify such default values.
Besides, this change allows user to put empty byte array to skip a value which can save storage and reduce computation cost. E.g. for a HyperLogLog column, if there is no value inside, user can simply put an empty array instead of the serialized byte array.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/aggregator/AvgValueAggregator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/aggregator/AvgValueAggregator.java
@@ -21,6 +21,7 @@ package org.apache.pinot.core.data.aggregator;
 import org.apache.pinot.common.data.FieldSpec.DataType;
 import org.apache.pinot.core.common.ObjectSerDeUtils;
 import org.apache.pinot.core.query.aggregation.function.AggregationFunctionType;
+import org.apache.pinot.core.query.aggregation.function.AvgAggregationFunction;
 import org.apache.pinot.core.query.aggregation.function.customobject.AvgPair;
 
 
@@ -40,7 +41,13 @@ public class AvgValueAggregator implements ValueAggregator<Object, AvgPair> {
   @Override
   public AvgPair getInitialAggregatedValue(Object rawValue) {
     if (rawValue instanceof byte[]) {
-      return deserializeAggregatedValue((byte[]) rawValue);
+      // Use default value for zero-length byte array
+      byte[] bytes = (byte[]) rawValue;
+      if (bytes.length != 0) {
+        return deserializeAggregatedValue(bytes);
+      } else {
+        return AvgAggregationFunction.getDefaultAvgPair();
+      }
     } else {
       return new AvgPair(((Number) rawValue).doubleValue(), 1L);
     }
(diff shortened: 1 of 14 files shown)


mayankshriv:
I had a similar PR a while back to filter out zero-length byte array (that represented null values for TDigest in that PR). But based on some offline concerns from @kishoreg that I don't recall right now, I had retracted the PR. Tagging @kishoreg to comment.

kishoreg:
Can you point us to a concrete query/usecase where this PR would help? What would be the behavior without this PR.

Jackie-Jiang:
@kishoreg Without this PR, in order to create default columns for serialized Object such as HyperLogLog, TDigest etc., user needs to serialize the default Object (empty HyperLogLog or TDigest) in order to get the default value and then put it into the schema. The default values are quite long for these Objects (HyperLogLog: "00000008000000ac00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"; QDigest: "3fa999999999999a0000000000000000000000000004b9467fffffffffffffff800000000000000000000000"; TDigest: "000000017ff0000000000000fff0000000000000405900000000000000000000").
With this PR, we directly treat empty byte array as null, so user doesn't need to specify such default values.

kishoreg:
This is equivalent of supporting NULL for BYTES column right?

Jackie-Jiang:
@kishoreg Yes, and we treat empty byte array as NULL for BYTES.

kishoreg:
ok. It's better to support NULL as first-class instead of checking for NULL every row right?. Generate a NULL bitmap for this column and filter it out in the filter phase.

kishoreg:
Let’s create an issue for supporting null and point to this PR.

Jackie-Jiang:
@kishoreg It cannot be done in filter phase because for each document, there is no guarantee that all metrics are null. We can filter them out in data fetching phase though. I think we need to add a new index type or reserve some dictId for null. Will create the issue.

kishoreg:
NULL value is per column. Also, I think I now remember the discussion with Mayank. This will be backward incompatible change when we add NULL support.

Jackie-Jiang:
@kishoreg I understand that NULL value is per column, but for example: SELECT SUM(A), SUM(B) FROM table, for the same document, column A might have NULL value while column B does not. We cannot apply an extra filter in such case.
Can you describe why this will be backward incompatible (forward incompatible maybe as we don't have NULL support yet)?

Jackie-Jiang:
Discussed with @kishoreg offline. Will hold on this PR and see how much effort it takes to support real NULL values, and then decide which route to take.

PR #4215, opened 2019-05-17
Minor fix for star-tree creation logs

Jackie-Jiang:


pinot-core/src/main/java/org/apache/pinot/core/startree/v2/builder/BaseSingleTreeBuilder.java
```diff
@@ -304,7 +304,7 @@ Record mergeStarTreeRecord(@Nullable Record aggregatedRecord, Record starTreeRec
   public void build()
       throws Exception {
     long startTime = System.currentTimeMillis();
-    LOGGER.info("Start building star-trees with config: {}", _builderConfig);
+    LOGGER.info("Start building star-tree with config: {}", _builderConfig);
```
mayankshriv: Nit: Started/Finished.

Jackie-Jiang: Fixed

PR #4217, opened 2019-05-17
[TE] distribute detection and notification tasks

xiaohui-sun:
1. Set cron for detection and notification tasks to start at random seconds.
2. Have a random delay between 0 to 1000 milliseconds when creating the detection and notification tasks.
3. Remove the logic for anomaly.setNotified() in notification task.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineJob.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineJob.java
@@ -83,10 +83,12 @@ public class DetectionPipelineJob implements Job {
     taskDTO.setStatus(TaskConstants.TaskStatus.WAITING);
     taskDTO.setTaskInfo(taskInfoJson);
 
-    // Sleep random 0 - 1000 milliseconds to distribute load to mysql.
+    // Here will write the task information to mysql.
+    // Sleep random 0 - 10 seconds to distribute load to mysql.
     Random random = new Random();
     try {
-      Thread.sleep(random.nextInt(1000));
+      LOG.info("Wait for " + random + " milliseconds.");
+      Thread.sleep(random.nextInt(10000));
       long taskId = taskDAO.save(taskDTO);
       LOG.info("Created detection pipeline task {} with taskId {}", taskDTO, taskId);
     } catch (InterruptedException e) {
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineJob.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineJob.java
@@ -83,12 +83,13 @@ public class DetectionPipelineJob implements Job {
     taskDTO.setStatus(TaskConstants.TaskStatus.WAITING);
     taskDTO.setTaskInfo(taskInfoJson);
 
+    // TODO: revisit it after identifying bottlenecks
     // Here will write the task information to mysql.
-    // Sleep random 0 - 10 seconds to distribute load to mysql.
+    // Sleep random 0 - 5 seconds to distribute load to mysql.
     Random random = new Random();
     try {
       LOG.info("Wait for " + random + " milliseconds.");
-      Thread.sleep(random.nextInt(10000));
+      Thread.sleep(random.nextInt(5000));
       long taskId = taskDAO.save(taskDTO);
       LOG.info("Created detection pipeline task {} with taskId {}", taskDTO, taskId);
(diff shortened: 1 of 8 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
```diff
@@ -379,16 +380,24 @@ private void fillInDetectorWrapperProperties(Map<String, Object> properties, Map
     return properties;
   }
 
+  //  Default schedule:
+  //  minute granularity: every 15 minutes, starts at 0 minute
+  //  hourly: every hour, starts at 0 minute
+  //  daily: every day, starts at 2 pm UTC
+  //  others: every day, start at 12 am UTC
   private String buildCron() {
+    // starts at random second to reduce task spike
```
akshayrai: Can you include a comment on what task spike you are referring to here? and why is that a problem? probably mention how this correlates to high mysql updates.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
```diff
@@ -379,16 +380,24 @@ private void fillInDetectorWrapperProperties(Map<String, Object> properties, Map
...
+    Random random = new Random();
+    String second = Integer.toString(random.nextInt(59));
     switch (this.datasetConfig.bucketTimeGranularity().getUnit()) {
       case MINUTES:
-        return "0 0/15 * * * ? *";
+        return second + " 0/15 * * * ? *";
```
akshayrai: Can we avoid adding buffer time to the cron and instead include the buffer when the task is picked up and executed. I feel it will be cleaner and easier to debug issues that way.

Moreover, we resort to buildCron method only if the cron is not defined in the yaml config. In other cases, the cron will remain plain without the buffer.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/DetectionAlertTaskRunner.java
```diff
@@ -107,13 +107,6 @@ private void updateAlertConfigWatermarks(DetectionAlertFilterResult result, Dete
       DetectionAlertFilter alertFilter = detAlertTaskFactory.loadAlertFilter(alertConfig, System.currentTimeMillis());
       DetectionAlertFilterResult result = alertFilter.run();
 
-      // TODO: The old UI relies on notified tag to display the anomalies. After the migration
-      // we need to clean up all references to notified tag.
-      for (MergedAnomalyResultDTO anomaly : result.getAllAnomalies()) {
```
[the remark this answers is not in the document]
xiaohui-sun: Sure. reverted.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
xiaohui-sun: Makes sense.  Will add more buffer when loading the task.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/CompositePipelineConfigTranslator.java
xiaohui-sun: From database log all tasks are executed within the first three seconds which maps to the spike in mysql write spike since the task object is serialized in the database. We should distribute it.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipelineJob.java
```diff
@@ -82,9 +83,18 @@ public void execute(JobExecutionContext jobExecutionContext) throws JobExecution
...
 
-    long taskId = taskDAO.save(taskDTO);
-    LOG.info("Created detection pipeline task {} with taskId {}", taskDTO, taskId);
-
+    // TODO: revisit it after identifying bottlenecks
+    // Here will write the task information to mysql.
+    // Sleep random 0 - 5 seconds to distribute load to mysql.
+    Random random = new Random();
+    try {
+      LOG.info("Wait for " + random + " milliseconds.");
```
akshayrai: Nit: Store `random.nextInt(5000)` and then log. Here and below.

PR #4218, opened 2019-05-17
Add RealtimeConsumptionCatchupServiceCallback

npawar:
Creating a ServiceCallback within ServiceStatus, to check that realtime consuming segments have caught up. This will help with latency issues which can happen during server startup, because the server has to catchup on a lot of consumption.
For the initial version, this only adds a static wait time (configurable). More smarts can be added later (monitor stability or idleness of stream consumption rate)

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
@@ -240,11 +240,9 @@ public class HelixBrokerStarter {
     LOGGER.info("Registering service status handler");
     double minResourcePercentForStartup = _brokerConf.getDouble(Broker.CONFIG_OF_BROKER_MIN_RESOURCE_PERCENT_FOR_START,
         Broker.DEFAULT_BROKER_MIN_RESOURCE_PERCENT_FOR_START);
-    ServiceStatus.setServiceStatusCallback(new ServiceStatus.MultipleCallbackServiceStatusCallback(ImmutableList
-        .of(new ServiceStatus.IdealStateAndCurrentStateMatchServiceStatusCallback(_participantHelixManager,
-                _clusterName, _brokerId, minResourcePercentForStartup),
-            new ServiceStatus.IdealStateAndExternalViewMatchServiceStatusCallback(_participantHelixManager,
-                _clusterName, _brokerId, minResourcePercentForStartup))));
+    ServiceStatus.setServiceStatusCallback(new ServiceStatus.MultipleCallbackServiceStatusCallback(ImmutableList.of(
+        new ServiceStatus.IdealStateServiceStatusCallback(_participantHelixManager, _clusterName, _brokerId,
+            minResourcePercentForStartup))));
 
     LOGGER.info("Finish starting Pinot broker");
   }
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
@@ -35,6 +35,7 @@ import org.apache.helix.HelixManagerFactory;
 import org.apache.helix.InstanceType;
 import org.apache.helix.SystemPropertyKeys;
 import org.apache.helix.ZNRecord;
+import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.Message;
 import org.apache.helix.participant.StateMachineEngine;
@@ -237,14 +238,30 @@ public class HelixBrokerStarter {
         .addPreConnectCallback(() -> brokerMetrics.addMeteredGlobalValue(BrokerMeter.HELIX_ZOOKEEPER_RECONNECTS, 1L));
 
     // Register the service status handler
-    LOGGER.info("Registering service status handler");
-    double minResourcePercentForStartup = _brokerConf.getDouble(Broker.CONFIG_OF_BROKER_MIN_RESOURCE_PERCENT_FOR_START,
-        Broker.DEFAULT_BROKER_MIN_RESOURCE_PERCENT_FOR_START);
-    ServiceStatus.setServiceStatusCallback(new ServiceStatus.MultipleCallbackServiceStatusCallback(ImmutableList.of(
-        new ServiceStatus.IdealStateServiceStatusCallback(_participantHelixManager, _clusterName, _brokerId,
-            minResourcePercentForStartup))));
+    registerServiceStatusHandler();
 
     LOGGER.info("Finish starting Pinot broker");
+  }
+
+  /**
+   * Fetches the resources to monitor and registers the {@link org.apache.pinot.common.utils.ServiceStatus.ServiceStatusCallback}s
+   */
+  private void registerServiceStatusHandler() {
+    List<String> resourcesToMonitor = new ArrayList<>(1);
+    IdealState resourceIdealState = _helixAdmin.getResourceIdealState(_clusterName, Helix.BROKER_RESOURCE_INSTANCE);
+    if (resourceIdealState != null && resourceIdealState.isEnabled()) {
+      resourcesToMonitor.add(Helix.BROKER_RESOURCE_INSTANCE);
+    }
+
+    double minResourcePercentForStartup = _brokerConf.getDouble(Broker.CONFIG_OF_BROKER_MIN_RESOURCE_PERCENT_FOR_START,
+        Broker.DEFAULT_BROKER_MIN_RESOURCE_PERCENT_FOR_START);
+
+    LOGGER.info("Registering service status handler");
+    ServiceStatus.setServiceStatusCallback(new ServiceStatus.MultipleCallbackServiceStatusCallback(ImmutableList.of(
+        new ServiceStatus.IdealStateAndCurrentStateMatchServiceStatusCallback(_participantHelixManager, _clusterName,
+            _brokerId, resourcesToMonitor, minResourcePercentForStartup),
+        new ServiceStatus.IdealStateAndExternalViewMatchServiceStatusCallback(_participantHelixManager, _clusterName,
+            _brokerId, resourcesToMonitor, minResourcePercentForStartup))));
   }
 
   private void addInstanceTagIfNeeded() {
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
@@ -248,9 +248,14 @@ public class HelixBrokerStarter {
    */
   private void registerServiceStatusHandler() {
     List<String> resourcesToMonitor = new ArrayList<>(1);
-    IdealState resourceIdealState = _helixAdmin.getResourceIdealState(_clusterName, Helix.BROKER_RESOURCE_INSTANCE);
-    if (resourceIdealState != null && resourceIdealState.isEnabled()) {
-      resourcesToMonitor.add(Helix.BROKER_RESOURCE_INSTANCE);
+    IdealState brokerResourceIdealState = _helixAdmin.getResourceIdealState(_clusterName, Helix.BROKER_RESOURCE_INSTANCE);
+    if (brokerResourceIdealState != null && brokerResourceIdealState.isEnabled()) {
+      for (String partitionName : brokerResourceIdealState.getPartitionSet()) {
+        if (brokerResourceIdealState.getInstanceSet(partitionName).contains(_brokerId)) {
+          resourcesToMonitor.add(Helix.BROKER_RESOURCE_INSTANCE);
+          break;
+        }
+      }
     }
 
     double minResourcePercentForStartup = _brokerConf.getDouble(Broker.CONFIG_OF_BROKER_MIN_RESOURCE_PERCENT_FOR_START,
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
@@ -131,7 +131,7 @@ public class TableNameBuilder {
   }
 
   /**
-   * Return whether the given resource name represents an realtime table resource.
+   * Return whether the given resource name represents a realtime table resource.
    */
   public static boolean isRealtimeTableResource(@Nonnull String resourceName) {
     return REALTIME.tableHasTypeSuffix(resourceName);
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
@@ -129,4 +129,11 @@ public class TableNameBuilder {
   public static boolean isOfflineTableResource(@Nonnull String resourceName) {
     return OFFLINE.tableHasTypeSuffix(resourceName);
   }
+
+  /**
+   * Return whether the given resource name represents an realtime table resource.
+   */
+  public static boolean isRealtimeTableResource(@Nonnull String resourceName) {
+    return REALTIME.tableHasTypeSuffix(resourceName);
+  }
 }
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
@@ -187,6 +187,9 @@ public class CommonConstants {
     public static final String CONFIG_OF_SERVER_MIN_RESOURCE_PERCENT_FOR_START =
         "pinot.server.startup.minResourcePercent";
     public static final double DEFAULT_SERVER_MIN_RESOURCE_PERCENT_FOR_START = 100.0;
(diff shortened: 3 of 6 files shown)


pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
```diff
@@ -129,4 +129,11 @@ public static boolean isTableResource(@Nonnull String resourceName) {
   public static boolean isOfflineTableResource(@Nonnull String resourceName) {
     return OFFLINE.tableHasTypeSuffix(resourceName);
   }
+
+  /**
+   * Return whether the given resource name represents an realtime table resource.
```
Jackie-Jiang: nit: **a** realtime table resource

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
       return statusDescription.toString();
     }
   }
+  
+  public static abstract class BaseServiceStatusCallback implements ServiceStatusCallback {
```
Jackie-Jiang: I don't think we need this base class. ServiceStatusCallback does not always associate with Helix cluster info (e.g. MultipleCallbackServiceStatusCallback), also the methods are not very commonly used. I don't like the way of returning statusDescription by setting a member variable from the parent class, it's not very readable. The base class IdealStateMatchServiceStatusCallback is already good enough IMO.

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
...
+   * TODO: In this initial version, we are simply adding a configurable static wait time
+   * This can be made smarter:
+   * 1) Keep track of average consumption rate for table in server stats
+   * 2) Monitor consumption rate during startup, report GOOD when it stabilizes to average rate
+   * 3) Monitor consumption rate during startup, report GOOD if it is idle
+   */
+  public static class RealtimeConsumptionCatchupServiceStatusCallback extends BaseServiceStatusCallback {
+
+    private final Map<String, List<String>> _tableToConsumingSegmentsMap;
+    private long _endWaitTime = 0;
```
Jackie-Jiang: final?

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
...
+        if (TableNameBuilder.isRealtimeTableResource(resourceName)) {
+
+          IdealState idealState = getResourceIdealState(resourceName);
+          if (idealState.isEnabled()) {
+            List<String> consumingSegments = new ArrayList<>();
+            for (String segmentName : idealState.getPartitionSet()) {
+              Map<String, String> instanceStateMap = idealState.getInstanceStateMap(segmentName);
+              if (CommonConstants.Helix.StateModel.RealtimeSegmentOnlineOfflineStateModel.CONSUMING.equals(
+                  instanceStateMap.get(instanceName))) {
+                consumingSegments.add(segmentName);
```
mcvsubbu: I suggest to leave it as a boolean and break if we have even one consuming segment we own.

We can add new logic when we decide to examine each segment as we need to.

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
...
+            if (!consumingSegments.isEmpty()) {
+              _tableToConsumingSegmentsMap.put(resourceName, consumingSegments);
+            }
+          }
+        }
+      }
+
+      if (_tableToConsumingSegmentsMap.isEmpty()) {
+        LOGGER.info(
+            "No consuming segments to monitor on instance. Setting realtime consumption catchup wait time to 0ms");
```
Jackie-Jiang: Directly set the service status to GOOD instead of relying on the _endWaitTime

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
...
+      } else {
+        long startTime = System.currentTimeMillis();
```
mcvsubbu: Should you take the time when this method is entered?

It may also be worth commenting that even in the not-so-fancy determination of  the time a consuming segment is ready to serve queries, the computation should be:
 end-time = (time of creation of partition consumer) + (confugured max time to catch up).

We are sort of approximating it to :+1: 
end-time = (time of server startup) + (configured max timeto catch up)

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
...
+        _endWaitTime = startTime + TimeUnit.SECONDS.toMillis(realtimeConsumptionCatchupWaitTimeSeconds);
+        LOGGER.info("Monitoring realtime consumption catchup for tables:{}. Will wait for time:{} ms",
+            _tableToConsumingSegmentsMap.keySet(), realtimeConsumptionCatchupWaitTimeSeconds);
+      }
+    }
+
+    @Override
+    public Status getServiceStatus() {
+      long now = System.currentTimeMillis();
```
Jackie-Jiang: Check for previous service status first, directly return GOOD if it is already GOOD. (Do not keep updating the statusDescription)

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
...
+        }
+      }
+
+      if (_tableToConsumingSegmentsMap.isEmpty()) {
+        LOGGER.info(
+            "No consuming segments to monitor on instance. Setting realtime consumption catchup wait time to 0ms");
+      } else {
+        long startTime = System.currentTimeMillis();
+        _endWaitTime = startTime + TimeUnit.SECONDS.toMillis(realtimeConsumptionCatchupWaitTimeSeconds);
+        LOGGER.info("Monitoring realtime consumption catchup for tables:{}. Will wait for time:{} ms",
```
mcvsubbu: We are not really waiting here, so the log could be misleading. May be say "will allow until time {} before marking status as GOOD"?

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
...
+            _tableToConsumingSegmentsMap.keySet(), realtimeConsumptionCatchupWaitTimeSeconds);
+      }
+    }
+
+    @Override
+    public Status getServiceStatus() {
+      long now = System.currentTimeMillis();
+      if (now < _endWaitTime) {
+        _statusDescription = String.format("Waiting for CONSUMING segments to catchup, timeRemaining=%dms", _endWaitTime - System.currentTimeMillis());
```
mcvsubbu: Use 'now' instead of calling System.currentMillis again

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -127,34 +129,129 @@ public String getStatusDescription() {
...
+        return Status.STARTING;
+      }
+      _statusDescription = "Wait time for CONSUMING segments catchup expired, now:" + now + ", endWaitTime:" + _endWaitTime;
```
mcvsubbu: I would omit the word 'wait' here (so we can grep for wait and see where waits are happening). How about: "Consuming segments caught up at " + _endWaitTime. 

Or, "Consuming segment status GOOD since " + _endTime.

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
npawar: RealtimeConsumptionCatchupServiceCallback doesn't really belong under IdealStateMatchServiceStatusCallback. What is the suggestion? Keep RealtimeConsumptionCatchupServiceCallback independent and directly implement ServiceStatusCallback interface?

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
npawar: Worked on all suggestions

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
npawar: Fixed all suggestions. Please take a look @mcvsubbu

mcvsubbu:
lgtm, @Jackie-Jiang had some comments on class hierarchy.  I am neutral on that.

kishoreg:
@npawar  I have a few questions. 

What is the impact on the latency and when does it kick in and for what kind of use cases? 
How will the user know whats the right wait-time? We might still see the same issue if the server has not caught up within the wait time right?

One of the concerns here is that we stop the entire server from serving any other segment.

Looks like we need two things
- limit the number of threads for consumption so that it does not hog all the cpu during bootstrap/server startup
-  Have the logic in the state transition for OFFLINE -> CONSUMING to not ack Helix until it's caught up. This has the added benefit that query results are up to date across all replicas.

I am ok with the temporary fix but was wondering what would be the right fix that requires minimal configuration from the user.

npawar:
> @npawar I have a few questions.
> 
> What is the impact on the latency and when does it kick in and for what kind of use cases?
> How will the user know whats the right wait-time? We might still see the same issue if the server has not caught up within the wait time right?
> 
> One of the concerns here is that we stop the entire server from serving any other segment.
> 
> Looks like we need two things
> 
> * limit the number of threads for consumption so that it does not hog all the cpu during bootstrap/server startup
> * Have the logic in the state transition for OFFLINE -> CONSUMING to not ack Helix until it's caught up. This has the added benefit that query results are up to date across all replicas.
> 
> I am ok with the temporary fix but was wondering what would be the right fix that requires minimal configuration from the user.

The impact we've observed is usually for high ingestion rate realtime tables. The tables can usually take 2-3 minutes to completely catchup after server restart. During this time, we not only see high latencies/slow queries/timeouts because of cpu bottleneck, but also the results returned by that server will not be in sync with what other replicas return. 

We decided to have a simple approach of static wait time to begin with, until we can get more information about the consumption rate. Yes we will still see the issue if the server has not caught up in the configured time. 
We plan to include consumption rate related stats in server stats. Once we have those, we will remove the static wait. We will then wait until consumption rate stabilizes to the average rate or until idle.

I can make the static wait to default 0 if having the default 5 is a problem?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -174,6 +174,8 @@ public ServerType getServerType() {
     public static final String CONFIG_OF_STARTER_ENABLE_SEGMENTS_LOADING_CHECK =
         "pinot.server.starter.enableSegmentsLoadingCheck";
     public static final String CONFIG_OF_STARTER_TIMEOUT_IN_SECONDS = "pinot.server.starter.timeoutInSeconds";
+    public static final String CONFIG_OF_STARTER_REALTIME_CONSUMPTION_CATCHUP_WAIT_TIME_SECONDS =
```
Jackie-Jiang: I would put this along with the **CONFIG_OF_SERVER_MIN_RESOURCE_PERCENT_FOR_START** since both of them are for the status check instead of the starter. Maybe rename it to "pinot.server.startup.realtimeConsumptionCatchupTimeMs"?

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
Jackie-Jiang: Yes, directly implement ServiceStatusCallback

kishoreg:
Agree with providing a simple wait for now but let's keep the default to 0 and allow users to override if needed.

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -128,11 +129,91 @@ public String getStatusDescription() {
...
+   * 3) Monitor consumption rate during startup, report GOOD if it is idle
+   */
+  public static class RealtimeConsumptionCatchupServiceStatusCallback implements ServiceStatusCallback {
+
+    String _statusDescription = STATUS_DESCRIPTION_INIT;
+    private Status _serviceStatus = Status.STARTING;
+    private final long _endWaitTime;
+
+    /**
+     * Realtime consumption catchup service which adds a static wait time for
```
sunithabeeram: Comment appears to be incomplete

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -128,11 +129,91 @@ public String getStatusDescription() {
...
+        // A consuming segment will actually be ready to serve queries after (time of creation of partition consumer) + (configured max time to catchup)
+        // We are approximating it to (time of server startup) + (configured max time to catch up)
+        _endWaitTime = startTime + TimeUnit.SECONDS.toMillis(realtimeConsumptionCatchupWaitTimeSeconds);
+        LOGGER.info("Monitoring realtime consumption catchup. Will allow {} seconds before marking status GOOD",
+            realtimeConsumptionCatchupWaitTimeSeconds);
+      }
+    }
+
+    @Override
+    public Status getServiceStatus() {
```
sunithabeeram: Shouldn't this be synchronized?

npawar:
Fixed suggestions @sunithabeeram 

@mcvsubbu I have refactored it such that we are not making the call to getResourcesInCluster and getResourceIdealState in every constructor. It is now happening only in 1 constructor.

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -129,56 +132,157 @@ public String getStatusDescription() {
...
+    private final List<ServiceStatusCallback> _statusCallbacks;
 
-    private final Set<String> _resourcesToMonitor;
-    private final int _numTotalResourcesToMonitor;
-    private Iterator<String> _resourceIterator = null;
-    // Minimum number of resources to be in converged state before we declare the service state as STARTED
-    private final int _minResourcesStartCount;
-
-    private String _statusDescription = STATUS_DESCRIPTION_INIT;
+    public IdealStateServiceStatusCallback(HelixManager helixManager, String clusterName, String instanceName,
```
mcvsubbu: So, this is a multi-service callback at the same time it is not? I am getting confused deciding one way or another.

Another approach could be that we provide a method in serviceStatus that gets the resources to be monitored, given the parameters. We have another constructor for the individual servicesStatus objects that also takes int he resources to be monitored, and HelixBrokerStarter (or server starter) invokes that constructor. I think that has two advantages:
1. It is better readable (or, at least I think so :)
2. We don't get all idealstates in the broker case. We only get the broker ideal state resource!

npawar: i changed it, to make the HelixServerStarter and HelixBrokerStarter decide the resources to monitor. I don't like the idea of adding a method in ServiceStatus to fetch resources to monitor. Firstly, there's no neat way to distinguish between whether the request is being made by server/broker/test (other than check prefix of instance), and we'll end up with a lot of if else. Secondly, we need to return resourcesToMonitor, as well as boolean indicating need for Realtiem handler (in future, another list to hold consuming segments). 
lmk if current iteration works. @mcvsubbu

pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -237,16 +238,30 @@ public void start()
...
+    LOGGER.info("Finish starting Pinot broker");
+  }
+
+  /**
+   * Fetches the resources to monitor and registers the {@link org.apache.pinot.common.utils.ServiceStatus.ServiceStatusCallback}s
+   */
+  private void registerServiceStatusHandler() {
+    List<String> resourcesToMonitor = new ArrayList<>(1);
+    IdealState resourceIdealState = _helixAdmin.getResourceIdealState(_clusterName, Helix.BROKER_RESOURCE_INSTANCE);
+    if (resourceIdealState != null && resourceIdealState.isEnabled()) {
```
mcvsubbu: dont we also need to check if this broker instance is on for any of the partitions?

npawar: Hmm yes, fixed. 
Follow up question: We still want to set the ServiceStatusHandlers, even if resourcesToMonitor is empty right? That way the health checks will get response GOOD instead of having no service handlers?

mcvsubbu: If each service handler we want to have end up being empty, then yes. we do need to set one that always returns GOOD. Perhaps that can be the default setting for ServiceStatus.getServiceStatus ?

PR #4222, opened 2019-05-19
Add startup/shutdown checks for HelixServerStarter

Jackie-Jiang:
Optional startup checks:
  - Service status check (ON by default)
Optional shutdown checks:
  - Query check (drains and finishes existing queries, ON by default)
  - Resource check (wait for all resources OFFLINE, OFF by default)

Change notes:
  - Service status check is a replacement of the segmentsLoadingCheck (OFF by default). When service status does not return GOOD, log the status description instead of looping over the segments which is potentially expensive with bunch of ZK reads.
  - Query check is a replacement of the shutdownDelay (ON by default) and waitUntilNoIncomingQueries (not configurable)
  - Resource check is a replacement of the noOnlineResources (not configurable)
  - The reason why resource check is OFF by default is that: 1. query check will ensure all queries are drained; 2. when the service status check is enabled, it will compare the currentState/externalView against the resource idealState, which is enough to ensure the segments are loaded before getting queries.
  - If service status check is enabled, server will wait until the check passed then mark itself as up so broker can send queries to it. This is the desired behavior because serving queries and loading segments at the same time might cause performance issue. But this will also delay the time when server start serving queries.

BACKWARD-INCOMPATIBLE:
The following config keys are removed or replaced, but the default
behavior should be desired:
  - pinot.server.starter.enableSegmentsLoadingCheck
  - pinot.server.starter.timeoutInSeconds
  - pinot.server.instance.enable.shutdown.delay
  - pinot.server.instance.starter.maxShutdownWaitTime
  - pinot.server.instance.starter.checkIntervalTime

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
@@ -219,10 +219,10 @@ public class CommonConstants {
     public static final String CONFIG_OF_SHUTDOWN_TIMEOUT_MS = "pinot.server.shutdown.timeoutMs";
     public static final long DEFAULT_SHUTDOWN_TIMEOUT_MS = 600_000L;
     // Shutdown: enable query check before shutting down the server
-    //           Will wait until no new query received and existing queries finished
+    //           Will drain queries (no incoming queries and all existing queries finished)
     public static final String CONFIG_OF_SHUTDOWN_ENABLE_QUERY_CHECK = "pinot.server.shutdown.enableQueryCheck";
     public static final boolean DEFAULT_SHUTDOWN_ENABLE_QUERY_CHECK = true;
-    // Shutdown: threshold to mark that there is no new query received, use max query time as the default threshold
+    // Shutdown: threshold to mark that there is no incoming queries, use max query time as the default threshold
     public static final String CONFIG_OF_SHUTDOWN_NO_QUERY_THRESHOLD_MS = "pinot.server.shutdown.noQueryThresholdMs";
     // Shutdown: enable resource check before shutting down the server
     //           Will wait until all the resources in the external view are neither ONLINE nor CONSUMING
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
@@ -171,13 +171,8 @@ public class CommonConstants {
     public static final String CONFIG_OF_REQUEST_HANDLER_FACTORY_CLASS = "pinot.server.requestHandlerFactory.class";
     public static final String CONFIG_OF_NETTY_PORT = "pinot.server.netty.port";
     public static final String CONFIG_OF_ADMIN_API_PORT = "pinot.server.adminapi.port";
-    public static final String CONFIG_OF_STARTER_ENABLE_SEGMENTS_LOADING_CHECK =
-        "pinot.server.starter.enableSegmentsLoadingCheck";
-    public static final String CONFIG_OF_STARTER_TIMEOUT_IN_SECONDS = "pinot.server.starter.timeoutInSeconds";
 
     public static final String CONFIG_OF_SEGMENT_FORMAT_VERSION = "pinot.server.instance.segment.format.version";
-    public static final String CONFIG_OF_ENABLE_DEFAULT_COLUMNS = "pinot.server.instance.enable.default.columns";
-    public static final String CONFIG_OF_ENABLE_SHUTDOWN_DELAY = "pinot.server.instance.enable.shutdown.delay";
     public static final String CONFIG_OF_ENABLE_SPLIT_COMMIT = "pinot.server.instance.enable.split.commit";
     public static final String CONFIG_OF_ENABLE_COMMIT_END_WITH_METADATA =
         "pinot.server.instance.enable.commitend.metadata";
@@ -186,10 +181,6 @@ public class CommonConstants {
         "pinot.server.instance.realtime.alloc.offheap.direct";
     public static final String PREFIX_OF_CONFIG_OF_PINOT_FS_FACTORY = "pinot.server.storage.factory";
     public static final String PREFIX_OF_CONFIG_OF_PINOT_CRYPTER = "pinot.server.crypter";
-    public static final String CONFIG_OF_INSTANCE_MAX_SHUTDOWN_WAIT_TIME =
-        "pinot.server.instance.starter.maxShutdownWaitTime";
-    public static final String CONFIG_OF_INSTANCE_CHECK_INTERVAL_TIME =
-        "pinot.server.instance.starter.checkIntervalTime";
     // Configuration to consider the server ServiceStatus as being STARTED if the percent of resources (tables) that
     // are ONLINE for this this server has crossed the threshold percentage of the total number of tables
     // that it is expected to serve.
@@ -198,8 +189,6 @@ public class CommonConstants {
     public static final double DEFAULT_SERVER_MIN_RESOURCE_PERCENT_FOR_START = 100.0;
 
     public static final int DEFAULT_ADMIN_API_PORT = 8097;
-    public static final boolean DEFAULT_STARTER_ENABLE_SEGMENTS_LOADING_CHECK = false;
-    public static final int DEFAULT_STARTER_TIMEOUT_IN_SECONDS = 600;
     public static final String DEFAULT_READ_MODE = "heap";
     public static final String DEFAULT_INSTANCE_BASE_DIR =
         System.getProperty("java.io.tmpdir") + File.separator + "PinotServer";
@@ -214,10 +203,37 @@ public class CommonConstants {
     public static final String DEFAULT_REQUEST_HANDLER_FACTORY_CLASS =
         "org.apache.pinot.server.request.SimpleRequestHandlerFactory";
     public static final String PREFIX_OF_CONFIG_OF_SEGMENT_FETCHER_FACTORY = "pinot.server.segment.fetcher";
-    public static final String DEFAULT_STAR_TREE_FORMAT_VERSION = "OFF_HEAP";
+
+    // Configs for server starter startup/shutdown checks
+    // Startup: timeout for the startup checks
+    public static final String CONFIG_OF_STARTUP_TIMEOUT_MS = "pinot.server.startup.timeoutMs";
+    public static final long DEFAULT_STARTUP_TIMEOUT_MS = 600_000L;
+    // Startup: enable service status check before claiming server up
+    public static final String CONFIG_OF_STARTUP_ENABLE_SERVICE_STATUS_CHECK =
+        "pinot.server.startup.enableServiceStatusCheck";
+    public static final boolean DEFAULT_STARTUP_ENABLE_SERVICE_STATUS_CHECK = true;
+    public static final String CONFIG_OF_STARTUP_SERVICE_STATUS_CHECK_INTERVAL_MS =
+        "pinot.server.startup.serviceStatusCheckIntervalMs";
+    public static final long DEFAULT_STARTUP_SERVICE_STATUS_CHECK_INTERVAL_MS = 10_000L;
+    // Shutdown: timeout for the shutdown checks
+    public static final String CONFIG_OF_SHUTDOWN_TIMEOUT_MS = "pinot.server.shutdown.timeoutMs";
+    public static final long DEFAULT_SHUTDOWN_TIMEOUT_MS = 600_000L;
+    // Shutdown: enable query check before shutting down the server
+    //           Will wait until no new query received and existing queries finished
+    public static final String CONFIG_OF_SHUTDOWN_ENABLE_QUERY_CHECK = "pinot.server.shutdown.enableQueryCheck";
+    public static final boolean DEFAULT_SHUTDOWN_ENABLE_QUERY_CHECK = true;
+    // Shutdown: threshold to mark that there is no new query received, use max query time as the default threshold
+    public static final String CONFIG_OF_SHUTDOWN_NO_QUERY_THRESHOLD_MS = "pinot.server.shutdown.noQueryThresholdMs";
+    // Shutdown: enable resource check before shutting down the server
+    //           Will wait until all the resources in the external view are neither ONLINE nor CONSUMING
+    //           No need to enable this check if startup service status check is enabled
+    public static final String CONFIG_OF_SHUTDOWN_ENABLE_RESOURCE_CHECK = "pinot.server.shutdown.enableResourceCheck";
+    public static final boolean DEFAULT_SHUTDOWN_ENABLE_RESOURCE_CHECK = false;
+    public static final String CONFIG_OF_SHUTDOWN_RESOURCE_CHECK_INTERVAL_MS =
+        "pinot.server.shutdown.resourceCheckIntervalMs";
+    public static final long DEFAULT_SHUTDOWN_RESOURCE_CHECK_INTERVAL_MS = 10_000L;
+
     public static final String DEFAULT_COLUMN_MIN_MAX_VALUE_GENERATOR_MODE = "TIME";
-    public static final long DEFAULT_MAX_SHUTDOWN_WAIT_TIME_MS = 600_000L;
-    public static final long DEFAULT_CHECK_INTERVAL_TIME_MS = 60_000L;
 
     public static class SegmentCompletionProtocol {
       public static final String PREFIX_OF_CONFIG_OF_SEGMENT_UPLOADER = "pinot.server.segment.uploader";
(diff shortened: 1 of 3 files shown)


pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -171,13 +171,8 @@ public ServerType getServerType() {
     public static final String CONFIG_OF_REQUEST_HANDLER_FACTORY_CLASS = "pinot.server.requestHandlerFactory.class";
     public static final String CONFIG_OF_NETTY_PORT = "pinot.server.netty.port";
     public static final String CONFIG_OF_ADMIN_API_PORT = "pinot.server.adminapi.port";
-    public static final String CONFIG_OF_STARTER_ENABLE_SEGMENTS_LOADING_CHECK =
```
jackjlli: Can we log these configs as `deprecated` in the server log when startup to notify these configs have been deprecated? Otherwise, it'll be hard for open source to notice this changes.

Jackie-Jiang: IMO we should not keep unused constants in the code base, it's better to put them in the release log.

Jackie-Jiang: Discussed offline, log warnings for usage of deprecated config keys

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -72,61 +63,66 @@
...
-      _instanceId = CommonConstants.Helix.PREFIX_OF_SERVER_INSTANCE + host + "_" + port;
-      _helixServerConfig.addProperty(CommonConstants.Server.CONFIG_OF_INSTANCE_ID, _instanceId);
+    _serverConf = ConfigurationUtils.cloneConfiguration(serverConf);
+
+    // Log warnings for usage of deprecated config keys
+    String[] deprecatedConfigKeys =
+        new String[]{CONFIG_OF_STARTER_ENABLE_SEGMENTS_LOADING_CHECK, CONFIG_OF_STARTER_TIMEOUT_IN_SECONDS, CONFIG_OF_ENABLE_SHUTDOWN_DELAY, CONFIG_OF_INSTANCE_MAX_SHUTDOWN_WAIT_TIME, CONFIG_OF_INSTANCE_CHECK_INTERVAL_TIME};
+    for (String deprecatedConfigKey : deprecatedConfigKeys) {
+      if (_serverConf.containsKey(deprecatedConfigKey)) {
+        LOGGER.warn("Found usage of deprecated key: {}", deprecatedConfigKey);
```
sunithabeeram: Might be good to indicate what this is replaced with.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -337,12 +237,47 @@ private void setupHelixSystemProperties() {
...
+   * @param endTimeMs Timeout for the check
+   */
+  private void startupServiceStatusCheck(long endTimeMs) {
+    LOGGER.info("Starting startup service status check");
+    long startTimeMs = System.currentTimeMillis();
+    long checkIntervalMs = _serverConf
+        .getLong(CONFIG_OF_STARTUP_SERVICE_STATUS_CHECK_INTERVAL_MS, DEFAULT_STARTUP_SERVICE_STATUS_CHECK_INTERVAL_MS);
+
+    while (System.currentTimeMillis() < endTimeMs) {
+      Status serviceStatus = ServiceStatus.getServiceStatus();
```
sunithabeeram: This can be considered as an enhancement and addressed later; ServiceStatus exposes polling based semantics; with this change it can be called by multiple 'actors'; one is the server that is starting up by itself, and external deploy tools calling health check. It would be cleaner if the status check can be turned into a something that runs in a thread and updates the state. All other callers conditionally-wait on that particular state instead of triggering the check yet again.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -72,61 +63,66 @@
...
+import static org.apache.pinot.common.utils.CommonConstants.Server.*;
+
 
 /**
- * Single server helix starter. Will start automatically with an untagged box.
- * Will auto join current cluster as a participant.
- *
- *
- *
+ * Starter for Helix-based Pinot server.
```
mcvsubbu: Suggest remove the words 'helix-based'. I suppose this is a left-over of file-based server that we had long ago, but we don't have anymore.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -352,135 +287,160 @@ public void stop() {
...
+      if (currentTimeMs >= latestQueryTimeMs + noQueryThresholdMs) {
+        LOGGER.info("Finished draining queries (no query received within {}ms) after {}ms",
+            currentTimeMs - latestQueryTimeMs, currentTimeMs - startTimeMs);
+        queriesDrained = true;
+        break;
       }
-
       try {
-        Thread.sleep(Math.min(_maxQueryTimeMs, (endTime - currentTime)));
+        Thread.sleep(Math.min(noQueryThresholdMs - latestQueryTimeMs, endTimeMs - currentTimeMs));
```
mcvsubbu: noQueryThresholdMs is an interval time, and latestQueryTime is an absolute value. Am I reading it right that we are subtracting one from the other and using that as an interval?

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -337,12 +237,47 @@ private void setupHelixSystemProperties() {
...
+      Status serviceStatus = ServiceStatus.getServiceStatus();
+      long currentTimeMs = System.currentTimeMillis();
+      if (serviceStatus == Status.GOOD) {
+        LOGGER.info("Service status is GOOD after {}ms", currentTimeMs - startTimeMs);
+        return;
+      } else if (serviceStatus == Status.BAD) {
+        throw new IllegalStateException("Service status is BAD");
+      }
+      try {
+        Thread.sleep(Math.min(checkIntervalMs, endTimeMs - currentTimeMs));
```
mcvsubbu: Useful to log service status string before the wait, so we know what we are waiting for

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
mcvsubbu: Useful to log that we are waiting for queries to drain. Include the last query time in the log.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -352,135 +287,160 @@ public void stop() {
...
+        queriesDrained = true;
+        break;
       }
-
       try {
-        Thread.sleep(Math.min(_maxQueryTimeMs, (endTime - currentTime)));
+        Thread.sleep(Math.min(noQueryThresholdMs - latestQueryTimeMs, endTimeMs - currentTimeMs));
       } catch (InterruptedException e) {
-        LOGGER.error("Interrupted when waiting for Pinot server not to receive any queries.", e);
+        LOGGER.error("Got interrupted while draining queries", e);
```
mcvsubbu: This may not be an error. It could be that the script that shuts down the server sent an interrupt because it got impatient and wants a quick shutdown (legitimate). If we are interrupted, we should be returning from here instead of waiting the full length

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -352,135 +287,160 @@ public void stop() {
...
+      }
+    }
+    if (queriesDrained) {
+      // Ensure all the existing queries are finished
+      long latestQueryFinishTimeMs = _serverInstance.getLatestQueryTime() + maxQueryTimeMs;
+      if (latestQueryFinishTimeMs > currentTimeMs) {
+        try {
+          Thread.sleep(latestQueryFinishTimeMs - currentTimeMs);
+        } catch (InterruptedException e) {
+          LOGGER.warn("Got interrupted while waiting for all the existing queries to be finished", e);
```
mcvsubbu: return if interrupted.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -352,135 +287,160 @@ public void stop() {
...
+          LOGGER.info("All resources are OFFLINE after {}ms", System.currentTimeMillis() - startTimeMs);
           return;
         }
-
         try {
-          Thread.sleep(Math.min(_checkIntervalTimeMs, (endTime - currentTime)));
+          Thread.sleep(Math.min(checkIntervalMs, endTimeMs - System.currentTimeMillis()));
         } catch (InterruptedException e) {
-          LOGGER.error("Interrupted when waiting for no online resources.", e);
+          LOGGER.warn("Got interrupted while waiting for all resources OFFLINE", e);
```
mcvsubbu: return if we got interrupted

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: Done.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: There is no one-to-one map for the keys. Changed the warning message to "Found usage of deprecated config key: {}, check CommonConstants for the new config keys"

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: Agree.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: Added.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: Good catch. Should be (latestQueryTimeMs + noQueryThresholdMs - currentTimeMs). Manually checked in the integration test

Jackie-Jiang: Added

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: Changed to warn.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: This is the last statement already. Added the `Thread.currentThread().interrupt();`

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: When we got interrupted, still log the unfinished resources and then return.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -72,61 +63,67 @@
...
-      _instanceId = CommonConstants.Helix.PREFIX_OF_SERVER_INSTANCE + host + "_" + port;
-      _helixServerConfig.addProperty(CommonConstants.Server.CONFIG_OF_INSTANCE_ID, _instanceId);
+    _serverConf = ConfigurationUtils.cloneConfiguration(serverConf);
+
+    // Log warnings for usage of deprecated config keys
+    String[] deprecatedConfigKeys =
+        new String[]{CONFIG_OF_STARTER_ENABLE_SEGMENTS_LOADING_CHECK, CONFIG_OF_STARTER_TIMEOUT_IN_SECONDS, CONFIG_OF_ENABLE_SHUTDOWN_DELAY, CONFIG_OF_INSTANCE_MAX_SHUTDOWN_WAIT_TIME, CONFIG_OF_INSTANCE_CHECK_INTERVAL_TIME};
+    for (String deprecatedConfigKey : deprecatedConfigKeys) {
+      if (_serverConf.containsKey(deprecatedConfigKey)) {
+        LOGGER.warn("Found usage of deprecated config key: {}, check CommonConstants for the new config keys",
```
mcvsubbu: I would keep a map of the old and new key (or, a map of old key and error message to be displayed). Try not to ask user to refer to source code.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
```diff
@@ -72,61 +63,66 @@
...
+ *   <li>
+ *     Optional shut-down checks:
+ *     <ul>
+ *       <li>Query check (drains and finishes existing queries, ON by default)</li>
+ *       <li>Resource check (wait for all resources OFFLINE, OFF by default)</li>
+ *     </ul>
+ *   </li>
+ * </ul>
  */
 public class HelixServerStarter {
```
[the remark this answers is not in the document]
Jackie-Jiang: We can rename the class, but all the client-side code need to be changed as well. Can be done in a separate PR.

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java
Jackie-Jiang: Added the map, and also the descriptions

Jackie-Jiang:
@sunithabeeram Updated the PR description for the behavior changes.

sunithabeeram:
Just remembered one thing that might be good to keep in mind: the reason a shutdown resource check was added was to handle the case of ZK/Helix slowness where the broker's EV was lagging. At the time, @kishoreg had suggested that waiting at shutdown time is better as its a stronger guarantee than at startup (if Broker's view is lagging, it may continue to - esp if the instance-config update processing is lagging).
This change flips the argument somewhat. We'll still need a startup wait because of the latency issues noted. We might have to keep a closer on the following deployments as we now have the resource offline check turned off.

Jackie-Jiang:
> Just remembered one thing that might be good to keep in mind: the reason a shutdown resource check was added was to handle the case of ZK/Helix slowness where the broker's EV was lagging. At the time, @kishoreg had suggested that waiting at shutdown time is better as its a stronger guarantee than at startup (if Broker's view is lagging, it may continue to - esp if the instance-config update processing is lagging).
> This change flips the argument somewhat. We'll still need a startup wait because of the latency issues noted. We might have to keep a closer on the following deployments as we now have the resource offline check turned off.

Both shutdown/startup check needs to wait for controller to pickup the current state changes and put them into external view, thus shutdown check won't give stronger guarantee. I'll monitor the following deployments and see if we need to tune the default behavior (that's why I keep everything configurable)

mcvsubbu:
Does it make sense to return a 'shutting down' code from the server when it gets a broker request? The broker can then stop sending more requests to the server.

Jackie-Jiang:
@mcvsubbu That's possible, but server also need to send a 'startting up' to broker in order to get new requests. Right now we rely on the 'shutDownInProgress' flag in the instance config.

mcvsubbu:
@Jackie-Jiang  we don't need to implement the 'starting up' part at all. When the broker sends a query to the server, it can do one of:
* send query response (either partial or complete)
* send a boolean flag (or, an exception, not sure) saying it is shutting down. The broker can record this as if the server went off from externalview
* timeout

On startup, the broker can use the same logic as checking the EV  change. Of course, we will need to account for the conditions where EV change can happen due to some other server (maybe broker should check instance config for 'isShuttingDown' before including a new server in the routing table).

Jackie-Jiang:
@mcvsubbu In ideal case, broker should get the "isShuttingDown" change immediately. If that is not the case, we can implement your proposal, but there will be plenty of corner cases we need to handle.

PR #4231, opened 2019-05-22
Bounds

harleyjj:
1) Adds confidence bounds to graph in Preview when Predicted baseline is selected
2) Adds spinner to graph when updating baseline only (Preview and Alert Overview)
3) Makes time range selection automatically call preview - necessary since the current time series now comes from the preview endpoint
4) Fixes legend to not show when there are more than 20 anomalies to display in the graph
5) Enables filtering by both rule and dimension in preview

### DIFF
--- a/thirdeye/thirdeye-pinot/pom.xml
+++ b/thirdeye/thirdeye-pinot/pom.xml
@@ -18,11 +18,6 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-configuration2</artifactId>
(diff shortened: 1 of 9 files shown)


thirdeye/thirdeye-pinot/pom.xml
```diff
@@ -17,6 +17,11 @@
   </properties>
 
   <dependencies>
+    <dependency>
```
jihaozh: do we need to add this dependency?

harleyjj: No, it's not needed

PR #4233, opened 2019-05-22
Remove depependency on Kafka stream implementation classes from pinot-core and pinot-controller tests

npawar:
As discussed in https://github.com/apache/incubator-pinot/issues/3998, we need to remove all dependencies of kafka stream implementation related classes from pinot-core, pinot-common, pinot-controller.


pinot-core/src/test/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
```diff
@@ -133,6 +135,30 @@ private LLCRealtimeSegmentZKMetadata createZkMetadata() {
     return segmentZKMetadata;
   }
 
+  public static class FakeStreamConsumerFactory extends StreamConsumerFactory {
```
kishoreg: Does it make sense to move this to another class/package so that we can use it else where?

pinot-core/src/test/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
```diff
@@ -717,7 +743,7 @@ protected void start() {
     protected boolean consumeLoop()
         throws Exception {
       if (_throwExceptionFromConsume) {
-        throw new PermanentConsumerException(Errors.OFFSET_OUT_OF_RANGE.exception());
+        throw new PermanentConsumerException(new Throwable("Offset out of range"));
```
kishoreg: why wrap exception within an exception?

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/sharding/SegmentAssignmentStrategyTest.java
```diff
@@ -255,15 +253,20 @@ public void testRealtimeReplicaGroupPartitionAssignment() {
 
     Map<String, String> streamConfigMap = new HashMap<>();
     String type = "kafka";
+    String consumerFactoryClass = "com.test.TestConsumerFactoryClass";
```
kishoreg: looks like this class does not exist.

pinot-core/src/test/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
npawar: I didn't really find any other places which need it. I was planning to add fake stream if I encountered a lot of places that need the stream (with mock implementations)

pinot-core/src/test/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
npawar: Not sure why the test was written that way in the first place. I'm just trying to remove dependency on the kafka class Errors.
PermanentConsumerException wraps stream specific exceptions

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/sharding/SegmentAssignmentStrategyTest.java
npawar: It doesn't need to exist unless the test is starting PartitionConsumers. It's just serving as a dummy value. We've been using the kafka class names, just because we had access to them

pinot-core/src/test/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
npawar: Added a FakeStream, to use in all the tests

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/sharding/SegmentAssignmentStrategyTest.java
npawar: Fixed

pinot-core/src/test/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
mcvsubbu: To add to Neha's comment, the class being tested behaves differently based on whether PermanentConsumerException is thrown or transient one is thrown. It is up to the stream to decide if an exception is permanent or transient. The higher layer has no idea. So we need to wrap.

pinot-core/src/test/resources/data/fakestream/fake_stream_pinot_schema.json
```diff
@@ -0,0 +1,335 @@
+{
+  "schemaName": "mytable",
```
mcvsubbu: Do we need to checkin the schema? Can we generate it from the avro file given? (since fake stream always reads from avro file)

npawar: I wanted to avoid extracting the data once more. I have no preference either way. It was just easier this way

npawar:
> Can we try not to hard-code the number of partitions to 2?? Each test should be able to start with as many partitions as needed? I do run the tests with different number of partitions on my desktop to test sometimes.

Fixed. The numPartitions can now be provided via the StreamConfig property. The StreamConfig can be created using `StreamConfig streamConfig = FakeStreamConfigUtils::getDefaultLowLevelStreamConfig(int num partitions)`

pinot-core/src/test/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
kishoreg: Got it.

pinot-core/src/test/resources/data/fakestream/fake_stream_pinot_schema.json
mcvsubbu: OK, then please add a comment in the schema file that it must be in sync with the avro file of specific path.

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/fakestream/FakePartitionLevelConsumer.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.pinot.core.util.AvroUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.pinot.core.realtime.impl.fakestream.FakeStreamConfigUtils.*;
+
+
+/**
+ * Implementation of {@link PartitionLevelConsumer} for fake stream
+ * Unpacks tar files in /resources/data/On_Time_Performance_2014_partition_<partition>.tar.gz as source of messages
```
mcvsubbu: Good to add a comment saying that we unpack the tar files into tempdir. Not sure if we will fill up tempdir

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/fakestream/FakePartitionLevelConsumer.java
```diff
@@ -0,0 +1,121 @@
...
+    } catch (Exception e) {
+      LOGGER.error("Could not create {}", FakePartitionLevelConsumer.class.getName(), e);
+    } finally {
+      FileUtils.deleteQuietly(outputDir);
+    }
+  }
+
+  @Override
+  public MessageBatch fetchMessages(long startOffset, long endOffset, int timeoutMillis) throws TimeoutException {
+    if (startOffset >= FakeStreamConfigUtils.getLargestOffset()) {
```
mcvsubbu: Instead of unpacking to a tempdir, is it possible to just include the loop in this place, and get the messages from the original avro file? If this does not seem straight-forward, you can add that as a TODO and work on it later. Just trying to make tests go faster, and try not to fill out tempdir.

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/fakestream/FakeStreamLevelConsumer.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.pinot.core.data.GenericRow;
+import org.apache.pinot.core.realtime.stream.StreamLevelConsumer;
+
+
+/**
+ * Test implementation of {@link StreamLevelConsumer}
+ * This is currently a no-op
+ */
+public class FakeStreamLevelConsumer implements StreamLevelConsumer {
```
snleee: Why `FakeStreamLevelConsumer` is no-op? I thought that this would provide some events that you feed as a list of `GenericRow` or avro files?

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/fakestream/FakePartitionLevelConsumer.java
snleee: It would be great if we can configure the input avro files instead of hard coding it to a specific file in case we want to reuse this fake consumer but use different data.

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/fakestream/FakePartitionLevelConsumer.java
mcvsubbu: Thanks for clarifying, @npawar . We cannot do the proposed method since we want to return records based on an offset.

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/fakestream/FakeStreamLevelConsumer.java
npawar: I've mocked that in the partition level consumer. We haven't been using a stream level consumer with mock data anywhere. Will implement it if we need to

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/fakestream/FakeStreamConsumerFactory.java
```diff
@@ -0,0 +1,101 @@
...
+  public StreamMetadataProvider createPartitionMetadataProvider(String clientId, int partition) {
+    return new FakeStreamMetadataProvider(_streamConfig);
+  }
+
+  @Override
+  public StreamMetadataProvider createStreamMetadataProvider(String clientId) {
+    return new FakeStreamMetadataProvider(_streamConfig);
+  }
+
+  public static void main(String[] args) throws Exception {
```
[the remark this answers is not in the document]
npawar: I've kept it as a demonstration of how to use this fake stream

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/fakestream/FakePartitionLevelConsumer.java
npawar: I think that's a little out of scope.. the kafka stream that I'm trying to replace also read only from files, so we have no tests that need this. We are not even using the partition level consumer of the fake stream anywhere in our code yet, so maybe for now we can leave it to hard coded

mcvsubbu: Agreed with @npawar . Current tests read from hard-coded files that are checked in, and push to kafka topic.

pinot-core/src/test/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManagerTest.java
```diff
@@ -88,7 +85,7 @@
           + "\": \"" + maxTimeForSegmentCloseMs + "\", \n" + "      \"stream.kafka.broker.list\": \"broker:7777\", \n"
           + "      \"stream.kafka.consumer.prop.auto.offset.reset\": \"smallest\", \n"
```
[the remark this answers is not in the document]
npawar: Fixed new line and kafka references

npawar:
> Don't you need to change BaseClusterIntegrationTest to not start kafka etc. Or, is that a different PR?

We will let the integration tests run on the kafka implementation (pinot-integration-tests can still have a dependency on pinot-kafka-0.9.x). If we decide to remove the pinot-kafka module entirely out of the  pinot repo, then we can move it to use fakestream.

PR #4234, opened 2019-05-22
Remove kafka dependency from PartitionFunction

npawar:
As discussed in issue https://github.com/apache/incubator-pinot/issues/3998, we want to remove kafka dependency from pinot-core and pinot-common. The Murmur and ByteArray PartitionFunctions depend on kafka classes. We do not really need to depend on those classes. The hash logic is generic and can be copied over to our internal implementations.
Have added a note to introduce a clean way to inject custom PartitionFunctions.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/partition/ByteArrayPartitionFunction.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/partition/ByteArrayPartitionFunction.java
@@ -19,8 +19,7 @@
 package org.apache.pinot.core.data.partition;
 
 import com.google.common.base.Preconditions;
-import kafka.producer.ByteArrayPartitioner;
-
+import java.util.Arrays;
 
 /**
  * Implementation of {@link Byte array partitioner}
@@ -29,7 +28,6 @@ import kafka.producer.ByteArrayPartitioner;
 public class ByteArrayPartitionFunction implements PartitionFunction {
   private static final String NAME = "ByteArray";
   private final int _numPartitions;
-  public ByteArrayPartitioner _byteArrayPartitioner;
 
   /**
    * Constructor for the class.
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/org/apache/pinot/core/data/partition/MurmurPartitionFunction.java
```diff
@@ -19,19 +19,11 @@
...
 /**
- * Implementation of {@link PartitionFunction} that mimics Kafka's
- * {@link org.apache.kafka.clients.producer.internals.DefaultPartitioner}.
- *
- * **** ALERT ****
- * However, this is not general purpose. It assumes that the partition key
- * was a string as used by the Tracking Producer. This is only a reference
- * implementation and users should create their own for their specific partitioner.
- *
+ * Implementation of {@link PartitionFunction} which partitions based on 32 bit murmur hash
```
snleee: Can you add the comment that this is taken from Kafka's util?

Also, have you check some examples to make sure they are the same?

pinot-core/src/main/java/org/apache/pinot/core/data/partition/ByteArrayPartitionFunction.java
```diff
@@ -55,5 +52,9 @@ public int getNumPartitions() {
   public String toString() {
     return NAME;
   }
+
+  private int abs(int n) {
+    return (n == Integer.MIN_VALUE) ? 0 : Math.abs(n);
```
snleee: Is this behavior the same as the original code?

snleee:
I think that it would be good if we can add some unit tests for `Murmur` partitioner with pre-computed result with kafka util.

mcvsubbu:
I would run the existing Kafka partition function on a few random strings (perhaps generated with some seed), record the answers, then run them with the newly coded functions, just to be sure.
This (or a subset) can then become the unit test sample.

pinot-core/src/main/java/org/apache/pinot/core/data/partition/MurmurPartitionFunction.java
npawar: Added comment.
Added tests to check equivalence

npawar:
> I think that it would be good if we can add some unit tests for `Murmur` partitioner with pre-computed result with kafka util.

I've added a tests to check equivalence of results of previous implementation with the new implementation. Please take a look @snleee

pinot-core/src/main/java/org/apache/pinot/core/data/partition/ByteArrayPartitionFunction.java
npawar: Verified it is with tests added

PR #4235, opened 2019-05-23
Add config to disable HLC realtime segment completion

jackjlli:
This PR adds a config to explicitly disable HLC realtime segment completion.
By default it is enabled.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
@@ -32,7 +32,6 @@ import org.apache.commons.configuration.PropertiesConfiguration;
 import org.apache.pinot.common.protocols.SegmentCompletionProtocol;
 import org.apache.pinot.common.utils.CommonConstants;
 import org.apache.pinot.common.utils.StringUtil;
-import org.apache.pinot.controller.helix.core.util.HelixSetupUtils;
 import org.apache.pinot.filesystem.LocalPinotFS;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -136,8 +135,8 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final String REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS =
       "controller.realtime.segment.metadata.commit.numLocks";
   private static final String ENABLE_STORAGE_QUOTA_CHECK = "controller.enable.storage.quota.check";
-
   private static final String ENABLE_BATCH_MESSAGE_MODE = "controller.enable.batch.message.mode";
+  private static final String ALLOW_HLC_TABLES = "controller.allow.hlc.tables";
 
   // Defines the kind of storage and the underlying PinotFS implementation
   private static final String PINOT_FS_FACTORY_CLASS_PREFIX = "controller.storage.factory.class";
@@ -155,6 +154,7 @@ public class ControllerConf extends PropertiesConfiguration {
   private static final int DEFAULT_REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS = 64;
   private static final boolean DEFAULT_ENABLE_STORAGE_QUOTA_CHECK = true;
   private static final boolean DEFAULT_ENABLE_BATCH_MESSAGE_MODE = false;
+  private static final boolean DEFAULT_ALLOW_HLC_TABLES = true;
   private static final String DEFAULT_CONTROLLER_MODE = ControllerMode.DUAL.name();
 
   private static final String DEFAULT_PINOT_FS_FACTORY_CLASS_LOCAL = LocalPinotFS.class.getName();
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -136,7 +136,7 @@ private static long getRandomInitialDelayInSeconds() {
   private static final String REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS =
       "controller.realtime.segment.metadata.commit.numLocks";
   private static final String ENABLE_STORAGE_QUOTA_CHECK = "controller.enable.storage.quota.check";
-
+  private static final String ENABLE_HLC_REALTIME_SEGMENT_COMPLETION = "controller.enable.hlc.realtime.segment.completion";
```
mcvsubbu: Can we rename the config as ALLOW_HLC_TABLES? Please document the config clearly. If the config says "true"(default),  then the behavior is unchanged. If the config  is set to "false", then the following will happen:
* Existing  HLC realtime tables will stop consumption
* Creation of mew HLC tables will be disallowed.

mcvsubbu:
Also, please change the LLCRealtimeCluster test to work with this disabled setting

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -136,8 +135,8 @@ private static long getRandomInitialDelayInSeconds() {
   private static final String REALTIME_SEGMENT_METADATA_COMMIT_NUMLOCKS =
       "controller.realtime.segment.metadata.commit.numLocks";
   private static final String ENABLE_STORAGE_QUOTA_CHECK = "controller.enable.storage.quota.check";
-
   private static final String ENABLE_BATCH_MESSAGE_MODE = "controller.enable.batch.message.mode";
+  private static final String ALLOW_HLC_TABLES = "controller.allow.hlc.tables";
```
mcvsubbu: please add a paragraph to describe what happns when this is turned off.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -273,8 +273,14 @@ private void setUpPinotController() {
         new SegmentCompletionManager(helixParticipantManager, _pinotLLCRealtimeSegmentManager, _controllerMetrics,
             _controllerLeadershipManager, _config.getSegmentCommitTimeoutSeconds());
 
-    _realtimeSegmentsManager = new PinotRealtimeSegmentManager(_helixResourceManager, _controllerLeadershipManager);
-    _realtimeSegmentsManager.start(_controllerMetrics);
+    if (_config.getHLCTablesAllowed()) {
+      LOGGER.info("HLC table is allowed. HLC realtime segment completion is enabled.");
```
mcvsubbu: I suggest the log message to be "Realtime tables with High Level consumers will be supported"

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -273,8 +273,14 @@ private void setUpPinotController() {
...
+      _realtimeSegmentsManager = new PinotRealtimeSegmentManager(_helixResourceManager, _controllerLeadershipManager);
+      _realtimeSegmentsManager.start(_controllerMetrics);
+    } else {
+      LOGGER.info("HLC table is disallowed. HLC realtime segment completion is disabled.");
```
mcvsubbu: Suggest change the log message to: "Realtime tables with High Level consumers will NOT be supported"

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1408,6 +1421,11 @@ public void updateIndexingConfigFor(String tableName, TableType type, IndexingCo
     setExistingTableConfig(tableConfig, tableNameWithType, type);
 
     if (type == TableType.REALTIME) {
+      // Check if HLC table is allowed
```
mcvsubbu: Did you look into the possibility of having a single validateIndexingConfig method?

jackjlli: Added.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
jackjlli: Updated.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
jackjlli: Updated.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
jackjlli: Added.

PR #4236, opened 2019-05-23
Add metric on ControllerLeadershipManager

jackjlli:
In large cluster, when a helix controller acquires the leadership, it needs to do a bunch of work to prepare for the leadership, like subscribing all the current state, etc. And Pinot leadership's related methods won't be called until Helix cluster finishes its work. Thus, it will take time for Pinot to start the leader's work while Helix immediately mark this controller as leader (Helix leader).

This PR marks the metric of checking Helix leader to be deprecated, and adds a new one of checking Pinot's leadership, so that we know it will always be the true Pinot leader when the metric is 1, and helix leadership preparation is completed.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
@@ -22,6 +22,8 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import org.apache.helix.HelixManager;
 import org.apache.helix.api.listeners.ControllerChangeListener;
+import org.apache.pinot.common.metrics.ControllerGauge;
+import org.apache.pinot.common.metrics.ControllerMetrics;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
```diff
@@ -35,12 +37,14 @@
...
   private HelixManager _helixManager;
+  private ControllerMetrics _controllerMetrics;
   private volatile boolean _amILeader = false;
 
   private Map<String, LeadershipChangeSubscriber> _subscribers = new ConcurrentHashMap<>();
 
-  public ControllerLeadershipManager(HelixManager helixManager) {
+  public ControllerLeadershipManager(HelixManager helixManager, ControllerMetrics controllerMetrics) {
     _helixManager = helixManager;
+    _controllerMetrics = controllerMetrics;
```
mcvsubbu: we should set it to 0 when the constructor is called

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -224,7 +224,6 @@ private void setUpHelixController() {
     // Emit helix controller metrics
     _controllerMetrics.addCallbackGauge(CommonConstants.Helix.INSTANCE_CONNECTED_METRIC_NAME,
         () -> _helixControllerManager.isConnected() ? 1L : 0L);
-    _controllerMetrics.addCallbackGauge("helix.leader", () -> _helixControllerManager.isLeader() ? 1L : 0L);
```
mcvsubbu: DO not remove the metric, Mark it deprecated.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java

jackjlli:
@mcvsubbu Log messages added.

mcvsubbu:
> @mcvsubbu Log messages added.

Thank you!

PR #4237, opened 2019-05-23
[TE] frontend - harleyjj/edit-alert - use new yaml components in edit…

harleyjj:
…-alert

This commit helps prepare the manage/yaml route (edit alert) for full-ui.  The PR:

1) Implements edit-alert using new separate yaml components
2) Migrates shared props to the controller
3) Fixes bug where preview button doesn't show unless user types
4) Fixes bug where template yaml text is shoved into editor when copy-pasting content
5) Makes all subscription groups available in dropdown on Edit Alert, since subscription group editing has been separated on Edit Alert anyway

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/manage/yaml/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/manage/yaml/route.js
@@ -9,10 +9,13 @@ import { set, get } from '@ember/object';
 import { inject as service } from '@ember/service';
 import yamljs from 'yamljs';
 import moment from 'moment';
-import { toastOptions } from 'thirdeye-frontend/utils/constants';
+import { yamlAlertSettings, toastOptions } from 'thirdeye-frontend/utils/constants';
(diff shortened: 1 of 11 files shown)


jihaozh:
As discussed offline, in the dropdown on Edit Alert, could you use make it into two separate sections to show subscribed groups and all other groups? Probably that's the most flexible design. This should be doable in ember power select component. https://ember-power-select.com/docs/groups

thirdeye/thirdeye-frontend/app/pods/manage/yaml/route.js
```diff
@@ -86,14 +89,58 @@ export default Route.extend({
...
+      .sortBy('name')
+      .filter(group => (group.get('active') && group.get('yaml')))
+      .map(group => {
+        return {
+          name: group.get('name'),
+          id: group.get('id'),
+          yaml: group.get('yaml')
+        };
+      });
+    const subscriptionGroupNamesDisplay = [
```
harleyjj: @jihaozh Here is the main change to make the dropdown accommodate groups.

PR #4240, opened 2019-05-24
Skip pipeline validation when disabling it

xiaohui-sun:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -343,14 +343,21 @@ void updateDetectionPipeline(long detectionID, String yamlDetectionConfig, long
...
-
-    // Validate updated config before saving it
-    detectionValidator.validateUpdatedConfig(detectionConfig, existingDetectionConfig);
-    // Save the detection config
-    Long id = this.detectionConfigDAO.save(detectionConfig);
-    Preconditions.checkNotNull(id, "Error while saving the detection pipeline");
+    // If it is to disable the pipeline then no need to do validation and parsing.
+    // It is possible that the metric or dataset was deleted so the validation will fail.
+    if (!MapUtils.getBooleanValue(newDetectionConfigMap, PROP_ACTIVE, true)) {
+      existingDetectionConfig.setActive(false);
```
akshayrai: The raw yaml config needs to be injected into the detection object before saving. Otherwise, the config will still show `active: true`.

xiaohui-sun: Good point. I put it into "finally" now. Do you have other ideas?

PR #4243, opened 2019-05-28
Add support for passing headers in pinot client

jitendratheta:
There are usecases where the service using the pinot client need to send the user defined headers to the http endpoint which seats behind load balancer. Adding support for the same in the client

### DIFF
--- a/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
+++ b/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
@@ -18,8 +18,8 @@
  */
 package org.apache.pinot.client;
 
-import org.apache.yetus.audience.InterfaceAudience;
-import org.apache.yetus.audience.InterfaceStability;
+import org.apache.pinot.annotations.InterfaceAudience;
+import org.apache.pinot.annotations.InterfaceStability;
 
 import java.util.Arrays;
 import java.util.List;
--- a/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
+++ b/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
@@ -18,8 +18,8 @@
  */
 package org.apache.pinot.client;
 
-import org.apache.pinot.annotations.InterfaceAudience;
-import org.apache.pinot.annotations.InterfaceStability;
+import org.apache.yetus.audience.InterfaceAudience;
+import org.apache.yetus.audience.InterfaceStability;
 
 import java.util.Arrays;
 import java.util.List;
--- a/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
+++ b/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
@@ -18,6 +18,9 @@
  */
 package org.apache.pinot.client;
 
+import org.apache.yetus.audience.InterfaceAudience;
+import org.apache.yetus.audience.InterfaceStability;
+
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -27,6 +30,8 @@ import java.util.Properties;
 /**
  * Creates connections to Pinot, given various initialization methods.
  */
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
 public class ConnectionFactory {
   static PinotClientTransportFactory _transportFactory = new JsonAsyncHttpPinotClientTransportFactory();
 
--- a/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
+++ b/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
@@ -19,6 +19,8 @@
 package org.apache.pinot.client;
 
 import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 
@@ -40,7 +42,7 @@ public class ConnectionFactory {
   public static Connection fromZookeeper(String zkUrl) {
     try {
       DynamicBrokerSelector dynamicBrokerSelector = new DynamicBrokerSelector(zkUrl);
-      return new Connection(dynamicBrokerSelector, _transportFactory.buildTransport());
+      return new Connection(dynamicBrokerSelector, _transportFactory.buildTransport(null));
     } catch (Exception e) {
       throw new PinotClientException(e);
     }
@@ -54,7 +56,7 @@ public class ConnectionFactory {
    */
   public static Connection fromProperties(Properties properties) {
     return new Connection(Arrays.asList(properties.getProperty("brokerList").split(",")),
-        _transportFactory.buildTransport());
+        _transportFactory.buildTransport(null));
   }
 
   /**
@@ -64,6 +66,17 @@ public class ConnectionFactory {
    * @return A connection to the set of brokers specified
    */
   public static Connection fromHostList(String... brokers) {
-    return new Connection(Arrays.asList(brokers), _transportFactory.buildTransport());
+    return new Connection(Arrays.asList(brokers), _transportFactory.buildTransport(null));
+  }
+
+  /**
+   * Creates a connection which sends queries randomly between the specified brokers.
+   *
+   * @param brokers The list of brokers to send queries to
+   * @param headers Map of key and values of header which need to be used during http call
+   * @return A connection to the set of brokers specified
+   */
+  public static Connection fromHostList(List<String> brokers, Map<String, String> headers) {
+    return new Connection(brokers, _transportFactory.buildTransport(headers));
   }
 }
--- a/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
+++ b/pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
@@ -18,9 +18,6 @@
  */
 package org.apache.pinot.client;
 
-import org.apache.yetus.audience.InterfaceAudience;
-import org.apache.yetus.audience.InterfaceStability;
-
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -30,8 +27,6 @@ import java.util.Properties;
 /**
  * Creates connections to Pinot, given various initialization methods.
  */
-@InterfaceAudience.Public
-@InterfaceStability.Evolving
 public class ConnectionFactory {
   static PinotClientTransportFactory _transportFactory = new JsonAsyncHttpPinotClientTransportFactory();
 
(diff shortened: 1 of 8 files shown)


jitendratheta:
@npawar @mcvsubbu Can you please review this diff.

pinot-api/src/main/java/org/apache/pinot/client/JsonAsyncHttpPinotClientTransport.java
```diff
@@ -58,7 +65,13 @@ public BrokerResponse executeQuery(String brokerAddress, String query)
 
       final String url = "http://" + brokerAddress + "/query";
 
-      final Future<Response> response = _httpClient.preparePost(url)
+      AsyncHttpClient.BoundRequestBuilder request = _httpClient.preparePost(url);
+
+      if(_headers != null) {
+        _headers.forEach( (k,v) -> request.addHeader(k, v));
```
mcvsubbu: nit: please remove the extra space bet "(" and add a space after "k,"

pinot-api/src/test/java/org/apache/pinot/client/PreparedStatementTest.java
```diff
@@ -69,9 +70,10 @@ public String getLastQuery() {
 
   class DummyPinotClientTransportFactory implements PinotClientTransportFactory {
     @Override
-    public PinotClientTransport buildTransport() {
+    public PinotClientTransport buildTransport(Map<String, String> headers) {
       return _dummyPinotClientTransport;
     }
+
```
npawar: nit: remove newline

jitendratheta: Thanks for pointing. Made the changes !

pinot-api/src/main/java/org/apache/pinot/client/JsonAsyncHttpPinotClientTransport.java
jitendratheta: Thanks. Made the changes.

jitendratheta:
> Since this is not common to pass additional headers from the client, keep the existing methods and just add the new method that accepts header map.

The idea here was to keep the interface level changes minimum. Thought of doing this also. It ran in more code changes with two exposed function in PinotClientTransportFactory.java  and its implementation in concrete classes. Following the same type of separation for class JsonAsyncHttpPinotClientTransport, it also needs to have two constructors. 

 With this in mind, I felt the current approach of having a single function and passing headers as null if not present makes a better fit here. Any thoughts are welcomed :)

mcvsubbu:
@kishoreg if you are ok with this, please go ahead and approve it, thanks.

kishoreg:
Thanks @jitendratheta  for the explanation. It looks good to me overall.  Have a few more questions before I sign off to make sure I understand the scenario. 
These headers seem to be set at the time of creating the connection but it's passed as part of every request. I am curious to know if adding another method to PinotClientTransport that takes the headers on every request solve your issue. 

Can you also list some sample headers

jitendratheta:
> Thanks @jitendratheta for the explanation. It looks good to me overall. Have a few more questions before I sign off to make sure I understand the scenario.
> These headers seem to be set at the time of creating the connection but it's passed as part of every request. I am curious to know if adding another method to PinotClientTransport that takes the headers on every request solve your issue.
> 
> Can you also list some sample headers

Some sample headers are as below
`curl -H 'RPC-Caller:curl' -H 'RPC-Service:pinot-staging' -d '{"pql":"select * from table"}'  localhost:1234/query` 
In this use case, all the calls go via the same endpoint and actual service is defined by "RPC-Service" header.

 We create a Connection object at our microservice startup and pass it to various Dao layer classes.
Dao layer class just pass the query to the connection object.
  Headers are mostly constant for a service. So passing it once at while creating connection seems to be appropriate. Also, the Dao layer does not have to deal with headers to pass in the connection object.

pinot-api/src/main/java/org/apache/pinot/client/PinotClientTransportFactory.java
```diff
@@ -18,9 +18,11 @@
...
 package org.apache.pinot.client;
 
+import java.util.Map;
+
 /**
  * Factory for client transports.
  */
 interface PinotClientTransportFactory {
-  PinotClientTransport buildTransport();
+  PinotClientTransport buildTransport(Map<String, String> headers);
```
mayankshriv: This is an interface change and can potentially impact other users of this module. Please follow Pinot guidelines for public interface changes.

sunithabeeram: +1

It looks like this can be extended by others and this signature change will break compatibility. This interface should have been marked as public:stable. We recently add annotations for a few foundation classes, we'll need to do the same here.

@jitendratheta, if you can review the guidelines here: https://github.com/apache/incubator-pinot/blob/master/pinot-common/src/main/java/org/apache/pinot/annotations that will be great. Javadocs for InterfaceStability.java provide more specific info. As noted there, to keep compatibility, the best option is to provide a default implementation for the new method while keeping or marking the old one as deprecated. 

We'll look into making one more pass at updating interface annotations, but for now, it looks like this class should have Audience.Public and Stability.Stable annotations. Could you add the same?

@mcvsubbu, @Jackie-Jiang - FYI. Please chime in as appropriate.

kishoreg: I had the same impression. But ConnectionFactory is the public interface and he has maintained backward compatibility by adding a new method.

jitendratheta: @sunithabeeram as per your recommendation I made this interface as public:stable.

This interface is not exposed as a public interface
  1. should we make it `public interface PinotClientTransportFactory`?
  2. If we do so, PinotClientTransportFactory is not accepted as an argument by any class. ConnectionFactory class is using it by implicit object creation inside the class. then does it make sense to have it public? or should we also make `PinotClientTransport` as public and one of the argument to be passed to function `ConnectionFactory.fromHostList` and let the caller decide the implementation of it?

jitendratheta: Or else I was thinking why not to keep the annotation as Private for class `PinotClientTransportFactory` as it is not exposed outside as public?

kishoreg: I don't see anyone extending PinotClientTransportFactory outside of Pinot code. I am fine with making it private

jitendratheta: @sunithabeeram @mayankshriv Can this thread be resolved if you are ok with the changes?

sunithabeeram: The code structure seems a bit weird to me. If ConnectionFactory is the one that can be extended, then shouldn't that actually be an interface and made Public.Stable? Its odd that we have a specific implementation there and expect clean extensibility.

Given the lack of clarity, it would be great if the javadocs for ConnectionFactory and PinotClientConnectionFactory be updated to expand on how they should be used - ie, what should change when new transports are considered/added. @kishoreg, hope you can provide appropriate input here for Jitendra.

jitendratheta: @sunithabeeram, Had a discussion with @kishoreg and added discussed changes.

pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
```diff
@@ -18,13 +18,20 @@
  */
 package org.apache.pinot.client;
 
+import org.apache.yetus.audience.InterfaceAudience;
```
sunithabeeram: Please use annotations from pinot-common.

pinot-api/src/main/java/org/apache/pinot/client/PinotClientTransportFactory.java
sunithabeeram: @kishoreg it looks like the current ConnectionFactory is tied to JsonHttp transport. How would things work if Https has to be supported? It appears to me that the transport factory itself needs to be injected to ConnectionFactory class in that case. 

The class has been marked as evolving. Is it because you see new methods getting added in the future or do you have other cleanup/changes in mind?

pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
jitendratheta: acknowledge

kishoreg:
@sunithabeeram  Annotating the interface requires us to bring pinot-common as a dependency. I am not sure about adding pinot-common as a dependency to pinot-api. This needs more discussion. Since this PR is not breaking backward compatibility, let's try to keep this PR simple and unblock Jitendra. 

@jitendratheta, please remove the annotation and the dependency on pinot-common and we should be good to go.

sunithabeeram:
@kishoreg can you please create an issue for what you see is wrong with pinot-common dependency? I've captured the other points I raised here in this https://github.com/apache/incubator-pinot/issues/4265.

kishoreg:
pinot-common will bring in a lot of additional jars and will add burden on the users to resolve conflicts/exclude jars. 

@jitendratheta Please revert the commit that brings in the dependency on pinot-common. We will clean up the interfaces when we address #4265.

jitendratheta:
Hi @kishoreg, reverted the last change that brings in the dependency on pinot-common.

pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
kishoreg: We can remove this.

pinot-api/src/main/java/org/apache/pinot/client/ConnectionFactory.java
```diff
@@ -18,13 +18,20 @@
...
 import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 
 
 /**
  * Creates connections to Pinot, given various initialization methods.
  */
+@InterfaceAudience.Public
```
kishoreg: Remove this for now. We will address these things later in another PR

pinot-api/src/main/java/org/apache/pinot/client/PinotClientTransportFactory.java
```diff
@@ -18,9 +18,28 @@
  */
 package org.apache.pinot.client;
 
+import org.apache.yetus.audience.InterfaceAudience;
```
kishoreg: Same.  remove this

kishoreg:
Thanks @jitendratheta. LGTM!

PR #4244, opened 2019-05-28
Add support for custom record-readers in the create-segment tool

sunithabeeram:
Currently,  via pinot-admin, we can only work with builtin record-readers. This change adds support to work with either the format option or recordReaderPath option in the generationConfigFile.

For users to successfully use this for custom record readers, the pom file will need to updated to bundle the appropriate jars. I've included commented out examples for using ORC.

Testing done:
Tested generating pinot segment from ORC file with the following options:
cat ~/config:
{
"recordReaderPath": "org.apache.pinot.orc.data.readers.ORCRecordReader"
}

Command used to generate segments:
bin/pinot-admin.sh CreateSegment -dataDir ~/segments -schemaFile ~/schema -tableName testTable -outDir ~/segments_out -generatorConfigFile ~/config -segmentName testTable

### DIFF
--- a/pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
+++ b/pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
@@ -228,15 +228,17 @@ public class CreateSegmentCommand extends AbstractBaseAdminCommand implements Co
       }
     }
 
-    FileFormat configFormat = segmentGeneratorConfig.getFormat();
-    if (_format == null) {
-      if (configFormat == null) {
-        throw new RuntimeException("Format cannot be null in config file.");
-      }
-      _format = configFormat;
-    } else {
-      if (configFormat != _format && configFormat != FileFormat.AVRO) {
-        LOGGER.warn("Find format conflict in command line and config file, use config in command line: {}", _format);
+    if (segmentGeneratorConfig.getRecordReaderPath() == null) {
+      FileFormat configFormat = segmentGeneratorConfig.getFormat();
+      if (_format == null) {
+        if (configFormat == null) {
+          throw new RuntimeException("Either recordReaderPath (via generatorConfigFile option) or format must be specified.");
+        }
+        _format = configFormat;
+      } else {
+        if (configFormat != _format && configFormat != FileFormat.AVRO) {
+          LOGGER.warn("Find format conflict in command line and config file, use config in command line: {}", _format);
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -228,15 +228,17 @@ public boolean execute()
...
-    FileFormat configFormat = segmentGeneratorConfig.getFormat();
-    if (_format == null) {
-      if (configFormat == null) {
-        throw new RuntimeException("Format cannot be null in config file.");
-      }
-      _format = configFormat;
-    } else {
-      if (configFormat != _format && configFormat != FileFormat.AVRO) {
-        LOGGER.warn("Find format conflict in command line and config file, use config in command line: {}", _format);
+    if (segmentGeneratorConfig.getRecordReaderPath() == null) {
```
jenniferdai: // Allow for instantiation general record readers from a record reader path passed into segment generator config
    // If this is set, this will override the file format
    if (recordReaderPath != null) {
      if (fileFormat != FileFormat.OTHER) {
        // NOTE: we currently have default file format set to AVRO inside segment generator config, do not want to break
        // this behavior for clients.
        LOGGER
            .warn("Using class: {} to read segment, ignoring configured file format: {}", recordReaderPath, fileFormat);
      }


Currently, in record reader factory, we have this default. Up to you, but potentially, we could follow the same model though requiring users to set file format to other seems annoying... maybe we can just leave a comment that the behavior is different or put it in relevant docs.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -291,7 +293,10 @@ public boolean execute()
     File[] files = dir.listFiles(new FilenameFilter() {
       @Override
       public boolean accept(File dir, String name) {
-        return name.toLowerCase().endsWith(_format.toString().toLowerCase());
+        if (_format != null) {
```
jenniferdai: do we want to validate the record reader class, eg: try to instantiate it and make sure it's valid?

sunithabeeram: Didn't get notified about your comments on this PR :( - sorry about the delay in response.

Let me check what the exception path message is to the users - if its sensible enough, I will just let it take that path instead of doing an additional validation.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
sunithabeeram: Let me check the other path and keep them consistent if possible.

npawar:

PR #4245, opened 2019-05-28
Make ControllerLeadershipManager thread-safe and register it to PARTICIPANT HelixManager

Jackie-Jiang:
ControllerLeadershipManager should be registered to PARTICIPANT HelixManager instead of CONTROLLER HelixManager so that the Helix callbacks and custom callbacks won't affect each other.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
@@ -18,10 +18,10 @@
  */
 package org.apache.pinot.controller;
 
+import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import javax.annotation.concurrent.ThreadSafe;
 import org.apache.helix.HelixManager;
-import org.apache.helix.api.listeners.ControllerChangeListener;
 import org.apache.pinot.common.metrics.ControllerGauge;
 import org.apache.pinot.common.metrics.ControllerMetrics;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
```diff
@@ -18,50 +18,66 @@
...
  */
+@ThreadSafe
 public class ControllerLeadershipManager {
-
   private static final Logger LOGGER = LoggerFactory.getLogger(ControllerLeadershipManager.class);
 
-  private HelixManager _helixManager;
-  private ControllerMetrics _controllerMetrics;
-  private volatile boolean _amILeader = false;
+  private final HelixManager _helixManager;
```
jackjlli: It'd be good to rename it to `_helixControllerManager` since it will always be from Helix controller instead of a Helix participant.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
```diff
@@ -18,50 +18,66 @@
...
+  private final ControllerMetrics _controllerMetrics;
 
-  private Map<String, LeadershipChangeSubscriber> _subscribers = new ConcurrentHashMap<>();
+  private Map<String, LeadershipChangeSubscriber> _subscribers = new HashMap<>();
+  private boolean _amILeader = false;
```
jackjlli: It might be good to mark it as `volatile`?

Jackie-Jiang: It does not need to be as the methods are synchronized

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
Jackie-Jiang: Done

PR #4246, opened 2019-05-28
[TE] Cube algorithm for ratio metrics

cyenjung:
__Changes__
1. Implement ratio cube related classes.
2. Add new unit tests for ratio cube.

__Tests__
1. Passed the existing and new unit tests.
2. The integration test is done through ThirdEye API.

Related PR:
https://github.com/apache/incubator-pinot/pull/4213

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/cost/OeRatioCostFunction.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/cost/OeRatioCostFunction.java
@@ -3,6 +3,7 @@ package org.apache.pinot.thirdeye.cube.cost;
 import com.google.common.base.Preconditions;
 import com.google.common.math.DoubleMath;
 import java.util.Map;
+import org.apache.pinot.thirdeye.cube.data.cube.CubeUtils;
 
 
 public class OeRatioCostFunction implements CostFunction {
@@ -27,13 +28,13 @@ public class OeRatioCostFunction implements CostFunction {
    * @param parentChangeRatio the changeRatio between baseline and current value of parent node.
    * @param baselineValue the baseline value of the current node.
    * @param currentValue the current value of the current node.
-   * @param baselineSize
-   * @param currentSize
+   * @param baselineSize the size of baseline node.
+   * @param currentSize the size of current node.
    * @param topBaselineValue the baseline value of the top node.
    * @param topCurrentValue the current value of the top node.
+   * @param topBaselineSize the size of baseline data cube .
+   * @param topCurrentSize the size of current data cube .
    *
-   * @param topBaselineSize
-   * @param topCurrentSize
    * @return the cost that consider change difference, change changeRatio, and node size.
    */
   @Override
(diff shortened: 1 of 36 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/summary/Summary.java
```diff
@@ -408,10 +407,10 @@ public OneSideErrorRowInserter(RowInserter basicRowInserter, boolean side) {
     public void insertRowToDPArray(DPArray dp, CubeNode node, double targetRatio)  {
       // If the row has the same change trend with the top row, then it is inserted.
       if ( side == node.side() ) {
-        // When do oneSide, we try to make the root's changeRatio close to 1 in order to see the major root causes.
-        if ( (side && Double.compare(targetRatio, 1d) > 0) || (!side && Double.compare(targetRatio, 1d) < 0)) {
-          targetRatio = 1d;
-        }
+//        // When do oneSide, we try to make the root's changeRatio close to 1 in order to see the major root causes.
+//        if ( (side && Double.compare(targetRatio, 1d) > 0) || (!side && Double.compare(targetRatio, 1d) < 0)) {
+//          targetRatio = 1d;
```
jihaozh: remove the commented code?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/dashboard/resources/SummaryResource.java
```diff
@@ -180,4 +181,69 @@ public String buildSummaryManualDimensionOrder(@QueryParam("dataset") String dat
     }
     return OBJECT_MAPPER.writeValueAsString(response);
   }
+
+  @GET
+  @Path(value = "/summary/autoRatioDimensionOrder")
```
jihaozh: how about sharing the same endpoint with additive metrics and run a separate code path in the back end? Instead of making different frontend calls.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/cost/OeRatioCostFunction.java
```diff
@@ -0,0 +1,103 @@
+package org.apache.pinot.thirdeye.cube.cost;
+
+import com.google.common.base.Preconditions;
+import com.google.common.math.DoubleMath;
+import java.util.Map;
+import org.apache.pinot.thirdeye.cube.data.cube.CubeUtils;
+
+
+public class OeRatioCostFunction implements CostFunction {
```
jihaozh: Could you add some documents for this class? Is this cost function works for all ratio metrics or it only works for OE ratio?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/summary/Summary.java
cyenjung: Nice catch! Removed.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/cube/cost/OeRatioCostFunction.java
cyenjung: Renamed to `RatioCostFunction` and added JavaDoc.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/dashboard/resources/SummaryResource.java
cyenjung: The integration is difficult to do for now because a ratio metric needs to be broken down to `numerator` and `denominator`. Although the information is available, it cannot be used directly. 

Specifically, `ratio = id123/id456` where `id123` and `id456` are the numerator and denominator. We need to create another query to do a reverse lookup and get the original metric name. Otherwise, ThirdEye won't be able to get metrics by the `id` names.

I suggest that we do the integration in another PR. This PR will focus on the implementation of the Ratio Cube. Users can use it through API for now and start to give us feedback. Does that work for you?

jihaozh: sure. sounds good.

PR #4249, opened 2019-05-28
[TE] make holter winters rule to return the current timeseries

jihaozh:
make holter winters rule to return the current time series so that in the preview UI it can show the current time series.


xiaohui-sun:
LTGM

PR #4250, opened 2019-05-28
SoC - Separate out Tuning from Translator

akshayrai:
Currently tuning is tied together with the translator. Each time we need to tune the detection, we have to call the translator which tries to re-convert the yaml config to detection object. This PR separates the tuner from the translator so that it can be triggered directly on the detection object.

Changes:
* Introduced `ConfigTranslator` interface for detection and subscription
* Enforce validation of raw and parsed config through interface (`template-pattern`)
* Moved out tuning related code under `DetectionConfigTuner`
* Move translator specific code under translator package
* Remove `YamlTranslationResult`; Cleaned up buildMetricUrn; Consistently use `componentKey` throughout instead of `componentName`
* Added and updated unit tests


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipeline.java
```diff
@@ -90,22 +90,21 @@ private void initComponents() throws Exception {
...
+      for (String componentKey : componentSpecs.keySet()) {
+        Map<String, Object> componentSpec = MapUtils.getMap(componentSpecs, componentKey);
+        if (!instancesMap.containsKey(componentKey)){
+          instancesMap.put(componentKey, createComponent(componentSpec));
         }
       }
 
-      for (String componentName : componentSpecs.keySet()) {
-        Map<String, Object> componentSpec = MapUtils.getMap(componentSpecs, componentName);
+      for (String componentKey : componentSpecs.keySet()) {
```
xiaohui-sun: Can we combine the two for loops?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionUtils.java
```diff
@@ -67,13 +67,22 @@ public static boolean isReferenceName(String key) {
     return key.startsWith("$");
   }
 
-  // get the component name from the reference key
-  // example "$myRule:ALGORITHM:0" -> "myRule:ALGORITHM:0"
-  public static String getComponentName(String key) {
-    if (isReferenceName(key)) return key.substring(1);
+  // Extracts the component key from the reference key
```
xiaohui-sun: Thanks for making  the naming more clear!

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipeline.java
akshayrai: Not really! In the second loop, we replace all component references with the components. This comes by looking up the componentKey to Component mapping which is generated in the first loop.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/DetectionConfigTuner.java
```diff
@@ -0,0 +1,158 @@
...
+    // round to daily boundary
+    DateTimeZone timezone = DateTimeZone.forID(this.datasetConfig.getTimezone() == null ? DEFAULT_TIMEZONE : this.datasetConfig.getTimezone());
+    DateTime start = new DateTime(startTime, timezone).withTimeAtStartOfDay();
+    DateTime end =  new DateTime(endTime, timezone).withTimeAtStartOfDay();
+    Interval window = new Interval(start, end);
+
+    // TODO can you use componentProps instead? what is the difference btw component and componentSpec
+    Map<String, Object> existingComponentSpec =
+        detectionConfig.getComponents().containsKey(componentKey) ?
+            MapUtils.getMap(detectionConfig.getComponents(), componentKey) : Collections.emptyMap();
```
jihaozh: should here be `MapUtils.getMap(detectionConfig.getComponentSpecs(), componentKey)  ` ? By calling `detectionConfig.getComponents()`, it will return a initilized  component object.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/DetectionConfigTuner.java
```diff
@@ -0,0 +1,158 @@
...
+      String componentKey = componentSpec.getKey();
+      Map<String, Object> componentProps = ConfigUtils.getMap(componentSpec.getValue());
+
+      // For tunable components, the model params are computed from user supplied yaml params and previous model params.
+      String componentClassName = componentProps.get(PROP_CLASS_NAME).toString();
+      String type = DetectionUtils.getComponentType(componentKey);
+      if (!TURNOFF_TUNING_COMPONENTS.contains(type) && DETECTION_REGISTRY.isTunable(componentClassName)) {
+        try {
+          componentProps.putAll(getTunedSpecs(componentKey, componentProps, tuningWindowStart, tuningWindowEnd));
+          detectionConfig.setComponentSpecs(componentProps);
```
jihaozh: is `componentProps` the properties for one component? should we set it as  `setComponentSpecs` of the whole detection config?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/rootcause/impl/MetricEntity.java
```diff
@@ -102,4 +104,15 @@ public static MetricEntity fromURN(String urn) {
   public static MetricEntity fromSlice(MetricSlice slice, double score) {
     return fromMetric(score, slice.getMetricId(), slice.getFilters());
   }
+
+  public static String buildMetricUrn(Map<String, Collection<String>> filterMaps, long metricId) {
```
jihaozh: IMO, this function should not be in MetricEntity class because the input and return types are not MetricEntity.  Maybe put it in DetectionUtils?

akshayrai: I think it makes sense to keep it here. I have updated the method to return a `MetricEntity` and renamed the method to `fromMetric` to be consistent with the other methods.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipeline.java
xiaohui-sun: You can do  update right after inserting.
for (String componentKey : componentSpecs.keySet()) {
// check if key exists
// if not then insert (key, value)
// update the map (value)
}

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/DetectionConfigTuner.java
```diff
@@ -112,13 +120,14 @@ public DetectionConfigTuner(DetectionConfigDTO config, DataProvider dataProvider
...
-        detectionConfig.getComponents().containsKey(componentKey) ?
-            MapUtils.getMap(detectionConfig.getComponents(), componentKey) : Collections.emptyMap();
-
     // TODO: if dimension drill down applied, pass in the metric urn of top dimension
-    return tunable.tune(existingComponentSpec, window, this.metricUrn);
+    tunedSpec.putAll(tunable.tune(componentProps, window, this.metricUrn));
+
+    // Hack to retain the raw yaml parameters.
+    // The tunable requires raw yaml params and previously tuned params to generate fresh params
+    tunedSpec.put(PROP_YAML_PARAMS, yamlParams);
```
jihaozh: After tuning, do we need to retain the raw yaml parameters?

akshayrai: That's a good question! I included it to keep it consistent with,
1. how we retain the raw params for other non-tunable components
2. how we include this in the spec before calling the tuner

jihaozh: make sense. it can also be used in re-tune.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionPipeline.java
akshayrai: Sorry, did I miss something! Are you talking about merging the 2 sequential loops on `componentSpecs` or the 2 nested loops?

My earlier comment was addressed assuming the former case. I am still not convinced even if it is the latter case. Note that the two loops are on `componentSpecs` and `componentSpec`. The parent loop iterates over each component and the nested loop iterates over the specifications of each component.

I do not have a complete understanding of the reference logic here either. Nevertheless, if required I will make this change in a separate PR as it is unrelated to this change.

xiaohui-sun: Sure. Thanks for the details.

Issue #3998, opened 2019-03-21
Upgrade to use  Kafka release 2.1.1


mcvsubbu:
Kafka release has a stable version of 2.1.1 (https://kafka.apache.org/downloads)

Pinot should upgrade to a more recent version of Kafka

This release is source incompatible with 0.9 (that Pinot currently uses).  We need to:
* Evaluate the compatiblity of new kafka client with older brokers. If they are not compatbile, then we should upgrade in steps getting to 2.1 eventually.
* Ensure that new Kafka client still provides the functionality we rely on, and move Pinot to use new Kafka clients.

Meanwhile, one solution is to use pluggable streams, but we need to make sure that there is no classpath collision and undesirable behavior

mcvsubbu:
Maybe we can consider moving to this
https://github.com/linkedin/li-apache-kafka-clients if it is compatible

kishoreg:
The first step would be decouple Pinot core from kafka right. Once we have that, we can add multiple implementations likafka, Kafka-0.9.x, Kafka-1.x kafka2.x. 

This will be a good test/use of pluggable streaming source we built.

mcvsubbu:
Sure, we can take all of the current Kafka code to a separate repo that depends on kafka=0.9.x and build other repos for newer Kafkas (if they change source compat). That should be easy to do since we have pluggable streams.

kishoreg:
I was thinking along the same lines. Let’s do that. That will allow us to make sure we are backwards compatible

mcvsubbu:
Not sure how the realtime integration tests will run, though.

mcvsubbu:
And we need to keep kafka client out of the distribution I suppose?

kishoreg:
Yes, we can keep a default implementation but have the ability to configure
something else. This will be similar to remoteFS - > NFS/Azure/HDFS

ananthdurai:
👋 Hi, any update on the Kafka 2.X support? We are running a different version of Kafka to maintain Pinot, would like to move along with the rest of the infrastructure.

npawar:
We're planning to work on this issue this quarter. We (@mcvsubbu  and I) will do some refactoring, such that the kafka code will be separated out as a stream plug in. This will make it easier for folks to plug in the versions of kafka they wish to use. 

Here's the tentative steps we plan to take:
[...]

PR #4252, opened 2019-05-28
pinot_kafka_0.9.0.1 module

npawar:
Step 3 of https://github.com/apache/incubator-pinot/issues/3998
As discussed in the issue, we are planning to move out kafka stream implementation related classes into its own module `pinot-kafka-0.9.0.1`. This should enable folks to plug in implementations for other kafka versions.
TODO: the pinot-integration-tests, pinot-tools and pinot-perf still depend on pinot-kafka-0.9.0.1 for the tests and dummy setup. We need to figure out a way by which we do not need to specify the dependency, but be able to inject the classes depending on the version we wish to use.


snleee:
What is the plan on the naming for the module for different Kafka version implementation? I think that we should keep one implementation for each major version (e.g. `0.9.x`, `1.0.x`,`2.0.x`...) instead of having each module for each Kafka version.

mcvsubbu:
> What is the plan on the naming for the module for different Kafka version implementation? I think that we should keep one implementation for each major version (e.g. `0.9.x`, `1.0.x`,`2.0.x`...) instead of having each module for each specific Kafka version.

In principle yes, but are there guarantees that kafka does not change the api within these smaller releases? I am not sure.

snleee:
> > What is the plan on the naming for the module for different Kafka version implementation? I think that we should keep one implementation for each major version (e.g. `0.9.x`, `1.0.x`,`2.0.x`...) instead of having each module for each specific Kafka version.
> 
> In principle yes, but are there guarantees that kafka does not change the api within these smaller releases? I am not sure.

I would still prefer to keep one package per major version. If we need to bump up the minor version and kafka is not backward compatible, we can change our code to use new api. I don't like the approach of creating module for each version. 

For example, let's say that we want to bump up the version from 2.1.1 to 2.1.2 and they are backward compatible (or not). Are we going to create another module for `2.1.2` and copy more or less the same code? or are we going to change the package module name?

mcvsubbu:
> > > What is the plan on the naming for the module for different Kafka version implementation? I think that we should keep one implementation for each major version (e.g. `0.9.x`, `1.0.x`,`2.0.x`...) instead of having each module for each specific Kafka version.
> > 
> > 
> > In principle yes, but are there guarantees that kafka does not change the api within these smaller releases? I am not sure.
> 
> I would still prefer to keep one package per major version. If we need to bump up the minor version and kafka is not backward compatible, we can change our code to use new api. I don't like the approach of creating module for each version.
> 
> For example, let's say that we want to bump up the version from 2.1.1 to 2.1.2 and they are backward compatible (or not). Are we going to create another module for `2.1.2` and copy more or less the same code? or are we going to change the package module name?

Each time they change API, we create a new version. If they never change API starting 2.1, we are good forever. How does that sound?

snleee:
> > > > What is the plan on the naming for the module for different Kafka version implementation? I think that we should keep one implementation for each major version (e.g. `0.9.x`, `1.0.x`,`2.0.x`...) instead of having each module for each specific Kafka version.
> > > 
> > > 
> > > In principle yes, but are there guarantees that kafka does not change the api within these smaller releases? I am not sure.
> > 
> > 
> > I would still prefer to keep one package per major version. If we need to bump up the minor version and kafka is not backward compatible, we can change our code to use new api. I don't like the approach of creating module for each version.
> > For example, let's say that we want to bump up the version from 2.1.1 to 2.1.2 and they are backward compatible (or not). Are we going to create another module for `2.1.2` and copy more or less the same code? or are we going to change the package module name?
> 
> Each time they change API, we create a new version. If they never change API starting 2.1, we are good forever. How does that sound?

I see. As long as we don't add too many modules and have a clear guideline, I'm fine with it. I just thought that associating a specific version number to the module name is not the flexible approach if we want to bump up Kafka versions frequently. Thanks for the quick comment!

mcvsubbu:
> > > > > What is the plan on the naming for the module for different Kafka version implementation? I think that we should keep one implementation for each major version (e.g. `0.9.x`, `1.0.x`,`2.0.x`...) instead of having each module for each specific Kafka version.
> > > > 
> > > > 
> > > > In principle yes, but are there guarantees that kafka does not change the api within these smaller releases? I am not sure.
> > > 
> > > 
> > > I would still prefer to keep one package per major version. If we need to bump up the minor version and kafka is not backward compatible, we can change our code to use new api. I don't like the approach of creating module for each version.
> > > For example, let's say that we want to bump up the version from 2.1.1 to 2.1.2 and they are backward compatible (or not). Are we going to create another module for `2.1.2` and copy more or less the same code? or are we going to change the package module name?
> > 
> > 
> > Each time they change API, we create a new version. If they never change API starting 2.1, we are good forever. How does that sound?
> 
> I see. As long as we don't add too many modules and have a clear guideline, I'm fine with it. I just thought that associating a specific version number to the module name is not the flexible approach if we want to bump up Kafka versions frequently. Thanks for the quick comment!

When the new kafka module is added (hopefully by @ananthdurai ) we will have a README in the new dir saying that attempts should be made to bump kafka revisions as needed until kafka interface changes. We can even name the directory as kafka-2.1-or-higher.

mcvsubbu:
I am fine with this. @ananthdurai , @kishoreg  anyone have any opinions? @ananthdurai  are you good  in terms of implementing a new adapter for newer versions of kafka? Please follow the conversation in this PR for naming conventions and decide accordingly.

kishoreg:
I missed the comment from @snleee of naming. I agree with him and prefer 0.9.x and I think Kafka guarantees that API's won't change in minor versions. The problem with naming it 0.9.0.1 is that its not a common convention and users might think that it works only for that version of Kafka which is not the case. See Flink naming convention for example https://github.com/apache/flink/tree/master/flink-connectors.

Another suggestion is to create a subfolder for all implementations if possible. We have too many modules at the root pinot-orc, pinot-parquet, pinot-hadoop-filesytem. If it's not too hard, we can move kafka module into a subfolder.

npawar:
FYI @chenboat @jamesyfshao  This directly affects the kafka implementation you currently use. There should be no impact, but please take a look

jamesyfshao:
Thanks for the update, we have some plans to upgrade to later version of Kafka (> 1.0). I think we should open a new module named like pinot-kafka-1.0 when we work on the new version? 

We also have plans to migrate away from the scala-based kafka simple consumer to the [new consumer model](https://www.confluent.io/blog/tutorial-getting-started-with-the-new-apache-kafka-0-9-consumer-client/) as it is deprecated since [0.10](https://cwiki.apache.org/confluence/display/KAFKA/KIP-109%3A+Old+Consumer+Deprecation) and the new consumer has compatible API as well more features. Do you guys have any work planned for it or we can investigate more into this effort? @chenboat

npawar:
> I missed the comment from @snleee of naming. I agree with him and prefer 0.9.x and I think Kafka guarantees that API's won't change in minor versions. The problem with naming it 0.9.0.1 is that its not a common convention and users might think that it works only for that version of Kafka which is not the case. See Flink naming convention for example https://github.com/apache/flink/tree/master/flink-connectors.
> 
> Another suggestion is to create a subfolder for all implementations if possible. We have too many modules at the root pinot-orc, pinot-parquet, pinot-hadoop-filesytem. If it's not too hard, we can move kafka module into a subfolder.

Sure, we can consider keeping it just 0.9, 1.0 etc, will check regarding the kafka guarantee about no changes in minor versions.

Regarding sub folder for all implementations, are you suggesting we put all implementations in 1 module (pinot-streams?)

npawar:
> Thanks for the update, we have some plans to upgrade to later version of Kafka (> 1.0). I think we should open a new module named like pinot-kafka-1.0 when we work on the new version?
> 
> We also have plans to migrate away from the scala-based kafka simple consumer to the [new consumer model](https://www.confluent.io/blog/tutorial-getting-started-with-the-new-apache-kafka-0-9-consumer-client/) as it is deprecated since [0.10](https://cwiki.apache.org/confluence/display/KAFKA/KIP-109%3A+Old+Consumer+Deprecation) and the new consumer has compatible API as well more features. Do you guys have any work planned for it or we can investigate more into this effort? @chenboat

Yes, you can create a new pinot-kafka-1.0 when you decide to work with newer version. We don't have any immediate plans of investigating it. It would be great if you can help out with that

kishoreg:
> > I missed the comment from @snleee of naming. I agree with him and prefer 0.9.x and I think Kafka guarantees that API's won't change in minor versions. The problem with naming it 0.9.0.1 is that its not a common convention and users might think that it works only for that version of Kafka which is not the case. See Flink naming convention for example https://github.com/apache/flink/tree/master/flink-connectors.
> > Another suggestion is to create a subfolder for all implementations if possible. We have too many modules at the root pinot-orc, pinot-parquet, pinot-hadoop-filesytem. If it's not too hard, we can move kafka module into a subfolder.
> 
> Sure, we can consider keeping it just 0.9, 1.0 etc, will check regarding the kafka guarantee about no changes in minor versions.
> 
> Regarding sub folder for all implementations, are you suggesting we put all implementations in 1 module (pinot-streams?)

@npawar Thanks. I am suggesting one module per implementation but just group them under one directory. Similar to the way Flink has organized its connectors. https://github.com/apache/flink/tree/master/flink-connectors. pinot-connectors is actually a pretty good name for the subfolder and we can have pinot-connector-kafka-0.9 module.

mcvsubbu:
> Thanks for the update, we have some plans to upgrade to later version of Kafka (> 1.0). I think we should open a new module named like pinot-kafka-1.0 when we work on the new version?
> 
> We also have plans to migrate away from the scala-based kafka simple consumer to the [new consumer model](https://www.confluent.io/blog/tutorial-getting-started-with-the-new-apache-kafka-0-9-consumer-client/) as it is deprecated since [0.10](https://cwiki.apache.org/confluence/display/KAFKA/KIP-109%3A+Old+Consumer+Deprecation) and the new consumer has compatible API as well more features. Do you guys have any work planned for it or we can investigate more into this effort? @chenboat

+1 to what @npawar said, but the naming of the folders are still up in the air.

 For example, if we name the directory 1.0.10 and it so happens that kafka does not change any APIs, and it is also valid for 2.0 then what happens?

snleee:
@mcvsubbu `kafka-1.0` module will import `1.0.x`  library. It may work with `kafka-2.0` if there was no API change but we don't guarantee it. We will create another module `kafka-2.0` that will import `2.0.x` and we can guarantee that it will work with kafka 2.0. 

@npawar I vote +1 for following flink approach to organize our modules. It's very clean. I also vote for `kafka-0.9` instead of `kafka-0.9.x`.

mcvsubbu:
> @mcvsubbu `kafka-1.0` module will import `1.0.x` library. It may work with `kafka-2.0` if there is no API change but we don't guarantee it. For explicit 2.0 support, we can create another module `kafka-2.0` that imports `2.0.x` and we can guarantee that it will work with kafka 2.0.
> 
> @npawar I vote +1 for following flink's approach to organize our modules. It's very clean. I also vote for `pinot-kafka-0.9` instead of `pinot-kafka-0.9.x`.

In other words, you are suggesting that we copy over all files even if nothing changes, and keep doing this for every (major) kafka release.

snleee:
@mcvsubbu We can still document that `1.0` module is compatible with `2.0` if we verified that they are instead of copying code. We can add module on demand when we need the new one after API change.

My suggestion was more about naming since I didn't like to put specific version because it seemed not to be the convention.

npawar:
The flink model looks good. I've changed the PR to follow that convention. Please take a look @mcvsubbu @kishoreg @snleee

pinot-connectors/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+
+    <!-- test -->
+    <dependency>
+      <groupId>org.testng</groupId>
+      <artifactId>testng</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+</project>
```
snleee: new line

pom.xml
```diff
@@ -55,6 +55,7 @@
     <module>pinot-azure-filesystem</module>
     <module>pinot-orc</module>
     <module>pinot-parquet</module>
+    <module>pinot-connectors</module>
```
snleee: Not related to this part but have a question with the main `pom.xml` file. In my understanding, we currently specify `kafka.version = 0.9.0.1` in the main pom.xml file.  After we add the new connector, each connector module will need to pull different kafka version. It will be good if we can start to think about the issue and how we will be managing that.

mcvsubbu: > Not related to this part but have a question with the main `pom.xml` file. In my understanding, we currently specify `kafka.version = 0.9.0.1` in the main pom.xml file. After we add the new connector, each connector module will need to pull different kafka version. It will be good if we can start to think about the issue and how we will be managing that.

I added some notes in Issue #3998

npawar: For this iteration itself, I can move it to the pinot-connector-kafka-0.9's pom.xml, something like this: https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-kafka-0.9/pom.xml
What say? @snleee

snleee: @npawar Does that mean that we can pass `kafka.version` as a property to mvn compilation command and pick default one (0.9 in our case) when it's not given?

If that's the case, it looks good to me 👍

kishoreg:
Thanks. This looks good. Please write up this convention when you get a chance.  Will be good to follow this convention for readers and deep storage.

pom.xml
npawar: This is one step in that direction. For now, I've just removed it from main pom, and kept it only in the connector.pom as a property. Next step will be to figure out a way to inject into it from the main pom

PR #4253, opened 2019-05-29
Add segment pre-processing Hadoop job

jackjlli:
We’ve seen some requests from client that we add partitioning, sorting and resizing functionalities in/before the actual Pinot related Hadoop jobs.
And without any control on the number/size of the segments, we may run into issues like generating too many small segments. It’s essential to add a layer before segment creation so that we lower the chances of making too small segments; the segment size could be more unified. On the other hand, when a new use case is onboard, clients don’t need to worry about partitioning and sorting.


Jackie-Jiang:
@kishoreg We never convert segment generation config to table config, it's always from table config to segment generation config in order to generate segments.

pinot-hadoop/pom.xml
```diff
@@ -77,6 +77,10 @@
           <groupId>org.apache.pinot</groupId>
```
jenniferdai: do we still need these exclusions? or did you change all the avro versions to a loewr version?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/PinotHadoopJobLauncher.java
```diff
@@ -29,7 +30,8 @@
 public class PinotHadoopJobLauncher {
```
jenniferdai: may be nice to also add an option to do preprocessing and other job types too

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/io/CombineAvroKeyInputFormat.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.avro.mapreduce.AvroKeyInputFormat;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.input.CombineFileInputFormat;
+import org.apache.hadoop.mapreduce.lib.input.CombineFileRecordReader;
+import org.apache.hadoop.mapreduce.lib.input.CombineFileRecordReaderWrapper;
+import org.apache.hadoop.mapreduce.lib.input.CombineFileSplit;
+
```
jenniferdai: Is this copied from hadoop? Can we use theirs?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
```diff
@@ -52,4 +52,17 @@
...
+  public static final String ENABLE_SORTING = "enable.sorting";
+  public static final String ENABLE_RESIZING = "enable.resizing";
+
+  public static final String PARTITION_COLUMN = "partition.column";
+  public static final String NUMBER_OF_PARTITIONS = "number.partitions";
+  public static final String PARTITION_FUNCTION = "partition.function";
+
+  public static final String SORTED_COLUMN = "sorted.column";
+
+  public static final String MINIMUM_NUMBER_OF_OUTPUT_FILES = "minimum.number.output.files";
```
jenniferdai: maybe min.num.output.files?

and max.num.records below to shorten parameters

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    logProperties();
+
+    if (_properties.getProperty(ENABLE_PARTITIONING) != null) {
+      _enablePartitioning = Boolean.parseBoolean(_properties.getProperty(ENABLE_PARTITIONING));
+    }
+
+    if (_properties.getProperty(ENABLE_SORTING) != null) {
+      _enableSorting = Boolean.parseBoolean(_properties.getProperty(ENABLE_SORTING));
+    }
+
```
jenniferdai: are we allowing both partitioning and sorting to proceed? i don't know too much about partitioning, but can anything break? do we always sort only after we've partitioned?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    if (!_enablePartitioning && _properties.getProperty(ENABLE_RESIZING) != null) {
+      _enableResizing = Boolean.parseBoolean(_properties.getProperty(ENABLE_RESIZING));
+    }
+
```
jenniferdai: Is this section already in opensource? I know it's part of the pbnj push job. If so, can we merge the logic?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    // get input/output paths.
+    _inputSegmentDir = properties.getProperty(PATH_TO_INPUT);
+    _preprocessedOutputDir = properties.getProperty(PATH_TO_OUTPUT);
+    _rawTableName = Preconditions.checkNotNull(_properties.getProperty(JobConfigConstants.SEGMENT_TABLE_NAME));
+
+    _hdfsDefaultClasspathDir = _properties.getProperty("hdfs.default.classpath.dir", null);
```
jenniferdai: nit: constant

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    if (_enablePartitioning) {
+      fetchPartitioningConfig();
+      _logger.info("partition column: {}", _partitionColumn);
+      _logger.info("number of partitions: {}", _numberOfPartitions);
+      _logger.info("partition column: {}", _partitionColumn);
+    }
+
+    if (_enableSorting) {
+      fetchSortingConfig();
+      _logger.info("sorted.column: {}", _sortedColumn);
```
jenniferdai: either use constant for parameter or words to avoid confusion/if we change param

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    if (_enableResizing) {
+      fetchResizingConfig();
+      _logger.info("minimum number of output files: {}", _numberOfOutputFiles);
+    }
+
+    _logger.info("Initializing a pre-computing job");
+    Job job = Job.getInstance(_configuration);
+
+    job.getConfiguration().set(JobContext.JOB_NAME, this.getClass().getName());
+    job.getConfiguration().set(MRJobConfig.MAPREDUCE_JOB_USER_CLASSPATH_FIRST, "true");
```
jenniferdai: please leave a comment about this param, (also, do we still need it with your latest avro fix?)

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    job.getConfiguration().set(FileOutputCommitter.SUCCESSFUL_JOB_OUTPUT_DIR_MARKER, "false");
```
jenniferdai: what is this parameter for?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    _logger.info("Number of reduce tasks for pre-computing: {}", numReduceTasks);
+    job.setNumReduceTasks(numReduceTasks);
+
+    // Sort config.
+    if (_sortedColumn != null) {
+      _logger.info("Adding sorted column: {} to job config", _sortedColumn);
+      job.getConfiguration().set(SORTED_COLUMN, _sortedColumn);
+
+      // Sorting is enabled. Adding sorted column value to mapper output key.
+      Schema sortedColumnSchema = schema.getField(_sortedColumn).schema();
```
jenniferdai: some users think they can add more than one sorted column we should make it clear here we don't support that by doing a check

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+      Schema sortedColumnAsKeySchema;
+      if (sortedColumnSchema.getType().equals(Schema.Type.UNION)) {
```
jenniferdai: is this for mv column? can we test this logic separately?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+        sortedColumnAsKeySchema = Schema.create(sortedColumnSchema.getType());
+      }
+      Schema.Field columnField = new Schema.Field(_sortedColumn, sortedColumnAsKeySchema, "sortedColumn", null);
+      fieldSet.add(columnField);
+    } else {
+      // If sorting is disabled, hashcode will be the only factor for sort/group comparator.
+      addHashCodeField(fieldSet);
+    }
+
+    Schema mapperOutputKeySchema = Schema.createRecord("record", "", "", false);
```
jenniferdai: can you leave a comment with what these parameters are?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+
+  /**
+   * Can be overridden to provide custom controller Rest API.
+   */
+  @Nullable
+  private ControllerRestApi getControllerRestApi() {
+    return _pushLocations != null ? new DefaultControllerRestApi(_pushLocations, _rawTableName) : null;
+  }
+
+  private void setDirPermission(Path dirPath)
```
jenniferdai: i think this is already in open source, right?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+        _partitionColumn = columnPartitionMap.keySet().iterator().next();
+        _numberOfPartitions = segmentPartitionConfig.getNumPartitions(_partitionColumn);
+        _partitionFunction = segmentPartitionConfig.getFunctionName(_partitionColumn);
+      }
+    } else {
+      _logger.info("Segment partition config is null for table: {}", _tableConfig.getTableName());
+    }
+
+    // Check whether configs overwritten by Hadoop properties.
+    if (_properties.getProperty(PARTITION_COLUMN) != null) {
```
jenniferdai: do we want to allow this? why don't we have table configs be source of truth?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+      _logger.info("Overriding partition number to: {}", _properties.getProperty(NUMBER_OF_PARTITIONS));
+      _numberOfPartitions = Integer.parseInt(_properties.getProperty(NUMBER_OF_PARTITIONS));
+    }
+  }
+
+  private void fetchSortingConfig() {
+    // Fetch sorting info from table config.
+    IndexingConfig indexingConfig = _tableConfig.getIndexingConfig();
+    List<String> sortedColumns = indexingConfig.getSortedColumn();
+    assert sortedColumns.size() <= 1;
```
jenniferdai: you sholud throw an informative error for clients

jenniferdai: also, another problem is that clients will declare sorted columns today but not expect that we sort in their hadoop job still, so we may need to inform everyone

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    if (sortedColumns.size() == 1) {
+      _sortedColumn = sortedColumns.get(0);
+    }
+    // Check whether configs overwritten by Hadoop properties.
```
jenniferdai: i don't think we should allow overriding in hadoop

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    if (_properties.getProperty(SORTED_COLUMN) != null) {
+      _logger.info("Overriding sorted column to: {}", _properties.getProperty(SORTED_COLUMN));
+      _sortedColumn = _properties.getProperty(SORTED_COLUMN);
+    }
+  }
+
+  private void fetchResizingConfig() {
+    if (_properties.getProperty(MINIMUM_NUMBER_OF_OUTPUT_FILES) != null) {
+      _numberOfOutputFiles = Integer.parseInt(_properties.getProperty(MINIMUM_NUMBER_OF_OUTPUT_FILES));
```
jenniferdai: does min mean we always use this number for output files? it seems like a misleading name in that case, either the variable or the param

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+    }
+  }
+
+  /**
+   * Finds the avro file in the input folder, and returns its avro schema
+   * @param inputPathDir Path to input directory
+   * @return Input schema
+   * @throws IOException exception when accessing to IO
+   */
+  private static Schema getSchema(String inputPathDir) throws IOException {
```
jenniferdai: we don't use this, right? we get schema from controller rest api

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+  private static DataFileStream<GenericRecord> getAvroReader(Path avroFile) throws IOException {
+    FileSystem fs = FileSystem.get(new Configuration());
+    if (avroFile.getName().endsWith("gz")) {
+      return new DataFileStream<>(new GZIPInputStream(fs.open(avroFile)), new GenericDatumReader<>());
+    } else {
+      return new DataFileStream<>(fs.open(avroFile), new GenericDatumReader<>());
+    }
+  }
+
+  private ArrayList<FileStatus> getDataFilesFromPath(FileSystem fs, Path inBaseDir) throws IOException {
```
jenniferdai: this code is elsewhere in open source, we should use that as there are more formats now

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,504 @@
...
+        if (fileName.endsWith(".json")) {
+          _logger.info("Adding json files: {}", fileStatus.getPath());
+          dataFileStatusList.add(fileStatus);
+        }
+      }
+    }
+    return dataFileStatusList;
+  }
+
+  private void addContainerLocalFiles(final Job job, final String cacheFolder, final Configuration conf)
```
jenniferdai: are we using the same distributed cache code that we use elsewhere in segment creation code? do we need to recopy this if so or is this for something else?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mapper/SegmentPreprocessingMapper.java
```diff
@@ -0,0 +1,85 @@
...
+    _enablePartition = Boolean.parseBoolean(configuration.get(ENABLE_PARTITIONING));
+    LOGGER.info("Enable partitioning? " + _enablePartition);
+  }
+
+  @Override
+  public void map(AvroKey<GenericRecord> record, NullWritable value, final Context context)
+      throws IOException, InterruptedException {
+    final GenericRecord inputRecord = record.datum();
+    final Schema schema = inputRecord.getSchema();
+    assert _outputSchema.equals(schema);
```
jenniferdai: whenever we do asserts, we should add a message for the user to help them

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/reducer/SegmentPreprocessingReducer.java
```diff
@@ -0,0 +1,84 @@
...
+  private int _maxNumberOfRecords;
+  private String _filePrefix;
+
+  @Override
+  public void setup(Context context) {
+    LOGGER.info("Using multiple outputs strategy.");
+    Configuration configuration = context.getConfiguration();
+    _multipleOutputs = new AvroMultipleOutputs(context);
+    _counter = new AtomicInteger();
+    _maxNumberOfRecords = configuration.getInt(MAXIMUM_NUMBER_OF_RECORDS, 0);
```
jenniferdai: why is default 0? if it's 0, what happens? will files sitll have records? which of max records and min files overrides? we should leave that documentation somewhere

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/reducer/SegmentPreprocessingReducer.java
```diff
@@ -0,0 +1,84 @@
...
+    LOGGER.info("Maximum number of records per file: {}", _maxNumberOfRecords);
+    _filePrefix = RandomStringUtils.randomAlphanumeric(4);
```
jenniferdai: why do we need this? Why is it 4?

pom.xml
```diff
@@ -706,6 +706,12 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>org.apache.avro</groupId>
```
jenniferdai: did we have a previous version of avro soemwhere?

pinot-server/src/test/java/org/apache/pinot/server/request/ScheduledRequestHandlerTest.java
```diff
@@ -126,9 +128,14 @@ public void testQueryProcessingException()
           @Nonnull
           @Override
           public ListenableFuture<byte[]> submit(@Nonnull ServerQueryRequest queryRequest) {
-            ListenableFuture<DataTable> dataTable = resourceManager.getQueryRunners().submit(() -> {
-              throw new RuntimeException("query processing error");
-            });
+            ListenableFuture<DataTable> dataTable = ((ListeningExecutorService)resourceManager.getQueryRunners()).submit(
```
jenniferdai: can you remove this portion? seems unrelated to the general one

pinot-hadoop/pom.xml
jackjlli: There's no harm with these exclusions. Let's keep it here.
And we don't need to change all the avro version as well. The only required changes is to downgrade the newly added version of avro-mapred-hadoop2.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/io/CombineAvroKeyInputFormat.java
jackjlli: No, the purpose of this class is to create splits across files, so that we can have larger splits and fewer long running mappers, which improves the performance of the Hadoop job.
We need to have a concrete class to extend `CombineFileInputFormat`, that's why we need this class here.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
jackjlli: Good idea. Updated.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mapper/SegmentPreprocessingMapper.java
jackjlli: Done.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/reducer/SegmentPreprocessingReducer.java
jackjlli: `MAXIMUM_NUMBER_OF_RECORDS` is an optional feature. If it isn't specified, we shouldn't truncate the output files. I'll add the documentation here.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/reducer/SegmentPreprocessingReducer.java
jackjlli: The purpose of having a radom prefix is to differentiate the output files generated from different reducers. The name of the output files would be like `tk2n3-r-00002`. So the output file would be listed as follow:
```
tk2n3-r-00001
tk2n3-r-00002
efh20-r-00001
efh20-r-00002
```

If we don't have this and instead have a fixed prefix, all the output files from different reducers will be mixed together, e.g. 
```
prefix1-00001
prefix1-00002
prefix1-00003
prefix2-00001
prefix2-00002
prefix2-00003
```
The reason of using 4 is that by default the name of output files like: `part-r-00001`. Just to have the similar length.

pom.xml
jackjlli: No, this is particularly for avro map reduce job. And there was no MR job before. Previously we just print out segment name as the output of hadoop job.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: We allow both partitioning and sorting. These two doesn't affect each other; they can be enabled/disabled respectively. If anything breaks, this preprocessing job will fail.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: It's best to always turn on this setting, since there could be other dependency jars that is required higher versions.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: By default, hadoop job will write an empty file called `_SUCCESS` to output dir. We actually don't need that empty file.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: If users don't want sort yet, they can still disable the preprocessing job. Plus, it's disabled by default.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: We'll use this to specify the schema of  output files. And IIRC Avro schema and Pinot Schema are not the same thing?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: Yes, we'll use this param to specify the number of reducers if partitioning is disabled. Each reducer will generate at least 1 output file.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: Which section do you refer to? If you refer to checking `_enablePartitioning` and `_enableSorting`, it's not in open source yet. The reason why we need this in this job is that it contains 3 components/features, whereas other existing hadoop job like segment creation job is a single feature.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: This is not only for mv column. In avro, the type can be union even when it's a single value.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: Another one is in segment creation job, which runs after this pre-processing job.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: Checked the code. The scenario here is different since we're only interested in some certain formats.

pinot-hadoop/pom.xml
jenniferdai: do they have conflicting classes at all?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: the whoel section to read parameters is what i meant

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: ok, may be good to document so users know that

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: what is the pinot type that corresponds to that?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: maybe we can put these utility methods somewhere so they can be reused

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: hmmm, i am thinking along the lines of when i want to resize by default, then i will enable preprocessing by default - do you think we should just turn sorting on for everyone? I think it will break, esp because some users think we can sort more than one column

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: can we rename it to make it clearer?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: oh i see what you mean, output files from preprocessing have to be avro

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: what about orc?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/reducer/SegmentPreprocessingReducer.java
jenniferdai: but youve made the default 0, right? won't that break when you do the check later on? it will be "specified"

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/reducer/SegmentPreprocessingReducer.java
jenniferdai: interesting, so each reducer no longer just generates one file, how do we determine that?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: talked offline, makes sense, we decided to validate this separately (perhaps create a ticket or small issue to fail/validate table configs so they only have one partition column setting)

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/reducer/SegmentPreprocessingReducer.java
jackjlli: @jenniferdai If `max.num.records` isn't specified, reducer will generate only 1 output file. If so, reducer will truncate output file into multiple files.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/PinotHadoopJobLauncher.java
jenniferdai: i see, talked offline, in open source we don't technically support this yet because we have to change some parameters to make that work - can we create a small ticket somewher eto add that functionality?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
```diff
@@ -52,4 +52,17 @@
 
   // The path to the record reader to be configured
   public static final String RECORD_READER_PATH = "record.reader.path";
+
+  public static final String ENABLE_PARTITIONING = "enable.partitioning";
+  public static final String ENABLE_SORTING = "enable.sorting";
+  public static final String ENABLE_RESIZING = "enable.resizing";
+
+  public static final String PARTITION_COLUMN = "partition.column";
+  public static final String NUMBER_OF_PARTITIONS = "number.partitions";
```
jenniferdai: num.partitions?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -263,23 +255,7 @@ protected void validateTableConfig(TableConfig tableConfig) {
...
-      if (fileStatus.isDirectory()) {
-        addDepsJarToDistributedCacheHelper(job, fileStatus.getPath());
-      } else {
-        Path depJarPath = fileStatus.getPath();
-        if (depJarPath.getName().endsWith(".jar")) {
-          _logger.info("Adding deps jar: {} to distributed cache", depJarPath);
-          job.addCacheArchive(depJarPath.toUri());
-        }
-      }
+      JobPreparationHelper.addDepsJarToDistributedCacheHelper(_fileSystem, job, _depsJarDir);
```
jenniferdai: nice, i like the job prep helper

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+  private String _partitionFunction;
+  private String _sortedColumn;
+  private int _numberOfOutputFiles;
+
+  private final String _inputSegmentDir;
+  private final String _preprocessedOutputDir;
+  protected final String _rawTableName;
+
+  private final String _pathToDependencyJar;
+  private final String _defaultPermissionsMask;
```
jenniferdai: hmm, do we need to set permissions on the files we create after the pre-processing step?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+      return;
+    } else {
+      _logger.info("Starting {}", getClass().getSimpleName());
+    }
+
+    _fileSystem = FileSystem.get(_configuration);
+    final List<String> inputDataPath = Arrays.asList(_inputSegmentDir.split("\\s*,\\s*"));
+    Path outputDir = new Path(_preprocessedOutputDir);
+
+    if (_fileSystem.exists(outputDir)) {
```
jenniferdai: want to confirm, which output dir is this? is this the same output dir as the pbnj segment creation step? (assuming we set the segment creation step properly?)
or is it just one level in?
perhaps we can use a different default path to output, not sure what our default is - in segment creation, we actually don't require one, we have a default, so we should have a default here too, perhaps /preprocess?

pinot-hadoop/pom.xml
jackjlli: No, we've tried our best to use the compatible method. ;)

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+      _logger.warn("Found the output folder {}, deleting it", _preprocessedOutputDir);
+      _fileSystem.delete(outputDir, true);
+    }
+    JobPreparationHelper.setDirPermission(_fileSystem, outputDir, _defaultPermissionsMask);
+
+    _tableConfig = getTableConfig();
```
jenniferdai: this will fail if we don't have push location set, right? so we should fail in constructor instead of just warn

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+
+    job.setJarByClass(SegmentPreprocessingJob.class);
+
+    String hadoopTokenFileLocation = System.getenv("HADOOP_TOKEN_FILE_LOCATION");
+    if (hadoopTokenFileLocation != null) {
+      job.getConfiguration().set("mapreduce.job.credentials.binary", hadoopTokenFileLocation);
+    }
+
+    // Schema configs.
+    Schema schema = getSchema(inputDataPath.get(0));
```
jenniferdai: are there any utils that can get this info? seems weird that we have to manually get 0th elem

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+    job.setMapperClass(SegmentPreprocessingMapper.class);
+    job.setMapOutputKeyClass(AvroKey.class);
+    job.setMapOutputValueClass(AvroValue.class);
+    job.getConfiguration().setInt(JobContext.NUM_MAPS, inputDataPath.size());
+
+    // Reducer configs.
+    job.setReducerClass(SegmentPreprocessingReducer.class);
+    job.setOutputKeyClass(AvroKey.class);
+    job.setOutputValueClass(NullWritable.class);
+    if (_properties.getProperty(MAXIMUM_NUMBER_OF_RECORDS) != null) {
```
jenniferdai: if this is set and min num files is set, what overrides, do we have error checking to handle that / clear overriding somewherE?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+      }
+    }
+    _logger.info("Total number of files to be pre-computed: {}", inputDataFiles.size());
+
+    // Set up mapper output key
+    Set<Schema.Field> fieldSet = new HashSet<>();
+
+    // Partition configs.
+    int numReduceTasks = (_numberOfPartitions != 0) ? _numberOfPartitions : inputDataFiles.size();
+    if (_partitionColumn != null) {
```
jenniferdai: do we have diff partition fns that we support? Or only the murmur one? IF only murmur, why do we allow this to be configurable?

also, might be nice to do extra error checking here, like make sure partition column and num partitions set and not partition if only one is set, for examplee

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+      job.getConfiguration().set(NUMBER_OF_PARTITIONS, Integer.toString(numReduceTasks));
+    } else {
+      if (_numberOfOutputFiles > 0) {
+        numReduceTasks = _numberOfOutputFiles;
+      }
+      // Partitioning is disabled. Adding hashcode as one of the fields to mapper output key.
+      // so that all the rows can be spread evenly.
+      addHashCodeField(fieldSet);
+    }
+    job.setInputFormatClass(CombineAvroKeyInputFormat.class);
```
jenniferdai: can it happen that files are repartitioned once more when sorted or do you always do 1:1 file mapping after partitioning? Seems not right, because we always use the comnbine avro key format?

also, does min # output files/max records change how files are partitioned? could that break partitioning?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+    job.waitForCompletion(true);
+    if (!job.isSuccessful()) {
+      throw new RuntimeException("Job failed : " + job);
+    }
+
+    _logger.info("Finished pre-computing job in {}ms", (System.currentTimeMillis() - startTime));
+  }
+
+  @Nullable
+  private TableConfig getTableConfig()
```
jenniferdai: we can also put this in some util since it's already in semgnet creation job

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,479 @@
...
+      throws IOException {
+    try (ControllerRestApi controllerRestApi = getControllerRestApi()) {
+      return controllerRestApi != null ? controllerRestApi.getTableConfig() : null;
```
jenniferdai: do we want to allow this to work without table config? we want our params to go in table config, right? what's the motivation for supporting non table config cases?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: Done.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: Any type. Actually this ambiguity comes from avro. 
E.g. the below is the schema of a column called `Carrier`:
```
 {
    "name" : "Carrier",
    "dataType" : "STRING"
  }
```
As you can see, although it shows `STRING` as data type in the table schema, it became `UNION` when we're trying to fetch the schema of that column.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: Done.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: These 3 features won't affect each other; they can be enabled respectively. If you need resizing, just make `enable.resizing` to be true.
And we only allow one sorted column per table. Same as partition column.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: It's optional. In some cases users would just like to run this job and they should be provided a way to access the output files.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: Good point. If output path isn't specified here, we should use a default path. I'll adjust the logic.

jackjlli: Just checked the code. Actually in segment creation job, the output path should not be null. The path modification should be done when initializing the job, not inside the job.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: It will fail if there's no push location or there's no table config.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: In Avro MR job, it requires to explicitly set the schema before running the job. Currently there's no way to get the avro schema beforehand though.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: It's not just murmur. Pinot also supports other functions like modulo, default(hashcode). I'll add the error check.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jackjlli: The config `min # output files` will be **disabled** if partitioning is **enabled**.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,431 @@
...
+   * Finds the avro file in the input folder, and returns its avro schema
+   * @param inputPathDir Path to input directory
+   * @return Input schema
+   * @throws IOException exception when accessing to IO
+   */
+  private static Schema getSchema(Path inputPathDir) throws IOException {
+    FileSystem fs = FileSystem.get(new Configuration());
+    Schema avroSchema = null;
+    for (FileStatus fileStatus : fs.listStatus(inputPathDir)) {
+      if (fileStatus.isFile() && fileStatus.getPath().getName().endsWith(".avro")) {
```
jenniferdai: can you add an item to support orc in the future?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/io/CombineAvroKeyInputFormat.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public RecordReader<AvroKey<T>, NullWritable> createRecordReader(InputSplit split, TaskAttemptContext context)
+      throws IOException {
+    Class cls = AvroKeyRecordReaderWrapper.class;
+
+    return new CombineFileRecordReader<>((CombineFileSplit) split, context,
+        (Class<? extends RecordReader<AvroKey<T>, NullWritable>>) cls);
+  }
+
+  public static class AvroKeyRecordReaderWrapper<T> extends CombineFileRecordReaderWrapper<AvroKey<T>, NullWritable> {
```
jenniferdai: can you check if orc has something similar that i can implement?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,431 @@
...
+      return new DataFileStream<>(fs.open(avroFile), new GenericDatumReader<>());
+    }
+  }
+
+  private void addSortedColumnField(Schema schema, Set<Schema.Field> fieldSet) {
+    // Sorting is enabled. Adding sorted column value to mapper output key.
+    Schema sortedColumnSchema = schema.getField(_sortedColumn).schema();
+    Schema sortedColumnAsKeySchema;
+    if (sortedColumnSchema.getType().equals(Schema.Type.UNION)) {
+      sortedColumnAsKeySchema = Schema.createUnion(sortedColumnSchema.getTypes());
```
jenniferdai: can we make sure people don't sort on mv for now? i think it may not work

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/reducers/SegmentPreprocessingReducer.java
```diff
@@ -0,0 +1,86 @@
...
+      _multipleOutputs = null;
+    }
+    LOGGER.info("Finished cleaning up reducer.");
+  }
+
+  private String generateFileName() {
+    if (_maxNumberOfRecords == 0) {
+      return _filePrefix;
+    } else {
+      return _filePrefix + (_counter.getAndIncrement() / _maxNumberOfRecords);
```
jenniferdai: as discussed offline, i think we should disable giving people the option to do this resizing for now, as worst case, we will end up with many tiny segments and many big segments (potentially one tiny segment for each reducer  * num of pushes people will be doing)

jackjlli: I don't think so. You cannot say it potentially generates too many segments so that it should not be used. At least this is the first phase functionality that users can tune the number of records by themselves. It is users' responsibility to provide a practical config. To enhance this, the next phase can have some ways to mitigate the size of the last file problem.

jenniferdai: Can we hide the parameter and disallow setting it for now? I think it's a pretty big risk for clients because each reducer can have 1 tiny and the rest giant, and if users use this setting over a long period of time, they will have to completely rebootstrap all their data. They will expect that we will do this work intelligently, but we won't have that until the next phase. In prod, this can be very dangerous - users can have downtime and will to rebootstrap their entire dataset to fix this problem

pinot-hadoop/pom.xml
```diff
@@ -117,6 +117,17 @@
       <artifactId>commons-math3</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.avro</groupId>
+      <artifactId>avro-mapred</artifactId>
+      <classifier>hadoop2</classifier>
+      <!--<exclusions>-->
```
jenniferdai: delete?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,451 @@
...
+              JobConfigConstants.PUSH_TO_PORT, pushPortString));
+    }
+
+    _logger.info("*********************************************************************");
+    _logger.info("enable.partitioning: {}", _enablePartitioning);
+    _logger.info("enable.sorting: {}", _enableSorting);
+    _logger.info("enable.resizing: {} {}", _enableResizing,
+        (resizingForcedDisabled ? "(forced to be disabled since partitioning is enabled)" : ""));
+    _logger.info("path.to.input: {}", _inputSegmentDir);
+    _logger.info("path.to.output: {}", _preprocessedOutputDir);
```
jenniferdai: change this to match or use the param so we don't need to

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -0,0 +1,451 @@
...
+  }
+
+  private void addSortedColumnField(Schema schema, Set<Schema.Field> fieldSet) {
+    // Sorting is enabled. Adding sorted column value to mapper output key.
+    Schema sortedColumnSchema = schema.getField(_sortedColumn).schema();
+    Schema sortedColumnAsKeySchema;
+    if (sortedColumnSchema.getType().equals(Schema.Type.UNION)) {
+      sortedColumnAsKeySchema = Schema.createUnion(sortedColumnSchema.getTypes());
+    } else if (sortedColumnSchema.getType().equals(Schema.Type.ARRAY)) {
+      sortedColumnAsKeySchema = Schema.createArray(sortedColumnSchema.getElementType());
```
jenniferdai: can we test this behavior to make sure it works as expected for a mv column?

kishoreg:
Is there a high level write up for this? I am looking for
- user doc - How to use this feature?
- what happens in the hadoop job
- how flexible is this? Can it handle any input format (avro, json, thrift etc). How easy is it to add support for other formats.
-

jenniferdai:
@kishoreg https://github.com/apache/incubator-pinot/issues/4353

Jackie-Jiang:
@jenniferdai Are all the concerns addressed? I thought we will have another discussion once all the design decisions settled.

jenniferdai:
Ive rewritten and posted the design https://github.com/apache/incubator-pinot/issues/4353
and am currently making code changes based on that.

PR #4262, opened 2019-05-30
[TE] add current time series if it's unavailable in predicted baseline endpoint

jihaozh:
In predicted baseline endpoint, add the current time series only when it's unavailable.


xiaohui-sun:
@jihaozh Can you give more details why what will leads to this scenario?

jihaozh:
> @jihaozh Can you give more details why what will leads to this scenario?
The previous PR added padding so that the graphs will have paddings. https://github.com/apache/incubator-pinot/pull/4260

This PR fixes an edge case that if the current time series is already returned from the baseline provider, it will skip adding it.

PR #4269, opened 2019-06-01
Add interface and implementations for the new segment assignment

Jackie-Jiang:
Added InstancePartitions implementations
Added InstancePartitionsUtils to fetch/persist it to ZK and also handle backward-compatibility if it does not exist in ZK
Added SegmentAssignmentStrategy to handle both segment assignment and table rebalance
Implemented the balance-number, replica-group segment assignment strategy for both OFFLINE and REALTIME tables that works on InstancePartitions

There is no change to the persistent state

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
@@ -1,0 +1,99 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.controller.helix.core.assignment;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import org.apache.helix.ZNRecord;
+
+
+/**
+ * Instance (server) partitions for the table.
+ *
+ * <p>The instance partitions is stored as a map from partition of the format: {@code <partitionId>_<replicaId>} to
+ * list of server instances, and is persisted under the ZK path: {@code <cluster>/PROPERTYSTORE/INSTANCE_PARTITIONS}.
+ * <p>The segment assignment will be based on the instance partitions of the table.
+ */
+public class InstancePartitions {
+  private static final char PARTITION_REPLICA_SEPARATOR = '_';
+
+  // Name will be of the format "<rawTableName>_<type>", e.g. "table_OFFLINE", "table_CONSUMING", "table_COMPLETED"
+  private final String _name;
+  private final Map<String, List<String>> _partitionToInstancesMap;
+  private int _numPartitions;
+  private int _numReplicas;
+
+  public InstancePartitions(String name) {
+    _name = name;
+    _partitionToInstancesMap = new TreeMap<>();
+  }
+
+  private InstancePartitions(String name, Map<String, List<String>> partitionToInstancesMap) {
+    _name = name;
+    _partitionToInstancesMap = partitionToInstancesMap;
+    for (String key : partitionToInstancesMap.keySet()) {
+      int splitterIndex = key.indexOf(PARTITION_REPLICA_SEPARATOR);
+      int partition = Integer.parseInt(key.substring(0, splitterIndex));
+      int replica = Integer.parseInt(key.substring(splitterIndex + 1));
+      _numPartitions = Integer.max(_numPartitions, partition + 1);
+      _numReplicas = Integer.max(_numReplicas, replica + 1);
+    }
+  }
+
+  public String getName() {
+    return _name;
+  }
+
+  public Map<String, List<String>> getPartitionToInstancesMap() {
+    return _partitionToInstancesMap;
+  }
+
+  @JsonIgnore
+  public int getNumPartitions() {
+    return _numPartitions;
+  }
+
+  @JsonIgnore
+  public int getNumReplicas() {
+    return _numReplicas;
+  }
+
+  public List<String> getInstances(int partitionId, int replicaId) {
+    return _partitionToInstancesMap.get(Integer.toString(partitionId) + PARTITION_REPLICA_SEPARATOR + replicaId);
+  }
+
+  public void setInstances(int partitionId, int replicaId, List<String> instances) {
+    String key = Integer.toString(partitionId) + PARTITION_REPLICA_SEPARATOR + replicaId;
+    _partitionToInstancesMap.put(key, instances);
+    _numPartitions = Integer.max(_numPartitions, partitionId + 1);
+    _numReplicas = Integer.max(_numReplicas, replicaId + 1);
+  }
+
+  public static InstancePartitions fromZNRecord(ZNRecord znRecord) {
+    return new InstancePartitions(znRecord.getId(), znRecord.getListFields());
+  }
+
+  public ZNRecord toZNRecord() {
+    ZNRecord znRecord = new ZNRecord(_name);
+    znRecord.setListFields(_partitionToInstancesMap);
+    return znRecord;
+  }
+}
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitionsUtils.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitionsUtils.java
@@ -1,0 +1,97 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.controller.helix.core.assignment;
+
+import java.util.Collections;
+import java.util.List;
+import org.apache.helix.AccessOption;
+import org.apache.helix.HelixManager;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.model.InstanceConfig;
+import org.apache.helix.store.HelixPropertyStore;
+import org.apache.pinot.common.config.OfflineTagConfig;
+import org.apache.pinot.common.config.RealtimeTagConfig;
+import org.apache.pinot.common.config.TableConfig;
+import org.apache.pinot.common.config.TableNameBuilder;
+import org.apache.pinot.common.metadata.ZKMetadataProvider;
+import org.apache.pinot.common.utils.InstancePartitionsType;
+import org.apache.pinot.common.utils.helix.HelixHelper;
+
+
+/**
+ * Utility class for instance partitions.
+ */
+public class InstancePartitionsUtils {
+  private InstancePartitionsUtils() {
+  }
+
+  /**
+   * Fetches the instance partitions from Helix property store if exists, or computes it for backward compatibility.
+   */
+  public static InstancePartitions fetchOrComputeInstancePartitions(HelixManager helixManager, TableConfig tableConfig,
+      InstancePartitionsType instancePartitionsType) {
+    String tableNameWithType = tableConfig.getTableName();
+    String instancePartitionsName =
+        instancePartitionsType.getInstancePartitionsName(TableNameBuilder.extractRawTableName(tableNameWithType));
+
+    // Fetch the instance partitions from property store if exists
+    String path = ZKMetadataProvider.constructPropertyStorePathForInstancePartitions(instancePartitionsName);
+    ZNRecord znRecord = helixManager.getHelixPropertyStore().get(path, null, AccessOption.PERSISTENT);
+    if (znRecord != null) {
+      return InstancePartitions.fromZNRecord(znRecord);
+    }
+
+    // Compute the instance partitions (for backward compatible)
+    // Sort all enabled instances with the server tag, rotate the list based on the table name to prevent creating
+    // hotspot servers
+    InstancePartitions instancePartitions = new InstancePartitions(instancePartitionsName);
+    List<InstanceConfig> instanceConfigs = HelixHelper.getInstanceConfigs(helixManager);
+    String serverTag;
+    switch (instancePartitionsType) {
+      case OFFLINE:
+        serverTag = new OfflineTagConfig(tableConfig).getOfflineServerTag();
+        break;
+      case CONSUMING:
+        serverTag = new RealtimeTagConfig(tableConfig).getConsumingServerTag();
+        break;
+      case COMPLETED:
+        serverTag = new RealtimeTagConfig(tableConfig).getCompletedServerTag();
+        break;
+      default:
+        throw new IllegalArgumentException();
+    }
+    List<String> instances = HelixHelper.getEnabledInstancesWithTag(instanceConfigs, serverTag);
+    instances.sort(null);
+    int numInstances = instances.size();
+    Collections.rotate(instances, -(Math.abs(tableNameWithType.hashCode()) % numInstances));
+    instancePartitions.setInstances(0, 0, instances);
+    return instancePartitions;
+  }
+
+  /**
+   * Persists the instance partitions to Helix property store.
+   *
+   * @return true if the instance partitions was successfully persisted, false otherwise
+   */
+  public static boolean persistInstancePartitions(HelixPropertyStore<ZNRecord> propertyStore,
+      InstancePartitions instancePartitions) {
+    String path = ZKMetadataProvider.constructPropertyStorePathForInstancePartitions(instancePartitions.getName());
+    return propertyStore.set(path, instancePartitions.toZNRecord(), AccessOption.PERSISTENT);
+  }
+}
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentStrategy.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentStrategy.java
@@ -1,0 +1,59 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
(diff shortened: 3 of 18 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitionsUtils.java
```diff
@@ -0,0 +1,84 @@
...
+    // hotspot servers
+    InstancePartitions instancePartitions = new InstancePartitions(tableNameWithType);
+    List<InstanceConfig> instanceConfigs = HelixHelper.getInstanceConfigs(helixManager);
+    String serverTag =
+        tableConfig.getTableType() == CommonConstants.Helix.TableType.OFFLINE ? new OfflineTagConfig(tableConfig)
+            .getOfflineServerTag() : new RealtimeTagConfig(tableConfig).getConsumingServerTag();
+    List<String> instances = HelixHelper.getEnabledInstancesWithTag(instanceConfigs, serverTag);
+    instances.sort(null);
+    int numInstances = instances.size();
+    Collections.rotate(instances, -(Math.abs(tableNameWithType.hashCode()) % numInstances));
```
snleee: Does the direction for rotation matter? (negative vs positive)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
```diff
@@ -0,0 +1,95 @@
...
+ * <p>The instance partitions are stored as a map from partition of the format: {@code <partitionId>_<replicaId>} to
+ * list of server instances.
+ */
+public class InstancePartitions {
+  private static final char SPLITTER = '_';
+
+  private final String _tableNameWithType;
+  private final Map<String, List<String>> _partitionToInstancesMap;
+  private int _numPartitions;
+  private int _numReplicas;
```
snleee: I guess that `numReplicas` for `InstancePartitions` is equivalent to `the number of groups in instance partition`. In that case, we may need to consider changing variable name for`numReplicas` to `numReplicaGroups`?  

In most cases, `numReplica` here is equal to `replication` factor; however, we use `numReplica=1` for supporting backward compatibility. In this case, `numReplica` and `replication` will mean different numbers. (`numReplica=1` while `replication can be 3`)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/RealtimeSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,155 @@
...
+    LOGGER.info(
+        "Rebalanced {} segments with {} replicas and instance partitions: {} for table: {}, number of segments to be moved to each instance: {}",
+        currentAssignment.size(), numReplicas, instancePartitions, _tableNameWithType,
+        SegmentAssignmentUtils.getNumSegmentsToBeMoved(currentAssignment, newAssignment));
+    return newAssignment;
+  }
+
+  private int getNumReplicas(InstancePartitions instancePartitions) {
+    int numReplicas = instancePartitions.getNumReplicas();
+    if (numReplicas == 1) {
```
snleee: This code is relevant to the above comment on variable name

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/ReplicaGroupSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,270 @@
...
+    int instanceIdWithLeastSegmentsAssigned = 0;
+    int numInstances = numSegmentsAssigned.length;
+    for (int instanceId = 1; instanceId < numInstances; instanceId++) {
+      if (numSegmentsAssigned[instanceId] < minNumSegmentsAssigned) {
+        minNumSegmentsAssigned = numSegmentsAssigned[instanceId];
+        instanceIdWithLeastSegmentsAssigned = instanceId;
+      }
+    }
+
+    // Mirror the assignment to all replicas
```
snleee: Are we planning to remove config for `mirror` and make it by default? +1 for removing config

The downside of this will be that if we ever want to support different # of servers for each replica group, we will not be able to do mirroring anymore. This code will add the constraint on that scenario while no-mirroring will naturally support that.

Currently, I cannot think of any use case that needs the relaxation. You can also double check on this.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitionsUtils.java
Jackie-Jiang: Yes, backward compatible with the current realtime behavior (UniformStreamPartitionAssignmentStrategy.class).

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
Jackie-Jiang: This `numReplicas` is for instance partitions (when we index it with partitionId and replicaId, replicaId must be smaller than this numReplicas).
You are correct that the instance partitions **numReplicas** might not match the table "replication", but it only represent the number of replicas inside instance partitions. Inside instance partitions, there is no concept of replica group, so I would prefer keep the variable name `numReplicas`

kishoreg:
Nice to see this. Give me couple of days to review.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/ReplicaGroupSegmentAssignmentStrategy.java
Jackie-Jiang: Yes. We are going to always use mirror. I cannot think of any use case which needs different number of servers for each replica group, and we don't support that as well. If we actually need that, we can add the support later.

Jackie-Jiang:
@kishoreg There will be a design doc for this and upcoming prs coming soon.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/BalanceNumSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,121 @@
...
+    return newAssignment;
+  }
+
+  private List<String> getInstances(InstancePartitions instancePartitions) {
+    Preconditions.checkArgument(instancePartitions.getNumPartitions() == 1,
+        "The instance partitions should contain only 1 partition");
+    Preconditions.checkArgument(instancePartitions.getNumReplicas() == 1,
+        "The instance partitions should contain only 1 replica");
+    List<String> instances = instancePartitions.getInstances(0, 0);
+    Preconditions.checkState(instances.size() >= _replication,
```
mcvsubbu: it will be useful to indicate how we arrived at less number of instances. Since instance assignment is now multiple steps, just returning the end result will leave an administrator confused as to what they did wrong.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public interface SegmentAssignmentStrategy {
+
+  /**
+   * Initializes the segment assignment strategy.
+   */
+  void init(ZkHelixPropertyStore<ZNRecord> propertyStore, TableConfig tableConfig);
+
+  /**
+   * Assigns a new segment.
```
mcvsubbu: Please document the arguments. What is the key and values of 'currentAssignment' map?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,54 @@
...
+   */
+  List<String> assignSegment(String segmentName, Map<String, Map<String, String>> currentAssignment,
+      InstancePartitions instancePartitions);
+
+  /**
+   * Rebalances the segments for a table.
+   *
+   * @return the rebalanced assignment for the segments
+   */
+  Map<String, Map<String, String>> rebalanceTable(Map<String, Map<String, String>> currentAssignment,
```
mcvsubbu: please doc the arguments

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/RealtimeSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,155 @@
...
+    for (int replicaId = 0; replicaId < numReplicas; replicaId++) {
+      instancesAssigned.add(getInstance(instancePartitions, partitionId, replicaId));
+    }
+    LOGGER.info("Assigned segment: {} with partition id: {} to instances: {} for table: {}", segmentName, partitionId,
+        instancesAssigned, _tableNameWithType);
+    return instancesAssigned;
+  }
+
+  @Override
+  public Map<String, Map<String, String>> rebalanceTable(Map<String, Map<String, String>> currentAssignment,
```
mcvsubbu: should include a flag whether we need to rebalance consuming segments or not?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import org.apache.helix.ZNRecord;
+
+
+/**
+ * Instance (server) partitions for the table.
```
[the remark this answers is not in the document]
Jackie-Jiang: All the segment assignment (both realtime and offline) will be handled with the new added classes. Once we plug them in, will clean up the existing ones.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
```diff
@@ -0,0 +1,97 @@
...
+    _partitionToInstancesMap.put(key, instances);
+    _numPartitions = Integer.max(_numPartitions, partitionId + 1);
+    _numReplicas = Integer.max(_numReplicas, replicaId + 1);
+  }
+
+  public static InstancePartitions fromZNRecord(ZNRecord znRecord) {
+    return new InstancePartitions(znRecord.getId(), znRecord.getListFields());
+  }
+
+  public ZNRecord toZNRecord() {
```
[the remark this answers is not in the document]
Jackie-Jiang: Added comments in the class javadoc

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/BalanceNumSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,121 @@
...
+    _tableNameWithType = tableConfig.getTableName();
+    _replication = tableConfig.getValidationConfig().getReplicationNumber();
+
+    LOGGER
+        .info("Initialized BalanceNumSegmentAssignmentStrategy for table: {} with replication: {}", _tableNameWithType,
+            _replication);
+  }
+
+  @Override
+  public List<String> assignSegment(String segmentName, Map<String, Map<String, String>> currentAssignment,
```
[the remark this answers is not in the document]
Jackie-Jiang: Added. It a map from segment name to instance state map

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/BalanceNumSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,121 @@
...
+      instancesAssigned.add(instances.get(heap.remove().getRight()));
+    }
+
+    LOGGER.info("Assigned segment: {} to instances: {} for table: {}", segmentName, instancesAssigned,
+        _tableNameWithType);
+    return instancesAssigned;
+  }
+
+  @Override
+  public Map<String, Map<String, String>> rebalanceTable(Map<String, Map<String, String>> currentAssignment,
```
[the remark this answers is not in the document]
Jackie-Jiang: Correct. Added comments

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/RealtimeSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.helix.store.zk.ZkHelixPropertyStore;
+import org.apache.pinot.common.config.TableConfig;
+import org.apache.pinot.common.utils.LLCSegmentName;
+import org.apache.pinot.controller.helix.core.assignment.InstancePartitions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Segment assignment strategy for LLC real-time segments.
```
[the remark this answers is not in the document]
Jackie-Jiang: Both. Added

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/RealtimeSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,155 @@
...
+ * <p>It is very similar to replica-group based segment assignment with the following differences:
+ * <ul>
+ *   <li>1. Each partition always contains one server</li>
```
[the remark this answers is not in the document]
Jackie-Jiang: Yes. Added

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/RealtimeSegmentAssignmentStrategy.java
```diff
@@ -0,0 +1,155 @@
...
+public class RealtimeSegmentAssignmentStrategy implements SegmentAssignmentStrategy {
+  private static final Logger LOGGER = LoggerFactory.getLogger(RealtimeSegmentAssignmentStrategy.class);
+
+  private String _tableNameWithType;
+  private int _replication;
+
+  @Override
+  public void init(ZkHelixPropertyStore<ZNRecord> propertyStore, TableConfig tableConfig) {
+    _tableNameWithType = tableConfig.getTableName();
+    _replication = tableConfig.getValidationConfig().getReplicasPerPartitionNumber();
```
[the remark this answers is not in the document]
Jackie-Jiang: That is correct. We don't allow assign/rebalance HLC segments, which will throw exception.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentStrategy.java
Jackie-Jiang: Added

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/RealtimeSegmentAssignmentStrategy.java
Jackie-Jiang: Discussed with @npawar. In order to benefit from replica-group routing (query less servers), we need to always rebalance consuming segments if partition for an instance changed (should be rare).

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentStrategy.java
Jackie-Jiang: Added

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/BalanceNumSegmentAssignmentStrategy.java
Jackie-Jiang: That will be handled in the instance assignment step (in the upcoming pr). This check is for cases where people modify table replication without re-calculate instance partitions.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
npawar: I agree with snlee. It is confusing to call this numReplicas. numReplicas already has a well known implied meaning, and this `numInstanceGroups`, or `numInstancePartitions` shouldn't be called numReplicas

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
```diff
@@ -0,0 +1,98 @@
...
+ * <p>The instance partitions is stored as a map from partition of the format: {@code <partitionId>_<replicaId>} to
+ * list of server instances, and is persisted under the ZK path: {@code <cluster>/PROPERTYSTORE/INSTANCE_PARTITIONS}.
+ * <p>The segment assignment will be based on the instance partitions of the table.
+ */
+public class InstancePartitions {
+  private static final char SPLITTER = '_';
+
+  private final String _tableNameWithType;
+  private final Map<String, List<String>> _partitionToInstancesMap;
+  private int _numPartitions;
```
npawar: numPartitions will always be 0 here right? Why do we need to maintain it in this class?

sunithabeeram:
Could you please call out in the description section the changes to persistent state (if any) made through this PR? What are the implications for cases where code is rolled-back with the new state?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
Jackie-Jiang: @npawar this `numReplicas` has the same meaning as the implied meaning. The only difference is that we use `numReplicas` **1** for backward compatibility.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
Jackie-Jiang: We support partitioning for replicas. Within a replica, each partition can have a subset of servers for the replica.

Jackie-Jiang:
> Could you please call out in the description section the changes to persistent state (if any) made through this PR? What are the implications for cases where code is rolled-back with the new state?

There is no persistent state change. Added in the commit message

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentUtils.java
```diff
@@ -0,0 +1,274 @@
...
+      InstancePartitions instancePartitions, int partitionId, Set<String> segments,
+      Map<String, Map<String, String>> newAssignment) {
+    // Fetch instances in replica 0
+    List<String> instances = instancePartitions.getInstances(partitionId, 0);
+    Map<String, Integer> instanceNameToIdMap = SegmentAssignmentUtils.getInstanceNameToIdMap(instances);
+
+    // Calculate target number of segments per instance
+    int numInstances = instances.size();
+    int numSegments = segments.size();
+    int targetNumSegmentsPerInstance = (numSegments + numInstances - 1) / numInstances;
```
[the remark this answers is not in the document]
Jackie-Jiang: This is equivalent to Math.ceil() for integer. E.g. for 7 segments, 3 servers, we will target 3 segments per server. Will add comment

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentStrategyFactory.java
```diff
@@ -0,0 +1,57 @@
...
+      if (tableConfig.getTableType() == TableType.OFFLINE) {
+        segmentAssignmentStrategy = new OfflineReplicaGroupSegmentAssignmentStrategy();
+      } else {
+        segmentAssignmentStrategy = new RealtimeReplicaGroupSegmentAssignmentStrategy();
+      }
+    } else {
+      if (tableConfig.getTableType() == TableType.OFFLINE) {
+        segmentAssignmentStrategy = new OfflineBalanceNumSegmentAssignmentStrategy();
+      } else {
+        segmentAssignmentStrategy = new RealtimeBalanceNumSegmentAssignmentStrategy();
```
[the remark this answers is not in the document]
Jackie-Jiang: For HLC you never need to assign segment. Everything is handled on the server side. If HLC table calls the segment assignment, it will throw exception.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignmentUtils.java
```diff
@@ -0,0 +1,274 @@
...
+ * Utility class for segment assignment.
+ */
+class SegmentAssignmentUtils {
+  private SegmentAssignmentUtils() {
+  }
+
+  /**
+   * Returns the number of segments assigned to each instance.
+   */
+  static int[] getNumSegmentsAssigned(Map<String, Map<String, String>> segmentAssignment, List<String> instances) {
```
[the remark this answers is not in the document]
Jackie-Jiang: Will rename

PR #4270, opened 2019-06-01
skip bit-map inverted index generation for sorted column

jasperjiaguo:
It is redundant for a sorted column to have a bitmap index. Skipping this here won't have side effect for meta data generation.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -159,11 +159,8 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
 
         // Initialize inverted index creator
         // Do not created inverted index if sorted, clean up invertedIndexColumns
-        if (invertedIndexColumns.contains(columnName)) {
-          if (indexCreationInfo.isSorted()){
-            invertedIndexColumns.remove(columnName);
-            //config.getInvertedIndexCreationColumns().remove(columnName); Do we clean up config?
-          } else if (segmentCreationSpec.isOnHeap()) {
+        if (invertedIndexColumns.contains(columnName) && !indexCreationInfo.isSorted()) {
+          if (segmentCreationSpec.isOnHeap()) {
             _invertedIndexCreatorMap
                 .put(columnName, new OnHeapBitmapInvertedIndexCreator(_indexDir, columnName, cardinality));
           } else {
--- a/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -158,8 +158,12 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
         }
 
         // Initialize inverted index creator
+        // Do not created inverted index if sorted, clean up invertedIndexColumns
         if (invertedIndexColumns.contains(columnName)) {
-          if (segmentCreationSpec.isOnHeap()) {
+          if (indexCreationInfo.isSorted()){
+            invertedIndexColumns.remove(columnName);
+            //config.getInvertedIndexCreationColumns().remove(columnName); Do we clean up config?
+          } else if (segmentCreationSpec.isOnHeap()) {
             _invertedIndexCreatorMap
                 .put(columnName, new OnHeapBitmapInvertedIndexCreator(_indexDir, columnName, cardinality));
           } else {
--- a/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -158,7 +158,7 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
         }
 
         // Initialize inverted index creator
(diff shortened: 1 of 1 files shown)


sunithabeeram:
@jasperjiaguo can you please note a few more details in the description section. What is the motivation/observation that lead to this change? And what testing was done to ensure that things worked correctly?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -158,8 +158,12 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
         }
 
         // Initialize inverted index creator
+        // Do not created inverted index if sorted, clean up invertedIndexColumns
         if (invertedIndexColumns.contains(columnName)) {
-          if (segmentCreationSpec.isOnHeap()) {
+          if (indexCreationInfo.isSorted()){
+            invertedIndexColumns.remove(columnName);
+            //config.getInvertedIndexCreationColumns().remove(columnName); Do we clean up config?
```
sunithabeeram: The way the execution code is currently structured, for us to use SortedInvertedIndexBasedFilterOperator, we still need the hint for inverted-index hint - so I don't think you can remove it from config. Please trace the code and ensure that that's not broken. Also make sure there are tests covering the execution for sorted columns.

BaseSingleValueQueriesTest.java is a comprehensive test suite that should cover this, but it looks like we are not covering the case there where a column is both sorted and has invertedIndex. You can extend that test to cover this case and ensure things are working as expected. You will most likely also need a test to assert that the bitmap based index isn't generated.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -158,8 +158,12 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
         }
 
         // Initialize inverted index creator
+        // Do not created inverted index if sorted, clean up invertedIndexColumns
         if (invertedIndexColumns.contains(columnName)) {
```
Jackie-Jiang: Change this line to `if (invertedIndexColumns.contains(columnName) && !indexCreationInfo.isSorted())` should be good. We don't need to clean up the invertedIndexColumns (it won't be populated outside even if you changed it).

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
Jackie-Jiang: BaseSingleValueQueriesTest is for query execution. If you want to test this, you can add a separate segment generation test.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -158,7 +158,8 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
         }
 
         // Initialize inverted index creator
-        if (invertedIndexColumns.contains(columnName)) {
+        // Do not created inverted index if sorted, clean up invertedIndexColumns
```
Jackie-Jiang: Please update the comment accordingly

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -158,7 +158,8 @@ public void init(SegmentGeneratorConfig segmentCreationSpec, SegmentIndexCreatio
         }
 
         // Initialize inverted index creator
-        if (invertedIndexColumns.contains(columnName)) {
+        // Skip creating inverted index if sorted; _invertedIndexCreatorMap is also correctly populated for metadata writing
```
Jackie-Jiang: I don't quite follow the second part of the comment. "Skip creating inverted index if sorted" is good enough IMO.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
sunithabeeram: @Jackie-Jiang, I was a requesting at test that covers the execution validation of sorted-column + inverted index as we don't seem to have any coverage for that currently.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
jasperjiaguo: Standalone test was performed with pinot-admin commands. No inverted index was generated for sorted columns when I asked the system to do so with generatorConfigFile option.

jasperjiaguo: Can you pls merge it? Thx.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
jasperjiaguo: I tested the sorted-column + inverted index case and it did skip the inverted index generation. The HAS_INVERTED_INDEX flag is also good.

sunithabeeram:
@jasperjiaguo thanks for testing this out. It will be great if you can add test coverage for sorted-inverted-index by updating BaseSingleValueQueriesTest (and related tests as needed).

PR #4272, opened 2019-06-03
[TE] frontend - harleyjj/alert-details - handle all dimensions case in alert-details component

harleyjj:
1) Handle 'All Dimensions' case in Alert Details component, where there are dimensions but some anomalies are for all dimensions.
2) Also fixes a bug where the detector names were not being converted into a set correctly.


sunithabeeram:
Could we please stick to the convention of including "[TE]" in the commit messages and PR descriptions? This will help us skip TE specific commits when we look up the changelog. Thanks!

PR #4276, opened 2019-06-04
Avoid negative sleep time in startupServiceStatusCheck

npawar:
We should not sleep if sleepTime <= 0 in startupServiceStatusCheck.


Jackie-Jiang:
Good catch. Sorry I misused the API... Thought it will return immediately if passing in negative sleep time (guava Uninterruptibles behavior). There are other place needs to be fixed. Will file a pr for them.

Jackie-Jiang:
#4277

PR #4279, opened 2019-06-04
Revert "Avoid negative sleep time in startupServiceStatusCheck"

npawar:
Reverts apache/incubator-pinot#4276


mcvsubbu:
In favor of #4277 having the complete fix. @Jackie-Jiang  please merge #4277

Jackie-Jiang:
Already rebased the change, no need to revert the previous one.

PR #4283, opened 2019-06-05
Reformat classes in pinot-common

Jackie-Jiang:
Includes thrift auto-generated files.

Auto-generated files from different Thrift version have different format.
Keep thrift files following Pinot Style can minimize the change when re-compiling using different Thrift version.


snleee:
This pr is a good example for the issue raised in https://github.com/apache/incubator-pinot/issues/4264

PR #4285, opened 2019-06-06
Fix the potential overflow for numGroupsLimit in CombineGroupByOperator

Jackie-Jiang:


s-sahay:
@Jackie-Jiang : Can we add a test that would catch this kind of issue? Doing it for all config values may be a stretch but can we add one for this which can test the code path with config values set to min/max?

Jackie-Jiang:
@s-sahay We can put these extreme configs into an integration test. While it is hard to prevent these kinds of issues for new added configs.

PR #4286, opened 2019-06-06
[TE] Translator code refactor to accommodate entity config translation

akshayrai:
Changes:

1. Translator should be like a black box - takes raw config and returns detection object. Update the code to make translator accept raw yaml config as input rather than a map.   
    - Removes duplicate logic in parsing of raw config to map
    - Eliminates the need to inject yaml config into detection object post translation.
2. Merge Composite pipeline translator into a single DetectionConfigTranslator. With assumption that we do not foresee any other types of translator. 
3. Remove `pipelineType` and Translator Loader in favor of a single Detection Translator.
4. Renamed `YamlDetectionConfigTranslator` to `DetectionConfigTranslator`
5. Renamed `YamlDetectionAlertConfigTranslator` to `SubscriptionConfigTranslator`


sunithabeeram:
Could we please stick to the convention of including "[TE]" in the commit messages and PR descriptions? This will help us skip TE specific commits when we look up the changelog. Thanks!

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/ConfigTranslator.java
```diff
@@ -19,25 +19,30 @@
...
  */
 public abstract class ConfigTranslator<T extends AbstractDTO, V extends ConfigValidator> {
 
-  Map<String, Object> yamlConfig;
-
   protected ConfigValidator validator;
   protected DataProvider dataProvider;
+  protected String yamlConfig;
+  protected Map<String, Object> yamlConfigMap;
+  protected Yaml yaml;
```
jihaozh: could you make these instance variables final when possible?

akshayrai: Done. I have also updated the interface making `translateConfig` accept the parsed config map as a parameter.

PR #4290, opened 2019-06-07
Prepare licenses for 0.2

mcvsubbu:
Updating LICENSE-binary in preparation for release 0.2.0

Ran the command 
`mvn -Ddependency.locations.enabled=false project-info-reports:dependencies`

and found the list of dependencies (including transitive ones) in 
`pinot-distribution/target/site/dependencies.html`

Used the list of dependencies found in updating the license files


mcvsubbu:
Best to review this one commit at a time.
We now have version numbers along with the package names, so that we can easily tell later on whether we need to look up the license for a specific package simply because the version number changed.
The first commit adds version numbers to the existing license file.
The second commit removes pinot-common and pinot-core jars from the components list under apache license.
The third commit makes the modifications to the license file from the time we released 0.1.0 until now.

An easy way of doing this is to cut/paste the contents of page from the browser onto a text file, and use awk scripts to get the package and version list. Assuming you cut/paste the contents of the html file in licenses.txt

You can then run script like this:
`
cat licenses.txt | awk '{printf("%s:%s:%s\n'", $1, $2, $3)}' | sort > pkg-versions
`
Now the file `pkg-versions` has all the packages we use, with the group id and version number.

licenses-binary/LICENSE-BSD-3-clause.txt
```diff
@@ -0,0 +1,27 @@
+
+Copyright (c) 2000-2011 INRIA, France Telecom
```
snleee: I put one license txt file for each library for BSD-3 because each library has different copyright. Please refer `LICENSE-jcabi.txt, LICENSE-paranamer.txt..etc`. Please create one file for each library.

LICENSE-binary
```diff
@@ -204,74 +204,104 @@
...
-org.roaringbitmap:RoaringBitmap
-org.webjars:swagger-ui
-org.xerial:xerial-core
-org.xerial.java:xerial-core
-org.xerial.larray:larray
-org.xerial.larray:larray-buffer
-org.xerial.larray:larray-mmap
-org.xerial.snappy:snappy-java
-org.yaml:snakeyaml
+
```
snleee: Remove extra lines

LICENSE-binary
```diff
@@ -323,57 +359,58 @@ Common Development and Distribution License (CDDL) 1.0
...
 
 Eclipse Public License (EPL) 1.0
 --------------------------------
 (see license/LICENSE-epl-1.0.txt)
 
-org.aspectj:aspectjrt - https://www.eclipse.org/aspectj
+org.aspectj:aspectjrt:1.8.4 - https://www.eclipse.org/aspectj
 
 
 Eclipse Public LIcense (EPL) 2.0
```
snleee: Can you rename `LIcense -> License`?

licenses-binary/LICENSE-public-domain.txt
```diff
@@ -0,0 +1,4 @@
+Text copied verbatim from http://aopalliance.sourceforge.net/
```
snleee: Which library is this for? `org.glassfish.hk2.external:aopalliance-repackaged` seems to be under EPL.

licenses-binary/LICENSE-BSD-3-clause.txt
mcvsubbu: They are all the same license, with the last clause either specifying the actual provider's name, or using a general term. Did you see a different one? I will take a look again.

licenses-binary/LICENSE-aopalliance.txt
```diff
@@ -0,0 +1,4 @@
+Text copied verbatim from http://aopalliance.sourceforge.net/
```
snleee: @felixcheung Is this good enough for `public domain` license?

snleee:
Can we fix the commit hash to whatever this pr is based and create `0.2.0` branch out of that? Otherwise, we may need to repeat this work if someone checks in any library change :)

mcvsubbu:
We cannot do that now, since we are planning to upgrade helix to  newer release anyway.
The idea of this exercise was to go through the process once and document it (since the earlier process was not documented.

The change after we add the new helix release should be small, so we should be good to go at that point.

licenses-binary/LICENSE-aopalliance.txt
felixcheung: https://apache.org/legal/resolved.html#handling-public-domain-licensed-works

do you see a reference to a license?

mcvsubbu: There is just that one line there in the page http://aopalliance.sourceforge.net/, with no link to the license clauses. There is a list of members right below the line that mentions LICENSE

mcvsubbu:
@felixcheung do you have any other comments?
Can we push this as a starter for 0.2? We do have at least one change coming to this one (we plan to integrate with a newer version of helix), so there will be one more round of update to these files as necessary.

snleee:
@mcvsubbu I found that both Druid and Spark package aopalliance. 
https://github.com/apache/incubator-druid/blob/master/LICENSE.BINARY#L1256
https://github.com/apache/spark/blob/master/LICENSE-binary#L510

They added it to `LICENSE-binary` but they did not add the license file. I think that we can do the following:

1. don't add `licenses-binary/LICENSE-aopalliance.txt`
2. or keep as it is


@felixcheung Are you fine with this?

felixcheung:
I think so.

this is a good reference: https://github.com/apache/spark/commit/f825847c82042a9eee7bd5cfab106310d279fc32#diff-d9821b8f126666b09f58dad60a32e546

licenses-binary/LICENSE-aopalliance.txt
snleee: @mcvsubbu I think that we can remove this line and just keep the following line to this file.

```
LICENCE: all the source code provided by AOP Alliance is Public Domain.
```

snleee:
@felixcheung Thanks for the confirmation 👍 

@mcvsubbu From the spark reference, it says that 
```
Public domain and similar works are treated like permissively licensed dependencies.
```

PR #4293, opened 2019-06-08
Add support for an aggregation function returning serialized hyperlog…

ianvkoeppe:
…logs.

Client-side batching is a common requirement for reducing the query and
response sizes to/from Pinot. In doing so, client's may then need to perform
their own aggregation of results from the multiple responses returned. This
is easy for sum, min, max, etc., but numerical values representing uniques cannot
simply be added.

Supporting the DISTINCTCOUNTRAWHLL and DISTINCTCOUNTRAWHLLMV will allow
clients to receive the serialized HLL and then aggregate them
client-side.

### DIFF
--- a/docs/pql_examples.rst
+++ b/docs/pql_examples.rst
@@ -148,6 +148,7 @@ Supported aggregations on single-value columns
 * ``MINMAXRANGE``
 * ``DISTINCTCOUNT``
 * ``DISTINCTCOUNTHLL``
+* ``DISTINCTCOUNTRAWHLL``: Returns HLL response serialized as string. The serialized HLL can be converted back into an HLL (see `pinot-core/\*\*/HllUtil.java` as an example) and then aggregated with other HLLs. A common use case may be to merge HLL responses from different Pinot tables, or to allow aggregation after client-side batching.
 * ``FASTHLL`` (**WARN**: will be deprecated soon. ``FASTHLL`` stores serialized HyperLogLog in String format, which performs
   worse than ``DISTINCTCOUNTHLL``, which supports serialized HyperLogLog in BYTES (byte array) format)
 * ``PERCENTILE[0-100]``: e.g. ``PERCENTILE5``, ``PERCENTILE50``, ``PERCENTILE99``, etc.
@@ -164,6 +165,7 @@ Supported aggregations on multi-value columns
 * ``MINMAXRANGEMV``
 * ``DISTINCTCOUNTMV``
 * ``DISTINCTCOUNTHLLMV``
+* ``DISTINCTCOUNTRAWHLLMV``: Returns HLL response serialized as string. The serialized HLL can be converted back into an HLL (see `pinot-core/**/HllUtil.java` as an example) and then aggregated with other HLLs. A common use case may be to merge HLL responses from different Pinot tables, or to allow aggregation after client-side batching.
 * ``FASTHLLMV`` (**WARN**: will be deprecated soon. It does not make lots of sense to configure serialized HyperLogLog
   column as a dimension)
 * ``PERCENTILE[0-100]MV``: e.g. ``PERCENTILE5MV``, ``PERCENTILE50MV``, ``PERCENTILE99MV``, etc.
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/aggregator/ValueAggregatorFactory.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/aggregator/ValueAggregatorFactory.java
@@ -50,6 +50,7 @@ public class ValueAggregatorFactory {
       case MINMAXRANGE:
         return new MinMaxRangeValueAggregator();
       case DISTINCTCOUNTHLL:
+      case DISTINCTCOUNTRAWHLL:
         return new DistinctCountHLLValueAggregator();
       case PERCENTILEEST:
         return new PercentileEstValueAggregator();
@@ -81,6 +82,7 @@ public class ValueAggregatorFactory {
       case MINMAXRANGE:
         return MinMaxRangeValueAggregator.AGGREGATED_VALUE_TYPE;
       case DISTINCTCOUNTHLL:
+      case DISTINCTCOUNTRAWHLL:
         return DistinctCountHLLValueAggregator.AGGREGATED_VALUE_TYPE;
       case PERCENTILEEST:
         return PercentileEstValueAggregator.AGGREGATED_VALUE_TYPE;
--- a/pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/AggregationFunctionFactory.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/AggregationFunctionFactory.java
@@ -81,6 +81,8 @@ public class AggregationFunctionFactory {
             return new DistinctCountAggregationFunction();
           case DISTINCTCOUNTHLL:
             return new DistinctCountHLLAggregationFunction();
+          case DISTINCTCOUNTRAWHLL:
+            return new DistinctCountRawHLLAggregationFunction();
           case FASTHLL:
             return new FastHLLAggregationFunction();
           case COUNTMV:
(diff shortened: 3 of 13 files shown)


sunithabeeram:
Thanks for the contribution @ianvkoeppe. This introduces a high-level aggregation/keyword, so will be good to get input from @kishoreg on this. 

Also, just wondering if there are other ways of getting this info. One possibility is that if you store the serializedHLL as a separate column already, you might be able to get it via a "select ... limit 1" on the column; I haven't played with serialized-hll columns before, so can't say for sure if this is valid - will check on it.

ianvkoeppe:
@sunithabeeram, thank you for the quick reply. I'm happy to speak more with @kishoreg about our use case.

Interestingly, we actually already are planning to use serialized HLLs, but still want Pinot to do the aggregation across multiple segments before returning the serialized HLL.

A more concrete example would be; given a Pinot table which has...
**Unique Page Views** [SerializedHLL]
**PageId** [Int]

...and I want to get the total unique page views for 10,000 pages. I can't supply those 10,000 page ids in a single query because URI length will be exceeded; especially when ids are represented as 40+ character urns. So I batch calls by 1,000 pages and get 10 responses.

Today, I would have 10 responses with unique page views as a longs. I can't logically add them. With this change, I'm hoping I can get 10 serialized HLLs back, and use the HLLUtil client side to deserialize, aggregate, and then find cardinality.

sunithabeeram:
@ianvkoeppe sounds good. Will let @kishoreg/@Jackie-Jiang chime in on this.

ianvkoeppe:
@kishoreg and @Jackie-Jiang, please let me know if there is any additional information you need on your end.

kishoreg:
What is the problem with passing 10k page ids in the query. I understand it will be long but can we not use Post instead of GET.  

Also, is the 10k page ids constant across queries? How did we select them.

ianvkoeppe:
Thank you for your questions @kishoreg.

> What is the problem with passing 10k page ids in the query. I understand it will be long but can we not use Post instead of GET.

This is certainly an option but only fixes half the problem. Limiting the number of ids in the request can be just as much about limiting the response size as it is about limiting the request size. For our use case, there can be many demographic dimensions which match a given filter.

> Also, is the 10k page ids constant across queries? How did we select them.

They aren't necessarily constant; the 10K ids was simply meant as a generalized example of our specific use case.

kishoreg:
> This is certainly an option but only fixes half the problem. Limiting the number of ids in the request can be just as much about limiting the response size as it is about limiting the request size. For our use case, there can be many demographic dimensions which match a given filter.

I didn't follow this. If we use POST, the query can be of any size right? 

Overall, I think this is a good feature to have. However, I am unable to justify that the client side batching is more efficient in terms of network (bytes moved) and compute(the total time to process the query) than passing all the page ids within the query.

ianvkoeppe:
I think I can provide a more concrete example of what I meant, but there are other reasonable use cases which may be more persuasive.

Use Case 1: Response Size too Large.

In the case of having 10,000 IDs, let's say we group by a persons job title. There are many job titles, so for 10,000 unique IDs, we could have a very large response size which exceeds the amount of data which can be returned. We can improve performance by querying smaller groups of data by batching the IDs and then aggregating matching records client size. (In a way, this is working around not having the ability to page results when using aggregation functions).

Use Case 2: Cross Column or Cross Table Aggregations.

Say I have two tables; one tracks page views by member, and another tracks ad clicks by member. If I want to see unique members who visited my website and also clicked on by ad, I would not be able to query both and add their raw numbers. Assuming I'm using the same member id in both HLLs, I could merge the raw HLL responses from Pinot to achieve this.

kishoreg:
Use case 2 makes sense. Use case 1 also makes sense but that can be achieved today without this change right?.

From the original description, I thought your requirement was something as follows -

``` select discountCountHLL(memberId) from T where pageId in (...... 10,000 IDs)```
and you wanted to solve this by running multiple (10 in this case) queries of the form 
``` select DISTINCTCOUNTRAWHLL(memberId) from T where pageId in (1000 IDs ...)```
 and you would then aggregate the raw HLL on the client side to get the actual distinctCount across all IDs.

Am I missing something?
Just to clarify, I think this is a valid feature request to have in Pinot and I will review it shortly. But wanted to get some clarity on the scenarios where this can be used. Use case 2 is one such use case.

**Comment by ianvkoeppe** [comment]
> From the original description, I thought your requirement was something as follows - select discountCountHLL(memberId) from T where pageId in (...... 10,000 IDs)
and you wanted to solve this by running multiple (10 in this case) queries of the form
select DISTINCTCOUNTRAWHLL(memberId) from T where pageId in (1000 IDs ...)

The only thing missing is that we might also have a dimension to group by...
`SELECT DISTINCTCOUNTRAWHLL(precomputedHLLColumn) FROM T WHERE pageId in (1000 IDS) GROUP BY JOB_TITLE`. Depending on the characteristics of the users visiting the pages, this could be tens of thousands of job titles. But I also mentioned, we have 50+ metrics in the table, so a request would be more like `SELECT SUM(col1), SUM(col2), ..., DISTINCTCOUNTRAWHLL(...)`. So the combination of a large table with many IDs, and a large dimension (job titles) means the response size could be arbitrarily large unless we make the filter more strict by batching IDs.

**Comment by mcvsubbu** [comment]
@ianvkoeppe once you know the full spec of this feature going in, please add documentation in docs/pql_examples.rst, and include it in this PR.

Thanks.

**Comment by ianvkoeppe** [comment]
@mcvsubbu, I added a little more color to the pql_examples.rst where I had mentioned the new functions. These are the only places I see HLLs mentioned, so it is the only place I updated.

File: docs/pql_examples.rst [hunk 1]
```diff
@@ -148,6 +148,7 @@ Supported aggregations on single-value columns
 * ``MINMAXRANGE``
 * ``DISTINCTCOUNT``
 * ``DISTINCTCOUNTHLL``
+* ``DISTINCTCOUNTRAWHLL``: Returns serialized HLL after initial aggregation. Allows post-query aggregation by clients.
```
mcvsubbu: You may wan to clarify that this is a string. Do you want to add example on how the result can be used?

ianvkoeppe:
Great idea @mcvsubbu. I added additional clarification on the HLL being represented as string and point to examples and use cases for handling the response.

kishoreg:
The code looks good to me. I think it's better to return byte[] as hexString instead of string. HllUtil has a toBytes method. The HLLUtil.toString has some additional overhead that can be avoided. This will allow you to use HyperLogLog library directly to parse the byte[] in HyperLogLog.

Coming back to the original problem. 
```SELECT func(m1), func(m2)..... FROM T WHERE pageId in <1000+ values> GROUP BY job_title (high cardinality)```

There are 4 possible options. 

1. No batching, just get all the results in one shot. Works if the cardinality of job_title < 100k
2. Batch by pageId
2. Batch by job_title
3. Batch by pageId and job_title (nested loop). 

If possible always pick 1. After that, batching by job_title is the right solution since each response is mutually exclusive and the client can simply stitch the responses together without additional processing.

But what you are suggesting is solution 1 - batch by pageId. I am not sure why this will be better unless there is some relationship between pageId and jobTitle such that restricting pageId will automatically limit jobTitle.

Does this line of reasoning make sense?

ianvkoeppe:
@kishoreg, thanks for your feedback.

> I think it's better to return byte[] as hexString instead of string. HllUtil has a toBytes method. The HLLUtil.toString has some additional overhead that can be avoided.

Awesome. I've converted it to use a HexString representation, and updated the tests accordingly.

> But what you are suggesting is solution 1 - batch by pageId. I am not sure why this will be better unless there is some relationship between pageId and jobTitle such that restricting pageId will automatically limit jobTitle. Does this line of reasoning make sense?

Absolutely, we are definitely in agreement overall. I think not batching makes sense in any case possible. My example above may not exactly outline the use case for needing to batch and merge large responses. I still think it could exist. In any case, being able to merge across columns or tables which have HLLs and use the same key for the HLL is a more straightforward use case and justification.

ianvkoeppe:
Anything else you need on your end @kishoreg?

ianvkoeppe:
@kishoreg and @mcvsubbu, is there anything else you need to be able to sign off on this change? We are wanting to leverage this in the near future.

PR #4296, opened 2019-06-10
[TE] Several small changes

xiaohui-sun:
* Changed session timeout to 24 hours.
 * Reverted the changes for adding additional delays when scheduling the tasks.


xiaohui-sun:
> Any particular reason for reverting the random backoff when saving the tasks?

This is not the root cause for the MySQL heavy load.

PR #4298, opened 2019-06-10
Package name changes for classes which moved from kafka->stream

npawar:
The file move in https://github.com/apache/incubator-pinot/pull/4252 didn't change the package declaration


pinot-connectors/pinot-connector-kafka-0.9/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaAvroMessageDecoder.java
```diff
@@ -41,6 +41,7 @@
 import org.apache.pinot.common.data.Schema;
 import org.apache.pinot.common.utils.retry.RetryPolicies;
 import org.apache.pinot.core.data.GenericRow;
+import org.apache.pinot.core.realtime.stream.AvroRecordToPinotRowGenerator;
```
mcvsubbu: hmm.. how did this compile?

PR #4301, opened 2019-06-11
[TE] populate the feedback in DTO for child anomalies

jihaozh:
Previously, the DTO for the child anomaly is loaded with the model mapper, which won't populate the feedbacks object. This PR fixes the issue by explicitly converting it into a DTO.


xiaohui-sun:
lgtm!

PR #4306, opened 2019-06-11
Added tests for RealtimeSegmentValidationManager

mcvsubbu:
Removed the dropping of realtime table and setting it up again. Relying on
the realtime table being created ahead of time

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
@@ -73,6 +73,8 @@ public class ControllerConf extends PropertiesConfiguration {
         "controller.validation.frequencyInSeconds";
     private static final String OFFLINE_SEGMENT_INTERVAL_CHECKER_FREQUENCY_IN_SECONDS =
         "controller.offline.segment.interval.checker.frequencyInSeconds";
+    private static final String SEGMENT_VALIDATION_INITIAL_DELAY_IN_SECONDS =
+        "controller.realtime.segment.validation.initialDelayInSeconds";
     private static final String REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS =
         "controller.realtime.segment.validation.frequencyInSeconds";
     private static final String BROKER_RESOURCE_VALIDATION_FREQUENCY_IN_SECONDS =
@@ -463,6 +465,10 @@ public class ControllerConf extends PropertiesConfiguration {
         getPeriodicTaskInitialDelayInSeconds());
   }
 
+  public void setSegmentLevelValidationInitialDelayInSeconds(long validationInitialDelayInSeconds) {
+    setProperty(ControllerPeriodicTasksConf.SEGMENT_LEVEL_VALIDATION_INTERVAL_IN_SECONDS, validationInitialDelayInSeconds);
+  }
+
   public void setBrokerResourceValidationInitialDelayInSeconds(long validationInitialDelayInSeconds) {
     setProperty(ControllerPeriodicTasksConf.BROKER_RESOURCE_VALIDATION_INITIAL_DELAY_IN_SECONDS,
         validationInitialDelayInSeconds);
@@ -635,6 +641,9 @@ public class ControllerConf extends PropertiesConfiguration {
   }
 
   public long getPeriodicTaskInitialDelayInSeconds() {
+    if (containsKey(ControllerPeriodicTasksConf.SEGMENT_LEVEL_VALIDATION_INTERVAL_IN_SECONDS)) {
+      return getLong(ControllerPeriodicTasksConf.SEGMENT_LEVEL_VALIDATION_INTERVAL_IN_SECONDS);
+    }
     return ControllerPeriodicTasksConf.getRandomInitialDelayInSeconds();
   }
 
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -73,6 +73,8 @@
         "controller.validation.frequencyInSeconds";
     private static final String OFFLINE_SEGMENT_INTERVAL_CHECKER_FREQUENCY_IN_SECONDS =
         "controller.offline.segment.interval.checker.frequencyInSeconds";
+    private static final String SEGMENT_VALIDATION_INITIAL_DELAY_IN_SECONDS =
```
npawar: should this be called REALTIME_SEGMENT_VALIDATION_INITIAL_DELAY_IN_SECONDS to match other variable conventions(like REALTIME_SEGMENT_VALIDATION_FREQUENCY_IN_SECONDS)

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -463,6 +465,10 @@ public long getBrokerResourceValidationInitialDelayInSeconds() {
         getPeriodicTaskInitialDelayInSeconds());
   }
 
+  public void setSegmentLevelValidationInitialDelayInSeconds(long validationInitialDelayInSeconds) {
```
npawar: setRealtimeSegmentValidationInitialDelayInSeconds ?

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -635,6 +641,9 @@ public void setOfflineSegmentIntervalCheckerInitialDelayInSeconds(long initialDe
   }
 
   public long getPeriodicTaskInitialDelayInSeconds() {
+    if (containsKey(ControllerPeriodicTasksConf.SEGMENT_LEVEL_VALIDATION_INTERVAL_IN_SECONDS)) {
```
npawar: This one is the generic PeriodicTask initial delay. The segment level validation seconds shouldn't be applied here. 
Maybe we need another on getRealtimeSegmentValidationInitialDelay()

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -635,6 +641,9 @@ public void setOfflineSegmentIntervalCheckerInitialDelayInSeconds(long initialDe
...
+      return getLong(ControllerPeriodicTasksConf.SEGMENT_LEVEL_VALIDATION_INTERVAL_IN_SECONDS);
+    }
```
npawar: getLong(key, defaultValue) should reduce this to 1 line

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ControllerPeriodicTasksIntegrationTests.java
```diff
@@ -499,6 +492,35 @@ public void testOfflineSegmentIntervalChecker()
...
+    Assert.assertTrue(controllerMetrics.getValueOfGlobalGauge(ControllerGauge.PERIODIC_TASK_NUM_TABLES_PROCESSED,
+        "RealtimeSegmentValidationManager") > 0);
+    RealtimeSegmentValidationManager validationManager = _controllerStarter.getRealtimeSegmentValidationManager();
+    ValidationMetrics validationMetrics = validationManager.getValidationMetrics();
+    // Make sure we processed the realtime table to get the total document count. Should have been done the first
+    // time RealtimeSegmentValidationManager ran on the default realtime table.
+    Assert.assertTrue(validationMetrics.getValueOfGauge(ValidationMetrics.makeGaugeName(getDefaultRealtimeTableName(),
+        "TotalDocumentCount")) > 0);
+  }
+
```
npawar: should we drop the realtime table in the after() for this test, or in the teardown() ?

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
mcvsubbu: All segment-level validation are done with a single call to get a random back-off. Therefore it makes sense to have a single variable. RealtimeSegemntValidationManager calls getPeriodicTasksInitialDelay(). Do we really want to introduce(even) more configs just for test sake?

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
mcvsubbu: See comment above regarding common configuration.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
mcvsubbu: No. The code calls this method  to get the initial delay

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
mcvsubbu: Disagree. We don't want users configuring this value, or provide a default if they don't. We want to choose random delay all the time except for tests.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ControllerPeriodicTasksIntegrationTests.java
mcvsubbu: Not necessary. We are tearing down the cluster at the end anyway.

PR #4308, opened 2019-06-12
Emit freshness lag at server level as well

sunithabeeram:
Currently, we emit a metric at broker level. This change makes it so that its emitted at the server level to make correlation/debugging easy.

Testing done: none

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
@@ -34,8 +34,10 @@ public interface RowMetadata {
 
   /**
    * Return the timestamp associated with when the row was ingested upstream.
-   *
    * Expected to be mainly used for stream-based sources.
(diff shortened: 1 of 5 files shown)


pinot-common/src/main/java/org/apache/pinot/common/metadata/RowMetadata.java
```diff
@@ -23,16 +23,19 @@
...
+ * Currently this is relevant for rows ingested into a mutable segment - the metadata is expected to be
+ * provided by the underlying stream.
  */
 @InterfaceAudience.Public
 @InterfaceStability.Evolving
 public interface RowMetadata {
 
   /**
-   * Return the ingestion timestamp of the row.
+   * Return the timestamp associated with when the row was ingested upstream.
```
mcvsubbu: Please also include in the comments, what value an implementation should return if it does not have the information. Not all streams may provide a timestamp when an event was ingested upstream.

PR #4313, opened 2019-06-13
[TE] Use multimap in metric urn as dimension

jihaozh:
Previously we use a map to store the dimension values so that for each dimension, there can only be one value. In this PR, we fix the issue by using the information stored in the metric urn to construct a multi-map.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionUtils.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import org.apache.pinot.thirdeye.common.dimension.DimensionMap;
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionUtils.java
```diff
@@ -55,8 +56,8 @@
   // TODO anomaly should support multimap
   public static DimensionMap toFilterMap(Multimap<String, String> filters) {
     DimensionMap map = new DimensionMap();
-    for (Map.Entry<String, String> entry : filters.entries()) {
-      map.put(entry.getKey(), entry.getValue());
+    for (Map.Entry<String, Collection<String>> entry: filters.asMap().entrySet()){
+      map.put(entry.getKey(), String.join(", ", entry.getValue()));
```
xiaohui-sun: Any reason use ", " instead of ","?

jihaozh: This dimension map is only used for presenting. Using ", " looks better than ",".

Issue #4317, opened 2019-06-14
Support variable length Offline Dictionary Indexes for bytes, strings and maps to save on storage


buchireddy:
**What?**
Currently, the dictionary index for offline segments for bytes and string types uses Fixed-size storage for each value (by picking the size of the max element and padding the smaller elements with "0"). See org.apache.pinot.core.io.util.FixedByteValueReaderWriter 
The idea is to avoid padding and support storing byte arrays/strings/maps of different length while not slowing down the lookups much (obviously).

**Why?**
Fixed size based storage is good for fast lookups but it's very inefficient for the storage. For example, if we have a String column and the size of the biggest string value is 100 bytes but the average size is only 10 bytes, there is about 90% padding. The same thing applies for byte[], maps, etc.

**How?**
Currently, `FixedByteValueReaderWriter` only writes the sorted values in the buffer directly starting from "0" offset and at fixed lengths. So, first Int is at index "0" and the second one at index "4", etc. There is no additional metadata needed in the buffer.
The idea is to maintain the index of each element at the beginning of the buffer so that the element sizes needn't be fixed. When looking up an element from the buffer, we first get it's offset and then read the actual element. This means we do two reads from the buffer (first int offset and then the actual element) but the offset read should be fast enough so it shouldn't slow down the overall operation that much.

Few things to note:
* If all values of a byte[], string or map column have fixed length, this approach rather adds storage overhead and one additional lookup and might not be preferable. Hence, we can have a flag/property at the column level to decide whether to use the VarLengthByteValueReaderWriter or not.
* Backward compatibility shouldn't be broken, which means we need to introduce some kind of header into the buffer to be able to distinguish the on-disk storage format.
* Need to run Benchmarks to see the lookup overhead added by this approach.
* If possible, we should do some benchmarking to get the storage savings with the new approach so that we can make data-driven decisions.

Thanks @kishoreg for pointing this problem and brainstorming.

P.S: This was originally tracked in https://github.com/winedepot/pinot/issues/24

buchireddy:
I've implemented a solution based on the approached discussed in the description and did some benchmarks with **String dictionary** to see the latency with VariableLength dictionary and the storage improvements it brings compared to the FixedLength dictionary. Here are the results and observations.

**Time taken to lookup 10M values:**
[...]

PR #4318, opened 2019-06-14
#4317 Feature/variable length bytes offline dictionary

buchireddy:
PR for the feature mentioned in #4317


snleee:
@buchireddy Can you rebase the code from `apache/incubator-pinot` master branch? Also, it seems that your pr includes some other commits from `winedepot` branch. If would be great if you can cherry pick the branch so that this pr only includes the code changes related to variable length byte & string support for dictionary.

buchireddy:
@snleee Closing this PR since I don't want to merge all other changes from winedepot fork. I've open https://github.com/apache/incubator-pinot/pull/4321 with the intended changes only.

mcvsubbu:
@buchireddy  did you consider the use of [MutableOffheapByteArrayStore](https://github.com/apache/incubator-pinot/blob/master/pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java)? I am not able to readily say whether your implementation is based on that or is something parallel.

buchireddy:
@mcvsubbu I did look at `MutableOffheapByteArrayStore` and took some concepts from it while implementing `VarLengthBytesValueReaderWriter` too. However, I haven't used the former fully because of the following reasons:
* In order to support a seamless transition from fixed length to variable length offline dictionary, we need a special magic header to be added at the beginning, which needs some changes to  `MutableOffheapByteArrayStore`.
* `MutableOffheapByteArrayStore` is handling the buffer expansions, multiple buffers logic because it needs to be mutable but the immutable one doesn't need all those.

So, it's a trade-off b/w simple code vs avoiding the code duplication (though we anyways need a new class for variable length dictionary class implementation). I'm open for debate and for further brainstorming.

P.S: This is my first time going through Pinot code so please let me know if I'm missing any aspects here.

mcvsubbu:
Mutable byte array store can be modified to have a header, no problem. It is currently used as a volatile store for comsuming segments, so changing it to include a header is not backward incompatible.

Why not modify it to have the header and also derive an unmutable class from it to be used during segment generation?

kishoreg:
Let’s move the discussion to the issue #4317.

PR #4321, opened 2019-06-14
#4317 Feature/variable length bytes offline dictionary for indexing bytes and string dicts.

buchireddy:
Implementation of the feature mentioned in https://github.com/apache/incubator-pinot/issues/4317


pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+ */
+public class VarLengthBytesValueReaderWriter implements Closeable, ValueReader {
+
+  /**
+   * Header used to identify the dictionary files written in variable length bytes
+   * format. Keeping this a mix of alphanumeric with special character to avoid
+   * collisions with the regular int/string dictionary values written in fixed size
+   * format.
+   */
+  private static final byte[] MAGIC_HEADER = StringUtil.encodeUtf8("....vl1;");
```
kishoreg: we need to think of a better way to store the magic header. See Gzip for  inspiration. https://en.wikipedia.org/wiki/Gzip 
```a 10-byte header, containing a magic number (1f 8b), compression ID (08 for DEFLATE), file flags, a 32-bit timestamp, compression flags and operating system ID.```

File: pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java [hunk 2]
```diff
@@ -0,0 +1,196 @@
...
+  private static final int MAGIC_HEADER_LENGTH = MAGIC_HEADER.length;
+  private static final int INDEX_ARRAY_START_OFFSET = MAGIC_HEADER.length + Integer.BYTES;
```
kishoreg: java doc for what is INDEX_ARRAY or use the same terms used in class level java doc

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+    return length;
+  }
+
+  public static boolean isVarLengthBytesDictBuffer(PinotDataBuffer buffer) {
+    // If the buffer is smaller than header + numElements size, it's not var length dictionary.
+    if (buffer.size() > INDEX_ARRAY_START_OFFSET) {
+      byte[] header = new byte[MAGIC_HEADER_LENGTH];
+      buffer.copyTo(0, header, 0, MAGIC_HEADER_LENGTH);
+
+      if (Arrays.equals(MAGIC_HEADER, header)) {
```
kishoreg: what happens when we move to vl2 and old files have vl1?

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+      _dataBuffer.putByte(offset, MAGIC_HEADER[offset]);
+    }
+  }
+
+  public void init(byte[][] byteArrays) {
+    this._numElements = byteArrays.length;
+
+    writeHeader();
+
+    // Add the number of elements as the first field in the buffer.
```
kishoreg: the class javadoc does not talk about writing the number of elements

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+    nextArrayStartOffset = INDEX_ARRAY_START_OFFSET + Integer.BYTES * byteArrays.length;
+    for (byte[] array: byteArrays) {
+      _dataBuffer.readFrom(nextArrayStartOffset, array);
+      nextArrayStartOffset += array.length;
+    }
+  }
+
+  int getNumElements() {
+    // Lazily initialize the numElements.
+    if (_numElements == 0) {
```
kishoreg: initialize it to -1 to handle the case of an empty segment.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+    return StringUtil.decodeUtf8(getBytes(index, numBytesPerValue, buffer));
+  }
+
+  @Override
+  public String getPaddedString(int index, int numBytesPerValue, byte[] buffer) {
+    return StringUtil.decodeUtf8(getBytes(index, numBytesPerValue, buffer));
+  }
+
+  @Override
+  public byte[] getBytes(int index, int numBytesPerValue, byte[] buffer) {
```
kishoreg: can we store an additional offset to make this code very simple without any branching?
length = buf[i+1] - buf[i]

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+    if (index == getNumElements() - 1) {
+      // This is the last byte array in the buffer so use buffer size to get the length.
+      length = (int) (_dataBuffer.size() - offset);
+    }
+    else {
+      length = _dataBuffer.getInt(INDEX_ARRAY_START_OFFSET + Integer.BYTES * (index + 1)) - offset;
+    }
+
+    byte[] b;
+    // If the caller didn't pass a buffer, create one with exact length.
```
kishoreg: what is the expectation here? lets have to methods if we really need to support the case where its null

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
```diff
@@ -211,6 +203,39 @@ public void build()
     }
   }
 
+  private void writeBytesValueDictionary(int numValues, byte[][] sortedByteArrays) throws IOException {
```
kishoreg: java doc for this method. This is invoked for string and BYTES right?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
```diff
@@ -234,7 +259,7 @@ public int indexOfSV(Object value) {
     }
   }
 
-  public int[] indexOfMV(Object value) {
+  int[] indexOfMV(Object value) {
```
kishoreg: why are we decreasing the scope

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
mcvsubbu: I would include a version number either in the header or following the header. This will allow us to auto-detect dictionary types on load, and evolve it

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
mcvsubbu: Why not use a structure where byte arrays are stored at one and the offsets at the other? I agree that is useful for mutability, but with a pre-calculated length, mutable can be same as immutable

pinot-common/src/main/java/org/apache/pinot/common/config/IndexingConfig.java
```diff
@@ -223,6 +231,14 @@ public boolean isAggregateMetrics() {
     return _aggregateMetrics;
   }
 
+  public List<String> getVarLengthDictionaryColumns() {
+    return _varLengthDictionaryColumns;
+  }
+
+  public void setVarLengthDictionaryColumns(List<String> _varLengthDictionaryColumns) {
+    this._varLengthDictionaryColumns = _varLengthDictionaryColumns;
```
mayankshriv: Nit: remove redundant this.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
```diff
@@ -145,6 +146,8 @@ public HLRealtimeSegmentDataManager(final RealtimeSegmentZKMetadata realtimeSegm
     // No DictionaryColumns
     noDictionaryColumns = new ArrayList<>(indexLoadingConfig.getNoDictionaryColumns());
 
+    varLengthDictionaryColumns = new ArrayList<>(indexLoadingConfig.getVarLengthDictionaryColumns());
```
mayankshriv: The PR description states this is for offline segments only, why does RealtimeSegmentDataManager need to know about this?

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -237,6 +237,7 @@ public void deleteSegmentFile() {
   private final String _timeColumnName;
   private final List<String> _invertedIndexColumns;
   private final List<String> _noDictionaryColumns;
+  private final List<String> _varLengthDictionaryColumns;
```
mayankshriv: Same here, why does RT need to know about var length dict columns if this PR is specifically for offline?

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public String getUnpaddedString(int index, int numBytesPerValue, byte paddingByte, byte[] buffer) {
+    return StringUtil.decodeUtf8(getBytes(index, numBytesPerValue, buffer));
+  }
+
+  @Override
+  public String getPaddedString(int index, int numBytesPerValue, byte[] buffer) {
```
mayankshriv: This should throw an exception instead? The callers are expecting a padded string, will any of the caller break if this just returns the raw string?

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+ * <ul>
+ *   <li> Integer offsets to start position of byte arrays. </li>
+ *   <li> All byte arrays. </li>
+ * </ul>
+ *
+ * Only sequential writes are supported.
+ *
+ * @see FixedByteValueReaderWriter
+ */
+public class VarLengthBytesValueReaderWriter implements Closeable, ValueReader {
```
mayankshriv: The convention is to use ReaderWriter for mutable segments, and those classes implement both Reader and writer (IIRC). This naming is breaking that convention and would create confusion.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+import org.apache.pinot.core.segment.memory.PinotDataBuffer;
+
+/**
+ * Implementation of {@link ValueReader} that will allow each value to be of variable
+ * length and there by avoiding the unnecessary padding.
+ *
+ * The layout of the file is as follows:
+ * <p> Header Section: </p>
+ * <ul>
+ *   <li> Magic header byte sequence. </li>
```
mayankshriv: Adding the following fields in the header will help with compatibility issues:

1. Version number
2. Start offset of (non-header) data. This way there is no implicit assumption on size of header and that data starts where header ends.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+   */
+  private static final byte[] MAGIC_HEADER = StringUtil.encodeUtf8("....vl1;");
+  private static final int MAGIC_HEADER_LENGTH = MAGIC_HEADER.length;
+  private static final int INDEX_ARRAY_START_OFFSET = MAGIC_HEADER.length + Integer.BYTES;
+
+  private final PinotDataBuffer _dataBuffer;
+  private transient int _numElements;
+
+  public VarLengthBytesValueReaderWriter(PinotDataBuffer dataBuffer) {
+    this._dataBuffer = dataBuffer;
```
mayankshriv: Nit: Remove redundant `this`.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  public static long getRequiredSize(byte[][] byteArrays) {
```
mayankshriv: The existing implementation has inefficiencies such as converting all strings to bytes (to compute max string length, for padding). Consider cleaning it up (ok to file an issue and addressing in separate PR). I understand this implementation still needs to know the required size, but there could be heuristics such as for large number of strings, may be better to resize the storage as needed, as opposed to converting all strings to byte[] upfront.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+    int length;
+    if (index == getNumElements() - 1) {
+      // This is the last byte array in the buffer so use buffer size to get the length.
+      length = (int) (_dataBuffer.size() - offset);
+    }
+    else {
+      length = _dataBuffer.getInt(INDEX_ARRAY_START_OFFSET + Integer.BYTES * (index + 1)) - offset;
+    }
+
+    byte[] b;
```
mayankshriv: Consider re-usable ThreadLocal buffer.

pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
```diff
@@ -54,10 +54,12 @@
   private List<String> invertedIndexColumns;
   private List<String> noDictionaryColumns;
   private StarTreeIndexSpec starTreeIndexSpec;
+  private List<String> varLengthDictionaryColumns;
 
   public RealtimeSegmentConverter(MutableSegmentImpl realtimeSegment, String outputPath, Schema schema,
       String tableName, String timeColumnName, String segmentName, String sortedColumn,
-      List<String> invertedIndexColumns, List<String> noDictionaryColumns, StarTreeIndexSpec starTreeIndexSpec) {
+      List<String> invertedIndexColumns, List<String> noDictionaryColumns,
+      List<String> varLengthDictionaryColumns, StarTreeIndexSpec starTreeIndexSpec) {
```
mayankshriv: I am not clear on why RT code needs to know about varLengthDictColumns, if we are not supporting them in this PR.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+  }
+
+  int getNumElements() {
+    // Lazily initialize the numElements.
+    if (_numElements == 0) {
+      _numElements = _dataBuffer.getInt(MAGIC_HEADER_LENGTH);
+      Preconditions.checkArgument(_numElements > 0);
+    }
+    return this._numElements;
```
mayankshriv: Remove redundant `this` (from all places).

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
```diff
@@ -211,6 +203,39 @@ public void build()
...
+
+    if (_useVarLengthDictionary) {
+      // Backward-compatible: index file is always big-endian
+      long size = VarLengthBytesValueReaderWriter.getRequiredSize(sortedByteArrays);
+      try (PinotDataBuffer dataBuffer = PinotDataBuffer
+          .mapFile(_dictionaryFile, false, 0, size, ByteOrder.BIG_ENDIAN,
+              getClass().getSimpleName());
+          VarLengthBytesValueReaderWriter writer = new VarLengthBytesValueReaderWriter(
+              dataBuffer)) {
+        writer.init(sortedByteArrays);
```
mayankshriv: Any reason why the `init` does the `write` as well (comparing with existing implementation)? This caller side code becomes less readable, as readers will question why only init is called, and then go to `init` and figure out that init is also writing the values.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
```diff
@@ -211,6 +203,39 @@ public void build()
     }
   }
 
+  private void writeBytesValueDictionary(int numValues, byte[][] sortedByteArrays) throws IOException {
+
+    if (_useVarLengthDictionary) {
```
mayankshriv: Can this `if/else` be refactored such that the writer is set conditionally and the rest of the code is outside?

pinot-core/src/main/java/org/apache/pinot/core/segment/index/loader/defaultcolumn/BaseDefaultColumnHandler.java
```diff
@@ -311,12 +311,14 @@ protected void createColumnV1Indices(String column)
             totalDocs, maxNumberOfMultiValueElements);
 
     ColumnIndexCreationInfo columnIndexCreationInfo =
-        new ColumnIndexCreationInfo(columnStatistics, true/*createDictionary*/, ForwardIndexType.FIXED_BIT_COMPRESSED,
+        new ColumnIndexCreationInfo(columnStatistics, true/*createDictionary*/,
```
mayankshriv: Is this a format-only change? Is Pinot styling being used here?

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/ImmutableDictionaryReader.java
```diff
@@ -244,6 +252,6 @@ protected String getPaddedString(int dictId, byte[] buffer) {
   }
 
   protected byte[] getBuffer() {
-    return new byte[_numBytesPerValue];
+    return (_valueReader instanceof VarLengthBytesValueReaderWriter) ? null : new byte[_numBytesPerValue];
```
mayankshriv: Instance of can be avoided by using ThreadLocal byte[] that is initialized to null in the constructor (already has the `if/else`)?

mcvsubbu:
@buchireddy , @kishoreg please provide your comments on https://github.com/apache/incubator-pinot/pull/4348. I need a ship-it from one of you

I have not addressed most of the comments by @mayankshriv which will be in the implementation of VarLengthBytesValueReaderWriter.   Basically, my proposal is to build mutable and immutable versions of OffHeapByteArrayStore. Either we can derive two classes out of this and use them, or we can play around with the ctor of OffHeapByteArraystore to make it independent of memory manager etc. (caller allocates) and provide only a pnot-databuffer that is mutable (vs the immutable, which could then becoime an empty constructor with an init call). We can work around these things. All I want is one way to encode strings in a compact fashion. thanks.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,196 @@
...
+  public String getPaddedString(int index, int numBytesPerValue, byte[] buffer) {
+    return StringUtil.decodeUtf8(getBytes(index, numBytesPerValue, buffer));
+  }
+
+  @Override
+  public byte[] getBytes(int index, int numBytesPerValue, byte[] buffer) {
+    // Read the offset of the byte array first and then read the actual byte array.
+    int offset = _dataBuffer.getInt(INDEX_ARRAY_START_OFFSET + Integer.BYTES * index);
+
+    // To get the length of the byte array, we use the next byte array offset. However,
```
Jackie-Jiang: I would suggest storing an extra end offset for the last element to avoid the if check for each read. CPU does not perform the best with such if condition.

pinot-common/src/main/java/org/apache/pinot/common/config/IndexingConfig.java
buchireddy: Done.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Done.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Done.

buchireddy:
FTR, I was waiting to address review comments on this PR until now because @mcvsubbu was working on pulling out `Buffer` as a common class (PR #4348) which can be reused in this PR as well. I personally think that's a nice-to-have change and will help to reuse the byte array store code for both mutable and immutable dictionaries.
However, @Jackie-Jiang is working on refactoring/cleaning the mutable byte array store related code so #4348 isn't going to be merged soon. Also, @Jackie-Jiang and I discussed that the byte array store used for VariableLength dictionary is different enough from the regular mutable store that it's okay for VarLength dictionary to have it's own immutable bytes store. We can definitely find better optimizations for immutable byte store that may not be applicable for mutable store.
Hence, I am now going to work on addressing the review comments on this PR and will get it ready to be merged. @mcvsubbu @Jackie-Jiang Please reply if you guys think otherwise.
cc @kishoreg

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Handled.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Added javadoc for this as well.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
buchireddy: I don't think the `public` scope is needed, hence removed but i'll revert that change.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Done.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Done.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
buchireddy: Done.

kishoreg:
@buchireddy Have you addressed all the comments? @Jackie-Jiang, can you take another look and approve.

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/ImmutableDictionaryReader.java
buchireddy: Fixed.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Yes. I've discuss this with @kishoreg in-person but the gist of it is, this new class is named exactly on the lines of an existing class `FixedByteValueReaderWriter` to be consistent. I'll do a new PR cleaning up both of these.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Sure, will take up in another PR.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Initializing this to `-1` now.
@mcvsubbu we can do that too but I'm not see any huge differences because of that. Can you think of any other advantages with that approach?

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: As per my knowledge, the callers of this method are assuming the result is padded only under `paddingByte != 0` but this class will never be used in those cases.

IMO, all this confusion is happening because `ValueReader.java` has the knowledge of padding/non-padding in the interface methods but it shouldn't be. The padding should be implementation detail. I'm planning to cleanup this along with the naming of `FixedByteValueReaderWriter` in another PR so leaving it for now.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: I've considered that but we need to know the size of the biggest element so that we can create a big enough buffer to handle all elements. For us to know the size of biggest element, we either need to store that length separately or should read all elements at once in the beginning when the object is initialized, both seemed like an overkill.

Do you think the garbage gain using ThreadLocal byte[] will be worth the complexity? I'm fine to implement that change.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: IMO, the intermediate buffers used to read a full element should be the concern of this class, not the caller. In this case, since the elements are of variable length, caller can't know the size of element and hence can't pass buffer with correct size hence it passes null.
I'm planning to do another PR where we cleanup the `ValueReader.java` and change the `getBytes()` method not to take any buffer so leaving it like this for now.

pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
buchireddy: We need to know this information while converting realtime segments into offline segments. Do you have any suggestions to get that without touching these realtime classes?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
buchireddy: Strictly speaking, `VarLengthBytesValueReaderWriter` and `FixedByteValueReaderWriter` are two implementations of readers but there is no common `Writer` interface here and hence the way the `byte[]` are written isn't uniform either.
Like I mentioned in another comment, I'm planning to do another PR to rename the classes and `ValueReader.java` interface so I'll see if this needs more cleanup then.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
buchireddy: I've renamed `init()` to `write()` to be clear. Please see my above comment for more details.

buchireddy:
@kishoreg @Jackie-Jiang @mayankshriv I've addressed the review comments and pushed the latest code. Please take a look and LMK if you have any more comments.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: I've added version number in the most recent commit.
About 2, having such offset will force us to read that value before reading any `byte[]` from the store. Currently, the read side code is simple because you know where the data starts already and can directly read. So, my take is, unless there is a strong reason, it might be better off leaving it like this.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: I've added version in the latest commit which will be used for the evolution of this store/data structure.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/HLRealtimeSegmentDataManager.java
mayankshriv: I see, makes sense.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
mayankshriv: Hmm, I am not sure about that. IIRC, '\0' is a valid padding char (ie used for padding), and does not mean absence of padding.

Also, it is likely that the exposing the api's (padded vs unpadded) was an optimization to perform operations on padded strings (ie avoid creating new unpadded strings), and strip padding at the end.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
mayankshriv: Please file an issue for all cleanups not being addressed in this PR, so we can track them.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
mayankshriv: For 2, one benefit is that if later we want to add more metadata in the header without changing the rest of the format, we can do so without having to create a new version number of worrying about backward compatibility. The old reader will still be able to read the new file (with new metadata), it will just not see the new metadata.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
mayankshriv: Please file an issue to track.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
mayankshriv: Well, another options is to start with a size, and resize as needed. Chances are that we will resize to max quickly, i.e. without creating a lot of garbage. Of course, in the pathalogical case we will resize once per value, but that is no worse than allocating memory each time. I have found reducing garbage at lowest level reader/writers to help with high throughput-low latency use cases.

pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
mayankshriv: I guess if RealtimeSegmentConverter took a segmentGeneratorConfig as param, then this would have been cleaner. May be file an issue for cleanup, if you agree?

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: @mayankshriv I've checked this again and you're correct that the variable length dict should throw exception for this method. Thanks for pointing it.

I've made the change and pushed it.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Filed https://github.com/apache/incubator-pinot/issues/4393 to do the cleanup.

pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
kishoreg: Instead, we should deprecate segmentGenerationConfig and use TableConfig everywhere.

buchireddy: I've not spent enough time on this part of code to give my take on what's the best thing to do.

If one of you can please log an issue with more details, I'll take it up next, check the code and can make the changes finally.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: I like the idea but I think we should split this into a separate change. Should we log an enhancement request and take up next?

pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
kishoreg: This is outside the scope of this PR. Dont worry about it.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: I've added version number in the header now which should let us evolve the store.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Filed https://github.com/apache/incubator-pinot/issues/4393 to handle this.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Logged https://github.com/apache/incubator-pinot/issues/4401 to handle this in a later PR.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Fixed this and pushed my changes again.

PR #4323, opened 2019-06-15
[Controller Separation] Add logic for lead controller resource

jackjlli:
This is the last PR of controller separation ([Design doc](https://cwiki.apache.org/confluence/display/PINOT/Controller+Separation+between+Helix+and+Pinot)).

This PR:

-  adds all the logic needed for lead controller resource. ControllerLeadershipManager is removed in this PR. All the periodic tasks are run depending on whether the current controller is the leader for given tables.
- Murmur2 is copied over from org.apache.kafka.common.utils.Utils::murmur2

Testing different partition functions:
In order to make all the tables more evenly distribute to all 24 partitions, we benchmarked the partition functions against 2300+ tables. The result "Diff" is calculated by  (max number of tables in partitions - min number of tables in partitions). Here are the details:
* Math.floorMod(rawTableName.hashCode(), 24)
Diff : 45
* Murmur2(rawTableName, 24)
Diff : 28
* (rawTableName.hashCode() & Integer.MAX_VALUE) % 24
Diff : 33
* Math.abs(rawTableName.hashCode()) % 24
Diff : 56

The result shows `Murmur2` and `(rawTableName.hashCode() & Integer.MAX_VALUE) % 24` performs better distribution on the tables.

Removed metrics in this PR:
*   PINOT_CONTROLLER_LEADER("PinotControllerLeader", true)


pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
```diff
@@ -1,117 +0,0 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-
-/**
- * Single place for listening on controller changes.
- * This should be created at controller startup and everyone who wants to listen to controller changes should subscribe.
- */
-@ThreadSafe
-public class ControllerLeadershipManager {
```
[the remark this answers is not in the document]
jackjlli: The behavior of running periodic tasks has changed. Previous it depends on the notification from controller listener. When it's Helix leader, we start periodic task scheduler; when it isn't, we stop periodic task scheduler.
The code change here will always run the scheduler. Inside the for loop we call `_leadControllerManager.isLeaderForTable(tableNameWithType)` to determine whether we should use Partition leader or Helix leader.
This class can still be useful for caching helix leader (which reduces ZK access) and emitting metric. I'll bring this class back and remove the logic with subscribers in the next push.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/HelixHelper.java
```diff
@@ -500,4 +503,42 @@ public IdealState apply(IdealState idealState) {
...
+  public static int getHashCodeForTable(String rawTableName) {
+    return rawTableName.hashCode();
+  }
+
+  /**
+   * Gets lead controller for table from lead controller resource.
+   * If the resource is disabled or no controller registered as participant, there is no instance in "MASTER" state.
+   * @param leadControllerResourceExternalView external view of lead controller resource
+   * @param rawTableName table name without type
+   * @return leader of partition, null if not found.
```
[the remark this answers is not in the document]
jackjlli: There's no need to throw exception. There'll be no difference between throwing an exception and logging a message saying partition leader is not found.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/HelixHelper.java
```diff
@@ -500,4 +503,42 @@ public IdealState apply(IdealState idealState) {
   public static Set<String> getBrokerInstancesForTenant(List<InstanceConfig> instanceConfigs, String tenant) {
     return new HashSet<>(HelixHelper.getInstancesWithTag(instanceConfigs, TagNameUtils.getBrokerTagForTenant(tenant)));
   }
+
+  /**
+   * Gets hash code for table.
+   * @param rawTableName table name
+   * @return hash code
+   */
+  public static int getHashCodeForTable(String rawTableName) {
```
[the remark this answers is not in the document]
jackjlli: Actually this hash function can be changed **only when** lead controller resource is disabled. It cannot be changed when the resource is enabled. I'll add it to the comment.

mcvsubbu:
I would wait for at least one other review on the API. Perhaps the API can be a separate checkin.

One thought that comes to mind is have an API URI like: "/leadController/tables" -- return the lead controller for all tables and "/leadController/tables/tablenameregex" return the lead controller for tables matching a regex.

The output in both cases can be like:
`
  [
    "tableName" : {
       "partition" : 9,
       "leader" : "hostName"
     },
     ...
  ]
`

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -133,6 +134,7 @@ private static long getRandomInitialDelayInSeconds() {
   // If it's set to false, existing HLC realtime tables will stop consumption, and creation of new HLC tables will be disallowed.
   // Please make sure there is no HLC table running in the cluster before disallowing it.
   private static final String ALLOW_HLC_TABLES = "controller.allow.hlc.tables";
+  private static final String HELIX_INSTANCE_TYPE = "controller.helix.instance.type";
```
mcvsubbu: let us not use a config to make sure that only controllers register as participants. Instead, let us have two constructors in PinotHelixResourceManager: One with (config) and the other with (config, instanceType). We use the first ctor everywhere except the ControllerStarter. In the controllerstarter, we can either use the second ctor, or use some static method like makePnotHelixResoirceManagerForController() and make the second ctor private.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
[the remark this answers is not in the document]
mcvsubbu: ok, so there is at least one more check-in before we enable the resource, right? Can you update your checkin comment to reflect that? thanks

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -139,6 +139,9 @@ public ControllerStarter(ControllerConf conf) {
     } else {
       _adminApp =
           new ControllerAdminApiApplication(_config.getQueryConsoleWebappPath(), _config.getQueryConsoleUseHttps());
+      // Helix instance type should explicitly be set to PARTICIPANT ONLY in {@link ControllerStarter}.
+      // Other places like {@link PerfBenchmarkDriver} which directly call {@link PinotHelixResourceManager} should NOT register as PARTICIPANT, which would be put to lead controller resource and mess up the leadership assignment. Those places should use ADMINISTRATOR other than PARTICIPANT.
+      _config.setHelixInstanceType(InstanceType.PARTICIPANT);
```
mcvsubbu: see comment above regarding the config

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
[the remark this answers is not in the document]
mcvsubbu: Also consider keeping the class disconnected and modifying it the next checkin to wire in as necessary if the helix resource is disabled.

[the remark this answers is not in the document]
mcvsubbu: ok, looks like there is some code in leadcontrolermanager for this...

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Override
+    public String toString() {
+      return "Partition index: " + getPartitionIndex() + ", lead controller id: " + getLeadControllerId();
+    }
+  }
+
+  /**
+   * Callback on changes in the controller. Should be registered to the controller callback.
```
mcvsubbu: can you elaborate this comment? This callback will only be triggered if the controller is in dual mode, right? It won't be triggered in any other mode.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -288,8 +278,10 @@ private void setUpPinotController() {
     // Setting up periodic tasks
     List<PeriodicTask> controllerPeriodicTasks = setupControllerPeriodicTasks();
     LOGGER.info("Init controller periodic tasks scheduler");
-    _controllerPeriodicTaskScheduler = new ControllerPeriodicTaskScheduler();
-    _controllerPeriodicTaskScheduler.init(controllerPeriodicTasks, _controllerLeadershipManager);
+    _periodicTaskScheduler = new PeriodicTaskScheduler();
```
mcvsubbu: let us avoid other gratuitous changes....

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -478,9 +475,10 @@ private void stopHelixController() {
 
   private void stopPinotController() {
     try {
-      // Stopping ControllerLeadershipManager has to be done before stopping HelixResourceManager.
-      LOGGER.info("Stopping controller leadership manager");
-      _controllerLeadershipManager.stop();
+      // Stopping periodic tasks has to be done before stopping HelixResourceManager.
+      // Stop controller periodic task.
+      LOGGER.info("Stopping controller periodic tasks");
```
mcvsubbu: We will need to stop lead controller manager so that we stop listening and reacting to state changes on leader controller resource?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/statemodel/LeadControllerResourceMasterSlaveStateModelFactory.java
```diff
@@ -0,0 +1,65 @@
...
+  public StateModel createNewStateModel(String resourceName, String partitionName) {
+    MasterSlaveStateModel stateModel = new LeadControllerResourceMasterSlaveStateModel();
+    stateModel.setPartitionName(partitionName);
+    return stateModel;
+  }
+
+  public class LeadControllerResourceMasterSlaveStateModel extends MasterSlaveStateModel {
+
+    @Override
+    public void onBecomeMasterFromSlave(Message message, NotificationContext context) {
```
mcvsubbu: Dont we need to unregister these during shutdown?

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,157 @@
...
+import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME;
+import static org.apache.pinot.common.utils.CommonConstants.Helix.NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE;
+
+
+public class LeadControllerManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(LeadControllerManager.class);
+
+  private final LeadControllerChecker _leadControllerChecker;
+  private PinotHelixResourceManager _pinotHelixResourceManager;
+  private boolean _amIHelixLeader = false;
```
mcvsubbu: volatile

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/HelixHelper.java
```diff
@@ -500,4 +503,55 @@ public IdealState apply(IdealState idealState) {
...
+
+  /**
+   * Gets hash code for table.
+   * Note: This method CANNOT be changed when lead controller resource has been enabled.
+   * Otherwise it will assign different controller for the same table, which will mess up the controller periodic tasks and realtime segment completion.
+   * @param rawTableName table name
+   * @return hash code
+   */
+  private static int getHashCodeForTable(String rawTableName) {
+    return rawTableName.hashCode();
```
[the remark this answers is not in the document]
jackjlli: Just did benchmark different partition functions and put the result in the description of this PR.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
jackjlli: Nope, the participant resource manager will register the controller listener. So once there's a controller change, all the participant will get notified, i.e. it'll work in both dual-mode and in Pinot-only mode.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
jackjlli: We don't need all the method in `ControllerPeriodicTaskScheduler`. Thus, we can just remove this class.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
jackjlli: No, stopping controller will mark all the periodic tasks to stop. Each periodic task will firstly check whether to stop/run, and then check whether to run the task for a given table.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/statemodel/LeadControllerResourceMasterSlaveStateModelFactory.java
jackjlli: Not necessary. All the cached partition number will be dropped after the shutdown.

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
```diff
@@ -78,35 +82,76 @@ public static ControllerLeaderLocator getInstance() {
...
+   * @param rawTableName table name without type
+   * @return the leader for this table.
+   */
+  private String getLeaderForTable(String rawTableName) {
+    String leaderForTable;
+    ExternalView leadControllerResourceExternalView =
+        _helixManager.getClusterManagmentTool().getResourceExternalView(_clusterName, LEAD_CONTROLLER_RESOURCE_NAME);
+    String partitionLeader = HelixHelper.getLeadControllerForTable(leadControllerResourceExternalView, rawTableName);
+    if (partitionLeader != null) {
+      // Converts participant id (with Prefix "Controller_") to controller id and assigns it as the leader.
```
mcvsubbu: Please co-locate the logic to convert a partition ID into a controller host name along with the same code that constructs the partition ID using controller host name

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
```diff
@@ -78,35 +82,76 @@ public static ControllerLeaderLocator getInstance() {
...
+          _controllerLeaderHostPort.getSecond());
+      return _controllerLeaderHostPort;
+    }
+  }
+
+  /**
+   * If partition leader exists, use this as the leader for realtime segment completion.
+   * Otherwise, try to use Helix leader.
+   * @param rawTableName table name without type
+   * @return the leader for this table.
```
mcvsubbu: clarify whether this returns hostname, or host_port or anythign else

pinot-core/src/test/java/org/apache/pinot/server/realtime/ControllerLeaderLocatorTest.java
```diff
@@ -19,6 +19,7 @@
 package org.apache.pinot.server.realtime;
 
 import org.apache.helix.BaseDataAccessor;
+import org.apache.helix.HelixAdmin;
```
mcvsubbu: Can you please add a test where the resource is disabled, we enable it, and then disable it, and make sure that the right code path s invoked in these cases?

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/HelixHelper.java
```diff
@@ -500,4 +503,55 @@ public IdealState apply(IdealState idealState) {
...
+    String partitionName = LEAD_CONTROLLER_RESOURCE_NAME + "_" + partitionIndex;
+    Map<String, String> partitionStateMap = leadControllerResourceExternalView.getStateMap(partitionName);
+
+    // Get master host from partition map. Return null if no master found.
+    for (Map.Entry<String, String> entry : partitionStateMap.entrySet()) {
+      if (MasterSlaveSMD.States.MASTER.name().equals(entry.getValue())) {
+        return entry.getKey();
+      }
+    }
+    LOGGER.info("No host in MASTER state in Partition: {} for Table: {}", partitionIndex, rawTableName);
```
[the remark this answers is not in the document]
jackjlli: Removed.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -255,30 +246,29 @@ private void setUpPinotController() {
...
-    LOGGER.info("Registering controller leadership manager");
-    // TODO: when Helix separation is completed, leadership only depends on the master in leadControllerResource, remove
-    //       ControllerLeadershipManager and this callback.
+    // Get lead controller manager from resource manager.
+    _leadControllerManager = _helixResourceManager.getLeadControllerManager();
+
+    LOGGER.info("Registering helix controller listener");
     helixParticipantManager.addControllerListener(
-        (ControllerChangeListener) changeContext -> _controllerLeadershipManager.onControllerChange());
+        (ControllerChangeListener) changeContext -> _leadControllerManager.onHelixControllerChange());
```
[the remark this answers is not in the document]
jackjlli: Added.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/statemodel/LeadControllerChecker.java
```diff
@@ -0,0 +1,57 @@
...
+import com.google.common.base.Preconditions;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.pinot.common.utils.CommonConstants.Helix.NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE;
+
+
+public class LeadControllerChecker {
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-core/src/test/java/org/apache/pinot/server/realtime/ControllerLeaderLocatorTest.java
jackjlli: We have such tests in `PinotControllerModeTest`

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
jackjlli: Done.

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
jackjlli: Done.

jackjlli:
An issue is created to track the API design: https://github.com/apache/incubator-pinot/issues/4441
We can remove the API out of this PR.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/HelixHelper.java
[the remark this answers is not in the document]
jackjlli: Murmur2 is in pinot-core module whereas this class is in pinot-common. It'll require us to refactor the code though.
I think using `(rawTableName.hashCode() & Integer.MAX_VALUE)` would be enough.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java

pinot-common/src/main/java/org/apache/pinot/common/utils/HashUtils.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+package org.apache.pinot.common.utils;
+
+import com.google.common.annotations.VisibleForTesting;
+
+
+public class HashUtils {
+
+  /**
+   * NOTE: This code has been copied over from org.apache.kafka.common.utils.Utils::murmur2
```
mcvsubbu: You can remove this note and just add a comment in the commit message.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return null;
+  }
+
+  public static String generateControllerParticipantId(String controllerHost, String controllerPort) {
+    return org.apache.pinot.common.utils.CommonConstants.Helix.PREFIX_OF_CONTROLLER_INSTANCE + controllerHost + "_"
+        + controllerPort;
+  }
+
+  public static String formPartitionName(int partitionIndex) {
```
mcvsubbu: generatePartitionName() to keep the method naming consistent?

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+    return org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME + "_" + partitionIndex;
+  }
+
+  public static int extractPartitionIndex(String partitionName) {
+    return Integer.parseInt(partitionName.substring(partitionName.lastIndexOf("_") + 1));
+  }
+
+  public static String extractLeadControllerHostNameAndPort(String controllerParticipantId) {
```
mcvsubbu: Please add javadoc on what the return format is.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Check whether the current controller is the leader for the given table. Return true if current controller is the leader for this table.
+   * Otherwise check whether the current controller is helix leader.
+   * @param tableName table name with/without table type.
+   */
+  public boolean isLeaderForTable(String tableName) {
+    String rawTableName = TableNameBuilder.extractRawTableName(tableName);
+    int partitionIndex =
+        LeadControllerUtils.getPartitionIdForTable(rawTableName, NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE);
+    if (_partitionIndexCache.containsKey(partitionIndex)) {
```
mcvsubbu: If there are 3 controllers, and controller A is leader for tables T1 and T2, and Controller B is leader for tables T3 and T4, and we call this method in contoller A for table T3 what happens?

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,163 @@
...
+      return isHelixLeader();
+    }
+  }
+
+  /**
+   * Get lead controller id and partition index for given table. Returns -1 as partition index and null as lead controller id if not found.
+   * @param tableName table name with/without table type.
+   * @return lead controller id and partition index
+   */
+  public LeadControllerResponse getLeadControllerAndPartitionIdForTable(String tableName) {
```
mcvsubbu: I thought we will add this API to a different PR. Can we remove this for now? And the LeadControllerResponse class as well? (we can use a pair?)

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,163 @@
...
+import org.slf4j.LoggerFactory;
+
+import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME;
+import static org.apache.pinot.common.utils.CommonConstants.Helix.NUMBER_OF_PARTITIONS_IN_LEAD_CONTROLLER_RESOURCE;
+
+
+public class LeadControllerManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(LeadControllerManager.class);
+
+  private Map<Integer, Integer> _partitionIndexCache;
```
mcvsubbu: Use  ConcurrentHashMap.KeySetView<K,Boolean> newKeySet()
 or just newSetFromMap, but newKeySet() is better.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
mcvsubbu: Please add a comment that this callback will be invoked whether we are in dual mode or not. It helps to clarify it. You can also probably repeat the comment in ControllerStarter that this is here to support the case when the lead controller resource is disabled, and is otherwise unused

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,163 @@
...
+  private PinotHelixResourceManager _pinotHelixResourceManager;
```
mcvsubbu: We seem to have a mutual dependency between PinotHelxResourceManager and this class. Can we break that?

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
```diff
@@ -78,35 +82,81 @@ public static ControllerLeaderLocator getInstance() {
...
+   * If partition leader exists, use this as the leader for realtime segment completion.
+   * Otherwise, try to use Helix leader.
+   * @param rawTableName table name without type
+   * @return the controller leader id with hostname and port for this table, e.g. localhost_9000
+   */
+  private String getLeaderForTable(String rawTableName) {
+    String leaderForTable;
+    ExternalView leadControllerResourceExternalView =
+        _helixManager.getClusterManagmentTool().getResourceExternalView(_clusterName, LEAD_CONTROLLER_RESOURCE_NAME);
+    String partitionLeader =
```
mcvsubbu: can we rename partitionLeader to leaderControllerInstance and maybe even change the method name to getLeadControllerInstanceForTable()

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,97 @@
...
+  public static int getPartitionIdForTable(String rawTableName, int numPartitions) {
+    return getHashCodeForTable(rawTableName) % numPartitions;
+  }
+
+  /**
+   * Gets lead controller for table from lead controller resource.
+   * If the resource is disabled or no controller registered as participant, there is no instance in "MASTER" state.
+   * @param leadControllerResourceExternalView external view of lead controller resource
+   * @param rawTableName table name without type
+   * @return leader of partition, null if not found or resource is disabled.
```
mcvsubbu: Please document the format that this return string is in. I believe it is the controller instance name, right?

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
```diff
@@ -78,35 +82,81 @@ public static ControllerLeaderLocator getInstance() {
...
+    } else {
+      // Gets Helix leader to be the leader to this table, otherwise returns null.
+      leaderForTable = getHelixClusterLeader();
+    }
+    return leaderForTable;
+  }
+
+  /**
+   * Gets Helix leader in the cluster. Null if there is no leader.
+   * @return Helix leader.
```
mcvsubbu: Please document the format in which this method returns the leader

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
jackjlli: If that's the case, the current controller returns `false` since it isn't the leader for the given table, which is mentioned in the javadoc of this method.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
jackjlli: The purpose of bringing `PinotHelxResourceManager` here is to fetch the EV of leadControllerResource, where all the requirements like `KeyBuilder` and `DataAccessor` live.
And `PinotHelxResourceManager` needs `LeadControllerManager` to register the cache for MasterSlaveStateModelFactory.
So it's hard to break that. That's why I added a method `registerResourceManager()` in `LeadControllerManager`. One thing we can do is add an `unregisterResourceManager()` in stop method.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
jackjlli: Yes, it's controller partition id like `Controller_localhost_9000`.

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
jackjlli: Done.

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
jackjlli: Removed.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/HashUtils.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/HashUtil.java
```diff
@@ -105,4 +106,52 @@ public static long hash64(final byte[] data, int length, int seed) {
     h ^= h >>> r;
     return h;
   }
+
+  /**
+   * Generates 32 bit murmur2 hash from byte array
+   * @param data byte array to hash
+   * @return 32 bit hash of the given array
+   */
+  @VisibleForTesting
```
mcvsubbu: Incorrect annotation. Please remove.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Gets hash code for table using murmur2 function, ignores the most significant bit.
+   * Note: This method CANNOT be changed when lead controller resource is enabled.
+   * Otherwise it will assign different controller for the same table, which will mess up the controller periodic tasks and realtime segment completion.
+   * @param rawTableName table name
+   * @return hash code ignoring the most significant bit.
+   */
+  private static int getHashCodeForTable(String rawTableName) {
+    return (HashUtil.murmur2(StringUtil.encodeUtf8(rawTableName)) & 0x7fffffff);
```
mcvsubbu:
```suggestion
    return (HashUtil.murmur2(StringUtil.encodeUtf8(rawTableName)) & Integer.MAX_VALUE);
```

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
mcvsubbu: What if we are running in dual mode, and A happens to be helix leader?

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/HashUtil.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -311,7 +311,7 @@ private SuccessResponse uploadSegment(FormDataMultiPart multiPart, boolean enabl
       // Validate segment
       SegmentValidatorResponse segmentValidatorResponse =
           new SegmentValidator(_pinotHelixResourceManager, _controllerConf, _executor, _connectionManager,
-              _controllerMetrics, _controllerLeadershipManager)
+              _controllerMetrics, _leadControllerManager)
```
mcvsubbu: Can we just get the current isLeader boolean for this table and pass it down instead of passing _leadControllerManager all the way down to execute one restli call? It will be cleaner to have smaller number of classes dependent on lead controller manager.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,138 @@
...
+    return _isLeadControllerResourceEnabled;
+  }
+
+  /**
+   * Callback on changes in the controller. Should be registered to the controller callback. This callback is not needed when the resource is enabled.
+   * However, the resource can be disabled sometime while the cluster is in operation, so we keep it here. Plus, it does not add much overhead.
+   * At some point in future when we stop supporting the disabled resource, we will remove this line altogether and the logic that goes with it.
+   */
+  void onHelixControllerChange() {
+    if (_pinotHelixResourceManager.isHelixLeader()) {
```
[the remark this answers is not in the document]
jackjlli: Even if we introduce a `_isShuttingDown` flag, it could still be possible that `_isShuttingDown` is false and then another thread immediately set it to be true before processing Helix controller change. Added synchronized on the methods.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -425,6 +416,23 @@ private void initPinotCrypterFactory() {
...
+    LOGGER.info("Registering helix controller listener");
+    helixParticipantManager.addControllerListener(
+        (ControllerChangeListener) changeContext -> _leadControllerManager.onHelixControllerChange());
+
+    LOGGER.info("Registering resource config listener");
+    try {
+      helixParticipantManager.addResourceConfigChangeListener(
+          (resourceConfigList, changeContext) -> _leadControllerManager.onResourceConfigChange());
+    } catch (Exception e) {
+      throw new RuntimeException("Error registering resource config listener", e);
```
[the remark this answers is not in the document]

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -425,6 +416,23 @@ private void initPinotCrypterFactory() {
     }
   }
 
+  private void registerListeners(HelixManager helixParticipantManager) {
+    // This registration is not needed when the resource is enabled.
```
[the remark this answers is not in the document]

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
mcvsubbu: thanks

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,160 @@
...
+  private volatile boolean _isShuttingDown = false;
+
+  public LeadControllerManager(String instanceId, HelixManager helixManager) {
+    _instanceId = instanceId;
+    _helixManager = helixManager;
+    _partitionIndexCache = ConcurrentHashMap.newKeySet();
+  }
+
+  /**
+   * Marks the cached indices invalid and unregisters {@link PinotHelixResourceManager}.
```
mcvsubbu: Please add a block of comment here that describes why syncrhonization is needed on the shutdown variable. 
thaanks.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/HelixHelper.java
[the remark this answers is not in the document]
jackjlli: Put `murmur2` function to `HashUtil` in pinot-common module.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
jackjlli: Refactored the code to use resource config to determine whether the resource is enabled.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
jackjlli: Refactored the code to pass `HelixManager` to the constructor of `LeadControllerManager`. No need to have methods to register/unregister `PinotHelxResourceManager`.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
jackjlli: Added.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -155,21 +160,32 @@ public PinotHelixResourceManager(@Nonnull String zkURL, @Nonnull String helixClu
...
+    this(controllerConf.getZkStr(), controllerConf.getHelixClusterName(), LeadControllerUtils
+            .generateControllerParticipantId(controllerConf.getControllerHost(), controllerConf.getControllerPort()),
+        controllerConf.getDataDir(), controllerConf.getExternalViewOnlineToOfflineTimeout(),
         controllerConf.tenantIsolationEnabled(), controllerConf.getEnableBatchMessageMode(),
-        controllerConf.getHLCTablesAllowed());
+        controllerConf.getHLCTablesAllowed(), instanceType);
+  }
+
+  public PinotHelixResourceManager(@Nonnull ControllerConf controllerConf) {
+    this(controllerConf, InstanceType.ADMINISTRATOR);
```
Jackie-Jiang: Don't change the default behavior, join as PARTICIPANT by default

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -146,7 +150,8 @@
 
   public PinotHelixResourceManager(@Nonnull String zkURL, @Nonnull String helixClusterName,
       @Nonnull String controllerInstanceId, String dataDir, long externalViewOnlineToOfflineTimeoutMillis,
-      boolean isSingleTenantCluster, boolean enableBatchMessageMode, boolean allowHLCTables) {
+      boolean isSingleTenantCluster, boolean enableBatchMessageMode, boolean allowHLCTables,
+      InstanceType helixInstanceType) {
```
Jackie-Jiang: Change this to a boolean joinAsParticipant

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: Use SPECTATOR instead of ADMINISTRATOR because seems SPECTATOR is the standard Helix type for monitoring the cluster

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -155,21 +160,32 @@ public PinotHelixResourceManager(@Nonnull String zkURL, @Nonnull String helixClu
     _isSingleTenantCluster = isSingleTenantCluster;
     _enableBatchMessageMode = enableBatchMessageMode;
     _allowHLCTables = allowHLCTables;
+    _helixInstanceType = helixInstanceType;
+
+    _helixZkManager =
```
Jackie-Jiang: Set up HelixManager and LeadControllerManager in the start()

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Gets lead controller participant id for table from lead controller resource.
+   * If the resource is disabled or no controller registered as participant, there is no instance in "MASTER" state.
+   * @param leadControllerResourceExternalView external view of lead controller resource
+   * @param rawTableName table name without type
+   * @return controller participant id for partition leader, e.g. Controller_localhost_9000. Null if not found or resource is disabled.
+   */
+  public static String getLeadControllerInstanceForTable(ExternalView leadControllerResourceExternalView,
```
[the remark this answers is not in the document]
mcvsubbu: I would like to understand why you prefer that. It is fine to make this class independent of HelixManager.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+   */
+  public static String generateControllerParticipantId(String controllerHost, String controllerPort) {
+    return org.apache.pinot.common.utils.CommonConstants.Helix.PREFIX_OF_CONTROLLER_INSTANCE + controllerHost + "_"
+        + controllerPort;
+  }
+
+  /**
+   * Extracts lead controller hostname and port from controller participant id, e.g. returns localhost_9000 given Controller_localhost_9000 as controller participant id.
+   */
+  public static String extractLeadControllerHostNameAndPort(String controllerParticipantId) {
```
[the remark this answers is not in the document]
Jackie-Jiang: It is very confusing here. For controller, we have **controllerInstanceId** and **participantInstanceId**, where **controllerInstanceId** is formatted as <hostName>_<port>, and **participantInstanceId** is formatted as Controller_<hostName>_<port>. Let's always return **controllerInstanceId** for this class.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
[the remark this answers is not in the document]
Jackie-Jiang: Right now the lead controller instance logic is in two class, and we return `null` for unexpected scenario, which can lead to NPE very easily. We should keep the logic at the same place (it's okay to move this method into ControllerLeaderLocator), and throw exception if the state is unexpected.
Also, please convert the return value from participantInstanceId to controllerInstanceId and make it more explicit from the method name.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Map;
+import java.util.Set;
+import org.apache.helix.model.ExternalView;
+import org.apache.helix.model.MasterSlaveSMD;
+import org.apache.pinot.common.utils.HashUtil;
+import org.apache.pinot.common.utils.StringUtil;
+
+
+public class LeadControllerUtils {
+  public static final String RESOURCE_ENABLED = "RESOURCE_ENABLED";
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Gets hash code for table using murmur2 function, ignores the most significant bit.
+   * Note: This method CANNOT be changed when lead controller resource is enabled.
+   * Otherwise it will assign different controller for the same table, which will mess up the controller periodic tasks and realtime segment completion.
+   * @param rawTableName table name
+   * @return hash code ignoring the most significant bit.
+   */
+  private static int getHashCodeForTable(String rawTableName) {
+    return (HashUtil.murmur2(StringUtil.encodeUtf8(rawTableName)) & Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Given a raw table name and number of partitions, returns the partition id in lead controller resource.
+   * @param rawTableName raw table name
+   * @param numPartitions number of partitions
+   * @return partition id in lead controller resource.
+   */
+  public static int getPartitionIdForTable(String rawTableName, int numPartitions) {
```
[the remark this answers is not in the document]
jackjlli: Updated.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+        return entry.getKey();
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Generates controller participant id, e.g. returns Controller_localhost_9000 given localhost as hostname and 9000 as port.
+   */
+  public static String generateControllerParticipantId(String controllerHost, String controllerPort) {
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+    return org.apache.pinot.common.utils.CommonConstants.Helix.PREFIX_OF_CONTROLLER_INSTANCE + controllerHost + "_"
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+   */
+  public static String extractLeadControllerHostNameAndPort(String controllerParticipantId) {
+    return controllerParticipantId.substring(controllerParticipantId.indexOf("_") + 1);
+  }
+
+  /**
+   * Generates partition name, e.g. returns leadControllerResource_0 given 0 as partition index.
+   */
+  public static String generatePartitionName(int partitionIndex) {
+    return org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME + "_" + partitionIndex;
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+   * Extracts lead controller hostname and port from controller participant id, e.g. returns localhost_9000 given Controller_localhost_9000 as controller participant id.
+   */
+  public static String extractLeadControllerHostNameAndPort(String controllerParticipantId) {
+    return controllerParticipantId.substring(controllerParticipantId.indexOf("_") + 1);
+  }
+
+  /**
+   * Generates partition name, e.g. returns leadControllerResource_0 given 0 as partition index.
+   */
+  public static String generatePartitionName(int partitionIndex) {
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -0,0 +1,111 @@
...
+    return org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME + "_" + partitionIndex;
+  }
+
+  /**
+   * Extracts partition index from partition name, e.g. returns 0 given leadControllerResource_0 as partition name.
+   */
+  public static int extractPartitionIndex(String partitionName) {
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -70,11 +69,14 @@
 import org.apache.pinot.controller.validation.RealtimeSegmentValidationManager;
 import org.apache.pinot.core.crypt.PinotCrypterFactory;
 import org.apache.pinot.core.periodictask.PeriodicTask;
+import org.apache.pinot.core.periodictask.PeriodicTaskScheduler;
 import org.apache.pinot.filesystem.PinotFSFactory;
 import org.glassfish.hk2.utilities.binding.AbstractBinder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME;
```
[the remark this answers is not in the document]

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: Removed the instance type from resource manager and put the logic to the caller instead.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: No need to keep this instance type any more.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: Pass HelixManager as param in start() method.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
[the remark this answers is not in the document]
jackjlli: Updated the code to return the host-port pair instead of a string.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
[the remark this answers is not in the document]
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,222 @@
...
+  private final String _instanceId;
+  private final HelixManager _helixManager;
+  private final ControllerMetrics _controllerMetrics;
+  private final Thread _controllerLeadershipFetchingThread;
+
+  private volatile boolean _isLeadControllerResourceEnabled = false;
+  private volatile boolean _amIHelixLeader = false;
+  private volatile boolean _isShuttingDown = false;
+
+  public LeadControllerManager(HelixManager helixManager, ControllerMetrics controllerMetrics) {
```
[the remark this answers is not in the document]
jackjlli: For Pinot-only mode, `controllerHelixManager` is null. So we shouldn't pass `controllerHelixManager` in here.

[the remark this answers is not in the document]
jackjlli: And I checked the logic of `isLeader()` in HelixManager. The additional check is whether sessionId matches, whereas in Pinot it's unnecessary to check session id for finding a leader for tables.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,222 @@
...
+
+import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY;
+import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME;
+
+
+public class LeadControllerManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(LeadControllerManager.class);
+  private static final long CONTROLLER_LEADERSHIP_FETCH_INTERVAL = 60_000L;
+
+  private final Set<Integer> _partitionIdCache;
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,222 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY;
+import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME;
+
+
+public class LeadControllerManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(LeadControllerManager.class);
+  private static final long CONTROLLER_LEADERSHIP_FETCH_INTERVAL = 60_000L;
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/LeadControllerManager.java
```diff
@@ -0,0 +1,222 @@
...
+import org.apache.helix.model.ResourceConfig;
+import org.apache.pinot.common.config.TableNameBuilder;
+import org.apache.pinot.common.metrics.ControllerMeter;
+import org.apache.pinot.common.metrics.ControllerMetrics;
+import org.apache.pinot.common.utils.CommonConstants;
+import org.apache.pinot.common.utils.helix.LeadControllerUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY;
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -147,6 +150,9 @@ public ControllerStarter(ControllerConf conf) {
     } else {
       _adminApp =
           new ControllerAdminApiApplication(_config.getQueryConsoleWebappPath(), _config.getQueryConsoleUseHttps());
+      // Helix instance type should be explicitly set to PARTICIPANT ONLY in {@link ControllerStarter}.
```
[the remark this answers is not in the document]
jackjlli: Updated.

PR #4326, opened 2019-06-17
[TE] Clean up and inject ThirdEye Credentials into SecurityContext

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auth/ThirdEyeCredentials.java
```diff
@@ -42,15 +51,23 @@ public void setPassword(String password) {
...
     if (this == o) {
       return true;
     }
-    if (!(o instanceof Credentials)) {
+    if (!(o instanceof ThirdEyeCredentials)) {
       return false;
     }
-    Credentials that = (Credentials) o;
+    ThirdEyeCredentials that = (ThirdEyeCredentials) o;
     return Objects.equals(principal, that.principal) && Objects.equals(password, that.password);
```
xiaohui-sun: Add token into comparison.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auth/ThirdEyeCredentials.java
```diff
@@ -22,9 +22,18 @@
 import java.util.Objects;
 
 
-public class Credentials {
+public class ThirdEyeCredentials {
   String principal;
   String password;
+  String token;
```
xiaohui-sun: Is it mainly for service account?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auth/ThirdEyeCredentials.java
akshayrai: Added

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auth/ThirdEyeCredentials.java
akshayrai: Token exists for all the users. This PR only addresses injecting the authenticated user into the `SecurityContext`. That way we can access the user from every api call.

I will send a followup PR to add the service restriction.

PR #4333, opened 2019-06-18
[TE] Entity Yaml (Composite Alert) Translator

akshayrai:
Changes:
- Introduced 2 types of alerts - METRIC_ALERT and COMPOSITE_ALERT (refer yaml examples)
- Modified and refactored the DetectionConfigTranslator to work for composite alerts by recursively calling the metric alert translator

Tests:
- Added unit tests


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigMetricCache.java
```diff
@@ -0,0 +1,105 @@
...
+  private void loadMetricCache(Map<String, Object> metricAlertConfigMap) {
+    String metricName = MapUtils.getString(metricAlertConfigMap, PROP_METRIC);
+    String datasetName = MapUtils.getString(metricAlertConfigMap, PROP_DATASET);
+
+    MetricConfigDTO metricConfig = this.dataProvider.fetchMetric(metricName, datasetName);
+    DatasetConfigDTO datasetConfig = this.dataProvider.fetchDatasets(Collections.singletonList(metricConfig.getDataset()))
+        .get(metricConfig.getDataset());
+    String cron = buildCron(datasetConfig.bucketTimeGranularity());
+
+    metricCache.put(metricName, new DetectionMetricProperties(cron, metricConfig, datasetConfig));
```
jihaozh: looks like the cache key is the `metricName`. What if there are two metrics with the same name but comes from different datasets? Maybe we should use separate Maps to cache dataset and metrics.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigMetricCache.java
```diff
@@ -0,0 +1,105 @@
...
+  MetricConfigDTO fetchMetric(Map<String, Object> metricAlertConfigMap) {
+    String metricName = MapUtils.getString(metricAlertConfigMap, PROP_METRIC);
+    if (!metricCache.containsKey(metricName)) {
+      loadMetricCache(metricAlertConfigMap);
+    }
+
+    return metricCache.get(metricName).getMetricConfigDTO();
+  }
+
+  String fetchCron(Map<String, Object> metricAlertConfigMap) {
```
jihaozh: What are the benefits of caching the cron expression?  Seems it is derived from the default granularity anyways. We can cache the metricConfig and datasetConfig and drive from that. Calling the `buildCron` method should be light.

We can also consider building the cache into the `DataProvider` so that it can be beneficial to more places.  Maybe leverage Guava's `LoadingCache`, it can handle cache expiration, eviction, etc.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigTranslator.java
```diff
@@ -210,16 +212,89 @@ DetectionConfigDTO translateConfig(Map<String, Object> yamlConfigMap) throws Ill
...
+      properties = translateCompositeAlert(yamlConfigMap);
+
+      // TODO: discuss strategy for default cron
+      Preconditions.checkArgument(yamlConfigMap.containsKey(PROP_CRON), "Missing property (" + PROP_CRON + ") in alert");
+      cron = MapUtils.getString(yamlConfigMap, PROP_CRON);
+    } else {
+      // The legacy type 'COMPOSITE' will be treated as a metric alert along with the new convention METRIC_ALERT.
+      // This is applicable only at the root level to maintain backward compatibility.
+      properties = translateMetricAlert(yamlConfigMap);
+      cron = metricCache.fetchCron(yamlConfigMap);
```
jihaozh: If YAML contains cron expression, it should override the default cron here

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigMetricCache.java
akshayrai: Although not implemented yet, the idea behind caching cron is to find the default cron in a composite alert. If we store a mapping of all the cron along with the metric/dataset, then we can potentially derive the default cron for the composite alert.

akshayrai: It would be awesome to have a caching at the `DataProvider` level. We can discuss more about it but would like to keep it outside the scope of this PR.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigMetricCache.java
akshayrai: This is a per translator cache. So it is unlikely that an alert will have 2 same metrics from different datasets. But I think it is safer to use metric and dataset name as the key. Will update it accordingly.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigMetricCache.java
akshayrai: This is not a cache per se. It is mostly a data holder. The datastructure is created during translation and destroyed after that. I will rename this to avoid confusion.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigTranslator.java
akshayrai: Thanks for pointing it out. Fixed it

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigMetricCache.java
```diff
@@ -0,0 +1,105 @@
...
+    DatasetConfigDTO datasetConfig = this.dataProvider.fetchDatasets(Collections.singletonList(metricConfig.getDataset()))
+        .get(metricConfig.getDataset());
+    String cron = buildCron(datasetConfig.bucketTimeGranularity());
+
+    metricCache.put(metricName, new DetectionMetricProperties(cron, metricConfig, datasetConfig));
+  }
+
+  DatasetConfigDTO fetchDataset(Map<String, Object> metricAlertConfigMap) {
+    String metricName = MapUtils.getString(metricAlertConfigMap, PROP_METRIC);
+    if (!metricCache.containsKey(metricName)) {
```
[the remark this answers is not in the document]
akshayrai: Cleaned it up.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigTranslator.java
```diff
@@ -157,18 +157,25 @@
...
+
+
   private static final DetectionRegistry DETECTION_REGISTRY = DetectionRegistry.getInstance();
   static {
     // do not tune for alerts migrated from legacy anomaly function.
     DetectionRegistry.registerComponent("com.linkedin.thirdeye.detection.components.AdLibAlertFilter",
         "MIGRATED_ALGORITHM_FILTER");
     DetectionRegistry.registerComponent("com.linkedin.thirdeye.detection.components.AdLibAnomalyDetector",
         "MIGRATED_ALGORITHM");
+    DetectionRegistry.registerComponent(MockGrouper.class.getName(), "MOCK_GROUPER");
```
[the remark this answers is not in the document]
akshayrai: Added while testing. Cleaned it up.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigMetricCache.java
jihaozh: make sense.

PR #4335, opened 2019-06-18
Refresh schema when reloading segments

haibow:
Now when reloading segments, schema is re-fetched only for OFFLINE table. For realtime tables, it will still reload with the old schema, ignoring new columns recently added.

This diff will reload schemas for REALTIME and OFFLINE tables before reloading segments, as a follow-up on https://github.com/apache/incubator-pinot/issues/4225#issuecomment-502896395

Tested on a few REALTIME (HLC and LLC) tables.

### DIFF
--- a/pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixInstanceDataManager.java
+++ b/pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixInstanceDataManager.java
@@ -180,10 +180,8 @@ public class HelixInstanceDataManager implements InstanceDataManager {
     Preconditions.checkNotNull(tableConfig);
 
     Schema schema = null;
-    // For OFFLINE table, try to get schema for default columns
-    if (TableNameBuilder.OFFLINE.tableHasTypeSuffix(tableNameWithType)) {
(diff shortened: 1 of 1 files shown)


haibow:
@Jackie-Jiang could you help to review?

pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixInstanceDataManager.java
```diff
@@ -180,10 +180,8 @@ public void reloadSegment(@Nonnull String tableNameWithType, @Nonnull String seg
     Preconditions.checkNotNull(tableConfig);
 
     Schema schema = null;
-    // For OFFLINE table, try to get schema for default columns
-    if (TableNameBuilder.OFFLINE.tableHasTypeSuffix(tableNameWithType)) {
-      schema = ZKMetadataProvider.getTableSchema(_propertyStore, tableNameWithType);
-    }
+    // For REALTIME and OFFLINE tables, try to get schema for default columns
```
Jackie-Jiang: (nit) Can you combine 182-184 into one line? The comment seems redundant here.

PR #4337, opened 2019-06-19
Take OFFLINE segment into account for real-time rebalancer

Jackie-Jiang:
When the consuming segment encounters error, the ideal state can be turned into OFFLINE
If the instance states for a segment is all OFFLINE, the segment is counted OFFLINE and won't be rebalanced
RealtimeSegmentValidationManager will periodically detect the OFFLINE segments and re-assign them


Jackie-Jiang:
@snleee Updated the comment and commit message

PR #4338, opened 2019-06-19
Within a partition, only allow querying the first CONSUMING segment for the real-time table routing

Jackie-Jiang:
Within a partition, we only allow querying the first CONSUMING segment (segment with instances in CONSUMING state) for the following reasons:
- Within a partition, there could be multiple CONSUMING segments (typically caused by the delay of CONSUMING to ONLINE state transition). We should only query the first CONSUMING segment because it might contain records that overlapped with the records in the next segment (over-consumed).
- If the instance states for a segment are partial CONSUMING (instances can be ONLINE if they have finished the CONSUMING to ONLINE state transition; instances can be OFFLINE if they encountered error while consuming and controller set the IdealState to OFFLINE; instances can be ERROR if they encountered error during the state transition), we count the segment as CONSUMING segment. If we don't count the segment as CONSUMING segment, then this segment is not allowed to be in the CONSUMING state for routing purpose, and we will not route queries to this segment if there is no ONLINE instances, or route all queries to the ONLINE instances which can potentially overwhelm instances.
- It is possible that the latest CONSUMING segment is not allowed for routing purpose and we won't query it, but it should only last for a short period of time. Once the older CONSUMING segment becomes ONLINE (all instances finished the CONSUMING to ONLINE state transition), the latest CONSUMING segment will become the first CONSUMING segment and will be allowed for routing purpose.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/LowLevelRoutingTableBuilderUtil.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/LowLevelRoutingTableBuilderUtil.java
@@ -37,17 +37,18 @@ public class LowLevelRoutingTableBuilderUtil {
    * state) for the following reasons:
    * <ul>
    *   <li>
-   *     If within a partition, there are multiple CONSUMING segments (typically caused by the delay of CONSUMING to
-   *     ONLINE state transition), we can only query the first CONSUMING segment because it might contain records that
-   *     overlapped with the records in the next segment (over-consumed).
+   *     Within a partition, there could be multiple CONSUMING segments (typically caused by the delay of CONSUMING to
+   *     ONLINE state transition). We should only query the first CONSUMING segment because it might contain records
+   *     that overlapped with the records in the next segment (over-consumed).
    *   </li>
    *   <li>
-   *     If the instance states for a segment is partial ONLINE and partial CONSUMING (some instances finished the
-   *     CONSUMING to ONLINE state transition, others didn't), we count the segment as CONSUMING segment (most likely
-   *     the first CONSUMING segment because it is already committed and is performing the CONSUMING to ONLINE state
-   *     transition). If we don't count the segment as CONSUMING segment, then this segment is not allowed to be in the
-   *     CONSUMING state for routing purpose, and we will route all queries to the ONLINE instances which can
-   *     potentially overwhelm instance.
+   *     If the instance states for a segment are partial CONSUMING (instances can be ONLINE if they have finished
+   *     the CONSUMING to ONLINE state transition; instances can be OFFLINE if they encountered error while consuming
+   *     and controller set the IdealState to OFFLINE; instances can be ERROR if they encountered error during the state
+   *     transition), we count the segment as CONSUMING segment. If we don't count the segment as CONSUMING segment,
+   *     then this segment is not allowed to be in the CONSUMING state for routing purpose, and we will not route
+   *     queries to this segment if there is no ONLINE instances, or route all queries to the ONLINE instances which can
+   *     potentially overwhelm instances.
    *   </li>
    *   <li>
    *     It is possible that the latest CONSUMING segment is not allowed for routing purpose and we won't query it, but
@@ -57,9 +58,9 @@ public class LowLevelRoutingTableBuilderUtil {
    *   </li>
    * </ul>
    *
-   * @param externalView helix external view
-   * @param sortedSegmentsByPartition map of partition to sorted set of segment names.
-   * @return map of allowed consuming segment for each partition for routing.
+   * @param externalView External view for the real-time table
+   * @param sortedSegmentsByPartition Map from partition to segments
+   * @return Map from partition to allowed CONSUMING segment for routing purpose
    */
   public static Map<String, SegmentName> getAllowedConsumingStateSegments(ExternalView externalView,
       Map<String, SortedSet<SegmentName>> sortedSegmentsByPartition) {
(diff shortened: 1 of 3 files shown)


mayankshriv:
Not sure if I follow the description. If for a given partition, there are 'n' replicas that are all in consuming  state, will the first instance get all the QPS? If yes, that will cause a huge spike for one instance (I am sure I am missing something here). If not, what does 'allow' mean?

Jackie-Jiang:
> Not sure if I follow the description. If for a given partition, there are 'n' replicas that are all in consuming state, will the first instance get all the QPS? If yes, that will cause a huge spike for one instance (I am sure I am missing something here). If not, what does 'allow' mean?

When a segment finishes consumption and the next segment starts consuming, there could be a short period of time where the ideal state looks like the follows:
```
-------- Server1   Server2   Server3
Segment1 ONLINE    CONSUMING CONSUMING
Segment2 CONSUMING CONSUMING CONSUMING
```
The current algorithm will allow only the segment2 (last one) in consuming state, which means Segment1 will only have one replica (Server1) up as Server2 and Server3 are in CONSUMING state. This is causing Server1 becoming the hotspot.

The new algorithm will allow only the segment1 in consuming state, which means Segment2 won't be queried.

kishoreg:
I can’t follow. Can you review your comment again and see if there is any typo.

Jackie-Jiang:
> I can’t follow. Can you review your comment again and see if there is any typo.

Rephrased the comment

kishoreg:
Got it. Good idea. One concern with this is if one replica of a segment stays in CONSUMING state for ever, then we will never query the new consuming segments. 

@npawar does validation manager handle take care of segments stuck in consuming state?

Jackie-Jiang:
> Got it. Good idea. One concern with this is if one replica of a segment stays in CONSUMING state for ever, then we will never query the new consuming segments.
> 
> @npawar does validation manager handle take care of segments stuck in consuming state?

In ideal state, if any instance is ONLINE for the segment, then it is guaranteed that all instances are set to ONLINE for the segment (in contrast to CONSUMING where there can be mixed CONSUMING and OFFLINE for the same segment).
If state transition failed, then the external view will turn into ERROR, and the current logic will shift to the next consuming segment.

pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/LowLevelRoutingTableBuilderUtil.java
```diff
@@ -32,59 +32,49 @@
...
-              .equalsIgnoreCase(CommonConstants.Helix.StateModel.RealtimeSegmentOnlineOfflineStateModel.CONSUMING)) {
-            replicasInConsumingState++;
-          }
-        }
-
-        // If all replicas have this segment in consuming state (and not all of them are in ERROR state), then pick this
-        // segment to be the last allowed segment to be in CONSUMING state
-        if (allInConsumingState && 0 < replicasInConsumingState) {
-          lastAllowedSegmentInConsumingState = segmentName;
+        if (externalView.getStateMap(segmentName.getSegmentName())
```
mcvsubbu: A revers iterator, if possible, will get us here faster most of the time? the logic may be a bit complex...

Jackie-Jiang: The current implementation has lots of redundancy. I have plan of re-implementing the routing table builder to handle new coming segment to avoid hotspot instance. For now, I'll just keep the change relatively small.

Jackie-Jiang:
@snleee @mcvsubbu Added unit test

pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/LowLevelRoutingTableBuilderUtil.java
```diff
@@ -32,59 +32,49 @@
 public class LowLevelRoutingTableBuilderUtil {
 
   /**
-   * Compute the map of allowed 'consuming' segments for each partition.
+   * Compute the map of allowed CONSUMING segments for each partition for routing purpose.
+   * <p>Within a partition, we only allow querying the first CONSUMING segment (segment with instances in CONSUMING
+   * state) for the following reasons:
+   * <ul>
+   *   <li>
+   *     If within a partition, there are multiple CONSUMING segments (typically caused by the delay of CONSUMING to
```
jackjlli: Within a partition, there could be multiple CONSUMING segments (typically caused by the delay of CONSUMING to ONLINE state transition). We should only allow querying the first CONSUMING segment because it might contain records that overlapped with the records in the next segment (over-consumed).

pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/LowLevelRoutingTableBuilderUtil.java
```diff
@@ -32,59 +32,49 @@
...
+   *   </li>
+   *   <li>
+   *     It is possible that the latest CONSUMING segment is not allowed for routing purpose and we won't query it, but
+   *     it should only last for a short period of time. Once the older CONSUMING segment becomes ONLINE (all instances
+   *     finished the CONSUMING to ONLINE state transition), the latest CONSUMING segment will become the first
+   *     CONSUMING segment and will be allowed for routing purpose.
+   *   </li>
+   * </ul>
    *
    * @param externalView helix external view
```
jackjlli: external view of realtime table.

pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/LowLevelRoutingTableBuilderUtil.java
Jackie-Jiang: Done

pinot-broker/src/main/java/org/apache/pinot/broker/routing/builder/LowLevelRoutingTableBuilderUtil.java
Jackie-Jiang: Done

Jackie-Jiang:
> thx for adding unit test.
> please add a TODO for other combinations (e.g. happy path of N segments online and 1 segment CONSUMING), introduce the last-but-one segment is OFFLINE or ERROR states, etc.

Updated comments to cover the cases

PR #4339, opened 2019-06-19
Only set Helix cluster properties when creating the cluster

Jackie-Jiang:
If Helix cluster does not exist, create the cluster with auto-join enabled
Do not set the STATE_TRANSITION.maxThreads when creating the cluster so Helix can go with the default value (40)

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
@@ -19,10 +19,9 @@
 package org.apache.pinot.controller.helix.core.util;
 
 import com.google.common.base.Preconditions;
-import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


Jackie-Jiang:
> Should we make `maxParallelStateTransition` configurable?
> 
> @jackjlli Can you comment if there was a particular reason that you set this to 1?

We can make it configurable, but in most of the cases Helix default (40) should be good enough.
The 1 is there in the legacy code. But before Jack's change, the 1 is applied only when we create a new cluster.

Jackie-Jiang:
@mcvsubbu The auto-join the number of threads should already be configured in the Helix cluster before Pinot controller joins. We cannot reset the value every time when pinot controller becomes alive, it should be a one time setup.

jackjlli:
@Jackie-Jiang The existing code will check whether auto-join is enabled or not. It will set it **only when** auto-join hasn't been enabled. So it will only be set once.
How do we make sure if it's enabled when the helix cluster is created by other code base instead of Pinot code base?

Jackie-Jiang:
@jackjlli When someone creates a cluster for Pinot, they can decide whether to allow auto-join or not, and how many threads to use for state transitions. Pinot should not override these properties, and that is exactly why we need this fix. Current code blindly overrides the number threads to 1, which is causing problem.

PR #4346, opened 2019-06-20
Adding support for udfs in Selection

kishoreg:
We currently support expressions and udf's in aggregation and group by but not in selection and filters. This PR adds support for expressions in selection.

Now we can run queries as follows

```
SELECT DaysSinceEpoch, timeConvert(DaysSinceEpoch,'DAYS','SECONDS') FROM mytable

SELECT DaysSinceEpoch, timeConvert(DaysSinceEpoch,'DAYS','SECONDS') FROM mytable order by DaysSinceEpoch limit 10000

SELECT DaysSinceEpoch, timeConvert(DaysSinceEpoch,'DAYS','SECONDS') FROM mytable order by timeConvert(DaysSinceEpoch,'DAYS','SECONDS') DESC limit 10000
```
I haven't deleted the old SelectionOnlyOperator and SelectionOrderByOperator yet.


mcvsubbu:
Please also update the documentation. thanks

mayankshriv:
Thanks for implementing this, will review it today.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformBlockDataFetcher.java
```diff
@@ -0,0 +1,340 @@
...
+        case BYTES:
+          return new SVBytesValueFetcher(blockValSet.getBytesValuesSV());
+      }
+    }
+    throw new UnsupportedOperationException();
+  }
+}
+
+
+interface Fetcher {
```
mayankshriv: Is there a reason why we cannot use the DataFetcher that already exists?

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformOperator.java
```diff
@@ -51,11 +54,13 @@
    * @param expressions Set of expressions to evaluate
    */
   public TransformOperator(@Nonnull ProjectionOperator projectionOperator,
-      @Nonnull Set<TransformExpressionTree> expressions) {
+      @Nonnull List<TransformExpressionTree> expressions) {
```
mayankshriv: Does the existing implementation break due to use of Set (ie when transforms are duplicated)? If not, what's the reason to change to List?

pinot-core/src/main/java/org/apache/pinot/core/plan/SelectionPlanNode.java
```diff
@@ -24,52 +24,65 @@
...
     _indexSegment = indexSegment;
     _selection = brokerRequest.getSelections();
-
-    if (_selection.getSize() > 0) {
-      int maxDocPerNextCall = DocIdSetPlanNode.MAX_DOC_PER_CALL;
-
-      // No ordering required, select minimum number of documents
-      if (!_selection.isSetSelectionSortSequence()) {
-        maxDocPerNextCall = Math.min(_selection.getOffset() + _selection.getSize(), maxDocPerNextCall);
+    if (enableUDFInSelection) {
```
mayankshriv: This can be removed, since it is hard-coded to be true?

pinot-common/src/main/java/org/apache/pinot/common/function/FunctionDefinitionRegistry.java
```diff
@@ -0,0 +1,36 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Place where all functions are registered.
+ */
+public class FunctionDefinitionRegistry {
+
+  public static boolean isAggFunc(String functionName) {
```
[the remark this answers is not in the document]
kishoreg: yes, but that means moving all the transform functions along with implementations to pinot-common.  we should probably define the registry in common and let the server register udfs at the startup time. Didn't want to do that as part of this PR

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -127,11 +128,21 @@ private void convertSelectList(PinotQuery pinotQuery, BrokerRequest brokerReques
           selection.addToSelectionColumns(expression.getIdentifier().getName());
           break;
         case FUNCTION:
-          AggregationInfo aggInfo = buildAggregationInfo(expression.getFunctionCall());
-          if (aggregationInfoList == null) {
-            aggregationInfoList = new ArrayList<>();
+
+          Function functionCall = expression.getFunctionCall();
+          String functionName = functionCall.getOperator();
+          if(FunctionDefinitionRegistry.isAggFunc(functionName)) {
```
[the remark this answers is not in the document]
kishoreg: Done.

pinot-core/src/main/java/org/apache/pinot/core/operator/query/SelectionOperator.java
```diff
@@ -0,0 +1,268 @@
...
+        _blockValSets[i] = transformBlock.getBlockValueSet(expression);
+      }
+      TransformBlockDataFetcher dataFetcher;
+      dataFetcher = new TransformBlockDataFetcher(_blockValSets, _dictionaries, _expressionResultMetadata);
+      if (selectionOnly) {
+        int docId = 0;
+        while (_rowEvents.size() < _limit && docId < transformBlock.getNumDocs()) {
+          Serializable[] row = dataFetcher.getRow(docId);
+          _rowEvents.add(row);
+          docId = docId + 1;
```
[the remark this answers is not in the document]
kishoreg: Habit :)

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformBlockDataFetcher.java
kishoreg: No, DataFetcher works on top of DataSources. Thought about refactoring the existing one to make it work for both use cases but we can do this in another PR.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformOperator.java
kishoreg: Yes, I remember that maintaining order was important. Can't remember why. I might have fixed that. Will look at the code again.

kishoreg: Its not important in terms of functionality but it matches the order in which we display the results for selection query(maintains the order in which columns appear in the query)

pinot-core/src/main/java/org/apache/pinot/core/plan/SelectionPlanNode.java
kishoreg: Done. Removed it along with old classes

mayankshriv:
Thanks for addressing the comments @kishoreg. Seems like there's tests failing, please address them. Also, would be good if @Jackie-Jiang can review as well.

kishoreg:
> Please also update the documentation. thanks

Will do this in another PR along with the filter udf

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
```diff
@@ -81,15 +93,36 @@ private void extractColumnsAndTransforms(@Nonnull BrokerRequest brokerRequest) {
         }
       }
     } else {
-      throw new UnsupportedOperationException("Transforms not supported in selection queries.");
-      // TODO: Add transform support.
-      // projectionColumns.addAll(brokerRequest.getSelections().getSelectionColumns());
+//      throw new UnsupportedOperationException("Transforms not supported in selection queries.");
```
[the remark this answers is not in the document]
kishoreg: Done!

PR #4348, opened 2019-06-20
A template implementation for VarLengthBytesValueReaderWriter

mcvsubbu:
This will not be checked in, but do review and give me comments

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
@@ -1,3 +1,21 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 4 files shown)


kishoreg:
Thanks Subbu.

private Buffer(long size, int startIndex, PinotDataBufferMemoryManager memoryManager, String allocationContext) {

Move this constructor out of this class and move it to a separate class.
Refactor the existing MutableOffHeapByteArrayStore to use this class.

Go ahead and check this in after that.

mcvsubbu:
@kishoreg does this help? @mayankshriv Do you have additional comments?

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private int getHeaderSize(byte[][] arrays) {
+    return 11;  // or whatever computed value for fixed/var sized headers, version etc..
+  }
+
+  private void writeHeader(PinotDataBuffer pinotDataBuffer, byte[][] arrays) {
+    // write headers here, maybe header has some metadata about values.
+  }
+
+  private PinotDataBuffer readHedader(PinotDataBuffer pinotDataBuffer) {
```
mayankshriv: Typo in name `s/readHedader/readHeader`

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/OffHeapByteArrayStore.java
```diff
@@ -0,0 +1,205 @@
...
+   * This constructor is useful when the values are not known ahead of time. If the caller uses this
+   * constructor, then it is expected that the caller invokes {@link #add(byte[])} method until it returns -1
+   * at which point it is known that no more strings can be added to the buffer.
+   *
+   * @param size
+   * @param memoryManager
+   * @param allocationContext
+   */
+  public OffHeapByteArrayStore(long size, PinotDataBufferMemoryManager memoryManager, String allocationContext) {
+    if (size >= Integer.MAX_VALUE) {
```
buchireddy: @mcvsubbu Though remote case, should we fail to create the buffer rather than silently trimming the size to `MAX_VALUE - 1`? Maybe making the size itself an `int` might make this explicit.

npawar:

PR #4349, opened 2019-06-20
Refresh schema when reloading all segments

haibow:
As a follow-up of #4335. Also refresh schema when reloading all segments.


haibow:
@Jackie-Jiang could you help to review?

PR #4350, opened 2019-06-20
For RANGE and REGEXP operators, if there is single matching dictionary id, use inverted index

Jackie-Jiang:
This is especially useful for time column with DAYS granularity and time values across two days
This optimization can apply to most hybrid use cases

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/filter/FilterOperatorUtils.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/filter/FilterOperatorUtils.java
@@ -53,18 +53,25 @@ public class FilterOperatorUtils {
     // TODO: make it exclusive
     int endDocId = numDocs - 1;
 
-    // Use inverted index if the predicate type is not RANGE or REGEXP_LIKE for efficiency
+    // Use scan-based operator if inverted index does not exist or the predicate type is RANGE or REGEXP_LIKE with more
+    // than 1 matching dictionary ids
+    // NOTE: allow RANGE with single matching dictionary id to use inverted index is very useful for time column with
+    //       DAYS granularity and time values across two days
+    // TODO: whether to use inverted index should be based on the number of matching dictionary ids and cardinality of
+    //       the column instead of the predicate type
+    // TODO: if column is sorted, should always use sorted index for RANGE predicate
     DataSourceMetadata dataSourceMetadata = dataSource.getDataSourceMetadata();
     Predicate.Type predicateType = predicateEvaluator.getPredicateType();
-    if (dataSourceMetadata.hasInvertedIndex() && (predicateType != Predicate.Type.RANGE) && (predicateType
-        != Predicate.Type.REGEXP_LIKE)) {
+    if (!dataSourceMetadata.hasInvertedIndex() || (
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/org/apache/pinot/core/operator/filter/FilterOperatorUtils.java
```diff
@@ -53,18 +53,25 @@ public static BaseFilterOperator getLeafFilterOperator(PredicateEvaluator predic
...
+    // TODO: whether to use inverted index should be based on the number of matching dictionary ids and cardinality of
+    //       the column instead of the predicate type
+    // TODO: if column is sorted, should always use sorted index for RANGE predicate
     DataSourceMetadata dataSourceMetadata = dataSource.getDataSourceMetadata();
     Predicate.Type predicateType = predicateEvaluator.getPredicateType();
-    if (dataSourceMetadata.hasInvertedIndex() && (predicateType != Predicate.Type.RANGE) && (predicateType
-        != Predicate.Type.REGEXP_LIKE)) {
+    if (!dataSourceMetadata.hasInvertedIndex() || (
+        (predicateType == Predicate.Type.RANGE || predicateType == Predicate.Type.REGEXP_LIKE)
+            && predicateEvaluator.getNumMatchingDictIds() > 1)) {
```
sunithabeeram: I think this can be higher (atleast for range) even for this version. We currently support IN clause for a LOT of values and after the roaring bit-map upgrade, the performance penalty is much lower. I would atleast make this configurable so that it can be updated easily (perhaps via an external or internal tuner).

Jackie-Jiang: Agree, but I want to keep this pr small and only solve the time column issue for now. I added TODOs and want to revisit this part later. The current way of handling IN can also backfire. The filter should be much smarter. Created #4351 to track this.

kishoreg:
I agree with Sunitha. What is the scenario where this helps?

Jackie-Jiang:
@kishoreg This one will help almost all hybrid use cases with DAYS granularity time column.
Broker will attach a time filter to both offline and realtime part. If the whole segment is within a day (same time column value), then the filter will match all/nothing, which is fine. If the segment is across two days, and the filter is at the boundary of two days, without this improvement we have to scan the time column (half of the values will match in average); with this improvement we just need to fetch a bitmap.

kishoreg:
we can easily extend that logic. We have N rows and D dictionary Ids and M
matching dictionary.

Create a Cost function that takes these three params. It’s ok if that
function is just uses M = 1 for now. But at least the code looks better and
it’s clear that we can try a few implementations.

Jackie-Jiang:
@kishoreg The purpose of this PR is to address the most common case and optimize a very important production use case. I left TODOs and created an issue and will revisit to create the cost function in another PR.

sunithabeeram:
@Jackie-Jiang, regarding:
>The purpose of this PR is to address the most common case and optimize a very important production use case.

IMHO, I don't think this is the critical issue for the prod usecase being referred to at this time. A change in how we handle RANGE/IN predicates has a broader impact, so it would be great if we can do this right. This is something that would fit nicely in the scope of things @jasperjiaguo is doing.

Jackie-Jiang:
Close this one and will make a complete optimization later

Issue #4007, opened 2019-03-22
Improve/Add completeness and freshness metrics for Pinot


sunithabeeram:
This issue tracks the need to expose better metrics in Pinot to expose completeness and freshness in the query path. Today, a lot of individual metrics are available for tracking, but its typically hard to say which queries were actually affected when issues do happen. Allowing tracking/metadata to be sent in query response helps with this. This issue tracks efforts being made for this. 
NOTE: the tracking is on a best effort basis - we would like to track the most common possibilities (without tying ourselves to specific individual scenarios) without compromising the execution time of the query and continuing to keep the design/implementation as simple/intuitive as possible. 

Details:
======
Completeness:
-------------------
Currently, we set a "partial-response" metadata flag when a broker is unable to get a connection to the server or response from a server within configured timeout. We would like to expand this to cover:
(a) cases where servers don't have the segments the broker is requesting.
(b) broker's routing table is incomplete - for ex, when all the replicas of a given segment are ofline or in error state.

Freshness:
---------------
In the current scope, we limit the freshness of the data to realtime tables. We would like to measure how "recent" the data available for querying in realtime is. 

More details for each of these will follow.

sunithabeeram:
https://github.com/apache/incubator-pinot/pull/3942 tracks Completeness (a) part of the issue. Note the following caveats:
- There is a legitimate case where the servers may not have segments available - this happens during retention. We don't consider to be a serving error. The change makes a best effort attempt to track recently deleted segments using an in-memory cache. This does mean that on server restarts the segments retained out segments can be marked as missing.
- When segments move or get refreshed with explicit state transitions (ie, an online->offline transition is triggered followed by a offline->online transition), and the query reaches a server after the online->offline transition), the server won't have the segment and we will count them as partial responses.

sunithabeeram:
More formal discussion around tracking Freshness can be found here: https://cwiki.apache.org/confluence/display/PINOT/Pinot+Freshness+Metric

hpvd:
Freshness is an really important topic several usecases. I really like @sunithabeeram proposal mentioned above.
If there is no metric its 
-  really hard to make decisions on this e.g. for deciding if we can always rely on data from inside Pinot or if have to craft a event sourcing solution for some parts
- check the result of automations in field usage

Are  you aware of any work in this field not yet mentioned in this issue?

hpvd:
would be also important to ad something like last ingest /not ingested for xxx ms to the metrics to easily find things like
[...]

PR #4356, opened 2019-06-24
Emit metrics to track performance against specific Service-Level-Objectives (SLO)

sunithabeeram:
This change builds on issue #4007. The main contribution of this PR is to emit metrics that can track performance of queries against a table for a given set of service-level objectives. We break the tracking into 5 categories: responses that are errors, responses that have partial-data, responses that don't meet latency objectives (latentResponses), responses that don't meet freshness objectives (staleResponses) - all others are considered "healthy", as in - they meet the SLO. The availability of the system can then be measured as: (healthyResponses)/(healthyResponses+partialResponses+errorResponses+latentResponses+staleResponses). This PR does not address the availability calculation as this can be computed outside of the system - and can be computed by assigning differnt weights for different categories. For e.g., the above equation assigns 0 healthyWeight for partialResponses - however, operators can choose to penalize partialResponses lesser and so on.

The SLO are now part of table-config that the broker fetches in order to emit the metrics natively.

Changes in this PR include:
- Updates to TableConfig to expose SloConfig
- Support a Lazy-loading Table-Config on the broker side to allow emitting the metrics.
- Categorize different scenarios into the specified response type classifications.
- Name change:  numConsumingSegmentsQueried -> numConsumingSegmentsProcessed

Testing done:
- Basic unit tests to cover newly added functionality
- Manual integration tests to cover metric emission and validations.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -40,6 +40,8 @@ import org.apache.pinot.broker.queryquota.QueryQuotaManager;
 import org.apache.pinot.broker.routing.RoutingTable;
 import org.apache.pinot.broker.routing.RoutingTableLookupRequest;
 import org.apache.pinot.broker.routing.TimeBoundaryService;
+import org.apache.pinot.common.config.SloConfig;
+import org.apache.pinot.common.config.TableConfig;
 import org.apache.pinot.common.config.TableNameBuilder;
 import org.apache.pinot.common.exception.QueryException;
 import org.apache.pinot.common.metrics.BrokerMeter;
@@ -73,6 +75,7 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   protected final TimeBoundaryService _timeBoundaryService;
   protected final AccessControlFactory _accessControlFactory;
   protected final QueryQuotaManager _queryQuotaManager;
+  protected final LazyTableConfigCache _tableConfigCache;
   protected final BrokerMetrics _brokerMetrics;
 
   protected final AtomicLong _requestIdGenerator = new AtomicLong();
@@ -90,12 +93,13 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
 
   public BaseBrokerRequestHandler(Configuration config, RoutingTable routingTable,
       TimeBoundaryService timeBoundaryService, AccessControlFactory accessControlFactory,
-      QueryQuotaManager queryQuotaManager, BrokerMetrics brokerMetrics) {
+      QueryQuotaManager queryQuotaManager, LazyTableConfigCache cache, BrokerMetrics brokerMetrics) {
     _config = config;
     _routingTable = routingTable;
     _timeBoundaryService = timeBoundaryService;
     _accessControlFactory = accessControlFactory;
     _queryQuotaManager = queryQuotaManager;
+    _tableConfigCache = cache;
     _brokerMetrics = brokerMetrics;
 
     _brokerId = config.getString(CONFIG_OF_BROKER_ID, getDefaultBrokerId());
@@ -125,6 +129,15 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   @SuppressWarnings("Duplicates")
   @Override
   public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentity requesterIdentity,
+      RequestStatistics requestStatistics)
+      throws Exception {
+
+    BrokerResponse response = _handleRequest(request, requesterIdentity, requestStatistics);
+    scoreResponse(response, requestStatistics);
+    return response;
+  }
+
+  private BrokerResponse _handleRequest(JsonNode request, @Nullable RequesterIdentity requesterIdentity,
       RequestStatistics requestStatistics)
       throws Exception {
     long requestId = _requestIdGenerator.incrementAndGet();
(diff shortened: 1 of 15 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -361,6 +384,52 @@ private boolean forceLog(BrokerResponse brokerResponse, long totalTimeMs) {
     return false;
   }
 
+
+  private void scoreResponse(BrokerResponse response, RequestStatistics stats) {
+
+    try {
+
+      if (response == BrokerResponseNative.EMPTY_RESULT || response == BrokerResponseNative.NO_TABLE_RESULT
```
mcvsubbu: No-table is the result of the user sending query to a non-existent table? Do you want to count that as an error response?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -361,6 +384,52 @@ private boolean forceLog(BrokerResponse brokerResponse, long totalTimeMs) {
...
+          || response.getNumServersResponded() == 0 || !response.getProcessingExceptions().isEmpty()) {
+        // error response
+        _brokerMetrics.addMeteredTableValue(stats.getTableName(), BrokerMeter.ERROR_RESPONSE, 1);
+        return;
+      }
+
+      if ((response.getNumServersQueried() > response.getNumServersResponded())
+          || stats.getFanoutType() == RequestStatistics.FanoutType.HYBRID
+          && response.getNumConsumingSegmentsQueried() <= 0 || response.isNumGroupsLimitReached()) {
```
mcvsubbu: is it worth noting here in a comment that it is possible that due to helix timings, the externalview may contain (at least theoretically) no consuming segments, and we will count that as a partial response even though the others are in the process of getting created?
I am not sure if it is useful to add a log only for the case when we have realtime tables but queries 0 consuming segments.
Lastly, does response.getNumConsumingSegmentsQueried() work correctly with HLC?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/LazyTableConfigCache.java
```diff
@@ -0,0 +1,105 @@
...
+        .build(new CacheLoader<String, TableConfig>() {
+          @Override
+          public TableConfig load(@Nonnull String rawTableName) {
+            return ZKMetadataProvider.getTableConfig(_propertyStore, rawTableName);
+          }
+
+          @Override
+          public ListenableFuture<TableConfig> reload(String tableName, TableConfig oldValue) {
+            ListenableFutureTask<TableConfig> task =
+                ListenableFutureTask.create(() -> ZKMetadataProvider.getTableConfig(_propertyStore, tableName));
```
mcvsubbu: this call takes a typed table name, not raw table name. We store configs for typed table names, not on raw table name

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -361,6 +384,52 @@ private boolean forceLog(BrokerResponse brokerResponse, long totalTimeMs) {
...
+        // partial response
+        _brokerMetrics.addMeteredTableValue(stats.getTableName(), BrokerMeter.PARTIAL_RESPONSE, 1);
+        return;
+      }
+
+      TableConfig tableConfig = _tableConfigCache.get(stats.getTableName());
```
mcvsubbu: note: stats.getTableName() gets raw table name. Which table config are you expecting to process below? I suppose realtime? I would imagine that freshness (in general) will also apply to offline tables. We are not impleemnting that now, I suppose?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -361,6 +384,52 @@ private boolean forceLog(BrokerResponse brokerResponse, long totalTimeMs) {
...
+            || stats.getFanoutType() == RequestStatistics.FanoutType.REALTIME)
+            && response.getMinConsumingFreshnessTimeMs() > 0 && (System.currentTimeMillis() - response.getMinConsumingFreshnessTimeMs() > sloConfig.getFreshnessLagMs())) {
+          // stale response
+          _brokerMetrics.addMeteredTableValue(stats.getTableName(), BrokerMeter.STALE_RESPONSE, 1);
+          return;
+        }
+
+        if (sloConfig.getLatencyMs() > 0 && stats.getProcessingTimeMillis() > sloConfig.getLatencyMs()) {
+          // latent response
+          _brokerMetrics.addMeteredTableValue(stats.getTableName(), BrokerMeter.LATENT_RESPONSE, 1);
```
mcvsubbu: LATE_RESPONSE? DELAYED_RESPONSE?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -361,6 +384,52 @@ private boolean forceLog(BrokerResponse brokerResponse, long totalTimeMs) {
...
+          return;
+        }
+      }
+
+      _brokerMetrics.addMeteredTableValue(stats.getTableName(), BrokerMeter.HEALTHY_RESPONSE, 1);
+    } catch (Exception e) {
+      // don't fail the request due to any issues related to tracking!
+      LOGGER.error("Failed to score request with id={} table={} cache={}", stats.getRequestId(), stats.getTableName(),
```
mcvsubbu: I can understand why you want to log here, but we need to be careful not to flood the logs.... You may want to consider the alternative of raising a metric? Or logging once every N requests? Not sure though.

pinot-common/src/main/java/org/apache/pinot/common/config/SloConfig.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Section of the table config that provides service-level objectives for specific dimensions such as
+ * latency, freshness.
+ *
+ * Note that analytical query latencies can be highly variable - the SLO should include the maximum tolerable latency
+ * for servicing the common, high latency queries.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class SloConfig {
```
mcvsubbu: SLOConfig?

pinot-common/src/main/java/org/apache/pinot/common/config/SloConfig.java
```diff
@@ -0,0 +1,117 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(SloConfig.class);
+
+  @ConfigKey("latencyMs")
+  @ConfigDoc("Latency SLO in milliseconds for queries of this table")
+  private long latencyMs;
+
+  @ConfigKey("freshnessLagMs")
+  @ConfigDoc("Freshness Lag SLO in milliseconds for queries of this table (realtime only)")
+  private long freshnessLagMs;
```
mcvsubbu: Consider allowing the user to specify freshness lag in terms of a string such as : "15m" or "1h". See TimeUtils.java

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
sunithabeeram: Non-existent table issue is caught earlier with the auth check, so I think this should be OK. This can also happen when we have a valid table, but EV doesn't reflect it. A signal on the second case seems important than a false alarm on the first. What do you think?

pinot-common/src/main/java/org/apache/pinot/common/config/SloConfig.java
sunithabeeram: Will do

sunithabeeram:
Closing this as the logic has been moved to higher layers.

PR #4359, opened 2019-06-24
[TE] Tuning of Composite alert - Store metric refs in component spec

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/translator/DetectionConfigTranslator.java
```diff
@@ -470,6 +471,9 @@ private void buildComponentSpec(Map<String, Object> yamlConfig, String type, Str
 
     String componentClassName = DETECTION_REGISTRY.lookup(type);
     componentSpecs.put(PROP_CLASS_NAME, componentClassName);
+    componentSpecs.put(PROP_METRIC, MapUtils.getString(yamlConfig, PROP_METRIC));
```
jihaozh: Where is this used? How about store the metric urn?

akshayrai: It is required to fetch the timezone info from the datasetConfig

jihaozh: Make sense. Since the composite alert can have different metrics at different levels.

PR #4360, opened 2019-06-24
Adding util methods for controller that will be used in offline push

jenniferdai:
-Adding util methods to support segment delete after push
-Split because it is necessary to refactor segment naming code and internal code that currently uses the same controller rest api interface
-Final Design: https://github.com/apache/incubator-pinot/issues/4353

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/FileUploadDownloadClient.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/FileUploadDownloadClient.java
@@ -98,7 +98,6 @@ public class FileUploadDownloadClient implements Closeable {
   private static final String SCHEMA_PATH = "/schemas";
   private static final String OLD_SEGMENT_PATH = "/segments";
   private static final String SEGMENT_PATH = "/v2/segments";
-  private static final String DELETE_SEGMENT_PATH = "/segments";
   private static final String SEGMENT_METADATA_PATH = "/segmentmetadata";
   private static final String TABLES_PATH = "/tables";
   private static final String TYPE_DELIMITER = "?type=";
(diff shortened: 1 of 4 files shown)


pinot-common/src/main/java/org/apache/pinot/common/utils/FileUploadDownloadClient.java
```diff
@@ -89,12 +91,14 @@ public static FileUploadType getDefaultUploadType() {
...
   public static final int DEFAULT_SOCKET_TIMEOUT_MS = 600 * 1000; // 10 minutes
   public static final int GET_REQUEST_SOCKET_TIMEOUT_MS = 5 * 1000; // 5 seconds
+  public static final int DELETE_REQUEST_SOCKET_TIMEOUT_MS = 10 * 1000; // 10 seconds
 
   private static final String HTTP = "http";
   private static final String HTTPS = "https";
   private static final String SCHEMA_PATH = "/schemas";
   private static final String OLD_SEGMENT_PATH = "/segments";
   private static final String SEGMENT_PATH = "/v2/segments";
+  private static final String DELETE_SEGMENT_PATH = "/segments";
```
jackjlli: Can we call it `SEGMENT_PATH` instead of `DELETE_SEGMENT_PATH`? Or why not just use `OLD_SEGMENT_PATH`?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/DefaultControllerRestApi.java
```diff
@@ -134,9 +134,56 @@ public void sendSegmentUris(List<String> segmentUris) {
...
+          LOGGER.error("Caught exception while deleting segment URI: {} to location: {}", segmentUri, pushLocation, e);
+          throw new RuntimeException(e);
+        }
+      }
+    }
+  }
+
+  @Override
+  public List<String> getAllSegments(String tableType) {
+    LOGGER.info("Getting all segments");
```
jackjlli: `"Getting all segments of Table: {}"`

pinot-common/src/main/java/org/apache/pinot/common/utils/FileUploadDownloadClient.java
jenniferdai: I wanted it to be clear that we are calling the delete API

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/DefaultControllerRestApi.java
jenniferdai: fixed, thanks

pinot-common/src/main/java/org/apache/pinot/common/utils/FileUploadDownloadClient.java
jackjlli: I think it's unnecessary, since you've already specify the behavior in the method name `getDeleteSegmentHttpUri`. Not a good idea to add more and more constants here.

jenniferdai: changed

PR #4361, opened 2019-06-24
Fix to wait until all docs loaded in quick start

jackjlli:
This PR fixes to wait until all docs loaded in quick start.

### DIFF
--- a/pinot-tools/src/main/java/org/apache/pinot/tools/Quickstart.java
+++ b/pinot-tools/src/main/java/org/apache/pinot/tools/Quickstart.java
@@ -24,6 +24,8 @@ import com.google.common.collect.Lists;
 import java.io.File;
 import java.net.URL;
 import org.apache.commons.io.FileUtils;
+import org.apache.pinot.common.utils.retry.RetryPolicies;
+import org.apache.pinot.common.utils.retry.RetryPolicy;
 import org.apache.pinot.core.data.readers.FileFormat;
 import org.apache.pinot.tools.admin.command.QuickstartRunner;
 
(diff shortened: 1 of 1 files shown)


jackjlli:
@kishoreg The code is in `pinot-tool` module, and we don't have any test in `pinot-tool` though.

kishoreg:
this is not the right fix. It just masking NPE. The real problem is in the following lines
```    
printStatus(Color.CYAN, "***** Waiting for 5 seconds for the server to fetch the assigned segment *****");   
Thread.sleep(5000);
```
Similar to what we have in integration tests, we need to wait until the segments are loaded and select count(*) returns the right number.

jackjlli:
@kishoreg Code updated

pinot-tools/src/main/java/org/apache/pinot/tools/Quickstart.java
```diff
@@ -209,6 +212,35 @@ public void run() {
     printStatus(Color.GREEN, "You can always go to http://localhost:9000/query/ to play around in the query console");
   }
 
+  protected void waitForAllDocsLoaded(QuickstartRunner runner, long timeoutMs)
+      throws Exception {
+    final long countStarResult = 97889L;
+    TestUtils.waitForCondition(new Function<Void, Boolean>() {
```
kishoreg: TestUtils? better to use RetryPolicies in PinotCommon. It allows you to try every 5 seconds until 60 seconds.

pinot-tools/src/main/java/org/apache/pinot/tools/Quickstart.java
```diff
@@ -158,8 +161,8 @@ public void execute()
     runner.buildSegment();
     printStatus(Color.CYAN, "***** Pushing segment to the controller *****");
     runner.pushSegment();
-    printStatus(Color.CYAN, "***** Waiting for 5 seconds for the server to fetch the assigned segment *****");
-    Thread.sleep(5000);
+    printStatus(Color.CYAN, "***** Waiting for 60 seconds for the server to fetch the assigned segment *****");
+    waitForAllDocsLoaded(runner, 6000L);
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-tools/src/main/java/org/apache/pinot/tools/Quickstart.java
jackjlli: Changed to `RetryPolicies`.

npawar:

PR #4362, opened 2019-06-25
Optimize MutableOffHeapByteArrayStore by directly calling the PinotDataBuffer API

Jackie-Jiang:
- Use PinotDataBuffer batch bytes read/write API instead of put/get byte by byte
- Change the buffer size to be int (long buffer size is not supported)
- Remove the deprecated OffHeapStringStore

Benchmark number of PinotDataBuffer batch/non-batch read/write API:
```
Benchmark                               (_valueLength)  Mode  Cnt    Score    Error  Units
BenchmarkPinotDataBuffer.batchRead                   1  avgt    5    1.915 ±  0.041  ms/op
BenchmarkPinotDataBuffer.batchRead                   2  avgt    5    2.477 ±  0.059  ms/op
BenchmarkPinotDataBuffer.batchRead                   4  avgt    5    4.468 ±  0.058  ms/op
BenchmarkPinotDataBuffer.batchRead                   8  avgt    5    5.729 ±  0.150  ms/op
BenchmarkPinotDataBuffer.batchRead                  16  avgt    5    5.122 ±  0.170  ms/op
BenchmarkPinotDataBuffer.batchRead                  32  avgt    5    4.876 ±  0.371  ms/op
BenchmarkPinotDataBuffer.batchRead                  64  avgt    5    5.332 ±  0.311  ms/op
BenchmarkPinotDataBuffer.batchRead                 128  avgt    5    6.705 ±  0.450  ms/op
BenchmarkPinotDataBuffer.batchRead                 256  avgt    5    8.245 ±  0.606  ms/op
BenchmarkPinotDataBuffer.batchRead                 512  avgt    5   10.590 ±  1.525  ms/op
BenchmarkPinotDataBuffer.batchRead                1024  avgt    5   17.031 ±  2.506  ms/op
BenchmarkPinotDataBuffer.batchWrite                  1  avgt    5    3.111 ±  0.107  ms/op
BenchmarkPinotDataBuffer.batchWrite                  2  avgt    5    3.355 ±  0.074  ms/op
BenchmarkPinotDataBuffer.batchWrite                  4  avgt    5    5.514 ±  0.173  ms/op
BenchmarkPinotDataBuffer.batchWrite                  8  avgt    5    7.277 ±  0.070  ms/op
BenchmarkPinotDataBuffer.batchWrite                 16  avgt    5    5.373 ±  0.036  ms/op
BenchmarkPinotDataBuffer.batchWrite                 32  avgt    5    5.211 ±  0.386  ms/op
BenchmarkPinotDataBuffer.batchWrite                 64  avgt    5    5.900 ±  0.516  ms/op
BenchmarkPinotDataBuffer.batchWrite                128  avgt    5    6.695 ±  0.735  ms/op
BenchmarkPinotDataBuffer.batchWrite                256  avgt    5    9.534 ±  0.656  ms/op
BenchmarkPinotDataBuffer.batchWrite                512  avgt    5   11.101 ±  0.693  ms/op
BenchmarkPinotDataBuffer.batchWrite               1024  avgt    5   14.272 ±  0.539  ms/op
BenchmarkPinotDataBuffer.nonBatchRead                1  avgt    5    2.392 ±  0.024  ms/op
BenchmarkPinotDataBuffer.nonBatchRead                2  avgt    5    3.496 ±  0.138  ms/op
BenchmarkPinotDataBuffer.nonBatchRead                4  avgt    5    4.546 ±  0.092  ms/op
BenchmarkPinotDataBuffer.nonBatchRead                8  avgt    5    5.727 ±  0.100  ms/op
BenchmarkPinotDataBuffer.nonBatchRead               16  avgt    5    8.145 ±  0.196  ms/op
BenchmarkPinotDataBuffer.nonBatchRead               32  avgt    5   13.088 ±  0.373  ms/op
BenchmarkPinotDataBuffer.nonBatchRead               64  avgt    5   23.039 ±  0.110  ms/op
BenchmarkPinotDataBuffer.nonBatchRead              128  avgt    5   43.596 ±  0.699  ms/op
BenchmarkPinotDataBuffer.nonBatchRead              256  avgt    5   82.396 ±  1.121  ms/op
BenchmarkPinotDataBuffer.nonBatchRead              512  avgt    5  161.428 ± 10.053  ms/op
BenchmarkPinotDataBuffer.nonBatchRead             1024  avgt    5  319.884 ±  5.365  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite               1  avgt    5    3.328 ±  0.034  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite               2  avgt    5    4.271 ±  0.240  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite               4  avgt    5    5.567 ±  0.095  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite               8  avgt    5    6.870 ±  0.164  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite              16  avgt    5    9.634 ±  0.188  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite              32  avgt    5   15.254 ±  0.588  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite              64  avgt    5   26.027 ±  0.377  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite             128  avgt    5   58.414 ±  1.256  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite             256  avgt    5  112.811 ±  2.249  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite             512  avgt    5  221.752 ±  2.795  ms/op
BenchmarkPinotDataBuffer.nonBatchWrite            1024  avgt    5  454.549 ± 12.985  ms/op
```


sunithabeeram:
Do you have numbers from tests for this?

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
```diff
@@ -113,41 +106,41 @@ private int add(byte[] value) {
         // full
         return -1;
       }
-      for (int i = 0, j = startOffset; i < value.length; i++, j++) {
-        _byteBuffer.put(j, value[i]);
-      }
-      _byteBuffer.putInt(_numValues * Integer.BYTES, startOffset);
+      _pinotDataBuffer.readFrom(startOffset, value);
```
mayankshriv: I take it that this is the batch read being referred to? Probably I am missing something, but it seemed that the underlying implementations were still iterating over individual bytes. If so, where is the speedup coming from?

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
```diff
@@ -157,7 +150,7 @@ public void close()
       _pinotDataBuffer.close();
     }
 
-    private long getSize() {
+    private int getSize() {
```
mcvsubbu: I think we should support long. At the time this was coded, Lbuffer support had bugs.

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
Jackie-Jiang: The underlying implementation is simply iterating over individual bytes. Check PinotByteBuffer for details.

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
Jackie-Jiang: The total size could be long, while each buffer's size is limited to 2G.
Besides, for real-time forward index/dictionary, I don't think we should allow buffers of gigabytes because that will definitely cause problem when committing the segment.

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/MutableOffHeapByteArrayStore.java
buchireddy: I'm also curious if this gives performance improvement and how much. As per the code in `org.apache.pinot.core.segment.memory.PinotByteBuffer#readFrom(long, byte[], int, int)`, if the buffer size is >10 (hard-coded value), the buffer is indeed duplicated during the batch operation so there could be garbage impact too.

Jackie-Jiang:
@sunithabeeram @buchireddy Added the benchmark number, and the performance difference is quite huge

buchireddy:
> @sunithabeeram @buchireddy Added the benchmark number, and the performance difference is quite huge

@Jackie-Jiang I'm assuming the lower the score is, the better it is and the results definitely look promising. Does this benchmarking consider the GC latencies as well? Like I mentioned above, batch copying seems to duplicate the buffers so you definitely need to analyze how could the buffer duplication impact the application.

Thanks for the benchmarking though.

Jackie-Jiang:
@buchireddy Yes, the unit of the score is ms/op.
You are correct that batch read/write will create a buffer duplicate, which is extra garbage. And that is why we put a threshold so that small read/write don't create the duplicate as an optimization (the old implementation always create the buffer). The duplicate of the buffer is always short-lived tiny garbage, which should have negligible GC impact. The benchmark is testing the read/write for hundreds millions times, so the GC should already be included.

mcvsubbu:
> @buchireddy Yes, the unit of the score is ms/op.
> You are correct that batch read/write will create a buffer duplicate, which is extra garbage. And that is why we put a threshold so that small read/write don't create the duplicate as an optimization (the old implementation always create the buffer). The duplicate of the buffer is always short-lived tiny garbage, which should have negligible GC impact. The benchmark is testing the read/write for hundreds millions times, so the GC should already be included.

The previous implementation created a byte buffer _once_ and used that for read/write. It did not do a bulk copy, but it did not generate garbage on each get() either.
Remember, get() is  called during value lookup of dictionary multiple times (once on each hash collision). Given that the whole off-heap effort is to save on GC, I am wondering if that optimization is worth it. 

Isn't it better to benchmark the whole dictionary operation for some datasets taken from production?

Jackie-Jiang:
@mcvsubbu As you can see in the benchmark number, for read/write of 128 bytes, there are about 7x performance penalty, and for 1k bytes, 20x if reading/writing byte-by-byte. And these numbers already took GC into account. The whole point of using off-heap is to reduce the long-lived large on-heap garbage but allow short-lived small garbages which can be easily collected (think about how to read from off-heap buffer).
The extra byte buffer in the previous implementation is not necessary as well because you can directly call PinotDataBuffer.getByte()/putByte().
This PR focuses on the performance gain from batch read/write, so as the benchmark.

PR #4363, opened 2019-06-25
When initializing mutable dictionary, preserve 10% buffer for cardinality to reduce the chance of re-sizing the dictionary

Jackie-Jiang:
The current off-heap mutable dictionary is designed to hold all values in a single buffer
If the buffer allocated is just enough for the cardinality estimation, there are quite high possibility that the dictionary needs to be expanded, which will impact performance
In order to reduce the chance of re-sizing the dictionary, preserve 10% buffer for cardinality

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
@@ -44,7 +44,6 @@ import org.apache.pinot.core.realtime.impl.RealtimeSegmentStatsHistory;
 import org.apache.pinot.core.realtime.impl.dictionary.MutableDictionary;
 import org.apache.pinot.core.realtime.impl.dictionary.MutableDictionaryFactory;
 import org.apache.pinot.core.realtime.impl.invertedindex.RealtimeInvertedIndexReader;
-import org.apache.pinot.core.realtime.stream.StreamMessageMetadata;
 import org.apache.pinot.core.segment.creator.impl.V1Constants;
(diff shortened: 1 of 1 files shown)


sunithabeeram:
Please add numbers from testing for PRs like this.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -165,9 +164,11 @@ public long getLatestIngestionTimestamp() {
           dictionaryColumnSize = dataType.size();
         }
         String allocationContext = buildAllocationContext(_segmentName, column, V1Constants.Dict.FILE_EXTENSION);
+        // NOTE: preserve 20% buffer for cardinality to reduce the chance of re-sizing the dictionary
+        int estimatedCardinality = (int) (_statsHistory.getEstimatedCardinality(column) * 1.2);
```
mcvsubbu: I think 10% should be sufficient, but then I agree that both 10 and 20 percent are arbitrary. Why not make this a per-machine config? Default can be 10, and we can play with it as needed.

mcvsubbu: @Jackie-Jiang I think we should merge this change soon. Do you plan to add a per0-machine config in this PR or later? Or change to 10%?

Jackie-Jiang: I can change it to 10%. The long term solution should be expanding the buffers as more values added. Allocating huge buffer at once is a quite slow operation (hundreds milliseconds), which should be avoided.

PR #4367, opened 2019-06-26
CompletionConfig for realtime tables

npawar:
Based on discussion in https://github.com/apache/incubator-pinot/issues/4345
Tested the change on a test cluster

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/CompletionConfig.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/CompletionConfig.java
@@ -30,7 +30,7 @@ import org.apache.pinot.common.utils.EqualityUtils;
 public class CompletionConfig {
 
   @ConfigKey(value = "completionMode")
-  @ConfigDoc("Mode to use when completing segment. DEFAULT for default strategy (build segment if segment is equivalent to the committed segment, else download). DOWNLOAD for always download the segment, never build. ")
+  @ConfigDoc(value = "Mode to use when completing segment. DEFAULT for default strategy (build segment if segment is equivalent to the committed segment, else download). DOWNLOAD for always download the segment, never build.", mandatory = false)
   private String _completionMode;
 
   public String getCompletionMode() {
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/CompletionConfig.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/CompletionConfig.java
@@ -1,0 +1,62 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


pinot-common/src/main/java/org/apache/pinot/common/config/CompletionConfig.java
```diff
@@ -0,0 +1,62 @@
...
+ * Class representing configurations related to realtime segment completion.
+ *
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CompletionConfig {
+
+  @ConfigKey(value = "completionMode")
+  private String _completionMode;
+
+  public String getCompletionMode() {
```
mcvsubbu: Can we add some documentation here on what values can exist for completion mode? Please use @ ConfigDoc annotation giving examples as well.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -320,6 +320,11 @@ public ServerType getServerType() {
         IN_PROGRESS, DONE
       }
 
+      public enum CompletionMode {
+        DEFAULT, // default behavior
+        DOWNLOAD // always download the segment
```
mcvsubbu: please add more document the behavior here,as verbose as you can (comments are free :)). In the current implementation, we are planning on the server paying attention to this config, and deciding to download or build a realtime segment that it knows are equivalent either way.

npawar:
> Can you also include the config for default starting numrows for segment size based threshold calcluations? We can then see how these two configs fit together
> 
> thanks

I added the comments. 
I would like to add the other config for default starting numrows in a future PR if that's okay (will send one out soon after this is merged)

pinot-common/src/main/java/org/apache/pinot/common/config/CompletionConfig.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Class representing configurations related to realtime segment completion.
+ *
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CompletionConfig {
+
+  @ConfigKey(value = "completionMode")
+  @ConfigDoc("Mode to use when completing segment. DEFAULT for default strategy (build segment if segment is equivalent to the committed segment, else download). DOWNLOAD for always download the segment, never build. ")
```
mcvsubbu:
```suggestion
  @ConfigDoc(value = "Mode to use when completing segment. DEFAULT for default strategy (build segment if segment is equivalent to the committed segment, else download). DOWNLOAD for always download the segment, never build.", mandatory = false)
```

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -320,6 +320,14 @@ public ServerType getServerType() {
         IN_PROGRESS, DONE
       }
 
+      /**
+       * Decides how the non-winner servers should get the segment
+       */
+      public enum CompletionMode {
+        DEFAULT, // default behavior - if the in memory segment in the non-winner segmis equivalent to the committed segment, then build and replace, else download
```
mcvsubbu: segmis?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -320,6 +320,14 @@ public ServerType getServerType() {
         IN_PROGRESS, DONE
       }
 
+      /**
+       * Decides how the non-winner servers should get the segment
```
mcvsubbu:
```suggestion
       * During realtime segment completion, the value of this enum decides how  non-winner servers should replace  the completed segment.
```

PR #4368, opened 2019-06-26
3891: Check for validity of segment start/end time during segment generation

siddharthteotia:
This will allow us to fail-fast during segment generation if there is incorrect data as opposed to detecting this later during segment upload. The check during segment upload is still retained though

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
@@ -95,6 +95,7 @@ public class RealtimeSegmentConverter {
       reader = new RealtimeSegmentRecordReader(realtimeSegmentImpl, dataSchema, sortedColumn);
     }
     SegmentGeneratorConfig genConfig = new SegmentGeneratorConfig(dataSchema);
+    genConfig.setCheckTimeColumnValidityDuringGeneration(false);
     if (invertedIndexColumns != null && !invertedIndexColumns.isEmpty()) {
       for (String column : invertedIndexColumns) {
         genConfig.createInvertedIndexForColumn(column);
--- a/pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
@@ -95,6 +95,11 @@ public class RealtimeSegmentConverter {
       reader = new RealtimeSegmentRecordReader(realtimeSegmentImpl, dataSchema, sortedColumn);
     }
     SegmentGeneratorConfig genConfig = new SegmentGeneratorConfig(dataSchema);
+    // The segment generation code in SegmentColumnarIndexCreator will throw
+    // exception if start and end time in time column are not in acceptable
+    // range. We don't want the realtime consumption to stop (if an exception
+    // is thrown) and thus the time validity check is explicitly disabled for
+    // realtime segment generation
     genConfig.setCheckTimeColumnValidityDuringGeneration(false);
     if (invertedIndexColumns != null && !invertedIndexColumns.isEmpty()) {
       for (String column : invertedIndexColumns) {
--- a/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -350,7 +350,14 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
         Object minTime = Preconditions.checkNotNull(timeColumnIndexCreationInfo.getMin());
         Object maxTime = Preconditions.checkNotNull(timeColumnIndexCreationInfo.getMax());
 
-        if (config.getTimeColumnType() == SegmentGeneratorConfig.TimeColumnType.SIMPLE_DATE) {
+        if (config.getSegmentTimeUnit() == TimeUnit.DAYS) {
+          final long minTimeMillis = TimeUnit.DAYS.toMillis(((Integer) minTime).longValue());
+          final long maxTimeMillis =  TimeUnit.DAYS.toMillis(((Integer) maxTime).longValue());
+          checkTime(config, minTimeMillis, maxTimeMillis, segmentName);
+          properties.setProperty(SEGMENT_START_TIME, minTimeMillis);
+          properties.setProperty(SEGMENT_END_TIME, maxTimeMillis);
+          properties.setProperty(TIME_UNIT, TimeUnit.MILLISECONDS);
+        } else if (config.getTimeColumnType() == SegmentGeneratorConfig.TimeColumnType.SIMPLE_DATE) {
           // For simple date format, convert time value into millis since epoch
           DateTimeFormatter dateTimeFormatter = DateTimeFormat.forPattern(config.getSimpleDateFormat());
           final Long minTimeMillis = dateTimeFormatter.parseMillis(minTime.toString());
@@ -416,8 +423,8 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
       throw new RuntimeException("Expecting non-null start/end time for segment: " + segmentName);
     }
 
-    long startMillis = 0;
-    long endMillis = 0;
+    long startMillis;
+    long endMillis;
 
     if (startTime instanceof Long && endTime instanceof Long) {
       startMillis = (long)startTime;
--- a/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
@@ -431,13 +431,22 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
       end = ((Integer) endTime).longValue();
     } else {
       final StringBuilder err = new StringBuilder();
-      err.append(
-          "Unable to interpret type of time column value. Failed to validate start and end time of segment time column")
-          .append(" uninterpreted type: ").append(startTime.getClass()).append(" start time: ").append(startTime)
-          .append(" end time: ").append(endTime).append(" time column name: ").append(config.getTimeColumnName())
-          .append(" segment name: ").append(segmentName).append(" segment time column unit: ")
-          .append(config.getSegmentTimeUnit().toString()).append(" segment time column type: ")
-          .append(config.getTimeColumnType().toString()).append(" time field spec data type: ")
+      err.append("Unable to interpret type of time column value. Failed to validate start and end time of segment")
+          .append(" uninterpreted type: ")
+          .append(startTime.getClass())
+          .append(" start time: ")
+          .append(startTime)
+          .append(" end time: ")
+          .append(endTime)
+          .append(" time column name: ")
+          .append(config.getTimeColumnName())
+          .append(" segment name: ")
+          .append(segmentName)
+          .append(" segment time column unit: ")
+          .append(config.getSegmentTimeUnit().toString())
+          .append(" segment time column type: ")
+          .append(config.getTimeColumnType().toString())
+          .append(" time field spec data type: ")
           .append(config.getSchema().getTimeFieldSpec().getDataType().toString());
       LOGGER.error(err.toString());
       throw new RuntimeException(err.toString());
@@ -480,8 +489,11 @@ public class SegmentColumnarIndexCreator implements SegmentCreator {
           if (config.getSegmentTimeUnit() != TimeUnit.MILLISECONDS) {
             // we should never be here
             final StringBuilder err = new StringBuilder();
-            err.append("Unexpected time unit: ").append(config.getSegmentTimeUnit().toString())
-                .append(" for time column: ").append(config.getTimeColumnName()).append(" for segment: ")
+            err.append("Unexpected time unit: ")
+                .append(config.getSegmentTimeUnit().toString())
+                .append(" for time column: ")
+                .append(config.getTimeColumnName())
+                .append(" for segment: ")
                 .append(segmentName);
             LOGGER.error(err.toString());
             throw new RuntimeException(err.toString());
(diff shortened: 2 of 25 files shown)


siddharthteotia:
cc @mcvsubbu, @jackjlli

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -390,6 +397,43 @@ void writeMetadata()
...
+
+    if (startTime instanceof Long && endTime instanceof Long) {
+      startMillis = (long)startTime;
+      endMillis = (long)endTime;
+    } else if (startTime instanceof Integer && endTime instanceof Integer) {
+      startMillis = ((Integer) startTime).longValue();
+      endMillis = ((Integer) endTime).longValue();
+    } else if (startTime instanceof String && endTime instanceof String) {
+      startMillis = Long.parseLong((String)startTime);
+      endMillis = Long.parseLong((String)endTime);
```
siddharthteotia: I am not quite sure about the type comparison. Have some questions on the format of time that is incoming here at the time of generation v/s how it is stored. By looking at the code in SegmentMetadataImpl, seems like it is stored as string and is always parsed as long

siddharthteotia:
@jackjlli , @mcvsubbu , I have been debugging the 31 unit test failures as reported here:
https://travis-ci.org/apache/incubator-pinot/jobs/550998101

Some of them I have fixed like MultiplePinotSegmentRecordReaderTest.  This test was generating its own test data and I tweaked it to account for generating time column values within acceptable range. Similarly, the unit test I added to test the newly added in SegmentGenerationWithTimeColumnTest  also piggy-backed on the random data generation in the test code. These tests first put the dummy data in a list of GenericRow and create a record reader with the schema and generic row set. Later SegmentIndexCreationDriver is initialized with the record reader for dummy data

However, many others like IntArraysTest, ColumnMetadataTest, SegmentMetadataImplTest, SegmentPrepropcessorTest seem to be reading the test data from avro file and thus there is no control in the test code over the data. SegmentIndexCreationDriver is initialized just with the generation config that internally has details of path to avro test dataset.

I am wondering how to fix these kinds of tests now that segment generation code does the time column validity check and the input data is in avro file?

I am still going through rest of the test failures but wanted an opinion w.r.t fixing tests that use external dataset.

jackjlli:
@siddharthteotia can we try modifying the test data? There's no point to allow generating un-pushable data. Thanks!

siddharthteotia:
@jackjlli , for the tests that I have fixed and the ones that I added, test data was generated by the test code using java random. However, for the other set of tests, the data is not generated by test code itself. Instead, it is an external avro file and we use that to build a segment. So no code change seems possible to change the data. As discussed offline, I will explore using avro-tools and see if the data in time column data in avro file can be updated.

mayankshriv:
@siddharthteotia One work-around would be write some throw-away java code that reads in the avro in question and writes another avro with valid/allowed time stamps. Should be pretty straight forward to do that.

siddharthteotia:
> @siddharthteotia One work-around would be write some throw-away java code that reads in the avro in question and writes another avro with valid/allowed time stamps. Should be pretty straight forward to do that.

@mayankshriv , thanks for the suggestion. Yes that is what I was doing (after intially trying to put the avro file in a hive table and then updating it via sql). After a brief discussion with @mcvsubbu , we also decided that it will be good to have an option in SegmentGeneratorConfig to enable/disable (if at all needed) this time validity check in segment generation code path. The check is enabled by default and we can see later if realtime code has any surprises in production -- in that case we can disable it. 

Secondly, we also discussed about the changes to failing unit tests. Very few of them were fixed by changing the data generation code. For rest of them, I was in the middle of changing the avro datasets. However, for current PR, I have just explicitly disabled the check for these unit tests that use avro file. I have another patch in progress that cleans up the avro data through a simple java program that updates the records. 

Also, a unit test has been added to test the exception code in SegmentGenerationWithTimeColumnTest.java

siddharthteotia:
@jackjlli . @mayankshriv , @mcvsubbu 

Got a clean build. Please review the changes when you get a chance. It took some iterations to make the time check code robust since it has to work with multiple TimeSpecs and both EPOCH and SIMPLE_DATE_FORMAT time column types.

Coming to tests where validity check is explicitly disabled -- as I mentioned earlier, I will create a PR (once this is merged) to clean up the avro data sets used by these tests. 

Some unit tests have been changed to generate valid data (for e.g PinotSegmentUtil) and a new unit test has been added to SegmentGenerationTimecolumnTest to test the newly added exception code in SegmentColumnarIndexCreator

I am seeing if a couple of more unit tests can be added. Meanwhile, it would be great to get feedback on the changes.

pinot-core/src/test/java/org/apache/pinot/core/segment/index/creator/SegmentGenerationWithTimeColumnTest.java
```diff
@@ -81,12 +84,25 @@ public void testSimpleDateSegmentGeneration()
...
     Assert.assertEquals(metadata.getStartTime(), minTime);
     Assert.assertEquals(metadata.getEndTime(), maxTime);
   }
 
+  @Test
+  public void testSegmentGenerationWithInvalidTime() {
+    boolean error = false;
+    try {
+      Schema schema = createSchema(false);
+      buildSegment(schema, false, true);
```
jackjlli: You can add a statement below this line like:
```
Assert.fail("...");
```
so that it will guarantee to fail the test if `buildSegment()` passes.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -391,6 +399,102 @@ void writeMetadata()
...
+                .append(" for time column: ").append(config.getTimeColumnName());
+            throw new RuntimeException(sb.toString());
+          }
+      }
+    }
+
+    if (!TimeUtils.timeValueInValidRange(start) || !TimeUtils.timeValueInValidRange(end)) {
+      final Date minDate = new Date(TimeUtils.getValidMinTimeMillis());
+      final Date maxDate = new Date(TimeUtils.getValidMaxTimeMillis());
+      LOGGER.error(
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-core/src/test/java/org/apache/pinot/core/data/readers/PinotSegmentUtil.java
```diff
@@ -47,16 +52,16 @@ private PinotSegmentUtil() {
 
   public static List<GenericRow> createTestData(Schema schema, int numRows) {
     List<GenericRow> rows = new ArrayList<>();
-    final Random random = new Random();
+    final ThreadLocalRandom random = ThreadLocalRandom.current();
```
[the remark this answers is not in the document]
siddharthteotia: For this class, there is no other use except for the fact that ThreadLocalRandom has the method nextLong(start, end) whereas java.util.Random doesn't provide any such method to randomly generate long values between a given lower and upper bound

pinot-core/src/test/java/org/apache/pinot/core/segment/index/creator/SegmentGenerationWithTimeColumnTest.java
siddharthteotia: Yes, we could do that. However, it won't let me test for the existence of message - exception detail. Right?

jackjlli: I think it's ok not to have assertion on the content of the exception; if the exception message changes in the future we'll still need to modify all the checks here. You can just write the code like this:
```
try {
  blabla();
  Assert.fail("...");
} catch (Exception e) {
  // Expected.
}
```

siddharthteotia: Done.

siddharthteotia:
Addressed the latest review comment. The build is clean

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -391,6 +399,134 @@ void writeMetadata()
...
+    }
+
+    if (startTime == null || endTime == null) {
+      throw new RuntimeException("Expecting non-null start/end time for segment: " + segmentName);
+    }
+
+    long start;
+    long end;
+
+    if (startTime instanceof Long && endTime instanceof Long) {
```
jackjlli: I'm thinking whether we can firstly check `startTime` ad `endTime` are of the same class, then use switch block to differentiate the cases on either one of the objects. By doing so we don't need to keep a list of `if else block` here. But you can decide.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -391,6 +399,134 @@ void writeMetadata()
...
+    } else if (startTime instanceof String && endTime instanceof String) {
+      start = Long.parseLong((String) startTime);
+      end = Long.parseLong((String) endTime);
+    } else if (startTime instanceof Integer && endTime instanceof Integer) {
+      start = ((Integer) startTime).longValue();
+      end = ((Integer) endTime).longValue();
+    } else {
+      final StringBuilder err = new StringBuilder();
+      err.append("Unable to interpret type of time column value. Failed to validate start and end time of segment")
+          .append(" uninterpreted type: ")
```
jackjlli: Could you re-format this file with Pinot check style?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -391,6 +399,134 @@ void writeMetadata()
     properties.save();
   }
 
+  /**
+   * Check for the validity of segment start and end time
+   * @param startTime segment start time
+   * @param endTime segment end time
+   * @param segmentName segment name
+   */
+  private void checkTime(final SegmentGeneratorConfig config, final Object startTime, final Object endTime,
```
mcvsubbu: we already have a time interval checker that the controller uses in the segment push code path. Can we re-use that (if necessary, move it to some common place)

siddharthteotia: The method in SegmentValidator does the check using segment metadata and compares the startMillis and endMillis against the valid range. 

I have written a common utility method in TimeUtils that takes the start and end time as millis from epoch and compared them against the valid range. This ensures that if in future the validity comparison logic changes, we can do it in single place and it will impact both scenarios -- check during generation and check during segment upload in controller.

It didn't seem feasible to abstract the complete method into a common method since new code (in segment generation), handles a lot more information -- data types (INT, LONG, STRING) along with time units (DAYS, MILLIS, HOURs etc) and whether we have EPOCH or SIMPLE_DATE_FORMAT. This also means that if the check fails, I dump few additional information into logger and error message. 

Let me know if this is fine.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
siddharthteotia: Done

siddharthteotia:
> also, we should set the check to false in the realtime segment generator code path

Done

siddharthteotia:
Addressed latest review comments

siddharthteotia:
The build is clean and all latest review comments have been addressed. Please take a look when you get a chance.

pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
```diff
@@ -95,6 +95,7 @@ public void build(@Nullable SegmentVersion segmentVersion, ServerMetrics serverM
       reader = new RealtimeSegmentRecordReader(realtimeSegmentImpl, dataSchema, sortedColumn);
     }
     SegmentGeneratorConfig genConfig = new SegmentGeneratorConfig(dataSchema);
+    genConfig.setCheckTimeColumnValidityDuringGeneration(false);
```
mcvsubbu: Good to add a comment here saying that we don't want realitme consumption to stop (if the exception fires, consumption _will_ stop) in case application is pushing bad data, so we explicitly ignore the time column check while generating realtime segments.

pinot-core/src/test/java/org/apache/pinot/core/plan/maker/MetadataAndDictionaryAggregationPlanMakerTest.java
```diff
@@ -94,6 +94,7 @@ public void buildSegment()
     segmentGeneratorConfig.setTableName("testTable");
     segmentGeneratorConfig.setSegmentName(SEGMENT_NAME);
     segmentGeneratorConfig.setOutDir(INDEX_DIR.getAbsolutePath());
+    segmentGeneratorConfig.setCheckTimeColumnValidityDuringGeneration(false);
```
mcvsubbu: maybe add a comment in these places as to the reason we are not checking for time column validity. I believe the reason is that the input avro files have invalid values for time column.

siddharthteotia: Good point. Done

pinot-core/src/main/java/org/apache/pinot/core/realtime/converter/RealtimeSegmentConverter.java
siddharthteotia: Done.

siddharthteotia:
Rebased

siddharthteotia:
It looks like the build (for integration tests) was stalled. The log seems to indicate it was running HadoopSegmentBuildPushOfflineClusterIntegrationTest at that point. The test passes locally. May be we can retrigger?

PR #4371, opened 2019-06-27
[TE] Support Anomaly Hierarchy for Composite alerts

akshayrai:
Extend anomaly hierarchy to multiple levels

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/bao/jdbc/MergedAnomalyResultManagerImpl.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/bao/jdbc/MergedAnomalyResultManagerImpl.java
@@ -19,6 +19,7 @@
 
 package org.apache.pinot.thirdeye.datalayer.bao.jdbc;
 
+import com.google.common.base.Preconditions;
 import com.google.inject.Singleton;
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/bao/jdbc/MergedAnomalyResultManagerImpl.java
```diff
@@ -98,24 +99,82 @@ public Long save(MergedAnomalyResultDTO mergedAnomalyResultDTO) {
...
+    return genericPojoDao.update(mergeAnomalyBean);
+  }
+
+  private Set<Long> saveChildAnomalies(MergedAnomalyResultDTO parentAnomaly,
+      Set<MergedAnomalyResultDTO> visitedAnomalies) {
+    Set<Long> childIds = new HashSet<>();
+    Set<MergedAnomalyResultDTO> childAnomalies = parentAnomaly.getChildren();
+    if (childAnomalies == null || childAnomalies.isEmpty()) {
+      // No child anomalies to save
+      return null;
```
jihaozh: How about return an empty set instead of null to prevent NullPointerException.

akshayrai: Updated accorgingly

PR #4373, opened 2019-06-27
Adding error if table does not exist to avoid null pointer

jenniferdai:


Jackie-Jiang:
This is not the correct fix. The check should be on the caller side

PR #4374, opened 2019-06-27
Enhance the time boundary service to get consistent result

Jackie-Jiang:
When pushing multiple offline segments, the first pushed offline segment will push forward the time boundary before the other offline segments arrived. If we directly push the time boundary to the segment end time, we might get inconsistent result before all the segments arrived. In order to address this issue:
- Case 1: segment has the same start and end time, directly use the end time as the time boundary. It is OK to directly use the end time as the time boundary because all the records in the new segments have the same time value and the time filter will filter all of them out, so the result is consistent. The reason why we handle this case separately is because there are use cases with time unit other than DAYS, but have time value rounded to the start of the day for offline table (offline segments have the same start and end time), and un-rounded time value for real-time table (real-time segments have different start and end time). In order to get the correct result, must attach filter 'time < endTime' to the offline side and filter 'time >= endTime' to the real-time side.
- Case 2. segment has different start and end time, rewind the end time with the push interval and use it as the time boundary. The assumption we made here is that the push will finish in the push interval, and next push will push the time boundary forward again and make the records within the last push interval queryable.

#4156 Broke the backward-compatibility for case 1, and this PR is restoring it.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
@@ -92,7 +92,8 @@ public class HelixExternalViewBasedTimeBoundaryService implements TimeBoundarySe
     List<OfflineSegmentZKMetadata> segmentZKMetadataList =
         ZKMetadataProvider.getOfflineSegmentZKMetadataListForTable(_propertyStore, tableNameWithType);
 
-    long maxTimeValue = -1L;
+    OfflineSegmentZKMetadata latestSegmentZKMetadata = null;
+    long latestSegmentEndTimeMs = 0L;
     for (OfflineSegmentZKMetadata segmentZKMetadata : segmentZKMetadataList) {
       String segmentName = segmentZKMetadata.getSegmentName();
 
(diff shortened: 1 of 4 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
```diff
@@ -111,25 +111,47 @@ public void updateTimeBoundaryService(ExternalView externalView) {
...
 
-      // Convert segment end time into table time unit
-      // NOTE: for now, time unit in segment ZK metadata should always match table time unit, but in the future we might
-      //       want to always use MILLISECONDS as the time unit in segment ZK metadata
-      maxTimeValue = Math.max(maxTimeValue, tableTimeUnit.convert(segmentEndTime, segmentZKMetadata.getTimeUnit()));
+      if (latestSegmentZKMetadata == null) {
+        latestSegmentZKMetadata = segmentZKMetadata;
+      } else {
+        long segmentEndTimeMs = segmentZKMetadata.getTimeUnit().toMillis(segmentEndTime);
+        long latestSegmentEndTimeMs =
```
mcvsubbu: maybe we can compute latestSegmentEndTimeMs each time we discover a later segment?

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
```diff
@@ -111,25 +111,47 @@ public void updateTimeBoundaryService(ExternalView externalView) {
...
-    } else {
-      timeBoundary = maxTimeValue - tableTimeUnit.convert(1L, TimeUnit.DAYS);
+    // Convert segment end time into table time unit
+    // NOTE: for now, time unit in segment ZK metadata should always match table time unit, but in the future we might
+    //       want to always use MILLISECONDS as the time unit in segment ZK metadata
+    long latestSegmentEndTime = latestSegmentZKMetadata.getEndTime();
+    long timeBoundary = tableTimeUnit.convert(latestSegmentEndTime, latestSegmentZKMetadata.getTimeUnit());
+
+    // Case 1 - segment has the same start and end time: directly use the end time as the time boundary for
+    //   backward-compatibility. There are use cases with time unit other than DAYS, but have time value rounded
```
mcvsubbu: You may want to remove the "backward compabitlity" phrase here. Just mention in the checkin comments that the other commit (reference the PR) broke backward compatibility and we are restoring it here.

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
Jackie-Jiang: Good improvement. Done

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
Jackie-Jiang: Done

pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedTimeBoundaryService.java
```diff
@@ -111,25 +112,53 @@ public void updateTimeBoundaryService(ExternalView externalView) {
         continue;
       }
 
-      // Convert segment end time into table time unit
-      // NOTE: for now, time unit in segment ZK metadata should always match table time unit, but in the future we might
-      //       want to always use MILLISECONDS as the time unit in segment ZK metadata
-      maxTimeValue = Math.max(maxTimeValue, tableTimeUnit.convert(segmentEndTime, segmentZKMetadata.getTimeUnit()));
+      long segmentEndTimeMs = segmentZKMetadata.getTimeUnit().toMillis(segmentEndTime);
+      if (latestSegmentZKMetadata == null) {
```
mcvsubbu: `if (latestSegmentZkMetadata == null || segmentEndTimeMs > latestSegmentEndTimeMs)`
works?

Jackie-Jiang: `if (segmentEndTimeMs < latestSegmentEndTimeMs)` is enough. Fixed

Jackie-Jiang:
Drop this PR. Offline and Realtime table should handle the time value in the same fashion.

PR #4376, opened 2019-06-27
Config for overriding initial rows threshold in segment size auto tuning

npawar:
When using segment size based auto tuning, the initial rows threshold is 100k currently. Adding a config to override the initial num rows, so that it can be changed if a table needs more/less as the initial.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamConfig.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamConfig.java
@@ -47,13 +47,13 @@ public class StreamConfig {
   private static final int DEFAULT_FLUSH_THRESHOLD_ROWS = 5_000_000;
   private static final long DEFAULT_FLUSH_THRESHOLD_TIME = TimeUnit.MILLISECONDS.convert(6, TimeUnit.HOURS);
   private static final long DEFAULT_DESIRED_SEGMENT_SIZE_BYTES = 200 * 1024 * 1024; // 200M
-  private static final int DEFAULT_INITIAL_ROWS_THRESHOLD = 100_000;
+  private static final int DEFAULT_FLUSH_AUTOTUNE_INITIAL_ROWS = 100_000;
   private static final String DEFAULT_CONSUMER_FACTORY_CLASS_NAME_STRING =
       "org.apache.pinot.core.realtime.impl.kafka.KafkaConsumerFactory";
 
-  protected static final long DEFAULT_STREAM_CONNECTION_TIMEOUT_MILLIS = 30_000;
-  protected static final int DEFAULT_STREAM_FETCH_TIMEOUT_MILLIS = 5_000;
-  protected static final String SIMPLE_CONSUMER_TYPE_STRING = "simple";
+  static final long DEFAULT_STREAM_CONNECTION_TIMEOUT_MILLIS = 30_000;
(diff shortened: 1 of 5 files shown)


pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamConfig.java
```diff
@@ -197,6 +199,19 @@ public StreamConfig(Map<String, String> streamConfigMap) {
       _flushSegmentDesiredSizeBytes = DEFAULT_DESIRED_SEGMENT_SIZE_BYTES;
     }
 
+    int initialRowsThreshold = 0;
+    String initialRowsThresholdValue = streamConfigMap.get(StreamConfigProperties.SEGMENT_INITIAL_ROWS_THRESHOLD);
+    if (initialRowsThresholdValue != null) {
+      try {
+        initialRowsThreshold = Integer.parseInt(initialRowsThresholdValue);
+      } catch (Exception e) {
+        LOGGER.warn("Caught exception while parsing {}:{}, defaulting to {} ms",
```
mcvsubbu: ms?

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamConfig.java
```diff
@@ -197,6 +199,19 @@ public StreamConfig(Map<String, String> streamConfigMap) {
       _flushSegmentDesiredSizeBytes = DEFAULT_DESIRED_SEGMENT_SIZE_BYTES;
     }
 
+    int initialRowsThreshold = 0;
+    String initialRowsThresholdValue = streamConfigMap.get(StreamConfigProperties.SEGMENT_INITIAL_ROWS_THRESHOLD);
+    if (initialRowsThresholdValue != null) {
+      try {
+        initialRowsThreshold = Integer.parseInt(initialRowsThresholdValue);
```
mcvsubbu: You may want to check if it is at least some threshold value? Or positive value?

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamConfigProperties.java
```diff
@@ -89,6 +89,11 @@
    */
   public static final String SEGMENT_FLUSH_DESIRED_SIZE = "realtime.segment.flush.desired.size";
 
+  /**
+   * The initial num rows to use for segment size auto tuning. By default 100_000 is used.
+   */
+  public static final String SEGMENT_INITIAL_ROWS_THRESHOLD = "realtime.segment.initial.rows.threshold";
```
mcvsubbu: I suggest "realtime.segment.flush.initial.rows.threshold". What do you think?

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamConfig.java
npawar: I've checked for positive, 2 lines below this. I dont think we should check for some minimum value... it might backfire in some cases if threshold was actually expected to be small

pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamConfigProperties.java
npawar: Yes, sounds good. Changed it

PR #4380, opened 2019-06-28
Define SLO config for tables

sunithabeeram:
Pulling out the SLO config part of the changes from https://github.com/apache/incubator-pinot/pull/4356. Addressed review comments.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/SLOConfig.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/SLOConfig.java
@@ -1,0 +1,135 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.common.config;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.lang.reflect.Field;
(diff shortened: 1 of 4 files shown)


kishoreg:
What do we need to add this as part of TableConfig?  Are we going to guarantee these numbers? - latency and freshness.

sunithabeeram:
@kishoreg these are required in the table config as they are per table parameters. The broker emits metrics that track against these numbers. There are no guarantees around this - they are used to obtain overall availability numbers of the service at the desired threshold (ie, 99.999/99.99 etc). The time interval over which the numbers are derived is not dictated by this implementation - its left to the users to derive. This change and the ones in #4356 emit a per response healthy/unhealthy score.

sunithabeeram:
@kishoreg - do you have further thoughts/comments on this?

kishoreg:
I missed the previous comment. SLOConfig is probably not the right name if we are not going to guarantee these latency numbers. 

I feel these config belong to the client and not Pinot. Also, shouldn’t the availability numbers be something that’s computed outside of Pinot on the client side. 

Maybe Pinot client should log these metrics and the clients can pass the threshold when Pinot client is initialized. 

I

sunithabeeram:
@kishoreg sorry about the delay in responses on my end as I am traveling.

Can you elaborate on what you mean by guaranteeing the latency numbers? Unless these are measured, there isn't a way to guarantee them, right? Once the numbers are available, Pinot operators can make appropriate changes to improve the system.

Client side measurements include noise (network and other client side induced latency) and are not reliable measures. 

The config helps emit the base metrics that are can be reliably used in measuring availability numbers from server side. Many services at LinkedIn use server side measurements - we have been emitting these currently from the broker using Jetty/container configs - but they make per table metrics harder to compute (as broker multi-tenancy is common).

kishoreg:
> @kishoreg sorry about the delay in responses on my end as I am traveling.
> 
> Can you elaborate on what you mean by guaranteeing the latency numbers? Unless these are measured, there isn't a way to guarantee them, right? Once the numbers are available, Pinot operators can make appropriate changes to improve the system.

What I meant is the name SLOConfig gives the impression that we will guarantee the latency somehow. Its like qps or quota config. My suggestion is to rename the config to something like log_slow_queries and slow_query_threshold. MySQL provides some config variables like this.
 
> Client side measurements include noise (network and other client side induced latency) and are not reliable measures.

I was referring to pinot client. Typically, network issues are also part of the SLA.

> The config helps emit the base metrics that are can be reliably used in measuring availability numbers from server side. Many services at LinkedIn use server side measurements - we have been emitting these currently from the broker using Jetty/container configs - but they make per table metrics harder to compute (as broker multi-tenancy is common).

That sounds good. Please rename SLOConfig to something else.

sunithabeeram:
Closing this as this has been moved to higher layers.

PR #4382, opened 2019-06-28
Fix java client connection close

sunny19930321:
Find the connection of the Java client, always connect after the query, add the close method in the connection, close the connection when the detection PinotClientTransport has registered


pinot-api/src/main/java/org/apache/pinot/client/Connection.java
```diff
@@ -156,4 +157,13 @@ public ResultSetGroup get(long timeout, TimeUnit unit)
       return new ResultSetGroup(response);
     }
   }
+
+  /**
+   * connection close.
+   */
+  public void close() {
+    if(Objects.nonNull(_transport)) {
+      System.exit(0);
```
kishoreg: we should never have system.exit in the code

sunny209868: Yes, So I put the deleted, in JsonAsyncHttpPinotClientTransport class increased the close method.

sunny209868: If you do not add the close method, the client consistently maintains the connection state, it'll experience threads hanging and resource leaks，Thanks

npawar:

PR #4385, opened 2019-06-29
Implement OffHeapMutableBytesStore for real-time var-length bytes store

Jackie-Jiang:
Comparing with the MutableOffHeapByteArrayStore:
1. Read/Write speed improvement
2. Buffer size is fixed and buffers are allocated on demand, so memory is always used effeciently
3. No need to estimate the buffer size beforehead, and performance is guaranteed (for MutableOffHeapByteArrayStore, if buffer size is under-estimated, pay huge penalty on allocating a huge buffer at runtime, and also read/write speed degration

Added benchmark to compare the read/write speed:
```
Benchmark                                                            (_maxValueLength)  Mode  Cnt     Score    Error  Units
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreRead                   8  avgt    5    29.875 ±  0.370  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreRead                  16  avgt    5    32.344 ±  0.765  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreRead                  32  avgt    5    35.027 ±  0.460  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreRead                  64  avgt    5    40.811 ±  0.414  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreRead                 128  avgt    5    51.767 ±  0.185  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreRead                 256  avgt    5    77.065 ±  1.627  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreRead                 512  avgt    5   127.167 ±  1.075  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreRead                1024  avgt    5   283.145 ±  2.829  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreWrite                  8  avgt    5    16.755 ±  0.374  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreWrite                 16  avgt    5    25.579 ±  0.293  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreWrite                 32  avgt    5    23.953 ±  0.507  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreWrite                 64  avgt    5    51.442 ±  1.230  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreWrite                128  avgt    5    64.334 ±  0.069  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreWrite                256  avgt    5   260.415 ±  1.823  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreWrite                512  avgt    5   615.528 ±  3.180  ms/op
BenchmarkOffHeapMutableBytesStore.mutableOffHeapByteArrayStoreWrite               1024  avgt    5  1168.779 ± 16.961  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreRead                       8  avgt    5    28.978 ±  0.535  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreRead                      16  avgt    5    33.420 ±  0.400  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreRead                      32  avgt    5    37.120 ±  0.272  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreRead                      64  avgt    5    39.552 ±  0.317  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreRead                     128  avgt    5    42.861 ±  0.588  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreRead                     256  avgt    5    47.025 ±  0.189  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreRead                     512  avgt    5    60.740 ±  0.677  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreRead                    1024  avgt    5    89.509 ±  1.838  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreWrite                      8  avgt    5    21.327 ±  0.509  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreWrite                     16  avgt    5    28.494 ±  0.304  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreWrite                     32  avgt    5    33.399 ±  0.533  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreWrite                     64  avgt    5    50.755 ±  1.700  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreWrite                    128  avgt    5    75.702 ±  1.079  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreWrite                    256  avgt    5   129.677 ±  1.942  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreWrite                    512  avgt    5   222.697 ±  2.109  ms/op
BenchmarkOffHeapMutableBytesStore.offHeapMutableBytesStoreWrite                   1024  avgt    5   392.053 ±  3.200  ms/op
```

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
@@ -1,0 +1,259 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.core.realtime.impl.dictionary;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+import javax.annotation.concurrent.ThreadSafe;
+import org.apache.pinot.core.io.readerwriter.PinotDataBufferMemoryManager;
+import org.apache.pinot.core.segment.memory.PinotDataBuffer;
+
+
+/**
+ * Off-heap variable length mutable bytes store.
+ * <p>The class is thread-safe for single writer and multiple readers.
+ * <p>There are two sets of buffers allocated for the store:
+ * <ul>
+ *   <li>
+ *     Offset buffer: stores the value end offsets. The end offset here is the global offset for all the value buffers
+ *     instead of the offset inside one value buffer (e.g. offset 9 in the 5th value buffer will have global offset of
+ *     4 * VALUE_BUFFER_SIZE + 9). New offset buffer is allocated when the previous buffer is filled up. The first entry
+ *     of each offset buffer stores the end offset of the last value (or 0 for the first offset buffer) so that we can
+ *     always get the previous and current value's end offset from the same buffer. Note that we use int type for the
+ *     offset, so each bytes store can hold at most 2^31 = 2GB total size of values.
+ *   </li>
+ *   <li>
+ *     Value buffer: stores the values. For performance concern, to avoid reading one value from multiple buffers, each
+ *     value is stored in one value buffer (no cross-buffer value allowed), and has a size limit of the
+ *     VALUE_BUFFER_SIZE (1MB). New value buffer is allocated when the previous buffer cannot hold the new added value.
+ *   </li>
+ * </ul>
+ * <p>The maximum capacity for the bytes store is 2^24 ~= 17M values and 2^31 = 2GB total size of values.
+ */
+@ThreadSafe
+public class OffHeapMutableBytesStore implements Closeable {
+  private static final byte[] EMPTY_BYTES = new byte[0];
+
+  // Each offset buffer holds 8_192 end offsets and one extra end offset for the last value in the previous buffer (or 0
+  // for the first buffer)
+  // OFFSET_BUFFER_SIZE = (8_192 + 1) * 4 = 32_772
+  private static final int OFFSET_BUFFER_SHIFT_OFFSET = 13;
+  private static final int OFFSET_BUFFER_SIZE = ((1 << OFFSET_BUFFER_SHIFT_OFFSET) + 1) << 2;
+  private static final int OFFSET_BUFFER_MASK = (1 << OFFSET_BUFFER_SHIFT_OFFSET) - 1;
+
+  // Each value buffer holds up to 1_048_576 bytes, which is also the size limit for each value
+  // VALUE_BUFFER_SIZE = 1_048_576
+  private static final int VALUE_BUFFER_SHIFT_OFFSET = 20;
+  private static final int VALUE_BUFFER_SIZE = 1 << VALUE_BUFFER_SHIFT_OFFSET;
+  private static final int VALUE_BUFFER_MASK = (1 << VALUE_BUFFER_SHIFT_OFFSET) - 1;
+
+  // With at most 2_048 buffers, we can store about 17M values of total size up to 2GB
+  // MAX_NUM_BUFFERS = 2_048
+  private static final int MAX_NUM_BUFFERS = 1 << 11;
+
+  private final AtomicReferenceArray<PinotDataBuffer> _offsetBuffers = new AtomicReferenceArray<>(MAX_NUM_BUFFERS);
+  private final AtomicReferenceArray<PinotDataBuffer> _valueBuffers = new AtomicReferenceArray<>(MAX_NUM_BUFFERS);
+  private final PinotDataBufferMemoryManager _memoryManager;
+  private final String _allocationContext;
+
+  // Global end offset of the previous added value
+  private int _previousValueEndOffset;
+  // Number of values added
+  private transient int _numValues;
+  // Total size of the OffHeap buffers (both offset buffers and value buffers)
+  private transient int _totalBufferSize;
+
+  public OffHeapMutableBytesStore(PinotDataBufferMemoryManager memoryManager, String allocationContext) {
+    _memoryManager = memoryManager;
+    _allocationContext = allocationContext;
+  }
+
+  /**
+   * Adds a value into the bytes store. The size of the given value should be smaller or equal to the VALUE_BUFFER_SIZE
+   * (1MB).
+   */
+  public int add(byte[] value) {
+    int offsetBufferIndex = _numValues >>> OFFSET_BUFFER_SHIFT_OFFSET;
+    int offsetIndex = _numValues & OFFSET_BUFFER_MASK;
+
+    PinotDataBuffer offsetBuffer;
+    // If this is the first entry in the offset buffer, allocate a new buffer and store the end offset of the previous
+    // value
+    if (offsetIndex == 0) {
+      offsetBuffer = _memoryManager.allocate(OFFSET_BUFFER_SIZE, _allocationContext);
+      offsetBuffer.putInt(0, _previousValueEndOffset);
+      _offsetBuffers.set(offsetBufferIndex, offsetBuffer);
+      _totalBufferSize += OFFSET_BUFFER_SIZE;
+    } else {
+      offsetBuffer = _offsetBuffers.get(offsetBufferIndex);
+    }
+
+    int valueLength = value.length;
+    if (valueLength == 0) {
+      offsetBuffer.putInt((offsetIndex + 1) << 2, _previousValueEndOffset);
+      return _numValues++;
+    }
+
+    int valueBufferIndex = (_previousValueEndOffset + valueLength - 1) >>> VALUE_BUFFER_SHIFT_OFFSET;
+    PinotDataBuffer valueBuffer;
+    int valueStartOffset;
+    // If the current value buffer does not have enough space, allocate a new buffer to store the value
+    if ((_previousValueEndOffset - 1) >>> VALUE_BUFFER_SHIFT_OFFSET != valueBufferIndex) {
+      valueBuffer = _memoryManager.allocate(VALUE_BUFFER_SIZE, _allocationContext);
+      _valueBuffers.set(valueBufferIndex, valueBuffer);
+      _totalBufferSize += VALUE_BUFFER_SIZE;
+      valueStartOffset = valueBufferIndex << VALUE_BUFFER_SHIFT_OFFSET;
+    } else {
+      valueBuffer = _valueBuffers.get(valueBufferIndex);
+      valueStartOffset = _previousValueEndOffset;
+    }
+
+    int valueEndOffset = valueStartOffset + valueLength;
+    offsetBuffer.putInt((offsetIndex + 1) << 2, valueEndOffset);
+    valueBuffer.readFrom(valueStartOffset & VALUE_BUFFER_MASK, value);
+    _previousValueEndOffset = valueEndOffset;
+
+    return _numValues++;
+  }
+
+  /**
+   * Returns the value at the given index. The given index should be smaller than the number of values already added.
+   */
+  @SuppressWarnings("Duplicates")
+  public byte[] get(int index) {
+    assert index < _numValues;
+
+    int offsetBufferIndex = index >>> OFFSET_BUFFER_SHIFT_OFFSET;
+    PinotDataBuffer offsetBuffer = _offsetBuffers.get(offsetBufferIndex);
+    int offsetIndex = index & OFFSET_BUFFER_MASK;
+    int previousValueEndOffset = offsetBuffer.getInt(offsetIndex << 2);
+    int valueEndOffset = offsetBuffer.getInt((offsetIndex + 1) << 2);
+
+    if (previousValueEndOffset == valueEndOffset) {
+      return EMPTY_BYTES;
+    }
+
+    int valueBufferIndex = (valueEndOffset - 1) >>> VALUE_BUFFER_SHIFT_OFFSET;
+    int startOffsetInValueBuffer;
+    int valueLength;
+    if ((previousValueEndOffset - 1) >>> VALUE_BUFFER_SHIFT_OFFSET != valueBufferIndex) {
+      // The first value in the value buffer
+      startOffsetInValueBuffer = 0;
+      valueLength = valueEndOffset & VALUE_BUFFER_MASK;
+    } else {
+      // Not the first value in the value buffer
+      startOffsetInValueBuffer = previousValueEndOffset & VALUE_BUFFER_MASK;
+      valueLength = valueEndOffset - previousValueEndOffset;
+    }
+
+    byte[] value = new byte[valueLength];
+    _valueBuffers.get(valueBufferIndex).copyTo(startOffsetInValueBuffer, value);
+    return value;
+  }
+
+  /**
+   * Returns whether the value at the given index equals the given value. The given index should be smaller than the
+   * number of values already added.
+   * <p>This method should be preferable when checking whether a value equals the value at a specific index in the bytes
+   * store compared to calling {@link #get(int)} and then compare the values because if the value length does not match,
+   * the value fetch can be skipped.
+   */
+  @SuppressWarnings("Duplicates")
+  public boolean equalsValueAt(int index, byte[] valueToCompare) {
+    assert index < _numValues;
+
+    int offsetBufferIndex = index >>> OFFSET_BUFFER_SHIFT_OFFSET;
+    PinotDataBuffer offsetBuffer = _offsetBuffers.get(offsetBufferIndex);
+    int offsetIndex = index & OFFSET_BUFFER_MASK;
+    int previousValueEndOffset = offsetBuffer.getInt(offsetIndex << 2);
+    int valueEndOffset = offsetBuffer.getInt((offsetIndex + 1) << 2);
+
+    int inputValueLength = valueToCompare.length;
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    int valueBufferIndex = (_valueStartOffset + valueLength - 1) >>> VALUE_BUFFER_SIFT_OFFSET;
+    PinotDataBuffer valueBuffer;
+    // If the current value buffer does not have enough space, allocate a new buffer to store the value
+    if ((_valueStartOffset - 1) >>> VALUE_BUFFER_SIFT_OFFSET != valueBufferIndex) {
+      valueBuffer = _memoryManager.allocate(VALUE_BUFFER_SIZE, _allocationContext);
+      _valueBuffers[valueBufferIndex] = valueBuffer;
+      _totalBufferSize += VALUE_BUFFER_SIZE;
+      _valueStartOffset = valueBufferIndex << VALUE_BUFFER_SIFT_OFFSET;
```
mcvsubbu: Doesn't this statement limit MAX_NUM_BUFFERS to (Integer.SIZE - VALUE_BUFFER_SHIFT_OFFSET)?

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,195 @@
...
+      valueLength = valueEndOffset - previousValueEndOffset;
+    }
+
+    byte[] value = new byte[valueLength];
+    _valueBuffers[valueBufferIndex].copyTo(startOffsetInValueBuffer, value);
+    return value;
+  }
+
+  public boolean equals(int index, byte[] value) {
+    int valueLength = value.length;
```
mcvsubbu: final. Also, it may be useful to rename this as inputValueLength (and 'value' as 'inputValue') fpor readability

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,195 @@
...
+    } else {
+      // Not the first value in the value buffer
+      if (valueEndOffset - previousValueEndOffset != valueLength) {
+        return false;
+      }
+      startOffsetInValueBuffer = previousValueEndOffset & VALUE_BUFFER_MASK;
+    }
+
+    byte[] valueAtIndex = new byte[valueLength];
+    _valueBuffers[valueBufferIndex].copyTo(startOffsetInValueBuffer, valueAtIndex);
```
mcvsubbu: Wouldnt it be faster if you just compared by fetching the bytes from the valueBuffer?

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,195 @@
...
+  private static final int VALUE_BUFFER_SIZE = 1 << VALUE_BUFFER_SIFT_OFFSET;
+  private static final int VALUE_BUFFER_MASK = 0xFFFFFFFF >>> (Integer.SIZE - VALUE_BUFFER_SIFT_OFFSET);
+
+  private final PinotDataBuffer[] _offsetBuffers = new PinotDataBuffer[MAX_NUM_BUFFERS];
+  private final PinotDataBuffer[] _valueBuffers = new PinotDataBuffer[MAX_NUM_BUFFERS];
+  private final PinotDataBufferMemoryManager _memoryManager;
+  private final String _allocationContext;
+
+  private int _numValues;
+  private int _valueStartOffset;
```
mcvsubbu: Please add a comment on how this value is computed. If I understand right, it is the GLOBAL start offset of each value assuming a linear array of (VALUE_ BUFFER_SIZE * numBuffers). If a value cannot be held in one buffer, then the _valueStartOffset is set to the start of the next buffer.

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStoreTest.java
```diff
@@ -0,0 +1,93 @@
...
+        int index = RANDOM.nextInt(NUM_VALUES);
+        assertTrue(offHeapMutableBytesStore.equals(index, _values[index]));
+        if (!Arrays.equals(_values[index], _values[0])) {
+          assertFalse(offHeapMutableBytesStore.equals(0, _values[index]));
+          assertFalse(offHeapMutableBytesStore.equals(index, _values[0]));
+        }
+      }
+    }
+  }
+}
```
mcvsubbu: Please add a test that adds values in one thread and reads values from (say) 4 threads
Please specific tests that test the boundary conditions (hitting buffer as well as offset limits, or hitting one of the limits) rather than relying on the randomness.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,195 @@
...
+ * <p>There are two sets of buffers allocated for the store, one set to store the value end offsets and the other set to
+ * store the values. The new buffer is allocated when the previous buffer is filled up. The first entry of each offset
+ * buffer stores the end offset of the last value (or 0 for the first offset buffer) so that we can always get the
+ * previous and current value's end offset from the same buffer.
+ */
+public class OffHeapMutableBytesStore implements Closeable {
+  private static final byte[] EMPTY_BYTES = new byte[0];
+
+  // MAX_NUM_BUFFERS = 8_192
+  private static final int MAX_NUM_BUFFERS = 1 << 13;
```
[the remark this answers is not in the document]
Jackie-Jiang: This number decides how many total values we can store. 8K buffers allow at most 67M values of total size up to 8GB.
Added more comments on these numbers

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,195 @@
...
+
+  // OFFSET_BUFFER_SIZE = 32_768 + 4
+  private static final int OFFSET_BUFFER_SHIFT_OFFSET = 13;
+  private static final int OFFSET_BUFFER_SIZE = ((1 << OFFSET_BUFFER_SHIFT_OFFSET) + 1) << 2;
+  private static final int OFFSET_BUFFER_MASK = 0xFFFFFFFF >>> (Integer.SIZE - OFFSET_BUFFER_SHIFT_OFFSET);
+
+  // VALUE_BUFFER_SIZE = 1_048_576
+  private static final int VALUE_BUFFER_SIFT_OFFSET = 20;
```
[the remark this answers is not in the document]
Jackie-Jiang: Good catch

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
Jackie-Jiang: Added

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,195 @@
...
+      offsetBuffer = _offsetBuffers[offsetBufferIndex];
+    }
+
+    int valueLength = value.length;
+    if (valueLength == 0) {
+      offsetBuffer.putInt((offsetIndex + 1) << 2, _valueStartOffset);
+      return _numValues++;
+    }
+
+    int valueBufferIndex = (_valueStartOffset + valueLength - 1) >>> VALUE_BUFFER_SIFT_OFFSET;
```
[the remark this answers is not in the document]
Jackie-Jiang: We don't allow storing one value cross multiple buffers, so this is the buffer index for the whole value. Added more comments in the class header.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,195 @@
...
+    PinotDataBuffer valueBuffer;
```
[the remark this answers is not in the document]
Jackie-Jiang: Preconditions at this level can hurt the performance. This bytes store can hold 8GB of data, so we should never exceed MAX_NUM_BUFFERS

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
Jackie-Jiang: No, these 2 are independent.
We use the shift operation to quickly get the startOffset because it is the first value in the current value buffer.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
Jackie-Jiang: I'm not a fan of using final on local variables. Changed the valueLength to inputValueLength.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
Jackie-Jiang: You mean byte-by-byte comparison? If we reach here, that means there is a pretty high chance that the two values are the same (we call this method when hash code matches, and we can only reach here if the length also matches), thus doing batch read should be preferable. I added one extra check to optimize the case of small reads to compare byte-by-byte.

mcvsubbu: But then there is extra garbage generated in the read.

Jackie-Jiang: True, but that is short lived garbage, and the benchmark in #4362 already proved that batch read can give big performance gain for big values.

Jackie-Jiang:
@mcvsubbu Addressed all the comments, please take another look

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
mcvsubbu: valueBufferIndex should fit in (Integer.SIZE - VALUE_BUFFER_SHIFT_OFFSET - 1) bits. Therefore, MAX_NUM_BUFFERS should be at most (2 ** (Integer.SIZE - VALUE_BUFFER_SHIFT_OFFSET - 1)).

If valueBufferIndex does not fit in these bits, then we have a problem with overflow in the expression (valueBufferIndex << VALUE_BUFFER_SHIFT_OFFSET)

Right?

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Off-heap variable length mutable bytes store.
+ * <p>The class is thread-safe for single writer and multiple readers.
+ * <p>There are two sets of buffers allocated for the store:
+ * <ul>
+ *   <li>
+ *     Offset buffer: stores the value end offsets. New offset buffer is allocated when the previous buffer is filled
+ *     up. The first entry of each offset buffer stores the end offset of the last value (or 0 for the first offset
+ *     buffer) so that we can always get the previous and current value's end offset from the same buffer.
```
mcvsubbu: Useful to define "offset" here. I believe you mean the global offset and not the offset within the buffer. I am not sure how to describe "global offset" but do the best you can. At least write that the offset we are talking about is NOT the offset within a valueBuffer.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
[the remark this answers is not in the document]
mcvsubbu: How can it hold 8GB of data when we store global offsets in 32 bits? Maybe I am missing something

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStoreTest.java
```diff
@@ -0,0 +1,137 @@
...
+      futures.add(executorService.submit(() -> {
+        for (int i = 0; i < NUM_VALUES; i++) {
+          offHeapMutableBytesStore.add(_values[i]);
+        }
+      }));
+
+      // Reader threads
+      for (int i = 0; i < numReaders; i++) {
+        futures.add(executorService.submit(() -> {
+          for (int j = 0; j < NUM_READS; j++) {
```
mcvsubbu: Instead of stopping with NUM_READS, stop when the writer thread is done.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
```diff
@@ -0,0 +1,252 @@
...
+@ThreadSafe
+public class OffHeapMutableBytesStore implements Closeable {
+  private static final byte[] EMPTY_BYTES = new byte[0];
+
+  // Each offset buffer holds 8_192 end offsets and one extra end offset for the last value in the previous buffer (or 0
+  // for the first buffer)
+  // OFFSET_BUFFER_SIZE = (8_192 + 1) * 4 = 32_772
+  private static final int OFFSET_BUFFER_SHIFT_OFFSET = 13;
+  private static final int OFFSET_BUFFER_SIZE = ((1 << OFFSET_BUFFER_SHIFT_OFFSET) + 1) << 2;
+  private static final int OFFSET_BUFFER_MASK = 0xFFFFFFFF >>> (Integer.SIZE - OFFSET_BUFFER_SHIFT_OFFSET);
```
mcvsubbu: Is'nt  (1<<OFFSET_BUFFER_SHIFT_OFFSET)-1 easier to understand?

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
Jackie-Jiang: Oh I see what you mean. Very good question. Yes, we should not store more than 2GB data in this bytes store, or we might trigger the int overflow for the global offset. This limitation is listed in the comments

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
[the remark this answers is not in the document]
Jackie-Jiang: I realized that, so I changed the upper bound to 2GB :-P

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
mcvsubbu: So, if we can add a pre-conditions check in the constructor that will be useful.  As written, it seems to imply that the static definitions can be set to whatever and things will work.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
Jackie-Jiang: Added

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStoreTest.java
Jackie-Jiang: Done

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/OffHeapMutableBytesStore.java
Jackie-Jiang: The limitation comes from the int type for offset, but not these constants. Meaning it still works for any buffer size/number of buffers, but we can only hold 2GB data.

Jackie-Jiang:
> It will be nice if you can add a single test case in which both valueBuffer and offsetBuffer limits are reached in the same add() call, instead of leaving it to happen at random. You may need to add getters to expose the limits of the buffers, but it is worth it, I believe.

This 2 buffers are handled independently (first process offsetBuffer, then valueBuffer), so I believe it is not necessary to test that explicitly. Also, with 100_000 writes per test, there should be pretty high chance to hit this scenario.

PR #4386, opened 2019-06-30
Run cursor selected query in query-console

xueyumusic:
Support running `cursor selected query` in code editor, in this way we don't need to clear editor to running new query, like this:
image
It will execute the second query only and don't need to clear first and third query.

### DIFF
--- a/pinot-controller/src/main/resources/static/js/init.js
+++ b/pinot-controller/src/main/resources/static/js/init.js
@@ -65,7 +65,7 @@ $(document).ready(function() {
 
   $("#execute-query").click(function() {
     // execute query and draw the results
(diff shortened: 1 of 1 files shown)


kishoreg:
This is a good feature. What happens if nothing is selected. It might be better to use ';' to indicate the end of SQL.

Most SQL editors follow that convention - http://sqlfiddle.com/#!9/a6c585/1

xueyumusic:
Hi, @kishoreg , if nothing is selected it behaves the same as currently, which uses the whole content in editor as single query. If selecting some part, the selected part is used as single query. The ';' ending looks supported right now for both whole content and select part, it just an optional not a force rule. Thanks.

npawar:

PR #4387, opened 2019-06-30
Adding Calcite SQL Parser and make another entry point to query Pinot

fx19880617:
As part of https://github.com/winedepot/pinot/issues/14, this PR implements a Calcite SQL parser to compile SQL to PinotQuery.

Something to be noticed as syntax backward incompatible with PQL.
1. '!=' is not supported, need to replace it with '<>';
2. Double quote is used for identifier;
3. Single quotes are used for string literal and it's part of the parsed string. It's up to us to keep it or remove it.

Currently I removed the single quote from StringLiteral to match PQL parsing behavior.
See: 

> org.apache.pinot.common.utils.request.RequestUtils 

`literal.setStringValue(node.toString().replaceAll("^\'|\'$", ""));`


pinot-common/src/main/java/org/apache/pinot/sql/parsers/CalciteSqlParser.java
```diff
@@ -0,0 +1,273 @@
...
+  //
+  // Multiple OPTIONs is also supported by:
+  // either
+  //   `OPTION (<k1> = <v1>, <k2> = <v2>, <k3> = <v3>)`
+  // or
+  //   `OPTION (<k1> = <v1>) OPTION (<k2> = <v2>) OPTION (<k3> = <v3>)`
+  private static final Pattern OPTIONS_REGEX_PATTEN =
+      Pattern.compile("option\\s*\\(([^\\)]+)\\)", Pattern.CASE_INSENSITIVE);
+
+  public static PinotQuery compileToPinotQuery(String sql)
```
kishoreg: Why static methods?

pinot-common/src/test/resources/pql_queries.list
```diff
@@ -812,3 +812,8 @@ SELECT column1, column5, column6 FROM testTable ORDER BY column5, column9 WHERE
 select $docId, $segmentName, $hostName from mytable where $docId < 5 limit 50
 select $docId, $segmentName, $hostName from mytable where $docId = 5 limit 50
 select $docId, $segmentName, $hostName from mytable where $docId > 19998 limit 50
+SELECT SUM(met) FROM mytable WHERE timeConvert(DaysSinceEpoch,'DAYS','SECONDS') = 1394323200
```
kishoreg: Map UDF is not yet merged

pinot-api/src/main/java/org/apache/pinot/client/JsonAsyncHttpPinotClientTransport.java
```diff
@@ -83,6 +88,22 @@ public BrokerResponse executeQuery(String brokerAddress, String query)
     }
   }
 
+  @Override
+  public BrokerResponse executeSqlQuery(String brokerAddress, String query)
```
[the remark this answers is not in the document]
fx19880617: Make Query an object

pinot-common/src/test/resources/pql_queries.list
fx19880617: it's ok, this is just for compile, not evaluate.

pinot-broker/src/test/java/org/apache/pinot/broker/request/PqlAndCalciteSqlCompatibleTest.java
```diff
@@ -0,0 +1,322 @@
...
+        Assert.assertTrue(validate(brokerRequestFromPQL, brokerRequestFromSQL));
+        seqId++;
+      } catch (Exception e) {
+        LOGGER.error("Failed to compare results from \n\tPQL: {}\nand\n\tSQL: {}", pql, sql, e);
+        throw e;
+      }
+    }
+  }
+
+  public boolean validate(BrokerRequest br1, BrokerRequest br2) {
```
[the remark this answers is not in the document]
fx19880617: done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/PinotQueryParserFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.pinot.sql.parsers.CalciteSqlCompiler;
+
+import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.PQL;
+import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.SQL;
+
+
+public class PinotQueryParserFactory {
+
+  private static final Pql2Compiler PQL_2_COMPILER = new Pql2Compiler();
+  private static final CalciteSqlCompiler CALCITE_SQL_COMPILER = new CalciteSqlCompiler();
```
mayankshriv: Is the CalciteSqlCompiler thread safe?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -338,6 +337,13 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
     return brokerResponse;
   }
 
+  private PinotQueryRequest getPinotQueryRequest(JsonNode request) {
+    if (request.has(SQL)) {
```
mayankshriv: Would be good to assert that only one of SQL|PQL is specified in the request.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -42,7 +43,7 @@ public FilterQueryTree optimize(FilterQueryOptimizerRequest request) {
 
   private FilterQueryTree optimize(FilterQueryTree filterQueryTree, FilterQueryTree parent) {
     if (filterQueryTree.getOperator() == FilterOperator.OR) {
-      Map<String, Set<String>> columnToValues = new HashMap<>();
+      Map<String, Set<String>> columnToValues = new TreeMap<>();
```
mayankshriv: Would be nice to add a comment on why TreeMap (ordering) is needed.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/PinotQueryParserFactory.java
mayankshriv: Is the CalciteSqlCompiler thread safe?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/PinotQueryRequest.java
```diff
@@ -0,0 +1,45 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.broker.requesthandler;
+
+public class PinotQueryRequest {
+  private String queryFormat;
```
mayankshriv: Name member variables starting with '_' and avoid using `this.`.

pinot-broker/src/test/java/org/apache/pinot/broker/request/PqlAndCalciteSqlCompatibilityTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+  // OPTIMIZER is used to flatten certain queries with filtering optimization.
+  // The reason is that SQL parser will parse the structure into a binary tree mode.
+  // PQL parser will flat the case of multiple children under AND/OR.
+  // After optimization, both BrokerRequests from PQL and SQL should look the same and be easier to compare.
+  private static final BrokerRequestOptimizer OPTIMIZER = new BrokerRequestOptimizer();
+  private static final Logger LOGGER = LoggerFactory.getLogger(PqlAndCalciteSqlCompatibilityTest.class);
+
+  @Test
+  public void testSinglePqlAndSqlCompatible() {
```
mayankshriv: Is this testing something different then what is covered in testPqlAndSqlCompatible()?

pinot-broker/src/test/java/org/apache/pinot/broker/request/PqlAndCalciteSqlCompatibilityTest.java
```diff
@@ -0,0 +1,123 @@
...
+    LOGGER.info("Trying to compare BrokerRequest -\nFrom PQL: {}\nFrom SQL: {}", brokerRequestFromPQL,
+        brokerRequestFromSQL);
+    Assert.assertTrue(BrokerRequestComparisonUtils.validate(brokerRequestFromPQL, brokerRequestFromSQL));
+  }
+
+  @Test
+  public void testPqlAndSqlCompatible()
+      throws Exception {
+    final BufferedReader brPql = new BufferedReader(new InputStreamReader(
+        PqlAndCalciteSqlCompatibilityTest.class.getClassLoader().getResourceAsStream("pql_queries.list")));
```
mayankshriv: Are the queries.list comprehensive enough to test all the various PQL syntax features?

pinot-common/pom.xml
```diff
@@ -96,6 +96,10 @@
       <groupId>org.antlr</groupId>
       <artifactId>antlr4-runtime</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.calcite</groupId>
+      <artifactId>calcite-core</artifactId>
```
mayankshriv: Does this have a lot more functionality than the complier that we won't need? If so, is there a different artifact to include that just has what we need?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
fx19880617: Since sql support is newly added, I feel it's ok to keep sql parsing as first priority comparing to PQL.
So existing PQL user won't get affected.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
fx19880617: It's not needed anymore after rewrote the comparison method.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/PinotQueryParserFactory.java
fx19880617: Yes. Need to create a SqlParser for each query. Please see the method `compileCalciteSqlToPinotQuery` in `CalciteSqlParser.java`

pinot-broker/src/test/java/org/apache/pinot/broker/request/PqlAndCalciteSqlCompatibilityTest.java
fx19880617: This test is more or less a simple example for devs to test out a query and make debug easier without going through the entire file.

pinot-broker/src/test/java/org/apache/pinot/broker/request/PqlAndCalciteSqlCompatibilityTest.java
fx19880617: I don't think this will cover full syntax features. I've added some more queries into it. We should keep adding more query patterns.

pinot-common/pom.xml
fx19880617: this is the artifact we need in order to use SqlParser(`https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/parser/SqlParser.java`).

pinot-broker/src/test/java/org/apache/pinot/broker/request/PqlAndCalciteSqlCompatibilityTest.java
mayankshriv: Ok, we should ensure that we have full coverage before we advertise the SQL option. Let's open a issue to track that.

pinot-common/pom.xml
mayankshriv: So are we bringing in a lot more (in calcite-core) than we need?

fx19880617: yes, we may exclude the unnecessary dependencies once conflict is found.

pinot-broker/src/test/java/org/apache/pinot/broker/request/PqlAndCalciteSqlCompatibilityTest.java
fx19880617: will do

PR #4388, opened 2019-07-02
Adding support for Map type fields

kishoreg:
This leverages existing features MultiValue feature in Pinot. A Map is represented as two Multi-valued columns - __KEYS and __VALUES. This convention is used in only one place - real-time ingestion and segment generation. When pinot sees a FieldSpec name that ends with __KEYS or __VALUES, it checks for the field in the input Record. If that object turns out to be a map, then we convert it into two columns -> keyArray and valueArray.

Once we ingest a map as two arrays, we can access map['keyName']  via udf map_value(keyColumnName, 'KeyName', valueColumnName). We can support the simpler syntax when we move to calcite.

Sample queries

```select map_value(myMap__KEYS, 'k1', myMap__VALUES) from myTable```

```select count(*) from myTable group by map_value(myMap__KEYS, 'k1', myMap__VALUES)```

```select count(*) from myTable where map_value(myMap__KEYS, 'k1', myMap__VALUES) = 'v1'```

NOTE:
While this solution is not elegant, it is a good first step towards supporting Map and Struct data types.
Supporting Map as a DataType was bit invasive and requires us to first introduce the concept of CompositeColumnDatasource.

### CODE REVIEW

File: pinot-core/src/main/java/org/apache/pinot/core/data/readers/AvroRecordReader.java [hunk 1]
```diff
@@ -81,8 +82,9 @@ public GenericRow next(GenericRow reuse)
       String fieldName = fieldSpec.getName();
       Object value = _reusableAvroRecord.get(fieldName);
       // Allow default value for non-time columns
+      //Why special treatment for Time Column?
```
mayankshriv: Open an issue so someone can respond to why was time column given special treatment. Hopefully, they can also add comment in the code explaining the same.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/MapValueTransformFunction.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.annotation.Nonnull;
+import org.apache.pinot.core.common.DataSource;
+import org.apache.pinot.core.operator.blocks.ProjectionBlock;
+import org.apache.pinot.core.operator.transform.TransformResultMetadata;
+import org.apache.pinot.core.plan.DocIdSetPlanNode;
+import org.apache.pinot.core.segment.index.readers.Dictionary;
+
+
+/**
+ * map_value(keyColName, 'keyName', valColName)
```
mayankshriv: Nit: can this be changed to `map_value('keyName', 'keyColName', valColName)` so that the column names are together?

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/MapValueTransformFunction.java
```diff
@@ -0,0 +1,91 @@
...
+  private int[] _outputValueDictIds;
+
+  @Override
+  public String getName() {
+    return FUNCTION_NAME;
+  }
+
+  @Override
+  public void init(@Nonnull List<TransformFunction> arguments, @Nonnull Map<String, DataSource> dataSourceMap) {
+    _keyColumnFunction = arguments.get(0);
```
mayankshriv: Add precondition on arguments.size() == 3?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
```diff
@@ -268,7 +268,7 @@ public int indexOfSV(Object value) {
         throw new UnsupportedOperationException("Unsupported data type : " + _fieldSpec.getDataType());
     }
 
-    Arrays.sort(indexes);
+//    Arrays.sort(indexes);
```
mayankshriv: Remove instead. Add comment in the PR description of this behavior change. Perhaps this should be done as a pre-requisite PR?

pinot-core/src/main/java/org/apache/pinot/core/util/AvroUtils.java
```diff
@@ -296,4 +302,33 @@ public static boolean isSingleValueField(Field field) {
       return fieldSchema;
     }
   }
+
+  public static void extractField(FieldSpec incomingFieldSpec, GenericRecord from, GenericRow to) {
+    String fieldName = incomingFieldSpec.getName();
+    //Handle MAP types
+    if (fieldName.toUpperCase().endsWith("__KEYS")) {
```
mayankshriv: Define __KEYS and __VALUES in CommonConstants.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/MapTypeClusterIntegrationTest.java
```diff
@@ -0,0 +1,246 @@
...
+import org.testng.annotations.Test;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.collect.Lists;
+
+
+public class MapTypeClusterIntegrationTest extends BaseClusterIntegrationTest {
```
mayankshriv: Does this need to be integration test? Or can we get the same coverage with a unit test that generates a couple of segments and performs query execution on top of it. That won't be testing the broker side merge, but unsure if that is warranted.

kishoreg: This takes care of testing ingestion and query together.

kishoreg:
> 1. Please open issues for things not handled in the PR.
Will create issues for
- supporting map type syntax map['key']. This will be handled automatically when we start using calcite.
Anything else you can think of?
> 2. The __KEYS and __VALUES columns have specific semantics in the code, however, they are stored as MV columns. Are there any issues in allowing these columns to be treated as regular MV columns when the queries are not of the form specified in the PR description?
Not really.
> 3. Would be good to add a real world example that this feature can help solve, in the description.
- will update the docs. I think the need for map type is pretty obvious.  classic cases - tags/labels, sparse dimensions, etc.


-

pinot-core/src/main/java/org/apache/pinot/core/util/AvroUtils.java
kishoreg: I will define them within the same class. this is very specific to the ingestion layer and we don't need it anywhere else.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentDictionaryCreator.java
kishoreg: Done.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/MapValueTransformFunction.java
kishoreg: I was trying to follow the syntax of map_column['key']. we need colName first and then keyName.

pinot-core/src/main/java/org/apache/pinot/core/data/readers/AvroRecordReader.java
kishoreg: I think I got the answer in NullValueTransformer PR.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/MapValueTransformFunction.java
kishoreg: Done

mayankshriv:
This PR seems to have commits that are already part of apache/pinot, and not related to map-type support. Probably the PR was not created correctly?

PR #4392, opened 2019-07-02
Allow customized metrics prefix in pinot controller/broker/server

jamesyfshao:


jamesyfshao:
pr aim at solving [4391](https://github.com/apache/incubator-pinot/issues/4391)

pinot-common/src/main/java/org/apache/pinot/common/metrics/ControllerMetrics.java
```diff
@@ -25,8 +25,18 @@
  * Metrics for the controller.
  */
 public class ControllerMetrics extends AbstractMetrics<AbstractMetrics.QueryPhase, ControllerMeter, ControllerGauge, ControllerTimer> {
+
+  // FYI this is incorrect as it generate metrics named without a dot after pinot.controller part,
+  // but we keep this default for backward compatibility in case someone relies on this format
+  // see Servermetrics or BrokerMetrics class for correct prefix should be use
+  public static final String METRICS_PREFIX_DEFAULT = "pinot.controller";
```
mcvsubbu: Missing "."

pinot-common/src/main/java/org/apache/pinot/common/metrics/MetricsHelper.java
```diff
@@ -48,7 +48,7 @@
       new WeakHashMap<MetricsRegistryRegistrationListener, Object>();
 
   /**
-   * Initializes the metrics system by initializing the registry registration listeners present in the configuration.
+   * Initializes the metrics system by initializing the registry rePinotJmxMetricsReporter is startedgistration listeners present in the configuration.
```
mcvsubbu: ??

pinot-common/src/main/java/org/apache/pinot/common/metrics/ControllerMetrics.java
jamesyfshao: see the comment, it is mostly for backward compatible as the old prefix don't have the-dot

pinot-common/src/main/java/org/apache/pinot/common/metrics/MetricsHelper.java
jamesyfshao: probably accidental copy-paste, will fix this

jamesyfshao:
> What is the motivation behind this change?
> Also, what about minion?

There are two reasons we are planning this change:
1. we used to rely on JMX trans for sending metrics from jmx interface to our internal metrics system, however this has been a very annoying thing to maintenance and make it very hard for us to catch up to the latest metrics added to pinot system. So we are migrating to direct metrics reporter that allow us to grab metrics from registry and send to our internal metrics report system. However, our internal metrics system expect a different prefix compared to the one we defined here and it will be great if we can customize the prefix with config

2. well, anything configurable is better than hard-coded values for flexibility :)

for minion we are not really using it, so we are not sure if our change would be correct or not. we could make similar change to minion if you think it is necessary, but it would be a bit hard for us to test it

mcvsubbu:
> > What is the motivation behind this change?
> > Also, what about minion?
> 
> There are two reasons we are planning this change:
> 
> 1. we used to rely on JMX trans for sending metrics from jmx interface to our internal metrics system, however this has been a very annoying thing to maintenance and make it very hard for us to catch up to the latest metrics added to pinot system. So we are migrating to direct metrics reporter that allow us to grab metrics from registry and send to our internal metrics report system. However, our internal metrics system expect a different prefix compared to the one we defined here and it will be great if we can customize the prefix with config
> 2. well, anything configurable is better than hard-coded values for flexibility :)

Not all the time. It is good to have a product with fewer knobs so a rollout can be faster, and adoption better.

> 
> for minion we are not really using it, so we are not sure if our change would be correct or not. we could make similar change to minion if you think it is necessary, but it would be a bit hard for us to test it

If we are supporting configurable prexies, might as well support it for all. Yes, please make change to minion as well

pinot-common/src/main/java/org/apache/pinot/common/metrics/ControllerMetrics.java
mcvsubbu: thanks

jamesyfshao:
> > > What is the motivation behind this change?
> > > Also, what about minion?
> > 
> > 
> > There are two reasons we are planning this change:
> > 
> > 1. we used to rely on JMX trans for sending metrics from jmx interface to our internal metrics system, however this has been a very annoying thing to maintenance and make it very hard for us to catch up to the latest metrics added to pinot system. So we are migrating to direct metrics reporter that allow us to grab metrics from registry and send to our internal metrics report system. However, our internal metrics system expect a different prefix compared to the one we defined here and it will be great if we can customize the prefix with config
> > 2. well, anything configurable is better than hard-coded values for flexibility :)
> 
> Not all the time. It is good to have a product with fewer knobs so a rollout can be faster, and adoption better.
> 
> > for minion we are not really using it, so we are not sure if our change would be correct or not. we could make similar change to minion if you think it is necessary, but it would be a bit hard for us to test it
> 
> If we are supporting configurable prexies, might as well support it for all. Yes, please make change to minion as well

updated for minion metrics

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -293,6 +294,7 @@ public ServerType getServerType() {
     public static final String METADATA_EVENT_OBSERVER_PREFIX = "metadata.event.notifier";
 
     // Config keys
+    public static final String METRICS_PREFIX_CONFIG = "metrics.prefix.config";
```
mcvsubbu: Can you rename this to "metricsPrefix" instead of "metrics.prefix.config"?

It is unfortunate that minion went on a different route for configuring metricsRegistryRegistrationListener (basically, minion picks up metrics related configs without any prefix string for the configuration keys of metrics related items. The other components have "pinot.controller.metrics.foo", "pinot.broker.metrics.foo", etc.). minion chooses to pick the listeners from the top-level config. Ideally, I would prefer deprecating the old config for minion (with empty prefix) in favor of the new one  (with a prefix "pinot.,minion.metrics", like the others) for one release and then removing the old one altogether. Of course, this will mean changing minion to parse the new configs if they are there, and then falling back to the old one, and then applying any defaults.

@Jackie-Jiang  can you chime in as to why minion took this route? Any thoughts?

mcvsubbu:
Sorry, that was not an approve. Misclicked. Please wait unti merge,

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
jamesyfshao: updated the metrics prefix naming per suggestion

mcvsubbu:
lgtm, unless @Jackie-Jiang has some other thoughts on minion.

pinot-common/src/main/java/org/apache/pinot/common/metrics/BrokerMetrics.java
```diff
@@ -26,17 +26,25 @@
  *
  */
 public class BrokerMetrics extends AbstractMetrics<BrokerQueryPhase, BrokerMeter, BrokerGauge, BrokerTimer> {
+
+  public static final String METRICS_PREFIX_DEFAULT = "pinot.broker.";
```
Jackie-Jiang: Can you add these constants into CommonConstants? Rename to DEFAULT_METRICS_PREFIX and DEFAULT_GLOBAL. Same for other classes

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -124,6 +124,7 @@ public ServerType getServerType() {
     public static final String ROUTING_TABLE_CONFIG_PREFIX = "pinot.broker.routing.table";
     public static final String ACCESS_CONTROL_CONFIG_PREFIX = "pinot.broker.access.control";
     public static final String METRICS_CONFIG_PREFIX = "pinot.broker.metrics";
+    public static final String METRICS_NAME_PREFIX = "pinot.broker.metrics.prefix";
```
Jackie-Jiang: CONFIG_OF_METRICS_PREFIX?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -293,6 +294,7 @@ public ServerType getServerType() {
     public static final String METADATA_EVENT_OBSERVER_PREFIX = "metadata.event.notifier";
 
     // Config keys
+    public static final String METRICS_PREFIX_CONFIG = "metricsPrefix";
```
Jackie-Jiang: METRICS_PREFIX_KEY?

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerConf.java
```diff
@@ -57,6 +58,7 @@
   private static final String CONSOLE_WEBAPP_USE_HTTPS = "controller.query.console.useHttps";
   private static final String EXTERNAL_VIEW_ONLINE_TO_OFFLINE_TIMEOUT = "controller.upload.onlineToOfflineTimeout";
   private static final String CONTROLLER_MODE = "controller.mode";
+  private static final String CONTROLLER_METRICS_PREFIX = "controller.metrics.prefix";
```
Jackie-Jiang: Should be in CommonConstants

jamesyfshao: updated

jamesyfshao:
@Jackie-Jiang updated a bunch of place per your feedback, please take a look when you have time. Thanks!

pinot-common/src/main/java/org/apache/pinot/common/metrics/ServerMetrics.java
```diff
@@ -42,10 +46,15 @@
   }
 
   public ServerMetrics(MetricsRegistry metricsRegistry) {
```
Jackie-Jiang: (nit) Can you please move the constructors to the top？

pinot-broker/src/main/java/org/apache/pinot/broker/broker/BrokerServerBuilder.java
```diff
@@ -70,7 +70,8 @@ public BrokerServerBuilder(Configuration config, RoutingTable routingTable, Time
     MetricsHelper.initializeMetrics(config.subset(Broker.METRICS_CONFIG_PREFIX));
     MetricsHelper.registerMetricsRegistry(_metricsRegistry);
     _brokerMetrics =
-        new BrokerMetrics(_metricsRegistry, !_config.getBoolean(Broker.CONFIG_OF_ENABLE_TABLE_LEVEL_METRICS, true));
+        new BrokerMetrics(config.getString(Broker.CONFIG_OF_METRICS_NAME_PREFIX, Broker.DEFAULT_METRICS_NAME_PREFIX), _metricsRegistry,
+            !_config.getBoolean(Broker.CONFIG_OF_ENABLE_TABLE_LEVEL_METRICS, true));
```
Jackie-Jiang: Change the true to !Broker.DEFAULT_METRICS_GLOBAL_ENABLED?

PR #4396, opened 2019-07-03
make kafka version number controlled by config number

jamesyfshao:


pinot-integration-tests/pom.xml
```diff
@@ -192,7 +192,7 @@
     <!-- TODO: figure out a way to avoid explicitly setting the kafka module dependency, and instead inject it, such that any kafka stream implementation can use this -->
```
npawar: Thanks for this change!
Please remove this TODO from the pom.xml in pinot-intergation-tests, pinot-tools, pinot-perf, and put it in the main pom.xml

jamesyfshao: updated per your feedback

PR #4398, opened 2019-07-03
Add more key/value pairs into LOG2M_TO_SIZE_IN_BYTES in HllSizeUtils

bozhang2820:
This change would allow more options for log2m in HLL config.


bozhang2820:
> This is ok but what are we gaining by this change. Is the accuracy changing dramatically

Yes without this change the best standard error (1.04 / sqrt(m)) is at ~4.6% when log2m = 9. With this change it goes down to ~0.57% (when log2m = 15).

In fact at Uber we have a table configured with log2m = 13 and we have a similar patch to accommodate that.

PR #4399, opened 2019-07-04
Moving handling of NULL values from RecordReaders to NullValueTransfo…

kishoreg:
…rmer

Prior to this PR, its the responsibility of each recordReader to populate defaulValues for every field in schema. This PR moves that logic into NullValueTransformer where we set defaultNullValues appropriately.

This is a precursor to supporting NULLs in Pinot.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
@@ -124,7 +124,6 @@ public class JsonUtils {
       if (jsonValue != null && !jsonValue.isNull()) {
         return extractSingleValue(jsonValue, fieldSpec.getDataType());
       } else {
-//        return fieldSpec.getDefaultNullValue();
         return null;
       }
     } else {
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
@@ -139,14 +139,12 @@ public class JsonUtils {
             return values;
           } else {
             return null;
-//            return new Object[]{fieldSpec.getDefaultNullValue()};
           }
         } else {
           return new Object[]{extractSingleValue(jsonValue, fieldSpec.getDataType())};
         }
       } else {
         return null;
-        //return new Object[]{fieldSpec.getDefaultNullValue()};
       }
     }
   }
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
@@ -124,7 +124,8 @@ public class JsonUtils {
       if (jsonValue != null && !jsonValue.isNull()) {
         return extractSingleValue(jsonValue, fieldSpec.getDataType());
       } else {
-        return fieldSpec.getDefaultNullValue();
+//        return fieldSpec.getDefaultNullValue();
+        return null;
       }
     } else {
       if (jsonValue != null && !jsonValue.isNull()) {
@@ -137,13 +138,15 @@ public class JsonUtils {
             }
             return values;
           } else {
-            return new Object[]{fieldSpec.getDefaultNullValue()};
+            return null;
+//            return new Object[]{fieldSpec.getDefaultNullValue()};
           }
         } else {
           return new Object[]{extractSingleValue(jsonValue, fieldSpec.getDataType())};
         }
       } else {
-        return new Object[]{fieldSpec.getDefaultNullValue()};
+        return null;
+        //return new Object[]{fieldSpec.getDefaultNullValue()};
       }
     }
   }
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderUtils.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderUtils.java
@@ -111,9 +111,6 @@ public class RecordReaderUtils {
    */
   public static Object convertSingleValue(FieldSpec fieldSpec, @Nullable Object value) {
     if (value == null) {
-      // Do not allow default value for time column
-//      assert fieldSpec.getFieldType() != FieldSpec.FieldType.TIME;
-//      return fieldSpec.getDefaultNullValue();
       return null;
     }
     if (value instanceof GenericData.Record) {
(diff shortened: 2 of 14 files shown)


pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
```diff
@@ -137,13 +138,15 @@ public static Object extractValue(@Nullable JsonNode jsonValue, FieldSpec fieldS
             }
             return values;
           } else {
-            return new Object[]{fieldSpec.getDefaultNullValue()};
+            return null;
```
Jackie-Jiang: Don't know if we should support zero-length array for MV columns, but it is not the same as null.

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
```diff
@@ -0,0 +1,53 @@
...
+
+  private final Collection<FieldSpec> _fieldSpecs;
+  private Schema _schema;
+
+  public NullValueTransformer(Schema schema) {
+    _schema = schema;
+    _fieldSpecs = _schema.getAllFieldSpecs();
+  }
+
+  @Nullable
```
Jackie-Jiang: Remove this annotation as the result will not be null.

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  public GenericRow transform(GenericRow row) {
+    for (FieldSpec fieldSpec : _fieldSpecs) {
+      String fieldName = fieldSpec.getName();
+      if (row.getValue(fieldName) == null && fieldSpec.getFieldType() != FieldSpec.FieldType.TIME) {
```
Jackie-Jiang: Make the `fieldSpec.getFieldType() != FieldSpec.FieldType.TIME` as a precondition?

pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderUtils.java
```diff
@@ -112,8 +112,9 @@ public static Object convert(FieldSpec fieldSpec, @Nullable Object value) {
   public static Object convertSingleValue(FieldSpec fieldSpec, @Nullable Object value) {
     if (value == null) {
       // Do not allow default value for time column
```
jackjlli: Move this comment or add a description to `NullValueTransformer`.

pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderUtils.java
```diff
@@ -112,8 +112,9 @@ public static Object convert(FieldSpec fieldSpec, @Nullable Object value) {
...
-      assert fieldSpec.getFieldType() != FieldSpec.FieldType.TIME;
-      return fieldSpec.getDefaultNullValue();
+//      assert fieldSpec.getFieldType() != FieldSpec.FieldType.TIME;
```
jackjlli: Clean up these lines. Same for the rest commented statements.

pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
kishoreg: I dint follow. I am doing something similar to this in NullValueTransformer
```
          row.putField(fieldName, new Object[]{fieldSpec.getDefaultNullValue()});
```

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
kishoreg: Done!

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
kishoreg: Why should it be a precondition?

pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderUtils.java
kishoreg: Done.

pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReaderUtils.java
kishoreg: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
Jackie-Jiang: IMO empty array should be a valid value for MV columns instead of treated as null. We don't have that support now, but maybe better to add some comments or TODOs

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
Jackie-Jiang: Time column should never have null value. Passing null as time value will cause exception later

kishoreg: Not sure if its the responsibility of the transformer to verify that. We can't do much in real-time case if the time column is null i.e. throwing Exception will not help.

pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
kishoreg: Want to maintain parity with the existing functionality. We can solve this in another PR.

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
Jackie-Jiang: We assume no null value in the following transformers, so this transformer should guarantee that, or throw exception. Performing a check and throw exception here is more clear than getting NPE somewhere in the following transformers, and easier to debug.

pinot-common/src/main/java/org/apache/pinot/common/utils/JsonUtils.java
```diff
@@ -124,7 +124,8 @@ public static Object extractValue(@Nullable JsonNode jsonValue, FieldSpec fieldS
       if (jsonValue != null && !jsonValue.isNull()) {
         return extractSingleValue(jsonValue, fieldSpec.getDataType());
       } else {
-        return fieldSpec.getDefaultNullValue();
+//        return fieldSpec.getDefaultNullValue();
```
Jackie-Jiang: Remove this line

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
kishoreg: what will happen in the real-time case when an exception is thrown

Jackie-Jiang: Actually TimeTransformer can handle null values correctly, and throw exception if both incoming and outgoing values are null, so we are good here.

PR #4405, opened 2019-07-08
[TE] add a thread pool to run preview tasks

jihaozh:
This PR adds a thread pool to run preview tasks for the dashboard. It adds the following limits to prevent excessive server usage.
1. The number of preview task running at the same time 
2. The timeout for a preview task

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
@@ -57,6 +57,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.pinot.thirdeye.anomaly.task.TaskConstants;
 import org.apache.pinot.thirdeye.api.Constants;
 import org.apache.pinot.thirdeye.auth.ThirdEyePrincipal;
+import org.apache.pinot.thirdeye.dashboard.DetectionPreviewConfiguration;
 import org.apache.pinot.thirdeye.dataframe.util.MetricSlice;
 import org.apache.pinot.thirdeye.datalayer.bao.DatasetConfigManager;
 import org.apache.pinot.thirdeye.datalayer.bao.DetectionAlertConfigManager;
@@ -119,10 +120,6 @@ public class YamlResource {
   private static final String PROP_SESSION_KEY = "sessionKey";
   private static final String PROP_PRINCIPAL_TYPE = "principalType";
   private static final String PROP_SERVICE = "SERVICE";
-  // 5 detection previews are running at the same time at most
-  private static final int PREVIEW_PARALLELISM = 5;
-  // max time allowed for a preview task
-  private static final long PREVIEW_TIMEOUT = TimeUnit.MINUTES.toMillis(5);
 
   // default onboarding replay period
   private static final long ONBOARDING_REPLAY_LOOKBACK = TimeUnit.DAYS.toMillis(30);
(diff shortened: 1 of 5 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -145,6 +154,7 @@ public YamlResource() {
     this.evaluationDAO = DAORegistry.getInstance().getEvaluationManager();
     this.sessionDAO = DAORegistry.getInstance().getSessionDAO();
     this.yaml = new Yaml();
+    this.executor = Executors.newFixedThreadPool(PREVIEW_PARALLELISM);
```
akshayrai: How about a using a `newCachedThreadPool`? That way the calls won't easily timeout at peak traffic.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -115,6 +119,10 @@
   private static final String PROP_SESSION_KEY = "sessionKey";
   private static final String PROP_PRINCIPAL_TYPE = "principalType";
   private static final String PROP_SERVICE = "SERVICE";
+  // 5 detection previews are running at the same time at most
+  private static final int PREVIEW_PARALLELISM = 5;
+  // max time allowed for a preview task
+  private static final long PREVIEW_TIMEOUT = TimeUnit.MINUTES.toMillis(5);
```
akshayrai: Can we make this configurable?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -724,13 +735,18 @@ private Response runPreview(long start, long end,
...
+      future = this.executor.submit(pipeline::run);
+      result = future.get(PREVIEW_TIMEOUT, TimeUnit.MILLISECONDS);
     } catch (IllegalArgumentException e) {
       return processBadRequestResponse(YamlOperations.PREVIEW.name(), YamlOperations.RUNNING.name(), payload, e);
     } catch (InvocationTargetException e) {
       responseMessage.put("message", "Failed to run the preview due to " + e.getTargetException().getMessage());
       return Response.serverError().entity(responseMessage).build();
+    } catch (TimeoutException e) {
+      // stop the preview
+      future.cancel(true);
```
akshayrai: would be safer to move this to a finally block?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
jihaozh: Actually, in this case, we want to use a `newFixedThreadPool`, because we would like to control the maximum number of preview tasks to keep it from eating all dashboard resource.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
jihaozh: sure

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
jihaozh: good idea. updated.

PR #4416, opened 2019-07-10
Improve controller tests

Jackie-Jiang:


Jackie-Jiang:
Close the temporary testing PR. The fix is merged in: #4426 #4430 #4431

PR #4418, opened 2019-07-10
Fix potential resource leak in the way we close a collection of closeables

siddharthteotia:
Fix for issue outlined here -- https://github.com/apache/incubator-pinot/issues/4417


pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
```diff
@@ -0,0 +1,61 @@
...
+ * under the License.
+ */
+package org.apache.pinot.common;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Arrays;
+
+
+public class Closeables {
```
jackjlli: It'd be great to add some descriptions on where/where to use this class.

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
```diff
@@ -0,0 +1,61 @@
...
+        if (topLevelException == null) {
+          topLevelException = e;
+        } else if (e != topLevelException) {
+          topLevelException.addSuppressed(e);
+        }
+      }
+    }
+
+    if (topLevelException != null) {
+      throw topLevelException;
```
jackjlli: Maybe it's good to log some messages here?

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/FixedByteSingleValueMultiColWriter.java
```diff
@@ -101,7 +101,9 @@ public void setBytes(int row, int col, byte[] bytes) {
 
   public void close()
       throws IOException {
-    this.indexDataBuffer.close();
-    this.indexDataBuffer = null;
+    if (indexDataBuffer != null) {
```
Jackie-Jiang: indexDataBuffer can never be null. You might want to change all the member variables as final for this class

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
Jackie-Jiang: +1. I think the intention is to close all Closeables even if some of them throwing exception? Another way to handle this is by using try with multiple resources:
```
    try (Closeable resource1 = closeable1; Closeable resource2 = closeable2) {
    }
```

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
```diff
@@ -0,0 +1,61 @@
...
+
+
+public class Closeables {
+
+  public static void close(Iterable<? extends Closeable> ac)
+      throws IOException {
+
+    if (ac == null) {
+      return;
+    } else if (ac instanceof Closeable) {
```
Jackie-Jiang: This if can never be true right?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/fwd/SingleValueSortedForwardIndexCreator.java
```diff
@@ -56,10 +56,13 @@ public void index(int docId, int dictId) {
   public void close()
       throws IOException {
     int cardinality = _maxDocIds.length;
-    for (int i = 0; i < cardinality; i++) {
-      _writer.setInt(i, 0, _minDocIds[i]);
-      _writer.setInt(i, 1, _maxDocIds[i]);
+    try {
```
Jackie-Jiang: Use try with resource

pinot-core/src/main/java/org/apache/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
```diff
@@ -68,14 +68,14 @@ public InvertedIndexHandler(@Nonnull File indexDir, @Nonnull SegmentMetadataImpl
   }
 
   public void createInvertedIndices()
-      throws IOException {
+      throws Exception {
```
Jackie-Jiang: Why changing this?

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: Added javadoc.
Yes, try with resource is a good pattern without having to worry about closing everything in finally. However, it will work only for resources that are local -- as in local to the function that are created. But for the resources that member variables of the class and are released only during the call to close(),  we can't use try with resource for them.

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: Yes, changed the condition. I was initially thinking to use this method for both collection of closeable resources or single resource (passed as a list). May be as we go about fixing rest of the code in future to use the helper methods here, we can add or modify this method based on what caller might pass.

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: done

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/FixedByteSingleValueMultiColWriter.java
siddharthteotia: Done. I introduced the null check to make the close() is idempotent. But since the close() on PinotDataBuffer is idempotent, calling this close() multiple times will be a NO-OP. 
Made the member variables final. rowSizeInBytes can't be final as it is being incremented in the constructor.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/fwd/SingleValueSortedForwardIndexCreator.java
siddharthteotia: May be I am missing something but how can we use try with resource here? It can only be used if the resource is being declared/created there for immediate usage and released thereafter soon after the try block ends. In this case, we are dealing with resources that are member variables.

pinot-core/src/main/java/org/apache/pinot/core/segment/index/loader/invertedindex/InvertedIndexHandler.java
siddharthteotia: Sorry, this was a temp change. I forgot to cleanup before creating PR. 
Removed.

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
```diff
@@ -0,0 +1,93 @@
...
+public class Closeables {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(Closeables.class);
+
+  /**
+   * Close a collection of {@link Closeable} resources
+   * @param closeables collection of resources to close
+   * @throws IOException
+   */
+  public static void close(Iterable<? extends Closeable> closeables)
```
mcvsubbu: We have one class with one static function. Can we move this static function to FileUtils? I understand not all closeables are files ...

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
mcvsubbu: I like moving methods to common utils like this, except that we just cannot log additional information about the object we are trying to work on (in this case, a Closeable). All we will see in the log is "Failed to close resource" . Hopefully the exception provides more information, but if it is just (say) "No such connection" and does not give any source or destination address/hostname, then we are stuck with an ERROR  log that does not tell us much more.

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
Jackie-Jiang: > Added javadoc.
> Yes, try with resource is a good pattern without having to worry about closing everything in finally. However, it will work only for resources that are local -- as in local to the function that are created. But for the resources that member variables of the class and are released only during the call to close(), we can't use try with resource for them.

The resource doesn't need to be local. You just need to assign the member variables to local variables as in my example

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: @mcvsubbu , to see the the extent of logging, I created a simple test:

`public static void main(String[] args) throws Exception {
    Closeables.close(new Closeable1(), new Closeable2(), new Closeable3());
  }

  private static class Closeable1 implements Closeable {
    @Override
    public void close() throws IOException {
      throw new IOException("io exception 1");
    }
  }

  private static class Closeable2 implements Closeable {
    @Override
    public void close() throws IOException {
      throw new FileNotFoundException("io exception 2");
    }
  }

  private static class Closeable3 implements Closeable {
    @Override
    public void close() throws IOException {
      throw new IOException("io exception 3");
    }
  }`

**10:15:06.577 [main] ERROR org.apache.pinot.common.Closeables - Failed to close resource: class org.apache.pinot.core.segment.creator.impl.SegmentColumnarIndexCreator$Closeable1
java.io.IOException: io exception 1**
    at SegmentColumnarIndexCreator$Closeable1.close(SegmentColumnarIndexCreator.java:548)
    at org.apache.pinot.common.Closeables.close(Closeables.java:66) [classes/:?]
    at org.apache.pinot.common.Closeables.close(Closeables.java:94) [classes/:?]
    at SegmentColumnarIndexCreator.main(SegmentColumnarIndexCreator.java:542)

**10:15:06.583 [main] ERROR org.apache.pinot.common.Closeables - Failed to close resource: class org.apache.pinot.core.segment.creator.impl.SegmentColumnarIndexCreator$Closeable2
java.io.FileNotFoundException: io exception 2**
        at SegmentColumnarIndexCreator$Closeable2.close(SegmentColumnarIndexCreator.java:555) 
	at org.apache.pinot.common.Closeables.close(Closeables.java:66) [classes/:?]
	at org.apache.pinot.common.Closeables.close(Closeables.java:94) [classes/:?]
	at SegmentColumnarIndexCreator.main(SegmentColumnarIndexCreator.java:542) [classes/:?]

**10:15:06.584 [main] ERROR org.apache.pinot.common.Closeables - Failed to close resource: class org.apache.pinot.core.segment.creator.impl.SegmentColumnarIndexCreator$Closeable3
java.io.IOException: io exception 3**
	at SegmentColumnarIndexCreator$Closeable3.close(SegmentColumnarIndexCreator.java:562) 
	at org.apache.pinot.common.Closeables.close(Closeables.java:66) [classes/:?]
	at org.apache.pinot.common.Closeables.close(Closeables.java:94) [classes/:?]
	at SegmentColumnarIndexCreator.main(SegmentColumnarIndexCreator.java:542) [classes/:?]

Above shows how each exception on close() is logged along with the exact resource type

The following shows the final log message with all the above exceptions (last 2 being suppressed)

**11:02:02.846 [main] ERROR org.apache.pinot.common.Closeables - Failed to close one or more resources
java.io.IOException: io exception 1**
	at SegmentColumnarIndexCreator$Closeable1.close(SegmentColumnarIndexCreator.java:548) 
	at org.apache.pinot.common.Closeables.close(Closeables.java:65) [classes/:?]
	at org.apache.pinot.common.Closeables.close(Closeables.java:93) [classes/:?]
	at SegmentColumnarIndexCreator.main(SegmentColumnarIndexCreator.java:542) [classes/:?]
	**Suppressed: java.io.FileNotFoundException: io exception 2**
		at SegmentColumnarIndexCreator$Closeable2.close(SegmentColumnarIndexCreator.java:555) 
		at org.apache.pinot.common.Closeables.close(Closeables.java:65) [classes/:?]
		at org.apache.pinot.common.Closeables.close(Closeables.java:93) [classes/:?]
		at SegmentColumnarIndexCreator.main(SegmentColumnarIndexCreator.java:542) 
	**Suppressed: java.io.IOException: io exception 3**
		at SegmentColumnarIndexCreator$Closeable3.close(SegmentColumnarIndexCreator.java:562)
		at org.apache.pinot.common.Closeables.close(Closeables.java:65) [classes/:?]
		at org.apache.pinot.common.Closeables.close(Closeables.java:93) [classes/:?]
		at .SegmentColumnarIndexCreator.main(SegmentColumnarIndexCreator.java:542) 

Finally this is what the end user exception looks like when we throw:

**Exception in thread "main" java.io.IOException: io exception 1**
	at SegmentColumnarIndexCreator$Closeable1.close(SegmentColumnarIndexCreator.java:548)
	at org.apache.pinot.common.Closeables.close(Closeables.java:65)
	at org.apache.pinot.common.Closeables.close(Closeables.java:93)
	at .SegmentColumnarIndexCreator.main(SegmentColumnarIndexCreator.java:542)
	**Suppressed: java.io.FileNotFoundException: io exception 2**
		at SegmentColumnarIndexCreator$Closeable2.close(SegmentColumnarIndexCreator.java:555)
		... 3 more
	Suppressed: java.io.IOException: io exception 3
		at SegmentColumnarIndexCreator$Closeable3.close(SegmentColumnarIndexCreator.java:562)
		... 3 more

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/fwd/SingleValueSortedForwardIndexCreator.java
Jackie-Jiang:
```
try (Closeable closeable = _writer) {
  ...
}
```

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: The final exception that is logged by the logger shows the call stack of top level along with any of the suppressed exceptions. Also, I added a small change to log (including the name of closeable) as and when we catch an exception

Jackie-Jiang: I don't think we should log anything in this util class as it will throw the exception out and whoever handles the exception should log the exception

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
```diff
@@ -0,0 +1,93 @@
...
+      throws IOException {
+
+    if (closeables == null) {
```
Jackie-Jiang: (nit) closeables are not supposed to be null. For nullable argument, add annotation in the method signature

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: We can move this to common/utils in Fileutils and probably rename FileUtils to IOUtils. 

In future, I think we should consider making these classes implement AutoCloseable instead of Closeable as that gives us flexibility of handling exceptions other than IOException. For example a close() on PinotDataBuffer might throw unpredictable exceptions as we release native memory. Once we move to AutoCloseable, then I think we should have these utility methods in their own class.

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/fwd/SingleValueSortedForwardIndexCreator.java
siddharthteotia: Got it. Made the change

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: Will it be better if instead of having each caller logging explicitly, we just log the error (with the call stack of top level exception and any of suppressed exceptions) here?

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
siddharthteotia: Got it. Thank you

siddharthteotia:
Based on the above conversations, I have addressed the review comments in latest commit.

pinot-common/src/main/java/org/apache/pinot/common/Closeables.java
Jackie-Jiang: The convention should be whoever handles the exception logs the exception (or ignore the exception if the exception is expected). This way we can prevent log the same exception multiple times, and also the caller has more context on what triggers the exception.

pinot-common/src/main/java/org/apache/pinot/common/utils/FileUtils.java
```diff
@@ -64,4 +70,63 @@ public static void transferBytes(FileChannel src, long position, long count, Fil
...
+   * }
+   *
+   * The helper methods provided here do this safely
+   * while keeping track of exception(s) raised during
+   * close() of each resource and still continuing to close
+   * subsequent resources.
+   * @param closeables collection of resources to close
+   * @throws IOException
+   */
+  public static void close(@Nonnull Iterable<? extends Closeable> closeables)
```
Jackie-Jiang: We follow the convention of only annotating nullable argument and tread non-annotated argument as non-null. (We did annotate non-null before and found there are too many annotations and Nullable and Nonnull look very similar :-P)

siddharthteotia: done

siddharthteotia:
Rebased

siddharthteotia:
Can this be merged?

pinot-common/src/main/java/org/apache/pinot/common/utils/FileUtils.java
```diff
@@ -18,16 +18,22 @@
...
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 public class FileUtils {
   private FileUtils() {
   }
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(FileUtils.class);
```
Jackie-Jiang: Remove this line as logger is never used

pinot-common/src/main/java/org/apache/pinot/common/utils/FileUtils.java
```diff
@@ -64,4 +70,60 @@ public static void transferBytes(FileChannel src, long position, long count, Fil
...
+   * The helper methods provided here do this safely
+   * while keeping track of exception(s) raised during
+   * close() of each resource and still continuing to close
+   * subsequent resources.
+   * @param closeables collection of resources to close
+   * @throws IOException
+   */
+  public static void close(Iterable<? extends Closeable> closeables)
+      throws IOException {
+
```
Jackie-Jiang: (nit) Remove empty line

pinot-core/src/main/java/org/apache/pinot/core/io/writer/impl/FixedByteSingleValueMultiColWriter.java
```diff
@@ -99,9 +100,12 @@ public void setBytes(int row, int col, byte[] bytes) {
     indexDataBuffer.readFrom(offset, bytes);
   }
 
+  @Override
   public void close()
       throws IOException {
-    this.indexDataBuffer.close();
-    this.indexDataBuffer = null;
+    if (this.indexDataBuffer != null) {
```
Jackie-Jiang: You can make this final and remove the null check as PinotDataBuffer.close() is idempotent, your choice

siddharthteotia: I had initially done that -- removed the setting to null and made it final. However, @mcvsubbu brought up a scenario in context of multi-threading to explain why we were setting to null intially. Looks like if two threads race in a particular way -- where one does a close and does not set it to null and soon after some other thread invokes a set***() operation on this class, then it will use the non-null buffer and invoke some set operation on it which may lead to arbitrary errors/bugs as opposed to hitting NPE if the first thread sets it to null. Based on this I reverted back to old behavior and made it non-final . 

I agree that PinotDataBuffer.close() is final and so we don't need the if check.

Jackie-Jiang: In case of multi-threading, final is safer than non-volatile variable. I don't think we should throw NPE if the buffer is already closed, which could be confusing, and not align with Closable's usual behavior. I prefer the final + no null check way, @mcvsubbu what do you think?

mcvsubbu: It is a balance here.

Pinot query processing threads are not interruptable (at least now). The parent thread lets the child thread run with the query while returning to the broker after some timeout. It is therefore possible that a query thread is using a segment (even after it has been released).

A retention or segment drop can cause the segment parts to be closed, and a read request may come along, causing us to access bad memory, and a JVM crash could result (we use direct access to memory, in unsafe manner)

NPE on a close is a much smaller evil, and I can live with that. 

In the early days we even saw JVM crashes during segment unload.

siddharthteotia: One thought was that in case this multi-thread scenario does happen then NPE might be better than a core dump / segmentation fault (as a result of doing set operation on buffer that has been closed/released especially since all set/get operations on direct buffer are unsafe).

Jackie-Jiang: This should not happen because we keep reference count on the segments. While it is true that if our code has bug then this could potentially cause JVM crash, so it is clean + fail fast model vs. try to survive failure model. I'm okay both way, but in FixedByteValueReaderWriter we are taking the first way, so I prefer keeping them consistent.

mcvsubbu: @Jackie-Jiang, this can happen because refcnt is kept only until the end of the main thread of the query. Like I said, a child-thread could be running after the main thread reduces the refcnt. If there are other classes that are not following this, I suggest we bring them back to follow this model.

Jackie-Jiang: Good point. Then let's make the data buffer volatile and set it to null when close it. Can you also change the fixed one to follow this model, and add some comments describing why we are doing this?

siddharthteotia: I have made the changes in FixedByteSingleValueMultiColWriter, FixedByteSingleValueMultiColReader,FixedByteValueReaderWriter and FixedBitIntReaderWriter based on the model we agreed to.

pinot-common/src/main/java/org/apache/pinot/common/utils/FileUtils.java
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/common/utils/FileUtils.java
siddharthteotia: Done

PR #4419, opened 2019-07-10
[TE] Fix exception handling - Propagate and display the error message/exception on frontend

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/validators/DetectionConfigValidator.java
```diff
@@ -72,20 +72,15 @@ public DetectionConfigValidator(DataProvider provider) {
...
-      detectionConfig.setId(-1L);
-
-      // try to load the detection pipeline and init all the components
-      this.loader.from(provider, detectionConfig, 0, 0);
-
-      // set id back
-      detectionConfig.setId(id);
-    } catch (Exception e){
-      // exception thrown in validate pipeline via reflection
-      throw new IllegalArgumentException("Semantic error: " + e.getCause().getMessage());
```
jihaozh: will delete this swallow the error message?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -374,8 +379,8 @@ public Response createDetectionPipelineApi(
     }
 
     LOG.info("Detection created with id " + detectionConfigId + " using payload " + payload);
-    responseMessage.put("detectionMsg", "Alert was created successfully.");
-    responseMessage.put("detectionMsg-moreInfo", "Record saved with id " + detectionConfigId);
+    responseMessage.put("message", "Alert was created successfully.");
```
jihaozh: is there any frontend changes to adapt for this?

akshayrai: Frontend uses the toast popup to display this error message.

Earlier, we used to display the messages just below the editor(by parsing detectionMsg and subscriptionMsg) but that seems to be broken. We can clean that up in the frontend and just rely on the toast popup.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/validators/DetectionConfigValidator.java
akshayrai: We would catch re-throw only if we add some additional information that can help us debug. But there we are simply changing the exception type without enriching the message.

Without the catch, the exception will be propagated up and handled in YamlResource.

jihaozh: make sense.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
jihaozh: sounds good. this should work

PR #4422, opened 2019-07-11
In ClusterChangeMediator, stop enqueue/process changes if already stopped

Jackie-Jiang:


pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/ClusterChangeMediator.java
```diff
@@ -145,15 +145,15 @@ private void processClusterChange(ChangeType changeType, List<ClusterChangeHandl
   /**
    * Starts the cluster change mediator.
    */
-  public void start() {
+  public synchronized void start() {
```
jackjlli: Do we need synchronized block here? These start and stop methods would only be called in brokerStarter though. Or maybe just adding the sync block on `_stopped` would be enough?

Jackie-Jiang: Just for safety. This method will be called only once, so no performance impact

PR #4423, opened 2019-07-11
[TE] add ml auto config into YAML template

jihaozh:
Add the ml auto-config into the default YAML template.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/utils/constants.js
+++ b/thirdeye/thirdeye-frontend/app/utils/constants.js
@@ -27,6 +27,7 @@ rules:                            # Can configure multiple rules with "OR" relat
         configuration:
           bucketPeriod: P1D       # Use PT1H for hourly and PT5M for minute level (ingraph metrics) data.
           pValueThreshold: 0.05   # Higher value means more sensitive to small changes.
+          mlConfig: true          # The machine learning auto config to select and maintain the configuration with the best performance.
   filter:                         # Filter out anomalies detected by rules to reduce noise.
     - name: filter_rule_1
       type: PERCENTAGE_CHANGE_FILTER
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/app/utils/constants.js
```diff
@@ -27,6 +27,7 @@ rules:                            # Can configure multiple rules with "OR" relat
         configuration:
           bucketPeriod: P1D       # Use PT1H for hourly and PT5M for minute level (ingraph metrics) data.
           pValueThreshold: 0.05   # Higher value means more sensitive to small changes.
+          mlConfig: true          # The machine learning auto config to select and maintain the configuration with the best performance.
```
akshayrai: Can we set the default setting to true instead?

cyy0714: Thanks for the comment. The current plan is to change the template (Only affects the new alerts, and monitor the performance) and set the default value to true (GA to all alerts). That's why we decide to change the template for evaluation.

jihaozh: The team decided to put it into the template so that users are aware of this change and can opt to turn it off if it takes too much time in preview.

akshayrai: Putting this in the template will not guarantee that very new alert will have this setting. Users will also copy paste configs from wiki and other alerts. We can monitor the performance by cloning some of our existing alerts right?

I would still prefer testing it offline and setting the default to true rather than an optional field in the template. Otherwise, I think it we can document it in the wiki rather than the template and keep the configuration simple.

jihaozh: As discussed offline, we will put it into the template for now.

PR #4426, opened 2019-07-11
Add URIUtils class to handle URI/URL encoding/decoding

Jackie-Jiang:
Fix the double encoding bug of encoding URI without scheme
"/segments/segment %" ->
Without fix: "file:/segments/segment+%2525"
With fix: "file:/segments/segment+%25"

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/URIUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/URIUtils.java
@@ -1,0 +1,88 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.common.utils;
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.util.StringJoiner;
+
+
+public class URIUtils {
+  private URIUtils() {
+  }
+
+  /**
+   * Returns the URI for the given base path and optional parts, appends the local (file) scheme to the URI if no
+   * scheme exists. All the parts will be appended to the base path with the file separator.
+   */
+  public static URI getUri(String basePath, String... parts) {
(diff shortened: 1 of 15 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -185,13 +181,8 @@ public Response downloadSegment(
...
     }
-    try {
-      segmentName = URLDecoder.decode(segmentName, "UTF-8");
-    } catch (UnsupportedEncodingException e) {
-      String errStr = "Could not decode segment name '" + segmentName + "'";
-      throw new ControllerApplicationException(LOGGER, errStr, Response.Status.BAD_REQUEST);
-    }
-    final File dataFile = new File(provider.getBaseDataDir(), StringUtil.join("/", tableName, segmentName));
+    segmentName = URIUtils.decode(segmentName);
+    File dataFile = new File(new File(provider.getBaseDataDir(), tableName), segmentName);
```
jackjlli: Why not use `toFile()` here?

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/PinotURIUploadIntegrationTest.java
```diff
@@ -267,12 +267,12 @@ public Integer call()
     }
   }
 
-  private List<String> getAllSegments(String tablename)
+  private List<String> getAllSegments(String tableName)
```
jackjlli: Good catch

pinot-common/src/main/java/org/apache/pinot/common/utils/URIUtils.java
```diff
@@ -0,0 +1,90 @@
...
+    String[] encodedSegments = new String[length + 1];
+    encodedSegments[0] = baseUrl;
+    for (int i = 0; i < length; i++) {
+      encodedSegments[i + 1] = encode(segments[i]);
+    }
+    return String.join(File.separator, encodedSegments);
+  }
+
+  public static String constructDownloadUrl(String baseUrl, String rawTableName, String segmentName) {
+    return getUrl(baseUrl, "segments", rawTableName, segmentName);
```
jackjlli: It may be good to use a static constant here for `segments`?

pinot-common/src/main/java/org/apache/pinot/common/utils/URIUtils.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class URIUtils {
+  private URIUtils() {
+  }
+
+  /**
+   * Returns the URI for the given base path and optional parts, appends the local (file) scheme to the URI if no
+   * scheme exists. All the parts will be appended to the base path with the file separator.
+   */
+  public static URI getUri(String basePath, String... parts) {
```
jackjlli: The second param is not a list of segments. It's just the path. Maybe we should use a better name?

jackjlli: Oh you changed it before I submitted the comments. :)

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
Jackie-Jiang: Which `toFile()`?

PR #4428, opened 2019-07-11
[TE] detection health - coverage fix

jihaozh:
- Fix the issue that anomaly coverage can be more than 100%
- Update the logic to get the latest success task to finish time in backend

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/components/detection-health/template.hbs
+++ b/thirdeye/thirdeye-frontend/app/pods/components/detection-health/template.hbs
@@ -26,7 +26,7 @@
       <div class="te-horizontal-metrics__container">
         <ul class="te-horizontal-metrics__metric">
           <li class="te-horizontal-metrics__number">{{executionTime}}</li>
-          <li class="te-horizontal-metrics__title">Last Detection Task Finish Time</li>
+          <li class="te-horizontal-metrics__title">Last detection task finish time</li>
         </ul>
       </div>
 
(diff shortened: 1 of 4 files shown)


thirdeye/thirdeye-frontend/app/pods/components/detection-health/template.hbs
```diff
@@ -26,7 +26,7 @@
       <div class="te-horizontal-metrics__container">
         <ul class="te-horizontal-metrics__metric">
           <li class="te-horizontal-metrics__number">{{executionTime}}</li>
-          <li class="te-horizontal-metrics__title">Last Detection Task Finish Time</li>
+          <li class="te-horizontal-metrics__title">Last detection task finish time</li>
```
xiaohui-sun: Last success detection task finish time

jihaozh: nice catch. updated

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/health/DetectionTaskStatus.java
```diff
@@ -25,6 +25,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import javafx.concurrent.Task;
```
ShubhamMMT1: I don't think that this package is available for higher jdk versions! This might create an issue!

ShubhamMMT1: I don't think that this package is available for higher jdk versions! This might create an issue!

jihaozh: Thanks for the comment. This import got sneaked in by the IDE. Will remove in following PR.

PR #4430, opened 2019-07-12
Cleanup un-necessary setups in HelixSetupUtils

Jackie-Jiang:
- Remove redundant empty brokerResource resource config
- Remove redundant empty leadControllerResource resource config
- No need to initialize property store sub-directories

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
@@ -19,6 +19,7 @@
 package org.apache.pinot.controller;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.primitives.Longs;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.yammer.metrics.core.MetricsRegistry;
@@ -31,6 +32,7 @@ import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import org.apache.commons.configuration.Configuration;
 import org.apache.commons.httpclient.HttpConnectionManager;
 import org.apache.commons.httpclient.MultiThreadedHttpConnectionManager;
@@ -219,7 +221,7 @@ public class ControllerStarter {
   private void setUpHelixController() {
     // Register and connect instance as Helix controller.
     LOGGER.info("Starting Helix controller");
-    _helixControllerManager = HelixSetupUtils.setup(_helixClusterName, _helixZkURL, _instanceId);
+    _helixControllerManager = HelixSetupUtils.setupHelixController(_helixClusterName, _helixZkURL, _instanceId);
 
     // Emit helix controller metrics
     _controllerMetrics.addCallbackGauge(CommonConstants.Helix.INSTANCE_CONNECTED_METRIC_NAME,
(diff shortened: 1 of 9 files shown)


pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
```diff
@@ -32,183 +32,185 @@
...
+    secondPinotOnlyController.start();
+    TestUtils.waitForCondition(
+        aVoid -> secondPinotOnlyController.getHelixResourceManager().getHelixZkManager().isConnected(), TIMEOUT_IN_MS,
+        "Failed to start the second Pinot-only controller");
+    TestUtils.waitForCondition(aVoid -> {
+      ExternalView leadControllerResourceExternalView =
+          helixAdmin.getResourceExternalView(getHelixClusterName(), LEAD_CONTROLLER_RESOURCE_NAME);
+      for (String partition : leadControllerResourceExternalView.getPartitionSet()) {
+        Map<String, String> stateMap = leadControllerResourceExternalView.getStateMap(partition);
+        if (stateMap.size() != 2 || stateMap.values().containsAll(Arrays.asList("MASTER", "SLAVE"))) {
```
jackjlli: Since the replica is set to 1, which means there'll be only 1 participant in each of the partitions, the participant would ultimately be either MASTER or OFFLINE. So there is no one in SLAVE state in the end.

pinot-core/src/test/java/org/apache/pinot/util/TestUtils.java
```diff
@@ -140,35 +140,32 @@ public static void ensureDirectoriesExistAndEmpty(@Nonnull File... dirs)
    * @param timeoutMs Timeout in milliseconds
    * @param errorMessage Error message if condition is not met before timed out
    */
-  public static void waitForCondition(@Nonnull Function<Void, Boolean> condition, long checkIntervalMs, long timeoutMs,
-      @Nullable String errorMessage)
-      throws Exception {
+  public static void waitForCondition(Function<Void, Boolean> condition, long checkIntervalMs, long timeoutMs,
```
jackjlli: I think it's unnecessary to refactor this method. It's better to put while loop inside the try block. Plus, in your current way you'll need two places to handle `errorMessage != null` cases.

Jackie-Jiang: You should not put while loop inside the try-catch because the outer catch will catch the exception thrown out from the inner loop. Merged the error message check to one place (in the old code we also have two places to handle `errorMessage != null`).

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
Jackie-Jiang: Good point! Re-factored both dual-mode and pinot-only controller test to test this

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerStarter.java
```diff
@@ -219,7 +221,7 @@ public void start() {
   private void setUpHelixController() {
     // Register and connect instance as Helix controller.
     LOGGER.info("Starting Helix controller");
-    _helixControllerManager = HelixSetupUtils.setup(_helixClusterName, _helixZkURL, _instanceId);
+    _helixControllerManager = HelixSetupUtils.setupHelixController(_helixClusterName, _helixZkURL, _instanceId);
```
jackjlli: It's not just setting up the controller, it also creates the cluster if needed as well.

Jackie-Jiang: True, while setting up Pinot controller also create the cluster, so keep them the same

jackjlli: Nope, setting up Pinot controller won't create the cluster; it'll throw exception if cluster is not found. The cluster has to be created by Helix controller.

PR #4431, opened 2019-07-12
Misc fix for controller tests

Jackie-Jiang:


pinot-controller/src/main/java/org/apache/pinot/controller/validation/StorageQuotaChecker.java
```diff
@@ -105,14 +105,16 @@ public QuotaCheckerResponse isSegmentStorageWithinQuota(@Nonnull File segmentFil
...
+      LOGGER.info("Storage quota is not configured for table: {}, skipping the check", tableNameWithType);
+      return success("Storage quota is not configured for table: " + tableNameWithType);
     }
 
     long allowedStorageBytes = numReplicas * quotaConfig.storageSizeBytes();
-    if (allowedStorageBytes < 0) {
-      LOGGER.warn("Storage quota is not configured for table: {}", tableNameWithType);
-      return success("Storage quota is not configured for table: " + tableNameWithType);
+    if (allowedStorageBytes <= 0) {
+      LOGGER.warn("Invalid storage quota: {} for table: {}, skipping the check", quotaConfig.getStorage(),
```
jackjlli: Can we reuse the same String for logging and throwing exception? Same question on Line 108.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/BaseClusterIntegrationTest.java
```diff
@@ -71,7 +71,6 @@
 
   protected final File _tempDir = new File(FileUtils.getTempDirectory(), getClass().getSimpleName());
   protected final File _avroDir = new File(_tempDir, "avroDir");
-  protected final File _preprocessingDir = new File(_tempDir, "preprocessingDir");
```
jackjlli: Hmm.. I remember we had a Hadoop integration test that tests the SegmentPreprocessing Job which uses this param `_preprocessingDir ` but couldn't find it in the code though.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1087,9 +1086,6 @@ public void addTable(@Nonnull TableConfig tableConfig)
         // lets add table configs
         ZKMetadataProvider.setOfflineTableConfig(_propertyStore, tableNameWithType, tableConfig.toZNRecord());
 
-        _propertyStore.create(ZKMetadataProvider.constructPropertyStorePathForResource(tableNameWithType),
```
jackjlli: Why don't we need to write to property store when adding a table?

mcvsubbu: yeah, and all the tests passed too!

Jackie-Jiang: We don't need to initialize this directory because it will be automatically created when the first segment gets pushed. Real-time table does not initialize this directory, so keep them consistent

pinot-controller/src/main/java/org/apache/pinot/controller/validation/StorageQuotaChecker.java
Jackie-Jiang: Done

Issue #4401, opened 2019-07-05
Avoid creating garbage in variable length bytes store for offline dictionary


buchireddy:
This is coming from https://github.com/apache/incubator-pinot/pull/4321#discussion_r299724111

From @mayankshriv 
> Well, another options is to start with a size, and resize as needed. Chances are that we will resize to max quickly, i.e. without creating a lot of garbage. Of course, in the pathalogical case we will resize once per value, but that is no worse than allocating memory each time. I have found reducing garbage at lowest level reader/writers to help with high throughput-low latency use cases.

The scope of this issue is to address the garbage creation issue by reusing some kind of buffer.

buchireddy:
Merged the PR to fix this.

PR #4432, opened 2019-07-12
#4401 Reuse a ThreadLocal byte[] when reading String elements from the variable length value reader.

buchireddy:
Fix for https://github.com/apache/incubator-pinot/issues/4401

@mayankshriv @kishoreg  Please review this when you get time.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/StringUtil.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/StringUtil.java
@@ -72,9 +72,9 @@ public class StringUtil {
     return decodeUtf8(bytes, 0, bytes.length);
   }
 
-  public static String decodeUtf8(byte[] bytes, int startIndex, int endIndex) {
+  public static String decodeUtf8(byte[] bytes, int startIndex, int length) {
     try {
-      return new String(bytes, startIndex, endIndex, charSet);
+      return new String(bytes, startIndex, length, charSet);
     } catch (UnsupportedEncodingException e) {
       throw new RuntimeException(e);
     }
--- a/pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
@@ -93,6 +93,12 @@ public class VarLengthBytesValueReaderWriter implements Closeable, ValueReader {
   private final int _numElements;
 
   /**
+   * ThreadLocal buffer to reuse the byte[] wherever we can. The byte[] is created with a decent
+   * size and later resized dynamically based on the size of elements.
+   */
+  private final ThreadLocal<byte[]> _reusableBytes = ThreadLocal.withInitial(() -> new byte[16]);
+
+  /**
    * Constructor to create a VarLengthBytesValueReaderWriter from a previously written buffer.
    */
   public VarLengthBytesValueReaderWriter(PinotDataBuffer dataBuffer) {
@@ -204,7 +210,13 @@ public class VarLengthBytesValueReaderWriter implements Closeable, ValueReader {
 
   @Override
   public String getUnpaddedString(int index, int numBytesPerValue, byte paddingByte, byte[] buffer) {
-    return StringUtil.decodeUtf8(getBytes(index, numBytesPerValue, buffer));
+    // Read the offset of the byte array first and then read the actual byte array.
+    int offset = _dataBuffer.getInt(_dataSectionStartOffSet + Integer.BYTES * index);
+
+    // To get the length of the byte array, we use the next byte array offset.
+    int length = _dataBuffer.getInt(_dataSectionStartOffSet + Integer.BYTES * (index + 1)) - offset;
+
+    return StringUtil.decodeUtf8(getBytes(buffer, offset, length, true), 0, length);
   }
 
   @Override
@@ -220,13 +232,25 @@ public class VarLengthBytesValueReaderWriter implements Closeable, ValueReader {
     // To get the length of the byte array, we use the next byte array offset.
     int length = _dataBuffer.getInt(_dataSectionStartOffSet + Integer.BYTES * (index + 1)) - offset;
 
+    // Since the byte[] is returned to the caller and caller could hold a reference to it,
+    // the buffer can't really be reused here. Hence, don't use ThreadLocal buffer in this case.
+    return getBytes(buffer, offset, length, false);
+  }
+
+  private byte[] getBytes(byte[] buffer, int offset, int length, boolean useThreadLocalBuffer) {
     byte[] b;
-    // If the caller didn't pass a buffer, create one with exact length.
-    if (buffer == null) {
-      b = new byte[length];
+    // If the caller has passed a buffer that can be used, use it.
+    if (buffer != null && buffer.length == length) {
+      b = buffer;
+    } else if (useThreadLocalBuffer) {
+      // Check if the current instance of ThreadLocal buffer is big enough. If not, resize it to double the size.
+      while (_reusableBytes.get().length < length) {
+        _reusableBytes.set(new byte[_reusableBytes.get().length * 2]);
+      }
+      b = _reusableBytes.get();
     } else {
-      // If the buffer passed by the caller isn't big enough, create a new one.
-      b = buffer.length == length ? buffer : new byte[length];
+      // Can't use caller's buffer or ThreadLocal buffer so create a new one.
+      b = new byte[length];
     }
     _dataBuffer.copyTo(offset, b, 0, length);
     return b;
--- a/pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
@@ -152,6 +152,22 @@ public class VarLengthBytesValueReaderWriter implements Closeable, ValueReader {
     return false;
   }
 
+  /**
+   * Returns the power of 2 >= n.
+   */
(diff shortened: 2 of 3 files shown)


pinot-common/src/main/java/org/apache/pinot/common/utils/StringUtil.java
```diff
@@ -72,9 +72,9 @@ public static String decodeUtf8(byte[] bytes) {
     return decodeUtf8(bytes, 0, bytes.length);
   }
 
-  public static String decodeUtf8(byte[] bytes, int startIndex, int endIndex) {
+  public static String decodeUtf8(byte[] bytes, int startIndex, int length) {
```
Jackie-Jiang: Good catch!

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -220,13 +232,25 @@ public String getPaddedString(int index, int numBytesPerValue, byte[] buffer) {
...
-    // If the caller didn't pass a buffer, create one with exact length.
-    if (buffer == null) {
-      b = new byte[length];
+    // If the caller has passed a buffer that can be used, use it.
+    if (buffer != null && buffer.length == length) {
+      b = buffer;
+    } else if (useThreadLocalBuffer) {
+      // Check if the current instance of ThreadLocal buffer is big enough. If not, resize it to double the size.
+      while (_reusableBytes.get().length < length) {
+        _reusableBytes.set(new byte[_reusableBytes.get().length * 2]);
```
Jackie-Jiang: Directly get the final size (next power of 2 for length) instead of doing this while loop

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -220,13 +232,25 @@ public String getPaddedString(int index, int numBytesPerValue, byte[] buffer) {
...
+    return getBytes(buffer, offset, length, false);
+  }
+
+  private byte[] getBytes(byte[] buffer, int offset, int length, boolean useThreadLocalBuffer) {
     byte[] b;
-    // If the caller didn't pass a buffer, create one with exact length.
-    if (buffer == null) {
-      b = new byte[length];
+    // If the caller has passed a buffer that can be used, use it.
+    if (buffer != null && buffer.length == length) {
```
Jackie-Jiang: Move this logic into the getUnpaddedString() to save the overhead of the if condition.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Done.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: I tried to reuse the code across two methods, which forces us to have the `if` condition but I've now made the code simple with some duplication. Unit tests should help us to catch any issues here too.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -146,6 +152,22 @@ public static boolean isVarLengthBytesDictBuffer(PinotDataBuffer buffer) {
     return false;
   }
 
+  /**
+   * Returns the power of 2 >= n.
+   */
+  private static int nextPowerOf2(int n) {
```
Jackie-Jiang: Faster way:
```
n -= 1;
n |= n >>> 1;
n |= n >>> 2;
n |= n >>> 4;
n |= n >>> 8;
n |= n >>> 16;
return n + 1;
```

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -204,7 +226,25 @@ public double getDouble(int index) {
...
+
+    // To get the length of the byte array, we use the next byte array offset.
+    int length = _dataBuffer.getInt(_dataSectionStartOffSet + Integer.BYTES * (index + 1)) - offset;
+
+    byte[] b;
+    if (buffer != null && buffer.length >= length) {
+      b = buffer;
+    } else {
+      // Check if the current instance of ThreadLocal buffer is big enough. If not, resize it to double the size.
+      if (_reusableBytes.get().length < length) {
```
Jackie-Jiang: To save a `_reusableBytes.get()`:
```
b = _reusableBytes.get();
if (b.length < length) {
  b = new byte[nextPowerOf2(length)];
  _reusableBytes.set(b);
}
```

buchireddy:
> Can you please post some perf number of this improvement?

I've done some quick tests with the same `StringDictionaryPerfTest` that I've used to compare the variable length dictionary with fixed length dictionary earlier and compared the cost of `getValues()` both with and without my changes. I see some improvement in the latency for cardinality <1M. Please note that the red line and blue line are with my changes and they both lie below the yellow and green ones for cardinality <1M. I'm wondering why it's going above the other lines for >1M cardinality though. Thinking further on that.

ReuseThreadLocalBuffer

Jackie-Jiang:
@buchireddy You might want to attach a profiler and compare the GC as well.

buchireddy:
@Jackie-Jiang I've done some testing using `StringDictionaryPerfTest.java` and captured the GC logs and parsed them using an online tool. Here are the stats and full reports.
Before change: https://gceasy.io/my-gc-report.jsp?p=c2hhcmVkLzIwMTkvMDcvMTMvLS1nY193aXRob3V0X2NoYW5nZS5sb2ctLTItMTYtNTE=&channel=WEB
With change: https://gceasy.io/my-gc-report.jsp?p=c2hhcmVkLzIwMTkvMDcvMTMvLS1nY193aXRoX2NoYW5nZS5sb2ctLTItMTgtMw==&channel=WEB


  | With Change | Before Change
-- | -- | --
Total GC count | 33 | 47
Total reclaimed bytes | 44.28 gb | 66.43 gb
Total GC time | 2 sec 405 ms | 3 sec 250 ms
Avg GC time | 72.9 ms | 69.1 ms
GC avg time std dev | 175 ms | 164 ms
GC min/max time | 0.848 ms / 690 ms | 0.651 ms / 833 ms
GC Interval avg time | 1 sec 154 ms | 911 ms
Minor GC count | 31 | 44
Minor GC reclaimed | 43.95 gb | 66.14 gb
Minor GC total time | 1 sec 96 ms | 1 sec 720 ms
Minor GC avg time | 35.4 ms | 39.1 ms
Minor GC avg time std dev | 96.6 ms | 97.5 ms
Minor GC min/max time | 0.848 ms / 535 ms | 0.651 ms / 553 ms
Minor GC Interval avg | 1 sec 231 ms | 975 ms
Full GC Count | 2 | 3
Full GC reclaimed | 339.23 mb | 303.5 mb
Full GC total time | 1 sec 309 ms | 1 sec 530 ms
Full GC avg time | 654 ms | 510 ms
Full GC avg time std dev | 35.1 ms | 269 ms
Full GC min/max time | 619 ms / 690 ms | 174 ms / 833 ms
Full GC Interval avg | 2 sec 392 ms | 6 sec 601 ms


As the results show, there is clearly improvement in both the minor and major GCs, GC time and the memory used. Please take a look.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Done.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
buchireddy: Done.

pinot-core/src/main/java/org/apache/pinot/core/io/util/VarLengthBytesValueReaderWriter.java
```diff
@@ -204,7 +223,26 @@ public double getDouble(int index) {
...
+    int length = _dataBuffer.getInt(_dataSectionStartOffSet + Integer.BYTES * (index + 1)) - offset;
+
+    byte[] b;
+    if (buffer != null && buffer.length >= length) {
+      b = buffer;
+    } else {
+      // Check if the current instance of ThreadLocal buffer is big enough. If not, resize it to double the size.
+      b = _reusableBytes.get();
+      if (b.length < length) {
+        b = new byte[nextPowerOf2(length)];
```
mayankshriv: Do we allow arbitary length strings? IIRC @Jackie-Jiang had put a max string length limit? If so, we could cap this allocation to that limit as well.

mcvsubbu: max string is set in the schema and changeable.

buchireddy: @mcvsubbu @mayankshriv Actually, good point. Jackie has merged the PR but this can be addressed in another PR too.

@mcvsubbu pointer to the config you're talking about please?

Jackie-Jiang: IMO the current code is okay. Usually we put some power of 2 as the max length. I don't think it's worth nesting the reader with schema to save small memory.

PR #4434, opened 2019-07-13
Remove the massive console print and log from CalciteSqlCompilerTest

Jackie-Jiang:


pinot-common/src/test/java/org/apache/pinot/sql/parsers/CalciteSqlCompilerTest.java
```diff
@@ -339,60 +336,23 @@ public void testStringLiteral() {
     Assert.assertEquals(groupbyList.get(1).getFunctionCall().getOperands().get(1).getIdentifier().getName(), "FOO");
   }
 
-  @Test
-  public void testConverter()
```
mayankshriv: While it makes sense to remove the logging, here's a prior conversation on the intention for this test.

https://github.com/apache/incubator-pinot/pull/4387#discussion_r300516547

pinot-common/src/test/java/org/apache/pinot/sql/parsers/CalciteSqlCompilerTest.java
```diff
@@ -339,60 +336,23 @@ public void testStringLiteral() {
...
-      throws IOException {
-    CalciteSqlParser.compileToPinotQuery("SELECT MIN(div(DaysSinceEpoch,2)) FROM mytable");
```
kishoreg: should we add these queries to the sql_queries.list? Its possible that these were added to test edge cases.

mayankshriv: As per the discussion here #4387, this was added as a sample for dev's to be able to debug. I had made the same comment to @Jackie-Jiang above.

PR #4435, opened 2019-07-14
Onboarding best practices doc

npawar:


s-sahay:
@npawar : Nice work on this! 
Minor suggestion on changing the section header and introduction to something like this:
**Pinot Best Practices**
Here are the recommended best practices for designing tables, loading data and writing queries (<anything else>) for Pinot.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+Considerations common to offline and realtime
+**********************************************
+
+1. Pre-aggregations
+###################
+Pre aggregate the data as much as the application logic allows. This means **rolling up the metric values for the unique dimensions and time column combinations**. This is beneficial as we will reduce the size of the data being stored in Pinot, as well as avoid aggregations to be done in Pinot for every query, hence improving query performance. 
+
+- For offline, perform the aggregations in your data preparation hadoop/spark job. 
+- For realtime, a samza job can be used to aggregate data at intervals based on the freshness requirements of the usecase.
```
mcvsubbu: Also mention that realtime supports aggregation on the fly for sum(). I think we mention it elsewhere, so you can point to it

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+2. Time column
+###################
+Think about what **granularity of time column** is needed for your application. This will usually be *HOURS, DAYS, (sometimes 15 MINUTES)*. It is not recommended to have the time column in MILLISECONDS or SECONDS granularity. If time granularity is in milliseconds, the cardinality of the time column becomes very high in realtime systems, causing the time column dictionary to get very big. Bucket your time column to the coarsest granularity possible (hoursSinceEpoch, daysSinceEpoch). A greater level of pre-aggregations can be achieved with coarser time granularity.
```
mcvsubbu: Instead of "it is not recommended", I may reword it as "If you don't need granularity in seconds or milliseconds, then it is not recommended"

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+If you are unsure of the granularity, or feel that the chosen granularity might have to be made finer at a later point, you can keep the column in *MILLISECONDS*, however, round off the column to a higher granularity (round off millis to nearest day for example). This way you still have flexibility later on to change the bucketing.
+
```
mcvsubbu: adding a couple of examples here illustrating how the granularity and time unit plays will be useful here.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+3. Segment Push Modes
+######################
+Pinot has 2 modes of segment push. *APPEND and REFRESH*.
+
+- APPEND - every data push is an incremental one, and the data received is appended to the existing data.  
+- REFRESH - The existing data is completely replaced by the new data pushed. The new segment names need to be exactly the same for the existing segments to be replaced
+
+In general, if you push a segment with the same name, it will overwrite the previous segment (in either mode).
```
mcvsubbu: Mention here that the NormalizedDateSegmentNameGenerator class provides a way to name the segments correctly depending on the type of table, and should be used as an example if customers are providing their own segment name generators.
Also note that APPEND tables will also allow you to replace segments.

Do you want to note that minion purge task does not work for REFRESH tables?

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+4. Prefer long over string where possible
+##########################################
+Strings in Java are inefficient. We end up serializing/deserializing them during query execution a lot of time.
```
mcvsubbu: "several times?"

mcvsubbu: I may omit the first sentence about inefficiency.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+Instead of storing a long string value, store an id, and use a stateless utility function for conversion. Similarly, for attributes of a dimension (*for example, companyId -> company details*) store only the id in Pinot, and use another key value store/pinot table for the lookup. This is applicable only if Pinot cannot handle desired latency/throughput, even after all other possible optimizations. The reason is that, inside Pinot, we don’t know the relation between different columns. If used in a query like group by (group by companyId, companyName, companySize)  we will try to gather all the combinations of these columns in order to get the group-by results. This is quite expensive inside Pinot, but with key-value store service (or even keep a separate Pinot table), this will be simply a value lookup, which is way cheaper, and can definitely give you better performance
```
mcvsubbu: "Instead of storing a long string, consider storing an id and using a stateless utility function for conversion if possible.

I think this paragraph needs re-work. I am not sure what is being said here.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+5. Sorted column
+###################
+Sort data on column if it will always be present in the queries. **Sorted index performs superior than inverted index.**
```
mcvsubbu: Sort the data on the column that will occur in WHERE clause on most of your queries.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+5. Sorted column
+###################
+Sort data on column if it will always be present in the queries. **Sorted index performs superior than inverted index.** 
+
+- For offline, Pinot cannot sort the data in segments that are pushed to pinot. This needs to be done prior to the segment creation, in the data preparation job. If the data is sorted before segment creation, the segment creation will detect and maintain this sorted column.
```
mcvsubbu: Suggest reword: "Offline segments pushed to pinot are immutable, so Pinot does not re-sort the rows before serving queries. Therefore, the sort order of rows needs to be decided before segment creation (include the segment generator config name here). Also mention that the "sorted" column setting in table config is ignored for Offline tables.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+5. Sorted column
+###################
+Sort data on column if it will always be present in the queries. **Sorted index performs superior than inverted index.** 
+
+- For offline, Pinot cannot sort the data in segments that are pushed to pinot. This needs to be done prior to the segment creation, in the data preparation job. If the data is sorted before segment creation, the segment creation will detect and maintain this sorted column. 
+- For realtime, set the column in “sortedColumn” in table config, and it will be sorted in Pinot when data is converted to a Pinot segment.
+
+6. Multi value columns
+#######################
+Pinot supports multi value columns. Arrays in an avro will be treated as multi value columns, by setting `“singleValueField” : “false”` in the dimensionFieldSpec in Pinot schema. However, in general they are more costly for query execution, and ensure that they are being used only when needed.
```
mcvsubbu: avro ? If the input file is avro? The paragraph is not clear, seems to indicate that pinot modifies the schema if input avro has an array field?

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+7. Inverted indexes
+###################
+For columns that will be used fairly frequently in your queries, enable inverted indexes. Inverted indexes are **more efficient and effective for high cardinality columns.**
```
mcvsubbu: mention pros and cons here. (and everywhere).

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+#######################
+Pinot supports multi value columns. Arrays in an avro will be treated as multi value columns, by setting `“singleValueField” : “false”` in the dimensionFieldSpec in Pinot schema. However, in general they are more costly for query execution, and ensure that they are being used only when needed.
+
+7. Inverted indexes
+###################
+For columns that will be used fairly frequently in your queries, enable inverted indexes. Inverted indexes are **more efficient and effective for high cardinality columns.** 
+
+8. Partitioning
+###################
+- For offline, if the data has been partitioned, Pinot can leverage this information for querying. This can be done by setting the partitioning function, number of partitions and column name in the segmentPartitionConfig in the table config
```
mcvsubbu: refer to the section where we describe partitioning pinot data (I hope there is one. Otherwise, start an empty section, we will fill it up)

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+
+
+
+Realtime specific considerations
+*********************************
+
+1. Time column
+###################
+Time column is a must for hybrid table
```
mcvsubbu: time column must be defined in both offline and realtime table configs to be the same

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+2. Num partitions
```
mcvsubbu: Number of stream partitions

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+###################
+When creating the stream topic (typically for your output of Samza job),  keep the number of partitions to a minimum. If partitions are increased later, Pinot is able to automatically and seamlessly start consuming from them.
```
mcvsubbu: When creating the stream topic that you expect pinot to consume, ...

mcvsubbu: remove seemlessly

docs/onboarding_best_practices.rst
kishoreg: +1. We should just say use the pick the right granularity needed for your application and explain the trade-off. I would just remove ```It is not recommended to have the time column in MILLISECONDS or SECONDS granularity.```

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+3. Retention
+###################
+The retention of the stream topic should be set depending on the lag in your offline push. In hybrid tables, depending on the push frequency (set in table config), Pinot will start using the offline data
```
mcvsubbu: clarify "lag". Perhaps by an example.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+The retention of the stream topic should be set depending on the lag in your offline push. In hybrid tables, depending on the push frequency (set in table config), Pinot will start using the offline data 
+
+- 1 day after offline data becomes available, if push frequency daily 
+- 1 hour after offline data becomes available, if push frequency is hourly
+
+Generally a stream retention of 4-5 days is sufficient, as this is enough time for the offline flow to generate data. The retention of the realtime table should no be more than the stream retention.
+
+4. Pre aggregations
+###################
+Consider reducing the number of stream events per second, by aggregating in a samza job. Keep the aggregation window reasonable (5 minutes, 15 minutes, 1 hour depending on the events per second). If the aggregation window is large (say 6 hours), a burst of events will be sent to Pinot periodically after long gaps of silence. This kind of ingestion pattern is not ideal for Pinot consumption, and a steady stream of events is preferred.
```
mcvsubbu: Can we move this to the pre-aggregation section under realtime+offline?

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+Generally a stream retention of 4-5 days is sufficient, as this is enough time for the offline flow to generate data. The retention of the realtime table should no be more than the stream retention.
+
+4. Pre aggregations
+###################
+Consider reducing the number of stream events per second, by aggregating in a samza job. Keep the aggregation window reasonable (5 minutes, 15 minutes, 1 hour depending on the events per second). If the aggregation window is large (say 6 hours), a burst of events will be sent to Pinot periodically after long gaps of silence. This kind of ingestion pattern is not ideal for Pinot consumption, and a steady stream of events is preferred. 
+
+5. Metrics aggregation in Pinot realtime
+#########################################################
+If aggregation is not possible at source (due to freshness concerns), pinot can aggregate after events are received, by enabling aggregateMetrics: true in table config. Note that Pinot supports only “sum” aggregations in realtime.
```
mcvsubbu: suggest merging with pre-aggregation as mentioned before

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+6. Aggregation buckets
+#####################################
+Keep the aggregation bucket size the same across realtime and offline. If aggregation buckets are not consistent across realtime and offline, there can be undesirable results for data which spans the boundary. For example, if in realtime you have MILLISECONDS time column rounded off to the nearest hour via the samza job, follow the same model in offline flow, and round off time column to the nearest hour in the pre aggregations. This means, the samza job will produce a new timestamp every hour, whereas the offline segment for the day will have 24 unique timestamps.
```
mcvsubbu: aggregation bucket size? Do we refer to it as granularity in the previous section? Better to use the same term everywhere. Also, consider adding one section on aggregation and mention everything there instead of having a numbered list.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+Keep the aggregation bucket size the same across realtime and offline. If aggregation buckets are not consistent across realtime and offline, there can be undesirable results for data which spans the boundary. For example, if in realtime you have MILLISECONDS time column rounded off to the nearest hour via the samza job, follow the same model in offline flow, and round off time column to the nearest hour in the pre aggregations. This means, the samza job will produce a new timestamp every hour, whereas the offline segment for the day will have 24 unique timestamps. 
+
+
+
+
+Querying Pinot
+^^^^^^^^^^^^^^
+`PQL doc <https://pinot.readthedocs.io/en/latest/pql_examples.html>`_
+
+1. Apply **time filters** to your queries where applicable, to avoid fetching too much data
```
mcvsubbu: Also to avoid fetching data that is awaiting retention. May want to explain how retention works, or point to a section

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+2. Apply **reasonable LIMIT and TOP** to your query results, based on what application needs
```
mcvsubbu: Not sure how an application writer can use this.  Would it be ok to word it like "Fetching a lot of data from Pinot is sub-optimal, so take care when specifying values for TOP and LIMIT clauses"

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+3. Pinot supports **distinct count**. However it is an expensive operation. Consider using **distinctCountHLL**, which is an approximate distinct count
+
+4. Implement **caching layer on the application side**, to reduce the amount of repeated queries being sent to Pinot
```
mcvsubbu: add a caveat here about keeping cache consistent. Sending periodic burst queries to keep cache consistent can also be bad.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+5. For **report generation** style applications, it might be better to **split your query** into smaller time periods
```
mcvsubbu: Or, use the offline hadoop mechanisms to generate reports.

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+6. At Pinot table side, **hard QPS quota limits** can be applied `tableConfig -> quotaConfig -> maxQueriesPerSecond`. If this quota is exceed, a 429 exception will be thrown with  message 
+
+.. code-block:: none
+
+    "Request <request id> exceeds query quota for table:<table name>, query:<query>"
+
+7. Pinot doesn’t have query throttling. It is recommended to have an **application side query throttling mechanism**
```
mcvsubbu: Combine this with the previous one?
Also, suggested wording: "Pinot does not throttle queries. Applications may choose to throttle queries to minimize 429 returns if necessary, or have administrators provision more hardware to support the rate of queries"

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+8. Even if it is a low QPS usecase (ie < 1 qps) having distinct count queries, top K queries, etc can greatly impact the query cost
```
mcvsubbu: Suggest take this out?

docs/onboarding_best_practices.rst
```diff
@@ -0,0 +1,165 @@
...
+
+9. **Count, max, min queries** with no filters on OFFLINE only or REALTIME only tables, will be answered only from **metadata**, and hence will be very efficient. For example, if you have a hybrid (OFFLINE + REALTIME) table, and you want to query
```
mcvsubbu: s/only/directly/
s/will be/are/g

PR #4438, opened 2019-07-16
Clean up the ServerInstance class and remove the un-necessary ip address stored

Jackie-Jiang:
- Remove the un-necessary IP address field
- Remove the error log for the UnknownHostException
- Always use the cached host name and short host name if exists
- This can reduce lots of errors in tests as host passed in might not exist, e.g. "testHost"

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/ServerInstance.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/ServerInstance.java
@@ -20,55 +20,46 @@ package org.apache.pinot.common.response;
 
 import com.google.common.net.InternetDomainName;
 import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.concurrent.ConcurrentHashMap;
-import org.apache.commons.lang3.tuple.Triple;
 import org.apache.pinot.common.utils.CommonConstants;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.apache.pinot.common.utils.EqualityUtils;
 
 
 /**
  * Service abstraction.
- * A service is identified by its hostname and port.
- * Internally, an ip address is also resolved.
- *
- * Nuances:
- * -------
- * A hostname "localhost" will not be resolved to the
- * local hostname and will retain the hostname as "localhost".
- * If the name passed is "127.0.0.1", then it is resolved to the
- * local hostname.
+ * A service is identified by its host, port and sequence id.
  */
-public class ServerInstance implements Comparable<ServerInstance> {
-  protected static final Logger LOGGER = LoggerFactory.getLogger(ServerInstance.class);
-
+public class ServerInstance {
   public static final String NAME_PORT_DELIMITER = ":";
   public static final String NAME_PORT_DELIMITER_FOR_INSTANCE_NAME = "_";
 
+  private static class HostNamePair {
+    String _hostName;
+    String _shortHostName;
+
+    HostNamePair(String hostName, String shortHostName) {
+      _hostName = hostName;
+      _shortHostName = shortHostName;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


pinot-common/src/main/java/org/apache/pinot/common/response/ServerInstance.java
```diff
@@ -140,8 +110,12 @@ public int getPort() {
     return _port;
   }
 
-  public InetAddress getIpAddress() {
-    return _ipAddress;
+  public ServerInstance withSeq(int seq) {
+    return new ServerInstance(_hostname, _port, seq, _shortHostName);
+  }
+
+  public String getShortHostName() {
```
jackjlli: It'd be good to put all the getter methods together.

pinot-transport/src/test/java/org/apache/pinot/transport/common/ServerInstanceTest.java
```diff
@@ -18,61 +18,27 @@
...
+    assertNotEquals(new ServerInstance("localhost", 8081), new ServerInstance("localhost", 8082));
 
-      ServerInstance instance1 = new ServerInstance("127.0.0.1", 8080);
-      Assert.assertEquals(instance1.getPort(), 8080, "Port check");
-      Assert.assertEquals(instance1.getPort(), 8080, "Hostname check");
-      Assert.assertEquals(instance1.getHostname(), ipAddr.getHostName(), "Host check");
-      Assert.assertEquals(instance1.getIpAddress(), ipAddr, "IP check");
-    }
+    // Same port but different host
+    assertNotEquals(new ServerInstance("foo", 8080), new ServerInstance("bar", 8080));
```
jackjlli: Can we include a test to test the case when host name is an ip address, like `127.0.0.1`?

pinot-common/src/main/java/org/apache/pinot/common/response/ServerInstance.java
Jackie-Jiang: Done

mcvsubbu:
I think the IP addresses are there for a reason. Perhaps azure/kubernetes clusters? I remember some discussion around it.

mcvsubbu:
Or, maybe it was in the context of caching the IP address while using connection pools. Are we totally disabled on connection pools now?

pinot-transport/src/test/java/org/apache/pinot/transport/common/ServerInstanceTest.java
Jackie-Jiang: Done

Jackie-Jiang:
@mcvsubbu We never fetch the ip address from the server instance. The reason why we have the ip address is to resolve "127.0.0.1" to "localhost" so they are considered the same host. That is not necessary because the server instance is always derived from the instance name

pinot-transport/src/test/java/org/apache/pinot/transport/common/ServerInstanceTest.java
jackjlli: Thanks. How about the test comparing with `127.0.0.1` as hostname and `localhost` as hostname? Would they be the same? Can we include this test as well?

Jackie-Jiang: That won't be the same as we do not resolve the ip address

Jackie-Jiang:
@mcvsubbu Actually the domain name expects the host name instead of the ip address, so I'll add back the ip address to host lookup

pinot-transport/src/test/java/org/apache/pinot/transport/common/ServerInstanceTest.java
Jackie-Jiang: Added back the IP address lookup and added test

PR #4440, opened 2019-07-17
Fix the potential resource leak from NettyServer

Jackie-Jiang:
- Release the ByteBuf immediately after reading the bytes
- De-couple RequestHandler from Netty ChannelHandlerContext and ByteBuf
- Change RequestHandler.processRequest() to directly take byte[] instead of the ByteBuf
- Fix the resource leak issue when RequestHandler.processRequest() throws Exception

### DIFF
--- a/pinot-server/src/main/java/org/apache/pinot/server/request/ScheduledRequestHandler.java
+++ b/pinot-server/src/main/java/org/apache/pinot/server/request/ScheduledRequestHandler.java
@@ -21,13 +21,11 @@ package org.apache.pinot.server.request;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import io.netty.buffer.ByteBuf;
-import io.netty.channel.ChannelHandlerContext;
-import java.net.InetSocketAddress;
 import org.apache.pinot.common.metrics.ServerMeter;
 import org.apache.pinot.common.metrics.ServerMetrics;
 import org.apache.pinot.common.metrics.ServerQueryPhase;
 import org.apache.pinot.common.request.InstanceRequest;
+import org.apache.pinot.common.utils.BytesUtils;
 import org.apache.pinot.core.query.request.ServerQueryRequest;
 import org.apache.pinot.core.query.scheduler.QueryScheduler;
 import org.apache.pinot.serde.SerDe;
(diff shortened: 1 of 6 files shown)


pinot-transport/src/test/java/org/apache/pinot/transport/perf/ScatterGatherPerfServer.java
```diff
@@ -152,17 +150,15 @@ public MyRequestHandler(String response, CountDownLatch responseHandlingLatch, l
...
     @Override
-    public ListenableFuture<byte[]> processRequest(ChannelHandlerContext channelHandlerContext, ByteBuf request) {
-      byte[] b = new byte[request.readableBytes()];
-      request.readBytes(b);
+    public ListenableFuture<byte[]> processRequest(byte[] request) {
       if (null != _responseHandlingLatch) {
         try {
           _responseHandlingLatch.await();
         } catch (InterruptedException e) {
           e.printStackTrace();
```
snleee: Can we put this as `LOGGER.error`?

pinot-server/src/main/java/org/apache/pinot/server/request/ScheduledRequestHandler.java
```diff
@@ -51,20 +49,14 @@ public ScheduledRequestHandler(QueryScheduler queryScheduler, ServerMetrics serv
...
-    request.readBytes(byteArray);
     SerDe serDe = new SerDe(new TCompactProtocol.Factory());
-    final InstanceRequest instanceRequest = new InstanceRequest();
-
-    if (!serDe.deserialize(instanceRequest, byteArray)) {
-      LOGGER.error("Failed to deserialize query request from broker ip: {}",
-          ((InetSocketAddress) channelHandlerContext.channel().remoteAddress()).getAddress().getHostAddress());
+    InstanceRequest instanceRequest = new InstanceRequest();
+    if (!serDe.deserialize(instanceRequest, request)) {
+      LOGGER.error("Failed to deserialize query request: {}", BytesUtils.toHexString(request));
```
snleee: Adding `BytesUtils.toHexString(request)` seems not to add much value on debugging?

pinot-transport/src/test/java/org/apache/pinot/transport/perf/ScatterGatherPerfServer.java
Jackie-Jiang: Good point

pinot-server/src/main/java/org/apache/pinot/server/request/ScheduledRequestHandler.java
Jackie-Jiang: This should not happen. While if it happens, we want to know the bytes of the request in order to debug.

pinot-transport/src/test/java/org/apache/pinot/transport/perf/ScatterGatherPerfServer.java
Jackie-Jiang: This class is just for perf benchmark purpose (not production code)

PR #4445, opened 2019-07-18
Refactor pinot-connectors to break the dependencies from kafka 0.9

fx19880617:
Currently `pinot-tools`/`pinot-integration-tests`/`pinot-perf` modules have compile dependency on `pinot-connector-kafka-0.9`, which prevents us to add new connectors.

This PR extracts common interfaces of StreamDataServer and StreamDataProducer. And create those instance based on runtime class name.

So we could make kafka-connector as a runtime dependency for `pinot-tools`/`pinot-integration-tests`/`pinot-perf`.


pinot-core/src/main/java/org/apache/pinot/core/realtime/stream/StreamDataServerStartable.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.apache.pinot.core.realtime.stream;
+
+import java.util.Properties;
+
+
+/**
+ * StreamDataServerStartable is the interface for stream data sources. E.g. KafkaDataServerStartable, KinesisDataServerStarable.
+ */
+public interface StreamDataServerStartable {
```
kishoreg: simple java docs for this.

fx19880617:
I've tested hybrid and realtime quick start. Both of them work.

pinot-tools/src/main/java/org/apache/pinot/tools/KafkaStarterUtils.java
```diff
@@ -0,0 +1,103 @@
...
+  public static int DEFAULT_KAFKA_PORT = 19092;
+  public static final String DEFAULT_KAFKA_BROKER = "localhost:" + DEFAULT_KAFKA_PORT;
+
+  public static final String PORT = "port";
+  public static final String BROKER_ID = "broker.id";
+  private static final String ZOOKEEPER_CONNECT = "zookeeper.connect";
+  private static final String LOG_DIRS = "log.dirs";
+
+  public static final String KAFKA_09_SERVER_STARTABLE_CLASS_NAME =
+      "org.apache.pinot.core.realtime.impl.kafka.server.KafkaDataServerStartable";
```
npawar: is the expectation now that every kafka implementation should have the KafkaDataServerStartable in the path `org.apache.pinot.core.realtime.impl.kafka.server` if we want the pinot-tools/pinot-perf/pinot-integration-tests to run?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/StreamAvroIntoKafkaCommand.java
```diff
@@ -104,8 +102,12 @@ public boolean execute()
     properties.put("serializer.class", "kafka.serializer.DefaultEncoder");
     properties.put("request.required.acks", "1");
 
-    ProducerConfig producerConfig = new ProducerConfig(properties);
-    Producer<byte[], byte[]> producer = new Producer<byte[], byte[]>(producerConfig);
+    StreamDataProducer streamDataProducer;
+    try {
+      streamDataProducer = StreamDataProvider.getStreamDataProducer(KafkaStarterUtils.KAFKA_09_PRODUCER_CLASS_NAME, properties);
```
npawar: why keep `09` in the name? I thought these same files are going to be able to handle other versions as well?

npawar:
Minor questions, LGTM otherwise. Thanks a lot for doing this!

pinot-tools/src/main/java/org/apache/pinot/tools/KafkaStarterUtils.java
kishoreg: yes, we can probably use profiles to add the right dependency? https://stackoverflow.com/questions/166895/different-dependencies-for-different-build-profiles

fx19880617: I think if we want to have an integration test, then data server/provider has to be implemented. Either we can have a mock one or we directly wrapper a real implementation.

PR #4446, opened 2019-07-18
Add support in the rebalancer for the user to provide minimum number of serving replicas

siddharthteotia:
The current implementation of no-downtime rebalancing keeps up at least 1 serving replica up for each segment. Having only 1 serving replica up can potentially hurt throughput.

- The changes here add support for the user to specify the minimum number of serving replicas that should be kept up for no-downtime rebalancing.
- Rebalancer bails out if external view will never converge (ERROR state)
- Added unit tests in TableRebalancerTest for the new code.
- Added cluster integration test to cover all scenarios for more coverage and end-to-end testing
using entry point in tools package -- RebalanceTable command in pinot-admin.sh script


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -212,39 +214,87 @@ private IdealState getNextState(IdealState currentState, IdealState targetState,
...
     }
 
-    MapDifference difference = Maps.difference(targetMap, srcMap);
-    if (!difference.entriesInCommon().isEmpty()) {
-      // if there are entries in common, there won't be downtime
+    // we are in no-downtime mode
+    int minReplicasToKeepUp = rebalanceUserConfig.getInt(RebalanceUserConfigConstants.MIN_REPLICAS_TO_KEEPUP_FOR_NODOWNTIME,
+        RebalanceUserConfigConstants.DEFAULT_MIN_REPLICAS_TO_KEEPUP_FOR_NODOWNTIME);
+    if (minReplicasToKeepUp >= currentIdealStateSegmentHosts.size()) {
+      // if the minimum number of serving replicas in rebalance config is
```
siddharthteotia: One thought here was that PinotTableRestletResource can first use an API of TableRebalancer to check if the user specified min replica number makes sense or not. TableRebalancer will look at ideal state and see the number of replicas for each segment. If the user specified value >= number of replicas then we throw an exception right there before rest api code submits the async rebalance task. 

Right now such an API is not implemented. Rather we take the user specified value as is and at the time of rebalancing if we found the value to be invalid, we default to 1

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/util/TableRebalancerTest.java
```diff
@@ -27,14 +27,16 @@
...
+import org.apache.pinot.controller.helix.ControllerTest;
 import org.apache.pinot.controller.helix.core.TableRebalancer;
 import org.apache.pinot.controller.helix.core.rebalance.RebalanceUserConfigConstants;
 import org.testng.Assert;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 
-public class TableRebalancerTest {
+public class TableRebalancerTest extends ControllerTest {
```
siddharthteotia: This was by mistake -- need to clean up.

siddharthteotia:
The latest changes have the following:

(1) Entry point in pinot-admin tool for table rebalancer
(2) Add a cluster integration test for the end-to-end path for the usage from pinot-admin tool

siddharthteotia:
cc @Jackie-Jiang , @snleee , @mcvsubbu

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -212,39 +218,132 @@ private IdealState getNextState(IdealState currentState, IdealState targetState,
...
+        // states should be ignored
+        if (!targetIdealStateSegmentHosts.containsKey(host)) {
+          hostToRemove = host;
+          break;
+        }
+      }
+
+      if (!hostToRemove.equals("")) {
+        idealStateToUpdate.getInstanceStateMap(segmentId).remove(hostToRemove);
+        LOGGER.info("Removing host: {} for segment: {}", hostToRemove, segmentId);
```
mcvsubbu: this will be a per segment log. u want this in info?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -212,39 +218,132 @@ private IdealState getNextState(IdealState currentState, IdealState targetState,
...
+      for (String host : targetIdealStateSegmentHosts.keySet()) {
+        // the hosts host that has already been added from target
+        // ideal state to new state should be ignored
+        if (!updatedSegmentInstancesMap.containsKey(host)) {
+          hostToAdd = host;
+          break;
+        }
+      }
+
+      if (!hostToAdd.equals("")) {
```
mcvsubbu: assert that it does not equal "", or, better, add a preconditions check

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -212,39 +218,132 @@ private IdealState getNextState(IdealState currentState, IdealState targetState,
...
+      // if there are enough hosts in common between current and target ideal states
+      // to satisfy the min replicas condition, then there won't be any downtime
+      // and we can directly set the current ideal state to target ideal state
+      LOGGER.debug("Current and target ideal states have common hosts. Will set to target state at one go");
+      setTargetState(idealStateToUpdate, segmentId, targetIdealStateSegmentHosts);
+      ++_rebalancerStats.directTransitions;
     } else {
       // remove one entry
-      idealState.getInstanceStateMap(segmentId).remove(srcMap.keySet().stream().findFirst().get());
+      String hostToRemove = "";
```
mcvsubbu: so the algorithm is to move one replica at a time until we reach a min number of replicas alive in target state, and then just move everything else. Would be good to note that in a comment above the method.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -300,4 +399,55 @@ private void waitForStable(String resourceName)
...
+     * rebalancing if there are sufficient common
+     * hosts between current and target ideal states
+     * @return direct transitions
+     */
+    public int getDirectTransitions() {
+      return directTransitions;
+    }
+
+    /**
+     * Get the number of times we updated the ideal
```
mcvsubbu: this is really number of segments we updated, because we bump this once for each segment that we change in idealstate. We may then update the idealstate in one go for all the segments.  We should also keep the number of idealstate updates as a separate stat.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/util/TableRebalancerTest.java
```diff
@@ -117,4 +119,332 @@ public void downtimeUpdateWithNoCommonElements() {
...
+
+    // STEP 1
+    rebalancer.updateSegmentIfNeeded(segmentId, currentIdealState.getRecord().getMapField(segmentId),
+        targetSegmentInstancesMap, toUpdateIdealState, noDowntime);
+    // After step 1, let's assume that
+    // toUpdateIdealState : { SEGMENT1 : { host4 : online, host2 : online, host3 : online } }
+    // Essentially, host1 from current state got replaced with host4 from target state
+    // Now, what we are checking is that between toUpdate and target ideal state,
+    // there is exactly 1 host in common
+    verifyStateForMinReplicaConstraint(toUpdateIdealState, targetIdealState, 1, true);
```
mcvsubbu: verify stats?

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -0,0 +1,352 @@
...
+  protected boolean isUsingNewConfigFormat() {
+    return true;
+  }
+
+  @BeforeClass
+  public void setup()
+      throws Exception {
+    startZk();
+    startController();
+    startBroker();
```
mcvsubbu: You should not need broker for this test

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -472,7 +472,8 @@ public String rebalance(
       @ApiParam(value = "offline|realtime") @Nonnull @QueryParam("type") String tableType,
       @ApiParam(value = "true|false") @Nonnull @DefaultValue("true") @QueryParam("dryrun") Boolean dryRun,
       @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("includeConsuming") Boolean includeConsuming,
-      @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("downtime") Boolean downtime) {
+      @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("downtime") Boolean downtime,
+      @ApiParam(value = "1|2|3|4") @DefaultValue("1") @QueryParam("minReplicasToKeepUpForNoDowntime") Integer minReplicasToKeepUp) {
```
snleee: Instead of number, can you put the explanation on what this variable means? It looks like we need to choose one of 1-4.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/RebalanceCommand.java
```diff
@@ -0,0 +1,114 @@
...
+            + "Use the -nodowntime option to rebalance without downtime. It takes"
+            + "care of keeping at least 1 serving replica alive while rebalancing. "
+            + "For better QPS, use the -minReplicasForNoDowntime <val> option along with"
+            + "-nodowntime to ensure that rebalancer keeps at least the required number of "
+            + "replicas alive for each segment while segments are being moved between hosts";
+    return description;
+  }
+
+  @Override
+  public void printUsage() {
```
snleee: Why are you trying to override `printUsage`?

I thought the framework would deal with printing information correctly (with the info that you wrote in `@Option` annotation

pinot-tools/src/main/java/org/apache/pinot/tools/PinotTableRebalancer.java
```diff
@@ -0,0 +1,123 @@
...
+    final TableRebalancer tableRebalancer = new TableRebalancer(helixManager, helixAdmin, clusterName);
+    final RebalanceSegmentStrategyFactory rebalanceSegmentStrategyFactory =
+        new RebalanceSegmentStrategyFactory(helixManager);
+    final RebalanceSegmentStrategy rebalanceSegmentsStrategy =
+        rebalanceSegmentStrategyFactory.getRebalanceSegmentsStrategy(tableConfig);
+    final RebalanceResult rebalanceResult =
+        tableRebalancer.rebalance(tableConfig, rebalanceSegmentsStrategy, rebalanceUserConfig);
+
+    if (rebalanceResult.getStatus() == RebalanceResult.RebalanceStatus.FAILED) {
+      System.out.println("Failed to rebalance table: " + tableName);
```
snleee: Are we not using `LOGGER` in tools package?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
snleee: +1

pinot-tools/src/main/java/org/apache/pinot/tools/PinotTableRebalancer.java
siddharthteotia: Initially didn't use it for manual testing. Switched to LOGGER now. It is being used in some of the other tools code as well.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/RebalanceCommand.java
siddharthteotia: printUsage() is overriden to also print some example usage of RebalanceCommand. So I let it print each option along with description by calling super.printUsage() and then print some sample usages.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
siddharthteotia: done

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
siddharthteotia: I am looking into it. But if I don't start broker, the table creation fails with HTTP bad request during the call to completeTableConfiguration that sends a POST request to controller for creating the new table.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/util/TableRebalancerTest.java
siddharthteotia: stats are verified further in the test

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: changed the name of these stats to avoid confusion, also tracking the number of times we updated the ideal state in ZK (which happens after we have made updates to instance map of each segment)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: updated javadoc

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -162,10 +164,12 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
         if (zkBaseDataAccessor
             .set(idealStateKey.getPath(), nextIdealState.getRecord(), currentIdealState.getRecord().getVersion(),
                 AccessOption.PERSISTENT)) {
+          LOGGER.debug("Successfully persisted the ideal state in ZK. Will wait for External view to converge");
           // if we succeeded, wait for the change to stabilize
           waitForStable(tableNameWithType);
           // clear retries as it tracks failures with each idealstate update attempt
           retries = 0;
+          LOGGER.debug("External view converged for the change in ideal state. Will start the next iteration now");
```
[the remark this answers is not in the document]
siddharthteotia: done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -162,10 +164,12 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
         if (zkBaseDataAccessor
             .set(idealStateKey.getPath(), nextIdealState.getRecord(), currentIdealState.getRecord().getVersion(),
                 AccessOption.PERSISTENT)) {
+          LOGGER.debug("Successfully persisted the ideal state in ZK. Will wait for External view to converge");
```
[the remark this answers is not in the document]
siddharthteotia: done

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -0,0 +1,352 @@
...
+    public StateModel createNewStateModel(String resourceName, String partitionName) {
+      return new FakeServerSegmentStateModelFactory.FakeSegmentStateModel();
+    }
+
+    @SuppressWarnings("unused")
+    @StateModelInfo(states = "{'OFFLINE', 'ONLINE', 'DROPPED'}", initialState = "OFFLINE")
+    public class FakeSegmentStateModel extends StateModel {
+
+      @Transition(from = "OFFLINE", to = "ONLINE")
+      public void onBecomeOnlineFromOffline(Message message, NotificationContext context) {
```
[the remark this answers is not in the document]
siddharthteotia: dont -- added a test to explicitly throw exception during state transition and check that rebalancer detects the error state in external view and bails out

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -0,0 +1,352 @@
...
+      if (idealState.getInstanceStateMap(segmentID) != null) {
+        idealState.getInstanceStateMap(segmentID).clear();
+      }
+    }
+    final ZkBaseDataAccessor zkBaseDataAccessor = (ZkBaseDataAccessor) dataAccessor.getBaseDataAccessor();
+    zkBaseDataAccessor.set(idealStateKey.getPath(), idealState.getRecord(), idealState.getRecord().getVersion(),
+        AccessOption.PERSISTENT);
+
+    // should be enough for the callbacks to come and go away
+    Thread.sleep(2000);
```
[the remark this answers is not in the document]
siddharthteotia: done -- using a new table in each test

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -0,0 +1,352 @@
...
+    Assert.assertEquals(0, stats.getIncrementalTransitions());
+
+    // as part of rebalancing, host2 lost segment1 and host3
+    // lost segment2 -- so 2 transitions from ON to OFF and
+    // OFF to DROP
+    Assert.assertEquals(2, _stateTransitionStats.offFromOn);
+    Assert.assertEquals(2, _stateTransitionStats.dropFromOff);
+
+    // clear for next test to avoid mixing state transition stats
+    clearIdealState(2);
```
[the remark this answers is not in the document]
siddharthteotia: done -- added a test

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/util/TableRebalancerTest.java
siddharthteotia: done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -162,10 +164,12 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
           // if we succeeded, wait for the change to stabilize
           waitForStable(tableNameWithType);
           // clear retries as it tracks failures with each idealstate update attempt
           retries = 0;
+          LOGGER.debug("External view converged for the change in ideal state. Will start the next iteration now");
           continue;
         }
         // in case of any error, we retry a bounded number of types
```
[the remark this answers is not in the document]
siddharthteotia: done, added info before attempting CAS on ZK. also dumped the changed ideal state (the one that we will try to atomically write in ZK)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: thought could be helpful for debugging -- this method has 3 debug level messages , one for entry in method and one each for adding and removing a host

siddharthteotia:
Addressed review comments and added more tests.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +390,162 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
-      }
-    } while (diff > 0);
+    try {
+      Thread.sleep(INITIAL_WAIT_MS);
+      while (true) {
+        if (isStable(resourceName)) {
+          break;
+        } else {
+          LOGGER.info("Waiting for externalView to match idealstate for table:" + resourceName);
+          Thread.sleep(EXTERNAL_VIEW_CHECK_INTERVAL_MS);
```
siddharthteotia: Although we now bail out on detecting error in external view, there is still a possibility of rebalancer waiting indefinitely for external view to converge simply because the changes haven't yet reflected and there is no error. I feel after updating ideal state in ZK when we come here to wait for stabilization, we should spend a fix amount of time in this method if isStable() never returns true

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,58 +145,88 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
+        // we will go back in the loop and reattempt by recomputing the target ideal state
         LOGGER.warn("Version changed while updating ideal state for resource: {}", tableNameWithType);
       } catch (Exception e) {
-        LOGGER.warn("Caught exception while updating ideal state for resource: {}", tableNameWithType, e);
+        if (e instanceof IllegalStateException && e.getCause() instanceof ExternalViewErrored) {
+          LOGGER.error("External view reported error for table {} after updating ideal state", tableNameWithType);
+          // remove the cause as it is private and is only used to detect exact error within this class
+          throw new IllegalStateException(e.getMessage());
+        } else {
+          LOGGER.warn("Caught exception while or after updating ideal state for resource: {}", tableNameWithType, e);
```
siddharthteotia: It makes total sense to continue (restart by regetting the current ideal state from ZK and recomputing the target) if the atomic operation failed in ZK due to version mismatch. 

However, if there is any other unpredictable exception (other than Illegal state exception due to error state in external view), I suggest we still bail out

siddharthteotia:
Ping.
All review comments have been addressed. Please take another look

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,58 +145,88 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
           // clear retries as it tracks failures with each idealstate update attempt
           retries = 0;
+          LOGGER.debug("External view converged for the change in ideal state for table {}. Will start the next iteration now",
+              tableNameWithType);
           continue;
         }
-        // in case of any error, we retry a bounded number of types
+        // in case of any error, we retry a bounded number of times
       } catch (ZkBadVersionException e) {
+        // we will go back in the loop and reattempt by recomputing the target ideal state
```
mcvsubbu: In the log in line 187, pls include the version you were expecting to see

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
mcvsubbu: agreed

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -212,39 +248,136 @@ private IdealState getNextState(IdealState currentState, IdealState targetState,
...
+   * @param rebalanceUserConfig rebalance configuration
    */
   @VisibleForTesting
-  public void updateSegmentIfNeeded(String segmentId, Map<String, String> srcMap, Map<String, String> targetMap,
-      IdealState idealState, Configuration rebalanceUserConfig) {
+  public void updateSegmentIfNeeded(String segmentId, Map<String, String> currentIdealStateSegmentHosts,
+      Map<String, String> targetIdealStateSegmentHosts, IdealState idealStateToUpdate,
+      Configuration rebalanceUserConfig) {
+
+    LOGGER.info("Will update instance map of segment: {}", segmentId);
```
mcvsubbu: i think this should move to debug since w will be logging once per segment. Tables can have 1000s of segments. All info logs in this method should turn into debug

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
mcvsubbu: I think it is better to error out ahead of time if the min replicas to keep alive is greater than or equal to the number of replicas of the table,

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
mcvsubbu: no, we can't have info logs per segment.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +390,162 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
   }
 
   /**
-   * Check if IdealState = ExternalView. If its not equal, return the number of differing segments.
+   * Check if external view has converged to ideal state
+   * @param tableName name of table that we are rebalancing
+   * @return true if external view is same as ideal state, false otherwise
    */
-  public int isStable(String tableName) {
+  private boolean isStable(String tableName) {
```
mcvsubbu: If we return a count of number of segments converged, that can show a progress bar. I am fine if that is a subsequent PR

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
mcvsubbu: agreed. Fixed, max time and the bail out.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +390,162 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
+   * Helper class that maintains stats that
+   * are later checked in tests to verify
+   * the behavior of the algorithm here
+   * that takes from current ideal state
+   * to a target ideal state
+   */
+  public static class RebalancerStats {
+    private int dryRun;
+    private int updatestoIdealStateInZK;
+    private int directUpdatesToSegmentInstanceMap;
```
mcvsubbu: which one of these equals numSegmentsMoved?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +390,162 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
+     * state in ZK as it progresses through
+     * rebalancing.
+     * @return ideal state updates in ZK
+     */
+    public int getUpdatestoIdealStateInZK() {
+      return updatestoIdealStateInZK;
+    }
+
+    /**
+     * Get the number of times we updated the instance-state
```
mcvsubbu: I am not clear on what this means. Is it the max number of times we updated in one step?

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
mcvsubbu: It could be that the table add specifies broker instances, and the controller checks if there are such instances.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,58 +145,88 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
             .set(idealStateKey.getPath(), nextIdealState.getRecord(), currentIdealState.getRecord().getVersion(),
                 AccessOption.PERSISTENT)) {
+          LOGGER.debug("Successfully persisted the ideal state for table {} in ZK. Will wait for External view to converge",
+              tableNameWithType);
+          ++_rebalancerStats.updatestoIdealStateInZK;
           // if we succeeded, wait for the change to stabilize
           waitForStable(tableNameWithType);
           // clear retries as it tracks failures with each idealstate update attempt
           retries = 0;
+          LOGGER.debug("External view converged for the change in ideal state for table {}. Will start the next iteration now",
```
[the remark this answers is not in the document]
siddharthteotia: Moved to info. We only check at the beginning of the while loop if next iteration is needed. I have changed the log message  to have "will start next iteration (if any)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,58 +145,88 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
-      // Check version and set ideal state
+      // Check version and set ideal state to nextIdealState
       try {
-        LOGGER.info("Updating IdealState for table {}", tableNameWithType);
+        LOGGER.info("Going to update current IdealState in ZK for table {}, current version {} and creation time {}",
+            tableNameWithType, currentIdealState.getRecord().getVersion(), currentIdealState.getRecord().getCreationTime());
         if (zkBaseDataAccessor
             .set(idealStateKey.getPath(), nextIdealState.getRecord(), currentIdealState.getRecord().getVersion(),
                 AccessOption.PERSISTENT)) {
+          LOGGER.debug("Successfully persisted the ideal state for table {} in ZK. Will wait for External view to converge",
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Moved to debug

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Done. using a max wait time of 5mins.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -110,6 +115,8 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
           strategy.getRebalancedIdealState(previousIdealState, tableConfig, rebalanceConfig, partitionAssignment);
       result.setIdealStateMapping(idealState.getRecord().getMapFields());
       result.setPartitionAssignment(partitionAssignment);
+      result.setStatus(RebalanceResult.RebalanceStatus.DONE);
+      _rebalancerStats.dryRun = 1;
```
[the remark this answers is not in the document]
siddharthteotia: For a given invocation of Rebalancer tool (either through command line or rest api), we create a single instance of TableRebalancer and invoke the rebalance() method. The stats are computed in a cumulative manner for that invocation. However, if the user invoked it in dry run mode then dryRun is set to 1 -- it can only be 1 for that invocation

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: The reason I didn't do it is that I thought number of replicas could be different for each segment. So to validate the user specified value, we have to go over the ideal state and for any segment if the condition is not met, we throw error. IdealState has a method getMinReplicaCount() and I am not sure if that can be used straight without iterating over all the segments.

Secondly, during the course of rebalancing, ideal state might change in ZK (due to some concurrent operation), and out version update will fail in that case and then we fetch the new ideal state and recompute target. So this requires that we still safeguard our code where we update one segment at a time to check for this condition as just checking once upfront upon entry into rebalancer code might not be enough.

I have now added a validation method we call upfront here before rebalancer does anything. I am wondering if there is any quick way to do so without going over the ideal state.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Sounds good.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: No, this is the total number of time we updated the host-state mapping of a segment to target mapping in a single step. This is cumulative across all segments.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Done in the same PR -- keeping track of the number of segments not converged and then printing a log message

pinot-common/src/main/java/org/apache/pinot/common/restlet/resources/RebalanceResult.java
```diff
@@ -28,17 +28,19 @@
 public class RebalanceResult {
   private Map<String, Map<String, String>> idealStateMapping;
   private PartitionAssignment partitionAssignment;
-  private String status;
+  private RebalanceStatus status;
```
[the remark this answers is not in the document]
siddharthteotia: done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: None of them -- I am deriving the num segments moved in the caller of the method that updates the instance map of each segment in a loop. Then I am also logging the difference between the updated ideal state and target

siddharthteotia:
Addressed latest review comments

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
mcvsubbu: You can use the Idealstate to get the number of replicas.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/RebalanceCommand.java
mcvsubbu: Can the --help option serve here?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -472,7 +472,9 @@ public String rebalance(
       @ApiParam(value = "offline|realtime") @Nonnull @QueryParam("type") String tableType,
       @ApiParam(value = "true|false") @Nonnull @DefaultValue("true") @QueryParam("dryrun") Boolean dryRun,
       @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("includeConsuming") Boolean includeConsuming,
-      @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("downtime") Boolean downtime) {
+      @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("downtime") Boolean downtime,
+      @ApiParam(value = "number of serving replicas to keep alive per segment while rebalancing in no downtime mode")
+      @DefaultValue("1") @QueryParam("minReplicasToKeepUpForNoDowntime") Integer minReplicasToKeepUp) {
```
mcvsubbu: Suggest "minAvailableReplicas". @snleee can you think of other possibly better param names?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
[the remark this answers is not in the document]
mcvsubbu: my bad, sorry

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
snleee: Sounds better than the current name :)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,112 +153,293 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
         LOGGER.error("Got interrupted while rebalancing table {}", tableNameWithType);
+        result.setStatus("Rebalancer got interrupted");
+        result.setStatusCode(RebalanceResult.RebalanceStatus.FAILED);
         Thread.currentThread().interrupt();
         return result;
       }
     }
   }
 
+  private void validateMinReplicas(final String tableNameWithType,
```
snleee: I think that checking replication factor from table config should be enough for validation.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,112 +153,293 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
+        // we will go back in the loop and reattempt by recomputing the target ideal state
+        LOGGER.info("Version changed while updating ideal state for resource: {}, was expecting version {}",
+            tableNameWithType, currentIdealState.getRecord().getVersion());
       } catch (Exception e) {
-        LOGGER.warn("Caught exception while updating ideal state for resource: {}", tableNameWithType, e);
+        if (e instanceof IllegalStateException && e.getCause() instanceof ExternalViewErrored) {
+          if (e.getCause() instanceof ExternalViewErrored) {
+            LOGGER.error("External view reported error for table {} after updating ideal state", tableNameWithType);
+          } else if (e.getCause() instanceof ExternalViewConvergeTimeout) {
+            LOGGER.error("Timedout while waiting for external view to converge for table {}", tableNameWithType);
```
snleee: `Timed out`

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,112 +153,293 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
-      LOGGER.info("Segment " + segmentId + " missing from current idealState. Skipping it.");
+      LOGGER.debug("Segment " + segmentId + " missing from current idealState. Skipping it.");
       return;
     }
 
-    if (rebalanceUserConfig
-        .getBoolean(RebalanceUserConfigConstants.DOWNTIME, RebalanceUserConfigConstants.DEFAULT_DOWNTIME)) {
-      setTargetState(idealState, segmentId, targetMap);
+    // dump additional detailed info for debugging
+    if (LOGGER.isDebugEnabled()) {
```
snleee: do we really need this check?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +452,198 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
+        // we have found that external view hasn't yet converged to ideal state.
+        // still go on to check for other segments just so that we can dump debug
+        // messages or we can detect error. that's why we don't return
+        // false immediately
+        stable = false;
+        ++segmentsNotConverged;
+        continue;
+      }
+
+      if (LOGGER.isDebugEnabled()) {
```
snleee: Do we need this check?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +452,198 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
+        prettyPrintMap(mapEV, Level.DEBUG);
+      }
 
       for (String server : mapIS.keySet()) {
-        String state = mapIS.get(server);
-        if (mapEV == null || mapEV.get(server) == null || !mapEV.get(server).equals(state)) {
-          LOGGER.debug("Mismatch: segment" + segment + " server:" + server + " state:" + state);
-          numDiff = numDiff + 1;
+        if (!mapEV.containsKey(server)) {
+          LOGGER.info("Host-state mapping of segment {} doesn't yet have server {} in external view",
```
snleee: I think that we should log `debug` level only in this loop. You're logging `num segments * replication` lines of log. This can easily be equivalent to thousands of lines of logs.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +452,198 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
+          LOGGER.error("Detected error state for segment {} for server {}", segment, server);
+          prettyPrintMap(mapIS, Level.ERROR);
+          prettyPrintMap(mapEV, Level.ERROR);
+          throw new IllegalStateException("External view reports error state for segment " + segment + " for host " + server,
+              new ExternalViewErrored());
+        } else {
+          final String stateInIdealState = mapIS.get(server);
+          final String stateInExternalView = mapEV.get(server);
+          if (!stateInIdealState.equalsIgnoreCase(stateInExternalView)) {
+            LOGGER.info("Host-state mapping of segment {} has state {} in external view and state {} in ideal state",
```
snleee: The same as above. I would prefer to use `debug` level.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +452,198 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
-        if (mapEV == null || mapEV.get(server) == null || !mapEV.get(server).equals(state)) {
-          LOGGER.debug("Mismatch: segment" + segment + " server:" + server + " state:" + state);
-          numDiff = numDiff + 1;
+        if (!mapEV.containsKey(server)) {
+          LOGGER.info("Host-state mapping of segment {} doesn't yet have server {} in external view",
+              segment, server);
+          // external view not yet converged
+          stable = false;
+          converged = false;
+        } else if (mapEV.get(server).equalsIgnoreCase("error")) {
```
snleee: I think that we should add the validation to check if there's any `ERROR` state in the external view before running any rebalance logic. In that case, we know for sure that `error` state caught here is caused by the rebalance. Then, it makes sense to stop the rebalance.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Done. No longer iterating over ideal state. Just using idealstate.getReplicas()

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,112 +153,293 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
+        RebalanceUserConfigConstants.DEFAULT_DOWNTIME)) {
+      final int minReplicasToKeepUp = rebalanceConfig
+          .getInt(RebalanceUserConfigConstants.MIN_REPLICAS_TO_KEEPUP_FOR_NODOWNTIME,
+              RebalanceUserConfigConstants.DEFAULT_MIN_REPLICAS_TO_KEEPUP_FOR_NODOWNTIME);
+      Map<String, Map<String, String>> currentMapFields = idealState.getRecord().getMapFields();
+      for (String segment : currentMapFields.keySet()) {
+        final Map<String, String> segmentHostsMap = currentMapFields.get(segment);
+        if (minReplicasToKeepUp >= segmentHostsMap.size()) {
+          final StringBuilder sb = new StringBuilder();
+          sb.append("User specified invalid number of min replicas: ").append(minReplicasToKeepUp)
```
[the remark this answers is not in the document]
siddharthteotia: Changed to use logger. My main reason for using string builder was to construct the message once and use it in both logger and exception

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: I thought we can check upfront to avoid running unnecessary logic that is needed to pretty print map by using joiner.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Same as before. prettyPrintMap() function first uses joiner to come up with a pretty print version of map. I thought we can avoid it all together upfront by checking if debug is enabled or not.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Only for the segments and servers where external view hasn't converge. Right?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: This log will be only for segment and server combination where state is not same as the one in ideal state. Is that fine in debug?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/RebalanceCommand.java
siddharthteotia: No that doesn't print example usage.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
siddharthteotia: Done.

siddharthteotia:
Addressed review comments. Added integration tests for realtime LLC (consuming, without consuming), realtime HLC and segment movement stats.

@snleee  brought up a point w.r.t throwing exception on detecting ERROR state in external view.

Currently we throw exception on detecting error in external view. What are the pros and cons of ignoring it and continuing with the hope that probably there is 1 of many segments that is in ERROR state and may be we can continue with rebalancing v/s the current approach of aborting the rebalancer midway by throwing exception.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
snleee: I was mentioning `replication` and `numReplicasPerPartition` from table config. If you want to use `idealstate.getReplicas()`, I think that you should double check on how it gets updated. I think that we usually don't depend on that number.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
snleee: oh i missed the if check :)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Initial thought was to use one of the methods from table config but @mcvsubbu suggested to use the ideal state method. I can see that setReplicas() method on ideal state is being invoked through PinotTableRestletResource -> PinotHelixResourceManager -> PinotTableIdealStateBuilder -> IdealStateBuilder. In my unit tests I see this happening every time a post request is sent to the controller for creating the table. In fact the value that is being passed down to IdealState.setReplicas() is coming from table config's SegmentValidationAndRetetionConfig.getReplication()

mcvsubbu: Checking from table config is a bit involved, different for offline, hlc and llc. This check is done during table creation, and the idealstate is created with the right number of replicas. Since we are getting idealstate here anyway, it makes sense to check for replicas on the idealstate

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -0,0 +1,877 @@
...
+  @Test
+  public void testHLCRealtimeTable() throws Exception {
+    final int numSegments = 4;
+    final int numReplicas = 3;
+    final int numAdditionalServers = 3;
+    final String tableName = getTableName() +"HLC";
+
+    try {
+      startFakeServers(NUM_INITIAL_SERVERS, CommonConstants.Helix.DEFAULT_SERVER_NETTY_PORT, true);
+      startKafka();
```
mcvsubbu: things may run faster if you startKafka and just keep it around and stopKafka() at the end. We don't consume any messages, so it is ok to just keep it running throughout.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/RebalanceCommand.java
mcvsubbu: Can you change the super method to call a printExamples() and printUsage()? That way other commands can also take advantage of this to print examples.  No better documentation than the one printed by the code!

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -0,0 +1,877 @@
...
+      // segment is moved from one server to another, the latter throws
+      // exception on state transition from OFFLINE to ONLINE
+      _raiseErrorOnSegmentStateTransition = true;
+
+      // rebalance
+      final PinotTableRebalancer tableRebalancer = new PinotTableRebalancer(ZKSTR, getHelixClusterName(), false, true, false, 2);
+      tableRebalancer.rebalance(tableName, "OFFLINE");
+      Assert.fail("Expecting exception");
+
+      final TableRebalancer.RebalancerStats stats = tableRebalancer.getRebalancerStats();
```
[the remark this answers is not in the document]
siddharthteotia: Good catch. Done.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -0,0 +1,877 @@
...
+      // we would have persisted the new ideal state in ZK and then
+      // waited for external view to converge and bailed out on seeing
+      // error state
+      Assert.assertEquals(stats.getUpdatestoIdealStateInZK(), 1);
+    } catch (Exception e) {
+      Assert.assertTrue(e.getMessage().contains("External view reports error state for segment segment0"));
+    }
+    _raiseErrorOnSegmentStateTransition = false;
+    stopFakeServers();
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/util/TableRebalancerTest.java
```diff
@@ -117,4 +119,332 @@ public void downtimeUpdateWithNoCommonElements() {
     Map<String, String> tempMap = current.getRecord().getMapField(segmentId);
     Assert.assertTrue(EqualityUtils.isEqual(tempMap, targetMap));
   }
+
+  /**
+   * Test for now downtime rebalance with no common hosts between
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -472,7 +472,9 @@ public String rebalance(
       @ApiParam(value = "offline|realtime") @Nonnull @QueryParam("type") String tableType,
       @ApiParam(value = "true|false") @Nonnull @DefaultValue("true") @QueryParam("dryrun") Boolean dryRun,
       @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("includeConsuming") Boolean includeConsuming,
-      @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("downtime") Boolean downtime) {
+      @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("downtime") Boolean downtime,
+      @ApiParam(value = "number of serving replicas to keep alive per segment while rebalancing in no downtime mode")
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -472,7 +472,9 @@ public String rebalance(
       @ApiParam(value = "offline|realtime") @Nonnull @QueryParam("type") String tableType,
       @ApiParam(value = "true|false") @Nonnull @DefaultValue("true") @QueryParam("dryrun") Boolean dryRun,
       @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("includeConsuming") Boolean includeConsuming,
-      @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("downtime") Boolean downtime) {
+      @ApiParam(value = "true|false") @DefaultValue("false") @QueryParam("downtime") Boolean downtime,
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -67,15 +72,18 @@
 
   private static final int MAX_RETRIES = 10;
   private static final int EXTERNAL_VIEW_CHECK_INTERVAL_MS = 30000;
+  private static final int EXTERNVAL_VIEW_STABILIZATON_MAX_WAIT_MS = 300000;
```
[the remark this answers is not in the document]
siddharthteotia: Done. using limit of 20mins for now. We can refine this in follow-up PR

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +451,208 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
+          }
+        }
+      } catch (InterruptedException e) {
+      LOGGER.error("Rebalancer got interrupted while waiting for external view to converge");
+      Thread.currentThread().interrupt();
+    }
+
+    if (!done) {
+      LOGGER.error("External view didn't converge for table {} within max wait time of 5mins. Bailing out",
+          resourceName);
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
siddharthteotia: Done. I feel we should return with FAILEd status code from the rebalancer once the pre-check on the external view detects error state. The reason being that there is probably no gain in going forward with rebalancing if external view is already not fine to begin with.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/RebalanceCommand.java
siddharthteotia: Done

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +451,208 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
+     * in no-downtime rebalancing if there are sufficient common
+     * hosts between current and target ideal states
+     * @return direct updates
+     */
+    public int getDirectUpdatesToSegmentInstanceMap() {
+      return directUpdatesToSegmentInstanceMap;
+    }
+
+    /**
+     * Get the number of times we updated the instance-state
```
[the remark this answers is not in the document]
siddharthteotia: Not sure what that implies. I have updated the javadoc if it is more clear now. We can discuss this

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -257,47 +451,208 @@ private void setTargetState(IdealState idealState, String segmentId, Map<String,
...
+     * rebalancing.
+     * @return ideal state updates in ZK
+     */
+    public int getUpdatestoIdealStateInZK() {
+      return updatestoIdealStateInZK;
+    }
+
+    /**
+     * Get the number of times we updated the instance-state
+     * mapping acrosss all segments in ideal state in one step. This
```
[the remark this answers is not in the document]
siddharthteotia: Not sure what that implies. I have updated the javadoc if it is more clear now. We can discuss this

siddharthteotia:
Addressed all review comments

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -0,0 +1,868 @@
...
+        idealStateCopy.setPartitionState(segmentID, instanceID, "ONLINE");
+      }
+    }
+    // this CAS serves no purpose in this single threaded execution test
+    final ZkBaseDataAccessor zkBaseDataAccessor = (ZkBaseDataAccessor) dataAccessor.getBaseDataAccessor();
+    zkBaseDataAccessor.set(idealStateKey.getPath(), idealStateCopy.getRecord(), idealState.getRecord().getVersion(),
+        AccessOption.PERSISTENT);
+  }
+
+  public class FakeServerSegmentStateModelFactory extends StateModelFactory<StateModel> {
```
mcvsubbu: would be a good thing to re-use methods to add/stop fake instances in ControllerTest (jackie cleaned up the tests recently), but we can leave that to a later PR. ControllerTest is a super-class of TableRebalancerAdminToolClusterIntegrationTest
Can you please add a TODO here?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -124,13 +123,32 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
       return result;
     }
 
+    // before running rebalancer if external view is fine to begin with
```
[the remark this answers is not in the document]
siddharthteotia: Moved it to before dry run

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/util/TableRebalancerTest.java
```diff
@@ -112,9 +106,322 @@ public void downtimeUpdateWithNoCommonElements() {
...
+    TableRebalancer.RebalancerStats rebalancerStats = rebalancer.getRebalancerStats();
+    // STEP 1 -- incremental update to segment's partition map
+    Assert.assertEquals(rebalancerStats.getIncrementalUpdatesToSegmentInstanceMap(), 1);
+    // STEP 2 -- final direct update to segment's partition map
+    Assert.assertEquals(rebalancerStats.getDirectUpdatesToSegmentInstanceMap(), 1);
+  }
+
+  private void verifyStateForMinReplicaConstraint(
+      final IdealState updated, final IdealState target,
+      final int same, final boolean sizeCheck) {
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/util/TableRebalancerTest.java
```diff
@@ -112,9 +106,322 @@ public void downtimeUpdateWithNoCommonElements() {
...
+    Assert.assertEquals(rebalancerStats.getDirectUpdatesToSegmentInstanceMap(), 1);
+  }
+
+  /**
+   * Test for no downtime rebalance with no common hosts between
+   * current and target ideal state and a request to keep an invalid
+   * number of minimum replicas up while rebalancing
+   */
+  @Test
+  public void noDowntimeUpdateWithNoCommonHostsAndInvalidMinReplicas() {
```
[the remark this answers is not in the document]
siddharthteotia: This test file doesn't exercise the full code in TableRebalancer. It was written that way originally. IT directly calls the updateSegment() method that just updates the ideal state -- the check is still done there to safeguard code.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
siddharthteotia: Done

siddharthteotia:
Addressed latest review comments

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/TableRebalancer.java
```diff
@@ -139,112 +171,293 @@ public RebalanceResult rebalance(TableConfig tableConfig, RebalanceSegmentStrate
...
       return;
     }
 
-    MapDifference difference = Maps.difference(targetMap, srcMap);
-    if (!difference.entriesInCommon().isEmpty()) {
-      // if there are entries in common, there won't be downtime
-      LOGGER.debug("Segment " + segmentId + " has common entries between current and expected ideal state");
-      setTargetState(idealState, segmentId, targetMap);
+    // we are in no-downtime mode
+    if (minAvailableReplicas >= currentIdealStateSegmentHosts.size()) {
```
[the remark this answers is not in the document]
mcvsubbu: Not sure what you mean by replicas going down .. we bail out if we see ERROR state in externalview.  We look at the idealstate for numreplicas, so if something changes in the tableconfig we are agnostic. we don't (yet) support the scenario of multiple instances of rebalancer tool running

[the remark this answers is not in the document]
siddharthteotia: Discussed offline that is fine to keep the check

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/util/TableRebalancerTest.java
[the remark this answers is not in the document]
mcvsubbu: ok, i think it is good to have that code with the minreplicas for possible race conditions when 2 of these are run, or a background and foreground versions are run together.

siddharthteotia:
Thanks @snleee , @mcvsubbu for reviewing and approving the changes.

Some follow up items I will address in upcoming PRs:

1 - Revisit the logic where we fetch the ideal state again and recompute the target. There is potential room for optimization here  -- fetch the ideal state again and recompute target only if we previously failed to write in ZK due to version mismatch.

2. For the rest api based access that runs rebalance in async mode, we need to show which controller is responding to the rebalance request.

3. Re-use the recently merged controller test apis in the integration test to start/stop fake servers.

4. See if logging can be improved/changed to ensure info level messages are giving enough information.

PR #4449, opened 2019-07-19
Add resource config for lead controller resource

jackjlli:
This PR:
* adds resource config for lead controller resource to enable/disable the resource.
* explicitly enable lead controller resource, since the logic for this resource hasn't been merged yet.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
@@ -20,9 +20,8 @@ package org.apache.pinot.controller.helix.core.util;
 
 import com.google.common.base.Preconditions;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import org.apache.helix.ConfigAccessor;
 import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixManager;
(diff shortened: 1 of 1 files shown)


Jackie-Jiang:
You want to add the field to enable/disable the resource right?

jackjlli:
@Jackie-Jiang Yes, but we can also add the code saying if the field is missing or the field is `false`, then the resource is disabled by default.

Jackie-Jiang:
You should still put the field when creating the resource config so it is more explicit, and easier for us to change

jackjlli:
> You should still put the field when creating the resource config so it is more explicit, and easier for us to change

Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/util/HelixSetupUtils.java
```diff
@@ -177,5 +180,16 @@ private static void createLeadControllerResourceIfNeeded(String helixClusterName
...
       helixAdmin.addResource(helixClusterName, LEAD_CONTROLLER_RESOURCE_NAME, idealState);
     }
+
+    // Create resource config for lead controller resource if it doesn't exist
+    ResourceConfig resourceConfig = configAccessor.getResourceConfig(helixClusterName, LEAD_CONTROLLER_RESOURCE_NAME);
+    if (resourceConfig == null) {
+      resourceConfig = new ResourceConfig(LEAD_CONTROLLER_RESOURCE_NAME);
+    }
+    // Set RESOURCE_ENABLED to false if it's absent in resource config
+    if (resourceConfig.getSimpleConfig("RESOURCE_ENABLED") == null) {
```
Jackie-Jiang: putIfAbsent?

jackjlli: I'd love to but `ResouceConfig` isn't a hashmap and there's no such method for now. :)

Jackie-Jiang: Make sense lol

PR #4451, opened 2019-07-19
Enable SegmentCompleteionIntegrationTest

mcvsubbu:
This test was not running because of a typo in the class name.
Also, re-organized the test so that it does not have to override
methods from upper level tests each time new tests are added.

### DIFF
--- a/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/BaseClusterIntegrationTest.java
+++ b/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/BaseClusterIntegrationTest.java
@@ -426,15 +426,10 @@ public abstract class BaseClusterIntegrationTest extends ClusterTest {
     completeTableConfiguration();
   }
 
-  protected void completeTableConfiguration() {
+  protected void completeTableConfiguration() throws IOException {
     if (isUsingNewConfigFormat()) {
       CombinedConfig combinedConfig = new CombinedConfig(_offlineTableConfig, _realtimeTableConfig, _schema);
-      try {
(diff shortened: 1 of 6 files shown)


pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/BaseClusterIntegrationTest.java
```diff
@@ -398,4 +403,38 @@ protected void testSqlQuery(@Nonnull String sqlQuery, @Nullable List<String> sql
...
+    completeTableConfiguration();
+  }
+
+  protected void completeTableConfiguration() {
+    if (isUsingNewConfigFormat()) {
+      CombinedConfig combinedConfig = new CombinedConfig(_offlineTableConfig, _realtimeTableConfig, _schema);
+      try {
+        sendPostRequest(_controllerRequestURLBuilder.forNewTableCreate(), Serializer.serializeToString(combinedConfig));
+      } catch (IOException e) {
+        throw new RuntimeException(e);
```
Jackie-Jiang: Allow method to throw Exception?

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/BaseClusterIntegrationTest.java
```diff
@@ -398,4 +403,38 @@ protected void testSqlQuery(@Nonnull String sqlQuery, @Nullable List<String> sql
...
+      }
+    }
+  }
+
```
Jackie-Jiang: (nit) extra empty line

Issue #4450, opened 2019-07-19
Support for SQL compliant Order By in PQL


mayankshriv:
Pinot currently does not support `Order By` in PQL. The current behavior is:

1. Throw exception for the following query:
    `select sum(m1) from table where ... group by d1, d2 order by d1`

2. Silently ignore `order by` for the following query. Each group is sorted based on its aggregation value.
 `select sum(m1), d1, d2 from table where ... group by d1, d2 order by d1`

The proposal is to support support the SQL compliant `Order By` in PQL for both the queries above. With this features. Upon initial investigation, it appears that the existing group-by execution code has several performance related optimizations that make it hard to support the new feature. So the initial proposal is as follows:

1. The existing `BrokerRequest` implementation has tied the `Order By` to `selection` queries only. In order to maintain backward compatibility, the proposal is to add a top level `Order By` in the `request.thrift` that would initially apply to `group-by`, and after drop the `SelectionSort` part.

2. In the presence of `Order By`, generate a new plan using new operator to have a parallel implementation for this new feature. Once this code is matured (for functionality and performance), the existing functionality can also be moved to use the new code.

3. The new operator will work on tabular data as opposed to Map with single group key and a value. This tabular structure would make it more generic and reusable for other queries (selection), as well as allow for creating customized sort orders (column based).

4. With SQL compliant behavior, we will now be able to perform segment level trimming of results in presence of multiple aggregation functions. This does not happen today because it breaks the existing behavior as dropping key for one aggregation function also drop it for all others (not desired for current non SQL behavior).

npawar:
Doc outlining approaches and thoughts: https://cwiki.apache.org/confluence/display/PINOT/Order+By

Jackie-Jiang:
Supported in SQL (PQL is deprecated)

PR #4453, opened 2019-07-19
Issue #4450: Support parsing of `Order By` (with group-by) clause in PQL compiler.

mayankshriv:
This is the first of several PR's to come for implementing group-by +
order-by support in PQL. The changes in this PR are limited to parsing
the order-by clause.

1. Added a top level orderBy field in BrokerRequest. The existing
selectionSort field is tied to selection only queries, and removing it
will break backward compatibility. Once the code is moved to use
PinotQuery instead of BrokerRequest, this will get cleaned up
automatically.

2. Moved to latest Thrift version 0.12.0. This required removing the
org.apache.commons.lang3 exclusion from pom.xml.

3. Added unit test for order-by compilation.


mayankshriv:
@snleee Thanks for bringing up the concern. Although I would be surprised if these versions of Thrift generates incompatible code for same input, I did go ahead and test the scenario you mentioned and it works as expected.

pinot-common/src/thrift/request.thrift
```diff
@@ -154,6 +154,7 @@ struct BrokerRequest {
  15: optional HavingFilterQuery havingFilterQuery;
  16: optional HavingFilterQueryMap havingFilterSubQueryMap;
  17: optional query.PinotQuery pinotQuery;
+ 18: optional list<SelectionSort> orderBy;
```
kishoreg: is it possible to rename SelectionSort to OrderBy?

mayankshriv: Yes, I thought about that as well. It might cause backward compatibility issues. Given that we are going to move to PinotQuery and drop BrokerReuqest this won't be an issue for long, so I decided it leave it as-is.

kishoreg: Makes sense.

PR #4455, opened 2019-07-22
Fix num entries scanned in filter for MultiValue scanner

jasperjiaguo:
The calculation of numEntriesScanned in MVScanDocIdIterator does not correctly reflect the actual number entries in multi-value case. The counter should be increase by the actual entries scanned in a multi-value instead of 1.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/dociditerators/MVScanDocIdIterator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/dociditerators/MVScanDocIdIterator.java
@@ -80,9 +80,12 @@ public class MVScanDocIdIterator implements ScanBasedDocIdIterator {
       return false;
     }
     valueIterator.skipTo(docId);
-    _numEntriesScanned++;
+    //_numEntriesScanned++;
     int length = valueIterator.nextIntVal(intArray);
-    return evaluator.applyMV(intArray, length);
+    int[] numEntriesScannedInEvaluator={0};
+    boolean ret=evaluator.applyMV(intArray, length,numEntriesScannedInEvaluator);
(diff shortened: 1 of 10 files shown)


jasperjiaguo:
@Jackie-Jiang Hi Jackie can you review this part of the code? Seems you implemented the counter. Thanks.

pinot-core/src/main/java/org/apache/pinot/core/operator/dociditerators/MVScanDocIdIterator.java
```diff
@@ -80,9 +81,11 @@ public boolean isMatch(int docId) {
       return false;
     }
     valueIterator.skipTo(docId);
-    _numEntriesScanned++;
     int length = valueIterator.nextIntVal(intArray);
-    return evaluator.applyMV(intArray, length);
+    MutableInt numEntriesScannedInEvaluator = new MutableInt(0);
+    boolean ret = evaluator.applyMV(intArray, length, numEntriesScannedInEvaluator);
+    _numEntriesScanned += numEntriesScannedInEvaluator.toInteger();
```
Jackie-Jiang: You can make _numEntriesScanned a MutableInt and keep passing it to the evaluators. In the getNumEntriesScanned(), return _numEntriesScanned.intValue()

Jackie-Jiang: Please also fix the tests

sunithabeeram: +1

sunithabeeram:
It looks like the tests are failing due to timeout. @jasperjiaguo can you kick off the travis job again?

PR #4459, opened 2019-07-22
Enable periodic rebalance as a temporary work-around for the Helix issue

Jackie-Jiang:
Helix issue: https://github.com/apache/helix/issues/331


mcvsubbu:
If this work-around indeed fixes the issue, we should code this up in the controller starter and enable it using configuration so that it can be used in production if needed. The configuration can be very specific (e.g. "workaround_helix_331") so that we can drop the config once the bug is fixed.

Jackie-Jiang:
@mcvsubbu This work-around is a cluster config, so if we want to enable it in production, we can simply add it into the cluster config on ZK.

PR #4463, opened 2019-07-23
Ensuring partition config is valid during Hadoop preprocess

jenniferdai:
* Ensuring partition function is valid prior to preprocess


pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -286,6 +287,12 @@ private void validateConfigsAgainstSchema(Schema schema) {
           .format("Partition column: %s is not found from the schema of input files.", _partitionColumn));
       Preconditions.checkArgument(_numberOfPartitions > 0, String.format("Number of partitions should be positive. Current value: %s", _numberOfPartitions));
       Preconditions.checkArgument(_partitionFunction != null, "Partition function should not be null!");
+      try {
+        PartitionFunctionFactory.PartitionFunctionType.fromString(_partitionFunction);
+      } catch (IllegalArgumentException e) {
+        _logger.info("Partition function needs to be one of Modulo, Murmur, ByteArray, HashCode");
```
jackjlli: It'd be good to print out the current value of the setting in the exception message.

Issue #4353, opened 2019-06-21
Pre-process


jenniferdai:
### Smart PBNJ
with contributions from @jackjlli (prototype: https://github.com/apache/incubator-pinot/pull/4253)

**Motivation**
Currently, users have to do a lot of pre-processing and have to know a lot about the internals of Pinot in order to generate the Pinot segment format. We want to prevent all of this and hope this project can improve cluster stability & reduce/eliminate issues with many small segments.

The scope of this project includes:
- Sorting and partitioning for our users
- Generating segments of a proper size for users. This prevents problems with having many small segments or super large segments.

### Design
![SPBNJ - Smart Pinot Build and Push Job](https://user-images.githubusercontent.com/5035129/59954163-33d22400-9438-11e9-8572-22b268362852.png)

**Configurations**
If set, we will fetch the following parameters from a new section of table config called hadoopConfigs. 
enable.partitioning
enable.sorting
min.num.output.files
max.num.records

_enable.partitioning_
False by default. If true, partition column, partition function and number of partitions within table config will be used to generate segments that contain data from a single partition. Number of partitions determines number of reducers.
_enable.sorting_
If this parameter is set and a valid sorted column is declared within the table indexing config (sometimes, people declare multiple which is invalid), we will sort user data.
_min.num.output.files_
This parameter determines the number of reducers when partitioning is disabled. 
_max.num.records_
This parameter refers to the maximum records that can be in a file. When not set, each reducer will generate one output file. When set, each reducer will start another file once it meets the configured limit. Caveat - see Smart Segment Resizing

**Smart Segment Resizing**
We will use max.num.records to resize segments to create performant datasets. This parameter will help us avoid problems with too many poorly-sized Pinot segments that will cause us to use many server threads.

One problem with this approach will be the possibility of segments that are unbalanced in size. We will correct this by setting a minimum percentage threshold for which a small number of extra records will be folded into the latest file created within that reducer.

**Multiple Formats**
To support each output format, we need to be plug in different format-specific classes within the pre-processing job.
job.setMapOutputKeyClass(AvroKey.class)
job.setMapOutputValueClass(AvroValue.class)
job.setOutputKeyClass(AvroKey.class)
AvroMultipleOutputs….
For ORC, for example, we have these classes available. 
http://hadoopathome.logdown.com/posts/277986-using-multipleoutputs-with-orc-in-mapreduce
https://orc.apache.org/docs/mapreduce.html

[...]

PR #4464, opened 2019-07-23
Delete extra refreshed segments after segment push

jenniferdai:
* Adds a config to make possible deleting extra segments that are pushed within the same configured time segment or refreshed
* Future work: deal with invalid prefixes (when people accidentally change their prefix, when they keep a huge amount of time in the same segment for append segments, etc)
* #4353

### DIFF
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
@@ -58,5 +58,5 @@ public class JobConfigConstants {
   public static final String ENABLE_SORTING = "enable.sorting";
   public static final String ENABLE_RESIZING = "enable.resizing";
 
-  public static final String DELETE_EXTRA_REFRESHED_SEGMENTS = "delete.extra.refreshed.segments";
+  public static final String IS_DELETE_EXTRA_SEGMENTS = "is.delete.extra.segments";
 }
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
@@ -57,4 +57,6 @@ public class JobConfigConstants {
   public static final String ENABLE_PARTITIONING = "enable.partitioning";
   public static final String ENABLE_SORTING = "enable.sorting";
   public static final String ENABLE_RESIZING = "enable.resizing";
+
+  public static final String DELETE_EXTRA_REFRESHED_SEGMENTS = "delete.extra.refreshed.segments";
 }
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
@@ -58,5 +58,7 @@ public class JobConfigConstants {
   public static final String ENABLE_SORTING = "enable.sorting";
   public static final String ENABLE_RESIZING = "enable.resizing";
 
+  // This setting should be used if you will generate less # of segments after
+  // push. In preprocessing, this is likely because we resize segments. 
   public static final String IS_DELETE_EXTRA_SEGMENTS = "is.delete.extra.segments";
 }
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
@@ -59,6 +59,6 @@ public class JobConfigConstants {
   public static final String ENABLE_RESIZING = "enable.resizing";
 
   // This setting should be used if you will generate less # of segments after
-  // push. In preprocessing, this is likely because we resize segments. 
-  public static final String IS_DELETE_EXTRA_SEGMENTS = "is.delete.extra.segments";
+  // push. In preprocessing, this is likely because we resize segments.
+  public static final String DELETE_EXTRA_SEGMENTS = "delete.extra.segments";
 }
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
@@ -19,6 +19,7 @@
 package org.apache.pinot.hadoop.job;
 
 import com.google.common.base.Preconditions;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
@@ -44,7 +45,7 @@ public class SegmentTarPushJob extends BaseSegmentJob {
     int port = Integer.parseInt(properties.getProperty(JobConfigConstants.PUSH_TO_PORT));
     _pushLocations = PushLocation.getPushLocations(hosts, port);
     _rawTableName = Preconditions.checkNotNull(_properties.getProperty(JobConfigConstants.SEGMENT_TABLE_NAME));
-    _deleteExtraSegments = Boolean.parseBoolean(properties.getProperty(JobConfigConstants.DELETE_EXTRA_REFRESHED_SEGMENTS, "false"));
+    _deleteExtraSegments = Boolean.parseBoolean(properties.getProperty(JobConfigConstants.IS_DELETE_EXTRA_SEGMENTS, "false"));
   }
 
   @Override
@@ -58,28 +59,7 @@ public class SegmentTarPushJob extends BaseSegmentJob {
     try (ControllerRestApi controllerRestApi = getControllerRestApi()) {
       // TODO: Deal with invalid prefixes in the future
       if (_deleteExtraSegments) {
-        List<String> allSegments = controllerRestApi.getAllSegments("OFFLINE");
-        Set<String> uniqueSegmentPrefixes = new HashSet<>();
-
-        // Get all relevant segment prefixes that we are planning on pushing
-        List<Path> segmentsToPushPaths = getDataFilePaths(_segmentPattern);
-        List<String> segmentsToPushNames = segmentsToPushPaths.stream().map(s -> s.getName()).collect(Collectors.toList());
-        for (String segmentName : segmentsToPushNames) {
-          String segmentNamePrefix = removeSequenceId(segmentName);
-          uniqueSegmentPrefixes.add(segmentNamePrefix);
-        }
-
-        List<String> relevantSegments = new ArrayList<>();
-        // Get relevant segments already pushed that we are planning on refreshing
-        for (String segmentName : allSegments) {
-          if (uniqueSegmentPrefixes.contains(removeSequenceId(segmentName))) {
-            relevantSegments.add(segmentName);
-          }
-        }
-
-        relevantSegments.removeAll(segmentsToPushNames);
-        controllerRestApi.pushSegments(fileSystem, getDataFilePaths(_segmentPattern));
-        controllerRestApi.deleteSegmentUris(relevantSegments);
+        deleteExtraSegments(controllerRestApi, fileSystem);
       } else {
         controllerRestApi.pushSegments(fileSystem, getDataFilePaths(_segmentPattern));
       }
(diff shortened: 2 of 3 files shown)


pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
```diff
@@ -48,10 +54,45 @@ public void run()
       throws Exception {
     FileSystem fileSystem = FileSystem.get(_conf);
     try (ControllerRestApi controllerRestApi = getControllerRestApi()) {
-      controllerRestApi.pushSegments(fileSystem, getDataFilePaths(_segmentPattern));
+      // TODO: Deal with invalid prefixes in the future
```
snleee: I'm not sure if segment name based solution is the correct approach for deleting segments pushed together. The logic will break for custom segment name builder. Also, it will break in the case where people are pushing to the table from multiple push jobs.

One approach that I can think of is to mark some id (e.g. batchId) to segments from offline push job and delete segments with the same batchId.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
```diff
@@ -57,4 +57,6 @@
   public static final String ENABLE_PARTITIONING = "enable.partitioning";
   public static final String ENABLE_SORTING = "enable.sorting";
   public static final String ENABLE_RESIZING = "enable.resizing";
+
+  public static final String DELETE_EXTRA_REFRESHED_SEGMENTS = "delete.extra.refreshed.segments";
```
snleee: Can we rename this if you have a better name?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
```diff
@@ -48,10 +54,45 @@ public void run()
...
     }
   }
 
+  /**
+   * Remove trailing sequence id
+   * @param segmentName
+   * @return
+   */
+  private String removeSequenceId(String segmentName) {
+    return segmentName.replaceAll("\\d*$", "");
```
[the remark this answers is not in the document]
jenniferdai: in the javadoc above i have the functionality but i will add an additional example

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
jenniferdai: I'm not plannign on deleting segments pushed together. I am deleting extra segments. For example,

If table_day1_0
table_day1_1
table_day1_2 are pushed today.

Table_day1_0
table_day1_1 are pushed tomorrow,

I will delete table_day1_2

siddharthteotia: I am a little confused here -- the PR description says that you are deleting segments that are pushed or refreshed within a time range.  What exactly is an extra segment?

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
```diff
@@ -0,0 +1,95 @@
...
+    Assert.assertEquals(segmentsToDelete.size(), 2);
+    Assert.assertEquals(segmentsToDelete.contains("mytable_2018-09-10_2018-09-10_0"), false);
+  }
+
+  @Test
+  public void checkDeleteWithRefresh() {
+    List<String> allSegments = new ArrayList<>();
+    allSegments.add("mytable_0");
+    allSegments.add("mytable_1");
+    allSegments.add("mytable_2");
```
siddharthteotia: how do we decide which one is an extra segment?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
```diff
@@ -47,9 +54,57 @@ protected boolean isDataFile(String fileName) {
...
+
+    List<String> relevantSegments = new ArrayList<>();
+    // Get relevant segments already pushed that we are planning on refreshing
+    for (String segmentName : allSegments) {
+      if (uniqueSegmentPrefixes.contains(removeSequenceId(segmentName))) {
+        relevantSegments.add(segmentName);
+      }
     }
+
+    relevantSegments.removeAll(segmentsToPushNames);
```
siddharthteotia: This will be a O(N^2) operation and potentially more heap heavy because of constant over-writing.
I think we should use a hashset (or any alternative) to make this constant time operation

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
jenniferdai: discussed offline

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    List<Path> currentSegments = new ArrayList<>();
+    currentSegments.add(new Path("mytable_0"));
+    SegmentTarPushJob segmentTarPushJob = new SegmentTarPushJob(_defaultProperties);
+    List<String> segmentsToDelete = segmentTarPushJob.getSegmentsToDelete(allSegments, currentSegments);
+    Assert.assertEquals(segmentsToDelete.size(), 2);
+    Assert.assertEquals(segmentsToDelete.contains("mytable_0"), false);
```
siddharthteotia: Using contains() on an arraylist is generally not recommended and not intuitive since it is essentially a linear scan for an operation that should be amortized constant. May be we should change the return type of getSegmentsToDelete() to Set<String> or build a set out in here in the test code before asserting

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
```diff
@@ -47,9 +54,57 @@ protected boolean isDataFile(String fileName) {
...
+    Set<String> uniqueSegmentPrefixes = new HashSet<>();
+
+    // Get all relevant segment prefixes that we are planning on pushing
+    List<String> segmentsToPushNames = segmentsToPush.stream().map(s -> s.getName()).collect(Collectors.toList());
+    for (String segmentName : segmentsToPushNames) {
+      String segmentNamePrefix = removeSequenceId(segmentName);
+      uniqueSegmentPrefixes.add(segmentNamePrefix);
+    }
+
+    List<String> relevantSegments = new ArrayList<>();
```
siddharthteotia: It would be great to complement this code with an example (in the javadoc or near the code) to walk through what list of segments we start with, how we filter them and got to a list of segments that we return (the ones to delete)

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
```diff
@@ -57,4 +57,6 @@
   public static final String ENABLE_PARTITIONING = "enable.partitioning";
   public static final String ENABLE_SORTING = "enable.sorting";
   public static final String ENABLE_RESIZING = "enable.resizing";
+
+  public static final String IS_DELETE_EXTRA_SEGMENTS = "is.delete.extra.segments";
```
jackjlli: It'd be good to add some descriptions to this setting on how/when to use it.

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
```diff
@@ -0,0 +1,95 @@
...
+  @BeforeClass
+  private void setup() {
+    _defaultProperties.setProperty(JobConfigConstants.PUSH_TO_HOSTS, "sample_host");
+    _defaultProperties.setProperty(JobConfigConstants.PUSH_TO_PORT, "1234");
+    _defaultProperties.setProperty(JobConfigConstants.PATH_TO_OUTPUT, "sample_output_path");
+  }
+
+  @Test
+  public void checkDelete() {
+    List<String> allSegments = new ArrayList<>();
```
jackjlli: rename the variable to sth like `allSegmentsInCluster` or sth like that?

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
```diff
@@ -0,0 +1,95 @@
...
+    allSegments.add("mytable_2018-09-10_2018-09-10_0");
+    allSegments.add("mytable_2018-09-10_2018-09-10_1");
+    allSegments.add("mytable_2018-09-10_2018-09-10_2");
+
+    List<Path> currentSegments = new ArrayList<>();
+    currentSegments.add(new Path("mytable_2018-09-10_2018-09-10_0"));
+    SegmentTarPushJob segmentTarPushJob = new SegmentTarPushJob(_defaultProperties);
+    List<String> segmentsToDelete = segmentTarPushJob.getSegmentsToDelete(allSegments, currentSegments);
+    Assert.assertEquals(segmentsToDelete.size(), 2);
+    Assert.assertEquals(segmentsToDelete.contains("mytable_2018-09-10_2018-09-10_0"), false);
```
jackjlli: Use `assertFalse()` instead of `assertEquals()`

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
jackjlli: Same here.

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
```diff
@@ -0,0 +1,95 @@
...
+    allSegments.add("mytable_02");
+    allSegments.add("mytable_12");
+    allSegments.add("mytable_23");
+
+    List<Path> currentSegments = new ArrayList<>();
+    currentSegments.add(new Path("mytable_02"));
+    SegmentTarPushJob segmentTarPushJob = new SegmentTarPushJob(_defaultProperties);
+    List<String> segmentsToDelete = segmentTarPushJob.getSegmentsToDelete(allSegments, currentSegments);
+    Assert.assertEquals(segmentsToDelete.size(), 2);
+    Assert.assertEquals(segmentsToDelete.contains("mytable_02"), false);
```
jackjlli: Same here.

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @AfterClass
+  private void shutdown() {
```
jackjlli: No need to keep this `shutdown` method?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
```diff
@@ -48,10 +54,45 @@ public void run()
...
     FileSystem fileSystem = FileSystem.get(_conf);
     try (ControllerRestApi controllerRestApi = getControllerRestApi()) {
-      controllerRestApi.pushSegments(fileSystem, getDataFilePaths(_segmentPattern));
+      // TODO: Deal with invalid prefixes in the future
+      if (_deleteExtraSegments) {
+        List<String> allSegments = controllerRestApi.getAllSegments("OFFLINE");
+        Set<String> uniqueSegmentPrefixes = new HashSet<>();
+
+        // Get all relevant segment prefixes that we are planning on pushing
+        List<Path> segmentsToPushPaths = getDataFilePaths(_segmentPattern);
```
snleee: `segmentPathsToPush`

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
jenniferdai: discussed and resolved this

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentTarPushJob.java
```diff
@@ -39,6 +45,7 @@ public SegmentTarPushJob(Properties properties) {
     int port = Integer.parseInt(properties.getProperty(JobConfigConstants.PUSH_TO_PORT));
     _pushLocations = PushLocation.getPushLocations(hosts, port);
     _rawTableName = Preconditions.checkNotNull(_properties.getProperty(JobConfigConstants.SEGMENT_TABLE_NAME));
+    _deleteExtraSegments = Boolean.parseBoolean(properties.getProperty(JobConfigConstants.IS_DELETE_EXTRA_SEGMENTS, "false"));
```
snleee: `DELETE_EXTRA_SEGMENTS` seems to be good enough :)

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/JobConfigConstants.java
snleee: `delete.extra.segments`

pinot-hadoop/src/test/java/org/apache/pinot/hadoop/io/DeleteExtraPushedSegmentsTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.pinot.hadoop.job.SegmentTarPushJob;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+/**
+ * Tests logic to delete extra segments within the same time unit for APPEND or extra segments during REFRESH cases.
+ */
+public class DeleteExtraPushedSegmentsTest {
```
snleee: Can you add the case where there's no sequence id? I think it may result in the wrong behavior.

In `NormalizedDateSegmentNameGenerator`, there is an explicit config `excludeSequenceId`. So, segment name with no sequence id is a valid case and we cover that case correctly.

jenniferdai: added a test to check, the code handles it correctly

jenniferdai:
Your comment seems to have disappeared. What would you like me to remove?

As the preconditions was added in your commit for multiple push and is tied with your feature, I think you probably have a better idea regarding all the edge cases you require this preconditions check, etc. It may be better for you to call out cases where that preconditions will be required as you merge subsequent prs for your feature.

PR #4466, opened 2019-07-23
[TE] Adding Mean Variance Rule Detector

HoraceChoi95:
Forecast using history mean and standard deviation.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+import org.joda.time.Interval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import static org.apache.pinot.thirdeye.dataframe.DoubleSeries.*;
+import static org.apache.pinot.thirdeye.dataframe.util.DataFrameUtils.*;
+
+@Components(title = "WoW mean and std-dev forecasting and detection",
+    type = "WOW_STD_RULE",
+    tags = {DetectionTag.RULE_DETECTION},
+    description = "Forecast using history WoW mean and std-dev",
```
xiaohui-sun: "Forecast using history mean and standard deviation."

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+import org.apache.pinot.thirdeye.detection.spi.model.TimeSeries;
+import org.apache.pinot.thirdeye.rootcause.impl.MetricEntity;
+import org.joda.time.DateTime;
+import org.joda.time.Interval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import static org.apache.pinot.thirdeye.dataframe.DoubleSeries.*;
+import static org.apache.pinot.thirdeye.dataframe.util.DataFrameUtils.*;
+
+@Components(title = "WoW mean and std-dev forecasting and detection",
```
xiaohui-sun: History mean and standard deviation based forecasting and detection.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+    type = "WOW_STD_RULE",
+    tags = {DetectionTag.RULE_DETECTION},
+    description = "Forecast using history WoW mean and std-dev",
+    params = {
+        @Param(name = "offset", defaultValue = "wo1w"),
+        @Param(name = "pattern", allowableValues = {"up", "down"}),
+        @Param(name = "lookback"),
+        @Param(name = "sensitivity")})
+public class WoWStdRuleDetector implements AnomalyDetector<WoWStdRuleDetectorSpec>,
```
xiaohui-sun: Suggest to change name to MeanVarianceRuleDetector.
It is not necessary "wow".  We can support "mom" .

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+import static org.apache.pinot.thirdeye.dataframe.util.DataFrameUtils.*;
+
+@Components(title = "WoW mean and std-dev forecasting and detection",
+    type = "WOW_STD_RULE",
+    tags = {DetectionTag.RULE_DETECTION},
+    description = "Forecast using history WoW mean and std-dev",
+    params = {
+        @Param(name = "offset", defaultValue = "wo1w"),
+        @Param(name = "pattern", allowableValues = {"up", "down"}),
+        @Param(name = "lookback"),
```
xiaohui-sun: What is the default value for "lookback" and "sensitivity"?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+
+    // Filter pattern
+    if (pattern.equals(Pattern.UP_OR_DOWN) ) {
+      df.addSeries(COL_PATTERN, BooleanSeries.fillValues(df.size(), true));
+    } else {
+      df.addSeries(COL_PATTERN, pattern.equals(Pattern.UP) ? df.getDoubles(COL_DIFF).gt(0) :
+          df.getDoubles(COL_DIFF).lt(0));
+    }
+    df.addSeries(COL_DIFF_VIOLATION, df.getDoubles(COL_DIFF).abs().gte(df.getDoubles(COL_ERROR)));
+//    df.addSeries(COL_DIFF_VIOLATION, map((Series.BooleanFunction) values -> values[0] || values[1],
```
xiaohui-sun: Remove unused comments.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/WoWStdRuleDetectorSpec.java
```diff
@@ -0,0 +1,60 @@
...
+ * under the License.
+ */
+
+package org.apache.pinot.thirdeye.detection.spec;
+
+import org.apache.pinot.thirdeye.dataframe.util.MetricSlice;
+import org.apache.pinot.thirdeye.detection.Pattern;
+
+
+public class WoWStdRuleDetectorSpec extends AbstractSpec {
```
xiaohui-sun: Add some comments to the public class.
Again suggest to use MeanVarianceRuleDetectorSpec to make it more generic.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/WoWStdRuleDetectorSpec.java
```diff
@@ -0,0 +1,60 @@
...
+  private String monitoringGranularity = MetricSlice.NATIVE_GRANULARITY.toAggregationGranularityString(); // use native granularity by default
+  private int lookback = 52;
+  private double sensitivity = 1.5;
```
xiaohui-sun: We should make the sensitivity from 0-10. Higher value means more sensitive.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+    return data.getTimeseries().get(sliceData);
+  }
+
+  /**
+   * Returns a data frame containing lookback number of data before prediction time
+   * @param originalDF the original dataframe
+   * @param time the prediction time, in unix timestamp
+   * @return DataFrame containing lookback number of data
+   */
+
```
xiaohui-sun: Remove extra space.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+  private DataFrame getLookbackDF(DataFrame originalDF, Long time) {
+    LongSeries longSeries = (LongSeries) originalDF.get(COL_TIME);
+    int indexFinish = longSeries.find(time);
```
xiaohui-sun: Call it "indexEnd"?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+    //get the trainingDF for each week, which is the number of lookback to 1 week before the each week predict start time
+    for (int k = 0; k < size; k++) {
+      DataFrame trainingDF;
+      trainingDF = getLookbackDF(inputDF, forecastDF.getLong(COL_TIME, k));
+      //the get historical WoW mean and std.
+      std[k]= trainingDF.getDoubles(COL_CHANGE).std().value();
+      mean[k] = trainingDF.getDoubles(COL_CHANGE).mean().value();
+
+      // We need at least 4 weeks of data
+      if (trainingDF.size() < 4) {
```
xiaohui-sun: When will we have less than 4 weeks? Should validate the configuration in ahead instead of silently drop here.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+
+    if (indexFinish != -1) {
+      int indexStart = Math.max(0, indexFinish - lookback);
+      df = df.append(originalDF.slice(indexStart, indexFinish));
+    }
+    // calculate percentage change
+    df.addSeries(COL_CURR,df.getDoubles(COL_VALUE).shift(-1));
+    df.addSeries(COL_CHANGE, map((Series.DoubleFunction) values -> {
+      if (Double.compare(values[1], 0.0) == 0) {
+        return Double.compare(values[0], 0.0) == 0 ? 0.0 : (values[0] > 0 ? Double.POSITIVE_INFINITY : Double.NEGATIVE_INFINITY);
```
xiaohui-sun: Return INFINITY will skew the result. Shall we just return 0?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
HoraceChoi95: The default lookback is 52 weeks and the default sensitivity is 1.5, which means we are using 3𝜎. It's set in the sepc file.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+   * @return Data Frame that has data from start to end
+   */
+  private DataFrame fetchData(MetricEntity metricEntity, long start, long end) {
+    List<MetricSlice> slices = new ArrayList<>();
+    MetricSlice sliceData = MetricSlice.from(metricEntity.getId(), start, end,
+        metricEntity.getFilters(), timeGranularity);
+    slices.add(sliceData);
+    LOG.info("Getting data for" + sliceData.toString());
+    InputData data = this.dataFetcher.fetchData(new InputDataSpec().withTimeseriesSlices(slices)
+        .withMetricIdsForDataset(Collections.singletonList(metricEntity.getId())));
```
jihaozh: Can remove the clause `withMetricIdsForDataset` because the dataset is never fetched from the `InputData`

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/WoWStdRuleDetectorSpec.java
HoraceChoi95: I think the 0-10 scale makes it very hard to map how many total sigmas we are trying to use or sigmas on each side. Typically the "sensitivity" here should be between 0.5 and 1.5 since the number is for sigmas on each side. The interval length is matching to 0-10, but then it would have a base value of 0.5, which is a little confusing. Maybe changing the name to "sigmas" instead of "sensitivity" to avoid confusion?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+import org.apache.pinot.thirdeye.rootcause.impl.MetricEntity;
+import org.joda.time.DateTime;
+import org.joda.time.Interval;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import static org.apache.pinot.thirdeye.dataframe.DoubleSeries.*;
+import static org.apache.pinot.thirdeye.dataframe.util.DataFrameUtils.*;
+
+@Components(title = "WoW mean and std-dev forecasting and detection",
+    type = "WOW_STD_RULE",
```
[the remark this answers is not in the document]
HoraceChoi95: Thanks for the name suggestion!

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
```diff
@@ -0,0 +1,275 @@
...
+    LongSeries longSeries = (LongSeries) originalDF.get(COL_TIME);
+    int indexFinish = longSeries.find(time);
+    DataFrame df = DataFrame.builder(COL_TIME, COL_VALUE).build();
+
+    if (indexFinish != -1) {
+      int indexStart = Math.max(0, indexFinish - lookback);
+      df = df.append(originalDF.slice(indexStart, indexFinish));
+    }
+    // calculate percentage change
+    df.addSeries(COL_CURR,df.getDoubles(COL_VALUE).shift(-1));
```
jihaozh: The `originalDF` might not be weekly granularity. Looks like it's set by the `monitoringGranularity`. Is this calculating WoW percentage change? Shifting by 1 might not get the correct result.

If we are always expecting the `originalDF` to be `7_DAYS` granularity, maybe we shouldn't let it be configurable in the specs.

thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetectorTest.java
```diff
@@ -0,0 +1,169 @@
...
+    spec.setLookback(52);
+    spec.setSensitivity(1.0);
+    detector.init(spec, new DefaultInputDataFetcher(this.provider, -1));
+    Interval window = new Interval(1530576000000L, 1562630400000L);
+    String metricUrn = "thirdeye:metric:1";
+    MetricEntity me = MetricEntity.fromURN(metricUrn);
+    MetricSlice slice = MetricSlice.from(me.getId(), window.getStartMillis(), window.getEndMillis(), me.getFilters(), TimeGranularity
+        .fromString(spec.getMonitoringGranularity()));
+    TimeSeries timeSeries = detector.computePredictedTimeSeries(slice);
+    Assert.assertEquals(timeSeries.getPredictedBaseline().size(), 53);
```
jihaozh: Maybe add an assertion for the predicted values?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/WoWStdRuleDetectorSpec.java
HoraceChoi95: I added some comments and change the name, but not sure what kind of comments should be here. Let me know if it's OK.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
HoraceChoi95: Thanks, I did not spot that cos' there's no divide by zero scenario in the test.

snleee:
Can you add `[TE]` to your git commit message so that we can differentiate commits for Thirdeye from Pinot.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
HoraceChoi95: I added spec validation part. Thanks!

HoraceChoi95:
> Can you add `[TE]` to your git commit message so that we can differentiate commits for Thirdeye from Pinot.

Sorry, I pushed one another commit without it. Will add prefix tag for future ones.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/WoWStdRuleDetector.java
HoraceChoi95: The `originalDF` could be weekly, monthly, quarterly(given enough data) or any period set by the user. I think it does not make sense for this detection component to look at DoD change if there is seasonality. However, if the metric is super stable, it can also use DoD change to predict too.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/WoWStdRuleDetectorSpec.java
HoraceChoi95: I change it to sigma and added validation part to make sure sigma should be between 0-1.5 Thanks!

HoraceChoi95:
> Nice work @HoraceChoi95.
> 
> Can you include the test csv file in the review? Looks like the tests failed in Travis.

@akshayrai I included it in the review. But it's still failing.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/WoWStdRuleDetectorSpec.java
xiaohui-sun: I think you can keep sigma. However the motivation of introducing 0-10 scale is to give a universal parameter setting across all anomaly detection functions. There are several benefits:
1. Easy to understand. Users don't need to understand the methodology behind it. It could be p-value or sigma or others.
2. Easy to switch between rules w/o changing parameters. E.g, user can set a sensitivity of 5 to expect a moderate sensitivity for all rules.
I strongly encourage you to generate a mapping between sensitivity and sigma and expose sensitivity to user. If you want to expose sigma for users to fine control the model (like p-value) it is fine.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MeanVarianceRuleDetector.java
```diff
@@ -0,0 +1,277 @@
...
+    LongSeries longSeries = (LongSeries) originalDF.get(COL_TIME);
+    int indexEnd = longSeries.find(time);
+    DataFrame df = DataFrame.builder(COL_TIME, COL_VALUE).build();
+
+    if (indexEnd != -1) {
+      int indexStart = Math.max(0, indexEnd - lookback);
+      df = df.append(originalDF.slice(indexStart, indexEnd));
+    }
+    // calculate percentage change
+    df.addSeries(COL_CURR,df.getDoubles(COL_VALUE).shift(-1));
```
xiaohui-sun: Just curious, why you want to calculate the percentage change, and then use the mean/std from the percentage change, instead of directly using the values?
Directly use the values can give you one more data point.

Also have you considered removing outliers? If there is one anomaly happens and this can affect the upper/lower boundaries for multiple weeks after it.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spec/WoWStdRuleDetectorSpec.java
HoraceChoi95: Thanks for the suggestion. I changed the spec and detection to use sensitivity.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/components/MeanVarianceRuleDetector.java
HoraceChoi95: For our use, we typically track metrics' WoW change and report that. It's what we actually tracking.

After discussion for the outlier removal, we decided to put it into phase 2. Phase one is just to have a detection method using the standard deviation that works for weekly.

PR #4469, opened 2019-07-25
Add crtl auto complete sql on query-console

sunny19930321:
Add use crtl keys auto complete sql, this PR use crtl keys  query result auto complete sql on query-console, like:
image


pinot-controller/src/main/resources/static/js/lib/codemirror/hint/sql-hint.js
```diff
@@ -0,0 +1,322 @@
+/*
```
snleee: remove apache header

sunny209868: Ok, I don't see through because there is no checks apache header, so direct copy apache header from the controller/src/main/resources/static/css/lib/codemirror-hint.css and controller/src/main/resources/static/js/init.js. Now, I remove apache header

npawar:

PR #4470, opened 2019-07-25
[TE] Include admin recipients in every email alert (cc address)

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/scheme/DetectionEmailAlerter.java
```diff
@@ -198,15 +202,15 @@ private void sendEmail(DetectionAlertFilterRecipients recipients, Set<MergedAnom
...
+    email.setSubject(emailEntity.getSubject());
     email.setFrom(this.config.getFrom());
     email.setTo(AlertUtils.toAddress(recipients.getTo()));
-    email.setSubject(emailEntity.getSubject());
-    if (CollectionUtils.isNotEmpty(recipients.getCc())) {
-      email.setCc(AlertUtils.toAddress(recipients.getCc()));
-    }
-    if (CollectionUtils.isNotEmpty(recipients.getBcc())) {
-      email.setBcc(AlertUtils.toAddress(recipients.getBcc()));
+    if (recipients.getCc() == null) {
```
xiaohui-sun: if Cc() is mandatory then it should get initialized as empty set in constructor.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/scheme/DetectionEmailAlerter.java
```diff
@@ -62,7 +63,10 @@
 
   private static final Comparator<AnomalyResult> COMPARATOR_DESC =
       (o1, o2) -> -1 * Long.compare(o1.getStartTime(), o2.getStartTime());
-  private static final String EMAIL_WHITELIST_KEY = "emailWhitelist";
+
+  private static final String PROP_EMAIL_WHITELIST = "emailWhitelist";
+  private static final String PROP_ADMIN_RECIPIENTS = "adminRecipients";
```
xiaohui-sun: Who are admin recipients?
Is it per subscription group or per application?
What is the different between "adminReceipients" and "cc" list?

akshayrai: Well, adminRecipients are not tied with either an application or subscription group. By admin, I am referring to the platform admins (us). We need to be part of all the alerts that are sent out.

For example, we include thirdeye-dev alias in cc for all the email alerts we send out. This will give us insights into which alerts are very actively monitored (email threads), which conversations ended up in critical GCNs and we can also help them with RCA.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/scheme/DetectionEmailAlerter.java
akshayrai: It is not mandatory but I think it makes sense to initialize them in the constructor itself.

akshayrai: On second thought, I'd rather keep it this way. cc and bcc are optional fields.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/scheme/DetectionEmailAlerter.java
xiaohui-sun: Sure that totally makes sense. Then is it configured in Yaml?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/scheme/DetectionEmailAlerter.java
xiaohui-sun: Makes sense.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/scheme/DetectionEmailAlerter.java
akshayrai: Yes, this will be configured in the ThirdEye application config - detector.yml

PR #4471, opened 2019-07-25
[TE] Include alert description in entity email alert

akshayrai:


xiaohui-sun:
LGTM. I remember there is another requirement that want to show rules that are violated in alert mail. We can think of how to include this information in anomaly and then show it in mail. @jihaozh  @akshayrai

jihaozh:
> LGTM. I remember there is another requirement that want to show rules that are violated in alert mail. We can think of how to include this information in anomaly and then show it in mail. @jihaozh @akshayrai

I think it's stored in the anomaly already. We can think of how to surface it in UI/Mail.

PR #4472, opened 2019-07-26
Use hit counter to track max QPS per minute for broker

siddharthteotia:
Using the hit counter in a configured manner (600 buckets of 100ms each to cover a time range of 60sec) to keep track of max QPS per minute for broker. 
Detailed explanation of algorithm is there is in HitCounter.java

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
@@ -24,8 +24,7 @@ import java.util.concurrent.atomic.AtomicLongArray;
 
 /**
  * This hit counter is for counting the number of hits within a range of time.
- * Right now the granularity we use is configured the users. Currently two users
- * are there
+ * Currently two callers are there
  *
  * (1) QueryQuota Manager {@link HelixExternalViewBasedQueryQuotaManager} which
  * uses the hit counter for 1sec range of time and 100 time buckets thus
@@ -100,7 +99,7 @@ public class HitCounter {
 
   /*
    * Explanation of algorithm to keep track of
-   * max QPS per minute:
+   * max QPS within a minute window
    *
    * Hit counter is configured with fixed number buckets
    * and each bucket covers a fixed time window to
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
@@ -22,23 +22,36 @@ import com.google.common.annotations.VisibleForTesting;
 import java.util.concurrent.atomic.AtomicIntegerArray;
 import java.util.concurrent.atomic.AtomicLongArray;
 
-
 /**
- * This hit counter is for counting the number of hits within a range of time. Right now the granularity we use is second.
- * In order to save the space and time, we store the number of hits over the last 100 time buckets. When the method hit
- * gets called, we put the timestamp to the specified bucket. When the method getHitCount gets called, we sum all the number
- * of hits within the last 100 time buckets.
+ * This hit counter is for counting the number of hits within a range of time.
+ * Right now the granularity we use is configured the users. Currently two users
+ * are there
+ *
+ * (1) QueryQuota Manager {@link HelixExternalViewBasedQueryQuotaManager} which
+ * uses the hit counter for 1sec range of time and 100 time buckets thus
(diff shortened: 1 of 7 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
```diff
@@ -77,11 +90,154 @@ public int getHitCount() {
...
+
+  /*
+   * Explanation of algorithm to keep track of
+   * max QPS per minute:
+   *
+   * Hit counter is configured with fixed number buckets
+   * and each bucket covers a fixed time window to
+   * cover an overall range of 60secs.
+   *
+   * Each bucket stores two values -- start time
```
mcvsubbu: it seems that we only need to store the start time for the first bucket, right? the others can be derived. Not sure if we implement it that way, just making a note here.

pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
```diff
@@ -77,11 +90,154 @@ public int getHitCount() {
...
+   *
+   * See the unit test for peakQPS in HitCounterTest
+   */
+
+  /**
+   * Currently invoked by {@link org.apache.pinot.broker.requesthandler.BrokerPeakQPSMetricsHandler}
+   * every time a query comes into {@link org.apache.pinot.broker.requesthandler.BaseBrokerRequestHandler}
+   */
+  public void hitAndUpdateLatestTime() {
+    hitAndUpdateLatestTime(System.currentTimeMillis());
```
mcvsubbu: If it helps, you can introduce a protected method called nowMillis() which returns current time in millis. You can then override this method to return whatever time you like in tests. Yet to read through the rest of the review, but just suggesting here. Not sure if it will help us get better test coverage.

pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
```diff
@@ -22,23 +22,36 @@
...
 import java.util.concurrent.atomic.AtomicLongArray;
 
-
 /**
- * This hit counter is for counting the number of hits within a range of time. Right now the granularity we use is second.
- * In order to save the space and time, we store the number of hits over the last 100 time buckets. When the method hit
- * gets called, we put the timestamp to the specified bucket. When the method getHitCount gets called, we sum all the number
- * of hits within the last 100 time buckets.
+ * This hit counter is for counting the number of hits within a range of time.
+ * Right now the granularity we use is configured the users. Currently two users
```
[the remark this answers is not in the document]
siddharthteotia: done

pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
```diff
@@ -77,11 +90,154 @@ public int getHitCount() {
...
+      if (numTimeUnits - _bucketStartTime.get(i) < _bucketCount) {
         count += _bucketHitCount.get(i);
       }
     }
     return count;
   }
+
+  /*
+   * Explanation of algorithm to keep track of
+   * max QPS per minute:
```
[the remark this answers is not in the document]
siddharthteotia: yep, corrected it

pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
siddharthteotia: I am not quite sure what that implies so we should discuss it. Let's take an example:

Start time of bucket b1 is 100 and bucket b2 is 101. Now if there are 6 buckets, then 106 and 107 will also map to b1 and b2 respectively. 

When there are only 2 hits recorded for 100 and 101 start times the latter can probably be derived from former. But let's say 3rd hit comes after a window of 1 min and let say start time is 107 then b2's start time will be set to 107 and this is something that can't be derived from b1's start time of 100 (although b1's hit count has now become irrelelevant as it is not out of 60sec window)

pinot-broker/src/main/java/org/apache/pinot/broker/queryquota/HitCounter.java
siddharthteotia: For coverage, I am explicitly calling the next method with a custom timestamp to test various values at different intervals.

siddharthteotia:
Addressed review comments

npawar:

PR #4473, opened 2019-07-26
Refactor HelixBrokerStarterTest

Jackie-Jiang:
- Use condition based wait instead of sleep to fix the flakiness of the test
- Fix the wrong assert which causes the test to last for more than 1 minute
- Use the TestUtils.waitForCondition so that the test can fail when condition cannot be met
- Reduce the default interval for TestUtils.waitForCondition to accelerate the test
- Remove some un-used class


pinot-broker/src/test/java/org/apache/pinot/broker/broker/HelixBrokerStarterTest.java
```diff
@@ -18,287 +18,197 @@
...
+        .addFakeBrokerInstancesToAutoJoinHelixCluster(getHelixClusterName(), ZkStarter.DEFAULT_ZK_STR, NUM_BROKERS - 1,
+            true);
     ControllerRequestBuilderUtil
-        .addFakeDataInstancesToAutoJoinHelixCluster(getHelixClusterName(), ZkStarter.DEFAULT_ZK_STR, 1, true);
-
-    while (_helixAdmin.getInstancesInClusterWithTag(getHelixClusterName(), "DefaultTenant_OFFLINE").size() == 0
-        || _helixAdmin.getInstancesInClusterWithTag(getHelixClusterName(), "DefaultTenant_BROKER").size() == 0) {
-      Thread.sleep(100);
-    }
+        .addFakeDataInstancesToAutoJoinHelixCluster(getHelixClusterName(), ZkStarter.DEFAULT_ZK_STR, NUM_SERVERS, true);
```
mcvsubbu: [since you are doing a bunch of cleanup] can you please change this to addFakeServerInstancesToHoinHelixCluster() ?

pinot-broker/src/test/java/org/apache/pinot/broker/broker/HelixBrokerStarterTest.java
```diff
@@ -18,287 +18,197 @@
...
+    _brokerStarter.start();
 
     ControllerRequestBuilderUtil
-        .addFakeBrokerInstancesToAutoJoinHelixCluster(getHelixClusterName(), ZkStarter.DEFAULT_ZK_STR, 5, true);
+        .addFakeBrokerInstancesToAutoJoinHelixCluster(getHelixClusterName(), ZkStarter.DEFAULT_ZK_STR, NUM_BROKERS - 1,
+            true);
     ControllerRequestBuilderUtil
-        .addFakeDataInstancesToAutoJoinHelixCluster(getHelixClusterName(), ZkStarter.DEFAULT_ZK_STR, 1, true);
-
-    while (_helixAdmin.getInstancesInClusterWithTag(getHelixClusterName(), "DefaultTenant_OFFLINE").size() == 0
```
mcvsubbu: Not sure why they had this wait. Does helix take time to recognize the instances added? Is it because of cached vs uncached references for instances in the getInstancesWithTag() call? Make sure that we don't need this wait.

pinot-broker/src/test/java/org/apache/pinot/broker/broker/HelixBrokerStarterTest.java
```diff
@@ -18,287 +18,197 @@
...
+    assertEquals(brokerResourceIdealState.getInstanceSet(newOfflineTableName).size(), NUM_BROKERS);
+
+    TestUtils.waitForCondition(aVoid -> {
+      Map<String, String> newTableStateMap =
+          _helixAdmin.getResourceExternalView(getHelixClusterName(), Helix.BROKER_RESOURCE_INSTANCE)
+              .getStateMap(newOfflineTableName);
+      return newTableStateMap != null && newTableStateMap.size() == NUM_BROKERS;
+    }, 30_000L, "Failed to find all brokers for the new table in the brokerResource ExternalView");
+
+    assertTrue(routing.routingTableExists(newOfflineTableName));
```
mcvsubbu: We seem to be testing for both realtime and offline segments to show up in the routing table,. Since the routing tables are different, isnt it better to have both in the tests? I think you have removed the realtime segment check?

pinot-core/src/test/java/org/apache/pinot/util/TestUtils.java
```diff
@@ -159,6 +97,6 @@ public static void waitForCondition(Function<Void, Boolean> condition, long chec
 
   public static void waitForCondition(Function<Void, Boolean> condition, long timeoutMs,
       @Nullable String errorMessage) {
-    waitForCondition(condition, 1000L, timeoutMs, errorMessage);
+    waitForCondition(condition, 100L, timeoutMs, errorMessage);
```
mcvsubbu: thanks for doing this

pinot-broker/src/test/java/org/apache/pinot/broker/broker/HelixBrokerStarterTest.java
Jackie-Jiang: This wait is not needed. There are several redundant wait in the old test

pinot-broker/src/test/java/org/apache/pinot/broker/broker/HelixBrokerStarterTest.java
Jackie-Jiang: That one is tested in line 142

pinot-broker/src/test/java/org/apache/pinot/broker/broker/HelixBrokerStarterTest.java
Jackie-Jiang: There are too many usages for this method, so I would do the name name in another pr.

PR #4475, opened 2019-07-28
Add sql +, -, *, / symbol arith

xueyumusic:
This PR adds sql +, -, *, / symbol arith based on recent new sql grammar support using calcite. Currently it look only support pql-like ADD, SUB, MULT, DIV function, it would be good to also support common used arith symbol.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/sql/parsers/CalciteSqlParser.java
+++ b/pinot-common/src/main/java/org/apache/pinot/sql/parsers/CalciteSqlParser.java
@@ -49,6 +49,17 @@ public class CalciteSqlParser {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(CalciteSqlParser.class);
 
+  private static final Map<String, String> FUNC_NAME_MAP;
+
+  // Calcite Sqlkind name to pinot operator name map
+  static {
+    FUNC_NAME_MAP = new HashMap<String, String>();
+    FUNC_NAME_MAP.put("PLUS", "ADD");
+    FUNC_NAME_MAP.put("MINUS", "SUB");
+    FUNC_NAME_MAP.put("TIMES", "MULT");
+    FUNC_NAME_MAP.put("DIVIDE", "DIV");
+  }
+
   /** Lexical policy similar to MySQL with ANSI_QUOTES option enabled. (To be
    * precise: MySQL on Windows; MySQL on Linux uses case-sensitive matching,
    * like the Linux file system.) The case of identifiers is preserved whether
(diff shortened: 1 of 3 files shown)


sunny209868:
> This PR adds sql +, -, *, / symbol arith based on recent new sql grammar support using calcite. Currently it look only support pql-like ADD, SUB, MULT, DIV function, it would be good to also support common used arith symbol.

Does this support addition, subtraction, multiplication and division of aggregate indices? like SUM(a) / SUM((b). @xueyumusic

xueyumusic:
Hi, @sunny19930321 . It looks this does not support `SUM(a) / SUM((b)` query. It maybe need more execute modification, thanks

sunny19930321:
> It maybe need more execute modification

I understand.

pinot-common/src/main/java/org/apache/pinot/sql/parsers/CalciteSqlParser.java
```diff
@@ -49,6 +49,17 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(CalciteSqlParser.class);
 
+  private static final Map<String, String> FUNC_NAME_MAP;
+
+  // Calcite Sqlkind name to pinot operator name map
+  static {
+    FUNC_NAME_MAP = new HashMap<String, String>();
```
fx19880617: Could you rename this to `CALCITE_TO_PINOT_FUNC_NAME_MAP`. Also it's better to make this an immutable map.

fx19880617:
@kishoreg do we support those arithmetic computations in query processing?

pinot-common/src/test/java/org/apache/pinot/sql/parsers/CalciteSqlCompilerTest.java
```diff
@@ -422,4 +422,36 @@ public void testSelectionTransformFunction() {
     Assert.assertEquals(
         pinotQuery.getFilterExpression().getFunctionCall().getOperands().get(1).getLiteral().getStringValue(), "v1");
   }
+
+  @Test
+  public void testSqlBaseArith() {
```
mayankshriv: Could you also add tests for nested expressions, for example: `(a * ((b  - c) / d) + e)`

xueyumusic: Thanks for review, @fx19880617 , @kishoreg , @mayankshriv , add a nested expression test case.

kishoreg:
@xueyumusic can we merge this?

fx19880617:
> @xueyumusic can we merge this?

this is already implemented in https://github.com/apache/incubator-pinot/pull/5018

PR #4476, opened 2019-07-29
Clean up test utils, disconnect fake instances

Jackie-Jiang:
- Remove redundant util classes
- Move controller test related methods into ControllerTest for easier usage
- Disconnect fake instances created in the test
- Update state model to include the missing transitions
- Fix the flakiness in ValidationManagerTest
- No change for the test logic


jackjlli:
Agree with @mcvsubbu that we should keep the method `getHelixClusterName()`.

Jackie-Jiang:
@mcvsubbu @jackjlli Done

PR #4479, opened 2019-07-30
[TE] Make the dashboard report summary to show the correct values

jihaozh:
This PR makes the dashboard report summary to show the correct values. 
- Fix the time range selector to return correct timestamps
- Make the report summary to show the full-time range and time zone information to avoid confusions.
- Fix the time range rounding/alignments.

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/home/share-dashboard/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/home/share-dashboard/route.js
@@ -19,6 +19,8 @@ const queryParamsConfig = {
   refreshModel: true
 };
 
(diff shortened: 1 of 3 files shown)


xiaohui-sun:
Thanks for fixing this? What is the root cause?

jihaozh:
> Thanks for fixing this. What is the root cause? Do we still face this problem if the data is late ?

If the data is late, the front end will show a dash, which means the data is not available, instead of showing a wrong number. 

This bug is caused by both frontend and backend. The frontend time range selector sometimes returns  misaligned time stamps. The time zone alignments in the endpoint also round the time stamp to the wrong direction.

PR #4482, opened 2019-07-30
[TE] frontend - harleyjj/full-ui - create alert form except rules component

harleyjj:
1) create-alert form and yaml editors share yaml state for both subscription and detection configs
2) consolidates yaml functions into util yaml-tools, including new redundantParse function for trying parse with jsyaml when yamljs fails
3) hides form unless debug=show param used 
4) sets ground work for dynamic component that renders based on yaml parsed to object


fx19880617:
Close this as thirdeye repo is moved to https://github.com/project-thirdeye/thirdeye

PR #4487, opened 2019-07-31
Fix the flakiness of PinotControllerModeTest by enabling the periodic rebalance

Jackie-Jiang:


Jackie-Jiang:
> So the periodic rebalance will be enabled in the controller of controllerTest instead of a separated helixOnlyController, right?

Periodic rebalance will be enabled when calling ControllerTest.startController() with Pinot-Ony or Dual mode.

jackjlli:
> > So the periodic rebalance will be enabled in the controller of controllerTest instead of a separated helixOnlyController, right?
> 
> Periodic rebalance will be enabled when calling ControllerTest.startController() with Pinot-Ony or Dual mode.

But in your PR you just pass the **helixOnly**ControllerConf to the test suite. How does that enable periodic rebalance?

Jackie-Jiang:
> > > So the periodic rebalance will be enabled in the controller of controllerTest instead of a separated helixOnlyController, right?
> > 
> > 
> > Periodic rebalance will be enabled when calling ControllerTest.startController() with Pinot-Ony or Dual mode.
> 
> But in your PR you just pass the **helixOnly**ControllerConf to the test suite. How does that enable periodic rebalance?

The first test does not require the periodic rebalance. I changed it just for consistency.

PR #4489, opened 2019-08-01
[Instance Assignment] Add configs for instance assignment

Jackie-Jiang:
- Also enhance JsonUtils to support TypeReference for deserialization

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
@@ -19,6 +19,7 @@
 package org.apache.pinot.common.config;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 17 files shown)


pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
```diff
@@ -82,6 +85,10 @@
   @NestedConfig
   private RoutingConfig _routingConfig;
 
+  @ConfigKey(INSTANCE_ASSIGNMENT_CONFIG_MAP_KEY)
+  @UseChildKeyHandler(InstanceAssignmentConfigMapChildKeyHandler.class)
+  private Map<InstancePartitionsType, InstanceAssignmentConfig> _instanceAssignmentConfigMap;
```
snleee: Map will be the best approach? We should consider other options to make this simpler. (e.g. put `instanceAssignmentConfigMap` + override config. `TenantConfig` is also using overriding config approach.

Jackie-Jiang: After consideration, I think explicitly putting the type (OFFLINE/CONSUMING/COMPLETED) is more clear than using the override. For realtime segments, if COMPLETED config does not exist, use CONSUMING config instead

snleee: In that case, do you think that we need to wrap those 3 config into some other class? It seems that adding 3 similar configs to the top level table config seems not too clean.

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
```diff
@@ -101,4 +102,18 @@ public static String getTenantNameFromTag(String tag) {
     }
     return tag;
   }
+
+  public static String getServerTagFromTableConfigAndInstancePartitionsType(TableConfig tableConfig,
+      InstancePartitionsType instancePartitionsType) {
+    switch (instancePartitionsType) {
+      case OFFLINE:
```
mcvsubbu: How does this relate to methods in RealtimeTagConfig class?

Jackie-Jiang: InstancePartitions is associated with server type (OFFLINE/CONSUMING/COMPLETED). Currently that is configured through the tenant config, and we need this method to extract the server tag from the tenant config.

PR #4490, opened 2019-08-02
Disable state transition stats verification in rebalancer cluster integration test

siddharthteotia:
Looks like state transition stats (from ONLINE to OFFLINE and OFFLINE to DROPPED) are potentially flakey as they are reliant on timing.

These stats are reliant on the fact that state transition has happened (callback invoked on the segment server) for the server that lost a segment as part of rebalancing.

While the rebalancer waits for all the ideal state changes to get reflected in external view, we don't yet check if the servers have indeed dropped the segments. So it is possible that server may not have done so by the time test finishes and thus we sometimes notice the segment dropped stats off by 1.

Disabling the stat verification to keep tests happy and until we make the external view check stronger by ensuring that it doesn't have any extra servers for a segment indicating that servers that were supposed to drop segments as part of rebalancing have truly done so. 

Will do this in a follow-up PR soon after but currently disabling to keep build clean.

### DIFF
--- a/pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
+++ b/pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
@@ -56,7 +56,7 @@ public class PinotControllerModeTest extends ControllerTest {
     stopController();
   }
 
-  @Test
+  @Test(enabled = false)
   public void testDualModeController() {
     // Start the first dual-mode controller
     ControllerConf firstDualModeControllerConfig = getDefaultControllerConfiguration();
--- a/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
+++ b/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
@@ -67,7 +67,6 @@ public class TableRebalancerAdminToolClusterIntegrationTest extends BaseClusterI
   private static final int NUM_INITIAL_SERVERS = 3;
   private SegmentStateTransitionStats _segmentStateTransitionStats;
   private boolean _raiseErrorOnSegmentStateTransition = false;
-  private final boolean _checkStateTransitionStats = false;
 
   // todo: re-use fake server start/stop methods from ControllerTest
   @BeforeClass
@@ -204,10 +203,9 @@ public class TableRebalancerAdminToolClusterIntegrationTest extends BaseClusterI
       // as part of rebalancing, host2 lost segment1 and host3
       // lost segment2 -- so 2 transitions from ON to OFF and
       // OFF to DROP
-      if (_checkStateTransitionStats) {
-        Assert.assertEquals(_segmentStateTransitionStats.offFromOn, 2);
-        Assert.assertEquals(_segmentStateTransitionStats.dropFromOff, 2);
-      }
+      // TODO: fix the flakey test behavior and re-enable these assertions
+      // Assert.assertEquals(_segmentStateTransitionStats.offFromOn, 2);
+      // Assert.assertEquals(_segmentStateTransitionStats.dropFromOff, 2);
     } finally {
       stopFakeServers();
     }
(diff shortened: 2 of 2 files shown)


siddharthteotia:
cc @jackjlli @mcvsubbu

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
```diff
@@ -67,6 +67,7 @@
   private static final int NUM_INITIAL_SERVERS = 3;
   private SegmentStateTransitionStats _segmentStateTransitionStats;
   private boolean _raiseErrorOnSegmentStateTransition = false;
+  private final boolean _checkStateTransitionStats = false;
```
Jackie-Jiang: Don't add this boolean, just comment out the lines and add TODO to add them back in the next pr

jackjlli:
Agree with @Jackie-Jiang that you can just comment out the assertions.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/TableRebalancerAdminToolClusterIntegrationTest.java
siddharthteotia: Done

siddharthteotia:
> Agree with @Jackie-Jiang that you can just comment out the assertions.

Done

pinot-common/src/main/java/org/apache/pinot/common/function/AggregationFunctionType.java
```diff
@@ -48,7 +48,8 @@
   DISTINCTCOUNTRAWHLLMV("distinctCountRawHLLMV"),
   PERCENTILEMV("percentileMV"),
   PERCENTILEESTMV("percentileEstMV"),
-  PERCENTILETDIGESTMV("percentileTDigestMV");
+  PERCENTILETDIGESTMV("percentileTDigestMV"),
+  DISTINCT("distinctValues");
```
Jackie-Jiang: Wrong merge?

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -82,8 +82,8 @@ public AggregationInfo buildAggregationInfo() {
       expression = "*";
     } else {
       List<? extends AstNode> children = getChildren();
-      if (children == null || children.size() != 1) {
-        throw new Pql2CompilationException("Aggregation function expects exact 1 argument");
+      if (children == null || (children.size() != 1 && !_name.equalsIgnoreCase("distinct"))) {
```
Jackie-Jiang: Same here?

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
```diff
@@ -56,7 +56,7 @@ public void testHelixOnlyController() {
     stopController();
   }
 
-  @Test
+  @Test(enabled = false)
```
Jackie-Jiang: Don't disable this test, we need to fix it

jackjlli: The test will be refactored in this PR:https://github.com/apache/incubator-pinot/pull/4323
We don't need to enable/disable resource from ideal state any more.
We'll re-enable the test in that PR.

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
siddharthteotia: sorry, it was bad merged. removed changes

pinot-common/src/main/java/org/apache/pinot/common/function/AggregationFunctionType.java
siddharthteotia: yes, removed changes. thanks for catching

siddharthteotia:
Got a clean build. Let's merge this.

PR #4494, opened 2019-08-04
Fix potential NPE from HelixDataAccessor

jackjlli:
When calling `getProperty` method from `HelixDataAccessor`, it could return null.
This PR handles potential NPE for that method in Pinot codebase.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
@@ -385,8 +385,12 @@ public class ServiceStatus {
     protected CurrentState getState(String resourceName) {
       PropertyKey.Builder keyBuilder = _helixDataAccessor.keyBuilder();
       LiveInstance liveInstance = _helixDataAccessor.getProperty(keyBuilder.liveInstance(_instanceName));
-      String sessionId = liveInstance.getSessionId();
-      return _helixDataAccessor.getProperty(keyBuilder.currentState(_instanceName, sessionId, resourceName));
+      if (liveInstance == null) {
+        return null;
+      } else {
+        String sessionId = liveInstance.getSessionId();
+        return _helixDataAccessor.getProperty(keyBuilder.currentState(_instanceName, sessionId, resourceName));
+      }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -223,17 +223,19 @@ public void cleanupLLC(final String realtimeTableName) {
...
-    IdealState idealState = HelixHelper.getTableIdealState(_helixManager, realtimeTableName);
     final List<String> segmentsToRemove = new ArrayList<String>();
-    Set<String> allSegments = idealState.getPartitionSet();
-    int removeCount = 0;
-    for (String segmentName : allSegments) {
-      if (SegmentName.isLowLevelConsumerSegmentName(segmentName)) {
-        segmentsToRemove.add(segmentName);
-        removeCount++;
+    IdealState idealState = HelixHelper.getTableIdealState(_helixManager, realtimeTableName);
+    if (idealState != null) {
```
chenboat: NIT for easier reading. 
       if (idealState == null) {
             return; 
        }
        .....

pinot-common/src/main/java/org/apache/pinot/common/utils/ServiceStatus.java
```diff
@@ -385,8 +385,12 @@ public IdealStateAndCurrentStateMatchServiceStatusCallback(HelixManager helixMan
     protected CurrentState getState(String resourceName) {
       PropertyKey.Builder keyBuilder = _helixDataAccessor.keyBuilder();
       LiveInstance liveInstance = _helixDataAccessor.getProperty(keyBuilder.liveInstance(_instanceName));
-      String sessionId = liveInstance.getSessionId();
-      return _helixDataAccessor.getProperty(keyBuilder.currentState(_instanceName, sessionId, resourceName));
+      if (liveInstance == null) {
```
siddharthteotia: Would it be appropriate here to raise error (like Preconditions.checkArgument) if it is null?

Jackie-Jiang:
@jackjlli We should drop this pr right?

jackjlli:
Closing this PR since only live instance can be null. Other znode should never be null.
PR to handle NPE from live instance: https://github.com/apache/incubator-pinot/pull/4528

PR #4495, opened 2019-08-05
[TE] [Composite-Alert] Adding Entity GroupKey with Whitelist Template and Formatter

akshayrai:
Idea: "Send alert on a top metric (whitelisted) along with anomalies in other affected modules."

Adding a composite alert template to display grouped anomalies(with groupKey) along with the anomalies from a whitelisted entity. For example: Send alert on the model performance metric along with anomalies co-occurring in the respective features.

* Extended the existing Composite alert template to support this by including a `entityWhitelist` property.
* Locally verified and tested along with unit test

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/alert/content/EntityGroupKeyContentFormatter.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/alert/content/EntityGroupKeyContentFormatter.java
@@ -23,6 +23,7 @@ import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -30,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import org.apache.pinot.thirdeye.anomaly.alert.util.EmailScreenshotHelper;
 import org.apache.pinot.thirdeye.anomalydetection.context.AnomalyResult;
 import org.apache.pinot.thirdeye.datalayer.bao.DetectionConfigManager;
 import org.apache.pinot.thirdeye.datalayer.dto.DetectionConfigDTO;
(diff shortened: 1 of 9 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/alert/content/EntityGroupKeyContentFormatter.java
```diff
@@ -48,32 +50,40 @@
...
+
+  static final String PROP_ENTITY_WHITELIST = "entityWhitelist";
   static final String PROP_ANOMALY_SCORE = "groupScore";
   static final String PROP_GROUP_KEY = "groupKey";
 
   private DetectionConfigManager configDAO = null;
   private Multimap<String, AnomalyReportEntity> entityToAnomaliesMap = ArrayListMultimap.create();
-  private Map<String, List<AnomalyReportEntity>> entityToSortedAnomaliesMap = new HashMap<>();
+  private Multimap<String, AnomalyReportEntity> entityToSortedAnomaliesMap = ArrayListMultimap.create();
+  private Multimap<String, AnomalyReportEntity> whitelistMetricToAnomaliesMap = ArrayListMultimap.create();
```
xiaohui-sun: Please add more comments what does "whitelistMetric" mean.

akshayrai: Added

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/alert/content/EntityGroupKeyContentFormatter.java
```diff
@@ -93,29 +110,51 @@ protected void updateTemplateDataByAnomalyResults(Map<String, Object> templateDa
     for (String entityName : entityToAnomaliesMap.keySet()) {
       List<AnomalyReportEntity> groupedAnomalies = (List<AnomalyReportEntity>) entityToAnomaliesMap.get(entityName);
       groupedAnomalies.sort((u1, u2) -> anomalyToGroupScoreMap.get(u2).compareTo(anomalyToGroupScoreMap.get(u1)));
-      entityToSortedAnomaliesMap.put(entityName, groupedAnomalies);
+      entityToSortedAnomaliesMap.putAll(entityName, groupedAnomalies);
     }
 
+    // Insert anomaly snapshot image
+    if (whitelistMetricToAnomaliesMap.size() == 1 && whitelistMetricToAnomaliesMap.values().size() == 1) {
```
jihaozh: If we only show it when the map size and value size is 1, can we use a Map instead of a multimap?

akshayrai: Actually, this condition here is to display the graph when we have only one anomaly. Otherwise, we will list all the anomalies. This map is passed below in line no. 127 which is then processed in the freemarker template file.

jihaozh: Make sense!

PR #4496, opened 2019-08-05
[TE] Remove deprecated stage classes and tests

akshayrai:


xiaohui-sun:
Thanks for cleaning it!

PR #4497, opened 2019-08-06
Add a thread in ControllerLeadershipManager to periodically fetch controller leadership info

Jackie-Jiang:
Add a thread in ControllerLeadershipManager to periodically fetch controller leadership info as a work-around of Helix callback delay
When the thread detects inconsistency between ControllerLeadershipManager and HelixManager, log a warning and emit metrics

The work-around is for the first issue described in: apache/helix#377

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import javax.annotation.concurrent.ThreadSafe;
 import org.apache.helix.HelixManager;
 import org.apache.pinot.common.metrics.ControllerGauge;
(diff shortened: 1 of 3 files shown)


mcvsubbu:
This is a work-around for the first problem described in https://github.com/apache/helix/issues/377

pinot-controller/src/main/java/org/apache/pinot/controller/ControllerLeadershipManager.java
```diff
@@ -60,17 +101,35 @@ public synchronized void subscribe(String name, LeadershipChangeSubscriber subsc
     }
   }
 
-  public boolean isLeader() {
+  public synchronized boolean isLeader() {
```
mcvsubbu: can remove synchronized here

Jackie-Jiang: Good catch

PR #4498, opened 2019-08-06
The external view check in rebalancer should also check for no extra servers

siddharthteotia:
The function to check if external view has converged now also ensures that servers that lost segments as part of rebalancing are not present in external view (for that segment).


siddharthteotia:
cc @mcvsubbu

siddharthteotia:
Although the external view stabilization check has been made stronger here in the code changes, I have removed the stat transition stats from the cluster integration test. This is actually baffling now -- I had even made them atomic so now there is no question of us reading stale value when checking for assertion. 

The fact that now the rebalancer ensures that for any segment, external view doesn't have any server that is not in ideal state should also ensure that segments have dropped since I believe the EV will not get updated until we get back from the callback.

siddharthteotia:
Filed https://issues.apache.org/jira/browse/HELIX-818 for a potential issue for why state transition callbacks are sometimes not invoked.

siddharthteotia:
Ping

PR #4500, opened 2019-08-06
Add file-evening to hadoop pre-process

jenniferdai:
* Currently, pre-process only allows users to set max.num.records. Now, what may happen is that we may have very skewed segment sizes if there is a tiny remainder in the segments. 
* This PR still guarantees max.num.records is the upper bound of record sin a segment but fixes the possible skew by calculating approximately-even segment record numbers prior to mandating the number of records that go into a reducer.

### DIFF
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
@@ -30,6 +30,7 @@ import java.util.zip.GZIPInputStream;
 import javax.annotation.Nullable;
 import org.apache.avro.Schema;
 import org.apache.avro.file.DataFileStream;
+import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericDatumReader;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.mapred.AvroKey;
@@ -177,7 +178,7 @@ public class SegmentPreprocessingJob extends BaseSegmentJob {
     }
 
     _fileSystem = FileSystem.get(_conf);
-    final List<Path> inputDataPath = getDataFilePaths(_inputSegmentDir);
+    final List<Path> inputDataPaths = getDataFilePaths(_inputSegmentDir);
 
     if (_fileSystem.exists(_preprocessedOutputDir)) {
       _logger.warn("Found the output folder {}, deleting it", _preprocessedOutputDir);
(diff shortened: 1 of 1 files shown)


pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -306,7 +309,15 @@ public void run()
       // so that all the rows can be spread evenly.
       addHashCodeField(fieldSet);
     }
-    setMaxNumRecordsConfigIfSpecified(job);
+    String enablePartitioningString = job.getConfiguration().get(JobConfigConstants.ENABLE_PARTITIONING);
```
Jackie-Jiang: boolean enablePartitioning = job.getConfiguration().getBoolean(JobConfigConstants.ENABLE_PARTITIONING, false);

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -412,17 +423,40 @@ private void fetchResizingConfig() {
     }
   }
 
-  private void setMaxNumRecordsConfigIfSpecified(Job job) {
+  // TODO: Handle case where #records exceed long max
```
Jackie-Jiang: Exceed long max? We don't need to consider Long overflow

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -446,6 +480,21 @@ private Schema getSchema(Path inputPathDir)
     return avroSchema;
   }
 
+  private long countAvroRecords(List<Path> inputPaths) throws IOException {
+    long count = 0;
+
+    for (Path path : inputPaths) {
+      DataFileStream<GenericRecord> dataFileStream = getAvroReader(path);
```
Jackie-Jiang: We need to scan all the files to get the record count? That will be super expensive

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: maybe.. for refresh use cases i coluld see it

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
jenniferdai: this seems to be the best approach from my research, do you have other suggestions?

jenniferdai:
others have benchmarked - MapReduce
When I use MapReduce job to iterate through all the records it completes in less than 2 minutes for about million records

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
Jackie-Jiang: That is more than the number of molecules on earth lol...

Jackie-Jiang:
> others have benchmarked - MapReduce
> When I use MapReduce job to iterate through all the records it completes in less than 2 minutes for about million records

So for 1 billion records, we need to wait for 2000 minutes (more than 1 day) just for the count?

PR #4502, opened 2019-08-06
Down-grade Helix to version 0.8.4

Jackie-Jiang:
We found some issues in Helix 0.9.0: apache/helix#377

NOTE: Helix version 0.8.4 has issues in Task Framework. Disabled the Minion tests for now. Don't use this version for Minion tasks.


mcvsubbu:
Does it make sense to remove the minion functionality by default ? (settable if needed)

Jackie-Jiang:
@mcvsubbu By default, there is no task scheduled, so minion functionality is not enabled

PR #4504, opened 2019-08-07
[Instance Assignment] Add implementations for instance assignment

Jackie-Jiang:
Generate InstancePartitions based on InstanceConfigs and InstanceAssignmentConfig
Generate default InstanceAssignmentConfig from TableConfig for backward-compatibility

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/instance/InstanceConstraintApplier.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/instance/InstanceConstraintApplier.java
@@ -1,0 +1,37 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.controller.helix.core.assignment.instance;
+
+import java.util.List;
+import java.util.Map;
+import org.apache.helix.model.InstanceConfig;
+
+
+/**
+ * The instance constraint applier is responsible for filtering out unqualified instances and sorts instances for
+ * picking priority.
+ */
+public interface InstanceConstraintApplier {
+
+  /**
+   * Applies the instance constraint to the given map from pool to instance configs to filter out unqualified instances
+   * and sort instances for picking priority.
+   */
+  Map<Integer, List<InstanceConfig>> applyConstraint(Map<Integer, List<InstanceConfig>> poolToInstanceConfigsMap);
+}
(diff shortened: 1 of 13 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/instance/InstanceReplicaPartitionSelector.java
```diff
@@ -0,0 +1,187 @@
...
+        int instanceIdInPool = 0;
+        for (int instanceIdInReplica = 0; instanceIdInReplica < numInstancesPerReplica; instanceIdInReplica++) {
+          for (int replicaId : replicaIdsInPool) {
+            replicaIdToInstancesMap[replicaId][instanceIdInReplica] =
+                instanceConfigsInPool.get(instanceIdInPool++).getInstanceName();
+          }
+        }
+      }
+
+      // Assign instances within a replica to one partition if not configured
```
snleee: I am a bit confused on the term that we are using.

`replica group - a set of instances that contains all segments for a partition or table`

`replica - we used to use this term for indicating replication factor for a segment (e.g. 3 replicas)`

Is this still correct in your comments? What do you mean by `instances within a replica`?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/instance/InstanceTagPoolSelector.java
```diff
@@ -0,0 +1,135 @@
...
+        poolsToSelect = new ArrayList<>(numPoolsToSelect);
+        List<Integer> poolsInCluster = new ArrayList<>(pools);
+        for (int i = 0; i < numPoolsToSelect; i++) {
+          poolsToSelect.add(poolsInCluster.get((_tableNameHash + i) % numPools));
+        }
+      }
+
+      // Keep the pools selected
+      LOGGER.info("Selecting pools: {} for table: {}", poolsToSelect, _tableNameWithType);
+      pools.retainAll(poolsToSelect);
```
snleee: Why do we need to modify `pools` here? it seems that it's a local variable and it's not used after here?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/instance/InstanceConstraintApplier.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * The instance constraint applier is responsible for filtering out unqualified instances and sorts instances for picking
+ * priority.
+ */
+public interface InstanceConstraintApplier {
+
+  /**
+   * Applies the instance constraint to the given map from pool to instance configs.
+   */
+  Map<Integer, List<InstanceConfig>> applyConstraint(Map<Integer, List<InstanceConfig>> poolToInstanceConfigsMap);
```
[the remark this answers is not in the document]
Jackie-Jiang: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/instance/InstanceReplicaPartitionSelector.java
Jackie-Jiang: The replica here refers to the table level replica. Added javadoc for the class

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/instance/InstanceTagPoolSelector.java
Jackie-Jiang: `pools` is the keySet() of the `poolToInstanceConfigsMap`, so any change to `pools` will reflect to the map.

snleee: ah i see :)

PR #4514, opened 2019-08-09
Reduce the number of calls to ZK to fetch the current ideal state in rebalancer

siddharthteotia:
In TableRebalancer, after updating the ideal state (after moving segments), if we successfully persisted this in ZK, we were still going back and fetching the current ideal state from ZK again.

The changes here fetch the ideal state from ZK and re-compute target ideal state only if we failed to persist updated ideal state in ZK earlier due to version error. If there is no version error, we simply start the next iteration (if there are more segments to be moved to get to target) by taking ideal state we just wrote in ZK as the current ideal state.


siddharthteotia:
cc @Jackie-Jiang , @mcvsubbu , @snleee

siddharthteotia:
Not needed with after changes to rebalancer as part of instance assignment

PR #4517, opened 2019-08-10
Adding config to use let controller/broker/server to set hostname

fx19880617:
The motivation for this PR is to support cloud deployment in k8s.
In a cloud deployment, usually ip address is not guaranteed to be the same.
So we need to use container provided hostname, e.g. in k8s, we can use `statefulset.apps` to deploy multiple pinot brokers, each instance will be named as `pinot-broker-0`, `pinot-broker-1`, ...

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
@@ -52,6 +52,7 @@ import org.apache.pinot.common.config.TagNameUtils;
 import org.apache.pinot.common.metadata.ZKMetadataProvider;
 import org.apache.pinot.common.metrics.BrokerMeter;
 import org.apache.pinot.common.metrics.BrokerMetrics;
+import org.apache.pinot.common.utils.CommonConstants;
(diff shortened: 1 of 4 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/broker/helix/HelixBrokerStarter.java
```diff
@@ -107,7 +108,9 @@ public HelixBrokerStarter(Configuration brokerConf, String clusterName, String z
     _zkServers = zkServer.replaceAll("\\s+", "");
 
     if (brokerHost == null) {
-      brokerHost = NetUtil.getHostAddress();
+      brokerHost =
+        _brokerConf.getBoolean(CommonConstants.Helix.PREFER_HOSTNAME_IN_DEFAULT_INSTANCD_ID_KEY, false) ? NetUtil
```
kishoreg: Typo Instance can we shorten the constant. SET_INSTANCEID_TO_HOSTNAME

fx19880617: done

fx19880617:
Instance_id is part of helix instance id (Broker_xxxx/Server_xxxx). So it will be logged everywhere along with helix stuffs during startup.

PR #4519, opened 2019-08-10
Add support sql query http api

xueyumusic:
This PR adds support calcite sql query http api, the updates are:
1. Rename PqlQueryResource to PinotQueryResource
2. Adds http post api `/queryByFormat` to support both pql and sql query based on parameter `queryFormat`.
3. Deprecated post api `/pql`.
4. Updates query console query api.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotQueryResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotQueryResource.java
@@ -20,6 +20,10 @@ package org.apache.pinot.controller.api.resources;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.PQL;
+import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.SQL;
+
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotQueryResource.java
```diff
@@ -66,8 +72,31 @@
   @Inject
   AccessControlFactory _accessControlFactory;
 
+  @POST
+  @Path("queryByFormat")
+  public String postQuery(String requestJsonStr, @Context HttpHeaders httpHeaders) {
+    try {
+      JsonNode requestJson = JsonUtils.stringToJsonNode(requestJsonStr);
+      String query = requestJson.get("query").asText();
```
mayankshriv: Use a CommonConstant or static final instead of "query".
Please do so for all other such hardcoded strings in this PR (e.g. "traceEnabled", "queryFormat" etc).

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotQueryResource.java
```diff
@@ -66,8 +72,31 @@
...
+      String traceEnabled = "false";
+      if (requestJson.has("trace")) {
```
mayankshriv: Minor nit: This can be a one-liner like:
`String traceEnabled = (requestJson.has("trace")) ? .... : false;`

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotQueryResource.java
```diff
@@ -66,8 +72,31 @@
...
+      return getQueryResponse(query, traceEnabled, httpHeaders, queryFormat);
+    } catch (Exception e) {
+      LOGGER.error("Caught exception while processing post request", e);
+      return QueryException.getException(QueryException.INTERNAL_ERROR, e).toString();
+    }
+  }
+
   @POST
   @Path("pql")
+  @Deprecated
```
mayankshriv: Another backward compatible way of making this change is to keep "pql" path for PQL queries 'as-is' and add a "sql" path for SQL queries. Unclear to me which is a better option at the moment, but stating so that others can also pitch in. At least, in this approach there is no need to deprecate existing api.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotQueryResource.java
```diff
@@ -86,24 +115,34 @@ public String post(String requestJsonStr, @Context HttpHeaders httpHeaders) {
...
-      brokerRequest = REQUEST_COMPILER.compileToBrokerRequest(pqlQuery);
+      switch (queryFormat) {
+        case PQL:
+          brokerRequest = PQL_COMPILER.compileToBrokerRequest(query);
+          break;
+        case SQL:
+          brokerRequest = SQL_COMPILER.compileToBrokerRequest(query);
+          break;
+        default:
+          throw new UnsupportedOperationException("Unknown query format - " + queryFormat);
```
mayankshriv: If we go the `/pql` and '/sql' route, then this condition never happens.

kishoreg:
@xueyumusic is this still valid?

fx19880617:
I think this is already implemented.

PR #4522, opened 2019-08-12
[Instance Assignment] Enhance Instance class to include the pool config

Jackie-Jiang:
- Enhance Instance class to include the pool config
- Allow multiple tags in the Instance
- Support pools and multiple tags for instance restlet API
- Make the behavior of joining the cluster via instance restlet API and auto-join consistent
  - HELIX_HOST will have the instance type prefix per Helix behavior


pinot-controller/src/main/java/org/apache/pinot/controller/api/pojos/Instance.java
```diff
@@ -19,119 +19,127 @@
 package org.apache.pinot.controller.api.pojos;
```
snleee: If possible, can we move this class to another package?

I think that this should be under `pinot-common`? For instance, `Tenant` is in `pinot-common/.../config`

Jackie-Jiang: Good point. Moved

PR #4524, opened 2019-08-13
Updating license and notice binary files for aea8fca7e

mcvsubbu:
Updated the license and notice binaries as per instructions in
https://cwiki.apache.org/confluence/display/PINOT/Creating+an+Apache+Release
These files are as of commit aea8fca7e from which we expect to cut
pinot-0.2.0 release


mcvsubbu:
Veriified. Excluding the pinot packages, there are a total of 154 packages.

PR #4532, opened 2019-08-15
Expose a method to determine if a QueryExceptionErrorCode represents a client-side error

sunithabeeram:
Add a method in the QueryException class to determine if an error code represents a client-side error. Adding it in the class directly helps keep the method up-to-date when new codes are added.

Also, expose the value used by RequestStatistics object to indicate the default value set for tables so higher layers can use it appropriately to determine whether the table name was parsed correctly or not.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/api/RequestStatistics.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/api/RequestStatistics.java
@@ -28,9 +28,12 @@ import org.apache.pinot.common.response.BrokerResponse;
  * post-processing at a finer level than metrics.
  */
 public class RequestStatistics {
+
+  public static final String DEFAULT_TABLE_NAME = "NotYetParsed";
+
   private int _errorCode = 0;
   private String _pql;
-  private String _tableName = "NotYetParsed";
+  private String _tableName = DEFAULT_TABLE_NAME;
   private long _processingTimeMillis = -1;
 
   private long _totalDocs;
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/api/RequestStatistics.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/api/RequestStatistics.java
@@ -29,7 +29,7 @@ import org.apache.pinot.common.response.BrokerResponse;
  */
 public class RequestStatistics {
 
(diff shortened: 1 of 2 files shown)


pinot-common/src/main/java/org/apache/pinot/common/exception/QueryException.java
```diff
@@ -147,4 +148,24 @@ public static ProcessingException getException(ProcessingException processingExc
...
+   * @return whether the code indicates client error or not
+   */
+  public static boolean isClientError(int errorCode) {
+    switch (errorCode) {
+      case QueryException.ACCESS_DENIED_ERROR_CODE:
+      case QueryException.JSON_COMPILATION_ERROR_CODE:
+      case QueryException.JSON_PARSING_ERROR_CODE:
+      case QueryException.QUERY_VALIDATION_ERROR_CODE:
+      case QueryException.PQL_PARSING_ERROR_CODE:
+      case QueryException.TOO_MANY_REQUESTS_ERROR_CODE: {
```
mcvsubbu: Also QueryException.BROKER_RESOURCE_MISSING_ERROR (unfortunate naming)?

mcvsubbu: Also, why is this a static method?

pinot-broker/src/main/java/org/apache/pinot/broker/api/RequestStatistics.java
```diff
@@ -28,9 +28,12 @@
  * post-processing at a finer level than metrics.
  */
 public class RequestStatistics {
+
+  public static final String DEFAULT_TABLE_NAME = "NotYetParsed";
```
mcvsubbu: private, and introduce a method isTableNameParsed()

pinot-common/src/main/java/org/apache/pinot/common/exception/QueryException.java
mcvsubbu: remove {

mcvsubbu: Please add a comment saying why we dont include BROKER_RESOURCE_MISSING_ERROR here as a client error.

PR #4533, opened 2019-08-16
[Instance Assignment] De-couple assignment strategy from SegmentAssignment

Jackie-Jiang:
For REALTIME segments, we should be able to assign CONSUMING and
COMPLETED segments with different strategies, i.e. one could
follows replica-group while the other follows balance-number. In
order to do so, we need to de-couple the assignment strategy from
the segment assignment, so each assignment can use different
strategies if necessary.

- Change SegmentAssignmentStrategy interface to SegmentAssignment
- Add OfflineSegmentAssignment and RealtimeSegmentAssignment
- Use InstancePartitions to determine the strategy to use
- For REALTIME table, support different strategies for CONSUMING
  and COMPLETED segments


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignment.java
```diff
@@ -23,37 +23,42 @@
 import org.apache.commons.configuration.Configuration;
 import org.apache.helix.HelixManager;
 import org.apache.pinot.common.config.TableConfig;
+import org.apache.pinot.common.utils.InstancePartitionsType;
+import org.apache.pinot.controller.helix.core.assignment.InstancePartitions;
 
 
 /**
- * Strategy to assign segment to instances or rebalance all segments in a table.
+ * Interface for segment assignment and table rebalancing.
```
mcvsubbu: Our documentation states that SegmentAssignmentStrategy is public. You should keep the existing interface and introduce a new one to suit the needs.
https://pinot.readthedocs.io/en/latest/customizations.html#segment-assignment-strategies

Otherwise it is an incompatible change

Jackie-Jiang: This is not the same interface as the one we mentioned in the documentation, but the new added one for the resource assignment feature. We will deprecate the old interface once the new feature is done, and I will add documentation for the new assignment.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/SegmentAssignment.java
```diff
@@ -23,37 +23,42 @@
...
  */
-public interface SegmentAssignmentStrategy {
+public interface SegmentAssignment {
```
snleee: Can you add the `TODO` document for adding support for custom segment assignment strategy?
I agree that it's not the right time to add the interface since we need the proper design for it. I can see multiple ways to achieve custom segment assignment (e.g. cost model - currently the cost is based on # of segments, we can extend this to add resources like storage factor into the cost).

Jackie-Jiang: Good point, added.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/segment/OfflineSegmentAssignment.java
```diff
@@ -0,0 +1,260 @@
...
+  public List<String> assignSegment(String segmentName, Map<String, Map<String, String>> currentAssignment,
+      Map<InstancePartitionsType, InstancePartitions> instancePartitionsMap) {
+    InstancePartitions instancePartitions = instancePartitionsMap.get(InstancePartitionsType.OFFLINE);
+    Preconditions.checkState(instancePartitions != null, "Failed to find OFFLINE instance partitions for table: %s",
+        _offlineTableName);
+    LOGGER.info("Assigning segment: {} with instance partitions: {} for table: {}", segmentName, instancePartitions,
+        _offlineTableName);
+
+    List<String> instancesAssigned;
+    if (instancePartitions.getNumReplicas() == 1) {
```
mcvsubbu: should this method be called getNumReplicaGroups()? It is confusing with the variable _replication (whcih, I suppose can be more than 1 even under this if condition)

Jackie-Jiang: Will address in another pr

PR #4534, opened 2019-08-16
Fixing bytes data type merge logic

fx19880617:
The old implementation convert bytes to string, for dictionary based column, but not for non-dictionary based.

Move this logic to merging part.


fx19880617:
Currently we only convert bytes column with dictionary to hex string. However for bytes column without dictionary, the value returned is still byte[]. this will cause cast exception.

PR #4535, opened 2019-08-18
Implement DISTINCT clause

siddharthteotia:
Implementation of DISTINCT in Pinot execution engine. 

Design doc -- https://docs.google.com/document/d/1Tv51HO5M5S0e18W6GzLYqkdESYHpOvExVbRZISSjj-k/edit#

Testing:

(1) Added unit tests to existing InnerSegment and InterSegment queries tests.
(2) Additional unit tests using custom data generator and result verification.
(3) Cluster integration tests and result verification by running the queries against H2
(4) Query compilation tests

Not done in this PR:

(1) Handling ORDER BY for DISTINCT queries. Once the 'group by + order by' feature PR merges, a follow-up PR will come to integrate with generic order by utility/service being implemented as part of that work

(2) Dictionary based results -- As indicated in the design doc, we can potentially get the resultset from dictionary if the query is on a single column and without filter.


pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/SelectAstNode.java
```diff
@@ -140,7 +144,23 @@ public void updateBrokerRequest(BrokerRequest brokerRequest) {
...
+    }
+
+    List<AggregationInfo> aggregationInfo = brokerRequest.getAggregationsInfo();
+    if (aggregationInfo != null && aggregationInfo.size() == 1) {
+      // Since DISTINCT is implemented as a function in the AggregationOperator,
+      // the compiler would have ensured by now that for a DISTINCT query,
+      // there will be exactly one aggregation info -- erroneous cases would have
+      // already resulted in error when we updated the broker request for child
+      // nodes of this Select node
+      if (aggregationInfo.get(0).getAggregationType().equalsIgnoreCase("distinct")) {
```
mayankshriv: The special casing for distinct seems like an issue to me.

pinot-common/src/thrift/request.thrift
```diff
@@ -155,6 +155,7 @@ struct BrokerRequest {
  16: optional HavingFilterQueryMap havingFilterSubQueryMap;
  17: optional query.PinotQuery pinotQuery;
  18: optional list<SelectionSort> orderBy;
+ 19: optional i32 limit = 0;
```
mayankshriv: Why do we need to add this, we already support limit, right?

pinot-core/src/main/java/org/apache/pinot/core/common/DataBlockCache.java
```diff
@@ -366,4 +366,74 @@ public boolean equals(Object obj) {
       return _column.equals(that._column) && _dataType == that._dataType;
     }
   }
+
+  /**
+   * Row Index based APIs for Single Value columns
+   */
+
+  public int getSVIntAtIndex(final String column, final int index) {
```
mayankshriv: It is unclear to me as to why we need to fetch values row-by-row for distinct, I'd rather avoid exposing this api to avoid function call per row? Also, there seems to be code duplication, `getSVIntAtIndex` should have just called `getIntValuesForSVColumn`?

pinot-core/src/main/java/org/apache/pinot/core/operator/blocks/IntermediateResultsBlock.java
```diff
@@ -231,41 +237,123 @@ private DataTable getSelectionResultDataTable()
...
       throws Exception {
     // Extract each aggregation column name and type from aggregation function context.
-    int numAggregationFunctions = _aggregationFunctionContexts.length;
-    String[] columnNames = new String[numAggregationFunctions];
-    DataSchema.ColumnDataType[] columnDataTypes = new DataSchema.ColumnDataType[numAggregationFunctions];
-    for (int i = 0; i < numAggregationFunctions; i++) {
-      AggregationFunctionContext aggregationFunctionContext = _aggregationFunctionContexts[i];
-      columnNames[i] = aggregationFunctionContext.getAggregationColumnName();
-      columnDataTypes[i] = aggregationFunctionContext.getAggregationFunction().getIntermediateResultColumnType();
+    if (AggregationFunctionUtils.isDistinct(_aggregationFunctionContexts)) {
```
mayankshriv: Yep, once we special case `distinct`, we have to do it at a lot of places, as you already found out. Let's discuss how to resolve it.

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
```diff
@@ -76,7 +83,19 @@ private void extractColumnsAndTransforms(@Nonnull BrokerRequest brokerRequest,
...
+        if (aggregationInfo.getAggregationType().equalsIgnoreCase(AggregationFunctionType.DISTINCT.getName())) {
+          // handle DISTINCT (col1, col2 ...) as an aggregate function
+          final String multiColumnExpression = AggregationFunctionUtils.getColumn(aggregationInfo);
+          final String[] distinctColumnExpressions = multiColumnExpression.split(FunctionCallAstNode.DISTINCT_MULTI_COLUMN_SEPARATOR);
+          int col = 0;
+          for (String distinctColumnExpr : distinctColumnExpressions) {
+            TransformExpressionTree transformExpressionTree = TransformExpressionTree.compileToExpressionTree(distinctColumnExpr);
+            transformExpressionTree.getColumns(_projectionColumns);
+            _expressionTrees.add(transformExpressionTree);
+            _projectionColumnIndices.put(distinctColumnExpr, col++);
```
mayankshriv: Where/how is this being used?

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
```diff
@@ -124,6 +143,10 @@ public TransformOperator run() {
     return new TransformOperator(_projectionPlanNode.run(), new ArrayList<>(_expressionTrees));
   }
 
+  public List<FieldSpec.DataType> getProjectedColumnTypes() {
```
mayankshriv: This should come from metadata, and not from PlanNode, right?

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/AggregationFunctionContext.java
```diff
@@ -54,4 +59,16 @@ public String getColumn() {
   public String getAggregationColumnName() {
     return _aggregationFunction.getColumnName(_column);
   }
+
+  public void setProjectedColumnTypes(final List<FieldSpec.DataType> projectedColumnTypes) {
```
mayankshriv: Same here, column types should be maintained in the metadata.

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DefaultAggregationExecutor.java
```diff
@@ -36,18 +40,40 @@
...
+      final String[] distinctColumnExpressions = multiColumnExpression.split(FunctionCallAstNode.DISTINCT_MULTI_COLUMN_SEPARATOR);
+      _expressions = new TransformExpressionTree[distinctColumnExpressions.length];
+      for (int i = 0; i < distinctColumnExpressions.length; i++) {
+        _expressions[i] = TransformExpressionTree.compileToExpressionTree(distinctColumnExpressions[i]);
+      }
+    } else {
+      _expressions = new TransformExpressionTree[_numFunctions];
+      for (int i = 0; i < _numFunctions; i++) {
+        AggregationFunction function = functionContexts[i].getAggregationFunction();
+        // parser should have already enforce the constraint of not allowing queries
```
mayankshriv: I'd rather make it a unit test.

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/EmptyAggregationFunction.java
```diff
@@ -0,0 +1,111 @@
...
+package org.apache.pinot.core.query.aggregation.function;
+
+import javax.annotation.Nonnull;
+import org.apache.pinot.common.function.AggregationFunctionType;
+import org.apache.pinot.common.utils.DataSchema;
+import org.apache.pinot.core.common.BlockValSet;
+import org.apache.pinot.core.query.aggregation.AggregationResultHolder;
+import org.apache.pinot.core.query.aggregation.groupby.GroupByResultHolder;
+
+public class EmptyAggregationFunction implements AggregationFunction<Object, Comparable> {
```
mayankshriv: s/Empty/Base?

pinot-common/src/thrift/request.thrift
siddharthteotia: The LIMIT is currently only supported in SELECT query path (BrokerRequest.setSelection()) has the LIMIT info (both start and end). Reusing this would have then led to change code down the line where existing code makes decisions based on the fact

(1) if brokerRequest.getSelections() is not null then it is guaranteed to be selection query
(2) if broerRequest.getAggregationInfo() is not null then it is guaranteed to be aggregation query

If we end up reusing the LIMIT stashed away by the parser in SelectAstNode for DISTINCT query as well then the above assumptions will break -- couple of places where existing code is dependent on these invariants are places like CombineService, BrokerReduceService

pinot-core/src/main/java/org/apache/pinot/core/common/DataBlockCache.java
siddharthteotia: At the end to judge uniqueness, we need to look at a row. So simply looking at the return value (which is an array containing all the projected values for the column) from getIntValuesForSVColumn will not help. We should look at each cell from each such array together with the values for other projection column arrays to check if row has already been stored in hashset and store it if now

That's why I introduced the index based APIs.

I see your point though w.r.t making a function call per row... what we can do is use the existing APIs and fetch the values(the array) and then iterate over them to build row, check for existence and store if new. We can avoid the function call overhead this way...

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/OutputColumnListAstNode.java
```diff
@@ -41,8 +42,54 @@ public void addChild(AstNode childNode) {
...
+            }
+          }
+        }
+      }
+    }
+  }
+
   @Override
   public void updateBrokerRequest(BrokerRequest brokerRequest) {
+    validate();
```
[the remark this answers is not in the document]
siddharthteotia: I may have used a not so good function name... by validation I still meant to do further parsing and syntax analysis to catch bad queries at parse time itself -- something that was mentioned in the design doc as well to catch queries like following at parse time itself.

SELECT DISTINCT(col1), col2 FROM foo
SELECT col1, SUM(col2) FROM foo

Pql2CompilerTest.java has more tests for such situations...

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -82,10 +85,30 @@ public AggregationInfo buildAggregationInfo() {
       expression = "*";
     } else {
       List<? extends AstNode> children = getChildren();
-      if (children == null || children.size() != 1) {
-        throw new Pql2CompilationException("Aggregation function expects exact 1 argument");
+      if (children == null) {
+        throw new Pql2CompilationException("Aggregation function expects non-null argument");
+      }
+      if (!_name.equalsIgnoreCase("distinct")) {
```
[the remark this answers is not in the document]
siddharthteotia: For this particular change, we need to specially check for DISTINCT because by doing the function based execution underneath we are breaking the invariant that all aggregation functions (SUM, MIN,MAX etc) will have exactly one argument. So the conditional check becomes something like:

If we are having an aggregate function other than DISTINCT then assert if we don't have one argument

As you see in the existing code, COUNT(*) is also treated specially since there the function node doesn't have any children

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/SelectAstNode.java
siddharthteotia: The reasoning here is same as in the comment below for LIMIT. Currently the LIMIT info is stashed away only for a selection query in BrokerRequest.setSelection(). However for a DISTINCT query (which is internally an aggregation based execution), we need to support LIMIT as well.

However, for this change it is probably fine to remove the check and simply do brokerRequest.setLimit() -- the only user will be aggregation executor for DISTINCT query so it will be harmless to other code

pinot-core/src/main/java/org/apache/pinot/core/operator/blocks/IntermediateResultsBlock.java
siddharthteotia: Other than the two special casing occurrences you pointed out earlier in FunctionCallAstNode and SelectAstNode (which I have mentioned in my comments as to why they are specifically needed), there are following places in the code where we do the check for DISTINCT

(1) DefaultAggregationExecutor  -- when operator initializes the executor
**Reason** -- earlier for each function we had to build column expression tree (since each function can only have a single column). Now for DISTINCT function, we will have to do this for each column

(2) IntermediateResultBlock -- when building the data table. Since now we are sending proper matrix for a DISTINCT query (similar to DataTable for Selection query). The above reason is also applicable here. 

(3) AggregationPlanNode -- to set the projected column types in function context to be used later during building the data table -- this is needed since for aggregation functions, the type was fixed at Long, Double.

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
siddharthteotia: This is being used in the ProjectionOperator to correctly store info on each projected column type and the type info is used later when we build DataTable in InstanceResponeBlock to correctly set the type in DataSchema for each projected column in distinct query

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
siddharthteotia: I see, but as mentioned in the response above https://github.com/apache/incubator-pinot/pull/4535#discussion_r315381929, this projected column type info is being used later to build DataTable in InstanceResponseBlock

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/AggregationFunctionContext.java
siddharthteotia: The source of information is actually coming from metadata as mentioned here https://github.com/apache/incubator-pinot/pull/4535#discussion_r315382449

Just stashing it away in function context to be able to use it during construction of DataTable in InstanceResponeBlock

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -82,10 +85,30 @@ public AggregationInfo buildAggregationInfo() {
...
+        if (children.size() != 1) {
+          throw new Pql2CompilationException("Aggregation function expects exactly 1 argument as column name");
+        } else {
+          expression = TransformExpressionTree.getStandardExpression(children.get(0));
+        }
+      } else {
+        if (children.size() < 1) {
```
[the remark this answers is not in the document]
siddharthteotia: Done -- Included the size as part of null check in parent else. Will throw compilation exception if either null or empty children

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -82,10 +85,30 @@ public AggregationInfo buildAggregationInfo() {
...
+          throw new Pql2CompilationException("Distinct expects 1 or more arguments as column names");
+        } else {
+          final StringBuilder distinctColumnExpr = new StringBuilder();
+          for (int i = 0; i < children.size() ; ++i) {
```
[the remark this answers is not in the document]
siddharthteotia: Good point. Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/OutputColumnListAstNode.java
```diff
@@ -41,8 +42,54 @@ public void addChild(AstNode childNode) {
     }
   }
 
+  private void validate() {
```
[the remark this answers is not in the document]
siddharthteotia: Done -- added comments and explained the query patterns being checked and why

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DefaultAggregationExecutor.java
siddharthteotia: Done -- unit tests were written in Pql2CompilerTest

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/EmptyAggregationFunction.java
siddharthteotia: Done

siddharthteotia:
Fixed test failures and rebased

siddharthteotia:
Made changes to handle DISTINCT queries via Calcite SQL compiler. 

All the changes aimed for this PR are done except for TODOs mentioned in the description. All review comments have been addressed. Please take a look

siddharthteotia:
Rebased

mayankshriv:
Seems like this will be WIP for sometime. I'd recommend not enabling the `distinct` support just yet, i.e. throw exception in parser, we have a complete solution.

siddharthteotia:
> Seems like this will be WIP for sometime. I'd recommend not enabling the `distinct` support just yet, i.e. throw exception in parser, we have a complete solution.

Yes the support is currently disabled. The parser (both pql and sql) will throw exception

siddharthteotia:
setRecordLimit() API is needed because we need to pass the upper bounds to restrict memory usage. Using a separate executor will not work as the bounds work at three different levels -- segment, server and broker. Executor is only used for the first one. Using the result holder will also not work for the same reason. 

That's why an API was introduced to trickle down this information to the function since the latter doesn't know if it is being used by the segment, server or broker.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformBlockDataFetcher.java
```diff
@@ -19,11 +19,7 @@
 package org.apache.pinot.core.operator.transform;
```
Jackie-Jiang: Why do we need to touch this class? This is for selection queries only

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
```diff
@@ -63,7 +65,8 @@ public TransformPlanNode(@Nonnull IndexSegment indexSegment, @Nonnull BrokerRequ
     _segmentName = indexSegment.getSegmentName();
     extractColumnsAndTransforms(brokerRequest, indexSegment);
     _projectionPlanNode =
-        new ProjectionPlanNode(indexSegment, _projectionColumns, new DocIdSetPlanNode(indexSegment, brokerRequest, _maxDocPerNextCall));
+        new ProjectionPlanNode(indexSegment, _projectionColumns,
```
Jackie-Jiang: (nit) Seems not formatted correctly

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
```diff
@@ -76,7 +79,18 @@ private void extractColumnsAndTransforms(@Nonnull BrokerRequest brokerRequest,
       IndexSegment indexSegment) {
     if (brokerRequest.isSetAggregationsInfo()) {
       for (AggregationInfo aggregationInfo : brokerRequest.getAggregationsInfo()) {
-        if (!aggregationInfo.getAggregationType().equalsIgnoreCase(AggregationFunctionType.COUNT.getName())) {
+        if (aggregationInfo.getAggregationType().equalsIgnoreCase(AggregationFunctionType.DISTINCT.getName())) {
+          // handle DISTINCT (col1, col2 ...) as an aggregate function
+          final String multiColumnExpression = AggregationFunctionUtils.getColumn(aggregationInfo);
```
Jackie-Jiang: (nit) final

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DefaultAggregationExecutor.java
```diff
@@ -41,13 +44,29 @@ public DefaultAggregationExecutor(@Nonnull AggregationFunctionContext[] function
...
-      _resultHolders[i] = _functions[i].createAggregationResultHolder();
-      if (function.getType() != AggregationFunctionType.COUNT) {
-        _expressions[i] = TransformExpressionTree.compileToExpressionTree(functionContexts[i].getColumn());
+    if (AggregationFunctionUtils.isDistinct(functionContexts)) {
+      // handle distinct (col1, col2..) function
+      // unlike other aggregate functions, distinct can work on multiple columns
+      // so we need to build expression tree for each column
+      _functions[0] = functionContexts[0].getAggregationFunction();
+      _resultHolders[0] = _functions[0].createAggregationResultHolder();
+      final String multiColumnExpression = functionContexts[0].getColumn();
```
Jackie-Jiang: (nit) final

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * This serves the following purposes:
+ *
+ * (1) Intermediate result object for Distinct aggregation function
+ * (2) The same object is serialized by the server inside the data table
+ * for sending the results to broker. Broker deserializes it.
+ */
+public class DistinctTable {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DistinctTable.class);
+  private final static int INITIAL_CAPACITY = 64000;
```
Jackie-Jiang: (nit) private static final

Jackie-Jiang: (Major) Why do we create map with such big initial capacity? It does not make sense for small limit right? We can make a cap on the initial capacity, and use `Math.min(MAX_INITIAL_CAPACITY, limit / loadFactor)`

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,235 @@
...
+  private int _recordLimit;
+  private Set<Key> _table;
+
+  /**
+   * Add a row to hash table
+   * @param key multi-column key to add
+   */
+  public void addKey(final Key key) {
+    if (_table.size() >= _recordLimit) {
+      LOGGER.info("Distinct table Reached allowed max cardinality of {}", _recordLimit);
```
Jackie-Jiang: (Major) This will flood the log

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,235 @@
...
+    _table = new HashSet<>(INITIAL_CAPACITY);
+  }
+
+  /**
+   * DESERIALIZE: Broker side
+   * @param byteBuffer data to deserialize
+   * @throws IOException
+   */
+  public DistinctTable(final ByteBuffer byteBuffer) throws IOException {
+    final DataTable dataTable = DataTableFactory.getDataTable(byteBuffer);
```
Jackie-Jiang: (nit) final

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,235 @@
...
+    final DataSchema dataSchema = dataTable.getDataSchema();
+    final int numRows = dataTable.getNumberOfRows();
+    final int numColumns = dataSchema.size();
+
+    _table = new HashSet<>();
+
+    // extract rows from the datatable
+    for (int rowIndex  = 0; rowIndex < numRows; rowIndex++) {
```
Jackie-Jiang: format the class

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,235 @@
...
+      for (int colIndex = 0; colIndex < numColumns; colIndex++) {
+        DataSchema.ColumnDataType columnDataType = dataSchema.getColumnDataType(colIndex);
+        switch (columnDataType) {
+          case INT:
+            columnValues[colIndex] = dataTable.getInt(rowIndex, colIndex);
+            break;
+          case LONG:
+            columnValues[colIndex] = dataTable.getLong(rowIndex, colIndex);
+            break;
+          case STRING:
```
Jackie-Jiang: (nit) put STRING after DOUBLE

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,235 @@
...
+ * This serves the following purposes:
+ *
+ * (1) Intermediate result object for Distinct aggregation function
+ * (2) The same object is serialized by the server inside the data table
+ * for sending the results to broker. Broker deserializes it.
+ */
+public class DistinctTable {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DistinctTable.class);
+  private final static int INITIAL_CAPACITY = 64000;
+  private FieldSpec.DataType[] _projectedColumnTypes;
```
Jackie-Jiang: Let's just name these two variables _columnTypes and _columnNames

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,235 @@
...
+    final int numColumns = dataSchema.size();
+    final FieldSpec.DataType[] columnTypes = new FieldSpec.DataType[numColumns];
+    for (int colIndex = 0; colIndex < numColumns; colIndex++) {
+      DataSchema.ColumnDataType columnDataType = dataSchema.getColumnDataType(colIndex);
+      switch (columnDataType) {
+        case INT:
+          columnTypes[colIndex] = FieldSpec.DataType.INT;
+          break;
+        case LONG:
+          columnTypes[colIndex] = FieldSpec.DataType.INT;
```
Jackie-Jiang: (Major) All INT???

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,235 @@
...
+            columnValues[colIndex] = dataTable.getString(rowIndex, colIndex);
+            break;
+          case FLOAT:
+            columnValues[colIndex] = dataTable.getFloat(rowIndex, colIndex);
+            break;
+          case DOUBLE:
+            columnValues[colIndex] = dataTable.getDouble(rowIndex, colIndex);
+            break;
+          default:
+            throw new UnsupportedOperationException("Unexpected column data type " + columnDataType + " in data table for DISTINCT query");
```
Jackie-Jiang: (nit) Prefer IllegalStateException

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/AggregationFunctionUtils.java
```diff
@@ -59,24 +63,35 @@ public static AggregationFunctionColumnPair getFunctionColumnPair(@Nonnull Aggre
...
+
   /**
    * Creates an {@link AggregationFunctionContext} from the {@link AggregationInfo}.
    */
   @Nonnull
   public static AggregationFunctionContext getAggregationFunctionContext(@Nonnull AggregationInfo aggregationInfo) {
-    String functionName = aggregationInfo.getAggregationType();
-    AggregationFunction aggregationFunction = AggregationFunctionFactory.getAggregationFunction(functionName);
-    return new AggregationFunctionContext(aggregationFunction, AggregationFunctionUtils.getColumn(aggregationInfo));
+   return getAggregationFunctionContext(aggregationInfo, new BrokerRequest());
```
Jackie-Jiang: Pass in `null`

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/AggregationFunctionUtils.java
```diff
@@ -89,12 +104,16 @@ public static AggregationFunctionContext getAggregationFunctionContext(@Nonnull
   }
 
   @Nonnull
-  public static AggregationFunction[] getAggregationFunctions(@Nonnull List<AggregationInfo> aggregationInfos) {
+  public static AggregationFunction[] getAggregationFunctions(BrokerRequest brokerRequest) {
+    List<AggregationInfo> aggregationInfos = brokerRequest.getAggregationsInfo();
     int numAggregationFunctions = aggregationInfos.size();
     AggregationFunction[] aggregationFunctions = new AggregationFunction[numAggregationFunctions];
     for (int i = 0; i < numAggregationFunctions; i++) {
+      AggregationInfo aggregationInfo = aggregationInfos.get(i);
```
Jackie-Jiang: Revert?

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * DISTINCT clause in SQL is implemented as function in the
+ * execution engine of Pinot.
+ */
+public class DistinctAggregationFunction implements AggregationFunction<DistinctTable, Comparable> {
+
+  private final DistinctTable _distinctTable;
+  private final String[] columnNames;
+  private FieldSpec.DataType[] _dataTypes;
```
Jackie-Jiang: No need for a separate init, just check `if (_dataTypes == null)`

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -0,0 +1,165 @@
...
+    if (!init) {
+      for (int i = 0; i < blockValSets.length; i++) {
+        _dataTypes[i] = blockValSets[i].getValueType();
+      }
+      _distinctTable.setProjectedColumnNames(columnNames);
+      _distinctTable.setProjectedColumnTypes(_dataTypes);
+      init = true;
+    }
+
+    // TODO: Follow up PR will make few changes to start using DictionaryBasedAggregationOperator
```
Jackie-Jiang: Without filter and on single column

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  @Nonnull
+  @Override
+  public AggregationResultHolder createAggregationResultHolder() {
+    return new ObjectAggregationResultHolder();
+  }
+
+  @Override
+  public void aggregate(int records, AggregationResultHolder aggregationResultHolder,
```
Jackie-Jiang: (nit) first parameter `length`

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -0,0 +1,165 @@
...
+      _distinctTable.setProjectedColumnTypes(_dataTypes);
+      init = true;
+    }
+
+    // TODO: Follow up PR will make few changes to start using DictionaryBasedAggregationOperator
+    // for DISTINCT queries without filter.
+    TransformBlockDataFetcher transformBlockDataFetcher = new TransformBlockDataFetcher(blockValSets, new Dictionary[0], new TransformResultMetadata[0]);
+
+    int rowIndex = 0;
+    while (rowIndex < records) {
```
Jackie-Jiang: (Major) track distinct table size and early terminate if size reached. No need to track size inside distinct table

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -0,0 +1,165 @@
...
+  @Override
+  public DistinctTable extractAggregationResult(AggregationResultHolder aggregationResultHolder) {
+    return _distinctTable;
+  }
+
+  @Nonnull
+  @Override
+  public DistinctTable merge(DistinctTable inProgressMergedResult, DistinctTable newResultToMerge) {
+    // do the union
+    final Iterator<Key> iterator = newResultToMerge.getIterator();
```
Jackie-Jiang: Same, track size here.

siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
siddharthteotia: Done -- right now only tracking in the function and returning if reached the limit. However, added a TODO to do this in operator to do early termination by not getting the next TransformBlock and calling function.aggregate() if the previous call to aggregate() hit the limit. But this requires interface change.

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
siddharthteotia: Exactly -- that's what we decided in the design doc

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
siddharthteotia: Yes. Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/AggregationFunctionUtils.java
siddharthteotia: Thanks. Had forgotten earlier

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/AggregationFunctionUtils.java
siddharthteotia: Done -- although I think making nonnullable would be better. It would then avoid unnecessary conditional check before dereferencing in getAggregationFunction()

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: Thanks. This was missed -- didn't get caught in tests as this was invoked by broker after deserializing and doing the reduce but at that point we already have the SelectionResult

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: Thanks. Had noticed during perf testing as well. Removed now

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: If the user specified limit is small (or in case user didn't specify and we defaulted to 10), then yes 64k is high and wastage. 

If not (may be the user is aware of the cardinality and appropriately used LIMIT N), then using high initial capacity will reduce the number of resize/rehash. 

Currently using max initial capacity as 64k and setting the initial capacity as min(max, limit/loadFactor)

siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DefaultAggregationExecutor.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformBlockDataFetcher.java
siddharthteotia: Discussed offline

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformBlockDataFetcher.java
```diff
@@ -49,7 +49,32 @@ public TransformBlockDataFetcher(BlockValSet[] blockValSets, Dictionary[] dictio
...
+      case DOUBLE:
+        return new SVDoubleValueFetcher(blockValSet.getDoubleValuesSV());
+      case BOOLEAN:
+      case STRING:
+        return new SVStringValueFetcher(blockValSet.getStringValuesSV());
+      case BYTES:
+        return new SVBytesValueFetcher(blockValSet.getBytesValuesSV());
+    }
+
+    throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
siddharthteotia: Based on today's discussion and with your new changes, this is not needed anyway

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformBlockDataFetcher.java
```diff
@@ -49,7 +49,32 @@ public TransformBlockDataFetcher(BlockValSet[] blockValSets, Dictionary[] dictio
...
+    switch (blockValSet.getValueType()) {
+      case INT:
+        return new SVIntValueFetcher(blockValSet.getIntValuesSV());
+      case LONG:
+        return new SVLongValueFetcher(blockValSet.getLongValuesSV());
+      case FLOAT:
+        return new SVFloatValueFetcher(blockValSet.getFloatValuesSV());
+      case DOUBLE:
+        return new SVDoubleValueFetcher(blockValSet.getDoubleValuesSV());
+      case BOOLEAN:
```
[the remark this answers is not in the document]
siddharthteotia: Based on today's discussion and with your new changes, this is not needed anyway

pinot-core/src/main/java/org/apache/pinot/core/common/ObjectSerDeUtils.java
```diff
@@ -425,8 +459,19 @@ public TDigest deserialize(ByteBuffer byteBuffer) {
...
+      DOUBLE_SER_DE,
+      DOUBLE_ARRAY_LIST_SER_DE,
+      AVG_PAIR_SER_DE,
+      MIN_MAX_RANGE_PAIR_SER_DE,
+      HYPER_LOG_LOG_SER_DE,
+      QUANTILE_DIGEST_SER_DE,
+      MAP_SER_DE,
+      INT_SET_SER_DE,
+      TDIGEST_SER_DE,
+      DISTINCT_TABLE_SER_DE};
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/common/ObjectSerDeUtils.java
```diff
@@ -277,6 +281,36 @@ public HyperLogLog deserialize(ByteBuffer byteBuffer) {
...
+        throw new RuntimeException("Caught exception while de-serializing HyperLogLog", e);
+      }
+    }
+
+    @Override
+    public DistinctTable deserialize(ByteBuffer byteBuffer) {
+      try {
+        return new DistinctTable(byteBuffer);
+      } catch (IOException e) {
+        throw new RuntimeException("Caught exception while de-serializing HyperLogLog", e);
```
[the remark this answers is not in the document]
siddharthteotia: Thanks -- copy paste error

pinot-core/src/main/java/org/apache/pinot/core/common/ObjectSerDeUtils.java
```diff
@@ -277,6 +281,36 @@ public HyperLogLog deserialize(ByteBuffer byteBuffer) {
...
+        throw new RuntimeException("Caught exception while serializing TupleHolder", e);
+      }
+    }
+
+    @Override
+    public DistinctTable deserialize(byte[] bytes) {
+      try {
+        return new DistinctTable(ByteBuffer.wrap(bytes));
+      } catch (IOException e) {
+        throw new RuntimeException("Caught exception while de-serializing HyperLogLog", e);
```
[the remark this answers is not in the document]
siddharthteotia: Thanks -- copy paste error

pinot-core/src/main/java/org/apache/pinot/core/common/ObjectSerDeUtils.java
```diff
@@ -277,6 +281,36 @@ public HyperLogLog deserialize(ByteBuffer byteBuffer) {
...
   };
 
+  public static final ObjectSerDe<DistinctTable> DISTINCT_TABLE_SER_DE = new ObjectSerDe<DistinctTable>() {
+
+    @Override
+    public byte[] serialize(DistinctTable tupleHolder) {
+      try {
+        return tupleHolder.toBytes();
+      } catch (IOException e) {
+        throw new RuntimeException("Caught exception while serializing TupleHolder", e);
```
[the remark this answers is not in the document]
siddharthteotia: Sorry -- this went through various naming rounds before we zeroed on DistinctTable

pinot-core/src/main/java/org/apache/pinot/core/common/ObjectSerDeUtils.java
```diff
@@ -93,6 +95,8 @@ public static ObjectType getObjectType(Object value) {
         return ObjectType.IntSet;
       } else if (value instanceof TDigest) {
         return ObjectType.TDigest;
+      } else if (value instanceof org.apache.pinot.core.query.aggregation.DistinctTable) {
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/sql/parsers/CalciteSqlParser.java
```diff
@@ -174,13 +182,23 @@ private static String removeOptionsFromSql(String sql) {
     return matcher.replaceAll("");
   }
 
-  private static List<Expression> convertSelectList(SqlNodeList selectList) {
+  private static List<Expression> convertSelectList(SqlNodeList selectList, SqlNode distinctNode) {
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/sql/parsers/CalciteSqlParser.java
```diff
@@ -213,6 +231,32 @@ private static Expression convertOrderBy(SqlNode node) {
...
 
+  /**
+   * DISTINCT is implemented as an aggregation function so need to take the select list items
+   * and convert them into a single function expression for handing over to execution engine
+   * either as a PinotQuery or BrokerRequest via conversion
+   * @param selectList select list items
+   * @return DISTINCT function expression
+   */
+  private static Expression convertDistinctAndSelectListToFunctionExpression(final SqlNodeList selectList) {
+    final String functionName = FunctionCallAstNode.DISTINCT_FUNCTION_NAME;
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/OutputColumnListAstNode.java
```diff
@@ -41,8 +43,87 @@ public void addChild(AstNode childNode) {
...
+            for (AstNode selectChild : child.getChildren()) {
+              if (selectChild instanceof IdentifierAstNode) {
+                if (distinctPresent) {
+                  throw new Pql2CompilationException("Syntax error: SELECT list columns should be part of DISTINCT clause");
+                } else {
+                  identifierPresent = true;
+                }
+              } else if (selectChild instanceof FunctionCallAstNode) {
+                FunctionCallAstNode function = (FunctionCallAstNode)selectChild;
+                if (function.getName().equalsIgnoreCase("distinct")) {
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/OutputColumnListAstNode.java
```diff
@@ -41,8 +43,87 @@ public void addChild(AstNode childNode) {
...
+   * again a valid SQL syntax for multi column distinct, we can support this since timeConvert
+   * is a valid supported transform function.
+   */
+  private void validate() {
+    boolean identifierPresent = false;
+    boolean distinctPresent = false;
+    boolean functionPresent = false;
+    if (hasChildren()) {
+      for (AstNode child : getChildren()) {
+        if (child instanceof OutputColumnAstNode) {
```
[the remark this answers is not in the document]
siddharthteotia: I don't think so. In fact it can only be of type OutputColumnAstNode which is an AstNode

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -75,22 +83,52 @@ public String getExpression() {
...
+          throw new Pql2CompilationException("Aggregation function expects exactly 1 argument as column name");
+        } else {
+          expression = TransformExpressionTree.getStandardExpression(children.get(0));
+        }
+      } else {
+        // DISTINCT
+        if (!Pql2Compiler.ENABLE_DISTINCT) {
+          throw new Pql2CompilationException("Support for DISTINCT is currently disabled");
+        }
+        if (children.size() == 1) {
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -33,6 +36,11 @@
   private String _expression;
   private boolean _isInSelectList;
 
+  public static final String DISTINCT_MULTI_COLUMN_SEPARATOR = ":";
+  public static final String COLUMN_KEY_IN_AGGREGATION_INFO = "column";
+  public static final String COUNT_FUNCTION_NAME = "count";
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -223,38 +226,65 @@ private String standardizeFunction(Function functionCall) {
...
+      if (operands.size() == 1) {
+        // single column DISTINCT
+        columnName = getColumnExpression(operands.get(0));
+      } else {
+        // multi column DISTINCT
+        final StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < operands.size() ; ++i) {
+          Expression expression = operands.get(i);
+          sb.append(getColumnExpression(expression));
+          if (i <= operands.size() - 2) {
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -223,38 +226,65 @@ private String standardizeFunction(Function functionCall) {
...
+
+    if (functionName.equalsIgnoreCase(FunctionCallAstNode.DISTINCT_FUNCTION_NAME)) {
+      // DISTINCT can support multiple arguments
+      if (operands.size() == 1) {
+        // single column DISTINCT
+        columnName = getColumnExpression(operands.get(0));
+      } else {
+        // multi column DISTINCT
+        final StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < operands.size() ; ++i) {
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -223,38 +226,65 @@ private String standardizeFunction(Function functionCall) {
...
+    String functionName = function.getOperator();
+
+    if (functionName.equalsIgnoreCase(FunctionCallAstNode.DISTINCT_FUNCTION_NAME)) {
+      // DISTINCT can support multiple arguments
+      if (operands.size() == 1) {
+        // single column DISTINCT
+        columnName = getColumnExpression(operands.get(0));
+      } else {
+        // multi column DISTINCT
+        final StringBuilder sb = new StringBuilder();
```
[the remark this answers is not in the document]
siddharthteotia: Done -- previous coding practice and I am getting used to get rid of it

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -223,38 +226,65 @@ private String standardizeFunction(Function functionCall) {
...
+        columnName = "*";
+      } else {
+        Expression functionParam = operands.get(0);
+        columnName = getColumnExpression(functionParam);
       }
     }
+
     AggregationInfo aggregationInfo = new AggregationInfo();
     aggregationInfo.setAggregationType(functionName);
     aggregationInfo.putToAggregationParams("column", columnName);
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: In fact this is not needed for deser. Removed the method

> Please add a unit test for DistinctTable
> 
> Bonus points if you can also cover BrokerReduceService via unit tests

Tests added in DistinctQueriesTest test the entire round trip of planning the query and executing from the broker -- thus all levels of execution code are exercised; segment, server level merge and broker level merge. Similar tests are also added in InterSegmentAggregationTests and OfflineClusterIntegrationTest

The tests exercise each and every method of DistinctTable -- adding the key, iterator, serialization by server to send DataTable to broker and deserialization at broker to do the merge and then set broker response. As part of asserts in these tests, I check that distinct table works fine (both actual data and metadata like column names and types) by comparing the end result (at the broker level and segment level) with the expected table that the test code builds as it generates custom data. So DistinctTable's code has got 100% coverage through the tests added as follows:

DistinctQueriesTest.java

- Exercises segment level execution and checks the data (unique rows) and metadata (column names and types) in Distinct Table.
- Exercises full inter segment and inter server execution via broker and checks the broker response data (actual unique rows) and metadata (column names) -- this broker response is set after deserializing the DistinctTable and doing the reduce/merge

OfflineClusterIntegrationTest.java

- Exercises full inter segment and inter server execution via broker and compares the result against H2

InnerSegmentAggregationSingleValueQueriesTest

- Exercises segment level execution and checks the data (unique rows) and metadata (column names and types) in Distinct Table.

InterSegmentAggregationSingleValueQueriesTest

- Exercises full inter segment and inter server execution via broker and checks the broker response data (actual unique rows) and metadata (column names) -- this broker response is set after deserializing the DistinctTable and doing the reduce/merge

siddharthteotia:
> Please add a unit test for DistinctTable
> 
> Bonus points if you can also cover BrokerReduceService via unit tests

Tests added in DistinctQueriesTest test the entire round trip of planning the query and executing from the broker -- thus all levels of execution code are exercised; segment, server level merge and broker level merge. Similar tests are also added in InterSegmentAggregationTests and OfflineClusterIntegrationTest

The tests exercise each and every method of DistinctTable -- adding the key, iterator, serialization by server to send DataTable to broker and deserialization at broker to do the merge and then set broker response. As part of asserts in these tests, I check that distinct table works fine (both actual data and metadata like column names and types) by comparing the end result (at the broker level and segment level) with the expected table that the test code builds as it generates custom data. So DistinctTable's code has got 100% coverage through the tests added as follows:

DistinctQueriesTest.java

- Exercises segment level execution and checks the data (unique rows) and metadata (column names and types) in Distinct Table.

- Exercises full inter segment and inter server execution via broker and checks the broker response data (actual unique rows) and metadata (column names) -- this broker response is set after deserializing the DistinctTable and doing the reduce/merge

OfflineClusterIntegrationTest.java

- Exercises full inter segment and inter server execution via broker and compares the result against H2

InnerSegmentAggregationSingleValueQueriesTest

- Exercises segment level execution and checks the data (unique rows) and metadata (column names and types) in Distinct Table.

InterSegmentAggregationSingleValueQueriesTest

- Exercises full inter segment and inter server execution via broker and checks the broker response data (actual unique rows) and metadata (column names) -- this broker response is set after deserializing the DistinctTable and doing the reduce/merge

pinot-common/src/thrift/request.thrift
siddharthteotia: Yes. If the user didn't specify LIMIT N, then we use the default 10 and set it here.

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -111,6 +113,7 @@ private void convertGroupBy(PinotQuery pinotQuery, BrokerRequest brokerRequest)
   }
 
   private void convertSelectList(PinotQuery pinotQuery, BrokerRequest brokerRequest) {
+
```
Jackie-Jiang: (nit) revert this new line

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -223,38 +226,66 @@ private String standardizeFunction(Function functionCall) {
 
   private AggregationInfo buildAggregationInfo(Function function) {
     List<Expression> operands = function.getOperands();
-    if (operands == null || operands.size() != 1) {
-      throw new Pql2CompilationException(
-          "Aggregation function" + function.getOperator() + " expects 1 argument. found: " + operands);
+    if (operands == null || operands.size() < 1) {
```
Jackie-Jiang: (nit) `if (operands == null || operands.isEmpty())`

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -223,38 +226,66 @@ private String standardizeFunction(Function functionCall) {
...
+    if (functionName.equalsIgnoreCase(AggregationFunctionType.DISTINCT.getName())) {
+      // DISTINCT can support multiple arguments
+      if (operands.size() == 1) {
+        // single column DISTINCT
+        columnName = getColumnExpression(operands.get(0));
+      } else {
+        // multi column DISTINCT
+        StringBuilder sb = new StringBuilder();
+        int numOperands = operands.size();
+        for (int i = 0; i < numOperands; ++i) {
```
Jackie-Jiang: (nit) `i++` for consistency

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -75,22 +80,53 @@ public String getExpression() {
...
     String expression;
     // COUNT aggregation function always works on '*'
-    if (_name.equalsIgnoreCase("count")) {
+    if (_name.equalsIgnoreCase(AggregationFunctionType.COUNT.getName())) {
       expression = "*";
     } else {
       List<? extends AstNode> children = getChildren();
-      if (children == null || children.size() != 1) {
-        throw new Pql2CompilationException("Aggregation function expects exact 1 argument");
+      if (children == null || children.size() < 1) {
```
Jackie-Jiang: (nit) if (children == null || children.isEmpty())

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -223,38 +226,66 @@ private String standardizeFunction(Function functionCall) {
...
     String columnName;
-    if (functionName.equalsIgnoreCase("count")) {
-      columnName = "*";
+    String functionName = function.getOperator();
+
+    if (functionName.equalsIgnoreCase(AggregationFunctionType.DISTINCT.getName())) {
+      // DISTINCT can support multiple arguments
+      if (operands.size() == 1) {
+        // single column DISTINCT
+        columnName = getColumnExpression(operands.get(0));
```
Jackie-Jiang: Do we want to support `SELECT DISTINCT * FROM table`?

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
```diff
@@ -223,38 +226,66 @@ private String standardizeFunction(Function functionCall) {
...
+      } else {
+        // multi column DISTINCT
+        StringBuilder sb = new StringBuilder();
+        int numOperands = operands.size();
+        for (int i = 0; i < numOperands; ++i) {
+          Expression expression = operands.get(i);
```
Jackie-Jiang: Deduplicate the columns

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -75,22 +80,53 @@ public String getExpression() {
...
+          expression = TransformExpressionTree.getStandardExpression(children.get(0));
+        } else {
+          // multi-column DISTINCT query
+          // e.g SELECT DISTINCT(col1, col2) FROM foo
+          // we will pass down the column expression to execution code
+          // as col1:col2
+          StringBuilder distinctColumnExpr = new StringBuilder();
+          int numChildren = children.size();
+          for (int i = 0; i < numChildren; ++i) {
+            expression = TransformExpressionTree.getStandardExpression(children.get(i));
```
Jackie-Jiang: Deduplicate the columns

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
```diff
@@ -75,22 +80,53 @@ public String getExpression() {
...
+        }
+      } else {
+        // DISTINCT
+        if (!Pql2Compiler.ENABLE_DISTINCT) {
+          throw new Pql2CompilationException("Support for DISTINCT is currently disabled");
+        }
+        if (children.size() == 1) {
+          // single column DISTINCT query
+          // e.g SELECT DISTINCT(col) FROM foo
+          expression = TransformExpressionTree.getStandardExpression(children.get(0));
```
Jackie-Jiang: Do we want to support `SELECT DISTINCT * FROM table`?

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/OutputColumnListAstNode.java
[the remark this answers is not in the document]
Jackie-Jiang: Can you add comments on that? Also, if StarAstNode exists, there should be no other nodes based on the sql standard. We should have check on that

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/OutputColumnListAstNode.java
```diff
@@ -41,8 +43,90 @@ public void addChild(AstNode childNode) {
...
+          if (child.hasChildren()) {
```
Jackie-Jiang: I think this if check is redundant?

pinot-common/src/main/java/org/apache/pinot/sql/parsers/CalciteSqlParser.java
```diff
@@ -174,13 +185,24 @@ private static String removeOptionsFromSql(String sql) {
...
+    return selectExpr;
+  }
+
   private static List<Expression> convertSelectList(SqlNodeList selectList) {
     List<Expression> selectExpr = new ArrayList<>();
+
     final Iterator<SqlNode> iterator = selectList.iterator();
     while (iterator.hasNext()) {
       final SqlNode next = iterator.next();
       selectExpr.add(toExpression(next));
```
Jackie-Jiang: Not part of this PR, but we should deduplicate the columns

pinot-core/src/main/java/org/apache/pinot/core/common/ObjectSerDeUtils.java
```diff
@@ -277,6 +281,36 @@ public HyperLogLog deserialize(ByteBuffer byteBuffer) {
...
   };
 
+  public static final ObjectSerDe<DistinctTable> DISTINCT_TABLE_SER_DE = new ObjectSerDe<DistinctTable>() {
+
+    @Override
+    public byte[] serialize(DistinctTable distinctTable) {
+      try {
+        return distinctTable.toBytes();
+      } catch (IOException e) {
+        throw new IllegalStateException("Caught exception while serializing DistinctHolder", e);
```
Jackie-Jiang: "Caught exception while serializing Distinct**Table**"

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DefaultAggregationExecutor.java
```diff
@@ -56,12 +77,25 @@ public DefaultAggregationExecutor(@Nonnull AggregationFunctionContext[] function
   public void aggregate(@Nonnull TransformBlock transformBlock) {
     int length = transformBlock.getNumDocs();
     for (int i = 0; i < _numFunctions; i++) {
-      AggregationFunction function = _functions[i];
-      AggregationResultHolder resultHolder = _resultHolders[i];
-
+      final AggregationFunction function = _functions[i];
```
Jackie-Jiang: (nit) revert the **final**s

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
```diff
@@ -0,0 +1,228 @@
...
+          case FLOAT:
+            columnValues[colIndex] = dataTable.getFloat(rowIndex, colIndex);
+            break;
+          case DOUBLE:
+            columnValues[colIndex] = dataTable.getDouble(rowIndex, colIndex);
+            break;
+          case STRING:
+            columnValues[colIndex] = dataTable.getString(rowIndex, colIndex);
+            break;
+          default:
```
Jackie-Jiang: BYTES support?

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -0,0 +1,169 @@
...
```
Jackie-Jiang: (nit) final

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -0,0 +1,169 @@
...
+
+
+/**
+ * DISTINCT clause in SQL is implemented as function in the
+ * execution engine of Pinot.
+ */
+public class DistinctAggregationFunction implements AggregationFunction<DistinctTable, Comparable> {
+
+  private final DistinctTable _distinctTable;
+  private final String[] columnNames;
```
Jackie-Jiang: `_columnNames`

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -0,0 +1,169 @@
...
+    int rowIndex = 0;
+    // TODO: Do early termination in the operator itself which should
+    // not call aggregate function at all if the limit has reached
+    // that will require the interface change since this function
+    // has to communicate back that required number of records have
+    // been collected
+    while (rowIndex < length && _distinctTable.size() < _limit) {
+      Object[] columnData = transformBlockDataFetcher.getRow(rowIndex);
+      _distinctTable.addKey(new Key(columnData));
+      ++rowIndex;
```
Jackie-Jiang: (nit) `rowIndex++`

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -377,19 +386,50 @@ private void setAggregationResults(@Nonnull BrokerResponseNative brokerResponseN
...
+    // The DISTINCT query is just another SELECTION style query from the user's point of view
+    // and will return one or records in the result table for the column selected.
+    // Internally the execution is happening as an aggregation function (but that is an implementation
+    // detail) and so for that reason, response from broker should be a selection query result
+    // up until now, we have treated DISTINCT similar to another aggregation function even in terms
+    // of the result from function since it has been implemented as an aggregation function.
+    // However, the broker response will be a selection query response as that makes sense from SQL
+    // perspective
+    if (isDistinct(aggregationFunctions)) {
+      final Object merged = intermediateResults[0];
```
Jackie-Jiang: (nit) final

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
siddharthteotia: Discussed offline. We don't support that

pinot-common/src/main/java/org/apache/pinot/pql/parsers/PinotQuery2BrokerRequestConverter.java
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
siddharthteotia: Done

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/FunctionCallAstNode.java
siddharthteotia: Discussed offline. We don't support that.

pinot-common/src/main/java/org/apache/pinot/pql/parsers/pql2/ast/OutputColumnListAstNode.java
siddharthteotia: Yes. Thanks for catching

pinot-common/src/main/java/org/apache/pinot/sql/parsers/CalciteSqlParser.java
siddharthteotia: Good catch. For the DISTINCT code via PQL, I deduplicate columns. For calcite compilation path, I can do that in a follow-up since it needs to be done for selection also.

pinot-core/src/main/java/org/apache/pinot/core/common/ObjectSerDeUtils.java
siddharthteotia: Done.

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DefaultAggregationExecutor.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/DistinctTable.java
siddharthteotia: I had initially planned to do it later. But I have added support now looking at what selection operator does

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
siddharthteotia: Done

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
siddharthteotia: Done

siddharthteotia:
@Jackie-Jiang , thanks for reviewing. Addressed your latest comments. We can merge this once "code yellow" ends.

siddharthteotia:
@Jackie-Jiang , @mcvsubbu , @mayankshriv , I have rebased on Jackie's latest order by change. Can we please merge this now that we are out of code-yellow?

PR #4536, opened 2019-08-19
[Instance Assignment] Rename instance level replica to replica-group

Jackie-Jiang:
To reduce the confusion of the code:
- Replica-group: a set of instances that serves one replica of all the segments
- Replica: one copy of the segment


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/assignment/InstancePartitions.java
```diff
@@ -33,19 +33,21 @@
 /**
  * Instance (server) partitions for the table.
  *
- * <p>The instance partitions is stored as a map from partition of the format: {@code <partitionId>_<replicaId>} to
+ * <p>The instance partitions is stored as a map from partition of the format: {@code <partitionId>_<replicaGroupId>} to
  * list of server instances, and is persisted under the ZK path: {@code <cluster>/PROPERTYSTORE/INSTANCE_PARTITIONS}.
+ * <p>When partition is not enabled, all instances will be stored as partition 0.
```
mcvsubbu: Suggest : s/partition/column value based partitioning/
- Also, clarify the znode name. Is it raw table name or typed table name?
Need to clarify somewhere that this has nothing to do with stream partitions.

Jackie-Jiang: Added javadoc to clarify these

PR #4537, opened 2019-08-19
Fixing metrics aggregation logic for Float and Double

icefury71:
There seems to an issue with metrics aggregation for float and double type columns. The current code seems to ignore the value of the row being ingested (and simply add the existing cumulative total). This PR fixes this minor bug.


icefury71:
Added a unit test for Float type metric

PR #4538, opened 2019-08-19
Fix bug in updating of metrics of float/double type in realtime

mayankshriv:
The current code has a bug where it is not updating the metrics of
float/double types correctly. This PR fixes the bug.


mayankshriv:
Closing as another fix was filed in #4537

PR #4539, opened 2019-08-19
[TE] Make HTTPS redirection configurable

jihaozh:
Make HTTPS redirection configurable. By default set to false.

### DIFF
--- a/thirdeye/thirdeye-frontend/config/environment.js
+++ b/thirdeye/thirdeye-frontend/config/environment.js
@@ -23,6 +23,8 @@ module.exports = function(environment) {
       directory: 'app/mirage'
     },
 
+    https_only: false,
+
     moment: {
       includeTimezone: 'all'
     },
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-frontend/config/environment.js
```diff
@@ -23,6 +23,8 @@ module.exports = function(environment) {
       directory: 'app/mirage'
     },
 
+    https_only: false,
```
xiaohui-sun: Will this disable https by default?

jihaozh: disable this by default in the open-source project otherwise opensource version will return an error when opening ThirdEye. I will config it in thirdeye-configs.

xiaohui-sun: I see. Thanks.

PR #4540, opened 2019-08-20
Fixing the issue of selection order by won't give correct ordering

fx19880617:
Current behavior:
Selection Order by query will always order the results based on the selection list sequence not the columns provided.

E.g. for query:
```
SELECT a,b,c FROM mytable ORDER BY c
```
The results will be ordered by a not c.

Sample bad queries from quickstart:
```
select AtBatting,G_old from baseballStats order by G_old desc ,hits desc limit 10
```


fx19880617:
I don't think we move column sequence now. Server data table is built based on selection column list

Jackie-Jiang:
@fx19880617 Check SelectionOperatorUtil.extractDataSchema, sort columns are put in front of the other columns. But seems @kishoreg recently re-write the SelectionOperator and that code is not invoked. I am not sure about the current behavior.

fx19880617:
Server side will return a data table with schema. broker side reduce logic should be agnostic to the ordering of schema from server side.

Jackie-Jiang:
If server side guarantees that the sorted columns are always put in front of other columns, then broker side does not need this extra index (where the value for each index is just the index) which will add overhead.
I tried your new test on the existing code and it passed, so I don't understand what is the purpose of this fix.

fx19880617:
Current behavior keeps the column ordering from selection list and not put ordered columns in front.

fx19880617:
The selection test query comparison logic only check the record existence, so it won't recognize the disordered behavior. You can run the sample query in PR description in quickstart to see the behavior.

Jackie-Jiang:
In that case, can you please make the selection test query comparison check the order? With that we can prevent breaking the order in the future.

pinot-core/src/main/java/org/apache/pinot/core/query/selection/SelectionOperatorUtils.java
```diff
@@ -67,14 +67,13 @@
...
 public class SelectionOperatorUtils {
-  private SelectionOperatorUtils() {
-  }
-
   private static final String INT_PATTERN = "##########";
   private static final String LONG_PATTERN = "####################";
   private static final String FLOAT_PATTERN = "#########0.0####";
   private static final String DOUBLE_PATTERN = "###################0.0#########";
   private static final DecimalFormatSymbols DECIMAL_FORMAT_SYMBOLS = DecimalFormatSymbols.getInstance(Locale.US);
+  private SelectionOperatorUtils() {
```
Jackie-Jiang: (nit) let's not move this as all the utils classes put the private constructor at the top

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterIntegrationTestUtils.java
```diff
@@ -766,11 +772,25 @@ public static void testQuery(@Nonnull String pinotQuery, @Nonnull String queryFo
           }
           String actualValue = actualValueBuilder.toString();
 
-          // Check actual value in expected values set
-          if (!expectedValues.contains(actualValue)) {
-            String failureMessage = "Selection result returned in Pinot but not in H2: " + actualValue;
-            failure(pinotQuery, sqlQueries, failureMessage);
-            return;
+          final BrokerRequest brokerRequest = PinotQueryParserFactory.get("pql").compileToBrokerRequest(pinotQuery);
```
Jackie-Jiang: (nit) We don't usually put final in method

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterIntegrationTestUtils.java
```diff
@@ -766,11 +772,25 @@ public static void testQuery(@Nonnull String pinotQuery, @Nonnull String queryFo
...
+          final List<String> selectionColumns = brokerRequest.getSelections().getSelectionColumns();
+          final Set<String> orderByColumns = getOrderByColumns(brokerRequest.getOrderBy());
```
Jackie-Jiang: Please use `brokerRequest.getSelections().getSelectionSortSequence()` as the orderByColumns

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterIntegrationTestUtils.java
```diff
@@ -766,11 +772,25 @@ public static void testQuery(@Nonnull String pinotQuery, @Nonnull String queryFo
...
+
+          final List<SelectionSort> orderBy = brokerRequest.getOrderBy();
+          if ((orderBy != null) && (!orderBy.isEmpty()) && selectionColumns.containsAll(orderByColumns))  {
```
Jackie-Jiang: Why do we need the selectionColumns.containsAll(orderByColumns)?

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterIntegrationTestUtils.java
```diff
@@ -766,11 +772,25 @@ public static void testQuery(@Nonnull String pinotQuery, @Nonnull String queryFo
...
+            // Check actual value in expected values set
+            if (!expectedValues.get(rowIndex).equals(actualValue)) {
```
Jackie-Jiang: We need to handle the tie case

fx19880617: it's impossible from the results. E.g. `select a from table, order by b,c`, if the tie results more than limit size, then we don't event know the ordering.

The only thing we can determine this is that order by columns are the same as selection columns.

PR #4545, opened 2019-08-20
Add APIs to get leader for all tables or a given table

jackjlli:
This PR adds APIs to get leader for all tables or a given table.

* If lead controller resource is enabled, returns the map of partition name and the relevant lead controller response, including lead controller instanceId and a list of table names that belongs to the same partition. 
* If lead controller resource is disabled, returns an empty response.

### Get leaders for all tables:
API GET /leader/tables
When resource enabled:
```
{
  "leadControllerResourceEnabled": true,
  "leadControllerEntryMap": {
    "leadControllerResource_0": {
      "tableNames": [
        "testTable_OFFLINE"
      ],
      "leadControllerId": "Controller_172.25.124.150_9000"
    },
    ...
    "leadControllerResource_23": {
      "tableNames": [],
      "leadControllerId": "Controller_172.25.124.150_9000"
    }
  }
}
```
When resource disabled:
```
{
  "leadControllerResourceEnabled": false,
  "leadControllerEntryMap": {}
}
``` 
### Get leader for a given table:
API GET /leader/tables/{tableName}
When resource enabled:
```
{
  "leadControllerResourceEnabled": true,
  "leadControllerEntryMap": {
    "leadControllerResource_7": {
      "tableNames": [
        "baseballStats_OFFLINE"
      ],
      "leadControllerId": "Controller_172.25.124.150_9000"
    }
  }
}
```
When resource disabled:
```
{
  "leadControllerResourceEnabled": false,
  "leadControllerEntryMap": {}
}
```

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
@@ -18,13 +18,17 @@
  */
 package org.apache.pinot.common.utils.helix;
 
+import org.apache.helix.AccessOption;
+import org.apache.helix.BaseDataAccessor;
 import org.apache.helix.HelixDataAccessor;
 import org.apache.helix.HelixManager;
 import org.apache.helix.PropertyKey;
+import org.apache.helix.ZNRecord;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.pinot.common.utils.CommonConstants.Helix;
 import org.apache.pinot.common.utils.HashUtil;
 import org.apache.pinot.common.utils.StringUtil;
+import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -68,12 +72,39 @@ public class LeadControllerUtils {
 
   /**
    * Checks from ZK if resource config of leadControllerResource is enabled.
+   * @param helixManager helix manager
    */
   public static boolean isLeadControllerResourceEnabled(HelixManager helixManager) {
-    HelixDataAccessor helixDataAccessor = helixManager.getHelixDataAccessor();
-    PropertyKey propertyKey = helixDataAccessor.keyBuilder().resourceConfig(Helix.LEAD_CONTROLLER_RESOURCE_NAME);
-    ResourceConfig resourceConfig = helixDataAccessor.getProperty(propertyKey);
-    String resourceEnabled = resourceConfig.getSimpleConfig(Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY);
-    return Boolean.parseBoolean(resourceEnabled);
+    try {
+      HelixDataAccessor helixDataAccessor = helixManager.getHelixDataAccessor();
+      PropertyKey propertyKey = helixDataAccessor.keyBuilder().resourceConfig(Helix.LEAD_CONTROLLER_RESOURCE_NAME);
+      ResourceConfig resourceConfig = helixDataAccessor.getProperty(propertyKey);
+      String resourceEnabled = resourceConfig.getSimpleConfig(Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY);
+      return Boolean.parseBoolean(resourceEnabled);
+    } catch (Exception e) {
+      LOGGER.warn("Could not get whether lead controller resource is enabled or not.", e);
+      return false;
+    }
+  }
+
+  /**
+   * Gets Helix leader in the cluster. Null if there is no leader.
+   * @param helixManager helix manager
+   * @return instance id of Helix cluster leader, e.g. localhost_9000.
+   */
+  public static String getHelixClusterLeader(HelixManager helixManager) {
+    BaseDataAccessor<ZNRecord> dataAccessor = helixManager.getHelixDataAccessor().getBaseDataAccessor();
+    Stat stat = new Stat();
+    try {
+      ZNRecord znRecord = dataAccessor.get("/" + helixManager.getClusterName() + "/CONTROLLER/LEADER", stat,
+          AccessOption.THROW_EXCEPTION_IFNOTEXIST);
+      String helixLeader = znRecord.getId();
+      LOGGER.info("Getting Helix leader: {} as per znode version {}, mtime {}", helixLeader, stat.getVersion(),
+          stat.getMtime());
+      return helixLeader;
+    } catch (Exception e) {
+      LOGGER.warn("Could not locate Helix leader!", e);
+      return null;
+    }
   }
 }
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
@@ -73,7 +73,7 @@ public class LeadControllerUtils {
     HelixDataAccessor helixDataAccessor = helixManager.getHelixDataAccessor();
     PropertyKey propertyKey = helixDataAccessor.keyBuilder().resourceConfig(Helix.LEAD_CONTROLLER_RESOURCE_NAME);
     ResourceConfig resourceConfig = helixDataAccessor.getProperty(propertyKey);
-    String enableResource = resourceConfig.getSimpleConfig(Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY);
-    return Boolean.parseBoolean(enableResource);
+    String resourceEnabled = resourceConfig.getSimpleConfig(Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY);
+    return Boolean.parseBoolean(resourceEnabled);
   }
 }
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
@@ -93,7 +93,8 @@ public class LeadControllerUtils {
       ZNRecord znRecord = dataAccessor.get("/" + helixManager.getClusterName() + "/CONTROLLER/LEADER", stat,
           AccessOption.THROW_EXCEPTION_IFNOTEXIST);
       String helixLeader = znRecord.getId();
-      LOGGER.info("Getting Helix leader: {} as per znode version {}, mtime {}", helixLeader, stat.getVersion(),
+      // It's ok to log the info since this method isn't called very often.
+      LOGGER.info("Getting Helix leader: {} as per ZNode version {}, mtime {}", helixLeader, stat.getVersion(),
           stat.getMtime());
       return helixLeader;
     } catch (Exception e) {
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
@@ -20,9 +20,8 @@ package org.apache.pinot.common.utils.helix;
 
 import org.apache.helix.AccessOption;
 import org.apache.helix.BaseDataAccessor;
-import org.apache.helix.HelixDataAccessor;
+import org.apache.helix.ConfigAccessor;
 import org.apache.helix.HelixManager;
-import org.apache.helix.PropertyKey;
 import org.apache.helix.ZNRecord;
 import org.apache.helix.model.ResourceConfig;
 import org.apache.pinot.common.utils.CommonConstants.Helix;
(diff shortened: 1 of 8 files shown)


jackjlli:
@mcvsubbu design doc updated.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
```diff
@@ -0,0 +1,165 @@
...
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("/tables")
+  @ApiOperation(value = "Gets leaders for all tables", notes = "Gets leaders for all tables")
+  public Map<String, LeadControllerResponse> getLeadersForAllTables() {
+    Map<String, LeadControllerResponse> leadControllerResponses = new LinkedHashMap<>();
+    HelixManager helixManager = _pinotHelixResourceManager.getHelixZkManager();
+    // returns an empty map if lead controller resource is disabled.
+    if (!LeadControllerUtils.isLeadControllerResourceEnabled(helixManager)) {
+      return leadControllerResponses;
```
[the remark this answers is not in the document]
jackjlli: Refactored the code to return a boolean flag that shows whether lead controller resource is enabled or not. If yes, return partition leader. Otherwise, returns helix leader.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -58,4 +65,15 @@ public static String generatePartitionName(int partitionId) {
...
   }
+
+  /**
+   * Checks from ZK if resource config of leadControllerResource is enabled.
+   */
+  public static boolean isLeadControllerResourceEnabled(HelixManager helixManager) {
+    HelixDataAccessor helixDataAccessor = helixManager.getHelixDataAccessor();
+    PropertyKey propertyKey = helixDataAccessor.keyBuilder().resourceConfig(Helix.LEAD_CONTROLLER_RESOURCE_NAME);
+    ResourceConfig resourceConfig = helixDataAccessor.getProperty(propertyKey);
+    String enableResource = resourceConfig.getSimpleConfig(Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY);
```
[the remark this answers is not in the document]
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -58,4 +69,42 @@ public static String generatePartitionName(int partitionId) {
...
+   */
+  public static boolean isLeadControllerResourceEnabled(HelixManager helixManager) {
+    try {
+      HelixDataAccessor helixDataAccessor = helixManager.getHelixDataAccessor();
+      PropertyKey propertyKey = helixDataAccessor.keyBuilder().resourceConfig(Helix.LEAD_CONTROLLER_RESOURCE_NAME);
+      ResourceConfig resourceConfig = helixDataAccessor.getProperty(propertyKey);
+      String resourceEnabled = resourceConfig.getSimpleConfig(Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY);
+      return Boolean.parseBoolean(resourceEnabled);
+    } catch (Exception e) {
+      LOGGER.warn("Could not get whether lead controller resource is enabled or not.", e);
```
mcvsubbu: Why are we not throwing exception here? 
At the minimum, we need to indicate in the API that a 'false' return does not mean much, but I prefer we throw exception and let the caller handle it.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
```diff
@@ -0,0 +1,197 @@
...
+    HelixManager helixManager = _pinotHelixResourceManager.getHelixZkManager();
+    boolean isLeadControllerResourceEnabled = LeadControllerUtils.isLeadControllerResourceEnabled(helixManager);
+
+    String helixLeader = null;
+    // Returns helix leader if lead controller resource is disabled.
+    if (!isLeadControllerResourceEnabled) {
+      helixLeader = LeadControllerUtils.getHelixClusterLeader(helixManager);
+    }
+
+    ExternalView leadControllerResourceExternalView = getLeadControllerResourceExternalView(helixManager);
```
mcvsubbu: Re-thinking, it is better to return an empty map when resource is disabled.  We can decouple helix leadership from this one

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
```diff
@@ -0,0 +1,206 @@
...
+  @Path("/tables")
+  @ApiOperation(value = "Gets leaders for all tables", notes = "Gets leaders for all tables")
+  public LeadControllerResponse getLeadersForAllTables() {
+    Map<String, LeadControllerEntry> leadControllerEntryMap = new LinkedHashMap<>();
+    HelixManager helixManager = _pinotHelixResourceManager.getHelixZkManager();
+    boolean isLeadControllerResourceEnabled;
+    try {
+      isLeadControllerResourceEnabled = LeadControllerUtils.isLeadControllerResourceEnabled(helixManager);
+    } catch (Exception e) {
+      throw new ControllerApplicationException(LOGGER,
```
mcvsubbu: Instantiate ControllerApplicaitonException with `e`

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    // Assigns all the tables to the relevant partitions.
+    List<String> tableNames = _pinotHelixResourceManager.getAllTables();
+    for (String tableName : tableNames) {
+      String rawTableName = TableNameBuilder.extractRawTableName(tableName);
+      int partitionId = LeadControllerUtils.getPartitionIdForTable(rawTableName);
+      String partitionName = LeadControllerUtils.generatePartitionName(partitionId);
+      LeadControllerEntry leadControllerEntry = leadControllerEntryMap.get(partitionName);
+      leadControllerEntry.getTableNames().add(tableName);
```
mcvsubbu: Is there a possibility that leadControllerEntry is null (say, due to some helix issue)? Perhaps better to check for null here. Better, we can add the entry here instead of a separate loop above?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
```diff
@@ -0,0 +1,206 @@
...
+  @ApiOperation(value = "Gets leader for a given table", notes = "Gets leader for a given table")
+  public LeadControllerResponse getLeaderForTable(
+      @ApiParam(value = "Table name", required = true) @PathParam("tableName") String tableName) {
+    Map<String, LeadControllerEntry> leadControllerEntryMap = new HashMap<>();
+    HelixManager helixManager = _pinotHelixResourceManager.getHelixZkManager();
+    boolean isLeadControllerResourceEnabled;
+    try {
+      isLeadControllerResourceEnabled = LeadControllerUtils.isLeadControllerResourceEnabled(helixManager);
+    } catch (Exception e) {
+      throw new ControllerApplicationException(LOGGER,
```
mcvsubbu: include `e` in ControllerApplicationException

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
```diff
@@ -0,0 +1,206 @@
...
+          Response.Status.INTERNAL_SERVER_ERROR);
+    }
+
+    ExternalView leadControllerResourceExternalView = getLeadControllerResourceExternalView(helixManager);
+    String rawTableName = TableNameBuilder.extractRawTableName(tableName);
+    int partitionId = LeadControllerUtils.getPartitionIdForTable(rawTableName);
+    String partitionName = LeadControllerUtils.generatePartitionName(partitionId);
+
+    // Returns controller Id from lead controller resource is enabled, otherwise returns empty map.
+    if (!isLeadControllerResourceEnabled) {
```
mcvsubbu: Move this if block to be before line 121

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -58,4 +69,43 @@ public static String generatePartitionName(int partitionId) {
...
+  public static boolean isLeadControllerResourceEnabled(HelixManager helixManager)
+      throws Exception {
+    try {
+      HelixDataAccessor helixDataAccessor = helixManager.getHelixDataAccessor();
+      PropertyKey propertyKey = helixDataAccessor.keyBuilder().resourceConfig(Helix.LEAD_CONTROLLER_RESOURCE_NAME);
+      ResourceConfig resourceConfig = helixDataAccessor.getProperty(propertyKey);
+      String resourceEnabled = resourceConfig.getSimpleConfig(Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY);
+      return Boolean.parseBoolean(resourceEnabled);
+    } catch (Exception e) {
+      LOGGER.warn("Could not get whether lead controller resource is enabled or not.", e);
```
mcvsubbu: Include the full message in constructing the exception, and log it in the place where we catch it. I guess this can only be Runtime Exception. Include LEAD_CONTRROLLER_RESOURCE_NAME and LEAD_CONTROLLER_RESOURCE_ENABLED_KEY in the message.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
```diff
@@ -161,8 +168,14 @@ public void testDualModeController() {
...
     final LeadControllerManager thirdLeadControllerManager = thirdDualModeController.getLeadControllerManager();
 
-    TestUtils.waitForCondition(aVoid -> !thirdLeadControllerManager.isLeadControllerResourceEnabled(), TIMEOUT_IN_MS,
-        "Lead controller resource should be disabled.");
+    TestUtils.waitForCondition(aVoid -> {
+      try {
+        return !LeadControllerUtils.isLeadControllerResourceEnabled(_helixManager);
+      } catch (Exception e) {
+        // Exception caught.
+        return false;
```
mcvsubbu: Assert false

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
```diff
@@ -214,6 +191,9 @@ private boolean isLeadControllerResourceEnabled() {
    * @param instanceId instance id without any prefix, e.g. localhost_9000
    * */
   private Pair<String, Integer> convertToHostAndPortPair(String instanceId) {
+    if (instanceId == null) {
```
mcvsubbu: ok, we are going to propagate exception  handling further I suppose. For now, this is ok to keep us from changing all code in one go. Maybe add a todo

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
jackjlli: `putIfAbsent` method is used so there's no way that leadControllerEntry is null.

pinot-core/src/main/java/org/apache/pinot/server/realtime/ControllerLeaderLocator.java
jackjlli: TODO added.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/PinotControllerModeTest.java
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
jackjlli: Removed the try catch block since any exception would be caught from the callers.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
jackjlli: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotLeadControllerRestletResource.java
jackjlli: Done.

pinot-common/src/main/java/org/apache/pinot/common/utils/helix/LeadControllerUtils.java
```diff
@@ -58,4 +68,37 @@ public static String generatePartitionName(int partitionId) {
...
+    String resourceEnabled = resourceConfig.getSimpleConfig(Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY);
+    return Boolean.parseBoolean(resourceEnabled);
+  }
+
+  /**
+   * Gets Helix leader in the cluster. Null if there is no leader.
+   * @param helixManager helix manager
+   * @return instance id of Helix cluster leader, e.g. localhost_9000.
+   */
+  public static String getHelixClusterLeader(HelixManager helixManager) {
```
mcvsubbu: Why is this method here and not in ControllerLeaderLocator?

jackjlli: This method can be called by `LeadControllerManager` from Pinot controller to check whether the current controller is helix leader as well. So it's needed to be here for both controller and server.

Issue #4511, opened 2019-08-09
Change the way we compute end-time of realtime consuming segments


mcvsubbu:
Right now, we have a table-level configuration for stream consuming segments, as mentioned in https://pinot.readthedocs.io/en/latest/pluggable_streams.html?highlight=stream%20realtime#requirements-to-support-partition-level-low-level-consumers. One such configuration is "realtime.segment.flush.threshold.time", that indicates how long a consuming segment should be kept alive before committing it.

When we start to consume a segment, we set the segment end time by adding the current to this configured value. If the consuming segment is still alive at that time, we stat executing the completion steps.

If it happens that a pinot server that is consuming a realtime stream partition has to be restarted (e.g for deployment), the data consumed so far is discarded, and the same segment starts to consume again (upon restart) from the offset that it was set to consume in the first place.

The problem is that we set the segment end time as current time plus the confgured value. Given that the segment already has backlogged events waiting for it to consume, a better way to determine the end time would be to set it to the original value that the previous incarnation of the pinot-server set it to.

We can get the original time be looking at creation time of the segment metadata in zookeeper, and add the configured value for the end time.

We will have to take care that the result is not very close to 0 (or less than 0), so I propose to add a configuration (say, "realtime.segment.flush.threshold.minTime", or something like that) so that we always consume for some small minimum amount of time. This can be set to (say) 15m by default.

There is still a risk that we keep restarting the server every 15 minutes, and the segment may never end, but we should reach the row limit in that case, and it should eventually commit.

@npawar may also have some suggestions on improving this further, or better config names and default values.

siddharthteotia:
Will work on this.

mcvsubbu:
PR merged

PR #4546, opened 2019-08-20
[Issue #4511] Set the consumption end time of a segment based on its creation time

mcvsubbu:
We used to set the end time of a consuming segment as the consumption
start time plus the configured time threshold. In case of server restarts,
that increases the time threshold by an amount depending on how long
the consumption has alread happened.

We now base the end time on the segment creation time instead of
consumption start time.


mcvsubbu:
This PR addresses issue #4511

PR #4548, opened 2019-08-21
[TE] correct time format before send query to Vertica

nadcharin:
found incorrect result return from Vertica because time format in query that send to Vertica is incorrect

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/sql/SqlUtils.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/sql/SqlUtils.java
@@ -581,6 +581,15 @@ public class SqlUtils {
     }
   }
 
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/sql/SqlUtils.java
```diff
@@ -581,6 +581,15 @@ private static String timeFormatToMySQLFormat(String timeFormat) {
     }
   }
 
+  private static String timeFormatToVerticaFormat(String timeFormat) {
+    if(timeFormat.contains("mm")){
+        return timeFormat.replace("mm","mi");
```
akshayrai: Can we perform a case-insensitive replace?
https://stackoverflow.com/questions/5054995/how-to-replace-case-insensitive-literal-substrings-in-java

Issue #4551, opened 2019-08-22
Mutable segments should not store index for virtual columns


mcvsubbu:
We are storing forward indices for virtual columns in mutable segments (we have 3 virtual columns by default). Although the memory is allocated off-heap, this takes up N* 3 * 4 bytes for forward index (where N is the max number of rows we expect to consume in a mutable segment). For schemas with a small number of rows, N can be large.

We also create a dictionary  for $docId, whose cardinality ends up being N.

So, a total of 13N bytes can be reduced for each mutable segment. 

For a segment that expects to consume 10M rows, we can reduce 1.3G of memory in realtime consuming segments.

PR #4555, opened 2019-08-23
[Issue #4551] Remove memory allocation for virtual columns in consumi…

mcvsubbu:
…ng segments

Removed allocation of memory for virtual columns in consuming segments.

Removed the getRecord() interface from IndexSegment since it is not used anywhere.
Restricted it to a method within MutableSegmentImpl (needed to build completed
segments from consuming segments).

Verified that segment metadata did not have virtual columns before and after my change.

Verified that we can still execute the following queries:

        SELECT distinctcount($segmentName) from mytable
        SELECT count(*) from mytable group by $segmentName top 50
        SELECT somefield, $docId, $segmentName, $hostName from mytable
        SELECT count(*) from mytable where $segmentName = "mytable__1__23__20190823T0142Z"

and they work fine

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/IndexSegment.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/IndexSegment.java
@@ -71,16 +71,6 @@ public interface IndexSegment {
   List<StarTreeV2> getStarTrees();
 
   /**
-   * Returns the record for the given document Id.
-   * <p>NOTE: don't use this method for high performance code.
-   *
-   * @param docId Document Id
-   * @param reuse Reusable buffer for the record
-   * @return Record for the given document Id
-   */
-  GenericRow getRecord(int docId, GenericRow reuse);
-
-  /**
    * Destroys segment in memory and closes file handlers if in MMAP mode.
    */
   void destroy();
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/IndexSegment.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/IndexSegment.java
@@ -71,6 +71,16 @@ public interface IndexSegment {
   List<StarTreeV2> getStarTrees();
 
   /**
+   * Returns the record for the given document Id. Virtual column values are not returned.
+   * <p>NOTE: don't use this method for high performance code.
+   *
+   * @param docId Document Id
+   * @param reuse Reusable buffer for the record
+   * @return Record for the given document Id
+   */
+  GenericRow getRecord(int docId, GenericRow reuse);
+
+  /**
    * Destroys segment in memory and closes file handlers if in MMAP mode.
    */
   void destroy();
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/immutable/ImmutableSegmentImpl.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/immutable/ImmutableSegmentImpl.java
@@ -158,16 +158,4 @@ public class ImmutableSegmentImpl implements ImmutableSegment {
   public List<StarTreeV2> getStarTrees() {
     return _starTreeIndexContainer != null ? _starTreeIndexContainer.getStarTrees() : null;
   }
-
-  @Override
-  public GenericRow getRecord(int docId, GenericRow reuse) {
-    for (FieldSpec fieldSpec : _segmentMetadata.getSchema().getAllFieldSpecs()) {
-      String column = fieldSpec.getName();
-      ColumnIndexContainer indexContainer = _indexContainerMap.get(column);
-      reuse.putField(column, IndexSegmentUtils
-          .getValue(docId, fieldSpec, indexContainer.getForwardIndex(), indexContainer.getDictionary(),
-              _segmentMetadata.getColumnMetadataFor(column).getMaxNumberOfMultiValues()));
-    }
-    return reuse;
-  }
 }
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/immutable/ImmutableSegmentImpl.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/immutable/ImmutableSegmentImpl.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Set;
 import javax.annotation.Nullable;
 import org.apache.pinot.common.data.FieldSpec;
+import org.apache.pinot.common.data.Schema;
 import org.apache.pinot.core.data.GenericRow;
 import org.apache.pinot.core.indexsegment.IndexSegmentUtils;
 import org.apache.pinot.core.io.reader.DataFileReader;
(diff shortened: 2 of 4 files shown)


pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -442,13 +470,20 @@ private ColumnDataSource getVirtualDataSource(String column) {
...
+   * @param reuse a GenericRow object that will be re-used if provided. Otherwise, this method will allocate a new one
+   * @return Generic row with physical columns of the specified row.
+   */
   public GenericRow getRecord(int docId, GenericRow reuse) {
     for (FieldSpec fieldSpec : _schema.getAllFieldSpecs()) {
       String column = fieldSpec.getName();
-      reuse.putField(column, IndexSegmentUtils
-          .getValue(docId, fieldSpec, _indexReaderWriterMap.get(column), _dictionaryMap.get(column),
-              _maxNumValuesMap.getOrDefault(column, 0)));
+      if (!_schema.isVirtualColumn(column)) {
```
Jackie-Jiang: Might be better to cache the physical fieldSpecs as we call getRecord() on per record level.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/IndexSegment.java
```diff
@@ -70,16 +70,6 @@
...
 
-  /**
-   * Returns the record for the given document Id.
-   * <p>NOTE: don't use this method for high performance code.
-   *
-   * @param docId Document Id
-   * @param reuse Reusable buffer for the record
-   * @return Record for the given document Id
-   */
-  GenericRow getRecord(int docId, GenericRow reuse);
```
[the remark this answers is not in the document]
mcvsubbu: i will add it back with a boolean to get (or not get) virtual columns

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -197,7 +201,18 @@ public long getLatestIngestionTimestamp() {
 
     // Metric aggregation can be enabled only if config is specified, and all dimensions have dictionary,
     // and no metrics have dictionary. If not enabled, the map returned is null.
-    _recordIdMap = enableMetricsAggregationIfPossible(config, _schema, noDictionaryColumns);
+    _recordIdMap = enableMetricsAggregationIfPossible(config, noDictionaryColumns);
+  }
+
+  private int computeNumKeyColumns() {
+    int numKeyColumns = 1;
```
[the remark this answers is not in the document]
jackjlli: +1. And I saw there is more than one field type which is a time column (i.e. TIME and DATE_TIME). Can they be in the same schema of a table?

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -611,15 +646,18 @@ private int getOrCreateDocId(Map<String, Object> dictIdMap) {
 
     // All metric columns should have no-dictionary index.
     // All metric columns must be single value
```
jackjlli: I don't fully understand the statements here if I didn't read the exact code. Can we add some comments on the description of this method to say aggregate metric is only supported for use cases when all the columns are single value?

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/IndexSegment.java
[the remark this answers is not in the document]
mcvsubbu: i kept the api the same and filtered out virtual columns from immutable segments

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/immutable/ImmutableSegmentImpl.java
```diff
@@ -158,16 +158,4 @@ public void destroy() {
   public List<StarTreeV2> getStarTrees() {
     return _starTreeIndexContainer != null ? _starTreeIndexContainer.getStarTrees() : null;
   }
-
-  @Override
-  public GenericRow getRecord(int docId, GenericRow reuse) {
-    for (FieldSpec fieldSpec : _segmentMetadata.getSchema().getAllFieldSpecs()) {
```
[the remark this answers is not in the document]
mcvsubbu: done

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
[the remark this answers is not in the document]
mcvsubbu: Added comment. The earlier code was:
numKeys = 1 + schema.getDimensions().size()

I translated that into a method to exclude the virtual columns. I don't want to add logic regarding date time columns and disrupt the existing code. That should be a separate PR. I will file an issue for it.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
mcvsubbu: done

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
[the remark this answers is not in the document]
mcvsubbu: Issue #4558 created

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
mcvsubbu: The comments at the head of the method specify the limitations.
The additional limitatation that a metric must be single value is not specified, I can move the comment from insie the method to outside of it.
I am not sure what you want me to change/improve here

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
Jackie-Jiang: You forgot this one, the most critical one lol

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
mcvsubbu: moved comments around

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
mcvsubbu: Actually, least critical since this is called only during segment build. Others are called at query time!
But yes, I will add it.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -95,6 +99,7 @@
   private volatile long _minTime = Long.MAX_VALUE;
   private volatile long _maxTime = Long.MIN_VALUE;
   private final int _numKeyColumns;
+  private final Collection<FieldSpec> _physicalColumnFieldSpecs;
```
Jackie-Jiang: Let's cache physical dimensions and physical metrics as well to eliminate all the record level check

mcvsubbu: I am not convinced that this is worth the change. In some places we look for fieldspec, and in others the column names. 
Why don't we get this PR through, and if we find that the 'if' checks are expensive, we can work on other ways to reduce them

Jackie-Jiang: For the places where we get the column name, we use the column name to fetch the field spec. Instead, we can directly cache the field spec.
These if checks won't cause big performance issue, but we are doing redundant work over and over again.

PR #4556, opened 2019-08-23
Support aws s3 with Parquet in pinot-tools

fx19880617:
sample usage to create segments:
```
bin/pinot-admin.sh CreateSegment -dataDir s3a://${aws_access_key}:${aws_secret_access_key}@${s3bucket}/${path}  -outDir /tmp/pinot-data -tableName sampleTable -segmentName sampleTable_0 -schemaFile /tmp/sample_schema.json -overwrite -format parquet
```

### DIFF
--- a/pinot-tools/pom.xml
+++ b/pinot-tools/pom.xml
@@ -56,6 +56,11 @@
     </dependency>
     <dependency>
       <groupId>org.apache.pinot</groupId>
+      <artifactId>pinot-parquet</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.pinot</groupId>
       <artifactId>pinot-connector-kafka-base</artifactId>
       <version>${project.version}</version>
     </dependency>
@@ -97,6 +102,15 @@
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-math3</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-aws</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
     </dependency>
   </dependencies>
   <build>
--- a/pinot-tools/pom.xml
+++ b/pinot-tools/pom.xml
@@ -111,6 +111,12 @@
     <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-annotations</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
   </dependencies>
   <build>
--- a/pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
+++ b/pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
@@ -56,28 +56,40 @@ import org.slf4j.LoggerFactory;
  */
 public class CreateSegmentCommand extends AbstractBaseAdminCommand implements Command {
   private static final Logger LOGGER = LoggerFactory.getLogger(CreateSegmentCommand.class);
-  @Option(name = "-enableStarTreeIndex", usage = "Enable Star Tree Index.")
-  boolean _enableStarTreeIndex = false;
+
   @Option(name = "-generatorConfigFile", metaVar = "<string>", usage = "Config file for segment generator.")
   private String _generatorConfigFile;
+
   @Option(name = "-dataDir", metaVar = "<string>", usage = "Directory containing the data.")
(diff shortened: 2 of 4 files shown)


fx19880617:
^^ @kishoreg just fyi

kishoreg:
Looks like we are pulling two different versions of hadoop-annotations 

Dependency convergence error for org.apache.hadoop:hadoop-annotations:2.2.0 paths to dependency are:
+-org.apache.pinot:pinot-distribution:0.2.0-SNAPSHOT
  +-org.apache.pinot:pinot-orc:0.2.0-SNAPSHOT
    +-org.apache.orc:orc-mapreduce:1.5.2
      +-org.apache.hadoop:hadoop-mapreduce-client-core:2.2.0
        +-org.apache.hadoop:hadoop-annotations:2.2.0
and
+-org.apache.pinot:pinot-distribution:0.2.0-SNAPSHOT
  +-org.apache.pinot:pinot-tools:0.2.0-SNAPSHOT
    +-org.apache.hadoop:hadoop-common:2.7.0
      +-org.apache.hadoop:hadoop-annotations:2.7.0

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -362,20 +357,31 @@ public boolean accept(File dir, String name) {
...
+            config.setInputFilePath(localFile);
             config.setSegmentName(_segmentName + "_" + segCnt);
             config.loadConfigFiles();
-
             final SegmentIndexCreationDriverImpl driver = new SegmentIndexCreationDriverImpl();
-            driver.init(config);
+            switch (config.getFormat()) {
+              case PARQUET:
+                RecordReader parquetRecordReader = new ParquetRecordReader();
+                parquetRecordReader.init(config);
```
npawar: can this not be done via `readerConfig`? why does this one need special handling, when all other formats are using the readerConfigFile?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -362,20 +357,31 @@ public boolean accept(File dir, String name) {
...
+    for (final Path dataFilePath : dataFilePaths) {
       final int segCnt = cnt;
 
       executor.execute(new Runnable() {
         @Override
         public void run() {
           try {
             SegmentGeneratorConfig config = new SegmentGeneratorConfig(segmentGeneratorConfig);
-            config.setInputFilePath(file.getAbsolutePath());
+
```
npawar: Just want to confirm that you've checked that this still works for create segment from local file system..?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
fx19880617: The reason is that `RecordReaderFactory` is in `pinot-core`.  If we want to use it, then  we have to make changes in `pinot-core` and bring in all parquet dependencies to `pinot-core`.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
fx19880617: yes. I've also tested:
```
bin/pinot-admin.sh CreateSegment -dataDir /tmp/sample-pinot-data  -outDir /tmp/pinot-data -tableName sampleTable -segmentName sampleTable_0 -schemaFile /tmp/sample_schema.json -overwrite -format parquet
``

File: pinot-tools/pom.xml [hunk 3]
```diff
@@ -98,6 +103,21 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-math3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
```
snleee: @kishoreg @fx19880617 Doesn't this defeat the purpose on the reason that we took out the separate package for avro, orc, parquet readers and azure filesystem? How do you guys think?

kishoreg: This is adding it only tools right not to core. We definitely need to extract out all things related to format and filesystems out of pinot-core and make it pluggable. But for now, this is the only solution I can think of adding support for different formats.

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -47,40 +56,28 @@
  */
 public class CreateSegmentCommand extends AbstractBaseAdminCommand implements Command {
   private static final Logger LOGGER = LoggerFactory.getLogger(CreateSegmentCommand.class);
-
+  @Option(name = "-enableStarTreeIndex", usage = "Enable Star Tree Index.")
+  boolean _enableStarTreeIndex = false;
   @Option(name = "-generatorConfigFile", metaVar = "<string>", usage = "Config file for segment generator.")
   private String _generatorConfigFile;
-
```
snleee: Can we add back these lines? Without lines, I think they are hard to read.

PR #4557, opened 2019-08-23
Support transform functions with AVG aggregation function

siddharthteotia:
Support queries like:

SELECT AVG(SUB(col1, col2) FROM foo
SELECT AVG(SUB(DIV(col1, col2), DIV(col3, col4))) FROM foo

They were failing in AvgAggregationFunction as TransformBlockValSet() did not implement getValueType() method

Also added operand checks for transform functions for acceptable data types

Testing:

Added TransformQueriesTest to test both inner-segment and inter-segment, inter-server queries with custom data

### DIFF
--- a/pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
+++ b/pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
@@ -1,0 +1,286 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.pinot.queries;
+
+import java.io.File;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.io.FileUtils;
+import org.apache.pinot.common.data.DimensionFieldSpec;
+import org.apache.pinot.common.data.FieldSpec;
+import org.apache.pinot.common.data.MetricFieldSpec;
+import org.apache.pinot.common.data.Schema;
+import org.apache.pinot.common.data.TimeFieldSpec;
+import org.apache.pinot.common.data.TimeGranularitySpec;
+import org.apache.pinot.common.response.broker.AggregationResult;
+import org.apache.pinot.common.response.broker.BrokerResponseNative;
+import org.apache.pinot.common.segment.ReadMode;
+import org.apache.pinot.common.utils.time.TimeUtils;
+import org.apache.pinot.core.data.GenericRow;
+import org.apache.pinot.core.data.manager.SegmentDataManager;
+import org.apache.pinot.core.data.manager.offline.ImmutableSegmentDataManager;
+import org.apache.pinot.core.data.readers.GenericRowRecordReader;
+import org.apache.pinot.core.data.readers.RecordReader;
+import org.apache.pinot.core.indexsegment.IndexSegment;
+import org.apache.pinot.core.indexsegment.generator.SegmentGeneratorConfig;
+import org.apache.pinot.core.indexsegment.immutable.ImmutableSegment;
+import org.apache.pinot.core.indexsegment.immutable.ImmutableSegmentLoader;
+import org.apache.pinot.core.operator.blocks.IntermediateResultsBlock;
+import org.apache.pinot.core.operator.query.AggregationOperator;
+import org.apache.pinot.core.query.aggregation.function.customobject.AvgPair;
+import org.apache.pinot.core.segment.creator.impl.SegmentIndexCreationDriverImpl;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class TransformQueriesTest extends BaseQueriesTest {
+
+  private static final String D1 = "STRING_COL";
+  private static final String M1 = "INT_COL1";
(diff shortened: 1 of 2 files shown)


siddharthteotia:
@Jackie-Jiang , the reason for creating a separate test file is because InnerSegment/InterSegment queries test extend BaseSingleValueQueriesTest and use the entire avro dataset to build test segments.

It is even difficult to predict if the queries are returning correct results since we don't know what data is there in avro files.

So I created a separate test that just extends BaseQueriesTest to get helper methods and created 2 segments with small data to do both inner and inter segment tests and verify correctness

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -0,0 +1,286 @@
...
+
+  private static final File INDEX_DIR = new File(FileUtils.getTempDirectory(), "TransformQueriesTest");
+
+  private Schema _schema;
+  private List<IndexSegment> _indexSegments = new ArrayList<>();
+  private List<SegmentDataManager> _segmentDataManagers;
+
+  @BeforeClass
+  public void setup() {
+    _schema = new Schema();
```
Jackie-Jiang: Use Schema.Builder to build the schema

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -0,0 +1,286 @@
...
+  private static final String SEGMENT_NAME_2 = "FOO_SEGMENT_2";
+
+  private static final File INDEX_DIR = new File(FileUtils.getTempDirectory(), "TransformQueriesTest");
+
+  private Schema _schema;
+  private List<IndexSegment> _indexSegments = new ArrayList<>();
+  private List<SegmentDataManager> _segmentDataManagers;
+
+  @BeforeClass
+  public void setup() {
```
Jackie-Jiang: (nit) We usually use setUp()

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -0,0 +1,286 @@
...
+
+    final MetricFieldSpec m4Spec = new MetricFieldSpec(M4, FieldSpec.DataType.LONG);
+    _schema.addField(m4Spec);
+
+    final TimeFieldSpec tSpec = new TimeFieldSpec(new TimeGranularitySpec(FieldSpec.DataType.LONG, TimeUnit.MILLISECONDS, TIME));
+    _schema.addField(tSpec);
+  }
+
+  @AfterClass
+  public void destroy() {
```
Jackie-Jiang: (nit) We usually use tearDown()

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+    _indexSegments.clear();
+  }
+
+  @Test
+  public void testTransformWithAvgInnerSegment() throws Exception {
+    try {
+      final List<GenericRow> rows = createDataSet(10);
+
+      final RecordReader recordReader = new GenericRowRecordReader(rows, _schema);
```
Jackie-Jiang: Resource leak, use try-with resource

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  @Test
+  public void testTransformWithAvgInterSegmentInterServer() throws Exception {
+    try {
+      final List<GenericRow> segmentOneRows = createDataSet(10);
+      final List<GenericRow> segmentTwoRows = createDataSet(10);
+
+      final RecordReader recordReaderOne = new GenericRowRecordReader(segmentOneRows, _schema);
+      final RecordReader recordReaderTwo = new GenericRowRecordReader(segmentTwoRows, _schema);
```
Jackie-Jiang: Resource leak, use try-with resource

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -0,0 +1,286 @@
...
+    return _indexSegments.get(0);
+  }
+
+  @Override
+  protected List<SegmentDataManager> getSegmentDataManagers() {
+    return _segmentDataManagers;
+  }
+
+  private void createSegment(
+      Schema schema,
```
Jackie-Jiang: Please reformat the file with the Pinot Style

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -0,0 +1,286 @@
...
+    final TimeFieldSpec tSpec = new TimeFieldSpec(new TimeGranularitySpec(FieldSpec.DataType.LONG, TimeUnit.MILLISECONDS, TIME));
+    _schema.addField(tSpec);
+  }
+
+  @AfterClass
+  public void destroy() {
+    FileUtils.deleteQuietly(INDEX_DIR);
+  }
+
+  private void delete() {
```
Jackie-Jiang: Rename to destroySegments() for clarity

siddharthteotia: done

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
siddharthteotia: done

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
siddharthteotia: Changed to try with resource but in this case we were not leaking resource as record reader.close() is a NO-OP

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
siddharthteotia: Changed to try with resource but in this case we were not leaking resource as record reader.close() is a NO-OP

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
siddharthteotia: done

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
siddharthteotia: done

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
siddharthteotia: done

siddharthteotia:
Addressed the comments

PR #4560, opened 2019-08-23
Refactor toggleInstanceState API

jackjlli:
This PR:
* refactors the inner logic of toggleInstanceState API.
* adds error message in the response message.

Existing error in TravisCI:
```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 38.034 sec <<< FAILURE! - in org.apache.pinot.controller.helix.ControllerInstanceToggleTest
testInstanceToggle(org.apache.pinot.controller.helix.ControllerInstanceToggleTest)  Time elapsed: 17.424 sec  <<< FAILURE!
java.io.IOException: Server returned HTTP response code: 500 for URL: http://localhost:18998/instances/Server_localhost_0/state
    at org.apache.pinot.controller.helix.ControllerInstanceToggleTest.testInstanceToggle(ControllerInstanceToggleTest.java:78)
```

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -2175,6 +2175,9 @@ public class PinotHelixResourceManager {
             for (String state : currentState.getPartitionStateMap().values()) {
               // If instance is enabled, all the partitions should not eventually be offline.
               // If instance is disabled, all the partitions should eventually be offline.
+              // TODO: Handle the case when realtime segments are in OFFLINE state because there're some problem with realtime segment consumption,
+              //  and realtime segment will mark itself as OFFLINE in ideal state.
+              //  Issue: https://github.com/apache/incubator-pinot/issues/4653
               if ((enableInstance && !offlineState.equals(state)) || (!enableInstance && offlineState.equals(state))) {
                 toggleSucceeded = false;
                 break;
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -2152,14 +2152,19 @@ public class PinotHelixResourceManager {
     _helixAdmin.enableInstance(_helixClusterName, instanceName, toggle);
     long deadline = System.currentTimeMillis() + 1000 * timeOutInSeconds;
     boolean toggleSucceed;
-    String beforeToggleStates =
-        (toggle) ? SegmentOnlineOfflineStateModel.OFFLINE : SegmentOnlineOfflineStateModel.ONLINE;
+    String offlineState = SegmentOnlineOfflineStateModel.OFFLINE;
 
     while (System.currentTimeMillis() < deadline) {
       toggleSucceed = true;
       PropertyKey liveInstanceKey = _keyBuilder.liveInstance(instanceName);
       LiveInstance liveInstance = _helixDataAccessor.getProperty(liveInstanceKey);
-      if (liveInstance != null) {
+      if (liveInstance == null) {
+        if (!toggle) {
+          // If we disable the instance, we actually don't care whether live instance being null. Thus, returning success should be good.
+          // Otherwise, wait for at most 10 seconds.
+          return PinotResourceManagerResponse.SUCCESS;
+        }
+      } else {
         // Checks all the current states fall into the target states
         PropertyKey instanceCurrentStatesKey = _keyBuilder.currentStates(instanceName, liveInstance.getSessionId());
         List<CurrentState> instanceCurrentStates = _helixDataAccessor.getChildValues(instanceCurrentStatesKey);
@@ -2168,7 +2173,9 @@ public class PinotHelixResourceManager {
         } else {
           for (CurrentState currentState : instanceCurrentStates) {
             for (String state : currentState.getPartitionStateMap().values()) {
-              if (beforeToggleStates.equals(state)) {
+              // If instance is enabled, all the partitions should not eventually be offline.
+              // If instance is disabled, all the partitions should eventually be offline.
+              if ((toggle && !offlineState.equals(state)) || (!toggle && offlineState.equals(state))) {
                 toggleSucceed = false;
                 break;
               }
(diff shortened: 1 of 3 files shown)


mcvsubbu:
I didn't get why liveinstance can be null. Can you add some comments in that regard in the test?

pinot-controller/src/test/java/org/apache/pinot/controller/helix/ControllerInstanceToggleTest.java
```diff
@@ -104,9 +107,21 @@ public void testInstanceToggle()
...
+  private void toggleInstanceState(String instanceName, String state) {
+    TestUtils.waitForCondition(aVoid -> {
+      try {
+        sendPostRequest(_controllerRequestURLBuilder.forInstanceState(instanceName), state);
+      } catch (IOException ioe) {
+        // receive non-200 status code
+        return false;
+      }
+      return true;
+    }, TIMEOUT_MS, "Failed to toggle instance state: \'" + state + "\' for instance: " + instanceName);
```
Jackie-Jiang: Why are you logging `\` lol

jackjlli:
Actually from the error message, it shouldn't be the case when live instance being null caused the failure. Because it failed at `disabling` the instance. 
It could be due to Helix slowness on updating the current state of all the segments.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/ControllerInstanceToggleTest.java
jackjlli: I'm logging `'` instead of `\`. Here's the simple program:
```
  public static void main(String[] args) {
    System.out.println("\'");
    System.out.println("'");
  }
```
Output:
```
'
'
```

Jackie-Jiang: Good to learn. While that is for escaping quote purpose, which is not needed here.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -2142,48 +2142,59 @@ public PinotResourceManagerResponse dropInstance(String instanceName) {
...
+        // Checks all the current states fall into the target states
+        PropertyKey instanceCurrentStatesKey = _keyBuilder.currentStates(instanceName, liveInstance.getSessionId());
+        List<CurrentState> instanceCurrentStates = _helixDataAccessor.getChildValues(instanceCurrentStatesKey);
+        if (instanceCurrentStates == null) {
+          return PinotResourceManagerResponse.SUCCESS;
+        } else {
+          for (CurrentState currentState : instanceCurrentStates) {
+            for (String state : currentState.getPartitionStateMap().values()) {
+              // If instance is enabled, all the partitions should not eventually be offline.
+              // If instance is disabled, all the partitions should eventually be offline.
```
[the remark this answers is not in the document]
mcvsubbu: That is true that OFFLINE can be a good state. The only way to check that is if we also pull the idealstate and do not compare segments that may be in OFFLINE state.

For this iteration, since it at least takes care of CONSUMING state, I am fine adding a TODO here for enhancement.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/ControllerInstanceToggleTest.java
mcvsubbu: I found it useful to quote user input so that we can identify cases where the user may have input a bunch of spaces or null string. In that case, the log will show **Failed to toggle instance state: ' '** as opposed to **Failed to toggle instance state: **. 
But yes, we should not need the back-slash

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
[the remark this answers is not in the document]
jackjlli: Just created an issue with the TODO description.
https://github.com/apache/incubator-pinot/issues/4653

pinot-controller/src/test/java/org/apache/pinot/controller/helix/ControllerInstanceToggleTest.java
jackjlli: Back-slash removed.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -2140,52 +2140,71 @@ public PinotResourceManagerResponse dropInstance(String instanceName) {
...
    *
    * @param instanceName: Name of Instance for which the status needs to be toggled.
-   * @param toggle: 'True' for ONLINE 'False' for OFFLINE.
-   * @param timeOutInSeconds: Time-out for setting ideal-state.
+   * @param enableInstance: 'True' for enabling the instance and 'False' for disabling the instance.
+   * @param timeOutMs: Time-out for setting ideal-state.
    * @return
    */
-  public PinotResourceManagerResponse toggleInstance(String instanceName, boolean toggle, int timeOutInSeconds) {
+  private PinotResourceManagerResponse enableInstance(String instanceName, boolean enableInstance, long timeOutMs) {
```
mcvsubbu: Not sure why you changed the name of the method to enableInstance. I would keep the method name the same. It takes a boolean that can  be either true or false.

Jackie-Jiang: By convention, toggle means ENABLE/DISABLE/DROP (Check PinotInstanceRestletResource)

jackjlli: In addition, for dropping instance there's already a method called `dropInstance(instanceName)` in `PinotHelixResourceManager`. 
So it's be good to name this method `enableInstance`.

mcvsubbu: I still think the method name should be operation-neutral since it does both. Since 2 out of 3 agree and this is not a very big issue, I will give in.
Some alternative names : "toggleInstance" (like before) or "enableOrDisableInstance()"

Jackie-Jiang: Actually this is consistent with `HelixAdmin.enableInstance()`

PR #4562, opened 2019-08-24
Handle the virtual column logic in FieldSpec instead of Schema

Jackie-Jiang:
Virtual column should be tied to FieldSpec, and should be determined by whether the virtualColumnProvider is configured
E.g. A column starts with '$' but without virtualColumnProvider should not be considered a virtual column because it cannot be initialized

Also fix a bug in MutableSegmentImpl where virtual columns are added to the physical dimensions/metrics list


pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
```diff
@@ -153,6 +154,11 @@ public void setVirtualColumnProvider(String virtualColumnProvider) {
     _virtualColumnProvider = virtualColumnProvider;
   }
 
+  @JsonIgnore
+  public boolean isVirtualColumn() {
+    return _virtualColumnProvider != null && !_virtualColumnProvider.isEmpty();
```
mcvsubbu: Are we allowing column names to start with $ for physical columns as well? Or, are we allowing virtual  column names to start with characters other than "$"? I did not see checks for either of these (so happens that the in0built virtual columns are starting with dollar sign). Would be good to add a comment here about that.

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplAggregateMetricsTest.java
```diff
@@ -51,15 +53,21 @@
...
-            .addMetric(METRIC, FieldSpec.DataType.LONG)
-            .addMetric(METRIC_2, FieldSpec.DataType.FLOAT)
-            .addTime(TIME_COLUMN, TimeUnit.DAYS, FieldSpec.DataType.INT)
-            .build();
+        .addSingleValueDimension(DIMENSION_2, FieldSpec.DataType.STRING).addMetric(METRIC, FieldSpec.DataType.LONG)
+        .addMetric(METRIC_2, FieldSpec.DataType.FLOAT).addTime(TIME_COLUMN, TimeUnit.DAYS, FieldSpec.DataType.INT)
+        .build();
+    // Add virtual columns, which should not be aggregated
+    DimensionFieldSpec virtualDimensionFieldSpec =
+        new DimensionFieldSpec("$virtualDimension", FieldSpec.DataType.INT, true, Object.class);
```
mcvsubbu: nice, thanks for catching the bug and adding this test.

pinot-common/src/main/java/org/apache/pinot/common/data/FieldSpec.java
Jackie-Jiang: Virtual column works only when the provider is provided. Added comments about that

PR #4573, opened 2019-08-30
Merge LLC related integration tests

Jackie-Jiang:
Clean up LLCRealtimeClusterIntegrationTest
Merge LLCRealtimeClusterSplitCommitIntegrationTest and LLCRealtimeClusterWithEnabledLeadControllerResourceIntegrationTest into LLCRealtimeClusterIntegrationTest to reduse the runtime of integration tests


pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/LLCRealtimeClusterIntegrationTest.java
```diff
@@ -19,58 +19,56 @@
...
-
-  public static final String CONSUMER_DIRECTORY = "/tmp/consumer-test";
-  public static final long RANDOM_SEED = System.currentTimeMillis();
-  public static final Random RANDOM = new Random(RANDOM_SEED);
-
-  public final boolean _isDirectAlloc = RANDOM.nextBoolean();
-  public final boolean _isConsumerDirConfigured = RANDOM.nextBoolean();
-  private final long _startTime = System.currentTimeMillis();
-
+  private static final String CONSUMER_DIRECTORY = "/tmp/consumer-test";
```
jenniferdai: Do we need to override the getServer method if split commit is enabled? Otherwise the server will not start with the config

Jackie-Jiang: We don't because getServer() will call overrideServerConf() where we override the config

Jackie-Jiang:
@mcvsubbu Once we have the module level test for split commit and lead controller, we can remove the random in this test. This PR can reduce the test time by nearly 10 minutes.

mcvsubbu:
> @mcvsubbu Once we have the module level test for split commit and lead controller, we can remove the random in this test. This PR can reduce the test time by nearly 10 minutes.

ok

PR #4578, opened 2019-09-04
Refactor TransformBlockDataFetcher so that DISTINCT and reuse it

Jackie-Jiang:
1. Refactor TransformBlockDataFetcher to take only the BlockValSets
2. Remove the redundant dictionary handling logic (already handled in DataFetcher)
3. Refactor SelectionOperator to simplify the logic, and only parse selection columns once
4. Move the sort columns (expressions) into the front of the data schema
5. Make MapValueTransformFunction able to transform values

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/query/SelectionOperator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/query/SelectionOperator.java
@@ -19,19 +19,16 @@
 package org.apache.pinot.core.operator.query;
 
 import java.io.Serializable;
-import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.PriorityQueue;
-import org.apache.pinot.common.data.FieldSpec.DataType;
 import org.apache.pinot.common.request.Selection;
 import org.apache.pinot.common.request.SelectionSort;
 import org.apache.pinot.common.request.transform.TransformExpressionTree;
 import org.apache.pinot.common.utils.DataSchema;
+import org.apache.pinot.common.utils.DataSchema.ColumnDataType;
 import org.apache.pinot.common.utils.primitive.ByteArray;
 import org.apache.pinot.core.common.BlockValSet;
 import org.apache.pinot.core.indexsegment.IndexSegment;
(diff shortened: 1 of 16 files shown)


pinot-core/src/main/java/org/apache/pinot/core/operator/query/SelectionOperator.java
```diff
@@ -42,212 +39,168 @@
...
-    _blockValSets = new BlockValSet[_expressions.size()];
-    _expressionResultMetadata = new TransformResultMetadata[_expressions.size()];
-    _dictionaries = new Dictionary[_expressions.size()];
-    DataSchema.ColumnDataType[] columnDataTypes = new DataSchema.ColumnDataType[_expressions.size()];
-    String[] columnNames = new String[_expressions.size()];
-    for (int i = 0; i < _expressions.size(); i++) {
+    int numExpressions = _expressions.size();
+    _blockValSets = new BlockValSet[numExpressions];
+    String[] columnNames = new String[numExpressions];
+    ColumnDataType[] columnDataTypes = new ColumnDataType[numExpressions];
```
siddharthteotia: I feel the changes here will break the ordering of columns -- TransformOperator has compiled expression trees as as list but that is given to it by TransformPlanNode which constructs them in a set (thus we lose the order as specified in the query). This is why here in the operator we build the list of expressions in the order as specified in the query and fetch block for each in the same order.

Looks like the new changes (although better that they look at the selection columns only once and directly work on the expressions from TransformOperator), we will probably project the columns in an order different than what was specified by the user in query.

This is the reason why existing implementation was getting the list of expressions from TransformOperator but still constructing its local list based on the ordering of columns in the query

pinot-core/src/main/java/org/apache/pinot/core/operator/query/SelectionOperator.java
```diff
@@ -42,212 +39,168 @@
...
-            return ret;
-          }
+        Serializable v1 = o1[i];
+        Serializable v2 = o2[i];
+
+        int ret = 0;
+        SelectionSort selectionSort = _sortSequence.get(i);
+        switch (_blockValSets[i].getValueType()) {
+          case INT:
+            if (!selectionSort.isIsAsc()) {
```
mayankshriv: Nit: can it not be replaced by one-liner:
`ret = (!selectionSort.isAsc()) ? ... : ...`

Jackie-Jiang:
This PR is split into multiple PRs and the one for TransformBlockDataFetcher: #4658

PR #4583, opened 2019-09-04
Support default value for Byte column

jackjlli:
This PR supports default value for Byte column.

Unit tests added.


pinot-common/src/test/java/org/apache/pinot/common/data/FieldSpecTest.java
```diff
@@ -117,6 +117,17 @@ public void testFieldSpec() {
     Assert.assertEquals(fieldSpec1.toString(), fieldSpec2.toString());
     Assert.assertEquals(fieldSpec1.hashCode(), fieldSpec2.hashCode());
     Assert.assertEquals(fieldSpec1.getDefaultNullValue(), 1L);
+
+    // Metric field with default null value for byte column.
+    fieldSpec1 = new MetricFieldSpec();
+    fieldSpec1.setName("btyeMetric");
```
siddharthteotia: typo?

pinot-core/src/main/java/org/apache/pinot/core/segment/index/loader/defaultcolumn/BaseDefaultColumnHandler.java
```diff
@@ -303,6 +304,10 @@ protected void createColumnV1Indices(String column)
         dictionaryElementSize = StringUtil.encodeUtf8(stringDefaultValue).length;
         sortedArray = new String[]{stringDefaultValue};
         break;
+      case BYTES:
+        Preconditions.checkState(defaultValue instanceof byte[]);
```
siddharthteotia: We should probably check if the default value byte array's length should be >= 1?

pinot-common/src/test/java/org/apache/pinot/common/data/FieldSpecTest.java
jackjlli: good catch 👍

pinot-core/src/test/resources/data/newColumnsSchema3.json
```diff
@@ -16,6 +16,11 @@
     {
       "dataType": "DOUBLE",
       "name": "newDoubleMetric"
+    },
+    {
+      "dataType": "BYTES",
+      "defaultNullValue": "00000008000000ac0000",
```
kishoreg: is this hexstring of the byte array

pinot-core/src/main/java/org/apache/pinot/core/segment/index/loader/defaultcolumn/BaseDefaultColumnHandler.java
```diff
@@ -303,6 +304,10 @@ protected void createColumnV1Indices(String column)
         dictionaryElementSize = StringUtil.encodeUtf8(stringDefaultValue).length;
```
kishoreg: we should use something that is independent of the encoding on the system. The file might be created/saved in a different encoding right

pinot-core/src/test/resources/data/newColumnsSchema3.json
jackjlli: This is one value of hyperLogLog

pinot-core/src/main/java/org/apache/pinot/core/segment/index/loader/defaultcolumn/BaseDefaultColumnHandler.java
```diff
@@ -303,6 +304,10 @@ protected void createColumnV1Indices(String column)
...
         sortedArray = new String[]{stringDefaultValue};
         break;
+      case BYTES:
+        Preconditions.checkState(defaultValue instanceof byte[]);
+        sortedArray = new ByteArray[] {new ByteArray((byte[]) defaultValue)};
```
mayankshriv: The defaultValue is coming from JSON schema, correct? The expectation is to specify ` Hex.encodeHexString(bytes)`  in the schema. Which means that this code would need to use `Hex.decodeHex` here.

Also, if we don't have this documented, let's also add the expectation on how to specify the default value in the schema for BYTES.

pinot-core/src/test/resources/data/newColumnsSchema3.json
jackjlli: Yes, this is helix string of the byte array.

pinot-core/src/main/java/org/apache/pinot/core/segment/index/loader/defaultcolumn/BaseDefaultColumnHandler.java
jackjlli: Actually the value is already of byte array type. We can document what/how to put a default value for byte column.

jackjlli:
@mayankshriv tests for both HLL and TDigest columns added.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -574,8 +575,14 @@ public static void addColumnMetadataInfo(PropertiesConfiguration properties, Str
     if (defaultNullValue == null) {
```
Jackie-Jiang: I don't think defaultNullValue can ever be null

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -574,8 +575,14 @@ public static void addColumnMetadataInfo(PropertiesConfiguration properties, Str
...
       defaultNullValue = fieldSpec.getDefaultNullValue();
     }
-    properties.setProperty(V1Constants.MetadataKeys.Column.getKeyFor(column, DEFAULT_NULL_VALUE),
-        String.valueOf(defaultNullValue));
+    if (fieldSpec.getDataType() == FieldSpec.DataType.BYTES && defaultNullValue instanceof byte[]) {
```
Jackie-Jiang: We just need the second check

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
jackjlli: Correct, we have default null value for all kinds of types in FieldSpec. 
Removed.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
jackjlli: Removed.

pinot-core/src/main/java/org/apache/pinot/core/segment/index/loader/defaultcolumn/BaseDefaultColumnHandler.java
Jackie-Jiang: @kishoreg We always encode String using UTF-8. As long as the files are generated with Pinot library, the String will be encoded with UTF-8. Record reader should be able to decode other format if necessary.

Issue #4230, opened 2019-05-21
NULL value support for all data types


Jackie-Jiang:
Currently in Pinot we don't have real NULL value support, but use some special default values for NULL. For dimensions, the default value is the minimum value for numeric types, "null" for STRING, empty byte array for BYTES; for metrics, the default value is 0 for numeric types, empty byte array for BYTES. #4214 adds support to treat empty byte array as NULL for BYTES, but it's not the general way to support NULL for all data types.
We should add another type of index to mark if the value is null, and apply that while filtering or fetching the values to skip all NULL values.

kishoreg:
Here is the high-level idea

- add isNullable=true/false in the FieldSpec
- if isNullable set to true, we can generate a presence vector or null vector. I think presence vector is better than null vector since the query will mostly be x != NULL. Having a presence vector means we don't have to invert the bitmap for every query
- Add presence bit vector to DataSource (similar to bloom filter etc)
- Nothing changes in the forward index for now. we will still use the default value. 
- We have two options during query execution 1. users explicitly add x != NULL in the query or we automatically add x != NULL if we see column x as part of the group by or aggregation function.

One open question is, should we restrict generating presence vector to segment creation phase or allow generating them on the fly when we load the segment (similar to inverted index). Generating them on the fly means we need to generate presence vector from the forward index. We can't do this as we represent NULL using the defaultValue in fieldSpec. One workaround is to treat the default value in the forward index as NULL. I think this might work but not 100% sure. Does anyone see any issues with treating default value in forward index implies NULL? 
   
For the first phase, I think restricting it to segment generation phase should be good enough.

Thoughts?

sunithabeeram:
Few thoughts:
- How would defaultValue and isNullable in the FieldSpec work? Does one take precedence over the other?
- Although I like the idea of presence vector, would it be better to have it a bit more dynamic by allowing for a metadata field that says whether it indicates the presence or absence depending on the type of data? This is an interesting consideration from storage pov (as we can use whatever results in better storage representation), but would make execution a bit non-straight forward.
- Do we want to generate the bit vector for metric fields?
- Regarding the open question, if we don't allow on the fly generation, how would this work with older segments? I think interpreting existing default values to generate the bit vector sounds ok to me.

kishoreg:
@sunithabeeram Changed bullet to the numbered list.
[...]

PR #4585, opened 2019-09-04
Presence vector

icefury71:
This PR adds support for a presence vector inside a mutable and immutable segment. This will enable the query layer to ignore null values in the corresponding columns. Please see this issue for more details: https://github.com/apache/incubator-pinot/issues/4230

High level gist is as follows:
* Create a presence vector per column per segment by default. This presence vector keeps track of which document ID has a null value in the corresponding column.
* Expose this through the DataSource interface to enable the caller to ignore such document IDs.

Presence vector is not really used anywhere as part of this PR. Subsequent PRs will use them to enable filtering out columns in predicates (eg: select ... from table where column_name != null)


pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/PresenceVectorReaderImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public PresenceVectorReaderImpl(PinotDataBuffer presenceVectorBuffer) throws IOException {
+    _nullBitmap = new ImmutableRoaringBitmap(presenceVectorBuffer.toDirectByteBuffer(0, (int)presenceVectorBuffer.size()));
+  }
+
+  public boolean isPresent(int docId) {
+    return !_nullBitmap.contains(docId);
+  }
+
+  public static void main(String[] args) {
```
snleee: remove

pinot-core/src/test/resources/data/test_presence_vector_data.json
```diff
@@ -0,0 +1,12 @@
...
+  "secondsSinceEpoch": 1567205394
+}
+{
+  "cityid" : -2053105565,
+  "description" : "q\u0012=cLp{-PDa\fT",
+  "signup_email" : {
+    "string" : "fI\u00194u@WI\f\u001Dt"
+  },
+  "secondsSinceEpoch": 1567205394
+}
```
snleee: add line

pinot-core/src/test/resources/data/test_presence_vector_pinot_schema.json
```diff
@@ -0,0 +1,26 @@
...
+    }
+  ],
+  "timeFieldSpec": {
+    "incomingGranularitySpec": {
+      "name": "secondsSinceEpoch",
+      "dataType": "LONG",
+      "timeType": "SECONDS"
+    }
+  }
+}
```
snleee: add line

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterIntegrationTestUtils.java
```diff
@@ -283,6 +283,7 @@ public static void buildSegmentsFromAvro(List<File> avroFiles, int baseSegmentIn
           TarGzCompressionUtils
               .createTarGzOfDirectory(segmentFile.getAbsolutePath(), new File(tarDir, segmentName).getAbsolutePath());
         } catch (Exception e) {
+          e.printStackTrace();
```
snleee: why do we need to add this?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -373,6 +373,8 @@ public ServerType getServerType() {
 
     public static final String LOCAL_SEGMENT_SCHEME = "file";
 
+    public static final String NULL_FIELDS = "null_fields";
```
snleee: `null.fields`

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
```diff
@@ -33,17 +36,35 @@ public NullValueTransformer(Schema schema) {
...
         if (fieldSpec.isSingleValueField()) {
           record.putField(fieldName, fieldSpec.getDefaultNullValue());
         } else {
           record.putField(fieldName, new Object[]{fieldSpec.getDefaultNullValue()});
         }
       }
     }
+
+    if (nullColumnNamesSet != null) {
+      record.putField(NULL_FIELDS, String.join(",", nullColumnNamesSet));
```
snleee: I think that we can directly store `Set` since `GenericRow` gets `Object` as value instead of keep repeating `join & parse`?

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/presence/RealtimePresenceVectorReaderWriter.java
```diff
@@ -0,0 +1,41 @@
...
+ * Defines a real-time presence vector to be used in realtime ingestion.
+ */
+public class RealtimePresenceVectorReaderWriter implements PresenceVectorReader {
+    private final MutableRoaringBitmap _nullBitmap;
+
+    public RealtimePresenceVectorReaderWriter() {
+        _nullBitmap = new MutableRoaringBitmap();
+    }
+
+    public void setIsNull(int docId) {
```
snleee: `setNull`?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/presence/PresenceVectorCreator.java
```diff
@@ -0,0 +1,77 @@
...
+    try (DataOutputStream outputStream = new DataOutputStream(new FileOutputStream(_presenceVectorFile))) {
+      _nullBitmap.serialize(outputStream);
+    }
+  }
+
+  public void setIsNull(int docId) {
+    _nullBitmap.add(docId);
+  }
+
+  public static void main(String[] args)
```
snleee: remove

pinot-core/src/test/resources/data/test_presence_vector_data.json
```diff
@@ -0,0 +1,12 @@
+{
```
snleee: should we also test the case something like

`"description" : null`?

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -479,6 +521,20 @@ public GenericRow getRecord(int docId, GenericRow reuse) {
           .getValue(docId, fieldSpec, _indexReaderWriterMap.get(column), _dictionaryMap.get(column),
               _maxNumValuesMap.getOrDefault(column, 0)));
     }
+
+    // Explicitly set it to null in case of reuse
+    reuse.putField(NULL_FIELDS, null);
+
+    // Look up bitmap for this docId and construct a comma separated
+    // null column names based on that bitmap
+    RoaringBitmap bitmap = _docIdToNullColumnsMap.get(docId);
```
snleee: why can't we rebuild the row using `_docIdToNullColumnsMap`? Because we need to call `isPresent()` for all columns?

Do we know the overheads on this data structure in memory?

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
icefury71: I think the concern is the size of a Java object which is potentially bigger in this case. 

One optimization here is that if we have a fixed order of column names (in the FieldSpec), we can simply use a BitSet to keep track of null columns (which will typically be around 64 bytes ish). This can save a lot of effort in other parts of the code base as well.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
icefury71: Yes - currently we need to call isPresent to check which columns have corresponding null values for this docId.

Yes there's considerable overhead using a BitMap but not as much as using a comma separated string / Set

icefury71: I think if we move to a BitSet based approach (as mentioned above) - it'll make this part much simpler. We can simply save the Bitset as is per docId and use that to recreate the row.

pinot-core/src/test/resources/data/test_presence_vector_data.json
icefury71: I wanted to leave one column non-null (in this case description).

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/ClusterIntegrationTestUtils.java
icefury71: This seems a mistake. Will remove this.

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplPresenceVectorTest.java
```diff
@@ -18,23 +18,23 @@
...
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
```
haibow: You may want to undo these wildcard imports

https://stackoverflow.com/questions/3587071/disable-intellij-starred-package-imports

pinot-core/src/main/java/org/apache/pinot/core/common/DataSource.java
```diff
@@ -19,9 +19,7 @@
 package org.apache.pinot.core.common;
 
 import org.apache.pinot.core.operator.BaseOperator;
-import org.apache.pinot.core.segment.index.readers.BloomFilterReader;
-import org.apache.pinot.core.segment.index.readers.Dictionary;
-import org.apache.pinot.core.segment.index.readers.InvertedIndexReader;
+import org.apache.pinot.core.segment.index.readers.*;
```
haibow: You may want to undo wildcard imports when formatting
https://stackoverflow.com/questions/3587071/disable-intellij-starred-package-imports.

Same with several other files in this diff.

icefury71: Aah good catch. Will fix this !

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
icefury71: FYI: I pushed a change to optimize this part. Please take a look.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/presence/RealtimePresenceVectorReaderWriter.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.pinot.core.realtime.impl.presence;
+
+import org.apache.pinot.core.segment.index.readers.PresenceVectorReader;
+import org.roaringbitmap.buffer.MutableRoaringBitmap;
+
+/**
+ * Defines a real-time presence vector to be used in realtime ingestion.
+ */
+public class RealtimePresenceVectorReaderWriter implements PresenceVectorReader {
+    private final MutableRoaringBitmap _nullBitmap;
```
snleee: We use 2 spaces. Can you reformat?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -254,6 +261,13 @@ private boolean createDictionaryForColumn(ColumnIndexCreationInfo info, SegmentG
 
   @Override
   public void indexRow(GenericRow row) {
+
```
snleee: remove line

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/PresenceVectorReader.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Reader interface to read from an underlying Presence vector. This is
+ * primarily used to check if a particular column value corresponding to
+ * a document ID is null or not.
+ */
+public interface PresenceVectorReader {
+
+    /**
+     * Check if the given docId is present in the corresponding column
```
snleee: 2 spaces

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplPresenceVectorTest.java
```diff
@@ -0,0 +1,104 @@
...
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+
+import static org.apache.pinot.common.utils.CommonConstants.Segment.NULL_FIELDS;
+
+public class MutableSegmentImplPresenceVectorTest {
```
snleee: 2spaces

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/presence/RealtimePresenceVectorReaderWriterTest.java
```diff
@@ -0,0 +1,42 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.core.realtime.impl.presence;
+
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+public class RealtimePresenceVectorReaderWriterTest {
```
snleee: 2 spaces

pinot-core/src/main/java/org/apache/pinot/core/operator/docidsets/BitmapDocIdSet.java
```diff
@@ -33,7 +36,8 @@
   public BitmapDocIdSet(ImmutableRoaringBitmap[] bitmaps, int startDocId, int endDocId, boolean exclusive) {
     int numBitmaps = bitmaps.length;
     if (numBitmaps > 1) {
-      MutableRoaringBitmap orBitmap = MutableRoaringBitmap.or(bitmaps);
+      Iterator iterator = Arrays.asList(bitmaps).iterator();
+      MutableRoaringBitmap orBitmap = MutableRoaringBitmap.or(iterator, startDocId, endDocId + 1);
```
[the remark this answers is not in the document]
icefury71: @kishoreg : can you please clarify ? I'm not sure we need this change for presence vector.

icefury71:
> Sorry for the delay. I was out for a while for the conference.
> 
> Would you add the following to the commit message?
> 
> 1. Link #4230 for the reference. After I read the issue, this pr made much more sense.
> 2. Add the description on how presence vector is populated. By reading the code, it seems that we generate presence vector by default.
> 3. Can you also add a little bit more explanation to the commit message on when you filter out NULL values? (e.g. when predicate is added by user -- column != NULL...)
> 
> I think that most comments are minor & style issues. I will do one more final review after this.

Thanks for pointing that out. Updated description. Please take a look

snleee:
@icefury71 Thanks for the reply. LGTM assuming that you address the comments above (2 spaces, comment on bloom filter etc). I will merge this once that part is updated. 

@kishoreg Can you reply to `BitmapDocIdSet.java#line 39` change?

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
```diff
@@ -22,28 +22,45 @@
 import org.apache.pinot.common.data.FieldSpec;
 import org.apache.pinot.common.data.Schema;
 import org.apache.pinot.core.data.GenericRow;
+import org.roaringbitmap.RoaringBitmap;
+
+import static org.apache.pinot.common.utils.CommonConstants.Segment.NULL_FIELDS;
```
Jackie-Jiang: (minor) please don't use static import for non-testing class. Same for other classes

icefury71:
> Why do we need to maintain a record level bitmap? That can easily cause memory issue (everything on heap).
> If we want to preserve `null` value, we should remove the NullValueTransformer from the record reader and let the record reader insert `null` into the GenericRow. The segment creator/mutable segment should be able to directly process `null` value.
> 

The bitmap added to Generic Row inside NullValueTransformer is a transient object (short lived) - which shouldn't cause any memory concern. 

However, I was maintaining a bitmap per docId inside MutableSegmentImpl. I've optimized that part of the code - its no longer necessary to keep track of all these bitmaps in memory.

icefury71:
> @icefury71 Thanks for the reply. LGTM assuming that you address the comments above (2 spaces, comment on bloom filter etc). I will merge this once that part is updated.
> 

I've reformatted code based on Pinot style.

Jackie-Jiang:
I'm still against storing the bitmap into the GenericRecord because that will couple the presence vector with the NullValueTransformer (real-time segment generation does not use NullValueTransformer, thus it won't work properly).
Instead, we should just allow null values (missing fields) in the GenericRecord when it is passed to the SegmentCreator, then it becomes much more flexible, and we can choose to fill in default values, or add them into the presence vector.

icefury71:
> I'm still against storing the bitmap into the GenericRecord because that will couple the presence vector with the NullValueTransformer (real-time segment generation does not use NullValueTransformer, thus it won't work properly).

By real-time I'm assuming you mean either HLC / LLC realtime segments ? From code inspection it does look like we're using NullValueTransformer in real-time path:

https://github.com/apache/incubator-pinot/blob/master/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java#L1158

(this is calling CompositeTransformer.getDefaultTransformer which in turn brings in NullValueTransformer)

> Instead, we should just allow null values (missing fields) in the GenericRecord when it is passed to the SegmentCreator, then it becomes much more flexible, and we can choose to fill in default values, or add them into the presence vector.

We considered this option. However, the problem is that it's extremely difficult to identify default null value which is outside the domain of legitimate values. In many cases there could be an overlap.

Given that memory is not a big concern, this could be a good way out. Thoughts ?

pinot-core/src/main/java/org/apache/pinot/core/operator/docidsets/BitmapDocIdSet.java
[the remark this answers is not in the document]
icefury71: I believe this might be slightly more efficient (based on conversation with Kishore)

icefury71:
Did a rebase against master to resolve conflicts.

Jackie-Jiang:
The record reader you mentioned is for ingesting data into real-time. When we convert real-time segment into offline segment, we use the RealtimeSegmentConverter which does not perform any transform on the records as records should already be in the desired format.
The concern is more about module isolation, meaning SegmentCreator interface should not expect the GenericRow to have a field of bitmap.
> 
> We considered this option. However, the problem is that it's extremely difficult to identify default null value which is outside the domain of legitimate values. In many cases there could be an overlap.
I don't quite follow. The logic should be very straight forward. If we allow null value, put the null value into the presence vector; if not, throw exception

pinot-common/src/main/java/org/apache/pinot/common/config/IndexingConfig.java
```diff
@@ -87,6 +87,9 @@
   @ConfigKey("aggregateMetrics")
   private boolean _aggregateMetrics;
 
+  @ConfigKey("nullHandlingEnabled")
```
Jackie-Jiang: `allowNullValue` is more clear?

pinot-common/src/main/java/org/apache/pinot/common/data/Schema.java
```diff
@@ -226,6 +226,7 @@ public void addField(@Nonnull String columnName, @Nonnull FieldSpec fieldSpec) {
 
   public boolean removeField(String columnName) {
     FieldSpec existingFieldSpec = _fieldSpecMap.remove(columnName);
+
```
Jackie-Jiang: (nit) revert this file?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -376,6 +376,8 @@ public ServerType getServerType() {
 
     public static final String LOCAL_SEGMENT_SCHEME = "file";
 
+    public static final String NULL_FIELDS = "null.fields";
```
Jackie-Jiang: This is not required

pinot-core/src/main/java/org/apache/pinot/core/common/DataSource.java
```diff
@@ -45,4 +46,7 @@
    * Returns the bloom filter for the data source if exists, or {@code null} if not.
    */
   public abstract BloomFilterReader getBloomFilter();
+
+  public abstract PresenceVectorReader getPresenceVector();
```
Jackie-Jiang: This is more like an NullValueVector

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/NullValueTransformer.java
```diff
@@ -19,18 +19,27 @@
...
 import java.util.Map;
+import java.util.Set;
 import org.apache.pinot.common.data.FieldSpec;
 import org.apache.pinot.common.data.FieldSpec.FieldType;
 import org.apache.pinot.common.data.Schema;
+import org.apache.pinot.common.utils.CommonConstants;
 import org.apache.pinot.core.data.GenericRow;
 
 
 public class NullValueTransformer implements RecordTransformer {
```
Jackie-Jiang: null values has been integrated into GenericRow in #4671

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/PresenceVectorReader.java
```diff
@@ -0,0 +1,36 @@
...
+ * under the License.
+ */
+package org.apache.pinot.core.segment.index.readers;
+
+/**
+ * Reader interface to read from an underlying Presence vector. This is
+ * primarily used to check if a particular column value corresponding to
+ * a document ID is null or not.
+ */
+public interface PresenceVectorReader {
```
Jackie-Jiang: We need APIs to get the whole presence vector/null vector

pinot-core/src/main/java/org/apache/pinot/core/common/DataSource.java
icefury71: PresenceVectorReader has a method called "isPresent(docId)" . So it is in fact a presence value vector (although internally it uses a null vector).

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/PresenceVectorReader.java
icefury71: I can add this in the next iteration (as needed during the query layer evolution)

pinot-common/src/main/java/org/apache/pinot/common/data/Schema.java
icefury71: Done

pinot-common/src/main/java/org/apache/pinot/common/config/IndexingConfig.java
icefury71: I'll keep this unless there's a strong objection.

pinot-common/src/main/java/org/apache/pinot/common/config/IndexingConfig.java
```diff
@@ -239,6 +242,12 @@ public void setVarLengthDictionaryColumns(List<String> varLengthDictionaryColumn
     _varLengthDictionaryColumns = varLengthDictionaryColumns;
   }
 
+  public boolean isNullHandlingEnabled() { return _nullHandlingEnabled; }
```
Jackie-Jiang: (Code style) Expand this into 3 lines

pinot-core/src/main/java/org/apache/pinot/core/data/GenericRow.java
```diff
@@ -90,6 +90,11 @@ public boolean isNullValue(String fieldName) {
     return _nullValueFields.contains(fieldName);
   }
 
+  /**
+   * Returns whether this row has null values for any of the columns
+   */
+  public boolean hasNullValues() { return _nullValueFields.size() > 0; }
```
Jackie-Jiang: (Code style) Expand this into 3 lines.
(nit) `return !_nullValueFields.isEmpty()` seems more readable

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -379,6 +393,25 @@ private void addInvertedIndex(int docId, Map<String, Object> dictIdMap) {
...
+   * column presence vectors accordingly
+   * @param row specifies row being ingested
+   * @param docId specified docId for this row
+   */
+  private void addPresenceVector(GenericRow row, int docId) {
+    if (!row.hasNullValues()) {
+      return;
+    }
+
+    for (FieldSpec fieldSpec : _schema.getAllFieldSpecs()) {
```
Jackie-Jiang: Iterate over `row.getNullValueFields()` and set presence vector

pinot-core/src/main/java/org/apache/pinot/core/operator/docidsets/BitmapDocIdSet.java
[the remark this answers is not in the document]
Jackie-Jiang: This could backfire because there is an extra step of selecting range for each bitmap. I checked the code and seems this step is always redundant because we always use 0 as startDocId and numDocs-1 as endDocId. Also this is not related to this PR. @kishoreg Do you see actual performance gain for this change?

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/RealtimeSegmentConfig.java
```diff
@@ -40,13 +40,25 @@
...
       Set<String> invertedIndexColumns, RealtimeSegmentZKMetadata realtimeSegmentZKMetadata,
       boolean offHeap, PinotDataBufferMemoryManager memoryManager,
       RealtimeSegmentStatsHistory statsHistory, SegmentPartitionConfig segmentPartitionConfig,
       boolean aggregateMetrics) {
+    this(segmentName, streamName, schema, capacity, avgNumMultiValues, noDictionaryColumns, varLengthDictionaryColumns,
+        invertedIndexColumns, realtimeSegmentZKMetadata, offHeap, memoryManager, statsHistory, segmentPartitionConfig,
+        aggregateMetrics, false);
+  }
+
+  private RealtimeSegmentConfig(String segmentName, String streamName, Schema schema, int capacity,
```
Jackie-Jiang: You don't need two constructors as it is private and always constructed from the builder.

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/RealtimeSegmentConfig.java
```diff
@@ -208,11 +224,20 @@ public Builder setAggregateMetrics(boolean aggregateMetrics) {
       return this;
     }
 
+    public boolean isNullHandlingEnabled() {
```
Jackie-Jiang: Remove the getter from the Builder

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/RealtimeSegmentConfig.java
```diff
@@ -119,6 +132,8 @@ public boolean aggregateMetrics() {
     return _aggregateMetrics;
   }
 
+  public boolean isNullHandlingEnabled() { return _nullHandlingEnabled; }
```
Jackie-Jiang: (Code style) 3 lines

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/PresenceVectorReader.java
```diff
@@ -0,0 +1,36 @@
...
+
+
+  /**
+   * Check if the given docId is present in the corresponding column
+   *
+   * @param docId specifies ID to check for presence
+   * @return true if docId is present (non null). False otherwise
+   */
+  boolean isPresent(int docId);
```
Jackie-Jiang: Going over the code, seems we only need `isNull(int docId)`? (All the returns of `isPresent()` are reversed)

pinot-core/src/main/java/org/apache/pinot/core/segment/index/column/ColumnIndexContainer.java
```diff
@@ -44,5 +45,15 @@
    */
   Dictionary getDictionary();
 
+  /**
+   *
+   * @return Get the bloom filter for the column if it exists else  {@code null}
```
Jackie-Jiang: Can you make the javadoc consistent with other methods?

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/V1Constants.java
```diff
@@ -39,6 +39,7 @@
     public static final String UNSORTED_MV_FORWARD_INDEX_FILE_EXTENSION = ".mv.fwd";
     public static final String BITMAP_INVERTED_INDEX_FILE_EXTENSION = ".bitmap.inv";
     public static final String BLOOM_FILTER_FILE_EXTENSION = ".bloom";
+    public static final String PRESENCE_VECTOR_FILE_EXTENSION = ".bitmap.presence";
```
Jackie-Jiang: Still feel this can cause confusion because the bitmap is actually null vector.

pinot-common/src/main/java/org/apache/pinot/common/config/IndexingConfig.java
icefury71: Done

pinot-core/src/main/java/org/apache/pinot/core/data/GenericRow.java
icefury71: Done

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
icefury71: Done

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/RealtimeSegmentConfig.java
icefury71: Done

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/nullvalue/NullValueVectorCreator.java
```diff
@@ -0,0 +1,62 @@
...
+  private MutableRoaringBitmap _nullBitmap;
+  private File _nullValueVectorFile;
+
+  public NullValueVectorCreator(File indexDir, String columnName) {
+    _nullValueVectorFile = new File(indexDir, columnName + V1Constants.Indexes.NULLVALUE_VECTOR_FILE_EXTENSION);
+    _nullBitmap = new MutableRoaringBitmap();
+  }
+
+  @Override
+  public void close()
```
Jackie-Jiang: Recommend having a `seal()` method similar to `InvertedIndexCreator` instead of using `close()` to persist the data.

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/NullValueVectorReader.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.apache.pinot.core.segment.index.readers;
+
+/**
+ * Reader interface to read from an underlying Null value vector. This is
+ * primarily used to check if a particular column value corresponding to
+ * a document ID is null or not.
+ */
+public interface NullValueVectorReader {
+
```
Jackie-Jiang: (nit) remove one empty line

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/NullValueVectorReaderImpl.java
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.pinot.core.segment.index.readers;
+
+import java.io.IOException;
+import org.apache.pinot.core.segment.memory.PinotDataBuffer;
+import org.roaringbitmap.buffer.ImmutableRoaringBitmap;
+
+
+public class NullValueVectorReaderImpl implements NullValueVectorReader {
+
+  ImmutableRoaringBitmap _nullBitmap;
```
Jackie-Jiang: private final

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/NullValueVectorReaderImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public NullValueVectorReaderImpl(PinotDataBuffer nullValueVectorBuffer) throws IOException {
+    _nullBitmap = new ImmutableRoaringBitmap(nullValueVectorBuffer.toDirectByteBuffer(0,
+        (int) nullValueVectorBuffer.size()));
+  }
+
+  public boolean isNull(int docId) {
+    return _nullBitmap.contains(docId);
+  }
+
```
Jackie-Jiang: (nit) remove this empty line

pinot-core/src/test/java/org/apache/pinot/core/segment/creator/impl/nullvalue/NullValueVectorCreatorTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+
+public class NullValueVectorCreatorTest {
+    private static final File TEMP_DIR = new File(FileUtils.getTempDirectory(), "NullValueVectorCreatorTest");
```
Jackie-Jiang: Change indentation to 2 spaces

pinot-core/src/test/java/org/apache/pinot/core/segment/creator/impl/nullvalue/NullValueVectorCreatorTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void testNullValueVectorCreation() {
+        try (NullValueVectorCreator creator = new NullValueVectorCreator(TEMP_DIR, COLUMN_NAME)) {
+            for (int i = 0; i < 100; i++) {
+                creator.setNull(i);
+            }
+            ImmutableRoaringBitmap nullBitmap = creator.getNullBitmap();
+            for (int i = 0; i < 100; i++) {
+                Assert.assertTrue(nullBitmap.contains(i));
+            }
+        } catch (IOException e) {
```
Jackie-Jiang: You can directly throw out the exception

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplNullValueVectorTest.java
```diff
@@ -0,0 +1,94 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+
+public class MutableSegmentImplNullValueVectorTest {
+  private static final String PINOT_SCHEMA_FILE_PATH = "data/test_null_value_vector_pinot_schema.json";
```
Jackie-Jiang: You can directly construct the Schema and Records, instead of reading from file as we won't want to test the functionality of Schema deserialization and RecordReader. You can use the Schema.Builder and GenericRowRecordReader.

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/nullvalue/RealtimeNullValueVectorReaderWriterTest.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.apache.pinot.core.realtime.impl.nullvalue;
+
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+public class RealtimeNullValueVectorReaderWriterTest {
+  private static RealtimeNullValueVectorReaderWriter readerWriter = null;
```
Jackie-Jiang: private RealtimeNullValueVectorReaderWriter _readerWriter;

pinot-core/src/test/java/org/apache/pinot/core/realtime/impl/nullvalue/RealtimeNullValueVectorReaderWriterTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @BeforeClass
+  public void setup() {
+    readerWriter = new RealtimeNullValueVectorReaderWriter();
+  }
+
+  @Test
+  public void testRealtimeNullValueVectorReaderWriter() {
+    for (int i = 0; i < 100; i++) {
```
Jackie-Jiang: Can we choose some random docIds to add?

pinot-core/src/test/java/org/apache/pinot/core/segment/index/readers/NullValueVectorReaderImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+
+public class NullValueVectorReaderImplTest {
+    private static final File TEMP_DIR = new File(FileUtils.getTempDirectory(), "NullValueVectorReaderImplTest");
```
Jackie-Jiang: Indentation

pinot-core/src/test/java/org/apache/pinot/core/segment/index/readers/NullValueVectorReaderImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+        if (TEMP_DIR.exists()) {
+            FileUtils.deleteQuietly(TEMP_DIR);
+        }
+        TEMP_DIR.mkdir();
+        try (NullValueVectorCreator creator = new NullValueVectorCreator(TEMP_DIR, COLUMN_NAME)) {
+            for (int i = 0; i < 100; i++) {
+                creator.setNull(i);
+            }
+        } catch (IOException e) {
+            Assert.fail("Unable to create NullValueVectorCreator", e);
```
Jackie-Jiang: You can directly throw out the exception.

pinot-core/src/test/java/org/apache/pinot/core/segment/index/readers/NullValueVectorReaderImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+        Assert.assertEquals(TEMP_DIR.list().length, 1);
+        File nullValueFile = new File(TEMP_DIR, TEMP_DIR.list()[0]);
+        try {
+            PinotDataBuffer buffer = PinotDataBuffer.loadBigEndianFile(nullValueFile);
+            NullValueVectorReader reader = new NullValueVectorReaderImpl(buffer);
+            for (int i = 0; i < 100; i++) {
+                Assert.assertTrue(reader.isNull(i));
+            }
+        } catch (IOException e) {
+            Assert.fail("Unable to create NullValueVectorReader from given file", e);
```
Jackie-Jiang: You can directly throw out the exception.

pinot-core/src/test/java/org/apache/pinot/core/segment/index/readers/NullValueVectorReaderImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.pinot.core.segment.memory.PinotDataBuffer;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+
+public class NullValueVectorReaderImplTest {
```
Jackie-Jiang: The creator test can be merged into this. Up to you

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/nullvalue/NullValueVectorCreator.java
icefury71: Any particular reason why ? It's using the same approach as Bloom Filter creator

Jackie-Jiang: The semantic for `close()` is "Closes this resource, relinquishing any underlying resources." and it should not be responsible for persisting the data. Bloom Filter should also follow this, but that is out of the scope of this PR.

pinot-core/src/test/java/org/apache/pinot/core/segment/index/readers/NullValueVectorReaderImplTest.java
icefury71: Keeping this as is for now.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/nullvalue/NullValueVectorCreator.java
icefury71: done -> added seal

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/PresenceVectorReader.java
icefury71: modified to isNull

pinot-core/src/main/java/org/apache/pinot/core/common/DataSource.java
icefury71: Modified to NullValueVector

icefury71:
> Good in general.
> I still feel it is better to make it NullValueVector for the following reasons:
> 
> * Underlying bitmap is nullValueBitmap
> * Always set null instead of set non-null
> * We only need to check whether the value is null (all the return for isPresent() is reversed)
> 
> I understand the intention for PresenceVector is to fast return the presenceBitmap, but if the underlying storage is actually a nullValueBitmap, it is quite confusing. In the following PRs, we can maintain two bitmaps in memory, but still only store nullValueBitmap on disk. @kishoreg Thoughts?

Modified to NullValueVector as per suggestion

PR #4586, opened 2019-09-04
Adding Predicate supports for Bytes Column without dictionary

fx19880617:
Support predicate for equal/not equ/in/not in/ range

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/filter/predicate/InPredicateEvaluatorFactory.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/filter/predicate/InPredicateEvaluatorFactory.java
@@ -32,6 +32,7 @@ import java.util.Set;
 import org.apache.pinot.common.data.FieldSpec;
 import org.apache.pinot.common.utils.BytesUtils;
 import org.apache.pinot.common.utils.HashUtil;
+import org.apache.pinot.common.utils.primitive.ByteArray;
 import org.apache.pinot.core.common.Predicate;
 import org.apache.pinot.core.common.predicate.InPredicate;
 import org.apache.pinot.core.segment.index.readers.Dictionary;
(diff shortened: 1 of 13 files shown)


kishoreg:
- There are cases where request id's can be generated as MD5 which can be represented as byte[]. Searching for a specific request is a common access pattern. 
- Query looks like select * from T where requestId = 'hexEncodedByteString'

Irrespective of the use case, we should support filtering on all data types for sake of completion. Currently, we support predicates on bytes if its dictionary encoded but not if it's stored as RAW.

pinot-core/src/main/java/org/apache/pinot/core/operator/filter/predicate/InPredicateEvaluatorFactory.java
```diff
@@ -233,4 +236,24 @@ public boolean applySV(String value) {
       return _matchingValues.contains(value);
     }
   }
+
+  private static final class BytesRawValueBasedInPredicateEvaluator extends BaseRawValueBasedPredicateEvaluator {
+    final Set<String> _matchingValues;
```
Jackie-Jiang: Storing `Set<ByteArray>` should be faster.

pinot-core/src/main/java/org/apache/pinot/core/operator/filter/predicate/InPredicateEvaluatorFactory.java
```diff
@@ -233,4 +236,24 @@ public boolean applySV(String value) {
...
+
+    BytesRawValueBasedInPredicateEvaluator(InPredicate inPredicate) {
+      String[] values = inPredicate.getValues();
+      _matchingValues = new HashSet<>(HashUtil.getMinHashSetSize(values.length));
```
Jackie-Jiang: Not related to this pr, but `HashUtil.getMinHashSetSize()` is implemented wrong

pinot-core/src/main/java/org/apache/pinot/core/operator/filter/predicate/NotInPredicateEvaluatorFactory.java
```diff
@@ -251,4 +254,25 @@ public boolean applySV(String value) {
       return !_nonMatchingValues.contains(value);
     }
   }
+
+  private static final class BytesRawValueBasedNotInPredicateEvaluator extends BaseRawValueBasedPredicateEvaluator {
+    final Set<String> _nonMatchingValues;
```
Jackie-Jiang: Use `Set<ByteArray>`

PR #4590, opened 2019-09-06
Disallowing multiple inputs in preprocess

jenniferdai:
* We cannot support this in preprocess because mapreduce takes complete control of the output path(s). In order to support this, we would need control to pipe the exact folders we receive as input to multiple outputs. While we can programmatically determine record by record what goes into each output path, this does not support our use case. Each folder is a separate "day," and frequently, our customers will have two dates in one file, due to timezone of data, so we are not able to distinguish what is "today's" vs. "tomorrow's" data by solely looking at the record.

### DIFF
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
@@ -83,14 +83,15 @@ public class SegmentPreprocessingJob extends BaseSegmentJob {
   private String _partitionFunction;
   private String _sortedColumn;
   private int _numOutputFiles;
+  private boolean _isMultipleInput = false;
 
-  private final Path _inputSegmentDir;
-  private final Path _preprocessedOutputDir;
-  protected final String _rawTableName;
-  protected final List<PushLocation> _pushLocations;
+  private Path _inputSegmentDir;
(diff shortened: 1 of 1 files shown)


pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentPreprocessingJob.java
```diff
@@ -101,8 +102,20 @@ public SegmentPreprocessingJob(final Properties properties) {
...
-    _inputSegmentDir = Preconditions.checkNotNull(getPathFromProperty(JobConfigConstants.PATH_TO_INPUT));
+    String inputPath = Preconditions.checkNotNull(properties.getProperty(JobConfigConstants.PATH_TO_INPUT));
+
+    // We cannot support this because mapreduce takes complete control of the output path. In order to support this, we
+    // would need control to pipe the exact folders we receive as input to multiple outputs. While we can programmatically
+    // determine record by record what goes into each output path, this does not support our use case. Each folder is a
+    // separate "day," and frequently, our customers will have two dates in one file, due to timezone of data, so we
+    // are not able to distinguish what is "today's" vs. "tomorrow's" data by solely looking at the record.
+    if (inputPath.split(",").length > 1) {
+      _isMultipleInput = true;
```
Jackie-Jiang: Let's directly throw exception here with proper error message

kishoreg:
Why are we doing this?

PR #4591, opened 2019-09-06
Deleting preprocess output post segment creation

jenniferdai:
* If the preprocessed output dir is defined, we will delete it after segment creation succeeds.

### DIFF
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentCreationJob.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentCreationJob.java
@@ -63,6 +63,7 @@ public class SegmentCreationJob extends BaseSegmentJob {
   protected final Path _schemaFile;
   protected final String _defaultPermissionsMask;
   protected final List<PushLocation> _pushLocations;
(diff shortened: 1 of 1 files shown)


pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -90,6 +91,11 @@ public SegmentCreationJob(Properties properties) {
       _pushLocations = null;
     }
 
+    String preprocessOutputPath = _properties.getProperty(JobConfigConstants.PREPROCESS_PATH_TO_OUTPUT);
```
Jackie-Jiang: SegmentCreationJob should have no concept on the pre-process path. Also, client might want to keep the pre-process output files, so we should not always delete them

jackjlli: I would say if user has specified `PREPROCESS_PATH_TO_OUTPUT`, it's ok to keep it.
If not, use a tmp dir as the preprocess output path, then remove this tmp dir when the whole workflow finished.

PR #4592, opened 2019-09-06
Refactor Realtime Segment Data Manager to remove repeat logic 

jenniferdai:
* Allows for tests specifically targeting the controller portion of split commit

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -768,9 +768,20 @@ public class LLRealtimeSegmentDataManager extends RealtimeSegmentDataManager {
     }
     SegmentCompletionProtocol.Response returnedResponse;
     boolean isSplitCommit = response.isSplitCommit() && _indexLoadingConfig.isEnableSplitCommit();
-    SegmentCommitter segmentCommitter = new SegmentCommitter(isSplitCommit, _segmentBuildDescriptor, _segmentNameStr,
-        _currentOffset, _numRowsConsumed, _instanceId, _isOffHeap, _protocolHandler, _memoryManager, _indexLoadingConfig, response);
-    returnedResponse = segmentCommitter.commitSegment();
+
+    SegmentCompletionProtocol.Request.Params params = new SegmentCompletionProtocol.Request.Params();
+
+    params.withSegmentName(_segmentNameStr).withOffset(_currentOffset).withNumRows(_numRowsConsumed)
+        .withInstanceId(_instanceId).withBuildTimeMillis(_segmentBuildDescriptor.getBuildTimeMillis())
+        .withSegmentSizeBytes(_segmentBuildDescriptor.getSegmentSizeBytes())
+        .withWaitTimeMillis(_segmentBuildDescriptor.getWaitTimeMillis());
+
+    if (_isOffHeap) {
+      params.withMemoryUsedBytes(_memoryManager.getTotalAllocatedBytes());
+    }
+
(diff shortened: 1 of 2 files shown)


pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -760,73 +760,28 @@ protected SegmentBuildDescriptor buildSegmentInternal(boolean forCommit) {
     }
   }
 
-  private SegmentCompletionProtocol.Response doSplitCommit(SegmentCompletionProtocol.Response prevResponse) {
-    final File segmentTarFile = new File(_segmentBuildDescriptor.getSegmentTarFilePath());
+  protected boolean commitSegment(SegmentCompletionProtocol.Response response) {
```
mcvsubbu: Extract value of `isSplitCommit` and `controllerVipUrl` from response, and pass those as arguments into the `commitSegment` method.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -760,73 +760,28 @@ protected SegmentBuildDescriptor buildSegmentInternal(boolean forCommit) {
...
-    if (response.isSplitCommit() && _indexLoadingConfig.isEnableSplitCommit()) {
-      // Send segmentStart, segmentUpload, & segmentCommitEnd to the controller
-      // if that succeeds, swap in-memory segment with the one built.
-      returnedResponse = doSplitCommit(response);
-    } else {
-      // Send segmentCommit() to the controller
-      // if that succeeds, swap in-memory segment with the one built.
-      returnedResponse = postSegmentCommitMsg();
-    }
+    SegmentCommitter segmentCommitter = new SegmentCommitter(isSplitCommit, _segmentNameStr, _protocolHandler, _indexLoadingConfig, response, params);
```
mcvsubbu: segmentNameStr should not be needed. Already there in `params`
Remove `_indexLoadingConfig`. Instead, extract  value of `isSplitCommitEndWithMetadata` and pass it to the constructor.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SegmentCommitter.java
```diff
@@ -0,0 +1,108 @@
...
+  private boolean _isSplitCommit;
+  private String _segmentNameStr;
+  private ServerSegmentCompletionProtocolHandler _protocolHandler;
+  private IndexLoadingConfig _indexLoadingConfig;
+  private SegmentCompletionProtocol.Response _prevResponse;
+  private SegmentCompletionProtocol.Request.Params _params;
+
+  private final Logger SEGMENT_LOGGER;
+
+  public SegmentCommitter(boolean isSplitCommit, String segmentNameStr, ServerSegmentCompletionProtocolHandler protocolHandler,
```
mcvsubbu: Do not include prevResponse here. See comments above

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SegmentCommitter.java
```diff
@@ -0,0 +1,108 @@
...
+      IndexLoadingConfig indexLoadingConfig, SegmentCompletionProtocol.Response prevResponse, SegmentCompletionProtocol.Request.Params params) {
+    _isSplitCommit = isSplitCommit;
+    _segmentNameStr = segmentNameStr;
+    _protocolHandler = protocolHandler;
+    _indexLoadingConfig = indexLoadingConfig;
+    _prevResponse = prevResponse;
+    _params = params;
+
+    SEGMENT_LOGGER = LoggerFactory.getLogger(LLRealtimeSegmentDataManager.class.getName() + "_" + _segmentNameStr);
```
mcvsubbu: this is already computed in LLRealtimeSegmentDataManager. Can we pass this in the constructor?

PR #4593, opened 2019-09-06
Do not update download URL and crypter when segment is not refreshed

Jackie-Jiang:
When segment is not refreshed (same crc so no need to refresh),
segment file won't be persisted, so should not change download URL
and crypter field in ZK metadata

### DIFF
--- a/pinot-controller/src/test/java/org/apache/pinot/controller/api/upload/ZKOperatorTest.java
+++ b/pinot-controller/src/test/java/org/apache/pinot/controller/api/upload/ZKOperatorTest.java
@@ -1,0 +1,141 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


Jackie-Jiang:
@icefury71 Good suggestion, added.

pinot-controller/src/test/java/org/apache/pinot/controller/api/upload/ZKOperatorTest.java
```diff
@@ -0,0 +1,141 @@
...
+    assertTrue(pushTime > 0);
+    assertEquals(segmentZKMetadata.getRefreshTime(), Long.MIN_VALUE);
+    assertEquals(segmentZKMetadata.getDownloadUrl(), "downloadUrl");
+    assertEquals(segmentZKMetadata.getCrypterName(), "crypter");
+
+    // Refresh the segment with unmatched IF_MATCH field
+    when(httpHeaders.getHeaderString(HttpHeaders.IF_MATCH)).thenReturn("123");
+    try {
+      zkOperator.completeSegmentOperations(RAW_TABLE_NAME, segmentMetadata, null, null, false, httpHeaders,
+          "otherDownloadUrl", false);
```
jackjlli: Shouldn't we use "downloadUrl" instead of "otherDownloadUrl" because it is not even checked yet?

Jackie-Jiang: Good catch :P

PR #4594, opened 2019-09-06
[TE] fix create alert error message null pointer exception

jihaozh:
This PR fixes the create alert error message null pointer exception.


xiaohui-sun:
Thanks for the fix @jihaozh. @harleyjj can you also make the error message copyable? I have seen several tickets complained they could not copy the message.

PR #4595, opened 2019-09-07
Implement all transform function APIs for MapValueTransformFunction

Jackie-Jiang:
- Enhance BaseTransformFunction to support bytes type and fetching values from dictionary
- Enhance MapValueTransformFunction to support filter with non-existing key, and use binary search
- Fix AvroUtils Map handling to sort keys based on the key data type
- Change MapValue function name to "mapValue" to follow the convention
- Add documentation for MapValue transform function
- Add more checks on the arguments for MapValue transform function
- Enhance the mapValue transform function test
- Simplify the logic for some transform functions (no need to check data type)
- Remove @nonnull annotation for transform functions

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/MapValueTransformFunction.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/MapValueTransformFunction.java
@@ -18,9 +18,10 @@
  */
 package org.apache.pinot.core.operator.transform.function;
 
+import com.google.common.base.Preconditions;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import javax.annotation.Nonnull;
 import org.apache.pinot.core.common.DataSource;
 import org.apache.pinot.core.operator.blocks.ProjectionBlock;
 import org.apache.pinot.core.operator.transform.TransformResultMetadata;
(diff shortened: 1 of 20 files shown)


pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/MapValueTransformFunction.java
```diff
@@ -29,20 +31,25 @@
...
 
 /**
- * map_value(keyColName, 'keyName', valColName)
+ * The MAP_VALUE transform function takes 3 arguments: keyColumn, keyValue, valueColumn, where keyColumn and valueColumn
+ * are dictionary-encoded multi-value columns, and keyValue must be a literal (number or string). In order to make
+ * MAP_VALUE transform function work, the keyValue provided must exist in the keyColumn. To ensure that, the query can
+ * have a filter on the keyColumn, e.g. SELECT mapValue(key, 'myKey', value) FROM myTable WHERE key = 'myKey'.
  */
 public class MapValueTransformFunction extends BaseTransformFunction {
+  public static final String FUNCTION_NAME = "mapValue";
```
kishoreg: why camel case. its common to use _ as the separator in sql udf's. See presto udf's for example https://prestodb.github.io/docs/current/functions/string.html

Jackie-Jiang: We already have `valueIn`, `timeConvert`, `dateTimeConvert`. Prefer keeping them the same way.

PR #4601, opened 2019-09-09
[TE] Exclude template/blacklist emails before sending out alert

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/scheme/DetectionEmailAlerter.java
```diff
@@ -106,6 +116,16 @@ private void whitelistRecipients(DetectionAlertFilterRecipients recipients) {
     }
   }
 
+  private void blacklistRecipients(DetectionAlertFilterRecipients recipients) {
+    if (recipients != null) {
```
xiaohui-sun: Combine two ifs.

PR #4602, opened 2019-09-10
First pass of GROUP BY with ORDER BY support

npawar:
This PR contains the implementation of ORDER BY support in group by.

In this first pass, the changes have been done from `CombineGroupByOrderByOperator` upwards. The AggregationGroupByOperator hasn't been changed.

`IndexedTable` is used wherever possible (to merge results in CombineGroupByOrderByOperator, and then to reduce results across servers in the BrokerReduceService)

`ResultTable` has been introduced, as a standard way to return results to the client.

2 `queryOptions` have been introduced:
1. groupByMode - pql/sql - whether to execute the group by in PQL style (split all aggregations and ignore order by) or standard SQL style
2. responseFormat - pql/sql - whether to present results using List<AggregationResults> (the PQL way), or use ResultTable which is closer to the SQL way.
By default, the modes are PQL, PQL
In order to get the order by results in ResultTable, modes should be SQL,SQL
In order to get the order by results, but in List<AggregationResult>, modes should be SQL,PQL
These modes can be added to the JSON payload:
`curl -H "Content-Type: application/json" -X POST -d '{"sql":"select count(*) from table group by dim1 order by dim1","queryOptions":"groupByMode=sql;responseFormat=sql"}' http://localhost:8099/query`

Pending: Benchmarking.
A comparison should be done of `SELECT agg1 FROM table GROUP BY group1, group2 ORDER by agg1 DESC` with the performance of the original `SELECT agg1 FROM table GROUP BY group1, group2` as the results are expected to be identical.
We can also compare `SELECT agg1,agg2... FROM table GROUP BY group1, group2 ORDER by agg1 DESC` with the performance of the original `SELECT agg1,agg2... FROM table GROUP BY group1, group2`. The groups will be different in the latter, but it is comparable in terms of result size.

Next steps: Push IndexedTable down into the AggregationGroupByOperator. We can introduce new operators, for each strategy we're trying out (1 ConcurrentIndexedTable, multiple SimpleIndexedTable, etc)

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -60,10 +60,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static org.apache.pinot.common.utils.CommonConstants.Broker.*;
-import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.DEBUG_OPTIONS;
-import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.PQL;
-import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.SQL;
-import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.TRACE;
+import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.*;
 
 
 @ThreadSafe
@@ -227,11 +224,17 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
       LOGGER.debug("Enable trace for request {}: {}", requestId, query);
       brokerRequest.setEnableTrace(true);
     }
+
     if (request.has(DEBUG_OPTIONS)) {
-      Map<String, String> debugOptions = Splitter.on(';').omitEmptyStrings().trimResults().withKeyValueSeparator('=')
-          .split(request.get(DEBUG_OPTIONS).asText());
+      Map<String, String> debugOptions = getOptionsFromRequest(request, DEBUG_OPTIONS);
       LOGGER.debug("Debug options are set to: {} for request {}: {}", debugOptions, requestId, query);
       brokerRequest.setDebugOptions(debugOptions);
+    }
+
+    if (request.has(QUERY_OPTIONS)) {
+      Map<String, String> queryOptions = getOptionsFromRequest(request, QUERY_OPTIONS);
+      LOGGER.debug("Query options are set to: {} for request {}: {}", queryOptions, requestId, query);
+      brokerRequest.setQueryOptions(queryOptions);
     }
 
     // Optimize the query
@@ -338,6 +341,14 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
       _numDroppedLog.incrementAndGet();
     }
     return brokerResponse;
+  }
+
+  private Map<String, String> getOptionsFromRequest(JsonNode request, String optionsKey) {
+    return Splitter.on(';')
+        .omitEmptyStrings()
+        .trimResults()
+        .withKeyValueSeparator('=')
+        .split(request.get(optionsKey).asText());
   }
 
   private PinotQueryRequest getPinotQueryRequest(JsonNode request) {
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -60,7 +60,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static org.apache.pinot.common.utils.CommonConstants.Broker.*;
-import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.*;
 
 
 @ThreadSafe
@@ -220,19 +219,19 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     }
 
     // Set extra settings into broker request
-    if (request.has(TRACE) && request.get(TRACE).asBoolean()) {
+    if (request.has(Request.TRACE) && request.get(Request.TRACE).asBoolean()) {
       LOGGER.debug("Enable trace for request {}: {}", requestId, query);
       brokerRequest.setEnableTrace(true);
     }
 
-    if (request.has(DEBUG_OPTIONS)) {
-      Map<String, String> debugOptions = getOptionsFromRequest(request, DEBUG_OPTIONS);
+    if (request.has(Request.DEBUG_OPTIONS)) {
+      Map<String, String> debugOptions = getOptionsFromRequest(request, Request.DEBUG_OPTIONS);
       LOGGER.debug("Debug options are set to: {} for request {}: {}", debugOptions, requestId, query);
       brokerRequest.setDebugOptions(debugOptions);
     }
 
-    if (request.has(QUERY_OPTIONS)) {
-      Map<String, String> queryOptions = getOptionsFromRequest(request, QUERY_OPTIONS);
+    if (request.has(Request.QUERY_OPTIONS)) {
+      Map<String, String> queryOptions = getOptionsFromRequest(request, Request.QUERY_OPTIONS);
       LOGGER.debug("Query options are set to: {} for request {}: {}", queryOptions, requestId, query);
       brokerRequest.setQueryOptions(queryOptions);
     }
@@ -352,10 +351,10 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   }
 
   private PinotQueryRequest getPinotQueryRequest(JsonNode request) {
-    if (request.has(SQL)) {
-      return new PinotQueryRequest(SQL, request.get(SQL).asText());
+    if (request.has(Request.SQL)) {
+      return new PinotQueryRequest(Request.SQL, request.get(Request.SQL).asText());
     }
-    return new PinotQueryRequest(PQL, request.get(PQL).asText());
+    return new PinotQueryRequest(Request.PQL, request.get(Request.PQL).asText());
   }
 
   /**
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -55,11 +55,10 @@ import org.apache.pinot.common.request.FilterQueryMap;
 import org.apache.pinot.common.response.BrokerResponse;
 import org.apache.pinot.common.response.broker.BrokerResponseNative;
 import org.apache.pinot.common.utils.CommonConstants;
+import org.apache.pinot.common.utils.CommonConstants.Broker;
 import org.apache.pinot.core.query.reduce.BrokerReduceService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import static org.apache.pinot.common.utils.CommonConstants.Broker.*;
 
 
 @ThreadSafe
@@ -96,12 +95,13 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     _queryQuotaManager = queryQuotaManager;
     _brokerMetrics = brokerMetrics;
 
-    _brokerId = config.getString(CONFIG_OF_BROKER_ID, getDefaultBrokerId());
-    _brokerTimeoutMs = config.getLong(CONFIG_OF_BROKER_TIMEOUT_MS, DEFAULT_BROKER_TIMEOUT_MS);
-    _queryResponseLimit = config.getInt(CONFIG_OF_BROKER_QUERY_RESPONSE_LIMIT, DEFAULT_BROKER_QUERY_RESPONSE_LIMIT);
-    _queryLogLength = config.getInt(CONFIG_OF_BROKER_QUERY_LOG_LENGTH, DEFAULT_BROKER_QUERY_LOG_LENGTH);
-    _queryLogRateLimiter = RateLimiter.create(
-        config.getDouble(CONFIG_OF_BROKER_QUERY_LOG_MAX_RATE_PER_SECOND, DEFAULT_BROKER_QUERY_LOG_MAX_RATE_PER_SECOND));
+    _brokerId = config.getString(Broker.CONFIG_OF_BROKER_ID, getDefaultBrokerId());
+    _brokerTimeoutMs = config.getLong(Broker.CONFIG_OF_BROKER_TIMEOUT_MS, Broker.DEFAULT_BROKER_TIMEOUT_MS);
+    _queryResponseLimit =
+        config.getInt(Broker.CONFIG_OF_BROKER_QUERY_RESPONSE_LIMIT, Broker.DEFAULT_BROKER_QUERY_RESPONSE_LIMIT);
+    _queryLogLength = config.getInt(Broker.CONFIG_OF_BROKER_QUERY_LOG_LENGTH, Broker.DEFAULT_BROKER_QUERY_LOG_LENGTH);
+    _queryLogRateLimiter = RateLimiter.create(config.getDouble(Broker.CONFIG_OF_BROKER_QUERY_LOG_MAX_RATE_PER_SECOND,
+        Broker.DEFAULT_BROKER_QUERY_LOG_MAX_RATE_PER_SECOND));
 
     _numDroppedLog = new AtomicInteger(0);
     _numDroppedLogRateLimiter = RateLimiter.create(1.0);
@@ -219,19 +219,19 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
     }
 
     // Set extra settings into broker request
-    if (request.has(Request.TRACE) && request.get(Request.TRACE).asBoolean()) {
+    if (request.has(Broker.Request.TRACE) && request.get(Broker.Request.TRACE).asBoolean()) {
       LOGGER.debug("Enable trace for request {}: {}", requestId, query);
       brokerRequest.setEnableTrace(true);
     }
 
-    if (request.has(Request.DEBUG_OPTIONS)) {
-      Map<String, String> debugOptions = getOptionsFromRequest(request, Request.DEBUG_OPTIONS);
+    if (request.has(Broker.Request.DEBUG_OPTIONS)) {
+      Map<String, String> debugOptions = getOptionsFromRequest(request, Broker.Request.DEBUG_OPTIONS);
       LOGGER.debug("Debug options are set to: {} for request {}: {}", debugOptions, requestId, query);
       brokerRequest.setDebugOptions(debugOptions);
     }
 
-    if (request.has(Request.QUERY_OPTIONS)) {
-      Map<String, String> queryOptions = getOptionsFromRequest(request, Request.QUERY_OPTIONS);
+    if (request.has(Broker.Request.QUERY_OPTIONS)) {
+      Map<String, String> queryOptions = getOptionsFromRequest(request, Broker.Request.QUERY_OPTIONS);
       LOGGER.debug("Query options are set to: {} for request {}: {}", queryOptions, requestId, query);
       brokerRequest.setQueryOptions(queryOptions);
     }
@@ -351,10 +351,10 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
   }
 
   private PinotQueryRequest getPinotQueryRequest(JsonNode request) {
-    if (request.has(Request.SQL)) {
-      return new PinotQueryRequest(Request.SQL, request.get(Request.SQL).asText());
+    if (request.has(Broker.Request.SQL)) {
+      return new PinotQueryRequest(Broker.Request.SQL, request.get(Broker.Request.SQL).asText());
     }
-    return new PinotQueryRequest(Request.PQL, request.get(Request.PQL).asText());
+    return new PinotQueryRequest(Broker.Request.PQL, request.get(Broker.Request.PQL).asText());
   }
 
   /**
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
@@ -32,8 +32,8 @@ import java.util.List;
  */
 @JsonPropertyOrder({"columns", "rows"})
 public class ResultTable {
-  final private List<String> _columns;
-  final private List<Serializable[]> _rows;
+  private final List<String> _columns;
+  private final List<Serializable[]> _rows;
 
   @JsonCreator
   public ResultTable(@JsonProperty("columns") List<String> columns,
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
@@ -1,0 +1,62 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.pinot.common.response.broker;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import java.io.Serializable;
+import java.util.List;
+
+
+/**
+ * Hosts the results in a standard tabular structure
+ */
+@JsonPropertyOrder({"columns", "results"})
+public class ResultTable {
+  private List<String> _columns;
+  private List<Serializable[]> _rows;
+
+  @JsonCreator
+  public ResultTable(@JsonProperty("columns") List<String> columns,
+      @JsonProperty("results") List<Serializable[]> results) {
+    _columns = columns;
+    _rows = results;
+  }
+
+  @JsonProperty("columns")
+  public List<String> getColumns() {
+    return _columns;
+  }
+
+  @JsonProperty("columns")
+  public void setColumns(List<String> columns) {
+    _columns = columns;
+  }
+
+  @JsonProperty("results")
+  public List<Serializable[]> getRows() {
+    return _rows;
+  }
+
+  @JsonProperty("results")
+  public void setRows(List<Serializable[]> rows) {
+    _rows = rows;
+  }
+}
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
@@ -28,16 +28,16 @@ import java.util.List;
 /**
  * Hosts the results in a standard tabular structure
  */
-@JsonPropertyOrder({"columns", "results"})
+@JsonPropertyOrder({"columns", "rows"})
 public class ResultTable {
   private List<String> _columns;
   private List<Serializable[]> _rows;
 
   @JsonCreator
   public ResultTable(@JsonProperty("columns") List<String> columns,
-      @JsonProperty("results") List<Serializable[]> results) {
+      @JsonProperty("rows") List<Serializable[]> rows) {
     _columns = columns;
-    _rows = results;
+    _rows = rows;
   }
 
   @JsonProperty("columns")
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
@@ -30,8 +30,8 @@ import java.util.List;
  */
 @JsonPropertyOrder({"columns", "rows"})
 public class ResultTable {
-  private List<String> _columns;
-  private List<Serializable[]> _rows;
+  final private List<String> _columns;
+  final private List<Serializable[]> _rows;
 
   @JsonCreator
   public ResultTable(@JsonProperty("columns") List<String> columns,
@@ -45,18 +45,8 @@ public class ResultTable {
     return _columns;
   }
 
-  @JsonProperty("columns")
-  public void setColumns(List<String> columns) {
-    _columns = columns;
-  }
-
   @JsonProperty("results")
   public List<Serializable[]> getRows() {
     return _rows;
-  }
-
-  @JsonProperty("results")
-  public void setRows(List<Serializable[]> rows) {
-    _rows = rows;
   }
 }
--- a/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
@@ -26,7 +26,9 @@ import java.util.List;
 
 
 /**
- * Hosts the results in a standard tabular structure
+ * Holds the results in a standard tabular structure
+ *
+ * FIXME: Column types and Multi-value support are missing, and deserialize might not work properly
  */
 @JsonPropertyOrder({"columns", "rows"})
 public class ResultTable {
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
@@ -171,9 +171,12 @@ public class CommonConstants {
       public static final String SQL = "sql";
       public static final String TRACE = "trace";
       public static final String DEBUG_OPTIONS = "debugOptions";
+      public static final String QUERY_OPTIONS = "queryOptions";
 
       public static class QueryOptionKey {
         public static final String PRESERVE_TYPE = "preserveType";
+        public static final String RESPONSE_FORMAT = "responseFormat";
+        public static final String GROUP_BY_MODE = "groupByMode";
       }
     }
   }
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/order/OrderByUtils.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/order/OrderByUtils.java
@@ -165,6 +165,8 @@ public final class OrderByUtils {
       String column = orderByInfo.getColumn();
       boolean ascending = orderByInfo.isIsAsc();
 
+      // TODO: avoid the index computation and index lookup in the comparison.
+      // we can achieve this by making order by operate on Object[], which contains only order by fields
       if (keyIndexMap.containsKey(column)) {
         int index = keyIndexMap.get(column);
         ColumnDataType columnDataType = keyColumnDataTypeMap.get(column);
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/order/OrderByUtils.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/order/OrderByUtils.java
@@ -38,6 +38,8 @@ import org.apache.pinot.core.query.aggregation.function.AggregationFunctionUtils
  */
 public final class OrderByUtils {
 
+  public static final int NUM_RESULTS_LOWER_LIMIT = 5000;
+
   private OrderByUtils() {
   }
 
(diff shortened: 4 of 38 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -224,13 +221,19 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
...
+
     if (request.has(DEBUG_OPTIONS)) {
-      Map<String, String> debugOptions = Splitter.on(';').omitEmptyStrings().trimResults().withKeyValueSeparator('=')
-          .split(request.get(DEBUG_OPTIONS).asText());
+      Map<String, String> debugOptions = getOptionsFromRequest(request, DEBUG_OPTIONS);
       LOGGER.debug("Debug options are set to: {} for request {}: {}", debugOptions, requestId, query);
       brokerRequest.setDebugOptions(debugOptions);
     }
 
+    if (request.has(QUERY_OPTIONS)) {
```
kishoreg: Is this the only place we are reading query options?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -171,9 +171,12 @@ public ServerType getServerType() {
       public static final String SQL = "sql";
       public static final String TRACE = "trace";
       public static final String DEBUG_OPTIONS = "debugOptions";
+      public static final String QUERY_OPTIONS = "queryOptions";
```
kishoreg: I thought we had this to support partition @mayankshriv ?

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -101,31 +108,51 @@ public int size() {
...
+      List<Record> sortedList = new ArrayList<>(_lookupMap.values());
+      sortedList.sort(_orderByComparator);
+      return sortedList.iterator();
+    }
     return _lookupMap.values().iterator();
   }
 
   private void resize(int trimToSize) {
 
     if (_lookupMap.size() > trimToSize) {
```
kishoreg: collect some metrics in this table such as num resizes, time taken to resize etc that can be printed/emitted in finish method to understand the perf bottlenecks

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -39,17 +41,22 @@
...
 
   private ConcurrentMap<Key, Record> _lookupMap;
-  private Comparator<Record> _minHeapComparator;
   private ReentrantReadWriteLock _readWriteLock;
 
+  private Comparator<Record> _minHeapComparator;
+  private Comparator<Record> _orderByComparator;
+
   @Override
   public void init(@Nonnull DataSchema dataSchema, List<AggregationInfo> aggregationInfos, List<SelectionSort> orderBy,
```
kishoreg: javadoc for this method?

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -101,31 +108,51 @@ public int size() {
...
 
-      for (Record evictRecord : minHeap) {
-        _lookupMap.remove(evictRecord.getKey());
+        for (Record evictRecord : minHeap) {
+          _lookupMap.remove(evictRecord.getKey());
+        }
+      } else {
+        // drop randomly
+
+        int numRecordsToDrop = _lookupMap.size() - trimToSize;
```
kishoreg: its better to not add new keys if there is no orderby rt

pinot-core/src/main/java/org/apache/pinot/core/operator/CombineGroupByOrderByOperator.java
```diff
@@ -0,0 +1,249 @@
...
+            if (aggregationGroupByResult != null) {
+              // Iterate over the group-by keys, for each key, update the group-by result in the indexedTable.
+              Iterator<GroupKeyGenerator.GroupKey> groupKeyIterator = aggregationGroupByResult.getGroupKeyIterator();
+              while (groupKeyIterator.hasNext()) {
+                GroupKeyGenerator.GroupKey groupKey = groupKeyIterator.next();
+                String[] stringKey = groupKey._stringKey.split(GroupKeyGenerator.DELIMITER);
+                Object[] objectKey = new Object[numGroupBy];
+                for (int i = 0; i < stringKey.length; i++) {
+                  DataSchema.ColumnDataType columnDataType = _dataSchema.getColumnDataType(i);
+                  switch (columnDataType) {
```
kishoreg: is it possible to avoid the switch statement for each row? looks like the dataSchema remains the same within the loop. we can probably have an interface to convert string to Object and have the right implementations set in advance?

pinot-core/src/main/java/org/apache/pinot/core/operator/blocks/IntermediateResultsBlock.java
```diff
@@ -224,11 +247,62 @@ public DataTable getDataTable()
...
+      }
+      dataTableBuilder.finishRow();
+    }
+    DataTable dataTable = dataTableBuilder.build();
+    return attachMetadataToDataTable(dataTable);
+  }
+
+  private void setDataTableColumn(ColumnDataType columnDataType, DataTableBuilder dataTableBuilder, int columnIndex, Object value)
+      throws IOException {
+    switch (columnDataType) {
```
kishoreg: same here, avoid switch if possible

pinot-core/src/main/java/org/apache/pinot/core/plan/CombinePlanNode.java
```diff
@@ -141,6 +146,12 @@ public Operator run() {
     // TODO: use the same combine operator for both aggregation and selection query.
     if (_brokerRequest.isSetAggregationsInfo() && _brokerRequest.getGroupBy() != null) {
       // Aggregation group-by query
+      Map<String, String> queryOptions = _brokerRequest.getQueryOptions();
+      // new Combine operator only when GROUP_BY_MODE explicitly set to SQL
+      if (queryOptions != null && SQL.equals(queryOptions.get(QueryOptionKey.GROUP_BY_MODE))) {
```
kishoreg: case insensitive check?

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
npawar: Yes, I didn't find any other place, for all other options as well

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
npawar: True. Would it be better to make the Operator keep track and stop adding more? Or let the operator keep adding everything, but make the table intelligent to ignore them?
Pro of the first approach is we don't add extra records. However, the operators have to do special handling depending on order by/no order by.
Pro of the second approach is operators need not know anything special. All intelligence is at the table. However, we will read some extra rows, create Record object, Key object, spend time converting raw values, only to ignore the values.

Seems like 1st would be more efficient.

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
npawar: Done. Printed num resizes, total time resizing, and avg resize time in finish.

pinot-core/src/main/java/org/apache/pinot/core/operator/CombineGroupByOrderByOperator.java
npawar: Fixed in CombineGroupByOperator and BrokerReduce service, please take a look. 
However, I'm not able to achieve this in IntermediateResultsBlock, because that one is conversion from Object -> primitive data types. Any suggestions?

npawar:
> Looks good overall. Few places where we can optimize further but let us do a benchmark before optimizing further.
> 
> I am assuming this code will only be triggered with query option =SQL

That's right. The if check in CombinePlanNode for `is groupByMode SQL` and then the same check in `BrokerReduceService::reduceOnDataTable` will ensure this only kicks in when `groupByMode=sql` is set in the json payload

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
kishoreg: not sure if we can simply ignore new rows once the limit is reached. We can only ignore new keys, we still need to update values for existing keys right?

pinot-core/src/main/java/org/apache/pinot/core/operator/CombineGroupByOrderByOperator.java
kishoreg: thanks. will look into that.

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
npawar: Ahh yes.. In that case, we either need the table to support `get(key)`, or we need to make the upsert check `if (isNewKey && isCapacityFull && !isOrderBy)`

npawar: Handled this. Please take a look. @mayankshriv FYI

pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
```diff
@@ -0,0 +1,62 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import java.io.Serializable;
+import java.util.List;
+
+
+/**
+ * Hosts the results in a standard tabular structure
+ */
+@JsonPropertyOrder({"columns", "results"})
```
mayankshriv: Nit: why not s/results/rows, especially when the member variable is called _rows?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
mayankshriv: I don't recall using it for partitioning. But seems this got introduced in: https://github.com/apache/incubator-pinot/pull/1490

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -19,37 +19,65 @@
...
-    super.init(dataSchema, aggregationInfos, orderBy, maxCapacity);
+      int maxCapacity, boolean sort) {
+    super.init(dataSchema, aggregationInfos, orderBy, maxCapacity, sort);
 
-    _minHeapComparator = OrderByUtils.getKeysAndValuesComparator(dataSchema, orderBy, aggregationInfos).reversed();
     _lookupMap = new ConcurrentHashMap<>();
     _readWriteLock = new ReentrantReadWriteLock();
+    _isOrderBy = CollectionUtils.isNotEmpty(orderBy);
+    if (_isOrderBy) {
+      _minHeapComparator = OrderByUtils.getKeysAndValuesComparator(dataSchema, orderBy, aggregationInfos).reversed();
```
mayankshriv: Will this support order-by with ASC/DESC?

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -204,18 +215,25 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
...
-            .getSelectionColumns(brokerRequest.getSelections().getSelectionColumns(), cachedDataSchema);
-        brokerResponseNative.setSelectionResults(new SelectionResults(selectionColumns, new ArrayList<>(0)));
+        if (brokerRequest.isSetSelections()) {
+          List<String> selectionColumns =
+              SelectionOperatorUtils.getSelectionColumns(brokerRequest.getSelections().getSelectionColumns(),
+                  cachedDataSchema);
+          brokerResponseNative.setSelectionResults(new SelectionResults(selectionColumns, new ArrayList<>(0)));
+        } else if (brokerRequest.isSetGroupBy() && queryOptions != null && SQL.equalsIgnoreCase(
+            queryOptions.get(QueryOptionKey.GROUP_BY_MODE)) && SQL.equalsIgnoreCase(
+            queryOptions.get(QueryOptionKey.RESPONSE_FORMAT))) {
```
mayankshriv: Is RESPONSE_FORMAT something that user specifies? If so, are we support query type SQL and response type PQL, and other such combinations? Not sure how that would useful, why not just dictate response based on query type (in which case we don't need this extra option)?

pinot-core/src/main/java/org/apache/pinot/core/operator/query/AggregationGroupByOperator.java
```diff
@@ -58,6 +64,34 @@ public AggregationGroupByOperator(@Nonnull AggregationFunctionContext[] function
     _transformOperator = transformOperator;
     _numTotalRawDocs = numTotalRawDocs;
     _useStarTree = useStarTree;
+
+    int numColumns = groupBy.getExpressionsSize() + _functionContexts.length;
```
mayankshriv: Seems like this is a change in the default path, purely due to signature change for constructor of IntermediateResultBlock? Ideally, I'd want to have no changes to default path while this project is still on-going. Of course, we can take smart risks depending on how the change. I'll let you assess that for this case.

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -246,22 +264,46 @@ public BrokerResponseNative reduceOnDataTable(@Nonnull BrokerRequest brokerReque
...
-              AggregationFunctionUtils.getAggregationFunctionsSelectStatus(brokerRequest.getAggregationsInfo());
-          setGroupByHavingResults(brokerResponseNative, aggregationFunctions, aggregationFunctionSelectStatus,
-              brokerRequest.getGroupBy(), dataTableMap, brokerRequest.getHavingFilterQuery(),
-              brokerRequest.getHavingFilterSubQueryMap(), preserveType);
-          if (brokerMetrics != null && (!brokerResponseNative.getAggregationResults().isEmpty())) {
-            // We emit the group by size when the result isn't empty. All the sizes among group-by results should be the same.
-            // Thus, we can just emit the one from the 1st result.
-            brokerMetrics.addMeteredQueryValue(brokerRequest, BrokerMeter.GROUP_BY_SIZE,
-                brokerResponseNative.getAggregationResults().get(0).getGroupByResult().size());
+        } else { // Aggregation group-by query.
```
mayankshriv: May be factorize this as a utility function?

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -392,6 +434,218 @@ private void setAggregationResults(@Nonnull BrokerResponseNative brokerResponseN
...
+   * Extract the results of group by order by into a List of {@link AggregationResult}
+   * There will be 1 aggregation result per aggregation. The group by keys will be the same across all aggregations
+   * @param brokerResponseNative broker response
+   * @param dataSchema data schema
+   * @param aggregationInfos aggregations info
+   * @param groupBy group by info
+   * @param orderBy order by info
+   * @param dataTableMap map from server to data table
+   */
+  private void setPQLGroupByOrderByResults(@Nonnull BrokerResponseNative brokerResponseNative,
```
mayankshriv: Is this needed because we want to support order-by with PQL?

pinot-common/src/main/java/org/apache/pinot/common/response/broker/ResultTable.java
npawar: done

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
npawar: Yes it will. It gets all necessary information from the List<SelectionSOrt> and constructs appropriate comparator. Some tests were included in the previous PR where the Table was introduced

npawar: Yes. All information is collected from List<SelectionSort>. OrderByUtilsTest has tests about it

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
npawar: Yes, user can specify `"queryOptions":"groupByMode=sql;responseFormat=sql"`
By default we will be `pql,pql`
`sql,sql` is the end state
In the interim, user can do `sql,pql`.
`pql,sql` is the only invalid combination, and it will default to `pql,pql`.

This is based on our (Mayank, Kishore, Neha) conversation, where we discussed that only allowing it on `sql:query..` won't be good, for the following reasons:
1. We are not sure how long it will take us to completely get to sql. Use will have to keep changing payload `pql:query` to `sql:query` and back, depending on query.  It's more flexible with options. It gives us a clean path from `pql,pql` to `sql,sql`.
2. Enforcing that only "sql" payload will trigger this, will bring in other problems, as we saw with the parser (Date column is a special keyword in sql) etc.

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
npawar: this is required for responseFormat=pql

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
npawar: Done

npawar: Done

pinot-core/src/main/java/org/apache/pinot/core/operator/query/AggregationGroupByOperator.java
npawar: I personally feel this much change is fine. If you/everyone feel strongly about not touching existing path, I'll separate it out

npawar:
> Will performance numbers be shared in this PR, or following PR's?

I'll include performance numbers in this PR

npawar:
Added a JMH Benchmark in the last commit. Simulated the logic in CombineGroupByOperator.
4 segments inserting 100k records each, concurrently.

``BenchmarkCombineGroupBy``:
``benchmarkConcurrentIndexedTableInCombineGroupBy`` - inserting into ConcurrentIndexedTable like in the PR code
``benchmarkOriginalCombineGroupBy`` - inserting into ConcurrentHashMap like in the original CombineGroupByOperator

Conclusions:
1. New code does at par/better in scenarios 
- No order by
- Order by, but buffer is so large that resize never triggered
2. In the case where it didn’t work, the bottleneck is clearly resizing
Screen Shot 2019-09-27 at 9 53 57 AM

pinot-core/src/main/java/org/apache/pinot/core/operator/blocks/IntermediateResultsBlock.java
```diff
@@ -45,7 +50,7 @@
  * The <code>IntermediateResultsBlock</code> class is the holder of the server side inter-segment results.
  */
 public class IntermediateResultsBlock implements Block {
-  private DataSchema _selectionDataSchema;
+  private DataSchema _dataSchema;
```
Jackie-Jiang: For aggregation, you don't really need data schema. Data schema is used to passing the data type for each column. For aggregation, the data type can be derived from the aggregation function.
For the column order, you can refer to the `SelectionOperatorUtils` where order-by expressions are always put in front of other expressions. With that you don't even need to maintain the column index.

Jackie-Jiang:
Also, for the benchmark number, why for several cases TOP 15k is faster than TOP 5k, which is then faster than TOP 500?

pinot-core/src/main/java/org/apache/pinot/core/data/order/OrderByUtils.java
```diff
@@ -109,12 +110,33 @@ private OrderByUtils() {
     return globalComparator;
   }
 
+  public static List<Integer> getAggregationIndexes(List<SelectionSort> orderBy, List<AggregationInfo> aggregationInfos) {
+    Map<String, Integer> aggregationColumnToIndex = new HashMap<>(aggregationInfos.size());
+    for (int i = 0; i < aggregationInfos.size(); i++) {
+      AggregationInfo aggregationInfo = aggregationInfos.get(i);
+      String aggregationColumn =
```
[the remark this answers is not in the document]
npawar: Done, added to AggregationFunctionUtil

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -19,37 +19,68 @@
...
-  private Comparator<Record> _minHeapComparator;
   private ReentrantReadWriteLock _readWriteLock;
 
+  private boolean _isOrderBy;
+  private Comparator<Record> _orderByComparator;
+  private Comparator<Record> _finalOrderByComparator;
+  private List<Integer> _aggregationIndexes;
+
+  private AtomicBoolean _noMoreNewRecords = new AtomicBoolean();
+  private LongAdder _numResizes = new LongAdder();
```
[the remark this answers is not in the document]
npawar: done

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -32,38 +32,44 @@
...
+    double bufferFactor;
+    /* Factor used to decide eviction threshold **/
+    /** The true capacity of the table is {@link IndexedTable::_bufferedCapacity},
+     * which is bufferFactor times the {@link IndexedTable::_maxCapacity}
+     *
+     * If records beyond {@link IndexedTable::_bufferedCapacity} are received,
+     * the table resize and evict bottom records, resizing it to {@link IndexedTable::_maxCapacity}
+     * The assumption here is that {@link IndexedTable::_maxCapacity} already has a buffer added by the caller (typically, we do max(top * 5, 5000))
+     */
+    if (maxCapacity > 50000) {
```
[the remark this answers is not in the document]
npawar: renamed to just capacity

pinot-core/src/main/java/org/apache/pinot/core/data/order/OrderByUtils.java
```diff
@@ -109,12 +110,33 @@ private OrderByUtils() {
     return globalComparator;
   }
 
+  public static List<Integer> getAggregationIndexes(List<SelectionSort> orderBy, List<AggregationInfo> aggregationInfos) {
```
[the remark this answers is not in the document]
npawar: It's not entirely similar. We can end up with 0 length list or upto numAggregations length list. To return int[], i'll still have to begin with list, and then make it an array, or scan twice to avoid that.

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -32,38 +32,44 @@
...
+  boolean _sort;
 
   int _maxCapacity;
-  int _evictCapacity;
   int _bufferedCapacity;
 
   @Override
   public void init(@Nonnull DataSchema dataSchema, List<AggregationInfo> aggregationInfos, List<SelectionSort> orderBy,
-      int maxCapacity) {
+      int maxCapacity, boolean sort) {
```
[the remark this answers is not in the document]
npawar: Sort tells us whether the caller is expecting the results sorted. For example, at server level, no need to have results sorted, but at broker level, we need them sorted.
Sort is better off in init, because it is currently needed in iterator() or in finish(), and maybe even elsewhere

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -61,25 +92,41 @@ public boolean upsert(@Nonnull Record newRecord) {
     Key key = newRecord.getKey();
     Preconditions.checkNotNull(key, "Cannot upsert record with null keys");
 
-    Record existingRecord = _lookupMap.putIfAbsent(key, newRecord);
-    if (existingRecord != null) {
-      _lookupMap.compute(key, (k, v) -> {
+    if (_noMoreNewRecords.get()) { // allow only existing record updates
+      _lookupMap.computeIfPresent(key, (k, v) -> {
```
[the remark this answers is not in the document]
npawar: Why is that necessary? computeIfPresent will lock the necessary buckets, ensuring same key is not operated on concurrently

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -61,25 +92,41 @@ public boolean upsert(@Nonnull Record newRecord) {
...
-    }
+    } else { // allow all records
 
-    // resize if exceeds capacity
-    if (_lookupMap.size() >= _bufferedCapacity) {
-      _readWriteLock.writeLock().lock();
-      try {
-        if (_lookupMap.size() >= _bufferedCapacity) {
-          resize(_evictCapacity);
+      Record existingRecord = _lookupMap.putIfAbsent(key, newRecord);
```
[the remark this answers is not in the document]
npawar: Several combinations were tried. This is the way I could ensure all the concurrency is handled correctly. It starts getting tricky with a single compute

pinot-core/src/main/java/org/apache/pinot/core/data/table/Table.java
```diff
@@ -37,9 +37,10 @@
    * @param aggregationInfos the aggregation info for the values if applicable
    * @param orderBy the order by information if applicable
    * @param maxCapacity the max capacity the table should have
+   * @param sort should the results be sorted in finish
    */
   void init(@Nonnull DataSchema dataSchema, List<AggregationInfo> aggregationInfos, List<SelectionSort> orderBy,
-      int maxCapacity);
+      int maxCapacity, boolean sort);
```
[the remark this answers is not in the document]
npawar: addressed above. sort flag is needed in iterator and in finish.

npawar:
> Also, for the benchmark number, why for several cases TOP 15k is faster than TOP 5k, which is then faster than TOP 500?

The code used in benchmark decided `capacity` and `bufferCapacity` entirely on the basis of `TOP`.  Capacity = Max(5 * TOP, 5k). As a result, more the TOP, lesser the number of resizes. That's why as TOP increased the performance improved. 
It has been since changed, to just have a very large `capacity`. We will focus on the optimization of resize after this PR is in.

pinot-core/src/main/java/org/apache/pinot/core/data/order/OrderByUtils.java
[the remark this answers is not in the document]
Jackie-Jiang: That is okay. The benefit of using int[] is the data access will be much faster. After calculating the indices, we need to access this array on a per value basis

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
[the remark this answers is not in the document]
Jackie-Jiang: I think you only need it in `finish()` right? Before `finish()` getting called, you never need sorted results

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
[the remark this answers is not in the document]
Jackie-Jiang: Why? the current code is equivalent to:
```
_lookupMap.compute(key, (k, v) -> {
  if (v == null) {
    return newRecord;
  } else {
    for (int i = 0; i < _aggregationFunctions.size(); i++) {
      v.getValues()[i] = _aggregationFunctions.get(i).merge(v.getValues()[i], newRecord.getValues()[i]);
    }
    return v;
  }
}
```

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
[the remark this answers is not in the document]
Jackie-Jiang: Then during resize, records can still be updated? That will cause issue right?

Jackie-Jiang:
> > Also, for the benchmark number, why for several cases TOP 15k is faster than TOP 5k, which is then faster than TOP 500?
> 
> The code used in benchmark decided `capacity` and `bufferCapacity` entirely on the basis of `TOP`. Capacity = Max(5 * TOP, 5k). As a result, more the TOP, lesser the number of resizes. That's why as TOP increased the performance improved.
> It has been since changed, to just have a very large `capacity`. We will focus on the optimization of resize after this PR is in.

Jackie-Jiang:
> > Also, for the benchmark number, why for several cases TOP 15k is faster than TOP 5k, which is then faster than TOP 500?
> 
> The code used in benchmark decided `capacity` and `bufferCapacity` entirely on the basis of `TOP`. Capacity = Max(5 * TOP, 5k). As a result, more the TOP, lesser the number of resizes. That's why as TOP increased the performance improved.
> It has been since changed, to just have a very large `capacity`. We will focus on the optimization of resize after this PR is in.

For the no resize case, TOP 500 should always be fastest right?

pinot-core/src/main/java/org/apache/pinot/core/operator/query/AggregationGroupByOperator.java
npawar: Separated this out into AggregationGroupByOrderBy PlanNode and Operator

pinot-core/src/main/java/org/apache/pinot/core/data/order/OrderByUtils.java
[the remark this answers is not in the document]
npawar: done

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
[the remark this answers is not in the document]
npawar: Refactored to pass sort flag in finish() instead of init()

pinot-core/src/main/java/org/apache/pinot/core/operator/blocks/IntermediateResultsBlock.java
npawar: I prefer not to make the change about reordering columns in this iteration. Can we take it up in following PRs? I've already noted it down in the list of possible optimizations.
The data schema is mainly needed for the Table structure.

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -19,37 +19,68 @@
...
-    super.init(dataSchema, aggregationInfos, orderBy, maxCapacity);
+      int maxCapacity, boolean sort) {
+    super.init(dataSchema, aggregationInfos, orderBy, maxCapacity, sort);
 
-    _minHeapComparator = OrderByUtils.getKeysAndValuesComparator(dataSchema, orderBy, aggregationInfos).reversed();
     _lookupMap = new ConcurrentHashMap<>();
     _readWriteLock = new ReentrantReadWriteLock();
+    _isOrderBy = CollectionUtils.isNotEmpty(orderBy);
+    if (_isOrderBy) {
+      _orderByComparator = OrderByUtils.getKeysAndValuesComparator(dataSchema, orderBy, aggregationInfos, false);
```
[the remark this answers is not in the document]
npawar: I agree that there is still another step for optimization. I've listed it down as next steps. I already optimized it by avoiding extractFinalResults in the comparison, and extracting before hand as per your suggestion. I'd like to do this additional optimization in a later PR to prevent the scope of this one from exploding too much. Also the aggregation functions which have expensive `extractFinalResults` are anyway not intermediate result comparable. The redundancy is just for count, sum, min, max, avg, minmax, but those are pretty light weight. Will certainly tackle this in the immediate next PR

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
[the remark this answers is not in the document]
npawar: True. Thanks for catching that. Added the read lock

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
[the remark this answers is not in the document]
npawar: Yes. I can't seem to remember why exactly I ended up changing it, because it was so long ago.
Changed it to a single compute.

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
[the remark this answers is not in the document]
mayankshriv: May be add a TODO here in the code as well?

pinot-core/src/main/java/org/apache/pinot/core/data/table/Record.java
```diff
@@ -43,4 +46,9 @@ public Key getKey() {
   public Object[] getValues() {
     return _values;
   }
+
+  @Override
+  public String toString() {
+    return _key.toString() + " " + Arrays.deepToString(_values);
```
mayankshriv: Will this String be used for de-serialization? And can key have whitespaces?

npawar: This String will never be used for de-serialization. We will always be using the granular values within the array for ser/deser.

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -428,6 +469,217 @@ private void setAggregationResults(BrokerResponseNative brokerResponseNative,
...
+      throws Throwable {
+
+    IndexedTable indexedTable = new ConcurrentIndexedTable();
+    int indexedTableCapacity = 1_000_000;
+    // FIXME: indexedTableCapacity should be derived from TOP. Hardcoding this value to a higher number until we can tune the resize
+    // int capacity = GroupByUtils.getTableCapacity((int) groupBy.getTopN());
+    indexedTable.init(dataSchema, aggregationInfos, orderBy, indexedTableCapacity);
+
+    for (DataTable dataTable : dataTableMap.values()) {
+      CheckedFunction2[] functions = new CheckedFunction2[dataSchema.size()];
```
Jackie-Jiang: Is this a commonly used interface? `BiFunction` from `java.util.function` might be better?

npawar: Changed. Thanks for the suggestion. BiFunction is better

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -202,10 +210,10 @@ public BrokerResponseNative reduceOnDataTable(BrokerRequest brokerRequest,
     String rawTableName = TableNameBuilder.extractRawTableName(tableName);
     if (brokerMetrics != null) {
       brokerMetrics.addMeteredTableValue(rawTableName, BrokerMeter.DOCUMENTS_SCANNED, numDocsScanned);
-      brokerMetrics
-          .addMeteredTableValue(rawTableName, BrokerMeter.ENTRIES_SCANNED_IN_FILTER, numEntriesScannedInFilter);
-      brokerMetrics
-          .addMeteredTableValue(rawTableName, BrokerMeter.ENTRIES_SCANNED_POST_FILTER, numEntriesScannedPostFilter);
+      brokerMetrics.addMeteredTableValue(rawTableName, BrokerMeter.ENTRIES_SCANNED_IN_FILTER,
```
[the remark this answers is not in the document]
npawar: Ah yes, changed it and reverted the formatting introduced by LinkedIn style

pinot-core/src/main/java/org/apache/pinot/core/operator/blocks/IntermediateResultsBlock.java
npawar: As per offline discussion, added TODO

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -101,37 +125,72 @@ public int size() {
...
+      _resizeTime.accumulate(timeElapsed);
     }
   }
 
   @Override
   public void finish() {
     resize(_maxCapacity);
+    long numResizes = _numResizes.sum();
+    long resizeTime = _resizeTime.get();
+    LOGGER.info("Num resizes : {}, Total time spent in resizing : {}, Avg resize time : {}", numResizes, resizeTime,
```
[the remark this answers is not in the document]
npawar: Changed it to debug.
Emitting a metric is tricky. This class can be created by the Server or Broker. Accordingly we will need access to BrokerMetrics or ServerMetrics. And it's a lot of wiring in to get these objects here..

npawar:

PR #4605, opened 2019-09-11
[TE] Add anomaly type for merged anomaly result

xiaohui-sun:
We will support multiple types of anomalies in TE.
Adding an anomaly type explicitly now.


akshayrai:
I don't see the AnomalyType being updated? Do you plan to send a followup PR?

xiaohui-sun:
Will return default anomaly type if it is not set.

PR #3705, opened 2019-01-17
Add Pinot code style

Jackie-Jiang:
Add code style config for eclipse and intellij
Add checkstyle config and maven plugin
All code style configs are borrowed from LinkedIn style

Command for checkstyle plugin: mvn checkstyle:check

Import code style to Intellij: https://www.jetbrains.com/help/idea/settings-code-style.html
1. Navigate to Preferences -> Editor -> Code Style -> Java
2. Select Import Scheme -> Intellij IDES code style XML
3. Choose codestyle-intellij.xml

PR #4608, opened 2019-09-12
Unit tests and bug fixes for DeleteTable rest API for controller.

chenboat:
(1) Add a unit test for DeleteTable rest API for controller.
(2) Fix an issue where the existing API always returns table deleted even the table does not exist.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -254,24 +254,39 @@ public class PinotTableRestletResource {
 
     List<String> tablesDeleted = new LinkedList<>();
     try {
-      if (tableType != TableType.REALTIME && !TableNameBuilder.REALTIME.tableHasTypeSuffix(tableName)
-          && _pinotHelixResourceManager.hasOfflineTable(tableName)) {
+      if (toDeleteOfflineTable(tableName, tableType) && _pinotHelixResourceManager
+          .hasOfflineTable(tableName)) {
         _pinotHelixResourceManager.deleteOfflineTable(tableName);
         tablesDeleted.add(TableNameBuilder.OFFLINE.tableNameWithType(tableName));
       }
-      if (tableType != TableType.OFFLINE && !TableNameBuilder.OFFLINE.tableHasTypeSuffix(tableName)
-          && _pinotHelixResourceManager.hasRealtimeTable(tableName)) {
+      if (toDeleteRealtimeTable(tableName, tableType) && _pinotHelixResourceManager
+          .hasRealtimeTable(tableName)) {
         _pinotHelixResourceManager.deleteRealtimeTable(tableName);
         tablesDeleted.add(TableNameBuilder.REALTIME.tableNameWithType(tableName));
       }
       if (tablesDeleted.size() == 0) {
         throw new ControllerApplicationException(LOGGER, "Table '" + tableName + "' does not exist",
-            Response.Status.BAD_REQUEST);
+            Response.Status.NOT_FOUND);
       }
       return new SuccessResponse("Tables: " + tablesDeleted + " deleted");
     } catch (Exception e) {
-      throw new ControllerApplicationException(LOGGER, e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR, e);
+      throw new ControllerApplicationException(LOGGER, e.getMessage(),
+          Response.Status.INTERNAL_SERVER_ERROR, e);
     }
+  }
+
+  // Return true if tableType is NOT offline (i.e., null or realtime) and table name does  not end
+  // with _offline suffix.
+  private boolean toDeleteRealtimeTable(String tableName, TableType tableType) {
+    return tableType != TableType.OFFLINE && !TableNameBuilder.OFFLINE
+        .tableHasTypeSuffix(tableName);
+  }
+
+  // Return true if tableType is NOT realtime (i.e., null or offline) and table name does  not end
+  // with _realtime suffix.
+  private boolean toDeleteOfflineTable(String tableName, TableType tableType) {
+    return tableType != TableType.REALTIME && !TableNameBuilder.REALTIME
+        .tableHasTypeSuffix(tableName);
   }
 
   @PUT
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -273,22 +273,22 @@ public class PinotTableRestletResource {
           tablesDeleted.add(TableNameBuilder.REALTIME.tableNameWithType(tableName));
         }
       }
-      if (!tableExist) {
-        throw new ControllerApplicationException(LOGGER,
-            "Table '" + tableName + "' with type " + tableType + " does not exist", Response.Status.NOT_FOUND);
+      if (!tablesDeleted.isEmpty()) {
+        return new SuccessResponse("Tables: " + tablesDeleted + " deleted");
       }
-      return new SuccessResponse("Tables: " + tablesDeleted + " deleted");
     } catch (Exception e) {
       throw new ControllerApplicationException(LOGGER, e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR, e);
     }
+    throw new ControllerApplicationException(LOGGER,
+          "Table '" + tableName + "' with type " + tableType + " does not exist", Response.Status.NOT_FOUND);
   }
 
-  // Return true iff the table is of the expectedType based on the given tableName and tableType. The truth table:
-  //          tableType           TableNameBuilder.getTableTypeFromTableName(tableName)          Return value
-  //   1.       null                            null (i.e., table has no type suffix)         true
-  //   2.       null                            not_null                               typeFromTableName == expectedType
-  //   3.      not_null                         null                                      tableType == expectedType
-  //   4.      not_null                         not_null                      tableType==typeFromTableName==expectedType
+  //   Return true iff the table is of the expectedType based on the given tableName and tableType. The truth table:
+  //        tableType   TableNameBuilder.getTableTypeFromTableName(tableName)   Return value
+  //     1. null      null (i.e., table has no type suffix)           true
+  //     2. null      not_null                              typeFromTableName == expectedType
+  //     3. not_null      null                                    tableType == expectedType
+  //     4. not_null      not_null                          tableType==typeFromTableName==expectedType
   private boolean verifyTableType(String tableName, TableType tableType, TableType expectedType) {
     if (tableType != null && tableType != expectedType) {
       return false;
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -254,17 +254,22 @@ public class PinotTableRestletResource {
 
     List<String> tablesDeleted = new LinkedList<>();
     try {
-      if (verifyTableType(tableName, tableType, tableType.OFFLINE) && _pinotHelixResourceManager
-          .hasOfflineTable(tableName)) {
+      boolean tableExist = false;
+      if (verifyTableType(tableName, tableType, tableType.OFFLINE)) {
+        tableExist = _pinotHelixResourceManager.hasOfflineTable(tableName);
+        // Even the table name does not exist, still go on to delete remaining table metadata in case a previous delete
+        // did not complete.
         _pinotHelixResourceManager.deleteOfflineTable(tableName);
         tablesDeleted.add(TableNameBuilder.OFFLINE.tableNameWithType(tableName));
       }
-      if (verifyTableType(tableName, tableType, tableType.REALTIME) && _pinotHelixResourceManager
-          .hasRealtimeTable(tableName)) {
+      if (verifyTableType(tableName, tableType, tableType.REALTIME)) {
+        tableExist = _pinotHelixResourceManager.hasRealtimeTable(tableName);
+        // Even the table name does not exist, still go on to delete remaining table metadata in case a previous delete
+        // did not complete.
         _pinotHelixResourceManager.deleteRealtimeTable(tableName);
         tablesDeleted.add(TableNameBuilder.REALTIME.tableNameWithType(tableName));
       }
-      if (tablesDeleted.size() == 0) {
+      if (!tableExist) {
         throw new ControllerApplicationException(LOGGER, "Table '" + tableName + "' does not exist",
             Response.Status.NOT_FOUND);
       }
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
@@ -260,18 +260,22 @@ public class PinotTableRestletResource {
         // Even the table name does not exist, still go on to delete remaining table metadata in case a previous delete
         // did not complete.
         _pinotHelixResourceManager.deleteOfflineTable(tableName);
-        tablesDeleted.add(TableNameBuilder.OFFLINE.tableNameWithType(tableName));
+        if (tableExist) {
+          tablesDeleted.add(TableNameBuilder.OFFLINE.tableNameWithType(tableName));
+        }
       }
       if (verifyTableType(tableName, tableType, tableType.REALTIME)) {
         tableExist = _pinotHelixResourceManager.hasRealtimeTable(tableName);
         // Even the table name does not exist, still go on to delete remaining table metadata in case a previous delete
         // did not complete.
         _pinotHelixResourceManager.deleteRealtimeTable(tableName);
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,14 +254,20 @@ public SuccessResponse deleteTable(
...
         _pinotHelixResourceManager.deleteOfflineTable(tableName);
         tablesDeleted.add(TableNameBuilder.OFFLINE.tableNameWithType(tableName));
       }
-      if (tableType != TableType.OFFLINE && !TableNameBuilder.OFFLINE.tableHasTypeSuffix(tableName)) {
+      if (tableType != TableType.OFFLINE && !TableNameBuilder.OFFLINE.tableHasTypeSuffix(tableName)
+          && _pinotHelixResourceManager.hasRealtimeTable(tableName)) {
         _pinotHelixResourceManager.deleteRealtimeTable(tableName);
         tablesDeleted.add(TableNameBuilder.REALTIME.tableNameWithType(tableName));
       }
+      if (tablesDeleted.size() == 0) {
```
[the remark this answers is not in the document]
chenboat: Right now if the table does not exist, the rest end point still returns "Success [table_non_existent] deleted". This is not consistent with system state -- that is why I think it is a bug (though minor).

Thank you for the overflow. It is interesting and new to me. Both sides has valid points. I think there is argument for idem-potency there -- but Pinot controller is not a stateless service (it has zk data for table state) IMO.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,14 +254,20 @@ public SuccessResponse deleteTable(
...
+        throw new ControllerApplicationException(LOGGER, "Table '" + tableName + "' does not exist",
+            Response.Status.BAD_REQUEST);
```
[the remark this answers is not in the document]
chenboat: Done. Good point.

mcvsubbu:
Is this somehow causing an issue in your environment? What is the motivation behind this change?

chenboat:
> Is this somehow causing an issue in your environment? What is the motivation behind this change?

Yes. We are running on ~April 2019 version of Pinot (0.1.0).The delete table REST API then is buggy and very confusing: when I did "curl control_url/tables/A_REALTIME", the returned response is 
_{"status":"Table deleted [A_REALTIME_OFFLINE, A_REALTIME]"}._ It was really scary for the first sight because it seemed to delete the offline table as well (actually upon very careful examination) it is fine. 

The delete table API is very important (probably more important than create) because it can remove data. Furthermore, Pinot has this offline and realtime sub-table concepts and make the logic not straightforward. The bug fix since April on this API fixed the above issue. But I still believe there is room to improve:

1. No UNIT test coverage for such an important API -- this is actually the main goal of this PR.
2. The double negative boolean checks in the current codes are not very readable -- I think at least for now we should abstract them out as methods.
3. The minor issue on non-existent tables -- this issue also justifies we should at test coverage on the method.

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotTableRestletResourceTest.java
```diff
@@ -273,6 +275,76 @@ public void rebalanceTableWithoutSegments()
...
+    Assert.assertEquals(deleteResponse, "{\"status\":\"Tables: [table2_REALTIME] deleted\"}");
+
+    deleteResponse = sendDeleteRequest(StringUtil.join("/", this._controllerBaseApiUrl,
+        "tables", "table2?type=offline"));
+    Assert.assertEquals(deleteResponse, "{\"status\":\"Tables: [table2_OFFLINE] deleted\"}");
+
+    // Case 5: Delete a non-existent table and expect a bad request expection.
+    try {
+      deleteResponse = sendDeleteRequest(
+          StringUtil.join("/", this._controllerBaseApiUrl, "tables", "no_such_table_OFFLINE"));
```
jackjlli: Add `Assert.fail()` right after this line.

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotTableRestletResourceTest.java
```diff
@@ -21,10 +21,12 @@
 import com.fasterxml.jackson.databind.JsonNode;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+
```
jackjlli: remove blank line

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,20 +254,41 @@ public SuccessResponse deleteTable(
...
     } catch (Exception e) {
-      throw new ControllerApplicationException(LOGGER, e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR, e);
+      throw new ControllerApplicationException(LOGGER, e.getMessage(),
+          Response.Status.INTERNAL_SERVER_ERROR, e);
     }
   }
 
+  // Return true if tableType is NOT offline (i.e., null or realtime) and table name does  not end
+  // with _offline suffix.
+  private boolean toDeleteRealtimeTable(String tableName, TableType tableType) {
```
jamesyfshao: I feel this two method can use better naming, for example, ensureTableNameMatchType(..) and potentially merge two method into one. We can probably even remove the comments if the method naming is improved

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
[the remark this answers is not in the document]
jamesyfshao: I think it would be better to think of this change as an improvement to the pinot admin API feedback. set aside the debates over the correct way to respond to a DELETE on a resource that is not presented, This change provides the caller of this method more information by providing feedback if such method doesn't exist. If the users desire the old behavior, he/she can just check if the method returns non-500 return codes. But now our users are also able to check if the table is really in pinot system. Given how this change improves the existing pinot system behavior, I think we should accept this diff after fixing any further minor issues @Jackie-Jiang

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
chenboat: Done. Combine two functions into one: I also use positive condition testing to make it easier to understand.

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotTableRestletResourceTest.java
chenboat: Done.

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotTableRestletResourceTest.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,20 +254,35 @@ public SuccessResponse deleteTable(
...
-      throw new ControllerApplicationException(LOGGER, e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR, e);
+      throw new ControllerApplicationException(LOGGER, e.getMessage(),
+          Response.Status.INTERNAL_SERVER_ERROR, e);
     }
   }
 
+  // Return true iff the table is of the expectedType based on the given tableName and tableType.
+  private boolean verifyTableType(String tableName, TableType tableType, TableType expectedType) {
+    if (tableType == null)
+      return TableNameBuilder.getTableTypeFromTableName(tableName) == expectedType ||
```
Jackie-Jiang: Please add braces for statement in the if clause, and also cache the tableType into a local variable to save one string comparison.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,20 +254,35 @@ public SuccessResponse deleteTable(
...
         tablesDeleted.add(TableNameBuilder.REALTIME.tableNameWithType(tableName));
       }
+      if (tablesDeleted.size() == 0) {
+        throw new ControllerApplicationException(LOGGER, "Table '" + tableName + "' does not exist",
+            Response.Status.NOT_FOUND);
+      }
       return new SuccessResponse("Tables: " + tablesDeleted + " deleted");
     } catch (Exception e) {
-      throw new ControllerApplicationException(LOGGER, e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR, e);
+      throw new ControllerApplicationException(LOGGER, e.getMessage(),
```
Jackie-Jiang: Please reformat with #3705

chenboat:
> The reason why we do not check if table exists is because when the previous table delete failed in the middle, the ideal state might already be removed. We still want to clean up the other properties for the table.

Thanks for pointing this out. Is it a common case? We could solve this issue in a periodically validation cleaner task?

Jackie-Jiang:
> > The reason why we do not check if table exists is because when the previous table delete failed in the middle, the ideal state might already be removed. We still want to clean up the other properties for the table.
> 
> Thanks for pointing this out. Is it a common case? We could solve this issue in a periodically validation cleaner task?

That is not the common case, but in exception path. Yes, that can be achieved by a periodically validation, but automatically delete table properties could also be risky. Right now we still rely on manually call the delete table again to clean up all properties.
Any reason why you want to check the table existence first?

chenboat:
> > > The reason why we do not check if table exists is because when the previous table delete failed in the middle, the ideal state might already be removed. We still want to clean up the other properties for the table.
> > 
> > 
> > Thanks for pointing this out. Is it a common case? We could solve this issue in a periodically validation cleaner task?
> 
> That is not the common case, but in exception path. Yes, that can be achieved by a periodically validation, but automatically delete table properties could also be risky. Right now we still rely on manually call the delete table again to clean up all properties.
> Any reason why you want to check the table existence first?

The reason is to existence check is to make sure the returned message does not show successful deletion of an non-existent table. This is also consistent with the other similar API DropTable() which does a table existence check.

The main purpose of this PR is to add unit test coverage. I did not know you also rely on this API to clear up some remaining metadata -- it is a minor but important use case i guess. I propose we at least modify the return message for the case of non-existent table: something along the line like "WARNING: table name does not exist. Delete table successfully".

Jackie-Jiang:
I agree we should keep all APIs in sync. When the table does not exist, let's still delete the table properties, then throw the NOT_FOUND exception.

chenboat:
> I agree we should keep all APIs in sync. When the table does not exist, let's still delete the table properties, then throw the NOT_FOUND exception.

Make sense and this is probably the best solution currently. Revised the PR and please take another look.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,20 +254,40 @@ public SuccessResponse deleteTable(
...
+            Response.Status.NOT_FOUND);
+      }
       return new SuccessResponse("Tables: " + tablesDeleted + " deleted");
     } catch (Exception e) {
       throw new ControllerApplicationException(LOGGER, e.getMessage(), Response.Status.INTERNAL_SERVER_ERROR, e);
     }
   }
 
+  // Return true iff the table is of the expectedType based on the given tableName and tableType.
+  private boolean verifyTableType(String tableName, TableType tableType, TableType expectedType) {
```
Jackie-Jiang: This logic is not the same as the current logic.
The more readable way of the current logic is:
```
if (tableType != null && tableType != expectedType) {
  return false;
}
TableType typeFromTableName = TableNameBuilder.getTableTypeFromTableName(tableName);
return typeFromTableName == null || typeFromTableName == expectedType;
```

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,20 +254,40 @@ public SuccessResponse deleteTable(
...
     List<String> tablesDeleted = new LinkedList<>();
     try {
-      if (tableType != TableType.REALTIME && !TableNameBuilder.REALTIME.tableHasTypeSuffix(tableName)) {
+      boolean tableExist = false;
+      if (verifyTableType(tableName, tableType, tableType.OFFLINE)) {
+        tableExist = _pinotHelixResourceManager.hasOfflineTable(tableName);
+        // Even the table name does not exist, still go on to delete remaining table metadata in case a previous delete
+        // did not complete.
         _pinotHelixResourceManager.deleteOfflineTable(tableName);
         tablesDeleted.add(TableNameBuilder.OFFLINE.tableNameWithType(tableName));
```
Jackie-Jiang: Add table to `tablesDeleted` only if table exists?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,20 +254,40 @@ public SuccessResponse deleteTable(
...
-      if (tableType != TableType.OFFLINE && !TableNameBuilder.OFFLINE.tableHasTypeSuffix(tableName)) {
+      if (verifyTableType(tableName, tableType, tableType.REALTIME)) {
+        tableExist = _pinotHelixResourceManager.hasRealtimeTable(tableName);
+        // Even the table name does not exist, still go on to delete remaining table metadata in case a previous delete
+        // did not complete.
         _pinotHelixResourceManager.deleteRealtimeTable(tableName);
         tablesDeleted.add(TableNameBuilder.REALTIME.tableNameWithType(tableName));
       }
+      if (!tableExist) {
+        throw new ControllerApplicationException(LOGGER, "Table '" + tableName + "' does not exist",
```
Jackie-Jiang: Add tableType into the error message?

chenboat: done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
chenboat: Good point. Your logic is the same as the current one. Added a truth table in the comments to define the boolean logic -- this is probably the best way to understand the logic behind.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
chenboat: Done.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,20 +254,49 @@ public SuccessResponse deleteTable(
...
     }
   }
 
+  // Return true iff the table is of the expectedType based on the given tableName and tableType. The truth table:
+  //          tableType           TableNameBuilder.getTableTypeFromTableName(tableName)          Return value
+  //   1.       null                            null (i.e., table has no type suffix)         true
+  //   2.       null                            not_null                               typeFromTableName == expectedType
+  //   3.      not_null                         null                                      tableType == expectedType
+  //   4.      not_null                         not_null                      tableType==typeFromTableName==expectedType
+  private boolean verifyTableType(String tableName, TableType tableType, TableType expectedType) {
```
Jackie-Jiang: The comments seems not properly aligned?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
```diff
@@ -254,20 +254,49 @@ public SuccessResponse deleteTable(
...
+        // Even the table name does not exist, still go on to delete remaining table metadata in case a previous delete
+        // did not complete.
         _pinotHelixResourceManager.deleteRealtimeTable(tableName);
-        tablesDeleted.add(TableNameBuilder.REALTIME.tableNameWithType(tableName));
+        if (tableExist) {
+          tablesDeleted.add(TableNameBuilder.REALTIME.tableNameWithType(tableName));
+        }
+      }
+      if (!tableExist) {
+        throw new ControllerApplicationException(LOGGER,
```
Jackie-Jiang: This exception will be caught by line 282. You need to move this outside of the try-catch block.
You can check whether `tablesDeleted.isEmpty()`

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
chenboat: Use tab instead of space now. There is no template for comments as far as I know.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotTableRestletResource.java
chenboat: Good point. Done. @Jackie-Jiang

chenboat:
> Please address the comments and we will merge the change

Done. Please take another look. @Jackie-Jiang

PR #4612, opened 2019-09-13
Add column types info to the selection and aggregation query results.

chenboat:
Add column types to the selection and aggregation query results. The column type info for group by columns are not added in this PR because the information right now is not present in the data schema available to the brokerReduceService. 

@kishoreg


npawar:

PR #4617, opened 2019-09-17
Adding CreateSegmentCommandV2 command

fx19880617:
Sample usage: 
```
pinot-admin.sh CreateSegmentV2 \
-dataDir /tmp/pinot/segmentv2/data \
-outDir /tmp/pinot/segmentv2/out \
-schemaFile /tmp/pinot/segmentv2/conn_schema.json \
-tableConfig /tmp/pinot/segmentv2/conn_table_config.json \
-segmentName conn-test \
-overwrite
```

Test table config/schema/data:
[segmentv2.zip](https://github.com/apache/incubator-pinot/files/3618890/segmentv2.zip)

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
@@ -177,6 +177,7 @@ public class SegmentGeneratorConfig {
     }
 
     IndexingConfig indexingConfig = tableConfig.getIndexingConfig();
+    _tableName = tableConfig.getTableName();
     List<String> noDictionaryColumns = indexingConfig.getNoDictionaryColumns();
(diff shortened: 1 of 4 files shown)


kishoreg:
Existing command is too complicated and needs segment generation config. I would like to simplify this. Touching existing will break compatibility, so it’s better add a new one and deprecate the existing one.

mcvsubbu:
In that case, let us add a print to the existing one to direct the user to this new command, and mention that it is deprecated. Over time, we can remove the older one. 
I have not looked through fully, but we must add a hook to pick up custom config, and naming schemes.
If one is not doing anything special like I mentioned above, the V2 command should work, right?

kishoreg:
Yes. Which custom config are useful? Can we put them as part of tableConfig. 

I am trying to see if we can move towards 3 main sections to setup a table 

Schema
TableConfig
IngestionConfig Batch/stream

fx19880617:
closed this as we moved to new ingestion spec things

PR #4618, opened 2019-09-17
Make Kafka offset out of range as an transient exception.

chenboat:
During LLC migration in Uber, we found LLC consumer often stopped consumption because it is assigned a bad Kafka broker (and resulted in OFFSET_OUT_OF_RANGE error). The current way of assigning OFFSET_OUT_OF_RANGE as permanent error causes the realtime consumption to completely stop -- until a background fixer thread to revive the consumption in ~10mins. 

In a company with large number of Kafka brokers, faulty brokers occur pretty frequently. A stop of realtime ingestion for 10 mins is not acceptable due to high SLA. We rather mark the error as a transient one and let the retry to kick in to select a different broker. It works well after the error code reassignment.

### DIFF
--- a/pinot-connectors/pinot-connector-kafka-0.9/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaConnectionHandler.java
+++ b/pinot-connectors/pinot-connector-kafka-0.9/src/main/java/org/apache/pinot/core/realtime/impl/kafka/KafkaConnectionHandler.java
@@ -388,7 +388,6 @@ public class KafkaConnectionHandler {
     final Errors kafkaError = Errors.forCode(kafkaErrorCode);
     switch (kafkaError) {
       case UNKNOWN:
-      case OFFSET_OUT_OF_RANGE:
       case CORRUPT_MESSAGE:
       case MESSAGE_TOO_LARGE:
       case OFFSET_METADATA_TOO_LARGE:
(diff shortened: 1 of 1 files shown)


chenboat:
By "a segment created before kafka rolled over data", do you mean the case where the consumer asks for data with offset starting with A but all brokers have their min offsets > A.

In such scenario, I think this change will fare as well as the current codes and in some cases work better. Here is the comparison:
1. Current code: It will stop ingestion immediately because the OUT_OF_RANGE is a permanent error. The Validation manager will try to restart the ingestion but that will fail again if the offset problem still persists. We would need to do manual fix.

2. New code: the transient error status means the retry will kick in: the consumer thread will try a few more brokers to check their data range. If all of them have the same issue, the behavior will be the same as the current code. If any broker still have the data, the ingestion will continue without waiting for the validation manager.

Our experience so far is that putting OUT_OF_RANGE as transient_error with more retries eliminated all ingestion stopping problems due to faulty brokers.

mcvsubbu:
Your first point is wrong. The validation manager will put in newer offsets that are valid before restarting. No manual fix is needed.

chenboat:
> Your first point is wrong. The validation manager will put in newer offsets that are valid before restarting. No manual fix is needed.

Thanks for pointing out this. The main point of this diff remains though: the validation manager only kicks in after a set period of 10 mins as we observed -- which is too long for strict SLA for realtime ingestion. The new error code reassignment does not change the validation manager behaviors -- it just give more chances for retries.

mcvsubbu:
A repeated transient exception is eventually turned into a permanent exception, so I am fine with this. Can you please re-start the tests and get them to pass? We can commit this.

PR #4619, opened 2019-09-18
[TE] Add data set display name and show it on UI

jihaozh:
- Add the ability to set dataset's display name. If the display name is set, the UI will show the display name instead of the dataset key.
- Add dataset and granularity filters in the alert list page.


akshayrai:
@jihaozh can you provide some context (example might help) behind why we need the display name?

harleyjj:
It looks good, but it's failing CI tests

jihaozh:
> @jihaozh can you provide some context (example might help) behind why we need the display name?

Added this per the requirements of UMP alert onboarding. I'll send a follow-up RB to fill the display name for UMP datasets.

jihaozh:
> It looks good, but it's failing CI tests

Thanks. Just pushed the fix for the front end tests.

PR #4621, opened 2019-09-18
[TE] fix the last time stamp filter out the generated anomalies 

jihaozh:
- Remove the check that to filter out generated anomalies based on the last timestamp. This may filter out useful anomalies.
- In the replay endpoint, add the option to turn on email sending for the generated anomalies.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/wrapper/AnomalyDetectorWrapper.java
@@ -216,10 +216,8 @@ public class AnomalyDetectorWrapper extends DetectionPipeline {
       anomaly.getProperties().put(PROP_SUB_ENTITY_NAME, this.entityName);
     }
     long lastTimeStamp = this.getLastTimeStamp();
-    List<MergedAnomalyResultDTO> anomalyResults = anomalies.stream().filter(anomaly -> anomaly.getEndTime() <= lastTimeStamp).collect(
-        Collectors.toList());
     PredictionResult predictedTimeSeries = new PredictionResult(this.detectorName, this.metricUrn, predictedResult.getDataFrame());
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
```diff
@@ -476,7 +477,11 @@ public Response detectionReplay(
       }
 
       for (MergedAnomalyResultDTO anomaly : result.getAnomalies()) {
-        anomaly.setAnomalyResultSource(AnomalyResultSource.ANOMALY_REPLAY);
+        // if would like to skip sending the alert emails, the anomalies should be marked with REPLAY
+        // otherwise, should be left as the default source
+        if (!sendEmail) {
+          anomaly.setAnomalyResultSource(AnomalyResultSource.ANOMALY_REPLAY);
```
akshayrai: Shouldn't the anomaly source be set to ANOMALY_REPLAY irrespective of the email?

jihaozh: If the source set to ANOMALY_REPLAY, the alerter will skip sending it. Actually, that's the only place this flag is used in the system. If that's not set, the alerter can pick up the anomaly and sent it out. 

I think we should have an option to send the anomalies generated by replay. This is useful when the system missed an anomaly but the email is still important for the customer.

akshayrai: I still don't think this might be a good way. let's discuss this offline. I have 2 concerns.
1. If we are replaying it then I think we should always set the source to ANOMALY_REPLAY to distinguish it from a regular detection, irrespective of the email report.
2. In case we are replaying the detection for some historical window, the alerter watermark can prevent it from sending the replayed anomalies right!

jihaozh: as discussed offline, will put that into a separate request

PR #4622, opened 2019-09-19
Standardize the Dictionary interface, ensure the BYTES support

Jackie-Jiang:
- Type conversion between INT/LONG/FLOAT/DOUBLE/STRING are allowed
- Type conversion between STRING/BYTES are allowed
- Push down and enhance the MutableDictionary range predicate evaluation
- Return dictId/dictIds while indexing value into MutableDictionary
- Add documentation for the contract of Dictionary interface
- Enhance PinotDataType for BYTES conversion


pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOffHeapMutableDictionary.java
```diff
@@ -53,102 +53,148 @@ public BytesOffHeapMutableDictionary(int estimatedCardinality, int maxOverflowHa
...
   }
 
   @Override
-  public boolean inRange(@Nonnull String lower, @Nonnull String upper, int dictIdToCompare, boolean includeLower,
-      boolean includeUpper) {
-    return valueInRange(new ByteArray(BytesUtils.toBytes(lower)), new ByteArray(BytesUtils.toBytes(upper)),
-        includeLower, includeUpper, new ByteArray(getBytesValue(dictIdToCompare)));
+  public int indexOf(String stringValue) {
+    byte[] bytesValue = BytesUtils.toBytes(stringValue);
+    return getDictId(new ByteArray(bytesValue), bytesValue);
```
jackjlli: Here you can just use `BytesUtils. toByteArray(stringValue)` which is already in this PR. 😉

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/Dictionary.java
```diff
@@ -21,18 +21,40 @@
...
-   * @return Index of object, or -1 if not found.
+   * NOTE: Immutable dictionary is always sorted; mutable dictionary is always unsorted.
+   */
+  boolean isSorted();
+
+  int length();
+
+  /**
+   * Returns the index of the string representation of the value in the dictionary, or {@link #NULL_VALUE_INDEX} (-1) if
+   * the value does not exist. This API is for cross-type predicate evaluation.
```
jackjlli: Please add a comment here saying the stringValue is always unpadded.

pinot-common/src/main/java/org/apache/pinot/common/utils/BytesUtils.java
```diff
@@ -20,27 +20,38 @@
...
+
+  /**
+   * Converts a Hex encoded string to a {@link ByteArray}.
+   *
+   * @param stringValue Hex encoded string
+   * @return Decoded {@link ByteArray}
+   */
+  public static ByteArray toByteArray(String stringValue) {
+    try {
+      return new ByteArray(Hex.decodeHex(stringValue.toCharArray()));
```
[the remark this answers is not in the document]
Jackie-Jiang: Done

pinot-core/src/main/java/org/apache/pinot/core/realtime/impl/dictionary/BytesOffHeapMutableDictionary.java
Jackie-Jiang: I need both byte[] and ByteArray, so I only fetched the byte[]

jackjlli: Oh I see

pinot-core/src/main/java/org/apache/pinot/core/segment/index/readers/Dictionary.java
Jackie-Jiang: This value is from the query, so no concept of padding here. Added comments for the String binary search.

PR #4629, opened 2019-09-23
Put order-by expressions in front of selection expreesions in selection DataSchema

Jackie-Jiang:
For selection order-by queries, put order-by expressions in front of selection expressions in the server response (DataTable)
- In the DataTable, all the expressions (both order-by and selection) should be deduplicated
- This is the behavior before PR #4346
- Fix the backward-incompatible issue introduced in PR #4346
- With this, no need to keep extra column indices to order the rows
On Broker side, when compiling the query, deduplicate the order-by expressions
- With this and the DataTable column order, only need to compare the first numOrderByColumns
On Broker side, add temporary code to handle the DataSchema with different expression orders
- Temporary code are manually tested
- Once all the servers are deployed, will remove the temporary code
When formatting the selection result, use thread-local DecimalFormat instead of creating one for each value

Also include some minor bug fixes
- 'SELECT *' should not select order-by expressions
- Make EmptySelectionOperator work with transform functions

Code convertion: only use Nullable annotation, all arguments and return value for the method are by default Nonnull


pinot-core/src/main/java/org/apache/pinot/core/operator/transform/TransformOperator.java
```diff
@@ -39,28 +37,23 @@
...
 
   /**
    * Constructor for the class
    *
    * @param projectionOperator Projection operator
    * @param expressions Set of expressions to evaluate
    */
-  public TransformOperator(@Nonnull ProjectionOperator projectionOperator,
-      @Nonnull List<TransformExpressionTree> expressions) {
+  public TransformOperator(ProjectionOperator projectionOperator, Collection<TransformExpressionTree> expressions) {
```
mayankshriv: I see this PR is removing `@Nonnull` annotations. If this is not related to this change, it would be good to take it out of this PR, or list down all such changes (along with the intention), in the PR description. It will make it easier to review the PR.

pinot-core/src/main/java/org/apache/pinot/core/query/selection/SelectionOperatorUtils.java
```diff
@@ -76,107 +73,74 @@ private SelectionOperatorUtils() {
...
-      @Nonnull IndexSegment indexSegment) {
-    Set<String> selectionColumns = new HashSet<>(getSelectionColumns(selection.getSelectionColumns(), indexSegment));
-    if (selection.getSelectionSortSequence() != null) {
-      for (SelectionSort selectionSort : selection.getSelectionSortSequence()) {
-        selectionColumns.add(selectionSort.getColumn());
-      }
-    }
-    return selectionColumns;
-  }
+  private static final ThreadLocal<DecimalFormat> THREAD_LOCAL_INT_FORMAT =
```
mayankshriv: Same here, capture this change in the PR description (as it seems not related to the intention of the PR).

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
```diff
@@ -50,37 +44,27 @@
   private final String _segmentName;
   private final ProjectionPlanNode _projectionPlanNode;
   private final Set<String> _projectionColumns = new HashSet<>();
-  private final Set<TransformExpressionTree> _expressionTrees = new LinkedHashSet<>();
+  private final Set<TransformExpressionTree> _expressions = new HashSet<>();
```
mayankshriv: Ordering not required anymore?

Jackie-Jiang:
@mayankshriv Addressed all the comments. The temporary handling code is manually tested. The plan is to first merge the PR, cut a release (open-source release candidate), then test it. If found issue, will apply fix in another PR.

pinot-core/src/main/java/org/apache/pinot/core/operator/query/SelectionOnlyOperator.java
```diff
@@ -0,0 +1,126 @@
...
+  private final DataSchema _dataSchema;
+  private final int _numRowsToKeep;
+  private final List<Serializable[]> _rows;
+
+  private ExecutionStatistics _executionStatistics;
+
+  public SelectionOnlyOperator(IndexSegment indexSegment, Selection selection, TransformOperator transformOperator) {
+    _indexSegment = indexSegment;
+    _transformOperator = transformOperator;
+    _expressions = SelectionOperatorUtils.extractExpressions(selection.getSelectionColumns(), indexSegment, null);
```
siddharthteotia: This gives in the same order as specified in the query right ? I mean broker does not need to do any re-ordering

pinot-core/src/main/java/org/apache/pinot/core/query/selection/SelectionOperatorUtils.java
```diff
@@ -437,20 +389,26 @@ public static SelectionResults renderSelectionResultsWithoutOrdering(@Nonnull Li
    * @param dataSchema data schema.
    * @return column indices
    */
-  public static int[] getColumnIndices(@Nonnull List<String> selectionColumns, @Nonnull DataSchema dataSchema) {
+  public static int[] getColumnIndices(List<String> selectionColumns, DataSchema dataSchema) {
```
siddharthteotia: Why do we need this if the order of columns as processed by the SelectionOperator is same as what user specified? I see you are using a function SelectionOperatorUtils.extractColumns and the comment seems to imply that broker relies on that ordering  to process so I am not sure why we need column indices

pinot-core/src/main/java/org/apache/pinot/core/operator/query/SelectionOnlyOperator.java
Jackie-Jiang: Yes, or all columns in alphabetical order if it is `SELECT *`. The re-ordering problem happens for selection order-by queries.

pinot-core/src/main/java/org/apache/pinot/core/query/selection/SelectionOperatorUtils.java
Jackie-Jiang: For selection order-by queries, the column order in the DataSchema is not the same as what user specified, and that is the reason why we need to calculate the column indices in the DataSchema to get the correct selection columns.

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
```diff
@@ -89,39 +73,36 @@ private void extractColumnsAndTransforms(@Nonnull BrokerRequest brokerRequest,
...
-          _expressionTrees.add(transformExpressionTree);
+          _expressions.add(transformExpressionTree);
         }
       }
     } else {
       Selection selection = brokerRequest.getSelections();
       // No ordering required, select minimum number of documents
       if (!selection.isSetSelectionSortSequence()) {
-        _maxDocPerNextCall = Math.min(selection.getOffset() + selection.getSize(), _maxDocPerNextCall);
+        _maxDocPerNextCall = Math.max(Math.min(selection.getSize(), _maxDocPerNextCall), 1);
```
npawar: why is this Math being changed? how is the behaviour staying the same with this?

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -527,4 +524,162 @@ private void setGroupByHavingResults(@Nonnull BrokerResponseNative brokerRespons
...
+   * Following part are temporary code to handle the selection query with different DataSchema returned from different
+   * servers.
+   * TODO: Remove the code after all servers are migrated to the current version.
+   */
+
+  private static DataSchema getCanonicalDataSchema(Selection selection, DataSchema dataSchema) {
+    String[] columnNames = dataSchema.getColumnNames();
+    Map<String, Integer> columnToIndexMap = SelectionOperatorUtils.getColumnToIndexMap(columnNames);
+
+    Set<String> canonicalColumnSet = new HashSet<>();
```
siddharthteotia: (nit): I think we know the exact size (as in the number of columns) so we should pass that as capacity to the constructor of these structures to prevent potential resize

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
Jackie-Jiang: For selection only queries, offset does not matter because no order is provided. We need to fetch at least one document to prevent exception (EmptySelectionOperator use TransformOperator to determine the expression type). These behavior are checked in the test

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
Jackie-Jiang: For HashSet, the initial capacity should not be configured the same as number of columns. The minimum capacity required should be (numColumns/loadFactor), but too small number can also increase the chance of hash collision.
Since the number of columns should be very few, the default size should be good enough IMO.

Jackie-Jiang: Added the initial size for ArrayList

pinot-core/src/main/java/org/apache/pinot/core/plan/TransformPlanNode.java
Jackie-Jiang: Added comments for that

PR #4633, opened 2019-09-23
Add metrics for the number of partitions for which controller becomes leader

jackjlli:
This PR adds metrics to track the number of partitions for which controller becomes leader, so that 
when the total number of this metric is aggregated from multiple controllers, we know every partitions have their own leader.


pinot-common/src/main/java/org/apache/pinot/common/metrics/ControllerGauge.java
```diff
@@ -54,6 +54,9 @@
   // Pinot controller resource enabled
   PINOT_LEAD_CONTROLLER_RESOURCE_ENABLED("PinotLeadControllerResourceEnabled", true),
 
+  // Number of partitions for which current controller becomes the leader
+  NUMBER_PARTITIONS_CONTROLLER_BECOMES_LEADER("NumberPartitionsControllerBecomesLeader", true),
```
mcvsubbu: Rename to `ControllerLeaderPartitionCount` or `NumControllerLeaderPartitions`

jackjlli: Done.

PR #4634, opened 2019-09-24
Remove import of sun.misc.Cleaner to make code compatible with jdk11

fx19880617:
sun.misc.Cleaner is deprecated since jdk9 and replaced with jdk.internal.ref.Cleaner.
This change aims to make Pinot compilable in higher jdk version.

Ref: https://bugs.openjdk.java.net/browse/JDK-8148117

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/segment/memory/PinotByteBuffer.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/segment/memory/PinotByteBuffer.java
@@ -328,9 +328,8 @@ public class PinotByteBuffer extends PinotDataBuffer {
 
   @Override
   protected void release() {
-    Cleaner cleaner = ((DirectBuffer) _buffer).cleaner();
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/org/apache/pinot/core/segment/memory/PinotByteBuffer.java
```diff
@@ -328,9 +328,8 @@ public void flush() {
 
   @Override
   protected void release() {
-    Cleaner cleaner = ((DirectBuffer) _buffer).cleaner();
-    if (cleaner != null) {
-      cleaner.clean();
+    if ( ((DirectBuffer) _buffer).cleaner() != null) {
```
Jackie-Jiang: (nit) Please reformat this line

Jackie-Jiang:
@fx19880617 I think you forgot to remove the import lol...

PR #4635, opened 2019-09-24
Support ORC files as data source in segment creation

fx19880617:
Sample commands:
build distribution:
```
mvn clean package  -DskipTests=true  -Pbin-dist
```
Run segment creation: 
```
pinot-distribution/target/apache-pinot-incubating-0.2.0-SNAPSHOT-bin/apache-pinot-incubating-0.2.0-SNAPSHOT-bin/bin/pinot-admin.sh CreateSegment -dataDir pinot-distribution/target/apache-pinot-incubating-0.2.0-SNAPSHOT-bin/apache-pinot-incubating-0.2.0-SNAPSHOT-bin/sample_data -format orc  -outDir /tmp/pinot-data/orc -tableName airlineStats -segmentName airlineStats_0 -schemaFile pinot-distribution/target/apache-pinot-incubating-0.2.0-SNAPSHOT-bin/apache-pinot-incubating-0.2.0-SNAPSHOT-bin/sample_data/airlineStats_schema.json  -overwrite
```


pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
```diff
@@ -291,27 +304,59 @@ public Double toDouble(Object value) {
   INTEGER_ARRAY {
     @Override
     public Integer[] convert(Object value, PinotDataType sourceType) {
+      if (value instanceof List) {
```
Jackie-Jiang: Please add the List support into `toIntegerArray()`, same for other types

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
```diff
@@ -291,27 +304,59 @@ public Double toDouble(Object value) {
...
+        final List list = (List) value;
```
Jackie-Jiang: (nit) we don't use final in method

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
```diff
@@ -291,27 +304,59 @@ public Double toDouble(Object value) {
...
+        Integer[] res = new Integer[list.size()];
```
Jackie-Jiang: cache the `list.size()`

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
```diff
@@ -263,21 +264,33 @@ public String convert(Object value, PinotDataType sourceType) {
   OBJECT {
     @Override
     public Integer toInteger(Object value) {
+      if (value instanceof Number) {
```
Jackie-Jiang: Why do we need this? Number should not be categorized as OBJECT

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
```diff
@@ -164,6 +165,8 @@ private Object getBaseObject(WritableComparable w) {
       obj = ((DoubleWritable) w).get();
     } else if (BytesWritable.class.isAssignableFrom(w.getClass())) {
       obj = ((BytesWritable) w).getBytes();
+    } else if (ByteWritable.class.isAssignableFrom(w.getClass())) {
```
Jackie-Jiang: Why `toString()`?

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
fx19880617: I've seen SHORT/INTEGER/LONGs are set as OBJECT from OrcRecordReader

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
fx19880617: the value is 8-bits tinyint, both `get()` and `toString()` will give same human readable numbers in a string format.

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
Jackie-Jiang: They should fall into the corresponding PinotDataType. Please check `DataTypeTransformer.transform()`. Also I think you need to add List into that method

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
Jackie-Jiang: Can you put this after the boolean type for readability?

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
```diff
@@ -355,6 +368,15 @@ public String toString(Object value) {
   }
 
   public Integer[] toIntegerArray(Object value) {
+    if (value instanceof List) {
```
Jackie-Jiang: You need to use the singleValueType to do the type conversion. Please refer to the code from line 383-390, and use the same naming convention

pinot-orc/src/main/java/org/apache/pinot/orc/data/readers/ORCRecordReader.java
fx19880617: will do

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
fx19880617: got it. will do

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/DataTypeTransformer.java
```diff
@@ -86,9 +87,24 @@ public GenericRow transform(GenericRow record) {
...
         }
         record.putField(column, value);
+        continue;
       } else {
         PinotDataType source;
-        if (value instanceof Object[]) {
+        if (value instanceof List) {
+          // Multi-valued column
+          List values = (List) value;
+          if (!values.isEmpty()) {
```
Jackie-Jiang: This check should be moved to line 80

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/DataTypeTransformer.java
```diff
@@ -86,9 +87,24 @@ public GenericRow transform(GenericRow record) {
...
+            source = MULTI_VALUE_TYPE_MAP.get(values.get(0).getClass());
+            if (source == null) {
+              source = PinotDataType.OBJECT_ARRAY;
+            }
+          } else {
+            source = PinotDataType.OBJECT_ARRAY;
+          }
+          value = dest.convert(value, source);
```
Jackie-Jiang: Please add some comments on why we need to convert even if source == dest

pinot-core/src/main/java/org/apache/pinot/core/data/recordtransformer/PinotDataType.java
```diff
@@ -357,6 +358,15 @@ public String toString(Object value) {
   public Integer[] toIntegerArray(Object value) {
     if (isSingleValue()) {
       return new Integer[]{toInteger(value)};
+    } else if (value instanceof List) {
```
Jackie-Jiang: Can you make it as following for readability:
```
if (isSingleValue()) {
   ...
} else {
  if (value instanceof List) {
    List valueList = (List) value;
    ...
  } else {
    Object[] valueArray = (Object[]) value;
    ...
  }
}
```

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -431,7 +437,25 @@ protected void getDataFilePathsHelper(FileSystem fileSystem, FileStatus[] fileSt
...
+        return fileName.endsWith(".parquet");
+      case CSV:
+        return fileName.endsWith(".csv");
+      case JSON:
+        return fileName.endsWith(".json");
+      case THRIFT:
+        return fileName.endsWith(".thrift");
+      case ORC:
+        return fileName.endsWith(".orc");
+      case OTHER:
```
Jackie-Jiang: Throw exception for unsupported format?

PR #4636, opened 2019-09-24
Documentation for completionConfig

npawar:
Add documentation for what completionConfig means, when and how to use it.

### DIFF
--- a/docs/tableconfig_schema.rst
+++ b/docs/tableconfig_schema.rst
@@ -113,6 +113,25 @@ The ``segmentsConfig`` section has information about configuring the following:
       "segmentAssignmentStrategy": "BalanceNumSegmentAssignmentStrategy"
     },
 
+* Completion Config
+
+  You can also add a ``completionConfig`` section under the ``segmentsConfig`` section. Completion config holds information related to realtime segment completion. There's just 1 field in this config as of now, which is the ``completionMode``. The value of the ``completioMode`` decides how non-winner servers should replace the completed segment during realtime segment completion. Currently, the supported values are
+
+  * ``DEFAULT``: if the in memory segment in the non-winner server is equivalent to the committed segment, then build and replace, else download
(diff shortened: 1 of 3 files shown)


docs/tableconfig_schema.rst
```diff
@@ -113,6 +113,25 @@ The ``segmentsConfig`` section has information about configuring the following:
       "segmentAssignmentStrategy": "BalanceNumSegmentAssignmentStrategy"
     },
 
+* Completion Config
+
+  You can also add a ``completionConfig`` section under the ``segmentsConfig`` section. Completion config holds information related to realtime segment completion. There's just 1 field in this config as of now, which is the ``completionMode``. The value of the ``completioMode`` decides how non-winner servers should replace the completed segment during realtime segment completion. Currently, the supported values are
```
mcvsubbu: s/1/one/

Please add a couple of sentences on winner or non-winner (better use commit leader and commit follower) in the architecture section and then add a sentence here to refer to the arch section for what these mean. You can also point to the realtime design document  segment completion protocol section. for more read.

docs/tableconfig_schema.rst
```diff
@@ -113,6 +113,25 @@ The ``segmentsConfig`` section has information about configuring the following:
...
+
+  * ``DEFAULT``: if the in memory segment in the non-winner server is equivalent to the committed segment, then build and replace, else download
+  * ``DOWNLOAD``: non-winner servers always download the segment, never build it
```
mcvsubbu: Add a sentence on when to use DOWNLOAD.

Also, does one really need to specify "Default" in the table config? Leaving it empty will be default, and the only value supported is "DOWNLOAD"?

docs/tableconfig_schema.rst
npawar: Added to the architecture section.

docs/tableconfig_schema.rst
npawar: Done

PR #4637, opened 2019-09-24
Adding documentation for Pinot Schema.

mayankshriv:
Added documentation for Pinot Schema.

### DIFF
--- a/docs/schema.rst
+++ b/docs/schema.rst
@@ -1,0 +1,124 @@
+..
+.. Licensed to the Apache Software Foundation (ASF) under one
+.. or more contributor license agreements.  See the NOTICE file
+.. distributed with this work for additional information
+.. regarding copyright ownership.  The ASF licenses this file
+.. to you under the Apache License, Version 2.0 (the
+.. "License"); you may not use this file except in compliance
+.. with the License.  You may obtain a copy of the License at
+..
+..   http://www.apache.org/licenses/LICENSE-2.0
+..
+.. Unless required by applicable law or agreed to in writing,
+.. software distributed under the License is distributed on an
+.. "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+.. KIND, either express or implied.  See the License for the
+.. specific language governing permissions and limitations
+.. under the License.
+..
+
+.. _schema-section:
+
+Pinot Schema
+============
+
+Pinot Schema consists of columns that can be categorized as dimensions, metric or time.
+
+**Dimensions** - These are columns that organize the data. For example, ``accountId``, ``country``, ``industry``, etc. These columns are used to slice/dice the data and typically appear in the ``selection``, ``filter`` and ``group-by`` sections in queries.
+
+**Metrics** - These are columns that represent quantative measurements. For example, ``numClicks``, ``pageViews``, etc. These columns typically appear in the aggregation section of the query, e.g., ``select sum(pageViews) from...``.
+
+**Time** - The time column represents the timestamp of the data. There is only one time column in a schema, and typically appears in the ``filter`` and ``group-by`` sections in queries. 
+
+A sample schema is shown below and will be used as an example to described the various fields.
+
+Sample Schema:
+~~~~~~~~~~~~~~
+
+.. code-block:: json
+
+   {
+     "schemaName": "flights",
+     "dimensionFieldSpecs": [
+       {
+         "name": "flightNumber",
+         "dataType": "LONG"
+       },
+       {
+         "name": "tags",
+         "dataType": "STRING",
+         "singleValueField": false
+       }
+     ],
+     "metricFieldSpecs": [
+       {
+         "name": "price",
+         "dataType": "DOUBLE"
+       }
+     ],
+     "timeFieldSpec": {
+       "incomingGranularitySpec": {
+         "name": "secondsSinceEpoch",
+         "dataType": "INT",
+         "timeFormat" : "EPOCH",
+         "timeType": "SECONDS"
+       },
+        "outgoingGranularitySpec": {
+         "name": "messageTimeHours",
+         "dataType": "INT",
+         "timeFormat" : "EPOCH",
+         "timeType": "DAYS"
+       }
+     }
+   }
+
+Schema Name:
+~~~~~~~~~~~~
+
+Every Pinot schema has a name that is used to identify it. For example, schema for a table is specified in the table conig by its name.
+
+
+FieldSpecs:
+~~~~~~~~~~~
+
+Each column in the schema is described using a ``fieldSpec`` that captures various attributes of the column, such as name, data-type, etc. A schema may contain an array of ``dimensionFieldSpecs`` that describe all the dimension columns, ``metricFieldSpecs`` that describe all the metric columns, and a ``timeFieldSpec`` that describes the time column.
+
+Dimensions:
+~~~~~~~~~~~
+
+The schema example above has two dimensions ``flightNumber`` of type ``Long`` and ``tags`` of type ``String``. Data types supported for dimension columns are:
+
+* **INT**
+* **FLOAT**
+* **LONG**
+* **DOUBLE**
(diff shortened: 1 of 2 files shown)


docs/schema.rst
```diff
@@ -0,0 +1,124 @@
...
+**Time** - The time column represents the timestamp of the data. There is only one time column in a schema, and typically appears in the ``filter`` and ``group-by`` sections in queries. 
+
+A sample schema is shown below and will be used as an example to described the various fields.
+
+Sample Schema:
+~~~~~~~~~~~~~~
+
+.. code-block:: json
+
+   {
```
kishoreg: can we pick a schema with all data types?

docs/schema.rst
```diff
@@ -0,0 +1,124 @@
...
+.. specific language governing permissions and limitations
+.. under the License.
+..
+
+.. _schema-section:
+
+Pinot Schema
+============
+
+Pinot Schema consists of columns that can be categorized as dimensions, metric or time.
```
kishoreg: We had a diagram describing these things with sample data?

docs/schema.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Time:
+~~~~~
+
+The schema above also contains a ``timeFieldSpec`` that is used to specify the attributes of the time column:
+
+* **incomingGranularitySpec** : Specifies the name, data type and time type for the time stamp present in the incoming data into Pinot.
+* **outgoingGranularitySpec** : Specifies the name, data type and time type for the time stamp as desired to be stored in Pinot.
+
+In this example, the input timestamp specified in ``SECONDS`` will be automatically converted into ``DAYS`` before storing into Pinot. The ``timeFieldSpec`` also has an optional attribute ``timeFormat`` that can take values ``EPOCH`` (default) and ``SIMPLE_DATE_FORMAT:<format>``.
```
kishoreg: The documentation in code had some nice examples. We also want to move towards DateTimeFieldSpec right?

mayankshriv: I believe @npawar will be adding that separately. We can link to that doc here.

docs/schema.rst
mayankshriv: I don't seem to recall, where can I find it?

mcvsubbu:
I think we should have a rule that all comments on a documentation PR should be constructive, preferably with text substitution. Good to have description of what is asked for :)

docs/schema.rst
```diff
@@ -0,0 +1,124 @@
...
+.. _schema-section:
+
+Pinot Schema
+============
+
+Pinot Schema consists of columns that can be categorized as dimensions, metric or time.
+
+**Dimensions** - These are columns that organize the data. For example, ``accountId``, ``country``, ``industry``, etc. These columns are used to slice/dice the data and typically appear in the ``selection``, ``filter`` and ``group-by`` sections in queries.
+
+**Metrics** - These are columns that represent quantative measurements. For example, ``numClicks``, ``pageViews``, etc. These columns typically appear in the aggregation section of the query, e.g., ``select sum(pageViews) from...``.
```
mcvsubbu: Please add that the types of these columns must be numeric

docs/schema.rst
```diff
@@ -0,0 +1,124 @@
...
+
+**Time** - The time column represents the timestamp of the data. There is only one time column in a schema, and typically appears in the ``filter`` and ``group-by`` sections in queries.
```
mcvsubbu: I would omit the appearance of time column in filter/group-by.

docs/schema.rst
```diff
@@ -0,0 +1,124 @@
...
+Every Pinot schema has a name that is used to identify it. For example, schema for a table is specified in the table conig by its name.
+
+
+FieldSpecs:
+~~~~~~~~~~~
+
+Each column in the schema is described using a ``fieldSpec`` that captures various attributes of the column, such as name, data-type, etc. A schema may contain an array of ``dimensionFieldSpecs`` that describe all the dimension columns, ``metricFieldSpecs`` that describe all the metric columns, and a ``timeFieldSpec`` that describes the time column.
+
+Dimensions:
+~~~~~~~~~~~~~~
```
mcvsubbu: I think the length of this line has to be same as previous line, otherwise we get a warning during `make html`

docs/schema.rst
```diff
@@ -0,0 +1,124 @@
...
+
+The schema example above has two dimensions ``flightNumber`` of type ``Long`` and ``tags`` of type ``String``. Data types supported for dimension columns are:
+
+* **INT**
```
mcvsubbu: Use two backticks instead of two asterisks to make it similar to other pages. Not sure how these show up.

docs/schema.rst
mayankshriv: Well, I listed down the valid data types for these columns, and that includes BYTES. I can add something like `these columns are typically numeric`.

docs/schema.rst
mayankshriv: I'd keep it, based on the type of questions I am used to receiving on what dim/metric/time columns mean. It seems relating to where the columns can appear in the query seems to make sense to the users.

docs/schema.rst
mayankshriv: I don't see any error in `make html`:

```sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v1.8.5
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: [] 0 added, 1 changed, 0 removed
reading sources... [100%] schema                                                                                                                                                                                                                                                          
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] schema                                                                                                                                                                                                                                                           
generating indices... genindex
writing additional pages... search
copying static files... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded.

The HTML pages are in _build/html.

Build finished. The HTML pages are in _build/html.```

docs/schema.rst
mayankshriv: The two backticks notations does not appear correct (in other pages as well). We should change it to bold by using this notations. Sample below: Note though, the backtick version in this comment looks much better than what we see in read-the-docs.

* ``INT``
* ``FLOAT``

v.s.

* **INT**
* **FLOAT**

mayankshriv:
> Please add sentence or two regarding:
> 
> Time column is optional in REFRESH use cases, but mandatory in APPEND use cases. You can add references to the table config seciton here.
> 
> How time column is used by broker to split between realtime and offline. I am nor sure if this belongs in the schema section, or a pointer to this belongs here. For now, if you can just add it here, we can figure out how to move it around. If you can find a better place that is fine as well.

I'll add a sentence on refresh/append use case. However, I don't think the other part about broker splitting between realtime/offline (and retention for that matter) belong here. There is some mention of that in Table Config section, already.

mcvsubbu:
> > Please add sentence or two regarding:
> > Time column is optional in REFRESH use cases, but mandatory in APPEND use cases. You can add references to the table config seciton here.
> > How time column is used by broker to split between realtime and offline. I am nor sure if this belongs in the schema section, or a pointer to this belongs here. For now, if you can just add it here, we can figure out how to move it around. If you can find a better place that is fine as well.
> 
> I'll add a sentence on refresh/append use case. However, I don't think the other part about broker splitting between realtime/offline (and retention for that matter) belong here. There is some mention of that in Table Config section, already.

Oh good. In that case just add a ref to the table config section for how time column is used during query federation. I think it is useful to cross-reference things to encourage people to go to more than one section as needed. thanks

docs/schema.rst
mcvsubbu: oh, then we need to standardize.

docs/schema.rst
mayankshriv: I have explicitly listed all the data types supported for various field specs. Adding them in the sample would make the schema unnecessarily large, and not sure if it provides much value.

PR #4645, opened 2019-09-25
Adding sample ADL settings to documentation

jenniferdai:
Updating documents

### DIFF
--- a/docs/pluggable_storage.rst
+++ b/docs/pluggable_storage.rst
@@ -58,31 +58,50 @@ AzurePinotFS requires the following configs according to your environment:
 adl.accountId, adl.authEndpoint, adl.clientId, adl.clientSecret
 
 Sample Controller Config
-    pinot.controller.storage.factory.class.adl: org.apache.pinot.filesystem.AzurePinotFS
-    pinot.controller.storage.factory.adl.accountId: xxxx
-    pinot.controller.storage.factory.adl.authEndpoint: xxxx
-    pinot.controller.storage.factory.adl.clientId: xxxx
-    pinot.controller.storage.factory.adl.clientId: xxxx
-    pinot.controller.segment.fetcher.protocols: adl
+
+.. code-block:: none
+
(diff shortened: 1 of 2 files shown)


docs/pluggable_storage.rst
```diff
@@ -57,6 +57,9 @@ AzurePinotFS requires the following configs according to your environment:
 
 adl.accountId, adl.authEndpoint, adl.clientId, adl.clientSecret
 
+You can find the parameters in your account as follows:
+https://stackoverflow.com/questions/56349040/what-is-clientid-authtokenendpoint-clientkey-for-accessing-azure-data-lake
```
mayankshriv: This does not help in showing an example of how the Pinot ADLS config looks like, right?

docs/pluggable_storage.rst
```diff
@@ -57,6 +57,25 @@ AzurePinotFS requires the following configs according to your environment:
 
 adl.accountId, adl.authEndpoint, adl.clientId, adl.clientSecret
 
+Sample Controller Config
```
mayankshriv: Can we format these configs as JSON?

jenniferdai: To format as JSON, it requires {}, and I don't want to confuse our users, as just the flat config without the braces is what they need to put in their configuration file

mayankshriv: Oh I see, this is not JSON config, but apache PropertiesConfiguration file. In that case, this is fine. Could you please add instruction on how this file is passed to `pinot-server`? Also, please make sure that the output html page is formatted correctly.

jenniferdai: I verified that the output is formatted. Can you clarify what you mean by instruction on how the file is passed? It's the config file that users can save and input to startServer, so we don't do anything special (they essentially just need to declare this path for their config)

docs/pluggable_storage.rst
jenniferdai: It displays where clients need to look for params, which could be helpful for clients confused about meaning

PR #4649, opened 2019-09-25
Use timed wait around get() calls in PinotInstanceRestletResourceTest

siddharthteotia:
cc @jackjlli , @mcvsubbu

### DIFF
--- a/pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
+++ b/pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
@@ -19,16 +19,19 @@
 package org.apache.pinot.controller.api;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.base.Function;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 import java.util.TreeMap;
+import javax.annotation.Nullable;
+import org.apache.pinot.common.config.Instance;
 import org.apache.pinot.common.utils.CommonConstants.Helix;
 import org.apache.pinot.common.utils.CommonConstants.Helix.InstanceType;
 import org.apache.pinot.common.utils.JsonUtils;
-import org.apache.pinot.common.config.Instance;
 import org.apache.pinot.controller.helix.ControllerTest;
+import org.apache.pinot.util.TestUtils;
 import org.testng.annotations.AfterClass;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
@@ -43,6 +46,8 @@ import static org.testng.Assert.fail;
  */
 public class PinotInstanceRestletResourceTest extends ControllerTest {
 
+  private static final long GET_CALL_TIMEOUT_MS = 10000;
+
   @BeforeClass
   public void setUp() {
     startZk();
@@ -54,9 +59,17 @@ public class PinotInstanceRestletResourceTest extends ControllerTest {
       throws Exception {
     // Check that there is only one CONTROLLER instance in the cluster
     String listInstancesUrl = _controllerRequestURLBuilder.forInstanceList();
-    JsonNode instanceList = JsonUtils.stringToJsonNode(sendGetRequest(listInstancesUrl));
-    assertEquals(instanceList.get("instances").size(), 1);
-    assertTrue(instanceList.get("instances").get(0).asText().startsWith(Helix.PREFIX_OF_CONTROLLER_INSTANCE));
+
+    TestUtils.waitForCondition(aVoid -> {
+      try {
+        String getResponse = sendGetRequest(listInstancesUrl);
+        JsonNode jsonNode = JsonUtils.stringToJsonNode(getResponse);
+        return (jsonNode != null) && (jsonNode.get("instances") != null) && (jsonNode.get("instances").size() == 1)
+            && (jsonNode.get("instances").get(0).asText().startsWith(Helix.PREFIX_OF_CONTROLLER_INSTANCE));
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+    }, GET_CALL_TIMEOUT_MS, "Expected one controller instance");
 
     // Create untagged broker and server instances
     String createInstanceUrl = _controllerRequestURLBuilder.forInstanceCreate();
@@ -66,8 +79,8 @@ public class PinotInstanceRestletResourceTest extends ControllerTest {
     Instance serverInstance = new Instance("1.2.3.4", 2345, InstanceType.SERVER, null, null);
     sendPostRequest(createInstanceUrl, serverInstance.toJsonString());
 
-    // Check that there are 3 instances
-    assertEquals(JsonUtils.stringToJsonNode(sendGetRequest(listInstancesUrl)).get("instances").size(), 3);
+    // Check that there are 3 instances -- controller, broker and server
+    checkNumInstances(listInstancesUrl, 3);
 
     // Create broker and server instances with tags and pools
     brokerInstance = new Instance("2.3.4.5", 1234, InstanceType.BROKER, Collections.singletonList("tag_BROKER"), null);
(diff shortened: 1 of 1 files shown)


pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
```diff
@@ -99,52 +110,100 @@ public void testInstanceListingAndCreation()
...
-    assertEquals(pools.get("tag_REALTIME").asText(), "1");
+    TestUtils.waitForCondition(new Function<Void, Boolean>() {
+      @Nullable
+      @Override
+      public Boolean apply(@Nullable Void aVoid) {
+        try {
+          String getResponse =
+              sendGetRequest(_controllerRequestURLBuilder.forInstanceInformation("Broker_1.2.3.4_1234"));
+          JsonNode instance = JsonUtils.stringToJsonNode(getResponse);
+          assertEquals(instance.get("instanceName").asText(), "Broker_1.2.3.4_1234");
```
jackjlli: The assertion statement will directly fail the test other than re-running it. No assertion should exist in `TestUtils.waitForCondition`

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
```diff
@@ -99,52 +110,100 @@ public void testInstanceListingAndCreation()
...
+      @Override
+      public Boolean apply(@Nullable Void aVoid) {
+        try {
+          String getResponse =
+              sendGetRequest(_controllerRequestURLBuilder.forInstanceInformation("Server_1.2.3.4_2345"));
+          JsonNode instance = JsonUtils.stringToJsonNode(getResponse);
+          assertEquals(instance.get("instanceName").asText(), "Server_1.2.3.4_2345");
+          assertEquals(instance.get("hostName").asText(), "Server_1.2.3.4");
+          assertEquals(instance.get("port").asText(), "2345");
+          assertTrue(instance.get("enabled").asBoolean());
```
jackjlli: Same here.

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
```diff
@@ -99,52 +110,100 @@ public void testInstanceListingAndCreation()
...
+      @Nullable
+      @Override
+      public Boolean apply(@Nullable Void aVoid) {
+        try {
+          String getResponse =
+              sendGetRequest(_controllerRequestURLBuilder.forInstanceInformation("Server_2.3.4.5_2345"));
+          JsonNode instance = JsonUtils.stringToJsonNode(getResponse);
+          assertEquals(instance.get("instanceName").asText(), "Server_2.3.4.5_2345");
+          assertEquals(instance.get("hostName").asText(), "Server_2.3.4.5");
+          assertEquals(instance.get("port").asText(), "2345");
```
jackjlli: Same here.

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
```diff
@@ -54,9 +57,17 @@ public void testInstanceListingAndCreation()
...
-    assertTrue(instanceList.get("instances").get(0).asText().startsWith(Helix.PREFIX_OF_CONTROLLER_INSTANCE));
+
+    TestUtils.waitForCondition(aVoid -> {
+      try {
+        String getResponse = sendGetRequest(listInstancesUrl);
+        JsonNode jsonNode = JsonUtils.stringToJsonNode(getResponse);
+        return jsonNode.get("instances").size() == 1 && jsonNode.get("instances").get(0).asText()
+            .startsWith(Helix.PREFIX_OF_CONTROLLER_INSTANCE);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
jackjlli: The thrown exception will directly fail the test instead of retry. 
The cache data accessor may return null, which will exactly be caught here. 
Returning false would be good.

siddharthteotia: We have to throw exception if getRequest fails for any arbitrary reason as in that case, retry might not help. Made the changes to check for null result though

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
siddharthteotia: thanks for catching. I have made the changes to check for null results in fields.

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
```diff
@@ -54,9 +57,17 @@ public void testInstanceListingAndCreation()
...
+    TestUtils.waitForCondition(aVoid -> {
+      try {
+        String getResponse = sendGetRequest(listInstancesUrl);
+        JsonNode jsonNode = JsonUtils.stringToJsonNode(getResponse);
+        return (jsonNode != null) && (jsonNode.get("instances") != null) && (jsonNode.get("instances").size() == 1)
+            && (jsonNode.get("instances").get(0).asText().startsWith(Helix.PREFIX_OF_CONTROLLER_INSTANCE));
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+    }, 500L, 10000L, "Expected one controller instance");
```
mcvsubbu: Another implementation of waitForCond () calls every 100ms. I think that should work fine here. We can still have the total time spent as 10s. Please declare that as a static final int and use it everywhere.thanks

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
```diff
@@ -99,60 +111,69 @@ public void testInstanceListingAndCreation()
...
-    assertEquals(pools.get("tag_REALTIME").asText(), "1");
-
-    // Check that an error is given for an instance that does not exist
-    try {
-      sendGetRequest(_controllerRequestURLBuilder.forInstanceInformation("Server_potato_8126"));
-      fail("Request to get instance information for an instance that does not exist did not fail");
-    } catch (IOException e) {
-      // Expected
-    }
+    checkInstanceInfo("Broker_1.2.3.4_1234", "Broker_1.2.3.4", 1234, new String[0], null, null);
```
mcvsubbu: nice.

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
```diff
@@ -99,60 +111,69 @@ public void testInstanceListingAndCreation()
...
+          } else {
+            result = result && instance.get("pools").isNull();
+          }
+
+          return result;
+        } catch (Exception e) {
+          throw new RuntimeException(e);
+        }
+      }
+    }, 500L, 10000L, "Failed to retrieve correct information for instance: " + instanceName);
```
mcvsubbu: same. Use the other signature

siddharthteotia: done

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
siddharthteotia: done

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
siddharthteotia: done

pinot-controller/src/test/java/org/apache/pinot/controller/api/PinotInstanceRestletResourceTest.java
siddharthteotia: Done

Issue #4626, opened 2019-09-20
Low level realtime consumer (LLC) got into ERROR state due to thread race condition.


chenboat:
Recently we observed LLC realtime consumer got into ERROR state during data consumption. We are running on mid April Pinot (5be8431d6a49) but by code inspection, the issue could also occur on the latest code.

 In a high level timeline, 
   (1) At the end of segment completion protocol, the controller asked the server to keep its segment and go online. 
   (2) the consumer thread T, after receiving the KEEP response, tried to build the segment but stuck in acquiring semaphore.
   (3) the main thread in the server received OnlineFromConsuming transition message from the helix. It then tried to stop the consumer thread T in (2) and waited for 10 mins but the consumer thread did not stop because it is waiting for semaphore. Then in RETAINNG state, the main thread chose to download the segment and go online.
   (4) Now there are two threads both trying to write to the final segment directory and caused file overwrite ERROR.

The detailed logs are attached below. Here some observations about the current codes and some fix ideas:
(1) In buildSegmentInternal() of LLRealtimeSegmentDataManager, _shouldStop is not checked after long ops like semaphore acquisition and segment build.
https://github.com/apache/incubator-pinot/blob/c0dbbfc81c1fa0d6be78c1b95448fff96803f0c6/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java#L673-L681

If the method re-checked the _shouldStop state after potential lengthy ops like acquireSemaphore() and buildSegment(), the PartitionConsumer thread can just stop as instructed by the main thread -- and there will be no overwriting issue. This fix alone could already fix the issue.

(2) The main thread chose to download and replace segment in RETAINING state -- this is not consistent with the comment below. 

https://github.com/apache/incubator-pinot/blob/c0dbbfc81c1fa0d6be78c1b95448fff96803f0c6/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java#L106-L108

chenboat:
[...]

PR #4652, opened 2019-09-27
[#4626] Fix a thread race condition during LLC realtime ingestion.

chenboat:
Fix a thread race condition by double check the _shouldStop value in the consumer thread after any lengthy blocking ops like build() and semaphore acquasition.

@mcvsubbu

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
@@ -677,7 +677,7 @@ public class LLRealtimeSegmentDataManager extends RealtimeSegmentDataManager {
         segmentLogger.info("Waiting to acquire semaphore for building segment");
         _segBuildSemaphore.acquire();
       }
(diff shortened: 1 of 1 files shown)


chenboat:
The bug #4652 occurs when segment download is ongoing in `goOnlineFromConsuming()`. The manager is in RETAINING state but the partition consumer thread ignores the manager's instruction to stop in `buildSegmentInternal()`.So I do not think moving to downloadAndReplace() for all case will solve the bug. The PR is a more defensive effort to cover such case. 

Recall that in #4652, the controller asks the server to KEEP its segment. Thus the server then proceeded to build the segment in the partitionConsumerThread. But this happen out side of the 
`goOnlineFromConsuming()`. Does #4546 change this behavior as well?

> We should just downloa and replace thesegment in all cases in the `goOnlineFromConsuming()` method.
> 
> The current implementation (the first one) assumes that it is a lot more expensive to download than to build. In general, (especially for high QPS use cases) we have seen that segment build is more expensive.
> 
> It also makes the code path simpler.
> 
> It should be more common for a segment to participate in the segment completion protocol than to get an online state transition while still consuming or catching up, so the code path should get hit less often, especially after PR 4546

npawar:

PR #4657, opened 2019-09-30
[TE] Clean up legacy detection related classes; update ADContentFormatterContext

akshayrai:
Changes:
* Updated `ADContentFormatterContext` with latest detectionConfig and detectionAlertConfig
* Selectively removed a lot of legacy detection related classes


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/bao/jdbc/MergedAnomalyResultManagerImpl.java
```diff
@@ -485,6 +491,12 @@ public MergedAnomalyResultDTO convertMergedAnomalyBean2DTO(MergedAnomalyResultBe
       mergedAnomalyResultDTO.setFunction(anomalyFunctionDTO);
     }
 
+    if (mergedAnomalyResultBean.getDetectionConfigId() != null) {
```
jihaozh: Is this necessary to put it in the `convertMergedAnomalyBean2DTO`? 
I'm worried that this will make the calls to get the anomalies slower because it needs to send 2 additional SQL queries to get the detection config whenever fetching an anomaly.

akshayrai: Good point

PR #4658, opened 2019-09-30
Refactor TransformBlockDataFetcher into RowBasedBlockValueFetcher

Jackie-Jiang:
Add BlockValSet.isSingleValue() interface
Remove the old value fetcher implementations


pinot-core/src/main/java/org/apache/pinot/core/common/RowBasedBlockValueFetcher.java
```diff
@@ -0,0 +1,217 @@
...
+        case LONG:
+          return new LongMultiValueFetcher(blockValSet.getLongValuesMV());
+        case FLOAT:
+          return new FloatMultiValueFetcher(blockValSet.getFloatValuesMV());
+        case DOUBLE:
+          return new DoubleMultiValueFetcher(blockValSet.getDoubleValuesMV());
+        case STRING:
+          return new StringMultiValueFetcher(blockValSet.getStringValuesMV());
+        default:
+          throw new IllegalStateException();
```
siddharthteotia: May be add a message with type in exception?

siddharthteotia:
LGTM

pinot-core/src/main/java/org/apache/pinot/core/common/RowBasedBlockValueFetcher.java
Jackie-Jiang: Done

PR #4662, opened 2019-10-01
Add wait for condition for toggle test

jackjlli:
This PR updates the method of getting instance configs, so that cache data accessor is used for segment operations, and direct data accessor is used for tenant/table operations.

Related issue: https://github.com/apache/incubator-pinot/issues/4665

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -254,7 +254,7 @@ public class PinotHelixResourceManager {
    * Add instance group tag for controller so that pinot controller can be assigned to lead controller resource.
    */
   private void addInstanceGroupTagIfNeeded() {
-    InstanceConfig instanceConfig = getHelixInstanceConfig(_instanceId);
+    InstanceConfig instanceConfig = getHelixInstanceConfig(_instanceId, false);
     if (!instanceConfig.containsTag(Helix.CONTROLLER_INSTANCE)) {
       LOGGER.info("Controller: {} doesn't contain group tag: {}. Adding one.", _instanceId, Helix.CONTROLLER_INSTANCE);
       instanceConfig.addTag(Helix.CONTROLLER_INSTANCE);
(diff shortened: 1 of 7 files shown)


Jackie-Jiang:
Don't change the test. We need to fix the code

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -268,44 +268,42 @@ private void addInstanceGroupTagIfNeeded() {
...
-   *
+   * Get all instance Ids. If it's related to segment operation like assigning segments to instances, cache accessor would be used.
+   * Otherwise, direct data accessor would be used.
+   * @param segmentRelatedOperation whether it's related to segment operation
    * @return List of instance Ids
    */
   @Nonnull
-  public List<String> getAllInstances() {
-    return _cacheInstanceConfigsDataAccessor.getChildNames("/", AccessOption.PERSISTENT);
+  public List<String> getAllInstances(boolean segmentRelatedOperation) {
```
Jackie-Jiang: You don't need to change this method

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -268,44 +268,42 @@ private void addInstanceGroupTagIfNeeded() {
...
+    } else {
+      return _helixAdmin.getInstancesInCluster(_helixClusterName);
+    }
   }
 
   /**
    * Returns the config for all the Helix instances in the cluster.
+   * All the callers are not for segment assignment. No need to use cached accessor.
    */
   public List<InstanceConfig> getAllHelixInstanceConfigs() {
```
Jackie-Jiang: I think the segment assignment is only calling this method

jackjlli: This method is called by `assignInstance` method, which is where we add/update table config.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: One of the usages of this method is to toggle the segment state, which can also be a segment operation though.

jackjlli:
This PR isn't needed any more. Closing this PR.

PR #4664, opened 2019-10-02
Support other filesystems for Pinot Hadoop job

fx19880617:
Make Pinot Hadoop job to support other file systems like s3a.

### DIFF
--- a/pinot-hadoop/pom.xml
+++ b/pinot-hadoop/pom.xml
@@ -43,6 +43,7 @@
         <plugins>
           <plugin>
             <artifactId>maven-shade-plugin</artifactId>
+            <version>3.2.1</version>
             <executions>
               <execution>
                 <phase>package</phase>
@@ -50,6 +51,16 @@
                   <goal>shade</goal>
                 </goals>
                 <configuration>
+                  <relocations>
+                    <relocation>
+                      <pattern>com.google.common.base.Preconditions</pattern>
+                      <shadedPattern>shaded.com.google.common.base.Preconditions</shadedPattern>
+                    </relocation>
+                    <relocation>
+                      <pattern>com.fasterxml.jackson</pattern>
+                      <shadedPattern>shaded.com.fasterxml.jackson</shadedPattern>
+                    </relocation>
+                  </relocations>
                   <transformers>
                     <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
                     <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
--- a/pinot-hadoop/pom.xml
+++ b/pinot-hadoop/pom.xml
@@ -51,6 +51,13 @@
                   <goal>shade</goal>
                 </goals>
                 <configuration>
+                  <!--
+                  Usually in hadoop environment, there are multiple jars with different versions.
+                  Most of the NoSuchMethodExceptions are caused by class loading conflicts.
+                  Class relocation ensures the reference of certain packages/classes in Pinot code to
+                  shaded libs, e.g. jackson or guava.
+                  Ref: https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html
+                  -->
                   <relocations>
                     <relocation>
                       <pattern>com.google.common.base.Preconditions</pattern>
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/BaseSegmentJob.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/BaseSegmentJob.java
@@ -107,8 +107,14 @@ public abstract class BaseSegmentJob extends Configured {
   protected List<Path> getDataFilePaths(Path pathPattern)
       throws IOException {
     List<Path> tarFilePaths = new ArrayList<>();
-    FileSystem fileSystem = FileSystem.get(_conf);
-    getDataFilePathsHelper(fileSystem, fileSystem.globStatus(pathPattern), tarFilePaths);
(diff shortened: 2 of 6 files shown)


pinot-hadoop/pom.xml
```diff
@@ -43,13 +43,24 @@
...
             <artifactId>maven-shade-plugin</artifactId>
+            <version>3.2.1</version>
             <executions>
               <execution>
                 <phase>package</phase>
                 <goals>
                   <goal>shade</goal>
                 </goals>
                 <configuration>
+                  <relocations>
```
Jackie-Jiang: Why shading these jars?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/BaseSegmentJob.java
```diff
@@ -107,8 +107,14 @@ protected Path getPathFromProperty(String key) {
   protected List<Path> getDataFilePaths(Path pathPattern)
       throws IOException {
     List<Path> tarFilePaths = new ArrayList<>();
-    FileSystem fileSystem = FileSystem.get(_conf);
-    getDataFilePathsHelper(fileSystem, fileSystem.globStatus(pathPattern), tarFilePaths);
+    FileSystem fileSystem = FileSystem.get(pathPattern.toUri(), _conf);
+    _logger.info("Using filesystem: {}", fileSystem);
+    final FileStatus[] fileStatuses = fileSystem.globStatus(pathPattern);
```
Jackie-Jiang: (nit) remove the final

pinot-hadoop/pom.xml
fx19880617: This is to relocate method call to use our shaded jar not from hadoop classpath. For Hadoop env, there could be older version of jackson and guava in classpath and cause NoSuchMethod exception.

Jackie-Jiang: Can you please add some comments to the commit message?

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/SegmentCreationJob.java
```diff
@@ -108,15 +108,15 @@ protected boolean isDataFile(String fileName) {
       return true;
     }
     return fileName.endsWith(".avro") || fileName.endsWith(".csv") || fileName.endsWith(".json") || fileName
-        .endsWith(".thrift");
+        .endsWith(".thrift") || fileName.endsWith(".parquet") || fileName.endsWith(".orc");
```
jenniferdai: You shouldn't need fileName.endsWith(".orc"), as the way ORC readers are currently configured go through JobConfigConstants.RECORD_READER_PATH

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mappers/SegmentCreationMapper.java
```diff
@@ -148,7 +148,8 @@ public void setup(Context context)
     Preconditions.checkState(_localSegmentDir.mkdir());
     Preconditions.checkState(_localSegmentTarDir.mkdir());
 
-    _fileSystem = FileSystem.get(context.getConfiguration());
+    // Note: we made an assumption that input file and output path are using same filesystem.
+    _fileSystem = FileSystem.get(_hdfsSegmentTarDir.toUri(), context.getConfiguration());
```
jenniferdai: If I'm not misunderstanding, this one will always be local, right? It's assigned with FileOutputFormat.getWorkOutputPath(context)

fx19880617: This filesystem should be same as where we want to store the segment tars.
If we use s3a path as segment tar dir. Then this should be s3a filesystem.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mappers/SegmentCreationMapper.java
```diff
@@ -277,6 +277,12 @@ protected FileFormat getFileFormat(String fileName) {
     if (fileName.endsWith(".thrift")) {
       return FileFormat.THRIFT;
     }
+    if (fileName.endsWith(".parquet")) {
+      return FileFormat.PARQUET;
+    }
+    if (fileName.endsWith(".orc")) {
```
jenniferdai: Correct me if you are understanding differently / configuring it differently, but when I configure orc, I configure. Same with the comment I left yesterday in SegmentCreationJob.

record.reader.path: ${PATH_TO_ORCREADER} 
therefore, it skips the fileName.endsWith

    if (_recordReaderPath != null) {
      segmentGeneratorConfig.setRecordReaderPath(_recordReaderPath);
      segmentGeneratorConfig.setFormat(FileFormat.OTHER);
    } else {
      FileFormat fileFormat = getFileFormat(inputFileName);
      segmentGeneratorConfig.setFormat(fileFormat);
      segmentGeneratorConfig.setReaderConfig(getReaderConfig(fileFormat));
    }

fx19880617: yes, so basically it requires extra configs in order to read from orc/parquet files

jenniferdai: We count ORC as FileFormat.OTHER and set the record reader path to the ORCRecordReader, so you don't need to include the orc-specific logic for segmentcreation to work

PR #4672, opened 2019-10-03
[TE] frontend - harleyjj/comparison - fix bugs from mini bugbash

harleyjj:
1) new granularity check protects against null value
2) remove anomaly id from investigate link
3) make cache key more robust against duplicate in anomaly service (preview anomalies can have duplicate ids since they are generated on the fly)
4) improve error handling for _getAnomalies task


harleyjj:
I updated the commit to also handle errors better in _getAnomalies, since one case now entails calling it twice (Edit Alert Preview being called when current anomalies aren't yet loaded)

PR #4674, opened 2019-10-03
Fix flakiness in BoundedAccountingExecutorTest

mcvsubbu:
Added comments to explain why the flakiness may happen in travis and
added the fix.

### DIFF
--- a/pinot-core/src/test/java/org/apache/pinot/core/query/scheduler/resources/BoundedAccountingExecutorTest.java
+++ b/pinot-core/src/test/java/org/apache/pinot/core/query/scheduler/resources/BoundedAccountingExecutorTest.java
@@ -18,12 +18,15 @@
  */
 package org.apache.pinot.core.query.scheduler.resources;
 
+import com.google.common.base.Function;
 import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.CyclicBarrier;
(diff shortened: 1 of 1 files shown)


pinot-core/src/test/java/org/apache/pinot/core/query/scheduler/resources/BoundedAccountingExecutorTest.java
```diff
@@ -83,17 +86,44 @@ public void run() {
...
+    // they exit, the pending jobs will acquire permits and start to increment
+    // the running counter and wait on startupBarrier.await().
     syncer.validationBarrier.await();
     // verify additional jobs are run as soon as current job finishes
     syncer.validationBarrier = new CyclicBarrier(pendingJobs + 1);
+    // When we run the test in a small number of cores, it is possible that the running jobs
+    // have not yet gotten to execute running.decrementAndGet(), but the pending jobs have already
+    // done the increment. So, we need to wait until we check the running counter to equal the
+    // pending jobs.
+    TestUtils.waitForCondition(new Function<Void, Boolean>() {
```
Jackie-Jiang: This can be simplified as:
`TestUtils.waitForCondition(aVoid -> running.get() == pendingJobs, 10_000, "Invalid number of running jobs" + running.get());`

PR #4677, opened 2019-10-03
Mocking out the LeadControllerManager in test

mcvsubbu:
We should not need the real one, so it is better to mock the class altogether

### DIFF
--- a/pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTest.java
+++ b/pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTest.java
@@ -1285,7 +1285,6 @@ public class PinotLLCRealtimeSegmentManagerTest {
   static class FakePinotLLCRealtimeSegmentManager extends PinotLLCRealtimeSegmentManager {
 
     private static final ControllerConf CONTROLLER_CONF = new ControllerConf();
-    private static final LeadControllerManager _mockLeadControllerManager = mock(LeadControllerManager.class);
     private List<String> _existingLLCSegments = new ArrayList<>(1);
 
     public List<String> _paths = new ArrayList<>(16);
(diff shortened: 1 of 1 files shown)


pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManagerTest.java
```diff
@@ -1321,8 +1322,7 @@ protected FakePinotLLCRealtimeSegmentManager(PinotHelixResourceManager pinotHeli
 
     protected FakePinotLLCRealtimeSegmentManager(PinotHelixResourceManager pinotHelixResourceManager,
         List<String> existingLLCSegments, ControllerMetrics controllerMetrics) {
-      super(pinotHelixResourceManager, CONTROLLER_CONF, controllerMetrics,
-          new LeadControllerManager(pinotHelixResourceManager.getHelixZkManager(), controllerMetrics));
+      super(pinotHelixResourceManager, CONTROLLER_CONF, controllerMetrics, _mockLeadControllerManager);
```
jackjlli: There's only one PinotLLCRealtimeSegmentManager object in controller starter. In that test, there're 3 PinotLLCRealtimeSegmentManager objects. So in general, there should be 3 LeadControllerManager objects.
I'd prefer mocking the object here instead of using a static one. Otherwise, it'd be good to add some comment for the static object.

mcvsubbu: The LeadControllerManager object is never used in the code path that the test goes through. If we refactor the code so that it gets used, then we will need to mock each one of them with its own behavior. I can add a comment saying so.

mcvsubbu: ok, i changed it so that we get one mock for each instance of the fake segment manager.

Issue #4665, opened 2019-10-02
Cached data accessor may be inconsistent even in the same controller


jackjlli:
In https://github.com/apache/incubator-pinot/pull/3633, we leveraged `ZkCacheBaseDataAccessor` to cache instance configs in order to reduce the zk access. 
While it has its drawback that it's only used to read instance config; writing/updating instance config is done in Helix code and it isn't controlled by Pinot. Thus, it's possible that the cached instance configs may be stale. 
Thus, the current workaround is only use cached accessor for segment operations. For tenant/table level operations, we use direct data accessor. It's because segment operations are more frequent than the ones of tenant/table.
In the future, the following actions can be done for this issue:
* Contact Helix team to make `updateCache` method public in cached data accessor.
* Maintain a zkClient in PinotHelixResourceManager and override the methods like addInstance.
* Implement our own guava cache to cache instance configs.

kishoreg:
Any kind of cache will always never be consistent when readers and writers are in different process/nodes. 

Instead of trying to achieve that, we should make sure our code is designed to handle the inconsistency gracefully

Jackie-Jiang:
@kishoreg We are able to handle the inconsistency from different process/nodes. The current problem is that, in the same thread, even in the same method, read after write might get inconsistent result, which we should avoid.

PR #3633, opened 2019-01-03
Use ZkCacheBaseDataAccessor to cache instance configs in PinotHelixResourceManager

jackjlli:
This PR uses `ZkCacheBaseDataAccessor` to cache instance configs in PinotHelixResourceManager, so that zk read access can be reduced.

PR #4678, opened 2019-10-04
Use guava cache for instance admin endpoints

jackjlli:
This PR:
* uses guava cache for instance admin endpoints.
* removes cached data accessor, which comes from Helix code.

Current usage of cached based data accessor has the issue described in this link: #4665
And the cached data accessor was to solve the problem of frequent zk access to get the instance admin endpoints when there's segment upload. Plus, we've seen some flaky tests in our build, which are also related to this cached data accessor.
This issue can be solved by just introducing a guava cache to cache the instance admin endpoint. So no need to maintain a cached data accessor any more.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -187,13 +187,11 @@ public class PinotHelixResourceManager {
     _helixAdmin = _helixZkManager.getClusterManagmentTool();
     _propertyStore = _helixZkManager.getHelixPropertyStore();
     _helixDataAccessor = _helixZkManager.getHelixDataAccessor();
-    // Cache instance zk paths.
-    BaseDataAccessor<ZNRecord> baseDataAccessor = _helixDataAccessor.getBaseDataAccessor();
+    _keyBuilder = _helixDataAccessor.keyBuilder();
 
     // Add instance group tag for controller
     addInstanceGroupTagIfNeeded();
 
-    _keyBuilder = _helixDataAccessor.keyBuilder();
     _segmentDeletionManager = new SegmentDeletionManager(_dataDir, _helixAdmin, _helixClusterName, _propertyStore);
     ZKMetadataProvider.setClusterTenantIsolationEnabled(_propertyStore, _isSingleTenantCluster);
   }
@@ -302,7 +300,7 @@ public class PinotHelixResourceManager {
    * @return Helix instance config
    */
   public InstanceConfig getHelixInstanceConfig(@Nonnull String instanceId) {
-    return _helixAdmin.getInstanceConfig(_helixClusterName, instanceId);
+    return _helixDataAccessor.getProperty(_keyBuilder.instanceConfig(instanceId));
   }
 
   /**
@@ -2220,9 +2218,7 @@ public class PinotHelixResourceManager {
    * @return True if instance exists in the Helix cluster, False otherwise.
    */
   public boolean instanceExists(String instanceName) {
-    HelixDataAccessor helixDataAccessor = _helixZkManager.getHelixDataAccessor();
-    InstanceConfig config = helixDataAccessor.getProperty(_keyBuilder.instanceConfig(instanceName));
-    return (config != null);
+    return getHelixInstanceConfig(instanceName) != null;
   }
 
   public boolean isSingleTenantCluster() {
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
@@ -21,7 +21,9 @@ package org.apache.pinot.controller.helix.core;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Strings;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.HashBiMap;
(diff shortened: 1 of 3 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -117,11 +118,13 @@
 public class PinotHelixResourceManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(PinotHelixResourceManager.class);
   private static final long DEFAULT_EXTERNAL_VIEW_UPDATE_RETRY_INTERVAL_MILLIS = 500L;
+  private static final long CACHE_TIMEOUT_HR = 6L;
```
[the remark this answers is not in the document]
Jackie-Jiang: Actually `CACHE_ENTRY_EXPIRE_TIME_HOURS` is more accurate. We don't want to use refresh as the performance is not that critical, and the entry can be removed.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManagerTest.java
```diff
@@ -173,24 +172,18 @@ private void addAndRemoveNewInstanceConfig(ZkClient zkClient)
...
     zkClient.createPersistent(instanceConfigPath, znRecord);
 
-    List<String> latestAllInstances = _helixResourceManager.getAllInstances();
-    long maxTime = System.currentTimeMillis() + MAX_TIMEOUT_IN_MILLISECOND;
-    while (!latestAllInstances.contains(instanceName) && System.currentTimeMillis() < maxTime) {
-      Thread.sleep(100L);
-      latestAllInstances = _helixResourceManager.getAllInstances();
-    }
-    Assert.assertTrue(System.currentTimeMillis() < maxTime, "Timeout when waiting for adding instance config");
+    TestUtils.waitForCondition(aVoid -> {
```
Jackie-Jiang: We don't need to wait for condition right?

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManagerTest.java
```diff
@@ -173,24 +172,18 @@ private void addAndRemoveNewInstanceConfig(ZkClient zkClient)
...
     zkClient.delete(instanceConfigPath);
 
-    latestAllInstances = _helixResourceManager.getAllInstances();
-    maxTime = System.currentTimeMillis() + MAX_TIMEOUT_IN_MILLISECOND;
-    while (latestAllInstances.contains(instanceName) && System.currentTimeMillis() < maxTime) {
-      Thread.sleep(100L);
-      latestAllInstances = _helixResourceManager.getAllInstances();
-    }
-    Assert.assertTrue(System.currentTimeMillis() < maxTime, "Timeout when waiting for removing instance config");
+    TestUtils.waitForCondition(aVoid -> {
```
Jackie-Jiang: Same here

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManagerTest.java
jackjlli: It's possible that direct data accessor still fails to get the instance config because of some other issues like zk connection. Let's still keep the wait for condition here.

kishoreg:
Can you please give me sometime. Adding caching on top of ZK is not trivial and in the end we will end up implementing Helix caching code in Pinot.

jackjlli:
@mcvsubbu The CI build failed even after we removed cached data accessor. Previously the test did have some flakiness issue.
I've refactored the test in the latest push to make the test more robust.

Jackie-Jiang:
> Can you please give me sometime. Adding caching on top of ZK is not trivial and in the end we will end up implementing Helix caching code in Pinot.

This PR is reverting the cache introduced in #3633 that caused some problem, so actually we are on the same page

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -304,8 +302,7 @@ private void addInstanceGroupTagIfNeeded() {
    * @return Helix instance config
    */
   public InstanceConfig getHelixInstanceConfig(@Nonnull String instanceId) {
-    ZNRecord znRecord = _cacheInstanceConfigsDataAccessor.get("/" + instanceId, null, AccessOption.PERSISTENT);
-    return znRecord != null ? new InstanceConfig(znRecord) : null;
+    return _helixAdmin.getInstanceConfig(_helixClusterName, instanceId);
```
Jackie-Jiang: This will throw exception when instance does not exist. You might want to use `_helixDataAccessor.getProperty(_keyBuilder.instanceConfig(instanceId))`

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -2223,8 +2220,9 @@ public RebalanceResult rebalanceTable(String tableNameWithType, Configuration re
    * @return True if instance exists in the Helix cluster, False otherwise.
    */
   public boolean instanceExists(String instanceName) {
-    ZNRecord znRecord = _cacheInstanceConfigsDataAccessor.get("/" + instanceName, null, AccessOption.PERSISTENT);
-    return (znRecord != null);
+    HelixDataAccessor helixDataAccessor = _helixZkManager.getHelixDataAccessor();
```
Jackie-Jiang: `return getHelixInstanceConfig(instanceName) != null`?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
jackjlli: Updated. 👍

PR #4680, opened 2019-10-05
Ensure segment converter not fail for schemas without time fieldSpec

fx19880617:


Jackie-Jiang:
@fx19880617 For real-time table, time field is mandatory. It should fail without time field.

npawar:
@fx19880617 i know this was a while ago, but do you remember the context of why this was done? I thought we don't want to support a realtime table without a time column. I think we also have validations in a bunch of places for this. Did we decide to make time column non-mandatory?

kishoreg:
I think it was to make the behavior similar to batch mode. One has a schema that works on offline table and then moves it to real-time and it suddenly doesn't work.

Another reason is - there is nothing in Pinot that makes time field mandatory. It's just a bunch of columns that one can query just like any other database. I  don't see any reason to enforce that

PR #4685, opened 2019-10-08
[TE] handle data insufficient exceptions in the detection pipeline

jihaozh:
When new metrics are on-boarded into ThirdEye, we expect the detector to throw a checked exception on the case that the data is insufficient to make a decision on whether it is an anomaly. 

This RB makes that if such a case happens, instead of failing the detection pipeline, it will check if there are any other detectors successfully executes and keep the result.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
@@ -21,9 +21,19 @@ package org.apache.pinot.thirdeye.detection.algorithm;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Multimap;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import org.apache.commons.collections4.MapUtils;
 import org.apache.pinot.thirdeye.dataframe.DataFrame;
 import org.apache.pinot.thirdeye.dataframe.util.MetricSlice;
 import org.apache.pinot.thirdeye.datalayer.dto.DetectionConfigDTO;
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -278,22 +286,48 @@ public DetectionPipelineResult run() throws Exception {
...
+   */
+  private void checkNestedMetricsStatus(long totalNestedMetrics, long successNestedMetrics, List<Exception> exceptions,
+      List<PredictionResult> predictions) throws DetectionPipelineException {
     // if all nested metrics failed, throw an exception
     if (successNestedMetrics == 0 && totalNestedMetrics > 0) {
-      throw new DetectionPipelineException(String.format(
-          "Detection failed for all nested dimensions for detection config id %d for monitoring window %d to %d.",
-          this.config.getId(), this.getStartTime(), this.getEndTime()), lastException);
+      // if all exceptions are caused by DetectorDataInsufficientException and
+      // there are other detectors run successfully, keep the detection running
```
vincentchenjl: Just curious what is the behavior if detectors in some dimensions throws other exceptions than `DetectorDataInsufficientException`. Do we still consider the detection is running successfully if one of the dimension fails due to other exceptions?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -278,22 +286,48 @@ public DetectionPipelineResult run() throws Exception {
...
+      if (exceptions.stream().allMatch(e -> e.getCause() instanceof DetectorDataInsufficientException)
+          && predictions.size() != 0) {
```
vincentchenjl: I think that `predictions.size() == 0` will always true if `successNestedMetrics == 0` is true. Please let me know if I miss something here.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
jihaozh: If any exception is thrown other than  `DetectorDataInsufficientException` in any dimension, it will fail the whole detection.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
jihaozh: Actually, `predictions` is the total number of detectors run. `successNestedMetrics` is the number of dimensions with all detector successfully executes. If `successNestedMetrics == 0` but `predictions.size() =! 0`, means that there are some detectors runs sucessfully.

PR #4690, opened 2019-10-09
[TE] Fix ignored timeout by checking Thread.interrupted

vincentchenjl:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -249,6 +249,12 @@ public DetectionPipelineResult run() throws Exception {
       checkEarlyStop(totalNestedMetrics, successNestedMetrics, i, lastException);
       MetricEntity metric = nestedMetrics.get(i);
       try {
+        if (Thread.interrupted()) {
```
jihaozh: Mind add some comment here specific when will the thread get interrupted? For example, when timing out or shutdown.

vincentchenjl: Just added the comment.

xiaohui-sun: Please also add more context like below "Detection with multiple dimensions may take a significantly long time, which would be cancelled from TaskDriver."

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -249,6 +249,13 @@ public DetectionPipelineResult run() throws Exception {
...
+          // this happens when Future.cancel() or ExecutorService.shutdown() is called.
+          LOG.error("Current thread is interrupted before running dimension {}/{} for detection {} metrics {}, "
+                  + "so skip the rest of dimensions and return the intermediate",
```
xiaohui-sun: we don't need  the first "+".

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -249,6 +249,13 @@ public DetectionPipelineResult run() throws Exception {
...
+              i + 1 , totalNestedMetrics, this.config.getId(), metric.getUrn());
+          break;
```
xiaohui-sun: Add a comment: "We still want to keep the current detection result and store it anyway."

PR #4691, opened 2019-10-09
[TE] remove erroneous import

jihaozh:
This PR removes an erroneous unused import which can block local compilation.


harleyjj:
Looks great!

PR #4693, opened 2019-10-09
Fetch new schema when loading committed real-time segments

Jackie-Jiang:
We added real-time reload support in #4335 and #4349 to generate
default columns for committed segments. But the consuming segments
won't have default columns generated. In order to support default
columns for real-time segments, when loading the committed real-time
segments, we need to fetch the latest schema from the cluster.

With this fix, after applying the new schema, new committed segments
will have the default columns generated for the new added columns.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
@@ -34,7 +34,6 @@ import org.apache.commons.io.FileUtils;
 import org.apache.pinot.common.Utils;
 import org.apache.pinot.common.config.IndexingConfig;
 import org.apache.pinot.common.config.TableConfig;
-import org.apache.pinot.common.config.TableNameBuilder;
 import org.apache.pinot.common.data.FieldSpec;
 import org.apache.pinot.common.data.Schema;
 import org.apache.pinot.common.metadata.ZKMetadataProvider;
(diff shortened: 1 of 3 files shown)


pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -205,9 +196,18 @@ public void notifySegmentCommitted(String tableNameWithType, RealtimeSegmentZKMe
...
+      _logger.warn("Skipping adding existing segment: {} for table: {} with data manager class: {}", segmentName,
+          _tableNameWithType, segmentDataManager.getClass().getSimpleName());
+      return;
+    }
+
     RealtimeSegmentZKMetadata realtimeSegmentZKMetadata =
         ZKMetadataProvider.getRealtimeSegmentZKMetadata(_propertyStore, _tableNameWithType, segmentName);
     Preconditions.checkNotNull(realtimeSegmentZKMetadata);
+    Schema schema = ZKMetadataProvider.getTableSchema(_propertyStore, _tableNameWithType);
+    Preconditions.checkNotNull(schema);
```
haibow: Why require not null here and not in other places (replaceHLSegment and replaceLLSegment)? 

Might be worth flagging it, but we could still keep the schema null and let it load with the segment metadata as fallback?

Jackie-Jiang: This code path is shared for committing segment as well as new consuming segment (check line242 in old file).
For real-time table, schema is mandatory, so this should not cause any problem.

haibow:
LGTM. Thanks for the fix!

haibow:
> LGTM. Just curious how we handle consuming(mutable) segment after adding the default value to the schema.

Good point - it remains an issue. The consuming segment would have the old schema, while previously committed segments would get the new schema after the reload.

As a result, if the user queries the newly added field (or select *) between the reload and the consuming segment is committed, some segments (either the consuming segment or all other segments, most likely the former) would be dropped from the query result.

This diff is part of the efforts to fully address the schema evolution story. Slack channel is #feat-better-schema-evolution if you are interested.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -205,9 +196,18 @@ public void notifySegmentCommitted(String tableNameWithType, RealtimeSegmentZKMe
   public void addSegment(@Nonnull String segmentName, @Nonnull TableConfig tableConfig,
       @Nonnull IndexLoadingConfig indexLoadingConfig)
       throws Exception {
+    SegmentDataManager segmentDataManager = _segmentDataManagerMap.get(segmentName);
+    if (segmentDataManager != null) {
+      _logger.warn("Skipping adding existing segment: {} for table: {} with data manager class: {}", segmentName,
```
mcvsubbu: you don't need table name or class name  since  _logger already has these in it. Please remove them. Also, please clarify the message in the log. "Skipping because we already found the segment loaded". We should not need to look into  the code to check the reason.

mcvsubbu:
@Jackie-Jiang

PR #4694, opened 2019-10-10
Add groupByMode=sql and responseFormat=sql options to query console

npawar:
Screen Shot 2019-10-09 at 5 48 23 PM

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PqlQueryResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PqlQueryResource.java
@@ -76,8 +76,12 @@ public class PqlQueryResource {
       if (requestJson.has("trace")) {
         traceEnabled = requestJson.get("trace").toString();
       }
+      String queryOptions = null;
+      if (requestJson.has("queryOptions")) {
+        queryOptions = requestJson.get("queryOptions").asText();
+      }
       LOGGER.debug("Trace: {}, Running query: {}", traceEnabled, pqlQuery);
-      return getQueryResponse(pqlQuery, traceEnabled, httpHeaders);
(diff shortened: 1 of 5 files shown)


kishoreg:
Nice!. Enable word seems redundant. It can be part of the tool tip if needed.
Group By Mode: Sql/sql Pql on by default.
Query Response Format: sql pql. pql on by default

npawar:
> Nice!. Enable word seems redundant. It can be part of the tool tip if needed.
> Group By Mode: Sql/sql Pql on by default.
> Query Response Format: sql pql. pql on by default

Reworded and added tooltip. Are you suggesting radio buttons or toggle button?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PqlQueryResource.java
```diff
@@ -76,8 +76,12 @@ public String post(String requestJsonStr, @Context HttpHeaders httpHeaders) {
       if (requestJson.has("trace")) {
         traceEnabled = requestJson.get("trace").toString();
       }
+      String queryOptions = null;
+      if (requestJson.has("queryOptions")) {
+        queryOptions = requestJson.get("queryOptions").asText();
+      }
       LOGGER.debug("Trace: {}, Running query: {}", traceEnabled, pqlQuery);
-      return getQueryResponse(pqlQuery, traceEnabled, httpHeaders);
+      return getQueryResponse(pqlQuery, traceEnabled, queryOptions, httpHeaders);
```
kishoreg: traceEnabled can also be part of queryOptions rt

npawar: If we move trace to queryOptions, the callers of the broker API will have to be changed to pass trace in queryOptions.

npawar:

PR #4695, opened 2019-10-10
[Resource Assignment] Plug in resource assignment to LLC REALTIME table

Jackie-Jiang:
This is the last PR to plug in the new resource assignment, after this PR, all the segment assignment/rebalance will use the new strategy.
Uniform the OFFLINE/REALTIME table segment assignment behavior to allow assigning segments to disabled instances to prevent segment shuffling if servers got disabled.

Re-write most of the PinotLLCRealtimeSegmentManager to use the new assignment strategy:
- Remove the boolean return and change it to throw exception when operation failed (old code can throw exception and also return false, which is hard to manage)
- Remove the controller leadership check inside the PinotLLCRealtimeSegmentManager to prevent responding FAILED in case of NOT_LEADER. The check is already performed in SegmentCompletionManager
- Remove the controller connection check because ZK access will implicitly check that
- Enhance the segment repair logic to correctly skip OFFLINE segments
- Introduce the LLC_ZOOKEEPER_FETCH_FAILURES in controller metrics to track ZK exceptions

Re-write the PinotLLCRealtimeSegmentManagerTest to test all the changes:
- testCommittingSegmentIfDisconnected is removed because we don't explicitly check disconnection but fail with the ZK exception
- testIdealStateAlreadyUpdated is covered by testCommitSegment committing the segment again


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1184,168 +898,127 @@ protected IdealState ensureAllPartitionsConsuming(final TableConfig tableConfig,
...
-        if (secondLatestMetadata == null && skipNewPartitions) {
+        if (!isExceededMaxSegmentCompletionTime(realtimeTableName, latestSegmentName, currentTimeMs)) {
           continue;
         }
-        if (secondLatestMetadata != null) {
-          onlineSegments.add(secondLatestMetadata.getSegmentName());
+        LOGGER.info("Repairing segment: {} which has segment ZK metadata but does not exist in IdealState",
+            latestSegmentName);
+
+        if (latestSegmentZKMetadata.getStatus() == Status.IN_PROGRESS) {
```
npawar: previously, we were getting metadata for the last 2 segments. Just curious, why did you change that to just latest segment, and instead opt to get the second last metadata here, by scanning all the instanceStatesMap for every partition?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1184,168 +898,127 @@ protected IdealState ensureAllPartitionsConsuming(final TableConfig tableConfig,
...
-    for (String instance : currentSegmentInstances) {
-      idealState.setPartitionState(currentSegmentId, instance,
-          PinotHelixSegmentOnlineOfflineStateModelGenerator.ONLINE_STATE);
+  private int getNumReplicas(TableConfig tableConfig, InstancePartitions instancePartitions) {
+    if (instancePartitions.getNumReplicaGroups() == 1) {
+      // Non-replica-group based
+      return tableConfig.getValidationConfig().getReplicasPerPartitionNumber();
+    } else {
+      // Replica-group based
+      return instancePartitions.getNumReplicaGroups();
```
npawar: even if it is replica group based, shouldn't the numReplicas match the `tableConfig.getValidationConfig().getReplicasPerPartitionNumber()` ?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1184,168 +898,127 @@ protected IdealState ensureAllPartitionsConsuming(final TableConfig tableConfig,
...
-        idealState.setPartitionState(newSegmentId, instance,
-            PinotHelixSegmentOnlineOfflineStateModelGenerator.CONSUMING_STATE);
-      }
+  private int getMaxNumPartitionsPerInstance(InstancePartitions instancePartitions, int numPartitions,
+      int numReplicas) {
+    if (instancePartitions.getNumReplicaGroups() == 1) {
+      // Non-replica-group based assignment:
+      // Uniformly spray the partitions and replicas across the instances.
+      // E.g. (6 instances, 3 partitions, 4 replicas)
+      // "0_0": [i0,  i1,  i2,  i3,  i4,  i5  ]
```
npawar: typo in this example -> p0r0 p0r1 p0r2 `p0r3`

pinot-common/src/main/java/org/apache/pinot/common/config/IndexingConfig.java
```diff
@@ -56,9 +56,6 @@
   @UseChildKeyHandler(SimpleMapChildKeyHandler.class)
   private Map<String, String> _streamConfigs = new HashMap<>();
 
-  @ConfigKey("streamConsumptionConfig")
```
mcvsubbu: Is this a backward incompatible change?

pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
```diff
@@ -642,11 +640,6 @@ public Builder setStreamConfigs(Map<String, String> streamConfigs) {
       return this;
     }
 
-    public Builder setStreamPartitionAssignmentStrategy(String streamPartitionAssignmentStrategy) {
```
mcvsubbu: Is this a backward incompatible change?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1161,27 +1161,31 @@ private void verifyIndexingConfig(String tableNameWithType, IndexingConfig index
...
   }
 
-  private void ensureRealtimeClusterIsSetUp(TableConfig config, String realtimeTableName,
-      IndexingConfig indexingConfig) {
-    StreamConfig streamConfig = new StreamConfig(realtimeTableName, indexingConfig.getStreamConfigs());
-    IdealState idealState = _helixAdmin.getResourceIdealState(_helixClusterName, realtimeTableName);
+  private void ensureRealtimeClusterIsSetUp(TableConfig realtimeTableConfig) {
+    String realtimeTableName = realtimeTableConfig.getTableName();
+    IdealState idealState = getTableIdealState(realtimeTableName);
+    // Skip processing disabled IdealState
```
mcvsubbu: We should be able to add/remove llc even if table is disabled, I would think

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1334,12 +1326,6 @@ public void updateIndexingConfigFor(String tableName, TableType type, IndexingCo
     }
     tableConfig.setIndexingConfig(newConfigs);
     setExistingTableConfig(tableConfig);
-
-    if (type == TableType.REALTIME) {
```
mcvsubbu: This was an optimization we had so that controllers do not need to set watches on HLC tables if they are not there in the cluster. Not sure why it has been removed, or if it has been moved elsewhere

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -181,183 +186,183 @@ public void stop() {
...
+  public void setUpNewTable(TableConfig tableConfig, IdealState idealState) {
+    Preconditions.checkState(!_isStopping, "Segment manager is stopping");
+
+    String realtimeTableName = tableConfig.getTableName();
+    LOGGER.info("Setting up new LLC table: {}", realtimeTableName);
+
+    // Make sure all the existing segments are HLC segments
+    List<String> currentSegments = getAllSegments(realtimeTableName);
+    for (String segmentName : currentSegments) {
+      Preconditions.checkState(SegmentName.isHighLevelConsumerSegmentName(segmentName),
```
mcvsubbu: Does this show up correctly in the table create command line on the client side?

pinot-common/src/main/java/org/apache/pinot/common/assignment/InstancePartitionsUtils.java
```diff
@@ -89,18 +91,32 @@ public static InstancePartitions fetchInstancePartitions(HelixPropertyStore<ZNRe
   }
 
   /**
-   * Computes the default instance partitions.
-   * <p>For backward-compatibility, sort all enabled instances with the server tag, rotate the list based on the table
-   * name name to prevent creating hotspot servers.
+   * Computes the default instance partitions. Sort all qualified instances and rotate the list based on the table name
+   * to prevent creating hotspot servers.
+   * <p>For backward-compatibility, choose only enabled instances with the server tag for OFFLINE table; choose both
```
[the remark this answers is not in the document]
Jackie-Jiang: Very good suggestion. Changed.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1190,32 +1194,20 @@ private void ensureRealtimeClusterIsSetUp(TableConfig config, String realtimeTab
...
-    }
-    LOGGER.info("Successfully created empty ideal state for  high level consumer for {} ", realtimeTableName);
-    // Finally, create the propertystore entry that will trigger watchers to create segments
-    String tablePropertyStorePath = ZKMetadataProvider.constructPropertyStorePathForResource(realtimeTableName);
-    if (!_propertyStore.exists(tablePropertyStorePath, AccessOption.PERSISTENT)) {
-      _propertyStore.create(tablePropertyStorePath, new ZNRecord(realtimeTableName), AccessOption.PERSISTENT);
+  private void ensurePropertyStoreEntryExistsForHighLevelConsumer(String realtimeTableName) {
+    String propertyStorePath = ZKMetadataProvider.constructPropertyStorePathForResource(realtimeTableName);
+    if (!_propertyStore.exists(propertyStorePath, AccessOption.PERSISTENT)) {
+      LOGGER.info("Creating property store entry for LLC table: {}", realtimeTableName);
```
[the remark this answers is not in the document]
Jackie-Jiang: Fixed. HLC relies on the watchers on the property store entry to create segments (commented on line 1187)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -181,183 +186,183 @@ public void stop() {
...
+
+    SegmentAssignment segmentAssignment = SegmentAssignmentFactory.getSegmentAssignment(_helixManager, tableConfig);
+    Map<InstancePartitionsType, InstancePartitions> instancePartitionsMap =
+        Collections.singletonMap(InstancePartitionsType.CONSUMING, instancePartitions);
+
+    long currentTimeMs = getCurrentTimeMs();
+    Map<String, Map<String, String>> instanceStatesMap = idealState.getRecord().getMapFields();
+    for (int partitionId = 0; partitionId < numPartitions; partitionId++) {
+      String segmentName =
+          setupNewPartition(tableConfig, partitionLevelStreamConfig, partitionId, currentTimeMs, instancePartitions,
```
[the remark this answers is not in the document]
Jackie-Jiang: I considered that, but that will make the method signature quite redundant. Since this is not performance sensitive path, I just cached the expensive operations such as extracting stream config from table config

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -181,183 +186,183 @@ public void stop() {
...
+  void setIdealState(String realtimeTableName, IdealState idealState) {
+    try {
+      _helixAdmin.setResourceIdealState(_clusterName, realtimeTableName, idealState);
+    } catch (Exception e) {
+      _controllerMetrics.addMeteredTableValue(realtimeTableName, ControllerMeter.LLC_ZOOKEEPER_UPDATE_FAILURES, 1L);
+      throw e;
     }
+  }
+
+  public void commitSegmentFile(String realtimeTableName, CommittingSegmentDescriptor committingSegmentDescriptor)
```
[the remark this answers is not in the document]
Jackie-Jiang: Added

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -181,183 +186,183 @@ public void stop() {
...
+    // Copy the segment file to the controller
     String segmentLocation = committingSegmentDescriptor.getSegmentLocation();
     URI segmentFileURI = URIUtils.getUri(segmentLocation);
-    URI tableDirURI = URIUtils.getUri(_controllerConf.getDataDir(), tableName);
-    URI uriToMoveTo = URIUtils.getUri(_controllerConf.getDataDir(), tableName, URIUtils.encode(segmentName));
+    URI tableDirURI = URIUtils.getUri(_controllerConf.getDataDir(), rawTableName);
+    URI uriToMoveTo = URIUtils.getUri(_controllerConf.getDataDir(), rawTableName, URIUtils.encode(segmentName));
     PinotFS pinotFS = PinotFSFactory.create(tableDirURI.getScheme());
-
-    if (!isConnected() || !isLeader(tableName)) {
```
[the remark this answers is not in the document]
Jackie-Jiang: Connection is checked when accessing the ZK, so no need to explicitly check connection.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: We cannot rely on the segment ZK metadata because previous segment might get all replicas OFFLINE, and we need to skip such segments (covered by the tests).
There is no ZK fetch for the instanceStatesMap scan, so it is actually much cheaper than fetching one more segment ZK metadata for each partition.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Not always. If we assign instances with instanceAssignmentConfig, it can be different from  `tableConfig.getValidationConfig().getReplicasPerPartitionNumber()`. We will log warning when assigning instances if these 2 numbers not match.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Good catch

pinot-common/src/main/java/org/apache/pinot/common/config/IndexingConfig.java
Jackie-Jiang: This config is for the real-time replica-group, which is not supported yet.
After checking in this pr, real-time replica-group will be supported with the instance partitions.

pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
Jackie-Jiang: Same here

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: It is already checked in `setExistingTableConfig`. This check is redundant

pinot-common/src/main/java/org/apache/pinot/common/config/TableConfig.java
Jackie-Jiang: Same here

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Yes, both (IllegalStateException and RuntimeException) are returned as INTERNAL_SERVER_ERROR

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: Correct. Removed the enable check

Jackie-Jiang:
> Can you minimize the changes to the ones you absolutely need? We can do a separate cleanup checkin. Otherwise it gets very hard for reviewers

@mcvsubbu I started that way, but with boolean type return as well as exception inside the method, it becomes really hard to handle unexpected behaviors.
Regardless of the changes in tests, most of the changes are in PinotLLCRealtimeSegmentManager because the segment assignment logic needs to be replaced.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
```diff
@@ -969,7 +971,13 @@ private int numReplicasToLookFor() {
       LOGGER
           .info("Instance {} stopped consuming segment {} at offset {}, state {}, createNew: {}, reason:{}", instanceId,
               _segmentName, offset, _state, createNew, reason);
-      _segmentManager.segmentStoppedConsuming(_segmentName, instanceId);
+      try {
+        _segmentManager.segmentStoppedConsuming(_segmentName, instanceId);
+      } catch (Exception e) {
+        LOGGER.error("Caught exception while processing stopped CONSUMING segment: {} on instance: {}",
```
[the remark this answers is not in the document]
Jackie-Jiang: Good catch, missed that. Added

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
```diff
@@ -1028,17 +1034,24 @@ private int numReplicasToLookFor() {
       // In case of splitCommit, the segment is uploaded to a unique file name indicated by segmentLocation,
       // so we need to move the segment file to its permanent location first before committing the metadata.
       if (isSplitCommit) {
-        if (!_segmentManager.commitSegmentFile(_segmentName.getTableName(), committingSegmentDescriptor)) {
+        try {
+          _segmentManager.commitSegmentFile(_realtimeTableName, committingSegmentDescriptor);
+        } catch (Exception e) {
+          LOGGER.error("Caught exception while committing segment file for segment: {}", _segmentName.getSegmentName());
```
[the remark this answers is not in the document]
Jackie-Jiang: Added

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionManager.java
```diff
@@ -1028,17 +1034,24 @@ private int numReplicasToLookFor() {
...
-      success = _segmentManager.commitSegmentMetadata(_segmentName.getTableName(), committingSegmentDescriptor);
-      if (success) {
-        _state = State.COMMITTED;
-        LOGGER.info("Committed segment {} at offset {} winner {}", _segmentName.getSegmentName(), offset, instanceId);
-        return SegmentCompletionProtocol.RESP_COMMIT_SUCCESS;
+      try {
+        _segmentManager.commitSegmentMetadata(_realtimeTableName, committingSegmentDescriptor);
+      } catch (Exception e) {
+        LOGGER
+            .error("Caught exception while committing segment metadata for segment: {}", _segmentName.getSegmentName());
```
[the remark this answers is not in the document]
Jackie-Jiang: Added

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -657,216 +567,88 @@ public long getCommitTimeoutMS(String tableName) {
...
    * Mark the state of the segment to be OFFLINE in idealstate.
    * When all replicas of this segment are marked offline, the {@link org.apache.pinot.controller.validation.RealtimeSegmentValidationManager},
    * in its next run, will auto-create a new segment with the appropriate offset.
    */
-  public void segmentStoppedConsuming(final LLCSegmentName segmentName, final String instance) {
-    String rawTableName = segmentName.getTableName();
-    String realtimeTableName = TableNameBuilder.REALTIME.tableNameWithType(rawTableName);
-    final String segmentNameStr = segmentName.getSegmentName();
+  public void segmentStoppedConsuming(LLCSegmentName llcSegmentName, String instanceName) {
+    Preconditions.checkState(!_isStopping, "Segment manager is stopping");
```
mcvsubbu: if the segment manger is stopping, we should return NOT_LEADER, so that pinot-server can connect to another leader and do the operation again. Same for other public operations initiated by the server.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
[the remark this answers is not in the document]
Jackie-Jiang: Early termination is not required, and can backfire in certain cases (transient disconnection/no leader).
When committing a segment, there are 3 steps of modifying the ZK states. As long as we ensure all states can be fixed by the validation manager, we are good.
I agree we should keep the existing known issue guarded. This PR keeps (also enhances) the tests for such scenarios addressed before.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Added the TODO. This PR keeps all the old API behaviors the same.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -657,216 +567,88 @@ public long getCommitTimeoutMS(String tableName) {
...
-      LOGGER.error("Could not get offset for topic {} partition {}, criteria {}", streamConfig.getTopicName(),
-          partitionId, offsetCriteria, fetcherException);
-      throw new RuntimeException(fetcherException);
+      throw new IllegalStateException(String
+          .format("Failed to fetch the offset for topic: %s, partition: %s with criteria: %s",
+              streamConfig.getTopicName(), partitionId, offsetCriteria));
     }
   }
 
-  private long getBetterStartOffsetIfNeeded(final String realtimeTableName, final int partition,
```
mcvsubbu: where did this logic go?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -182,183 +186,188 @@ public void stop() {
...
-    int removeCount = 0;
-    for (String segmentName : allSegments) {
+    List<String> segmentsToRemove = new ArrayList<>();
+    for (String segmentName : idealState.getRecord().getMapFields().keySet()) {
       if (SegmentName.isLowLevelConsumerSegmentName(segmentName)) {
         segmentsToRemove.add(segmentName);
-        removeCount++;
       }
     }
-    LOGGER.info("Attempting to remove {} LLC segments of table {}", removeCount, realtimeTableName);
```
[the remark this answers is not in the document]
Jackie-Jiang: Yes, it is logged in `PinotHelixResourceManager.deleteSegments()`

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -367,48 +376,44 @@ public boolean commitSegmentFile(String tableName, CommittingSegmentDescriptor c
     try {
       for (String uri : pinotFS.listFiles(tableDirURI, false)) {
         if (uri.contains(SegmentCompletionUtils.getSegmentNamePrefix(segmentName))) {
-          LOGGER.warn("Deleting " + uri);
-          pinotFS.delete(new URI(uri), true);
+          LOGGER.warn("Deleting temporary segment file: {}", uri);
+          Preconditions.checkState(pinotFS.delete(new URI(uri), true), "Failed to delete file: %s", uri);
```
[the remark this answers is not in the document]
Jackie-Jiang: Yes, but it can only recognize `%s`

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionTest.java
```diff
@@ -1153,44 +1151,32 @@ protected MockPinotLLCRealtimeSegmentManager(PinotHelixResourceManager pinotHeli
...
-        return false;
-      } else {
-        return true;
-      }
+    public void commitSegmentFile(String rawTableName, CommittingSegmentDescriptor committingSegmentDescriptor) {
+      Preconditions.checkState(!committingSegmentDescriptor.getSegmentLocation().equals("doNotCommitMe"));
     }
 
     @Override
-    protected void writeSegmentsToPropertyStore(List<String> paths, List<ZNRecord> records, String realtimeTableName) {
```
mcvsubbu: Not sure how we are testing the callers of this method

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/realtime/segment/FlushThresholdUpdaterTest.java
```diff
@@ -18,699 +18,354 @@
...
-import org.testng.Assert;
-import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
-/*
- * TODO: FIXME
- * - Consider changing the test to use mock streamConfig than the one generated via real configurations.
- *   Makes things a lot easier. Or, have a class that inherits from StreamConfig and has setters/getters
- *   to return the values that we care about.
- * - Change the SegmentSizeBasedFlushThresholdUpdater class to use a method to get time, and fake it
```
mcvsubbu: not sure if this TODO has been implementated. if so, then please change the TODO to the test to use fake time instead (this PR is already too big)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Discussed offline. The leader check are performed before getting into this class.

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/realtime/SegmentCompletionTest.java
Jackie-Jiang: `persistSegmentZKMetadata()` can be overridden by tests

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Line 874-884

pinot-controller/src/test/java/org/apache/pinot/controller/helix/core/realtime/segment/FlushThresholdUpdaterTest.java
Jackie-Jiang: Added the TODO

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1151,34 +866,32 @@ protected IdealState ensureAllPartitionsConsuming(final TableConfig tableConfig,
...
-                tableNameWithType, partition);
-
-            // To begin with, set startOffset to the oldest available offset in the stream. Fix it to be the one we want,
-            // depending on what the prev segment had.
-            long startOffset = getPartitionOffset(streamConfig, OffsetCriteria.SMALLEST_OFFSET_CRITERIA, partition);
-            LOGGER.info("Found smallest offset {} for table {} for partition {}", startOffset, tableNameWithType,
-                partition);
-            startOffset = getBetterStartOffsetIfNeeded(tableNameWithType, partition, segmentName, startOffset,
-                newLLCSegmentName.getSequenceNumber());
+          if (isAllInstancesInState(instanceStateMap, RealtimeSegmentOnlineOfflineStateModel.OFFLINE)) {
```
mcvsubbu: You need to also check if it is too soon to correct this segment.

mcvsubbu: You are not handling the case where all instances are in ONLINE state, but there are no CONSUMING segments. Basically, handle PR #2891

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1151,34 +866,32 @@ protected IdealState ensureAllPartitionsConsuming(final TableConfig tableConfig,
...
-              consumingSegments.add(newLLCSegmentName.getSegmentName());
-            }
+                new CommittingSegmentDescriptor(latestSegmentName, startOffset, 0);
+            createNewSegmentZKMetadata(tableConfig, streamConfig, newLLCSegmentName, currentTimeMs,
+                committingSegmentDescriptor, latestSegmentZKMetadata, instancePartitions, numPartitions, numReplicas);
+            String newSegmentName = newLLCSegmentName.getSegmentName();
+            updateInstanceStatesForNewConsumingSegment(instanceStatesMap, null, newSegmentName, segmentAssignment,
+                instancePartitionsMap);
+          } else {
+            LOGGER.error("Got unexpected instance state map: {} for segment: {}", instanceStateMap, latestSegmentName);
```
mcvsubbu: Why is this unexpected state?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -992,37 +755,24 @@ private IdealState setupTable(TableConfig tableConfig, IdealState idealState, in
...
-    long metadataUpdateTime = stat.getMtime();
-    if (now > metadataUpdateTime + MAX_SEGMENT_COMPLETION_TIME_MILLIS) {
-      LOGGER
-          .info("Segment:{}, Now:{}, metadataUpdateTime:{}, Exceeded MAX_SEGMENT_COMPLETION_TIME_MILLIS:{}", segmentId,
-              now, metadataUpdateTime, MAX_SEGMENT_COMPLETION_TIME_MILLIS);
-      return true;
-    }
-    return false;
+    getSegmentZKMetadata(realtimeTableName, segmentName, stat);
+    return currentTimeMs > stat.getMtime() + MAX_SEGMENT_COMPLETION_TIME_MILLIS;
```
mcvsubbu: We found it useful to log here. Can you please add back the log?

Jackie-Jiang: Added back

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: We don't need to check for all OFFLINE case. That is not a valid state for segment commit.
The `all in ONLINE state` scenario is handled in RetentionManager (#2890 as in the comments). #2891 does not handle that scenario as well.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: This is for debugging purpose to log unexpected state that is not properly handled. E.g. `all in ONLINE state` scenario is not handled properly in RetentionManager

mcvsubbu: @npawar  and I discussed and came to the conclusion that we were being conservative when we handle the all ONLINE case. Can you please add a comment to that effect?

mcvsubbu:
I have asked @npawar  to look at the class from the standpoint of:
- What if an exception is raised where we would have returned a boolean 'false' before. It seemed like we would continue to do what we can, but now we abort.
- Log messages: Used to be that every step is logged, so that we can reconstruct what happened during a repair or segment assignment. It will help in reasoning about  problems like : "this segment was not repaired", "this partition stopped consuming". With the removal of logs, we need to be careful that we clearly log an exception that may have happened while we try to do an operation, or otherwise log the branches that we may take that may lead us to NOT completing an operation.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Added more comment explaining what needs to be handled when getting this error message.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1103,46 +835,35 @@ protected IdealState ensureAllPartitionsConsuming(final TableConfig tableConfig,
...
-            if (isTooSoonToCorrect(tableNameWithType, segmentId, now)) {
+            if (!isExceededMaxSegmentCompletionTime(realtimeTableName, latestSegmentName, currentTimeMs)) {
               continue;
             }
-            LOGGER.info("{}:Repairing segment for partition {}. "
-                    + "Old segment metadata {} has status DONE, but segments are still in CONSUMING state in ideal STATE",
-                tableNameWithType, partition, segmentId);
-
-            LLCSegmentName newLLCSegmentName = makeNextLLCSegmentName(segmentName, partition, now);
-            LOGGER.info("{}: Creating new segment metadata for {}", tableNameWithType,
```
npawar: Can we add this log back in some form? Basically to know which new segment metadata were created by the ValidationManager

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1151,34 +872,36 @@ protected IdealState ensureAllPartitionsConsuming(final TableConfig tableConfig,
           // 1. all replicas OFFLINE and metadata IN_PROGRESS/DONE - a segment marked itself OFFLINE during consumption for some reason
           // 2. all replicas ONLINE and metadata DONE - Resolved in https://github.com/linkedin/pinot/pull/2890
           // 3. we should never end up with some replicas ONLINE and some OFFLINE.
-          if (isAllInstancesInState(instanceStateMap, PinotHelixSegmentOnlineOfflineStateModelGenerator.OFFLINE_STATE)
-              || !(isTooSoonToCorrect(tableNameWithType, segmentId, now))) {
-
-            // No instances are consuming, so create a new consuming segment.
-            LLCSegmentName newLLCSegmentName = makeNextLLCSegmentName(segmentName, partition, now);
-            LOGGER.info("Creating CONSUMING segment {} for {} partition {}", newLLCSegmentName.getSegmentName(),
```
npawar: same for this

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -1151,34 +872,36 @@ protected IdealState ensureAllPartitionsConsuming(final TableConfig tableConfig,
...
-                tableNameWithType, partition);
-
-            // To begin with, set startOffset to the oldest available offset in the stream. Fix it to be the one we want,
-            // depending on what the prev segment had.
-            long startOffset = getPartitionOffset(streamConfig, OffsetCriteria.SMALLEST_OFFSET_CRITERIA, partition);
-            LOGGER.info("Found smallest offset {} for table {} for partition {}", startOffset, tableNameWithType,
```
npawar: ALso for this one. Can we somewhere log if we are adjusting the offset? (even without data loss)

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: This is logged in `createNewSegmentZKMetadata()`

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Logged in `createNewSegmentZKMetadata()`

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
Jackie-Jiang: Also logged in `createNewSegmentZKMetadata()` lol

PR #4702, opened 2019-10-12
Enhance Selection Comparator

Jackie-Jiang:
- Enhance the comparator for selection order-by queries to avoid doing row based switch
- Enhance the 'SELECT *' to not compile the expression
- Do not project order-by columns for 'LIMIT 0' (EmptySelection) case

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/query/selection/SelectionOperatorUtils.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/query/selection/SelectionOperatorUtils.java
@@ -23,6 +23,7 @@ import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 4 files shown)


pinot-core/src/main/java/org/apache/pinot/core/query/selection/SelectionOperatorUtils.java
```diff
@@ -568,4 +579,31 @@ private static Serializable getFormattedValue(Serializable value, DataSchema.Col
       queue.offer(value);
     }
   }
+
+  /**
+   * Helper class to compare rows.
+   */
```
siddharthteotia: If I understand it correctly, the purpose of valueIndices is to record the index of order by column(s) in a row. Since this comparator is fed the entire row, we need to know which column is the order by column. If that is correct, can you please add this info in javadoc for anyone using this to understand what exactly is value indices>

Jackie-Jiang: Good suggestion. Added javadoc

siddharthteotia:
LGTM. We can merge this once build passes.

PR #4706, opened 2019-10-14
Adding option to load a Pinot segment as post creation verification

fx19880617:
Per https://github.com/apache/incubator-pinot/issues/4705, this PR

1. adds a flag `-postCreationVerification` to turn on segment verification to ensure we could load segment metadata and column data file as well did a crc check. Segment verification process will copy data from `-outDir` to local then load it using `MMAP` mode.
2. add a `-retry` flag to re-generate segment for any exception thrown during segment creation or verification phase.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/segment/index/converter/SegmentV1V2ToV3FormatConverter.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/segment/index/converter/SegmentV1V2ToV3FormatConverter.java
@@ -200,9 +200,9 @@ public class SegmentV1V2ToV3FormatConverter implements SegmentFormatConverter {
       ColumnIndexType indexType)
       throws IOException {
     PinotDataBuffer oldBuffer = reader.getIndexFor(column, indexType);
-    PinotDataBuffer newBuffer = writer.newIndexFor(column, indexType, oldBuffer.size());
-    oldBuffer.copyTo(0, newBuffer, 0, oldBuffer.size());
-    newBuffer.flush();
+    long oldBufferSize = oldBuffer.size();
+    PinotDataBuffer newBuffer = writer.newIndexFor(column, indexType, oldBufferSize);
+    oldBuffer.copyTo(0, newBuffer, 0, oldBufferSize);
   }
 
   private void createMetadataFile(File currentDir, File v3Dir)
--- a/pinot-core/src/main/java/org/apache/pinot/core/segment/index/converter/SegmentV1V2ToV3FormatConverter.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/segment/index/converter/SegmentV1V2ToV3FormatConverter.java
@@ -200,8 +200,9 @@ public class SegmentV1V2ToV3FormatConverter implements SegmentFormatConverter {
       ColumnIndexType indexType)
       throws IOException {
     PinotDataBuffer oldBuffer = reader.getIndexFor(column, indexType);
-    PinotDataBuffer newDictBuffer = writer.newIndexFor(column, indexType, oldBuffer.size());
-    oldBuffer.copyTo(0, newDictBuffer, 0, oldBuffer.size());
+    PinotDataBuffer newBuffer = writer.newIndexFor(column, indexType, oldBuffer.size());
+    oldBuffer.copyTo(0, newBuffer, 0, oldBuffer.size());
+    newBuffer.flush();
   }
 
   private void createMetadataFile(File currentDir, File v3Dir)
--- a/pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
@@ -31,11 +31,15 @@ import java.util.List;
 import java.util.zip.Adler32;
 import java.util.zip.Checksum;
 import org.apache.pinot.core.segment.creator.impl.V1Constants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 @SuppressWarnings("Duplicates")
 public class CrcUtils {
+  private static final Logger LOGGER = LoggerFactory.getLogger(CrcUtils.class);
   private static final int BUFFER_SIZE = 65536;
+  private static final String CRC_FILE_EXTENSTION = ".crc";
 
   private final List<File> _files;
 
@@ -59,7 +63,8 @@ public class CrcUtils {
     Preconditions.checkNotNull(files);
     for (File file : files) {
       if (file.isFile()) {
-        if (!file.getName().equals(V1Constants.SEGMENT_CREATION_META)) {
+        // We should ignore both SEGMENT_CREATION_META and generated '.crc' files.
+        if (!file.getName().equals(V1Constants.SEGMENT_CREATION_META) && !file.getName().endsWith(CRC_FILE_EXTENSTION)) {
           normalFiles.add(file);
         }
       } else {
@@ -80,6 +85,7 @@ public class CrcUtils {
           checksum.update(buffer, 0, len);
         }
       }
+      LOGGER.info("Current file = {}, computed crc = {}", file, checksum.getValue());
     }
 
     return checksum.getValue();
--- a/pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
@@ -85,10 +85,10 @@ public class CrcUtils {
           checksum.update(buffer, 0, len);
         }
       }
-      LOGGER.info("Current file = {}, computed crc = {}", file, checksum.getValue());
     }
-
-    return checksum.getValue();
+    long crc = checksum.getValue();
+    LOGGER.info("Computed crc = {}", crc);
+    return crc;
   }
 
   public String computeMD5()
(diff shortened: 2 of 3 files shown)


pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -413,6 +445,32 @@ public void run() {
...
+    File localTempDir;
+    try {
+      localTempDir = new File("segmentVerification", org.apache.pinot.common.utils.FileUtils.getRandomFileName());
+      FileSystem.get(URI.create(indexDirStr), new Configuration())
+          .copyToLocalFile(new Path(indexDirStr), new Path(localTempDir.toString()));
+    } catch (IOException e) {
+      LOGGER.error("Failed to copy segment {} to local directory for verification.", indexDirStr, e);
+      return false;
+    }
+    try {
```
Jackie-Jiang: I don't think this will work. You need to untar the segment first.
I would recommend untar the segment, then use `ImmutableSegmentLoader.load()` to verify if you can load the segment.

fx19880617: `CreateSegment` script will not tar files.  `UploadSegment` script will do so. See: https://github.com/apache/incubator-pinot/blob/master/pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/UploadSegmentCommand.java#L128

pinot-core/src/main/java/org/apache/pinot/core/segment/index/converter/SegmentV1V2ToV3FormatConverter.java
```diff
@@ -200,8 +200,9 @@ private void readCopyBuffers(SegmentDirectory.Reader reader, SegmentDirectory.Wr
       ColumnIndexType indexType)
       throws IOException {
     PinotDataBuffer oldBuffer = reader.getIndexFor(column, indexType);
-    PinotDataBuffer newDictBuffer = writer.newIndexFor(column, indexType, oldBuffer.size());
-    oldBuffer.copyTo(0, newDictBuffer, 0, oldBuffer.size());
+    PinotDataBuffer newBuffer = writer.newIndexFor(column, indexType, oldBuffer.size());
```
Jackie-Jiang: Cache the old buffer size

pinot-core/src/main/java/org/apache/pinot/core/segment/index/converter/SegmentV1V2ToV3FormatConverter.java
```diff
@@ -200,8 +200,9 @@ private void readCopyBuffers(SegmentDirectory.Reader reader, SegmentDirectory.Wr
...
+    oldBuffer.copyTo(0, newBuffer, 0, oldBuffer.size());
+    newBuffer.flush();
```
Jackie-Jiang: We should flush the buffer in SegmentLocalFSDirectory.save() which is empty right now.

pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
```diff
@@ -59,7 +63,8 @@ private static void getAllNormalFiles(File dir, List<File> normalFiles) {
     Preconditions.checkNotNull(files);
     for (File file : files) {
       if (file.isFile()) {
-        if (!file.getName().equals(V1Constants.SEGMENT_CREATION_META)) {
+        // We should ignore both SEGMENT_CREATION_META and generated '.crc' files.
+        if (!file.getName().equals(V1Constants.SEGMENT_CREATION_META) && !file.getName().endsWith(CRC_FILE_EXTENSTION)) {
```
Jackie-Jiang: Where is the `.crc` file generated?

pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
```diff
@@ -80,6 +85,7 @@ public long computeCrc()
           checksum.update(buffer, 0, len);
         }
       }
+      LOGGER.info("Current file = {}, computed crc = {}", file, checksum.getValue());
```
Jackie-Jiang: Can we move this out of the loop so that we don't flood the log for v1 segment? We can log the list of files and the final crc.
Maybe log this for MD5 as well?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -413,6 +447,32 @@ public void run() {
     return executor.awaitTermination(1, TimeUnit.HOURS);
   }
 
+  private boolean verifySegment(File indexDir) {
+    File localTempDir = null;
+    try {
+      try {
+        localTempDir = new File("segmentVerification", org.apache.pinot.common.utils.FileUtils.getRandomFileName());
```
Jackie-Jiang: Use `FileUtils.getTempDirectory()` as the parent temp dir?

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -413,6 +447,32 @@ public void run() {
...
+        FileSystem.get(URI.create(indexDir.toString()), new Configuration())
+            .copyToLocalFile(new Path(indexDir.toString()), new Path(localTempDir.toString()));
+      } catch (IOException e) {
+        LOGGER.error("Failed to copy segment {} to local directory for verification.", indexDir, e);
+        return false;
+      }
+      try {
+        ImmutableSegment segment = ImmutableSegmentLoader.load(localTempDir, ReadMode.mmap);
+        LOGGER.info("Successfully loaded Pinot segment {} (size: {} Bytes).", segment.getSegmentName(),
+            segment.getSegmentSizeBytes());
```
Jackie-Jiang: Destroy the segment

pinot-core/src/main/java/org/apache/pinot/core/segment/index/converter/SegmentV1V2ToV3FormatConverter.java
fx19880617: correct, removed this line.

pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
fx19880617: done

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
fx19880617: done

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
fx19880617: done

pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
fx19880617: I think I've seen certain data copy operations may leave .crc files in the directory, which could cause the crc check failed.

Jackie-Jiang: Are you referring to the behavior of some FileSystem? If so, can you please add more comments about that?

pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
```diff
@@ -81,8 +86,9 @@ public long computeCrc()
         }
       }
     }
-
-    return checksum.getValue();
+    long crc = checksum.getValue();
+    LOGGER.info("Computed crc = {}", crc);
```
Jackie-Jiang: You might want to include all the files used to generate the crc for debugging purpose? Only logging crc does not provide a lot of value IMO. Same for MD5

pinot-tools/src/main/java/org/apache/pinot/tools/admin/command/CreateSegmentCommand.java
```diff
@@ -413,6 +447,32 @@ public void run() {
     return executor.awaitTermination(1, TimeUnit.HOURS);
   }
 
+  private boolean verifySegment(File indexDir) {
+    File localTempDir = FileUtils.getTempDirectory();
```
Jackie-Jiang: You cannot directly use the tempDir and delete it after the verification. Multiple segments will end up modifying and deleting the same tempDir.
You can use `
    new File(FileUtils.getTempDirectory(), org.apache.pinot.common.utils.FileUtils.getRandomFileName())` as the tempDir

pinot-core/src/main/java/org/apache/pinot/core/util/CrcUtils.java
```diff
@@ -81,8 +88,9 @@ public long computeCrc()
         }
       }
     }
-
-    return checksum.getValue();
+    long crc = checksum.getValue();
+    LOGGER.info("Computed crc = {}, based on files {}", crc, Arrays.toString(_files.toArray()));
```
Jackie-Jiang: `LOGGER.info("Computed crc = {}, based on files {}", crc, _files);`
Same for MD5

PR #4713, opened 2019-10-16
Refactoring realtime segment committer

jenniferdai:
* Pave the path for different ways to commit realtime segments, including committing realtime segments in ways that do not depend on accessible deep storage
* See https://cwiki.apache.org/confluence/display/PINOT/By-passing+deep-store+requirement+for+Realtime+segment+completion


pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SegmentCommitter.java
```diff
@@ -0,0 +1,29 @@
...
+package org.apache.pinot.core.data.manager.realtime;
+
+import org.apache.pinot.common.protocols.SegmentCompletionProtocol;
+
+
+/**
+ * For committing realtime segments.
+ */
+public interface SegmentCommitter {
+  SegmentCompletionProtocol.Response commit(long currentOffset, int numRows, LLRealtimeSegmentDataManager.SegmentBuildDescriptor segmentBuildDescriptor);
```
mcvsubbu: pls add javadocs here

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -837,22 +779,24 @@ protected boolean commitSegment(SegmentCompletionProtocol.Response response) {
...
         .withWaitTimeMillis(_segmentBuildDescriptor.getWaitTimeMillis());
     if (_isOffHeap) {
       params.withMemoryUsedBytes(_memoryManager.getTotalAllocatedBytes());
     }
-    SegmentCompletionProtocol.Response response = _protocolHandler.segmentCommit(params, segmentTarFile);
-    if (!response.getStatus().equals(SegmentCompletionProtocol.ControllerResponseStatus.COMMIT_SUCCESS)) {
-      segmentLogger.warn("Commit failed  with response {}", response.toJsonString());
+
+    if (isSplitCommit) {
+      SplitSegmentCommitter splitSegmentCommitter = new SplitSegmentCommitter(params, _protocolHandler, response, _indexLoadingConfig.isEnableSplitCommitEndWithMetadata(), segmentLogger);
```
mcvsubbu: Just pass the _indexLoadingConfig.
Use a factory to create the committer class. 
You can construct the factory up front, with constriuctor(segmentLogger, _protocolHandler, and indexLoadingConfig)
A createCommitter() method on the factory that can take (params, response) as arguments and create the right kind of committer.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -837,22 +781,28 @@ protected boolean commitSegment(SegmentCompletionProtocol.Response response) {
...
       params.withMemoryUsedBytes(_memoryManager.getTotalAllocatedBytes());
     }
-    SegmentCompletionProtocol.Response response = _protocolHandler.segmentCommit(params, segmentTarFile);
-    if (!response.getStatus().equals(SegmentCompletionProtocol.ControllerResponseStatus.COMMIT_SUCCESS)) {
-      segmentLogger.warn("Commit failed  with response {}", response.toJsonString());
+
+    Configuration config = new PropertiesConfiguration();
+
+    if (response.isSplitCommit() && _indexLoadingConfig.isEnableSplitCommit()) {
+      config.setProperty(CommonConstants.Segment.Realtime.COMMITTER_CLASS, SplitSegmentCommitter.class.getName());
```
mcvsubbu: Like we discussed, I think Config is a bit too much of overhead here. We really cannot avoid the if statement before creating a committer (since it still depends on the controller's response to the server's segmentConsumed() call), we can call different methods in the factory (like createDefaultCommitter() vs createSplitCommitter()). The factory itself can be constructed in the ctor of LLRealtimeSegmentDataManager with parameters that are needed for both committers (indexLoadingConfig, segmentLogger, and perhaps even protocolHandler). thanks

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -839,22 +783,27 @@ protected boolean commitSegment(SegmentCompletionProtocol.Response response) {
...
         .withSegmentSizeBytes(_segmentBuildDescriptor.getSegmentSizeBytes())
         .withWaitTimeMillis(_segmentBuildDescriptor.getWaitTimeMillis());
     if (_isOffHeap) {
       params.withMemoryUsedBytes(_memoryManager.getTotalAllocatedBytes());
     }
-    SegmentCompletionProtocol.Response response = _protocolHandler.segmentCommit(params, segmentTarFile);
-    if (!response.getStatus().equals(SegmentCompletionProtocol.ControllerResponseStatus.COMMIT_SUCCESS)) {
-      segmentLogger.warn("Commit failed  with response {}", response.toJsonString());
+
+    Configuration config = new PropertiesConfiguration();
```
mcvsubbu: do you still need config?

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -627,7 +630,7 @@ private CompletionMode getSegmentCompletionMode() {
     return CompletionMode.DEFAULT;
   }
 
-  private File makeSegmentDirPath() {
+  public File makeSegmentDirPath() {
```
mcvsubbu: why this change?

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
```diff
@@ -0,0 +1,227 @@
...
+    startZk();
+    startController();
+    addFakeBrokerInstancesToAutoJoinHelixCluster(1, true);
+    addFakeServerInstancesToAutoJoinHelixCluster(1, true);
+
+    ControllerLeaderLocator.create(_helixManager);
+
+    // Start Kafka
+    startKafka();
+    List<File> avroFiles = unpackAvroData(_tempDir);
```
mcvsubbu: checkin 2 or 3  segments with 1 or 2 rows in them that is already built. That will save run time. Maybe we already have some checked in segments, check first.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
```diff
@@ -0,0 +1,227 @@
...
+  @Override
+  protected boolean useLlc() {
+    return true;
+  }
+
+  @Test
+  public void testDefaultCommitter() {
+    IndexLoadingConfig indexLoadingConfig = mock(IndexLoadingConfig.class);
+    ServerMetrics serverMetrics = mock(ServerMetrics.class);
+    doNothing().when(serverMetrics).addMeteredGlobalValue(ServerMeter.LLC_CONTROLLER_RESPONSE_COMMIT_SUCCESS, 1);
```
mcvsubbu: you can create new server metrics class.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
```diff
@@ -0,0 +1,227 @@
...
+    ServerSegmentCompletionProtocolHandler
+        protocolHandler = new ServerSegmentCompletionProtocolHandler(serverMetrics, getTableName());
+    SegmentCompletionProtocol.Request.Params params = mock(SegmentCompletionProtocol.Request.Params.class);
```
mcvsubbu: why are you mocking these classes? Just set them to whatever you want.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
```diff
@@ -0,0 +1,227 @@
...
+    SegmentCompletionProtocol.Response prevResponse = mock(SegmentCompletionProtocol.Response.class);
+    LLRealtimeSegmentDataManager.SegmentBuildDescriptor segmentBuildDescriptor = mock(
+        LLRealtimeSegmentDataManager.SegmentBuildDescriptor.class);
+
+    RealtimeSegmentZKMetadata metadata = _helixResourceManager.getRealtimeSegmentMetadata(getTableName()).get(0);
+
+    when(params.getSegmentName()).thenReturn(metadata.getSegmentName());
+    when(params.getInstanceId()).thenReturn("Server_localhost_0");
+    when(params.getOffset()).thenReturn(END_OFFSET);
+    when(segmentBuildDescriptor.getSegmentTarFilePath()).thenReturn(_realtimeSegmentUntarred + ".tar.gz");
```
mcvsubbu: point this to a real segment

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
```diff
@@ -0,0 +1,227 @@
...
+    when(segmentBuildDescriptor.getBuildTimeMillis()).thenReturn(0L);
+    Map<String, File> metadataPaths = new HashMap();
+    metadataPaths.put(metadata.getSegmentName(), new File(untarredSegmentFile, "metadata.properties"));
+
+    when(segmentBuildDescriptor.getMetadataFiles()).thenReturn(metadataPaths);
+    when(segmentBuildDescriptor.getOffset()).thenReturn(END_OFFSET);
+    when(segmentBuildDescriptor.getSegmentSizeBytes()).thenReturn(0L);
+    when(segmentBuildDescriptor.getWaitTimeMillis()).thenReturn(0L);
+
+    SegmentCompletionManager segmentCompletionManager = _controllerStarter.getSegmentCompletionManager();
```
mcvsubbu: why are you doing this? you call the commit code, after callinug  the segmentConsumed API like I demonstrated with curl command

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
jenniferdai: It points to a real segment currently.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/DefaultSegmentCommitter.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.pinot.server.realtime.ServerSegmentCompletionProtocolHandler;
+import org.slf4j.Logger;
+
+
+/**
+ * Sends segmentCommit() to the controller.
+ * If that succeeds, swap in-memory segment with the one built.
+ */
+public class DefaultSegmentCommitter implements SegmentCommitter {
+  private SegmentCompletionProtocol.Request.Params _params;
```
mcvsubbu: final for all three

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -260,6 +261,7 @@ public void deleteSegmentFile() {
   private final Semaphore _segBuildSemaphore;
   private final boolean _isOffHeap;
   private final boolean _nullHandlingEnabled;
+  private SegmentCommitterFactory _segmentCommitterFactory;
```
mcvsubbu: final

pinot-controller/src/test/java/org/apache/pinot/controller/helix/ControllerTest.java
```diff
@@ -76,12 +76,9 @@
 import org.slf4j.LoggerFactory;
 import org.testng.Assert;
 
-import static org.apache.pinot.common.utils.CommonConstants.Helix.Instance.ADMIN_PORT_KEY;
-import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_ENABLED_KEY;
-import static org.apache.pinot.common.utils.CommonConstants.Helix.LEAD_CONTROLLER_RESOURCE_NAME;
-import static org.apache.pinot.common.utils.CommonConstants.Helix.UNTAGGED_BROKER_INSTANCE;
-import static org.apache.pinot.common.utils.CommonConstants.Helix.UNTAGGED_SERVER_INSTANCE;
-import static org.apache.pinot.common.utils.CommonConstants.Server.DEFAULT_ADMIN_API_PORT;
+import static org.apache.pinot.common.utils.CommonConstants.Helix.Instance.*;
```
mcvsubbu: Not sure why you made this change. We try not to import * but have individual imports as needed. There seems to be no other change to this file.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -1208,6 +1155,7 @@ public LLRealtimeSegmentDataManager(RealtimeSegmentZKMetadata segmentZKMetadata,
     segmentLogger
         .info("Starting consumption on realtime consuming segment {} maxRowCount {} maxEndTime {}", _segmentName,
             _segmentMaxRowCount, new DateTime(_consumeEndTime, DateTimeZone.UTC).toString());
+    _segmentCommitterFactory = new SegmentCommitterFactory(segmentLogger, _indexLoadingConfig, _protocolHandler);
```
mcvsubbu: Please move this line above the log. 

We would like the start() be right below the log (we claim to start consumption in the log, right?)

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SegmentCommitter.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.apache.pinot.common.protocols.SegmentCompletionProtocol;
+
+
+/**
+ * For committing realtime segments.
+ */
+public interface SegmentCommitter {
+  /**
+   * Commits a realtime segment
```
mcvsubbu:
```suggestion
   * Commits a realtime segment to persistent store
```

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SegmentCommitter.java
```diff
@@ -0,0 +1,36 @@
...
+   * @param currentOffset current offset
```
mcvsubbu:
```suggestion
   * @param currentOffset current offset in the stream
```

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SegmentCommitter.java
```diff
@@ -0,0 +1,36 @@
...
+   * @param numRows num rows committing
```
mcvsubbu:
```suggestion
   * @param numRows num rows consumed in the segment being committed
```

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SegmentCommitterFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.pinot.server.realtime.ServerSegmentCompletionProtocolHandler;
+import org.slf4j.Logger;
+
+
+/**
+ * Factory for the SegmentCommitter interface
+ */
+public class SegmentCommitterFactory {
+  private static Logger LOGGER;
+  private IndexLoadingConfig _indexLoadingConfig;
```
mcvsubbu: final please

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SplitSegmentCommitter.java
```diff
@@ -0,0 +1,87 @@
...
+    final File segmentTarFile = new File(segmentBuildDescriptor.getSegmentTarFilePath());
+
+    SegmentCompletionProtocol.Response segmentCommitStartResponse = _protocolHandler.segmentCommitStart(_params);
+    if (!segmentCommitStartResponse.getStatus()
+        .equals(SegmentCompletionProtocol.ControllerResponseStatus.COMMIT_CONTINUE)) {
+      _segmentLogger.warn("CommitStart failed  with response {}", segmentCommitStartResponse.toJsonString());
+      return SegmentCompletionProtocol.RESP_FAILED;
+    }
+
+    SegmentCompletionProtocol.Request.Params params = new SegmentCompletionProtocol.Request.Params();
```
mcvsubbu: why not reuse _params here?

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SplitSegmentCommitter.java
```diff
@@ -0,0 +1,87 @@
...
+    params.withInstanceId(_params.getInstanceId()).withOffset(currentOffset).withSegmentName(_params.getSegmentName());
+    SegmentCompletionProtocol.Response segmentCommitUploadResponse =
+        _protocolHandler.segmentCommitUpload(params, segmentTarFile, _prevResponse.getControllerVipUrl());
+    if (!segmentCommitUploadResponse.getStatus()
+        .equals(SegmentCompletionProtocol.ControllerResponseStatus.UPLOAD_SUCCESS)) {
+      _segmentLogger.warn("Segment upload failed with response {}", segmentCommitUploadResponse.toJsonString());
+      return SegmentCompletionProtocol.RESP_FAILED;
+    }
+
+    _params.withOffset(currentOffset).withSegmentLocation(segmentCommitUploadResponse.getSegmentLocation())
```
mcvsubbu: curent offset and numrows consumed should already be set, and is not going to change between first and this step. We only need to add segmentLocation

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
```diff
@@ -0,0 +1,204 @@
...
+    FileUtils.copyDirectory(_indexDir, _realtimeSegmentUntarred);
+
+    TarGzCompressionUtils.createTarGzOfDirectory(_realtimeSegmentUntarred.getAbsolutePath());
+
+    when(segmentBuildDescriptor.getSegmentTarFilePath()).thenReturn(_realtimeSegmentUntarred + TARGZ_SUFFIX);
+    when(segmentBuildDescriptor.getBuildTimeMillis()).thenReturn(0L);
+    Map<String, File> metadataPaths = new HashMap();
+    metadataPaths.put(metadata.getSegmentName(), new File(_indexDir, METADATA_FILE_NAME));
+
+    when(segmentBuildDescriptor.getMetadataFiles()).thenReturn(metadataPaths);
```
mcvsubbu: since you are testing default commit in this test, you can omit the getMetadtaFiles mock

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
```diff
@@ -0,0 +1,204 @@
...
+    sendGetRequest("http://localhost:" + DEFAULT_CONTROLLER_PORT +
+            "/segmentConsumed?instance=" + instanceId + "&name=" + segmentName + "&offset=" + END_OFFSET);
+
+    SegmentCommitterFactory
+        segmentCommitterFactory = new SegmentCommitterFactory(LOGGER, indexLoadingConfig, protocolHandler);
+    SegmentCommitter segmentCommitter = segmentCommitterFactory.createDefaultSegmentCommitter(params);
+    segmentCommitter.commit(END_OFFSET, 3, segmentBuildDescriptor);
+  }
+
+  public void buildSegment()
```
mcvsubbu: can we create a segment and check-in? How long does it take for the test to run?

jenniferdai: 8 seconds, my segment is built with only one row, so I think it should be pretty fast - thoughts?

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SplitSegmentCommitter.java
jenniferdai: The old code created a new params object, I assume to make sure that state from the other params object wasn't maintained - I kept this behavior.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SplitSegmentCommitter.java
jenniferdai: Because I wanted to maintain the integrity of the previous code, I created a new Params object to wipe state - therefore, for the params object passed in I will keep this for now.

pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/DefaultCommitterRealtimeIntegrationTest.java
mcvsubbu: 8 seconds seems reasonable. Let us leave it like that then.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/LLRealtimeSegmentDataManager.java
```diff
@@ -761,74 +763,15 @@ protected SegmentBuildDescriptor buildSegmentInternal(boolean forCommit) {
...
-      // Send segmentStart, segmentUpload, & segmentCommitEnd to the controller
-      // if that succeeds, swap in-memory segment with the one built.
-      returnedResponse = doSplitCommit(response);
-    } else {
-      // Send segmentCommit() to the controller
-      // if that succeeds, swap in-memory segment with the one built.
-      returnedResponse = postSegmentCommitMsg();
-    }
+
+    returnedResponse = commit(response);
```
mcvsubbu: Please change this to declare and assign returnedResponse in one line.  Can you rename this var to commitResponse?

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SplitSegmentCommitter.java
```diff
@@ -0,0 +1,84 @@
...
+
+    SegmentCompletionProtocol.Response segmentCommitUploadResponse =
+        _protocolHandler.segmentCommitUpload(_params, segmentTarFile, _controllerVipUrl);
+    if (!segmentCommitUploadResponse.getStatus()
+        .equals(SegmentCompletionProtocol.ControllerResponseStatus.UPLOAD_SUCCESS)) {
+      _segmentLogger.warn("Segment upload failed with response {}", segmentCommitUploadResponse.toJsonString());
+      return SegmentCompletionProtocol.RESP_FAILED;
+    }
+
+    _params.withSegmentLocation(segmentCommitUploadResponse.getSegmentLocation());
```
mcvsubbu: Use the new constructor here instead of _params, so that you don't modify the object in this class.

pinot-core/src/test/java/org/apache/pinot/core/data/readers/PinotSegmentUtil.java
```diff
@@ -80,7 +79,8 @@ static boolean compareMultiValueColumn(Object value1, Object value2) {
     return value1Set.containsAll(value2Set);
   }
 
-  static File createSegment(Schema schema, String segmentName, String segmentOutputDir, RecordReader recordReader)
+  @VisibleForTesting
```
mcvsubbu: this class is already in test, so you don't need the annotation here

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/SplitSegmentCommitter.java
mcvsubbu: ah, my bad. I did not notice that you were already constructing a new one. thanks.

PR #4716, opened 2019-10-17
Package pinot-parquet and pinot-orc module into pinot-hadoop shaded jar

fx19880617:
So that users could just drop shaded jar and run it in hadoop environment to read both parquet and orc files.

Sample conf in job.property
```
record.reader.path=org.apache.pinot.parquet.data.readers.ParquetRecordReader
```

### DIFF
--- a/pinot-hadoop/pom.xml
+++ b/pinot-hadoop/pom.xml
@@ -60,8 +60,8 @@
                   -->
                   <relocations>
                     <relocation>
-                      <pattern>com.google.common.base.Preconditions</pattern>
-                      <shadedPattern>shaded.com.google.common.base.Preconditions</shadedPattern>
+                      <pattern>com.google.common.base</pattern>
+                      <shadedPattern>shaded.com.google.common.base</shadedPattern>
                     </relocation>
                     <relocation>
                       <pattern>com.fasterxml.jackson</pattern>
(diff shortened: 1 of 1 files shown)


pinot-hadoop/pom.xml
```diff
@@ -103,6 +103,16 @@
       <version>${project.version}</version>
       <classifier>shaded</classifier>
     </dependency>
+    <dependency>
+      <groupId>org.apache.pinot</groupId>
```
Jackie-Jiang: You probably need to skip pinot-core for these 2 dependencies?

pinot-hadoop/pom.xml
```diff
@@ -103,6 +103,28 @@
...
+      <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.pinot</groupId>
+          <artifactId>pinot-core</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.pinot</groupId>
```
jenniferdai: are you trying to shade the orc and parquet dependencies? this pr only shades the google dependency, right?

fx19880617: This pr aims to shade orc/parquet, so user could use this jar to create segment from parquent/orc file. Otherwise they need to drop pinot-parquet/pinot-orc jar into their classpath or do their own shading with pinot-parquet/pinot-orc to generate a fat jar.

PR #4718, opened 2019-10-17
[TE] support row count metric for Pinot

jihaozh:
This PR adds the support for the ROW_COUNT metric for Pinot. It requires two changes:
- Don't remove the ROW_COUNT metric during Pinot metric auto-onboarding.
- Remove the special condition for the "COUNT" aggregation function when parsing the Pinot response.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/AutoOnboardPinotMetadataSource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/AutoOnboardPinotMetadataSource.java
@@ -68,6 +68,10 @@ public class AutoOnboardPinotMetadataSource extends AutoOnboard {
 
   private static final Set<String> DIMENSION_SUFFIX_BLACKLIST = new HashSet<>(Arrays.asList("_topk", "_approximate", "_tDigest"));
 
+  /**
+   * Use "ROW_COUNT" as the special token for the count(*) metric for a pinot table
+   */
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/AutoOnboardPinotMetadataSource.java
```diff
@@ -154,7 +158,7 @@ private boolean shouldRemoveDataset(DatasetConfigDTO datasetConfigDTO, Set<Strin
       List<MetricConfigDTO> metrics = metricDAO.findByDataset(datasetConfigDTO.getDataset());
       int metricCount = metrics.size();
       for (MetricConfigDTO metric : metrics) {
-        if (!metric.isDerived()) {
+        if (!metric.isDerived() && !metric.getName().equals(ROW_COUNT)) {
```
xiaohui-sun: Are you planning to explicitly add a column called ROW_COUNT for all datasets?

jihaozh: No, won't add a column. In ThirdEye, we will map "ROW_COUNT" to a count(*) metric. 
This line is to ensure that the metric won't be removed automatically.

xiaohui-sun: Thanks for the clarification.

PR #4722, opened 2019-10-18
Fix segmentMetadata for MutableSegment

haibow:
The current way of initializing `_segmentMetadata` would make its type to be `MutableSegmentImpl` instead of `SegmentMetadataImpl`.

Fixing it and adding a test.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
@@ -115,29 +115,7 @@ public class MutableSegmentImpl implements MutableSegment {
     _segmentName = config.getSegmentName();
     _schema = config.getSchema();
     _capacity = config.getCapacity();
-    _segmentMetadata = new SegmentMetadataImpl(config.getRealtimeSegmentZKMetadata(), _schema) {
-      @Override
-      public int getTotalDocs() {
-        return _numDocsIndexed;
-      }
-
-      @Override
-      public int getTotalRawDocs() {
-        // In realtime total docs and total raw docs are the same currently.
-        return _numDocsIndexed;
-      }
-
-      @Override
-      public long getLastIndexedTimestamp() {
-        return _lastIndexedTimeMs;
-      }
-
-      @Override
-      public long getLatestIngestionTimestamp() {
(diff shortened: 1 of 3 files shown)


Jackie-Jiang:
Why do you need the class to be the same? As long as you can cast it to `SegmentMetadataImpl` (a subclass of `SegmentMetadataImpl`), you should be able to call all its API.

We should not allow setters for SegmentMetadata.

haibow:
> Why do you need the class to be the same? As long as you can cast it to `SegmentMetadataImpl` (a subclass of `SegmentMetadataImpl`), you should be able to call all its API.
> 
> We should not allow setters for SegmentMetadata.

It seems ok casting to `SegmentMetadataImpl` where I'm using it - just looks weird SegmentMetadata being a MutableSegment class, and need to cast it wherever it's used to avoid potential issues.

Is it required to have no setters for SegmentMetadata? Overriding the getter methods kind of shows there is such need for mutable segments.

Jackie-Jiang:
> > Why do you need the class to be the same? As long as you can cast it to `SegmentMetadataImpl` (a subclass of `SegmentMetadataImpl`), you should be able to call all its API.
> > We should not allow setters for SegmentMetadata.
> 
> It seems ok casting to `SegmentMetadataImpl` where I'm using it - just looks weird SegmentMetadata being a MutableSegment class, and need to cast it wherever it's used to avoid potential issues.
> 
> Is it required to have no setters for SegmentMetadata? Overriding the getter methods kind of shows there is such need for mutable segments.

The `SegmentMetadata` interface should not have any setters (segment metadata represents the state of the segment, and allow user to set it seems weird to me).
I agree it looks weird to have `SegmentMetadataImpl` inside MutableSegment. The reason why we override `SegmentMetadataImpl` instead of implementing `SegmentMetadata` (which is cleaner) for MutableSegment is because some historical package issue, where `SegmentMetadata` interface cannot include `ColumnMetadata`, and this is also the reason why we have to cast the `SegmentMetadata` to `SegmentMetadataImpl` in order to read `ColumnMetadata`.
Long term wise, we want to re-organize the package so that `SegmentMetadata` can include `ColumnMetadata`, and MutableSegment should implement its own segment metadata.

haibow:
@Jackie-Jiang thanks for the input! Agree a more thorough refactor is more ideal. Since this is not a blocker for me, I will abandon this fix.

PR #4723, opened 2019-10-18
[TE] put default settings for subscription YAML

xiaohui-sun:
1. Set the default type as "DEFAULT_ALERTER_PIPELINE".
2. Remove optional parameters from template.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
@@ -658,9 +658,6 @@ public class YamlResource {
       }
     }
 
-    if (!properties.containsKey(PROP_CLASS_NAME)) {
-      properties.put(PROP_CLASS_NAME, "DEFAULT_ALERTER_PIPELINE");
-    }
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/yaml/YamlResource.java
```diff
@@ -657,6 +657,10 @@ void updateSubscriptionGroup(ThirdEyePrincipal user, long oldAlertConfigID, Stri
         }
       }
     }
+
+    if (!properties.containsKey(PROP_CLASS_NAME)) {
```
akshayrai: The parsing of "DEFAULT_ALERTER_PIPELINE" type to the appropriate class happens in the translator. So you need to move this inside the translator class.

```
private Map<String, Object> buildAlerterProperties(Map<String, Object> alertYamlConfigs) {
    // By default we assume the subscription pipeline follows DEFAULT_ALERTER_PIPELINE
    alertYamlConfigs.putIfAbsent(PROP_TYPE, "DEFAULT_ALERTER_PIPELINE");  // ADD here
   ...
} 
```

xiaohui-sun: Makes sense! Moved the code.

PR #4727, opened 2019-10-21
Enable full stack trace when there is an exception during tests

mcvsubbu:
By default maven does not print the full stack when exceptions are encountered during tests.
Enabled surefire plugin to print the full stack for easy debugging.  In some cases we may not
have access to the full surefire report to get the stack trace.


mcvsubbu:
Checked one test that encounters quite some exceptions (FlakyConsumerRealtimeClusterIntegrationTests) and the logs increase by about 3% since all exceptions are logged.

PR #4728, opened 2019-10-21
Optimizations for  IndexedTable resize

npawar:
2 optimizations introduced:
1) Convert the Record to an IntermediateRecord before sorting/putting into PQ. IntermediateRecord will contain only the order by columns, in the right sequence.
2) When converting to IntermediateRecord, extract final result if intermediate result is non-comparable

Also made the logic for Simple and Concurrent table very identical (minus the Concurrent data structures used in ConcurrentIndexedTable)

Removed the OrderByUtils in favor of IndexedTableResizer.


pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,236 @@
...
+        AggregationFunction aggregationFunction = orderByInfo.getAggregationFunction();
+        if (!aggregationFunction.isIntermediateResultComparable()) {
+          comparable = aggregationFunction.extractFinalResult(aggregationColumn);
+        } else {
+          comparable = (Comparable) aggregationColumn;
+        }
+      }
+      intermediateRecordValues[i] = comparable;
+    }
+    return new IntermediateRecord(record.getKey(), intermediateRecordValues);
```
mayankshriv: Do we need to create this additional object for all cases (for example, when we don't need to extract final results)?

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,236 @@
...
+        recordsMap.remove(evictRecord.getKey());
+      }
+    }
+  }
+
+  /**
+   * Sort the records in the recordsMap according to order by
+   */
+  @Override
+  List<Record> sortRecordsMap(Map<Key, Record> recordsMap) {
```
mayankshriv: Will there be a case where the caller will call both resize and sort? If so, in that case there will be an additional sorting that can be avoided.

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,260 @@
...
+  @VisibleForTesting
+  IntermediateRecord getIntermediateRecord(Record record) {
+    Comparable[] intermediateRecordValues = new Comparable[_numOrderBy];
+
+    Object[] keyColumns = record.getKey().getColumns();
+    Object[] aggregationColumns = record.getValues();
+    for (int i = 0; i < _numOrderBy; i++) {
+      OrderByInfo orderByInfo = _orderByInfos.get(i);
+      Comparable comparable;
+      if (orderByInfo.getOrderByType().equals(OrderByType.KEY_COLUMN)) {
```
Jackie-Jiang: To avoid these if checks, we should add an inner class to extract comparable from the Record. E.g. have a KeyExtractor and an AggregationResultExtractor

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,260 @@
...
+      String column = selectionSort.getColumn();
+
+      if (keyIndexMap.containsKey(column)) {
+        int index = keyIndexMap.get(column);
+        _orderByInfos.add(createOrderByInfoForKeyColumn(index));
+      } else if (aggregationColumnToIndex.containsKey(column)) {
+        int index = aggregationColumnToIndex.get(column);
+        _orderByInfos.add(createOrderByInfoForAggregationColumn(index, aggregationColumnToInfo.get(column)));
+      } else {
+        throw new UnsupportedOperationException("Could not find column " + column + " in data schema");
```
Jackie-Jiang: This is IllegalStateException

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,260 @@
...
+      }
+
+      comparators[i] = ComparableComparator.getInstance();
```
Jackie-Jiang: `Comparator.naturalOrder()`

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,260 @@
...
+      if (!selectionSort.isIsAsc()) {
+        comparators[i] = comparators[i].reversed();
+      }
+    }
+
+    _intermediateRecordComparator = (o1, o2) -> {
+
+      for (int i = 0; i < _numOrderBy; i++) {
+        int result = comparators[i].compare(o1.getValues()[i], o2.getValues()[i]);
```
Jackie-Jiang: Cache o1.getValues() and o2.getValues() (or you can directly access the array if you make the class inner class)

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,260 @@
...
+   */
+  private enum OrderByType {
+    KEY_COLUMN, // order by on a key column (e.g. group by or distinct)
+    AGGREGATION_COLUMN // order by on aggregation (e.g. aggregations in aggregation only or group by)
+  }
+
+  /**
+   * Defines the information needed for one SelectionSort of an order by clause
+   */
+  private class OrderByInfo {
```
Jackie-Jiang: This class can be replaced with the OrderByValueExtractor to avoid condition checks

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
Jackie-Jiang: +1. In the final step, you should use PriorityQueue to resize and sort

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,260 @@
...
+          }
+        }
+      }
+
+      if (numRecordsToEvict < trimToSize) { // PQ contains records to evict
+        for (IntermediateRecord evictRecord : priorityQueue) {
+          recordsMap.remove(evictRecord.getKey());
+        }
+      } else { // PQ contains records to retain
+        Set<Key> keysToRetain = new HashSet<>(priorityQueue.size());
```
Jackie-Jiang: This will almost always cause a resize. You can use ObjectOpenHashSet to avoid this, or take the load factor into the initial capacity

pinot-core/src/main/java/org/apache/pinot/core/data/table/SimpleIndexedTable.java
```diff
@@ -66,82 +58,65 @@ public void init(@Nonnull DataSchema dataSchema, List<AggregationInfo> aggregati
...
 
-    Integer index = _lookupTable.get(keys);
-    if (_noMoreNewRecords) { // only update existing records
-      if (index != null) {
-        Record existingRecord = _records.get(index);
+    if (_noMoreNewRecords) { // allow only existing record updates
+      _lookupMap.computeIfPresent(key, (k, v) -> {
         for (int i = 0; i < _numAggregations; i++) {
-          existingRecord.getValues()[i] = _aggregationFunctions.get(i).merge(existingRecord.getValues()[i], newRecord.getValues()[i]);
+          v.getValues()[i] = _aggregationFunctions[i].merge(v.getValues()[i], newRecord.getValues()[i]);
```
Jackie-Jiang: Cache v.getValues() and newRecord.getValues()

pinot-core/src/main/java/org/apache/pinot/core/data/table/SimpleIndexedTable.java
```diff
@@ -66,82 +58,65 @@ public void init(@Nonnull DataSchema dataSchema, List<AggregationInfo> aggregati
...
-      if (size() >= _bufferedCapacity) {
-        if (_isOrderBy) { // capacity reached, order and resize
-          sortAndResize(_maxCapacity);
-        } else { // capacity reached, but no order by. Allow no more records
+      _lookupMap.compute(key, (k, v) -> {
+        if (v == null) {
+          return newRecord;
+        } else {
+          for (int i = 0; i < _numAggregations; i++) {
+            v.getValues()[i] = _aggregationFunctions[i].merge(v.getValues()[i], newRecord.getValues()[i]);
```
Jackie-Jiang: Cache v.getValues() and newRecord.getValues()

pinot-core/src/main/java/org/apache/pinot/core/data/table/RandomIndexedTableResizer.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+
+/**
+ * Helper class for trimming and sorting records in the IndexedTable
+ */
+class RandomIndexedTableResizer extends IndexedTableResizer {
```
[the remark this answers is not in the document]
npawar: Removed the other one, renamed this to just `IndexedTableResizer`

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -98,7 +78,7 @@ public boolean upsert(@Nonnull Record newRecord) {
     if (_noMoreNewRecords.get()) { // allow only existing record updates
       _lookupMap.computeIfPresent(key, (k, v) -> {
         for (int i = 0; i < _numAggregations; i++) {
-          v.getValues()[i] = _aggregationFunctions.get(i).merge(v.getValues()[i], newRecord.getValues()[i]);
+          v.getValues()[i] = _aggregationFunctions[i].merge(v.getValues()[i], newRecord.getValues()[i]);
```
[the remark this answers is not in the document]
npawar: Done in both places for both files

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -218,22 +157,16 @@ private void resize(int trimToSize) {
...
   @Override
   public void finish(boolean sort) {
-    resize(_maxCapacity);
+    resize(_capacity);
     int numResizes = _numResizes.get();
     long resizeTime = _resizeTime.get();
     LOGGER.debug("Num resizes : {}, Total time spent in resizing : {}, Avg resize time : {}", numResizes, resizeTime,
         numResizes == 0 ? 0 : resizeTime / numResizes);
 
     _iterator = _lookupMap.values().iterator();
```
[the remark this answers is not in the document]
npawar: changed this some more

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -33,43 +33,41 @@
  */
 public abstract class IndexedTable implements Table {
 
-  List<AggregationFunction> _aggregationFunctions;
+  AggregationFunction[] _aggregationFunctions;
```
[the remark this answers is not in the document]
npawar: kept them as package-private because they're not and will not be used in any other packages.
Also, the IDE flags them if they are made protected as "Access can be package-private"

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.pinot.core.data.table;
+
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Helper class for trimming and sorting records in the IndexedTable
+ */
+abstract class IndexedTableResizer {
```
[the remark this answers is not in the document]
npawar: Removed it entirely

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -33,43 +33,41 @@
...
-          AggregationFunctionUtils.getAggregationFunctionContext(aggregationInfo).getAggregationFunction());
+    _aggregationFunctions = new AggregationFunction[_numAggregations];
+    for (int i = 0; i < _numAggregations; i++) {
+      _aggregationFunctions[i] =
+          AggregationFunctionUtils.getAggregationFunctionContext(aggregationInfos.get(i)).getAggregationFunction();
+    }
+
+    _capacity = capacity;
+    // TODO: tune these numbers
+    if (capacity <= 100_000) { // Capacity is small, make a very large buffer. Make PQ of records to retain, during resize
```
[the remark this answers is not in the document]
npawar: done

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -33,43 +33,41 @@
...
-    if (capacity > 50000) {
-      // if max capacity is large, buffer capacity is kept smaller, so that we do not accumulate too many records for sorting/resizing
-      bufferFactor = 1.2;
+    _isOrderBy = CollectionUtils.isNotEmpty(orderBy);
+    if (_isOrderBy) {
+      _indexedTableResizer = new OrderedIndexedTableResizer(dataSchema, aggregationInfos, orderBy);
     } else {
-      // if max capacity is small, buffer capacity is kept larger, so that we avoid frequent resizing
-      bufferFactor = 2.0;
+      _indexedTableResizer = new RandomIndexedTableResizer();
```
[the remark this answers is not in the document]
npawar: done

pinot-core/src/main/java/org/apache/pinot/core/data/table/IntermediateRecord.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * IntermediateRecord is derived from a Record
+ * Some of the main properties of an IntermediateRecord are:
+ *
+ * 1. Key in IntermediateRecord is expected to be identical to the one in the Record
+ * 2. For values, IntermediateRecord should only have the columns needed for order by
+ * 3. Inside the values, the columns should be ordered by the order by sequence
+ * 4. For order by on aggregations, final results should extracted if the intermediate result is non-comparable
+ */
+public class IntermediateRecord {
```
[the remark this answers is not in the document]
npawar: done

pinot-core/src/main/java/org/apache/pinot/core/data/table/RandomIndexedTableResizer.java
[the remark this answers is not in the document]
npawar: removed

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,260 @@
...
+
+  private List<OrderByInfo> _orderByInfos;
+  private Comparator<IntermediateRecord> _intermediateRecordComparator;
+  private int _numOrderBy;
+
+  OrderedIndexedTableResizer(DataSchema dataSchema, List<AggregationInfo> aggregationInfos,
+      List<SelectionSort> orderBy) {
+
+    int numAggregations = aggregationInfos.size();
+    int numKeyColumns = dataSchema.size() - numAggregations;
```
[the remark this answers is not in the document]
npawar: this is not a valid query. order by can only be done on aggregations or group by already in the query.

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
npawar: Changed it to resize and sort in one go using the same PQ when finishing.

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
npawar: True, we do not need to create it if we don't have to extract final result. However, the new IntermediateRecord would only have as many elements as the size of order by clause, which should be really small. It will only be alive during resize. The Key within IntermediateRecord is referring to the Key within the original Record (so no new Object there).
I'll add a TODO here, to re-evaluate this during further optimizations, and possibly avoid creating IntermediateRecord when possible. Right now, I'd prefer to avoid handling this, as it'll bring in a lot more if else.

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -33,43 +33,41 @@
...
 
-  int _maxCapacity;
+  int _capacity;
   int _bufferedCapacity;
 
+  boolean _isOrderBy;
+  IndexedTableResizer _indexedTableResizer;
+
   @Override
   public void init(@Nonnull DataSchema dataSchema, List<AggregationInfo> aggregationInfos, List<SelectionSort> orderBy,
```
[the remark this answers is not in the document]
npawar: No specific reason. I think since we designed the interface first, we ended up starting with an init.
I  changed it to constructor (read up about it and realized, as better practice, should not be using init if it can be done in constructor)

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
npawar: Great suggestion. Added an inner class `OrderByValueExtractor` with implementations `KeyColumnExtractor` and `ComparableAggregationColumnExtractor` and `NonComparableAggregationColumnExtractor`. Separated these 2 to further avoid the if check `if (aggregationFunction.isIntermediateResultComparable()`

pinot-core/src/main/java/org/apache/pinot/core/data/table/OrderedIndexedTableResizer.java
```diff
@@ -0,0 +1,260 @@
...
+    Comparable[] intermediateRecordValues = new Comparable[_numOrderBy];
+
+    Object[] keyColumns = record.getKey().getColumns();
+    Object[] aggregationColumns = record.getValues();
+    for (int i = 0; i < _numOrderBy; i++) {
+      OrderByInfo orderByInfo = _orderByInfos.get(i);
+      Comparable comparable;
+      if (orderByInfo.getOrderByType().equals(OrderByType.KEY_COLUMN)) {
+        Object keyColumn = keyColumns[orderByInfo.getIndex()];
+        comparable = (Comparable) keyColumn; // FIXME: is this the right way to get Comparable? will it work for BYTES?
```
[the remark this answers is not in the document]
npawar: changed the switch in CombineGroupByOrderByOperator to store ByteArray

npawar:
Performance numbers from BenchmarkCombineGroupBy:
This PR makes the numbers slightly better than the previous numbers with IndexedTable. We are still not completely at par with non IndexedTable approach 
Screen Shot 2019-10-24 at 2 16 06 PM

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -33,43 +32,44 @@
  */
 public abstract class IndexedTable implements Table {
 
-  List<AggregationFunction> _aggregationFunctions;
+  AggregationFunction[] _aggregationFunctions;
   int _numAggregations;
   DataSchema _dataSchema;
 
+  int _capacity;
```
mcvsubbu: A comment on what the difference is between 'capacity' and 'maxCapacity' will be nice. If I get it right, capacity is the number we need to trim to (at most) and maxCapacity is how high we are going to go in order to not drop new keys coming in.

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
```diff
@@ -0,0 +1,307 @@
...
+/**
+ * Helper class for trimming and sorting records in the IndexedTable, based on the order by information
+ */
+class IndexedTableResizer {
+
+  private OrderByValueExtractor[] _orderByValueExtractors;
+  private Comparator<IntermediateRecord> _intermediateRecordComparator;
+  private int _numOrderBy;
+
+  IndexedTableResizer(DataSchema dataSchema, List<AggregationInfo> aggregationInfos, List<SelectionSort> orderBy) {
```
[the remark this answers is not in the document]
Jackie-Jiang: Please check if `SELECT SUM(A) GROUP BY B ORDER BY SUM(C)` is valid in Calcite (doesn't need to be addressed in this PR)

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
```diff
@@ -0,0 +1,307 @@
...
+      return (Comparable) keyColumn;
+    }
+  }
+
+  /**
+   * Extractor for aggregation column
+   */
+  private static class AggregationColumnExtractor extends OrderByValueExtractor {
+    final int _index;
+    final AggregationFunction _aggregationFunction;
```
Jackie-Jiang: No need to store the aggregation function

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
```diff
@@ -0,0 +1,307 @@
...
+    int numAggregations = aggregationInfos.size();
+    int numKeyColumns = dataSchema.size() - numAggregations;
+
+    Map<String, Integer> keyIndexMap = new HashMap<>();
+    for (int i = 0; i < numKeyColumns; i++) {
+      String columnName = dataSchema.getColumnName(i);
+      keyIndexMap.put(columnName, i);
+    }
+
+    Map<String, Integer> aggregationColumnToIndex = new HashMap<>();
```
Jackie-Jiang: Store key index and aggregation column index into the same map, use the index to determine whether it is key or aggregation column

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
```diff
@@ -0,0 +1,307 @@
...
+   * Trim recordsMap to trimToSize, based on order by information
+   * Resize only if number of records is greater than trimToSize
+   * The resizer smartly chooses to create PQ of records to evict or records to retain, based on the number of records and the number of records to evict
+   */
+  void resizeRecordsMap(Map<Key, Record> recordsMap, int trimToSize) {
+
+    int numRecordsToEvict = recordsMap.size() - trimToSize;
+
+    if (numRecordsToEvict > 0) {
+      int size;
```
Jackie-Jiang: (nit) no need to declare these variables outside of the if clause. Easier to read if declared inside the if clause IMO

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
```diff
@@ -0,0 +1,307 @@
...
+    PriorityQueue<IntermediateRecord> priorityQueue = new PriorityQueue<>(size, comparator);
+
+    for (Record record : recordsMap.values()) {
+
+      IntermediateRecord intermediateRecord = getIntermediateRecord(record);
+      if (priorityQueue.size() < size) {
+        priorityQueue.offer(intermediateRecord);
+      } else {
+        IntermediateRecord peek = priorityQueue.peek();
+        if (priorityQueue.comparator().compare(peek, intermediateRecord) < 0) {
```
Jackie-Jiang: `comparator.compare()`

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
npawar: The set of indexes for keys and columns is not a disjoint set. For example, 
`select sum(a), sum(b) from table group by c order by c`
the keyToIndex map is `{c -> 0}` and aggregationToIndex map is `{sum(a) -> 0, sum(b) -> 1}`
This is because the key[] and value[] are stored as separate arrays in the Table.
So I cannot use the index to determine the column type right?

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
npawar: That's right. Added the comment in the IndexedTable

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -142,7 +124,7 @@ public boolean upsert(@Nonnull Record newRecord) {
   }
 
   @Override
-  public boolean merge(@Nonnull Table table) {
+  public boolean merge(Table table) {
```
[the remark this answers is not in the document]
npawar: done

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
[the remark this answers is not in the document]
npawar: Checked `SELECT SUM(column1) FROM testTable GROUP BY column11 ORDER BY column12` and `SELECT SUM(column1) FROM testTable GROUP BY column11 ORDER BY SUM(column6)`.
They are both valid! 
Added a TODO in the AggregationGroupByOrderByOperator, to start supporting reading order by columns

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTableResizer.java
npawar: after the offline discussion, changed it to 1 map

npawar:

PR #4730, opened 2019-10-21
Adding logs for HybridClusterIntegrationTest

jenniferdai:
* Determine whether virtual column test or teardown is flaky


Jackie-Jiang:
How is adding logs going to help with this?
Whenever a test failed, we should analyze the stack trace instead of the the log inside the tests.

jenniferdai:
spoke offline, will see if it is worth it to throw exception instead of log for zkstop

PR #4733, opened 2019-10-22
Add an integration test for LLC with PinotFS on non-local FS.

chenboat:
1. Add a new integration test on Low Level Consumer on non-local Pinot FS. A mini hdfs cluster is used in test instead.

2. Fix a bug in move() of PinotFS.java: it does not work with a non-local FS (i.e., uri starting with hdfs://)because  Paths.get() needs an external configured FilesystemProvider. A simpler and more reliable solution is to use URI.resolve(): it does not need any external dependency.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/filesystem/PinotFS.java
+++ b/pinot-common/src/main/java/org/apache/pinot/filesystem/PinotFS.java
@@ -22,7 +22,6 @@ import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
-import java.nio.file.Paths;
 import org.apache.commons.configuration.Configuration;
 import org.apache.pinot.annotations.InterfaceAudience;
 import org.apache.pinot.annotations.InterfaceStability;
@@ -101,7 +100,7 @@ public abstract class PinotFS implements Closeable {
       }
     } else {
       // ensures the parent path of dst exists.
-      URI parentUri = Paths.get(dstUri).getParent().toUri();
+      URI parentUri = dstUri.getPath().endsWith("/") ? dstUri.resolve("..") : dstUri.resolve(".");
       mkdir(parentUri);
     }
     return doMove(srcUri, dstUri);
--- a/pinot-hadoop-filesystem/src/main/java/org/apache/pinot/filesystem/HadoopPinotFS.java
+++ b/pinot-hadoop-filesystem/src/main/java/org/apache/pinot/filesystem/HadoopPinotFS.java
@@ -51,7 +51,7 @@ import static org.apache.pinot.common.utils.CommonConstants.SegmentOperations.RE
  */
 public class HadoopPinotFS extends PinotFS {
   private static final Logger LOGGER = LoggerFactory.getLogger(HadoopPinotFS.class);
-  private org.apache.hadoop.fs.FileSystem _hadoopFS = null;
+  protected org.apache.hadoop.fs.FileSystem _hadoopFS = null;
   private int _retryCount = RETRY_DEFAULT;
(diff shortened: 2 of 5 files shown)


pinot-common/src/main/java/org/apache/pinot/filesystem/PinotFS.java
```diff
@@ -101,7 +100,7 @@ public boolean move(URI srcUri, URI dstUri, boolean overwrite)
       }
     } else {
       // ensures the parent path of dst exists.
-      URI parentUri = Paths.get(dstUri).getParent().toUri();
+      URI parentUri = dstUri.getPath().endsWith("/") ? dstUri.resolve("..") : dstUri.resolve(".");
```
Jackie-Jiang: To me, this also has dependency on certain FS. Maybe better to add a `getParent()` API to `PinotFS`?

pinot-hadoop-filesystem/src/main/java/org/apache/pinot/filesystem/HadoopPinotFS.java
```diff
@@ -51,7 +51,7 @@
  */
 public class HadoopPinotFS extends PinotFS {
   private static final Logger LOGGER = LoggerFactory.getLogger(HadoopPinotFS.class);
-  private org.apache.hadoop.fs.FileSystem _hadoopFS = null;
+  protected org.apache.hadoop.fs.FileSystem _hadoopFS = null;
```
Jackie-Jiang: Can this be replaced with `import org.apache.hadoop.fs.FileSystem;`?

Jackie-Jiang (commented):
Thanks for the fix. I think the scope of this PR should be fix the PinotFS.move() for non-local FS. Adding an integration test seems too much for this fix (which will add 2-3 minutes to each Travis run), is it possible to cover this by enhancing the unit test for PinotFS?

chenboat:
> Thanks for the fix. I think the scope of this PR should be fix the PinotFS.move() for non-local FS. Adding an integration test seems too much for this fix (which will add 2-3 minutes to each Travis run), is it possible to cover this by enhancing the unit test for PinotFS?

I can add a unit test to PinotFS to cover this function. The integration test is important to us though because we use LLC+HDFS and needs some integration test coverage. The bug is found when we run the integration test. Currently we run an integration test internally and wanted to contribute it back to open source. Is there any way to make Travis run faster (e.g., to do parallel run of tests)? If that is hard, I could think about other ways to do integration tests in separate PRs.

Jackie-Jiang:
> > Thanks for the fix. I think the scope of this PR should be fix the PinotFS.move() for non-local FS. Adding an integration test seems too much for this fix (which will add 2-3 minutes to each Travis run), is it possible to cover this by enhancing the unit test for PinotFS?
> 
> I can add a unit test to PinotFS to cover this function. The integration test is important to us though because we use LLC+HDFS and needs some integration test coverage. The bug is found when we run the integration test. Currently we run an integration test internally and wanted to contribute it back to open source. Is there any way to make Travis run faster (e.g., to do parallel run of tests)? If that is hard, I could think about other ways to do integration tests in separate PRs.

Travis is free for open-source project, but has very limited resource. One way to test this in integration test but not adding another test is by using random in the existing LLC integration test to randomly use localFS or HadoopFS. The reason why I'm against adding another integration test for this is because all the query engine tests are the same as the existing LLC integration test. PinotFS tests should be able to to cover all the FS related code.

xiangfu0:
Closing this PR as stale because it has had no activity for more than one year (last updated before 2025-06-18). If this is still relevant, please reopen it or open a fresh PR against current master. Thanks for the contribution.

PR #4734, opened 2019-10-22
Enhance the exception handling in TableRebalancer

Jackie-Jiang:
Catch more types of exceptions in TableRebalancer.rebalance() and wrap the exception into the RebalanceResult
This can make the rebalance failure more gracefully, and easier to understand

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
@@ -98,8 +98,15 @@ public class TableRebalancer {
     }
 
     // Check that table exists and is enabled
-    IdealState currentIdealState =
-        _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().idealStates(tableNameWithType));
+    IdealState currentIdealState;
+    try {
+      currentIdealState =
+          _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().idealStates(tableNameWithType));
+    } catch (Exception e) {
+      LOGGER.warn("Caught exception while fetching IdealState for table: {}", tableNameWithType, e);
+      return new RebalanceResult(RebalanceResult.Status.FAILED, "Caught exception while fetching IdealState: " + e,
+          null, null);
+    }
     if (currentIdealState == null) {
       return new RebalanceResult(RebalanceResult.Status.FAILED, "Cannot find the table", null, null);
     }
@@ -110,11 +117,17 @@ public class TableRebalancer {
     // Check if any segment is in ERROR state in ExternalView before rebalancing the table. ERROR state segment cannot
     // be transferred to other state, thus will cause the rebalance to never converge.
     // NOTE: ExternalView might be null when table is just created
-    ExternalView externalView =
-        _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().externalView(tableNameWithType));
-    if (externalView != null && hasSegmentInErrorState(externalView.getRecord().getMapFields())) {
-      LOGGER.warn("Found segments in ERROR state for table: {}", tableNameWithType);
-      return new RebalanceResult(RebalanceResult.Status.FAILED, "Found segments in ERROR state", null, null);
+    try {
+      ExternalView externalView =
+          _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().externalView(tableNameWithType));
+      if (externalView != null && hasSegmentInErrorState(externalView.getRecord().getMapFields())) {
+        LOGGER.warn("Found segments in ERROR state for table: {}", tableNameWithType);
+        return new RebalanceResult(RebalanceResult.Status.FAILED, "Found segments in ERROR state", null, null);
+      }
+    } catch (Exception e) {
+      LOGGER.warn("Caught exception while fetching ExternalView for table: {}", tableNameWithType, e);
+      return new RebalanceResult(RebalanceResult.Status.FAILED, "Caught exception while fetching ExternalView: " + e,
+          null, null);
     }
 
     boolean dryRun =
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
```diff
@@ -98,8 +98,15 @@ public RebalanceResult rebalance(TableConfig tableConfig, Configuration rebalanc
...
 
     // Check that table exists and is enabled
-    IdealState currentIdealState =
-        _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().idealStates(tableNameWithType));
+    IdealState currentIdealState;
+    try {
+      currentIdealState =
+          _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().idealStates(tableNameWithType));
+    } catch (Exception e) {
+      LOGGER.warn("Caught exception while fetching IdealState for table: {}", tableNameWithType, e);
```
mcvsubbu: Since these exceptions are caused by operator action, we should log them as INFO, not WARN

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
```diff
@@ -110,11 +117,17 @@ public RebalanceResult rebalance(TableConfig tableConfig, Configuration rebalanc
...
-    ExternalView externalView =
-        _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().externalView(tableNameWithType));
-    if (externalView != null && hasSegmentInErrorState(externalView.getRecord().getMapFields())) {
-      LOGGER.warn("Found segments in ERROR state for table: {}", tableNameWithType);
-      return new RebalanceResult(RebalanceResult.Status.FAILED, "Found segments in ERROR state", null, null);
+    try {
+      ExternalView externalView =
+          _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().externalView(tableNameWithType));
+      if (externalView != null && hasSegmentInErrorState(externalView.getRecord().getMapFields())) {
+        LOGGER.warn("Found segments in ERROR state for table: {}", tableNameWithType);
```
mcvsubbu: Same here

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
```diff
@@ -110,11 +117,17 @@ public RebalanceResult rebalance(TableConfig tableConfig, Configuration rebalanc
...
+        return new RebalanceResult(RebalanceResult.Status.FAILED, "Found segments in ERROR state", null, null);
+      }
+    } catch (Exception e) {
+      LOGGER.warn("Caught exception while fetching ExternalView for table: {}", tableNameWithType, e);
```
mcvsubbu: Same, will not repeat.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
```diff
@@ -98,8 +98,15 @@ public RebalanceResult rebalance(TableConfig tableConfig, Configuration rebalanc
...
     // Check that table exists and is enabled
-    IdealState currentIdealState =
-        _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().idealStates(tableNameWithType));
+    IdealState currentIdealState;
+    try {
+      currentIdealState =
+          _helixDataAccessor.getProperty(_helixDataAccessor.keyBuilder().idealStates(tableNameWithType));
+    } catch (Exception e) {
+      LOGGER.warn("Caught exception while fetching IdealState for table: {}", tableNameWithType, e);
+      return new RebalanceResult(RebalanceResult.Status.FAILED, "Caught exception while fetching IdealState: " + e,
```
jackjlli: Not related to this PR but you can create a new constructor like `RebalanceResult(Status status, String description)`, so that you don't need to pass the null value all the time.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
```diff
@@ -126,24 +139,37 @@ public RebalanceResult rebalance(TableConfig tableConfig, Configuration rebalanc
...
     SegmentAssignment segmentAssignment = SegmentAssignmentFactory.getSegmentAssignment(_helixManager, tableConfig);
     Map<String, Map<String, String>> currentAssignment = currentIdealState.getRecord().getMapFields();
-    Map<String, Map<String, String>> targetAssignment =
-        segmentAssignment.rebalanceTable(currentAssignment, instancePartitionsMap, rebalanceConfig);
+    Map<String, Map<String, String>> targetAssignment;
+    try {
+      targetAssignment = segmentAssignment.rebalanceTable(currentAssignment, instancePartitionsMap, rebalanceConfig);
+    } catch (Exception e) {
+      LOGGER.warn("Caught exception while calculating target assignment for table: {}", tableNameWithType, e);
+      return new RebalanceResult(RebalanceResult.Status.FAILED, "Caught exception while fetching ExternalView: " + e,
```
jackjlli: This catch block only catches exception when rebalancing the table. Maybe this message needs modification?

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
Jackie-Jiang: These exceptions are unusual system exception instead of invalid operator action (ZK exception instead of table does not exist). Also, real-time segment relocation is also done by TableRebalancer, so I would not treat it as a user facing tool. Same for other WARNING. For invalid operator action such as table does not exist/rebalance HLC table, no warning is logged.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
Jackie-Jiang: I prefer making it more explicit haha

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
Jackie-Jiang: Good catch, I copy-pasted from the previous message and missed it

PR #4735, opened 2019-10-22
Changing error log to warn in SegmentPreprocessingMapper

jenniferdai:
* SegmentPreprocessingMapper currently logs an error -> changing it to a warn.

### DIFF
--- a/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mappers/SegmentPreprocessingMapper.java
+++ b/pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mappers/SegmentPreprocessingMapper.java
@@ -94,7 +94,7 @@ public class SegmentPreprocessingMapper extends Mapper<AvroKey<GenericRecord>, N
       if (!normalizedTimeColumnValue.equals(_sampleNormalizedTimeColumnValue)) {
         // TODO: Create a custom exception and gracefully catch this exception outside, changing what the path to input
         // into segment creation should be
-        LOGGER.error("This segment contains multiple time units. Sample is {}, current is {}", _sampleNormalizedTimeColumnValue, normalizedTimeColumnValue);
+        LOGGER.warn("This segment contains multiple time units. Sample is {}, current is {}", _sampleNormalizedTimeColumnValue, normalizedTimeColumnValue);
       }
     }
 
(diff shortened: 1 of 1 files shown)


pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mappers/SegmentPreprocessingMapper.java
```diff
@@ -94,7 +94,7 @@ public void map(AvroKey<GenericRecord> record, NullWritable value, final Context
       if (!normalizedTimeColumnValue.equals(_sampleNormalizedTimeColumnValue)) {
```
snleee: Can you explain on why we need to put this log? We don't enforce that a single segment need to contain only one day's of data.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mappers/SegmentPreprocessingMapper.java
```diff
@@ -94,7 +94,7 @@ public void map(AvroKey<GenericRecord> record, NullWritable value, final Context
...
         // TODO: Create a custom exception and gracefully catch this exception outside, changing what the path to input
         // into segment creation should be
-        LOGGER.error("This segment contains multiple time units. Sample is {}, current is {}", _sampleNormalizedTimeColumnValue, normalizedTimeColumnValue);
+        LOGGER.warn("This segment contains multiple time units. Sample is {}, current is {}", _sampleNormalizedTimeColumnValue, normalizedTimeColumnValue);
```
snleee: We should avoid to put record level log unless it would really provide the good information. Even if changing the log level to `WARN` won't log to Travis, this will be logged to other jobs (actual hadoop job, TMC test..etc).

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mappers/SegmentPreprocessingMapper.java
jenniferdai: When you decide to rollup segments, we want to make sure that only one time unit is rolled up, so it is theoretically safer to fail preprocessing if more than one time unit is contained (although I know we have some segments that cross timezones, in which case there will be more than one timeunit). I wanted to fail the job entirely and not allow customers to continue, but I settled for an error.

pinot-hadoop/src/main/java/org/apache/pinot/hadoop/job/mappers/SegmentPreprocessingMapper.java
jenniferdai: This will provide good information - see above

PR #4737, opened 2019-10-23
Druid-Pinot Segment Converter Tool

dbadapon:
Adding a reader for Druid segment files so that Druid segments can be converted to Pinot segments.
Part of a new tool for automating migration of data from Druid to Pinot.


pinot-druid/src/main/java1/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,183 @@
...
+  @Override
+  public GenericRow next(GenericRow reuse) throws IOException {
+    // TODO: use Pinot schema to fill the values to handle missing column and default values properly
+    // Only the columns whose names are in the Pinot schema will get processed
+    for (int i = 0; i < _columnNames.size(); i++) {
+      final String columnName = _columnNames.get(i);
+      final Object value = _selectors.get(i).getObject();
+      // Figure out how Pinot handles time; see TimeGranularitySpec and TimeFieldSpec
+      if (_columnNames.get(i).equals(ColumnHolder.TIME_COLUMN_NAME)) {
+        reuse.putField(columnName, new DateTime());
```
snleee: Druid has a special time column `_time`. We need to appropriately handle the timestamp. By putting `new DateTime()`, you're overwriting the time column value. We should put the value what's read from Druid segment.

What the type of `Object value` for the time column returned by Druid segment?

pinot-druid/src/main/java1/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  @Override
+  public Schema getSchema() {
+    return _pinotSchema;
+  }
+
+  @Override
+  public void close() throws IOException {
+    // Not sure what to close; the Yielder was the only Closeable.
```
snleee:
```
QueryableIndex.java
  /**
   * The close method shouldn't actually be here as this is nasty. We will adjust it in the future.
   * @throws IOException if an exception was thrown closing the index
   */
  //@Deprecated // This is still required for SimpleQueryableIndex. It should not go away unitl SimpleQueryableIndex is fixed
  @Override
  void close();
```

pinot-druid/src/main/java1/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,183 @@
+package org.apache.pinot.druid.data.readers;
```
snleee: Add the header

kishoreg:
we have too many modules at the top level. lets move the package to a sub-folder (pinot-contrib). renamed the module to druid-2-pinot-migration or something like that.

How can a user this tool?

Will it be something like, Druid2Pinot <druid_cluster_endpoint_config>  <pinot_cluster_endpoint_config>?

At a high level, what are the various steps involved in this? A simple issue with some description will help in understanding the goal and reviewing this PR.

pinot-druid/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,219 @@
...
+public class DruidSegmentRecordReader implements RecordReader {
+  // TODO: Figure out why logger stuff isn't showing up in console
+  private static final Logger LOGGER = LoggerFactory.getLogger(DruidSegmentRecordReader.class);
+
+  private Schema _pinotSchema;
+  private Cursor _cursor;
+  private ArrayList<String> _columnNames;
+  private List<BaseObjectColumnValueSelector> _selectors;
+  private QueryableIndex _index;
+  private QueryableIndexStorageAdapter _adapter;
```
snleee: I think that we don't need to keep this as the instance variable. It seems to be only used in init().

pinot-druid/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,219 @@
...
+
+  @Override
+  public GenericRow next(GenericRow reuse) {
+    // Only the columns whose names are in the Pinot schema will get processed
+    for (int i = 0; i < _columnNames.size(); i++) {
+      final String columnName = _columnNames.get(i);
+      BaseObjectColumnValueSelector selector = _selectors.get(i);
+      // If the column does not exist in the segment file, skip it.
+      if (selector != null) {
+        final Object value = selector.getObject();
```
snleee: Have we checked what types of object returned by Druid for all possible Pinot columns types? Can you add this mapping information to the design document?

If we are sure that for all 

e.g. For long column, `selector.getObject()` returns `Long` object in Druid, and Pinot also uses `Long` obejct for `GenericRow`

For multi-value long column, `select.getObject()` returns `???` object in Druid, and Pinot uses `Long[]` Long array for `GenericRow` etc.

pinot-druid/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * Note that Druid uses LONG and not INT, so construct the Pinot schema accordingly.
+ */
+public class DruidSegmentRecordReader implements RecordReader {
+  // TODO: Figure out why logger stuff isn't showing up in console
+  private static final Logger LOGGER = LoggerFactory.getLogger(DruidSegmentRecordReader.class);
+
+  private Schema _pinotSchema;
+  private Cursor _cursor;
+  private ArrayList<String> _columnNames;
```
snleee: Since we can get this information easily by calling `_pinotSchema.getColumnNames()`, I think that we can remove this from putting it as an instance variable.

pinot-druid/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,219 @@
...
+  private List<BaseObjectColumnValueSelector> _selectors;
+  private QueryableIndex _index;
+  private QueryableIndexStorageAdapter _adapter;
+
+  private void validateColumns()
+      throws IllegalArgumentException {
```
snleee: Remove `throw IllegalArgumentExcpetion` since it's runtime exception.

pinot-druid/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,219 @@
...
+          throw new IllegalArgumentException("Column " + columnName
+              + " in Pinot schema is single-valued, but column {} in " + "record is multi-valued.");
+        }
+        if (!fieldSpec.isSingleValueField() && !capabilities.hasMultipleValues()) {
+          throw new IllegalArgumentException("Column " + columnName
+              + " in Pinot schema is multi-valued, but column {} in" + "record is single-valued");
+        }
+        String typeFromSchema = _pinotSchema.getFieldSpecFor(columnName).getDataType().toString();
+        String typeFromRecord = capabilities.getType().toString();
+        if (typeFromRecord != typeFromSchema) {
```
snleee: 1. You need to use `typeFromRecord.equals(typeFromSchema)` to compare the string. `!=` will only check if they are the same reference. https://stackoverflow.com/questions/513832/how-do-i-compare-strings-in-java
2. Pinot will give you `DataType` and Druid will give you `ValueType`. Instead of comparing `String` value directly, I think that it's better to add a compare function something like the following:

```
private boolean compareType(DataType pinotColumnType, ValueType druidColumnType) 
```
You can handle the following type equality within the function.

```
 public enum DataType {
    INT, LONG, FLOAT, DOUBLE, BOOLEAN/* Stored as STRING */, STRING, BYTES;
  }
```

```
public enum ValueType
{
  FLOAT, DOUBLE, LONG, STRING, COMPLEX 
}
```

pinot-druid/src/test/java/org/apache/pinot/druid/data/readers/TestDruidSegmentRecordReader.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+import org.apache.pinot.common.data.Schema;
+import org.apache.pinot.core.data.GenericRow;
+import org.apache.pinot.core.indexsegment.generator.SegmentGeneratorConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testng.annotations.Test;
+
+
+public class TestDruidSegmentRecordReader {
```
snleee: Let's improve testing code in general :) 

1. add more tests as you listed in `TODO`
2. make them complete (e.g. use `Assert.xxx` instead of print out).

pinot-druid/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
dbadapon: I also use it in validateColumns. should I still make it not an instance variable?

pinot-druid/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
dbadapon: I also use it in validateColumns. should I still make it not an instance variable?

dbadapon: I also use it in next()
`    ArrayList<String> columnNames = new ArrayList<>();
    columnNames.addAll(_pinotSchema.getColumnNames());`

dbadapon: I need to make it into a list of column names that I can use in init, validateColumns, and next()

dbadapon: I need to convert the result of `getColumnNames()` (Set) into a List so that I can iterate them in a certain order throughout the record reader because it will correspond to a list of selectors that are specific for each column.

snleee:
> we have too many modules at the top level. lets move the package to a sub-folder (pinot-contrib). renamed the module to druid-2-pinot-migration or something like that.
> 
> How can a user this tool?
> 
> Will it be something like, Druid2Pinot <druid_cluster_endpoint_config> <pinot_cluster_endpoint_config>?
> 
> At a high level, what are the various steps involved in this? A simple issue with some description will help in understanding the goal and reviewing this PR.

We will address the package location before we check in this pr. We were also debating top level vs `pinot-contrib`. 

What's your opinion on adding druid dependency to `pinot-tools` or `pinot-hadoop`? If we want to use existing segment generation tools (e.g. `pinot-admin's SegmentGenerationCommand` or in `PinotSegmentGeneration` hadoop job), we need to pull druid dependency. (`pinot-tools` and `pinot-hadoop` will pull `pinot-druid-migration`, which pulls druid.

Another approach is to provide this tool as a completely separate code outside of Pinot project. In that case, we can add the standalone class with its own Main function for cmd line tool and provide a standalone hadoop job. All code will be under `pinot-contrib/pinot-druid-migration` and other part of Pinot code won't import this package.

For user interface, our current approach is to reuse the existing segment generation tools (cmd line, hadoop job) and add `DRUID_SEGMENT` as an extra input data format that we support.

We can add more tools around segment converter to automatize end-to-end migration process.

kishoreg:
> > we have too many modules at the top level. lets move the package to a sub-folder (pinot-contrib). renamed the module to druid-2-pinot-migration or something like that.
> > How can a user this tool?
> > Will it be something like, Druid2Pinot <druid_cluster_endpoint_config> <pinot_cluster_endpoint_config>?
> > At a high level, what are the various steps involved in this? A simple issue with some description will help in understanding the goal and reviewing this PR.
> 
> @kishoreg
> 
> We will address the package location before we check in this pr. We were also debating top level vs `pinot-contrib`.
> 
> What's your opinion on adding druid dependency to `pinot-tools` or `pinot-hadoop`? If we want to use existing segment generation tools (e.g. `pinot-admin's SegmentGenerationCommand` or in `PinotSegmentGeneration` hadoop job), we need to pull druid dependency. (`pinot-tools` and `pinot-hadoop` will pull `pinot-druid-migration`, which pulls druid.
> 
> Another approach is to provide this tool as a completely separate code outside of Pinot project. In that case, we can add the standalone class with its own Main function for cmd line tool and provide a standalone hadoop job. All code will be under `pinot-contrib/pinot-druid-migration` and other part of Pinot code won't import this package.
> 
> For user interface, our current approach is to reuse the existing segment generation tools (cmd line, hadoop job) and add `DRUID_SEGMENT` as an extra input data format that we support.
> 
> We can add more tools around segment converter to automatize end-to-end migration process.

Let's not add druid dependency in pinot-tools/hadoop.  Its better to keep this as a separate project for now.

pinot-druid/src/test/java/org/apache/pinot/druid/data/readers/TestDruidSegmentRecordReader.java
```diff
@@ -0,0 +1,71 @@
...
+    SegmentGeneratorConfig segmentGeneratorConfig = new SegmentGeneratorConfig();
+    segmentGeneratorConfig.setInputFilePath(indexPath);
+    segmentGeneratorConfig.setSchema(schema);
+
+    try {
+      testReader.init(segmentGeneratorConfig);
+    } catch (Exception e) {
+      System.out.println("INIT RECORDREADER FAILED");
+      LOGGER.error("Failed to initialize DruidSegmentRecordReader: ", e);
+      return;
```
snleee: If some error happens `init()` here, the test will log error message and return, which means that the test will pass if we reach here. Is this expected behavior?

pinot-druid/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,248 @@
...
+    // Only the columns whose names are in the Pinot schema will get processed
+    for (int i = 0; i < _columnNames.size(); i++) {
+      final String columnName = _columnNames.get(i);
+      BaseObjectColumnValueSelector selector = _selectors.get(i);
+      // If the column does not exist in the segment file, skip it.
+      if (selector != null) {
+        FieldSpec fieldSpec = _pinotSchema.getFieldSpecFor(_columnNames.get(i));
+        Object value = selector.getObject();
+        if (value != null && !fieldSpec.isSingleValueField()) {
+          // Assuming that multi-valued dimensions in Druid are stored as Arrays.ArrayList (this has been checked)
```
snleee: Instead of assuming, I think it's better to add `Precondition` check?

pinot-druid/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.pinot.common.data.Schema;
+import org.apache.pinot.core.data.GenericRow;
+import org.apache.pinot.core.data.recordtransformer.CompositeTransformer;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+
+public class DruidSegmentRecordReaderTest {
+  private static int wikipedia_size = 39244;
```
snleee: `private static final int`? 

Also, Java convention for `static final` variable name is all capital. `e.g. WIKIPEDIA_SIZE`

Please make the change for all variables defined here.

pinot-druid/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,141 @@
...
+    csv_test_schema = createSchemaFromFilename("csv-test-schema.json");
+
+    all_types_test = createFileFromFilename("test-druid-all-types");
+    all_types_test_schema = createSchemaFromFilename("test-druid-all-types-schema.json");
+  }
+
+  @Test
+  public void testReadAllRows()
+      throws IOException {
+    DruidSegmentRecordReader testReader;
```
snleee: `DruidSegmentRecordReader testReader = new DruidSegmentRecordReader(wikipedia_test, wikipedia_test_schema);`

pinot-druid/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,141 @@
...
+    testReader = new DruidSegmentRecordReader(wikipedia_test, wikipedia_test_schema);
+
+    List<GenericRow> genericRows = new ArrayList<>();
+    while (testReader.hasNext()) {
+      GenericRow nextRow = testReader.next();
+      genericRows.add(nextRow);
+    }
+    Assert.assertEquals(genericRows.size(), wikipedia_size, "Generic row size must be " + wikipedia_size + ".");
```
snleee: I think that you're only checking `genericRows.size()`

pinot-druid/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  @Test
+  public void testCSVRecord()
+      throws IOException {
+    final String[] columms = {"sum_INT_SV", "INT_MV", "__time"};
```
snleee: `INT_MV` looks like `integer multivalue` column. Can you change this to `STRING_MV`? (String multivalue column is what you are actually testing)

pinot-druid/src/test/resources/test_druid.csv
```diff
@@ -0,0 +1,4 @@
+INT_SV,INT_MV
```
snleee: `STRING_MV`?

pinot-druid/src/test/resources/test_druid.csv
```diff
@@ -0,0 +1,4 @@
+INT_SV,INT_MV
+5,"10";"15";"20"
```
snleee: Also, where do you use this file?

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSegmentConverter.java
```diff
@@ -0,0 +1,77 @@
...
+    _schemaFilePath = args[2];
+    _segmentName = args[3];
+    _tableName = args[4];
+
+    try {
+      convertSegment();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+
```
snleee: Remove the line

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSegmentConverter.java
```diff
@@ -0,0 +1,77 @@
...
+    final SegmentIndexCreationDriverImpl driver = new SegmentIndexCreationDriverImpl();
+    SegmentGeneratorConfig config = new SegmentGeneratorConfig(segmentGeneratorConfig);
+    DruidSegmentRecordReader recordReader = new DruidSegmentRecordReader(segment, segmentGeneratorConfig.getSchema());
+    driver.init(config, recordReader);
+    driver.build();
+  }
+
+  public static void main(String[] args) {
+    // Arg format is segmentPath, outDir, schemaPath, segmentName, tableName
```
snleee: When user runs this with no input, you can print out the information on how to use this command. Otherwise, we don't have the way to know which arguments this cmd takes.

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSegmentConverter.java
```diff
@@ -0,0 +1,77 @@
...
+    Schema schema = Schema.fromFile(new File(_schemaFilePath));
+
+    final SegmentGeneratorConfig segmentGeneratorConfig = new SegmentGeneratorConfig();
+    segmentGeneratorConfig.setDataDir(_druidSegmentPath);
+    segmentGeneratorConfig.setOutDir(_outPath);
+    segmentGeneratorConfig.setOverwrite(true);
+    segmentGeneratorConfig.setTableName(_tableName);
+    segmentGeneratorConfig.setSegmentName(_segmentName);
+    segmentGeneratorConfig.setSchemaFile(_schemaFilePath);
+    segmentGeneratorConfig.setSchema(schema);
```
snleee: Remove `setSchema(schema)`

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,252 @@
...
+    for (int i = 0; i < _columnNames.size(); i++) {
+      final String columnName = _columnNames.get(i);
+      BaseObjectColumnValueSelector selector = _selectors.get(i);
+      // If the column does not exist in the segment file, skip it.
+      if (selector != null) {
+        FieldSpec fieldSpec = _pinotSchema.getFieldSpecFor(_columnNames.get(i));
+        Object value = selector.getObject();
+        if (value != null && !fieldSpec.isSingleValueField()) {
+          // Multi-valued dimensions in Druid are stored as Arrays.ArrayList (this has been checked)
+          Preconditions.checkArgument(value.getClass() == java.util.Arrays.asList().getClass(), "The multi-valued dimension " + columnName +
```
snleee: Try to emit the error message like the following:
```
String errorMsg = String.format(...);
Precondition.checkArgument(..., errorMsg);
```

snleee: For java, when you do the equality check, you should always use `.equals()` if you are comparing java objects. You can only use `==` for primitive types.

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,252 @@
...
+        }
+        reuse.putField(columnName, value);
+      }
+    }
+    _cursor.advance();
+    return reuse;
+  }
+
+  @Override
+  public void rewind() {
```
snleee: Can we add the test to see if `rewind()` works correctly?

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,252 @@
...
+ * The DruidSegmentRecordReader allows us to convert all of the rows in a Druid segment file
+ * into GenericRows, which are made into Pinot segments.
+ *
+ * Note that Druid uses LONG and not INT, so construct the Pinot schema accordingly.
+ */
+public class DruidSegmentRecordReader implements RecordReader {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DruidSegmentRecordReader.class);
+
+  private Schema _pinotSchema;
+  private ArrayList<String> _columnNames;
```
snleee: `private List<String> _columnNames`

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,252 @@
...
+        return null;
+      }
+    });
+    sequence.accumulate(null, (accumulated, in) -> null);
+
+    // There should only be one single Cursor for every segment, so there should only be one Cursor in the cursorList
+    Preconditions.checkArgument(cursorList.size() == 1, "There should only be one Cursor in the Sequence.");
+    _cursor = cursorList.get(0);
+
+    _columnNames = new ArrayList<>();
```
snleee: Do we need to save this as an instance variable since we can easily access to this by calling `_pinotSchema.getColumnNames()`?

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,252 @@
...
+
+  @Override
+  public void close() {
+    _index.close();
+  }
+
+  private boolean compareTypes(FieldSpec.DataType pinotColumnType, ValueType druidColumnType) {
+    switch (pinotColumnType) {
+      case INT:
+        return false;
```
snleee: If Druid doesn't support `int`, I think that it's better to handle this together with `BOOLEAN, BYTES`.

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,252 @@
...
+      case FLOAT:
+        return druidColumnType == ValueType.FLOAT;
+      case DOUBLE:
+        return druidColumnType == ValueType.DOUBLE;
+      case STRING:
+        return druidColumnType == ValueType.STRING;
+      case BOOLEAN:
+      case BYTES:
+      default:
+        throw new UnsupportedOperationException("Unsupported Druid type: " + pinotColumnType.name());
```
snleee: You should print both Druid and Pinot types.

contrib/pinot-druid-migration/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <artifactId>pinot-druid-migration</artifactId>
+  <groupId>org.apache.pinot</groupId>
+  <version>0.2.0-SNAPSHOT</version>
+  <modelVersion>4.0.0</modelVersion>
+
+  <name>Pinot Druid</name>
```
snleee: Pinot Druid Migration?

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+        .addSingleValueDimension("unknown_dimension", FieldSpec.DataType.STRING)
+        .addMetric("met_impressionCount", FieldSpec.DataType.LONG).addMetric("unknown_metric", FieldSpec.DataType.DOUBLE)
+        .build();
+  // Test based on CSVRecordReaderTest
+  private static final File CSV_TEST = new File(Preconditions
+      .checkNotNull(DruidSegmentRecordReaderTest.class.getClassLoader().getResource("test_druid_from_csv"))
+      .getFile());
+  private static final Schema CSV_TEST_SCHEMA = new Schema.SchemaBuilder().addSingleValueDimension("__time", FieldSpec.DataType.LONG)
+      .addMultiValueDimension("STR_MV", FieldSpec.DataType.STRING)
+      .addMetric("INT_SV", FieldSpec.DataType.LONG)
```
snleee: `LONG_SV`?

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+      .build();
+  private static final File ALL_TYPES_TEST = new File(Preconditions
+      .checkNotNull(DruidSegmentRecordReaderTest.class.getClassLoader().getResource("test_druid_all_types"))
+      .getFile());
+  private static final Schema ALL_TYPES_TEST_SCHEMA = new Schema.SchemaBuilder().addSingleValueDimension("__time", FieldSpec.DataType.LONG)
+      .addSingleValueDimension("stringval", FieldSpec.DataType.STRING)
+      .addMultiValueDimension("tags", FieldSpec.DataType.STRING)
+      .addMetric("intnum", FieldSpec.DataType.LONG)
+      .addMetric("floatnum", FieldSpec.DataType.FLOAT)
```
snleee: Druid also supports `double`. Can we add it to the test?

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+        Assert.assertTrue(row < records.length);
+
+        GenericRow nextRow = testReader.next();
+        GenericRow transformedRow = defaultTransformer.transform(nextRow);
+
+        for (int col = 0; col < columns.length; col++) {
+          if (records[row][col] != null) {
+            if (transformedRow.getValue(columns[col]) != null && records[row][col] != null) {
+              Assert.assertTrue(transformedRow.getValue(columns[col]).getClass() == records[row][col].getClass(),
+                  String.format("In row {} and column {}, type of value in segment ({}) does not match type of value in test record ({}).",
```
snleee: Fix `String.format()`

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+                Assert.assertEquals(transformedRow.getValue(columns[col]), records[row][col]);
+              }
+            }
+          }
+        }
+        genericRows.add(nextRow);
+        row += 1;
+      }
+      Assert.assertFalse(testReader.hasNext());
+      Assert.assertEquals(genericRows.size(), records.length, String.format("Size of GenericRows ({}) must be equal to records size ({}).", genericRows.size(), records.length));
```
snleee: Fix `String.format()`

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSegmentConverter.java
```diff
@@ -0,0 +1,78 @@
...
+    SegmentGeneratorConfig config = new SegmentGeneratorConfig(segmentGeneratorConfig);
+    DruidSegmentRecordReader recordReader = new DruidSegmentRecordReader(segment, segmentGeneratorConfig.getSchema());
+    driver.init(config, recordReader);
+    driver.build();
+  }
+
+  public static void main(String[] args) {
+    if (args.length != 5) {
+      System.out.println("Convert a Druid segment into Pinot format.");
+      System.out.println("Use with arguments: <table-name> <segment-name> <segment-path> <schema-path> <output-directory>");
```
snleee:
```
System.out.println("Usage: ");
System.out.println("DruidToPinotSegmentConverter <pinot_table_name> <pinot_segment_name>  <pinot_schema_path> <druid_input_path> <output_path>");

```

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSegmentConverter.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.pinot.core.indexsegment.generator.SegmentGeneratorConfig;
+import org.apache.pinot.core.segment.creator.impl.SegmentIndexCreationDriverImpl;
+import org.apache.pinot.druid.data.readers.DruidSegmentRecordReader;
+
+
+/**
+ * The DruidToPinotSegmentConverter is a CLI tool that converts a Druid segment to a Pinot segment.
+ */
+public class DruidToPinotSegmentConverter {
+  private static String _schemaFilePath;
```
snleee: Can you change the variable names to be a bit more expressive? (e.g. `_schemaFilePath` -> `_pinotSchemaPath`)

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSegmentConverter.java
```diff
@@ -0,0 +1,78 @@
...
+      _tableName = args[0];
+      _segmentName = args[1];
+      _druidSegmentPath = args[2];
+      _schemaFilePath = args[3];
+      _outPath = args[4];
+
+      try {
+        convertSegment();
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
snleee: Why do we catch exception and re-throw it as a runtime exception here?

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+      .addMultiValueDimension("STR_MV", FieldSpec.DataType.STRING)
+      .addMetric("INT_SV", FieldSpec.DataType.LONG)
+      .build();
+  private static final File ALL_TYPES_TEST = new File(Preconditions
+      .checkNotNull(DruidSegmentRecordReaderTest.class.getClassLoader().getResource("test_druid_all_types"))
+      .getFile());
+  private static final Schema ALL_TYPES_TEST_SCHEMA = new Schema.SchemaBuilder().addSingleValueDimension("__time", FieldSpec.DataType.LONG)
+      .addSingleValueDimension("stringval", FieldSpec.DataType.STRING)
+      .addMultiValueDimension("tags", FieldSpec.DataType.STRING)
+      .addMetric("intnum", FieldSpec.DataType.LONG)
```
snleee: `longnum` ?

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -222,28 +222,28 @@ private void validateColumns(QueryableIndexStorageAdapter adapter) {
       final String columnName = _columnNames.get(i);
       ColumnCapabilities capabilities = adapter.getColumnCapabilities(columnName);
       if (capabilities == null) {
-        LOGGER.warn("Column {} is not in record", columnName);
+        LOGGER.warn("Column %s is not in record", columnName);
```
snleee: For LOGGER, You need to use `Column {} is not in record` for LOGGER.

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -171,8 +171,8 @@ public GenericRow next(GenericRow reuse) {
         Object value = selector.getObject();
         if (value != null && !fieldSpec.isSingleValueField()) {
           // Multi-valued dimensions in Druid are stored as Arrays.ArrayList (this has been checked)
-          Preconditions.checkArgument(value.getClass() == java.util.Arrays.asList().getClass(), "The multi-valued dimension " + columnName +
-              " should be java.util.Arrays$ArrayList, but it is " + value.getClass());
+          Preconditions.checkArgument(value.getClass() == java.util.Arrays.asList().getClass(),
```
snleee: As I mentioned in the previous comment, you need to use `equals()` for object comparison.
Moreover, you can try to use `instanceof` instead of checking class object by calling `getClass()`.
e.g. 
```
Preconditions.checkState(value instanceof List, errorMessage);
```

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -222,28 +222,28 @@ private void validateColumns(QueryableIndexStorageAdapter adapter) {
       final String columnName = _columnNames.get(i);
       ColumnCapabilities capabilities = adapter.getColumnCapabilities(columnName);
```
snleee: let's rename this to `druidColumnCapabilities`

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -171,8 +171,8 @@ public GenericRow next(GenericRow reuse) {
         Object value = selector.getObject();
         if (value != null && !fieldSpec.isSingleValueField()) {
```
snleee: let's rename `fieldSpec` to `pinotFieldSpec`

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
snleee: Sorry. I actually put the same comment above and you added the answer on why you had to keep `_columnNames`. Resolving this.

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+      while (testReader.hasNext()) {
+        Assert.assertTrue(row < records.length);
+
+        GenericRow nextRow = testReader.next();
+        GenericRow transformedRow = defaultTransformer.transform(nextRow);
+
+        for (int col = 0; col < columns.length; col++) {
+          if (records[row][col] != null) {
+            if (transformedRow.getValue(columns[col]) != null && records[row][col] != null) {
+              Assert.assertTrue(transformedRow.getValue(columns[col]).getClass() == records[row][col].getClass(),
```
snleee: Let's use `instanceof`

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+                  String.format("In row {} and column {}, type of value in segment ({}) does not match type of value in test record ({}).",
+                      row, col, transformedRow.getValue(columns[col]).getClass(), records[row][col].getClass()));
+              if (transformedRow.getValue(columns[col]).getClass() == String[].class) {
```
snleee: `instanceof`

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+    try (DruidSegmentRecordReader druidRecordReader = new DruidSegmentRecordReader(TEST_SAMPLE_DATA, TEST_SAMPLE_DATA_SCHEMA);
+        JSONRecordReader jsonRecordReader = new JSONRecordReader(TEST_SAMPLE_DATA_JSON, TEST_SAMPLE_DATA_SCHEMA)) {
+      int numRecords = 0;
+      while (druidRecordReader.hasNext()) {
+        assertTrue(jsonRecordReader.hasNext());
+        numRecords++;
+
+        GenericRow druidRecord = defaultTransformer.transform(druidRecordReader.next());
+        GenericRow jsonRecord = defaultTransformer.transform(jsonRecordReader.next());
+        assertEquals(druidRecord, jsonRecord);
```
snleee: Let's use a single style. `Assert.assertEquals` or `asserEquals`.

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+      .addMetric("floatnum", FieldSpec.DataType.FLOAT)
+      .build();
+
+  private void testReadRowsWithRecordList(Object[][] records, String[] columns, File dataFile, Schema schema)
+      throws IOException {
+    try (DruidSegmentRecordReader testReader = new DruidSegmentRecordReader(dataFile, schema)) {
+      CompositeTransformer defaultTransformer = CompositeTransformer.getDefaultTransformer(schema);
+
+      List<GenericRow> genericRows = new ArrayList<>();
+      int row = 0;
```
snleee: `rowCount`

contrib/pinot-druid-migration/src/test/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReaderTest.java
```diff
@@ -0,0 +1,180 @@
...
+      while (testReader.hasNext()) {
+        Assert.assertTrue(row < records.length);
+
+        GenericRow nextRow = testReader.next();
```
snleee: `currentRow`

snleee:
@dbadapon Travis test fails due to some weird error. `Commit range is invalid.` Can you try to fix this as well?

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSegmentConverter.java
dbadapon: I saw it in a class that i was referencing so I guess i'll just take it out

snleee: I think that this is not a good pattern. Put the following to the main function

```
public static void main(String[] args) throws Exception {
}
```

contrib/pinot-druid-migration/HEADER
```diff
@@ -0,0 +1,16 @@
+Licensed to the Apache Software Foundation (ASF) under one
```
snleee: This file may not needed anymore since we removed `contrib` from checking license header.

pom.xml
```diff
@@ -322,15 +322,15 @@
...
-          <exclusion>
-            <groupId>log4j</groupId>
-            <artifactId>log4j</artifactId>
-          </exclusion>
-          <exclusion>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-log4j12</artifactId>
-          </exclusion>
-        </exclusions>
+        <exclusion>
```
snleee: revert the indentation. Previous indentation looks correct.
```
<exclusions>
  <excluston>
  </exclusion>
</exclusions>
```

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSegmentConverter.java
```diff
@@ -31,9 +31,9 @@
 public class DruidToPinotSegmentConverter {
   private static String _pinotSchemaFilePath;
```
snleee: We usually use either `private` (instance variable) or `private static final` (constant). Let's actually remove these variables.

We can do either of the followings:
1. Do everything in `main()`
2. Add variables to `convertSegment()` function

pinot-tools/pom.xml
```diff
@@ -123,6 +118,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
snleee: It seems that `pinot-orc`was included in the original code. Let's not modify this part and revert back to the original copy.

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSchemaGenerator.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.druid.segment.column.ValueType;
+import org.apache.druid.segment.data.Indexed;
+import org.apache.pinot.common.data.FieldSpec;
+import org.apache.pinot.common.data.Schema;
+
+/**
+ * The DruidToPinotSchemaConverter is a tool that takes a Druid segment and creates a Pinot schema with the
+ * segment's information.
+ */
+public class DruidToPinotSchemaGenerator {
```
snleee: The input for schema converter should not be `druid segment`. It should be `ingestion spec` https://druid.apache.org/docs/latest/ingestion/index.html#ingestion-specs

pom.xml
```diff
@@ -330,7 +330,7 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
           </exclusion>
-        </exclusions>
+      </exclusions>
```
snleee: Revert this change. (Previous indentation is correct)

pinot-tools/pom.xml
```diff
@@ -123,6 +123,10 @@
...
```
snleee: I commented 2 times on this change. Those comments got marked as resolved without any change! We `should not` touch anything on `pinot-tools/pom.xml`. Please revert the change back.

dbadapon: So weird, I specifically remember fixing this exact thing. I'll go back and fix it again, hopefully it will actually work this time.

contrib/pinot-druid-migration/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Build and Usage
+To build the project:
+
+```
+mvn clean install -DskipTests
+```
+
+
+This will create `druid-to-pinot-schema-generator-jar-with-dependencies.jar` and
```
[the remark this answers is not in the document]
dbadapon: Currently working on the single jar.

pom.xml
dbadapon: I'm looking at the file and I don't see anything wrong with the indentation... what is happening?

pinot-tools/pom.xml
dbadapon: I think the same thing is happening here as above. I'm looking at the file and there does not seem to be anything wrong with the indentation, and I'm not sure how this diff was created in the first place.

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/DruidToPinotSchemaGenerator.java
dbadapon: It turns out that the Druid ingestion spec does not have enough info to generate a Pinot schema from it. It does not specify which columns are single/multi-valued, so I had to go ahead and use this original method anyway. I only create an `adapter` from the segment, and I don't have to read any rows.

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/hadoop/DruidToPinotSegmentConverterHadoopJob.java
```diff
@@ -0,0 +1,438 @@
...
+      @Override
+      public boolean accept(File pathname)
+      {
+        return pathname.exists()
+            && pathname.isDirectory();
+      }
+    };
+
+    if (file.getAbsoluteFile().exists()) {
+      File untaredInput = new File(file.getParentFile().getPath(), "untaredInput");
```
jenniferdai: super nitpick but untarred*

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/hadoop/DruidToPinotSegmentConverterHadoopJob.java
```diff
@@ -0,0 +1,438 @@
...
+
+    _logger.info("*********************************************************************");
+    _logger.info("Input Pattern: {}", _inputPattern);
+    _logger.info("Output Directory: {}", _outputDir);
+    _logger.info("Staging Directory: {}", _stagingDir);
+    _logger.info("Raw Table Name: {}", _rawTableName);
+    _logger.info("*********************************************************************");
+  }
+
+  protected boolean isDataFile(URI s) {
```
jenniferdai: Maybe we can add a quick unit test for this method since it's a little tricky

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/data/readers/DruidSegmentRecordReader.java
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  @Override
+  public Schema getSchema() {
+    return _pinotSchema;
+  }
+
+  @Override
+  public void close() {
+    _index.close();
```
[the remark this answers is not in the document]
dbadapon: I don't think it needs to be closed?

contrib/pinot-druid-migration/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+    </plugins>
+  </build>
+
+  <!-- dependencyManagement section added to resolve dependency convergence issues with Druid -->
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-all</artifactId>
+        <version>4.1.30.Final</version>
```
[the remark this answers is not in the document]
dbadapon: oh wow... I don't know why we didn't think of this earlier lol

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/hadoop/DruidToPinotSegmentConverterHadoopJob.java
```diff
@@ -0,0 +1,438 @@
...
+    _depsJarDir = getPathFromProperty(JobConfigConstants.PATH_TO_DEPS_JAR);
+    _schemaFile = getPathFromProperty(JobConfigConstants.PATH_TO_SCHEMA);
+    _defaultPermissionsMask = _properties.getProperty(JobConfigConstants.DEFAULT_PERMISSIONS_MASK);
+
+    // Optional push location and table parameters. If set, will use the table config and schema from the push hosts.
+    String pushHostsString = _properties.getProperty(JobConfigConstants.PUSH_TO_HOSTS);
+    String pushPortString = _properties.getProperty(JobConfigConstants.PUSH_TO_PORT);
+    if (pushHostsString != null && pushPortString != null) {
+      _pushLocations =
+          PushLocation.getPushLocations(StringUtils.split(pushHostsString, ','), Integer.parseInt(pushPortString));
```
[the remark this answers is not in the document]
dbadapon: I don't see it in base segment job

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/hadoop/DruidToPinotSegmentConverterHadoopJob.java
```diff
@@ -0,0 +1,438 @@
...
+    _logger.info("Output Directory: {}", _outputDir);
+    _logger.info("Staging Directory: {}", _stagingDir);
+    _logger.info("Raw Table Name: {}", _rawTableName);
+    _logger.info("*********************************************************************");
+  }
+
+  protected boolean isDataFile(URI s) {
+    if (_properties.getProperty(JobConfigConstants.RECORD_READER_PATH) != null) {
+      return true;
+    }
```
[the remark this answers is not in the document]
dbadapon: Yeah I'm actually not sure why this is here... copypasta

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/hadoop/DruidToPinotSegmentConverterHadoopJob.java
dbadapon: Do I just add another test class to the test folder?

jenniferdai: yes :) just for this method if possible

contrib/pinot-druid-migration/src/main/java/org/apache/pinot/druid/tools/hadoop/DruidToPinotSegmentConverterHadoopJob.java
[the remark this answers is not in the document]
jenniferdai: hmm maybe my code is diff but i see it in the constructor
protected BaseSegmentJob(Properties properties) {
    _properties = properties;
    _conf = new Configuration();
    setConf(_conf);
    Utils.logVersions();
    logProperties();

    // Optional push location and table parameters. If set, will use the table config and schema from the push hosts.
    String pushHostsString = _properties.getProperty(JobConfigConstants.PUSH_TO_HOSTS);
    String pushPortString = _properties.getProperty(JobConfigConstants.PUSH_TO_PORT);
    if (pushHostsString != null && pushPortString != null) {
      _pushLocations =
          PushLocation.getPushLocations(StringUtils.split(pushHostsString, ','), Integer.parseInt(pushPortString));
    } else {
      _pushLocations = null;
    }

    _rawTableName = Preconditions.checkNotNull(_properties.getProperty(JobConfigConstants.SEGMENT_TABLE_NAME));
  }

jenniferdai:
Can you rebase on master? I will continue reviewing in the meantime

PR #4739, opened 2019-10-23
Fix state change when consumption is stopped

mcvsubbu:
When a server stops consumption due to any exception or error, it invokes
a controller operation to mark itself as being OFFLINE. It may happen that
this condition takes place right after a segment is marked as ONLINE,
but the consumer thread on the server is still alive and waiting for the
completion response (of KEEP/COMMIT/DISCARD, etc.)

While updating the ideal state, the controller should ensure that the
Helix state of the segment is CONSUMING and not anything else. If it
is not CONSUMING, then we don't want to error out, but just quietly report
back success to the server, and let it do its thing.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -604,11 +604,12 @@ public class PinotLLCRealtimeSegmentManager {
       HelixHelper.updateIdealState(_helixManager, realtimeTableName, idealState -> {
         assert idealState != null;
         Map<String, String> stateMap = idealState.getInstanceStateMap(segmentName);
-        if (stateMap != null) {
-          String state = stateMap.get(instanceName);
-          if (state != null && state.equals(RealtimeSegmentOnlineOfflineStateModel.CONSUMING)) {
-            stateMap.put(instanceName, RealtimeSegmentOnlineOfflineStateModel.OFFLINE);
-          }
+        String state = stateMap.get(instanceName);
+        if (RealtimeSegmentOnlineOfflineStateModel.CONSUMING.equals(state)) {
+          stateMap.put(instanceName, RealtimeSegmentOnlineOfflineStateModel.OFFLINE);
+        } else {
+          LOGGER.info("Segment {} in state {} when trying to register consumption stop from {}",
+              segmentName, state, instanceName);
         }
         return idealState;
       }, RetryPolicies.exponentialBackoffRetryPolicy(10, 500L, 1.2f), true);
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
@@ -589,7 +589,7 @@ public class PinotLLCRealtimeSegmentManager {
 
   /**
    * An instance is reporting that it has stopped consuming a topic due to some error.
-   * Mark the state of the segment to be OFFLINE in idealstate.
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -603,9 +603,15 @@ public void segmentStoppedConsuming(LLCSegmentName llcSegmentName, String instan
     try {
       HelixHelper.updateIdealState(_helixManager, realtimeTableName, idealState -> {
         assert idealState != null;
-        idealState.getInstanceStateMap(segmentName).put(instanceName, RealtimeSegmentOnlineOfflineStateModel.OFFLINE);
+        Map<String, String> stateMap = idealState.getInstanceStateMap(segmentName);
+        if (stateMap != null) {
```
Jackie-Jiang: The `stateMap` should never be null. We should not silently return if it is null.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -603,9 +603,15 @@ public void segmentStoppedConsuming(LLCSegmentName llcSegmentName, String instan
...
+          String state = stateMap.get(instanceName);
+          if (state != null && state.equals(RealtimeSegmentOnlineOfflineStateModel.CONSUMING)) {
```
Jackie-Jiang: `RealtimeSegmentOnlineOfflineStateModel.CONSUMING.equals(state)`

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
```diff
@@ -603,9 +603,15 @@ public void segmentStoppedConsuming(LLCSegmentName llcSegmentName, String instan
...
+            stateMap.put(instanceName, RealtimeSegmentOnlineOfflineStateModel.OFFLINE);
+          }
```
Jackie-Jiang: Please add some log such as cannot find instance in CONSUMING state for debugging purpose

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: Agreed, the statemap can be null only when the segment has somehow been removed from the table (a bug elsewhere). I can log something to that effect.
We should always return "DONE" to the server, otherwise the server may retry the operation, to no effect. So, no point throwing exceptions here, if that is what you mean

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: sure

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: ok

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/realtime/PinotLLCRealtimeSegmentManager.java
mcvsubbu: I decided to throw NPE here if that happens. That is a nasty bug, and we dont want to be checking for nulls everywhere.

PR #4740, opened 2019-10-23
[New Transformation function] Presto compatible DateTrunc

agrawaldevesh:
This will only be called by Presto connector (https://github.com/prestodb/presto/pull/13504) and it has identically semantics
to presto's SQL date_trunc, albeit with a timezone specialization.

This is needed so that the presto's date_trunc invocations can be
faithfully translated as is to this new function. Its a new function so
that it is trivial to roll out to harmlessly without a lot of regression
testing.

Without this function, we cannot handle timezones nor week truncations
in the existing pinot's dateTimeConvert function.

It basically copies the PrestoDB code:
https://github.com/prestodb/presto/blob/master/presto-main/src/main/java/com/facebook/presto/operator/scalar/DateTimeFunctions.java,
(specializations of date_trunc for TIMESTAMP and
TIMESTAMP_WITH_TIME_ZONE

I am even checking in the zone-index.properties used by presto to ensure that
even the time zones are 1:1 b/w presto and this function. (sync'd to the
latest prestodb repo)

Understanding this UDF requires knowledge of the joda-time API. I am not
documenting this heavily since it is a copy of the Presto UDF.


agrawaldevesh:
@snleee @mayankshriv can you please review this PR. 

The context is that we would like presto's use of the date_trunc (using PR https://github.com/prestodb/presto/pull/13504 in the prestodb) to be translated to this new presto compatible date_trunc I am adding in pinot. This will ensure faithful translation supporting both timezones and week truncations. 

Currently the Pinot dateTimeConvert function truncates to the week starting at Thursday and that is incorrect from presto's standpoint.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTrunc.java
```diff
@@ -0,0 +1,204 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Locale.ENGLISH;
+import static org.apache.pinot.core.operator.transform.function.TimeZoneKey.MAX_TIME_ZONE_KEY;
+import static org.apache.pinot.core.operator.transform.function.TimeZoneKey.getTimeZoneKeys;
+
+public class PrestoDateTrunc
```
kishoreg: why do we need the Presto prefix? can we not use this without Presto

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
```diff
@@ -0,0 +1,262 @@
...
+ * // Assume that the presto session timezone is 'Europe/Berlin'
+ * date_trunc(from_unixtime(time_in_seconds), 'week') to prestoDateTrunc(time_in_seconds, 'SECONDS', 'Europe/Berlin', 'week')
+ * <p>
+ * // timestamp with timezone (override whatever the default session timezone is)
+ * date_trunc(from_unixtime(time_in_seconds, 'Americas/New_York'), 'week') to prestoƒDateTrunc(time_in_seconds, 'SECONDS', 'Americas/New_York', 'week')
+ * </li>
+ * </ul>
+
+ */
+public class PrestoDateTruncFunction
```
Jackie-Jiang: Is this the same as `PrestoDateTrunc.java`? I believe `PrestoDateTrunc.java` should be deleted.
+1 on removing `Presto` from the function name, and rename it to `DateTruncTransformFunction`

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
```diff
@@ -0,0 +1,262 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import static java.util.Locale.ENGLISH;
+import static org.apache.pinot.core.operator.transform.function.TimeZoneKey.MAX_TIME_ZONE_KEY;
+import static org.apache.pinot.core.operator.transform.function.TimeZoneKey.getTimeZoneKeys;
+
+/**
+ * The <code>PrestoDateTruncFunction</code> class implements the sql compatible date_trunc function for TIMESTAMP type.
```
Jackie-Jiang: Can you please add proper indentation to this javadoc?

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/TimeZoneKey.java
```diff
@@ -0,0 +1,262 @@
+package org.apache.pinot.core.operator.transform.function;
```
Jackie-Jiang: Move this line after the header

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
```diff
@@ -0,0 +1,262 @@
...
+import org.joda.time.field.ScaledDurationField;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import static java.util.Locale.ENGLISH;
```
Jackie-Jiang: (Code style) avoid static import

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/TransformFunctionFactory.java
```diff
@@ -58,6 +58,8 @@ private TransformFunctionFactory() {
           put(TimeConversionTransformFunction.FUNCTION_NAME.toLowerCase(), TimeConversionTransformFunction.class);
           put(DateTimeConversionTransformFunction.FUNCTION_NAME.toLowerCase(),
               DateTimeConversionTransformFunction.class);
+          put(PrestoDateTruncFunction.FUNCTION_NAME.toLowerCase(),
```
Jackie-Jiang: (nit) one line

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
```diff
@@ -0,0 +1,262 @@
...
+import org.joda.time.DurationField;
+import org.joda.time.DurationFieldType;
+import org.joda.time.chrono.ISOChronology;
+import org.joda.time.field.DividedDateTimeField;
+import org.joda.time.field.OffsetDateTimeField;
+import org.joda.time.field.ScaledDurationField;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
```
Jackie-Jiang: (Code style) do not annotate non-null, only annotate nullable

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    TransformFunction firstArgument = arguments.get(0);
+    if (firstArgument instanceof LiteralTransformFunction || !firstArgument.getResultMetadata().isSingleValue()) {
+      throw new IllegalArgumentException(
+              "The first argument of prestoDateTrunc transform function must be a single-valued column or a transform function");
+    }
+    _mainTransformFunction = firstArgument;
+    _resultMetadata = LONG_SV_NO_DICTIONARY_METADATA;
+    String inputTimeUnitStr = ((LiteralTransformFunction) arguments.get(1)).getLiteral().toUpperCase(ENGLISH);
```
Jackie-Jiang: (nit) Use `toUpperCase()` (default locale) for readability and consistency. Same for line 238 and 240

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
```diff
@@ -0,0 +1,262 @@
...
+    try {
+      _inputTimeUnit = TimeUnit.valueOf(inputTimeUnitStr);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("Invalid input time unit " + _inputTimeUnit, e);
```
Jackie-Jiang: _inputTimeUnit will always be null here. No need to catch the exception, directly throw it out is fine

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
```diff
@@ -0,0 +1,262 @@
...
+      case "quarter":
+        return QUARTER_OF_YEAR.getField(chronology);
+      case "year":
+        return chronology.year();
+    }
+    throw new IllegalArgumentException("'" + unitString + "' is not a valid Timestamp field");
+  }
+
+  // Usage:
+  // prestoDateTrunc(input, 'tz', 'unit') -> output.
```
Jackie-Jiang: 4 arguments are expected right?

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTrunc.java
agrawaldevesh: This has nothing to do with Presto and the function can be used independantly. 

I just wanted to distinguish it with the existing dateTimeConvert since it is not compatible with that. 

dateTimeConvert truncates the week to Thursday (5pm PT perhaps !) whereas this function truncates to Monday midnight in the specified timezone. 

What name would you suggest ? How about PrestoCompatibleDateTrunc or SqlCompatibleDateTrunc ? Although I imagine there is nothing "Sql standard" about date truncation.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
agrawaldevesh: Ah ... Sure. This is where the presto and pinot code conventions differ :-). Will do.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/TransformFunctionFactory.java
agrawaldevesh: It crosses the line limit and the checked in codestyle mandates two lines here :)

agrawaldevesh:
Thanks for the review @Jackie-Jiang and @kishoreg. Updated per your feedback.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
agrawaldevesh: Sorry, I didn't follow. I followed the instructions given here: https://pinot.readthedocs.io/en/latest/dev_env.html. 

Can you let me know if the new indentation matches the requirement. Thanks. It passed checkstyle etc.

agrawaldevesh:
I am not sure what lead to the massive decrease in code coverage. The report looks fishy and lists several unrelated files. I did add unit tests for these new classes :). Can you help me figure out if the code coverage decrease is legit or not. Thanks !

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/PrestoDateTruncFunction.java
Jackie-Jiang: I was referring to the indentation of the javadoc. Right now it is a little bit hard to read. You can check the javadoc in SelectionOperatorService.java as a reference.

Jackie-Jiang:
@agrawaldevesh You don't need to worry about the code coverage. It could be that some coverage files are not sent to the server.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/DateTruncFunction.java
```diff
@@ -0,0 +1,242 @@
...
+ * <li>
+ * // Assume that the presto session timezone is 'Europe/Berlin'
+ * date_trunc(from_unixtime(time_in_seconds), 'week') to dateTrunc(time_in_seconds, 'SECONDS', 'Europe/Berlin', 'week')
+ * <p>
+ * // timestamp with timezone (override whatever the default session timezone is)
+ * date_trunc(from_unixtime(time_in_seconds, 'Americas/New_York'), 'week') to dateTrunc(time_in_seconds, 'SECONDS', 'Americas/New_York', 'week')
+ * </li>
+ * </ul>
+ */
+public class DateTruncFunction extends BaseTransformFunction {
```
Jackie-Jiang: Rename to DataTruncTransformFunction. We want to differentiate AggregationFunction and TransformFunction

agrawaldevesh:
@Jackie-Jiang Thanks for your comments earlier. I have incorporated all of your feedback. Thank you !

pinot-core/src/test/java/org/apache/pinot/core/operator/transform/function/DateTruncTransformFunctionTest.java
```diff
@@ -0,0 +1,172 @@
...
+      driver.build();
+
+      IndexSegment indexSegment = ImmutableSegmentLoader.load(new File(indexDirPath, segmentName), ReadMode.heap);
+      Set<String> columnNames = indexSegment.getPhysicalColumnNames();
+      HashMap<String, DataSource> dataSourceMap = new HashMap<>(columnNames.size());
+      for (String columnName : columnNames) {
+        dataSourceMap.put(columnName, indexSegment.getDataSource(columnName));
+      }
+
+      ProjectionBlock projectionBlock = new ProjectionOperator(dataSourceMap,
```
siddharthteotia: I think it would be better to test this transform function using overall query execution code path as well in addition to this test. Here you have to build the expression tree, operator, projection block etc and this is something that query execution code already does. So may be we should also write a test extending BaseQueriesTest similar to how other query execution tests are written.

siddharthteotia: That way we will know that this transform function fits into query execution code flow

pinot-core/src/test/java/org/apache/pinot/core/operator/transform/function/DateTruncTransformFunctionTest.java
```diff
@@ -0,0 +1,172 @@
...
+      Assert.assertEquals(transformFunction.getName(), DateTruncTransformFunction.FUNCTION_NAME);
+      long[] longValues = transformFunction.transformToLongValuesSV(projectionBlock);
+      Assert.assertEquals(longValues[0], expected);
+    } finally {
+      FileUtils.deleteDirectory(new File(indexDirPath));
+    }
+  }
+
+  @Test
+  public void testPrestoCompatibleDateTimeConversionTransformFunction() throws Exception {
```
siddharthteotia: Since dateTrunc() function can accept transform function as an argument, tt would be good to add tests that verify the nested nature. You can look at TransformQueriesTest for example.

pinot-core/src/test/java/org/apache/pinot/core/operator/transform/function/DateTruncTransformFunctionTest.java
agrawaldevesh: Thanks for your comments and looking over this diff.

I am wondering if there is an existing test framework/example I can "copy" ? I think it is a bit onerous for UDF authors to develop whole new testing frameworks. So if you can point me in the right direction with what I should base this test on, I will be happy to author it and amend the diff.

pinot-core/src/test/java/org/apache/pinot/core/operator/transform/function/DateTruncTransformFunctionTest.java
agrawaldevesh: Oh cool. So I will add the test for this new UDF also there.

pinot-core/src/test/java/org/apache/pinot/core/operator/transform/function/DateTruncTransformFunctionTest.java
siddharthteotia: Yes, please look at TransformQueriesTest and how it extends BaseQueriesTest. The test:

(1) Creates a sample data set
(2) Build segment(s) out of the dataset
(3) Runs inner segment level query
(4) Runs inter segment (and inter server) query.

Probably (4) is not necessary for this test. 

You can also look at DistintQueriesTest.java in the same directory as TransformQueriesTest.

agrawaldevesh:
Hi @siddharthteotia ... take a look at the PR now. I added the requested e2e unit test and also added documentation. So this UDF is now "unhidden" and can be used by anyone. 

I believe this should allay your concerns and allow this PR to be merged in. Let me know if you think something else needs to be done here.

Thanks for the review !

siddharthteotia:
> Hi @siddharthteotia ... take a look at the PR now. I added the requested e2e unit test and also added documentation. So this UDF is now "unhidden" and can be used by anyone.
> 
> I believe this should allay your concerns and allow this PR to be merged in. Let me know if you think something else needs to be done here.
> 
> Thanks for the review !

Thanks, @agrawaldevesh. LGTM

docs/pql_examples.rst
```diff
@@ -246,6 +246,19 @@ Supported transform functions
    format ``1:SECONDS:EPOCH`` with a granularity of ``15:MINUTES`` (*i.e.* nearest 15-minute value lower than the value
    of ``date`` column.
 
+``DATETRUNC``
+   (Presto) SQL compatible date truncation, equivalent to the Presto function `date_trunc
+<https://mode.com/blog/date-trunc-sql-timestamp-function-count-on>`_. Takes 4 or 5 arguments, converts the value into another date time format, and buckets time based on the given time granularity. It is timezone aware. If 4 arguments are given, the resulting number is in ``Milliseconds``, otherwise the fifthe argument is the output time granularity.
```
Jackie-Jiang: Type `fifth`

docs/pql_examples.rst
```diff
@@ -246,6 +246,19 @@ Supported transform functions
    format ``1:SECONDS:EPOCH`` with a granularity of ``15:MINUTES`` (*i.e.* nearest 15-minute value lower than the value
    of ``date`` column.
 
+``DATETRUNC``
+   (Presto) SQL compatible date truncation, equivalent to the Presto function `date_trunc
```
Jackie-Jiang: Presto, PostgreSQL and Redshift all supports DATE_TRUNC, and all of them take two arguments: DATE_TRUNC('unit', timestamp). What is the specific reason why we need 4 or 5 arguments? Should we also add support for 2 arguments, where timestamp is millisSinceEpoch?

agrawaldevesh:
HI @Jackie-Jiang .. 

First some context: This diff didn't start its life out (and it is not my intention) to be a general time truncation UDF. Instead, it was meant to be called only by the presto-pinot connector and thus generate a value that presto can understand natively. Thus it returned "long milliseconds since epoch in UTC" because that is what presto natively understands as time.

The typical way for this function to be invoked is:

Pinot lacks a proper timestamp type. A type that can encode other bits of information associated with it: the timezone and the granularity. 

Presto achieves this with two timestamp types: 

* timestamp -> Stores the milliseconds since epoch and is always in UTC
* timestamp_with_tz -> Stores the 'timestamp above' and a couple of bits to denote the timezone.

Given this:
* date_trunc('hour', timestamp) -> returns a timestamp
* date_trunc('hour', timestamp_with_tz) -> returns a timestamp with timezone.

They also have functions to change timezones, ie to go from a timestamp to a timestamp_with_tz and to change the timezone burned into a timestamp_with_tz.

In the absence of a proper pinot type representing time or timezones, we need to inline that into the return type and thus have arguments to configure them. For example, we can have additional input arguments saying what the output timezone and the output time granularity are in. 

We similarly need additional arguments to specify what the input granularity and timezone are. 

And thus the need for 6 or so arguments: 
* Input long value
* Truncation interval: hour, day, quarter etc
* Input granularity (ms or seconds etc)
* Input timezone
* Output granularity. (When called from Presto this is always in Milliseconds)
* Output timezone. (When called from Presto this is always UTC)

(For full disclosure, I don't think we should allow the output timezone to be changed. What would a user do with that ? Because now they need to remember what the output timezone is)

What do you think ? Should we make this function be fully general with these 6 arguments or should we keep it very custom and specific to Presto and thereby reduce it down to three arguments:

* Input long value
* Truncation interval: hour, day, quarter etc
* ~~Input granularity (Hardcoded to seconds when called via Presto)~~
* Input timezone
* ~~Output granularity. (Hardcoded to milliseconds when called via Presto)~~
* ~~Output timezone. (When called from Presto this is always UTC)~~

I like your idea of thinking about this more fully and introducing more time UDFs like timezone conversion, or something to print timestamps with timezones. But I think that is widening the scope of this PR: I need something minimal such that our presto-pinot customers can translate their presto expressions like below to pinot:
```
date_trunc('month', from_unixtime(ts_in_millis / 1000, 'America/Los_Angeles'))
```

If it unblocks this PR: I can create an issue for better time handling in Pinot and narrow this PR to be very Presto specific and remove this function from the Pinot documentation (ie let it remain hidden). 

Please let me know what would you need to see in this PR such that it can be merged in.

Jackie-Jiang:
@agrawaldevesh Thanks for the explanation. 
IMO, `date_trunc()` should not have any knowledge of the timezone, but only truncate the time based on the given timestamp. Given the input value, `date_trunc()` should be able to figure out the input granularity automatically, and truncate it to the specified truncation interval. With these, we can achieve the standard `date_trunc()` semantic with only 2 arguments, e.g. `date_trunc('DAYS', timestamp)`.
Then the missing part is a timezone conversion function from arbitrary timezone & value to a timestamp in another timezone. The time value can be of numeric type or string type (e.g. DateTimeFormat).
With these 2 functions de-coupled and chain-able, it becomes much more flexible. We can support timestamp/DateTimeFormat input in any timezone, and return truncated timestamp in another timezone. Also, we have the standard semantic for `date_trunc()` function.
What do you think? Does this proposal meet all the requirements for the Presto connector?

agrawaldevesh:
HI @Jackie-Jiang ... I understand that splitting up the functions and making them be composable is flexible and great. But I am afraid that the approach you have pointed, won't work:

In the library joda-time we are using: Truncation depends on the timezone. Which makes sense: How do you truncate to a quarter, when the quarter starts at different timestamps across the world ?

We cannot work around the fact that Pinot does not have a "Timestamp with TZ" type: so there is no way to known of the input timezone without specifying that as an argument. So at the minimum we need to have three arguments: Namely the input timezone, the truncation granularity 'hour'/'week' etc, and the input value. 

We need the timezone from truncation: either as an input field or as a whole new type. The latter is obviously a much bigger change :-)

Jackie-Jiang:
@agrawaldevesh I think I might have missed something? Within the same timezone (both input and output value in the same timezone), will `date_trunc` always return the same value?
E.g. 
- Input value in timezone1, truncate the time in timezone1, output value in timezone 2: `timezone_convert(date_trunc('DAYS', input), timezone1, timezone2)`
- Input value in timezone1, output value in timezone2, truncate the time in timezone2:
`date_trunc('DAYS', timezone_convert(input, timezone1, timezone2))`

If this does not work, and we need to have the timezone info, then I agree we should put at as an input field. We can make it optional and default to UTC.

agrawaldevesh:
> * Input value in timezone1, truncate the time in timezone1, output value in timezone 2: `timezone_convert(date_trunc('DAYS', input), timezone1, timezone2)`

The problem is that "input" is just a long. How would date_trunc know that 'input' is in timezone1 ?

Jackie-Jiang:
@agrawaldevesh I see where I was missing. So both the input and output values are unix time (in UTC), and we just want to truncate the time based on some timezone right?
Then the current approach is good. My suggestion would be making timezone (default UTC), input granularity (default millis) and output granularity (default millis) optional. The function can take 2-5 arguments, and by default using UTC to truncate in millis.

Jackie-Jiang:
Also, based on the input value, input granularity can also be automatically figured out

agrawaldevesh:
Hi @Jackie-Jiang ... thanks for discussing the shape of this function with me on Slack. I have updated it per what we discussed:

The new argument format is: 
* truncation granularity
* Input value that is always since UTC epoch
* input granularity: milliseconds, seconds etc
* optional: Truncation TZ defaulting to UTC
* optional: Output granularity defaulting to Input granularity

Thanks for helping me nail this interface.

pinot-core/src/main/java/org/apache/pinot/core/operator/transform/function/DateTruncTransformFunction.java
```diff
@@ -0,0 +1,261 @@
...
+    String outputTimeUnitStr =
+        arguments.size() >= 5 ? ((LiteralTransformFunction) arguments.get(4)).getLiteral().toUpperCase()
+            : inputTimeUnitStr;
+    TimeZoneKey timeZoneKey = TimeZoneKey.getTimeZoneKey(timeZone);
+
+    _field = getTimestampField(DateTimeZoneIndex.getChronology(timeZoneKey), unit);
+    _resultMetadata = LONG_SV_NO_DICTIONARY_METADATA;
+    _outputTimeUnit = TimeUnit.valueOf(outputTimeUnitStr);
+
+    if (arguments.size() > 5) {
```
Jackie-Jiang: This can be combined with the first check in line 130 `Preconditions.checkArgument(numArguments >= 3 && numArguments <= 5, "...")` and provide an example of the expected arguments in the exception message

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -143,6 +149,54 @@ public void testTransformWithAvgInnerSegment()
...
+      row.putField(D1, columns[0]);
+      row.putField(M1, columns[1]);
+      row.putField(M2, columns[2]);
+      row.putField(M3, columns[3]);
+      row.putField(M4, columns[4]);
+      row.putField(TIME, ThreadLocalRandom.current().nextLong(start, start + 5000));
+      rows.add(row);
+    }
+
+    try {
```
Jackie-Jiang: This try is redundant, `finally` can be applied to the `try` in next line

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -143,6 +149,54 @@ public void testTransformWithAvgInnerSegment()
...
+      try (final RecordReader recordReader = new GenericRowRecordReader(rows, _schema)) {
+        createSegment(_schema, recordReader, SEGMENT_NAME_1, TABLE_NAME);
+        final ImmutableSegment segment = loadSegment(SEGMENT_NAME_1);
+        _indexSegments.add(segment);
+
+        String query =
+            "SELECT COUNT(1) FROM foo GROUP BY DATETRUNC('week', ADD(SUB(DIV(T, 1000), INT_COL2), INT_COL2), 'SECONDS', 'Europe/Berlin')";
```
Jackie-Jiang: `COUNT(*)` as we don't support accessing column with index

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -143,6 +149,54 @@ public void testTransformWithAvgInnerSegment()
...
+        verifyDateTruncationResult(query, rows.size(), "95295600");
+
+        query =
+            "SELECT COUNT(1) FROM foo GROUP BY DATETRUNC('week', DIV(MULT(DIV(ADD(SUB(T, 5), 5), 1000), INT_COL2), INT_COL2), 'SECONDS', 'Europe/Berlin', 'MILLISECONDS')";
```
Jackie-Jiang: `COUNT(*)`

pinot-core/src/test/java/org/apache/pinot/queries/TransformQueriesTest.java
```diff
@@ -143,6 +149,54 @@ public void testTransformWithAvgInnerSegment()
...
+      rows.add(row);
+    }
+
+    try {
+      try (final RecordReader recordReader = new GenericRowRecordReader(rows, _schema)) {
+        createSegment(_schema, recordReader, SEGMENT_NAME_1, TABLE_NAME);
+        final ImmutableSegment segment = loadSegment(SEGMENT_NAME_1);
+        _indexSegments.add(segment);
+
+        String query =
```
Jackie-Jiang: You might want to add a query without timezone argument

PR #4741, opened 2019-10-24
Make the download url of a realtime LLC segment as a list of urls.

chenboat:
As planned in the design in [deep storage by-pass for realtime ingestion](https://cwiki.apache.org/confluence/display/PINOT/By-passing+deep-store+requirement+for+Realtime+segment+completion), the PR makes the download url of a LLC segments as a list of uris instead of one single uris.

The high level rationale is that Pinot servers can then download the segment from a list of candidate uris (deep storage or peer servers). This PR is backward compatible for the existing codes.

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/metadata/segment/LLCRealtimeSegmentZKMetadata.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/metadata/segment/LLCRealtimeSegmentZKMetadata.java
@@ -20,6 +20,7 @@ package org.apache.pinot.common.metadata.segment;
 
 import java.util.Map;
 import org.apache.helix.ZNRecord;
+import org.apache.pinot.common.utils.StringUtil;
 
 import static org.apache.pinot.common.utils.EqualityUtils.hashCodeOf;
 import static org.apache.pinot.common.utils.EqualityUtils.isEqual;
@@ -36,8 +37,8 @@ public class LLCRealtimeSegmentZKMetadata extends RealtimeSegmentZKMetadata {
   private long _startOffset;
   private long _endOffset;
   private int _numReplicas;
-
-  private String _downloadUrl = null;
+  // A comma separated list of urls to download the segment.
+  private String _downloadUrls = null;
 
   public LLCRealtimeSegmentZKMetadata() {
     super();
@@ -48,7 +49,7 @@ public class LLCRealtimeSegmentZKMetadata extends RealtimeSegmentZKMetadata {
     _startOffset = Long.valueOf(znRecord.getSimpleField(START_OFFSET));
     _numReplicas = Integer.valueOf(znRecord.getSimpleField(NUM_REPLICAS));
     _endOffset = Long.valueOf(znRecord.getSimpleField(END_OFFSET));
-    _downloadUrl = znRecord.getSimpleField(DOWNLOAD_URL);
+    _downloadUrls = znRecord.getSimpleField(DOWNLOAD_URL);
   }
 
   public long getStartOffset() {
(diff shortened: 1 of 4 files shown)


Jackie-Jiang:
Let's hold on this change. I'm against putting server url into the segment ZK metadata for the following reasons:
- Server should not modify segment ZK metadata, segment ZK metadata should only be touched by controller. We use segment ZK metadata as the segment lock, and if we allow server to modify segment ZK metadata, we might run into certain unknown race conditions.
- Whenever segment moves from one server to another server, it is hard to keep the segment ZK metadata consistent with the segment location. Think of rebalance where all segments might be moved, which will cause multiple modifications on all segment ZK metadata for the table.
- Storing server url along with deep storage url implies server should return the segment in the same format as deep storage, which is not the case. Server keeps segment un-compressed while deep storage stores compressed segment.

Instead, I am proposing the following approach:
- Keep the download uri in segment ZK metadata only points to deep storage
- Introduce a concept of downloading segment from peer, which allow server to download segment from other servers serving the same segment if deep storage is down.
- When deep storage is down, server can read the table's external view, find other ONLINE servers with the segment, and construct download uri based on the ONLINE server instance id.

With the new approach:
- No need to touch segment ZK metadata
- The logic is cleaner. First try the deep storage, if failed, go to the peer download approach
- Segment move can be automatically picked up by the external view
- When downloading the segment from peer, expect the segment to be uncompressed, without extra handling for both compressed and uncompressed segment.
- Offline table can also benefit from the peer download.

mcvsubbu:
@chenboat we briefly discussed this approach proposed by @Jackie-Jiang  and it does have advantages in terms of keeping segment metadata update to one entity (something we had discussed when the proposal was first brought up). It is worth thinking through one more level, and make sure that it satisfies the needs of the project. 
Note that we can still make the metadata field a list, just that servers are not there in the list.

It also does not prevent a PR from changing the server to provide an endpoint to download a segment.

Can you give it some thought?

One thing that I am not very clear about is any race conditions:
- Are there any subtle situations where while transitioning from CONSUMING to ONLINE it is possible that the server with the segment is not yet available in externalview? Can retries solve the problem or do we need more things to work around this?
- Are there similar race conditions during rebalance or while moving completed segments to a different set of servers?

At any rate, it will be good if you can capture the idea in the design doc, thanks.

chenboat:
@Jackie-Jiang @mcvsubbu First thank you for the comments and proposals.  I have updated the design doc with a new[ section](https://cwiki.apache.org/confluence/display/PINOT/By-passing+deep-store+requirement+for+Realtime+segment+completion#By-passingdeep-storerequirementforRealtimesegmentcompletion-SegmentDownload
) to capture the key issue above discussion so far. Please comment there if you have further issues. 

As for Jackie's points, 
 - the update to Zk metadata need NOT be done by servers: server can invoke controller API to do the update. 

- The more important issue is how to discover such server list, I do agree that using External View is better (i remember we briefly discussed about this idea as well). We also need to double check all current download processes using the download_url can fallback to use EV to download from peer servers.

- For segment download format, currently the commit server already Gzips its segment and then upload to the controller as a single file. The new download REST API on the Pinot server uses the same process and also sends such GZip'd file to the server which needs the segment. Therefore Gzip compression is mainly for REST download easiness. I am not sure if there is an easy not to Gzip a segment here. I am not sure if we want to do too much optimization here -- given segment download in LLC is in fact a rare event and the change is further reduced by available deep storage copies.

Jackie-Jiang:
@chenboat I didn't find the new section to discuss this downloading from peer proposal without touching segment ZK metadata but using external view.

- Server call controller API has no difference with server directly modify the segment ZK metadata. The point here is that it is very hard to keep segment ZK metadata consistent with the actual segment locations. Who is in charge of updating the segment ZK metadata? When segment moves from one server to another server, how to update the segment ZK metadata? What if one server is down during the move? What if the segment goes into ERROR state? There are so many things to consider, and it is very hard to cover all scenarios. On contrast, external view already gathered all these information, and it is just way simpler than keep modifying the segment ZK metadata. As long as we provide the same REST API on each server, we can easily construct the download url from the instance id.
- Segment download format is another independent design discussion. Compressing segment is a CPU intensive operation, which could hurt server's performance.

chenboat:
> @chenboat I didn't find the new section to discuss this downloading from peer proposal without touching segment ZK metadata but using external view.

The section is here with title "Segment Download"
https://cwiki.apache.org/confluence/display/PINOT/By-passing+deep-store+requirement+for+Realtime+segment+completion#By-passingdeep-storerequirementforRealtimesegmentcompletion-SegmentDownload

chenboat:
Close this PR based on the discussion. I will follow up with a external view based discovery to download segment from servers.

PR #4745, opened 2019-10-25
[SegmentGeneratorConfig Cleanup] Replace checkTimeColumnValidityDuringGeneration with skipTimeValueCheck

Jackie-Jiang:
- Reverse the config so that the default is false (easier to use)
- Simplify the logic of time value check
- Fix the invalid time value in some tests

NOTE: This commit has backward-incompatible change on the config introduced in #4368 merged in 07/16/2019. In SegmentGeneratorConfig, `checkTimeColumnValidityDuringGeneration` (default true) is replaced with `skipTimeValueCheck` (default false).

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/utils/time/TimeUtils.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/utils/time/TimeUtils.java
@@ -23,12 +23,18 @@ import javax.annotation.Nullable;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.Duration;
+import org.joda.time.Interval;
 import org.joda.time.Period;
 import org.joda.time.format.PeriodFormatter;
 import org.joda.time.format.PeriodFormatterBuilder;
 
 
 public class TimeUtils {
+  public static final long VALID_MIN_TIME_MILLIS = new DateTime(1971, 1, 1, 0, 0, 0, 0, DateTimeZone.UTC).getMillis();
+  public static final long VALID_MAX_TIME_MILLIS = new DateTime(2071, 1, 1, 0, 0, 0, 0, DateTimeZone.UTC).getMillis();
+  public static final Interval VALID_TIME_INTERVAL =
+      new Interval(VALID_MIN_TIME_MILLIS, VALID_MAX_TIME_MILLIS, DateTimeZone.UTC);
+
   private static final String UPPER_CASE_DAYS = "DAYS";
   private static final String UPPER_CASE_DAYS_SINCE_EPOCH = "DAYSSINCEEPOCH";
   private static final String UPPER_CASE_HOURS = "HOURS";
@@ -46,9 +52,6 @@ public class TimeUtils {
   private static final String UPPER_CASE_NANOSECONDS = "NANOSECONDS";
   private static final String UPPER_CASE_NANOS_SINCE_EPOCH = "NANOSSINCEEPOCH";
   private static final String UPPER_CASE_NANOSECONDS_SINCE_EPOCH = "NANOSECONDSSINCEEPOCH";
-
-  private static final long VALID_MIN_TIME_MILLIS = new DateTime(1971, 1, 1, 0, 0, 0, 0, DateTimeZone.UTC).getMillis();
-  private static final long VALID_MAX_TIME_MILLIS = new DateTime(2071, 1, 1, 0, 0, 0, 0, DateTimeZone.UTC).getMillis();
 
   /**
    * Converts a time unit string into {@link TimeUnit}, ignoring case. For {@code null} or empty time unit string,
@@ -108,7 +111,16 @@ public class TimeUtils {
    * <p>The current valid range used is between beginning of 1971 and beginning of 2071.
    */
   public static boolean timeValueInValidRange(long timeValueInMillis) {
-    return (VALID_MIN_TIME_MILLIS <= timeValueInMillis && timeValueInMillis <= VALID_MAX_TIME_MILLIS);
+    return timeValueInMillis >= VALID_MIN_TIME_MILLIS && timeValueInMillis <= VALID_MAX_TIME_MILLIS;
+  }
+
+  /**
+   * Given a time interval, returns true if the interval is between a valid range, false otherwise.
+   * <p>The current valid range used is between beginning of 1971 and beginning of 2071.
+   */
+  public static boolean isValidTimeInterval(Interval timeInterval) {
+    return timeInterval.getStartMillis() >= VALID_MIN_TIME_MILLIS
+        && timeInterval.getEndMillis() <= VALID_MAX_TIME_MILLIS;
   }
 
   /**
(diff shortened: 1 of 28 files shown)


pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
```diff
@@ -112,7 +112,7 @@
   private String _simpleDateFormat = null;
   // Use on-heap or off-heap memory to generate index (currently only affect inverted index and star-tree v2)
   private boolean _onHeap = false;
-  private boolean _checkTimeColumnValidityDuringGeneration = true;
+  private boolean _skipTimeValueCheck = false;
```
siddharthteotia: If the default is false, how will we catch invalid time values during segment generation?  We will have to push this to users (outside Pinot code) who are using SegmentGeneratorConfig. Earlier, the default was true but the check was disabled for realtime and some unit tests where fixing the invalid time values required rewriting a set of Avro files that were used by tests. Still in some other tests, the check was enabled because those tests were doing data generation programmatically (and so easier to control generation of valid time values) as opposed to Avro test files.

I feel the default should be true.

pinot-core/src/test/java/org/apache/pinot/queries/FastHllQueriesTest.java
```diff
@@ -240,7 +240,7 @@ private void buildAndLoadSegment(boolean hasPreGeneratedHllColumns)
     // range. For this test, we first need to fix the input avro data
     // to have the time column values in allowed range. Until then, the check
     // is explicitly disabled
-    segmentGeneratorConfig.setCheckTimeColumnValidityDuringGeneration(false);
+    segmentGeneratorConfig.setSkipTimeValueCheck(true);
```
siddharthteotia: If we really go with default as false (check disabled), we really don't need to do setSkipTimeValueCheck(true) in any of the tests to skip the check.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
```diff
@@ -339,33 +339,42 @@ void writeMetadata()
...
-          properties.setProperty(TIME_UNIT, Preconditions.checkNotNull(config.getSegmentTimeUnit()));
+          startTime = Long.parseLong(startTimeStr);
+          endTime = Long.parseLong(endTimeStr);
+          timeUnit = Preconditions.checkNotNull(config.getSegmentTimeUnit());
         }
       }
+
+      if (!config.isSkipTimeValueCheck()) {
+        Interval timeInterval = new Interval(timeUnit.toMillis(startTime), timeUnit.toMillis(endTime));
+        Preconditions.checkState(TimeUtils.isValidTimeInterval(timeInterval),
```
siddharthteotia: I am not sure how exactly Interval will be printed. I think we should probably make sure the error includes the startTime in millis and endTime in millis to tell the user about invalid times. Also, we should ideally include the valid start/end times in millis from TimeUtils.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
mcvsubbu: +1 for default being true.
The reason this check was added was because the time column check was done when the segment was pushed, and the segment push failed. The idea was to flag the error earlier since the segment is invalid and can never be pushed anyway

Jackie-Jiang: @siddharthteotia @mcvsubbu What I am saying is that for a config, the default value for boolean should be false (easier to use because user usually assumes that), and that is the reason why I changed the variable from `_checkTimeColumnValidityDuringGeneration` to `_skipTimeValueCheck`. I agree we should check time value by default.

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
Jackie-Jiang: It will be printed in human readable format, which is better IMO because user can easily figure out their timestamp is invalid. I'll add start/end time in millis as well.
E.g. `Invalid segment start/end time: 2071-12-31T16:00:00.000/2071-12-31T16:00:00.000 (in millis: 3218832000000/3218832000000) for time column: date`

pinot-core/src/test/java/org/apache/pinot/queries/FastHllQueriesTest.java
Jackie-Jiang: We still enable the check by default, it is just the boolean value default is false.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
siddharthteotia: Got it. Thanks @Jackie-Jiang

pinot-core/src/main/java/org/apache/pinot/core/segment/creator/impl/SegmentColumnarIndexCreator.java
siddharthteotia: Sounds good.

siddharthteotia:
LGTM. Let's merge once we get a clean build.

pinot-common/src/main/java/org/apache/pinot/common/utils/time/TimeUtils.java
```diff
@@ -108,7 +111,16 @@ public static TimeUnit timeUnitFromString(@Nullable String timeUnitString) {
    * <p>The current valid range used is between beginning of 1971 and beginning of 2071.
    */
   public static boolean timeValueInValidRange(long timeValueInMillis) {
-    return (VALID_MIN_TIME_MILLIS <= timeValueInMillis && timeValueInMillis <= VALID_MAX_TIME_MILLIS);
+    return timeValueInMillis >= VALID_MIN_TIME_MILLIS && timeValueInMillis <= VALID_MAX_TIME_MILLIS;
```
mcvsubbu: Not sure why this change was made

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
```diff
@@ -595,12 +595,12 @@ public void setOnHeap(boolean onHeap) {
     _onHeap = onHeap;
   }
 
-  public boolean isCheckTimeColumnValidityDuringGeneration() {
-    return _checkTimeColumnValidityDuringGeneration;
+  public boolean isSkipTimeValueCheck() {
```
mcvsubbu: This is a backward-incompatible change , please flag the PR in the commit message. Will be good if you can identify the checkin/date when the config was introduced, so that we can use that during pinot release.

pinot-common/src/main/java/org/apache/pinot/common/utils/time/TimeUtils.java
Jackie-Jiang: Just want keep the `timeValueInValidRange()` and `isValidTimeInterval()` checking order consistent.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/generator/SegmentGeneratorConfig.java
Jackie-Jiang: Good suggestion. Added to the PR description

PR #4751, opened 2019-10-29
AvroSchemaToPinotSchema should return immediately if outputDir doesn't exist

sekikn:
I ran `pinot-admin.sh AvroSchemaToPinotSchema` with non-existent output directory and got the following error. It should exit immediately after the error message is displayed.

```
$ bin/pinot-admin.sh AvroSchemaToPinotSchema -outputDir /tmp/foo -pinotSchemaName weather -dimensions station -metrics temp -timeColumnName time -avroDataFile ~/repos/avro/share/test/data/weather.avro 
ERROR: Output directory: %s does not exist or is not a directory
Store Pinot schema to file: /tmp/foo/weather.json
Exception caught: 
java.io.FileNotFoundException: /tmp/foo/weather.json (No such file or directory)
(snip)
```


mcvsubbu:
@sekikn  thanks for your contribution to pinot

PR #4761, opened 2019-10-30
[TE] Centralized cache proof-of-concept for anomaly detection via Couchbase

bxji:
todo still (will add later before merging): 
- unit tests
- move files to proper packages 
- add index creation for couchbase to build script? 

This code is the foundation for using Couchbase as a centralized cache for time-series data. It helps speed up anomaly detection jobs, especially jobs with dimension exploration enabled. Previews may not see too big of a performance improvement, but recurring alerts should see significant boosts in latency over time.

- By default, using the centralized cache is disabled. Using the in-memory cache is also enabled by default.
- The config for authentication is currently in /config/data-sources/cache-config.yml, along with the TTL (time-to-live) for documents in the bucket.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,154 @@
...
+    // we can consider using fixedThreadPool instead.
+    ExecutorService executor = Executors.newCachedThreadPool();
+
+    // insert points in parallel
+    for (MetricFunction metric : response.getMetricFunctions()) {
+      String metricUrn = MetricEntity.fromMetric(response.getRequest().getFilterSet().asMap(), metric.getMetricId()).getUrn();
+      for (int i = 0; i < response.getNumRowsFor(metric); i++) {
+        Map<String, String> row = response.getRow(metric, i);
+        TimeSeriesDataPoint dp = new TimeSeriesDataPoint(metricUrn, Long.parseLong(row.get(CacheConstants.TIMESTAMP)), metric.getMetricId(), row.get(metric.toString()));
+        executor.execute(() -> this.cacheDAO.insertTimeSeriesDataPoint(dp));
```
vincentchenjl: Maybe consider batch upsert or at least add TODO here to optimize.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
```diff
@@ -0,0 +1,106 @@
...
+      JsonObject documentBody = CacheUtils.buildDocumentStructure(point);
+      doc = JsonDocument.create(point.getDocumentKey(), CacheConfig.TTL, documentBody);
+    } else {
+      JsonObject dimensions = doc.content();
+      if (dimensions.containsKey(point.getMetricUrnHash())) {
+        return;
+      }
+
+      dimensions.put(point.getMetricUrnHash(),
+          (point.getDataValue() == null || point.getDataValue().equals("null")) ? "0" : point.getDataValue());
```
vincentchenjl: Do we need to check how many dimensions we already have to ensure that the size of documents is reasonable before we add more dimension?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.pinot.thirdeye.datasource.cache.QueryCache;
+import org.apache.pinot.thirdeye.rootcause.impl.MetricEntity;
+import org.apache.pinot.thirdeye.util.ThirdEyeUtils;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class DefaultTimeSeriesCache implements TimeSeriesCache {
```
vincentchenjl: Better add test for this class, as this class has all the core logic of this feature.

vincentchenjl: Also please add comments to this class and all public class.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
```diff
@@ -0,0 +1,106 @@
...
+        .put(CacheConstants.START, request.getStartTimeInclusive())
+        .put(CacheConstants.END, request.getEndTimeExclusive() - request.getRequest().getGroupByTimeGranularity().toMillis());
+
+    String query = CacheUtils.buildQuery(parameters);
+
+    N1qlQueryResult queryResult = bucket.query(N1qlQuery.simple(query));
+
+    if (!queryResult.finalSuccess()) {
+      LOG.error("cache error occurred for window startTime = {} to endTime = {}", request.getStartTimeInclusive(), request.getEndTimeExclusive());
+      throw new Exception("query to Couchbase failed");
```
xiaohui-sun: Is it possible to pull the error message?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
```diff
@@ -0,0 +1,106 @@
...
+import com.couchbase.client.java.query.N1qlQueryResult;
+import com.couchbase.client.java.query.N1qlQueryRow;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.pinot.thirdeye.util.CacheUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class CouchbaseCacheDAO {
```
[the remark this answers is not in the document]
xiaohui-sun: Suggest to add more comments on the cache layout.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
```diff
@@ -0,0 +1,106 @@
...
+      LOG.error("cache error occurred for window startTime = {} to endTime = {}", request.getStartTimeInclusive(), request.getEndTimeExclusive());
+      throw new Exception("query to Couchbase failed");
+    }
+
+    List<TimeSeriesDataPoint> timeSeriesRows = new ArrayList<>();
+
+    for (N1qlQueryRow row : queryResult) {
+      long timestamp = row.value().getLong(CacheConstants.TIME);
+      String dataValue = row.value().getString(dimensionKey);
+      timeSeriesRows.add(new TimeSeriesDataPoint(request.getMetricUrn(), timestamp, request.getMetricId(), dataValue));
```
xiaohui-sun: I would suggest to explorer columnar storage type.
With each data point maps to each object has big overheads.
Can you use DataFrame in ThirdEye?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -238,6 +244,16 @@ public DetectionPipelineResult run() throws Exception {
           this.config.getId(), nestedMetrics.size(), MAX_DIMENSION_COMBINATIONS));
     }
 
+    // don't use in-memory cache for dimension exploration, it will cause thrashing
+    if (CacheConfig.useCentralizedCache()) {
+      if (this.cachingPeriodLookback >= 0) {
+        this.provider.fetchTimeseries(nestedMetrics.stream()
```
vincentchenjl: Are we doing cache warm-up here? If yes, can this call be slow?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,154 @@
...
+      rows.add(row);
+    }
+
+    return new RelationalThirdEyeResponse(request, rows, timeSpec);
+  }
+
+  public void insertTimeSeriesIntoCache(ThirdEyeResponse response) {
+
+    // we can consider using fixedThreadPool instead.
+    ExecutorService executor = Executors.newCachedThreadPool();
```
xiaohui-sun: 1. we need to add a limit to the number of threads.
2. the executor should be reused. Why init it each time?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/loader/DefaultTimeSeriesLoader.java
```diff
@@ -36,16 +37,18 @@
 
   private final MetricConfigManager metricDAO;
   private final DatasetConfigManager datasetDAO;
-  private final QueryCache cache;
+  private final QueryCache queryCache;
+  private final DefaultTimeSeriesCache timeSeriesCache;
```
vincentchenjl: My suggestion here is that we should put `timeSeriesCache` in `DefaultDataProvider`, as this cache is in the same level as the memory cache we had, from the perspective of the architecture. It is much easier to understand if all the logic of loading from cache or data source is in one place.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
```diff
@@ -0,0 +1,35 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.pinot.thirdeye.detection.cache;
+
+public class CacheConfig {
```
vincentchenjl: All these config values should be configurable.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,154 @@
...
+
+    // insert points in parallel
```
xiaohui-sun: Any reason we want to insert points in parallel? Is the inserting very slow?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/MergeSliceType.java
```diff
@@ -0,0 +1,25 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.pinot.thirdeye.detection.cache;
+
+public enum MergeSliceType {
```
xiaohui-sun: Add comments

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/ThirdEyeCacheRequest.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package org.apache.pinot.thirdeye.detection.cache;
+
+import org.apache.pinot.thirdeye.datasource.ThirdEyeRequest;
+import org.apache.pinot.thirdeye.rootcause.impl.MetricEntity;
+import org.apache.pinot.thirdeye.util.CacheUtils;
+
+
+public class ThirdEyeCacheRequest {
```
xiaohui-sun: Add comments.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
```diff
@@ -0,0 +1,35 @@
...
+  private static boolean useCentralizedCache = false;
+  private static boolean useInMemoryCache = true;
+
+  public static final String COUCHBASE_AUTH_USERNAME = "thirdeye";
+  public static final String COUCHBASE_AUTH_PASSWORD = "thirdeye";
+  public static final String COUCHBASE_BUCKET_NAME = "cache";
+
+  // time-to-live for documents in the cache.
+  public static final int TTL = 3600;
```
xiaohui-sun: Please put that into the configuration file.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
```diff
@@ -0,0 +1,106 @@
...
+
+    JsonDocument doc = bucket.getAndTouch(point.getDocumentKey(), CacheConfig.TTL);
+
+    if (doc == null) {
+      JsonObject documentBody = CacheUtils.buildDocumentStructure(point);
+      doc = JsonDocument.create(point.getDocumentKey(), CacheConfig.TTL, documentBody);
+    } else {
+      JsonObject dimensions = doc.content();
+      if (dimensions.containsKey(point.getMetricUrnHash())) {
+        return;
```
xiaohui-sun: Should we update cache if it already exists?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
```diff
@@ -0,0 +1,106 @@
...
+      }
+
+      dimensions.put(point.getMetricUrnHash(),
```
xiaohui-sun: point.getMetricUrnHash() is used multiple times. Should store and reuse.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
```diff
@@ -0,0 +1,35 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.pinot.thirdeye.detection.cache;
+
+public class CacheConfig {
+  private static boolean useCentralizedCache = false;
+  private static boolean useInMemoryCache = true;
```
vincentchenjl: User should be able to configure which cache should be used.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
xiaohui-sun: The pattern "point.getDataValue().equals("null")) ? "0" : point.getDataValue())" is used multiple times. Should put into getDataValue().

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public static final String COUCHBASE_AUTH_USERNAME = "thirdeye";
+  public static final String COUCHBASE_AUTH_PASSWORD = "thirdeye";
```
vincentchenjl: We should not put the credential in code.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,154 @@
...
+  // fetches missing time-series slices and merges them into the cacheResponse object.
+  private void fetchMissingSlices(ThirdEyeCacheResponse cacheResponse) throws Exception {
+
+    ThirdEyeRequest request = cacheResponse.getCacheRequest().getRequest();
+    long metricId = request.getMetricFunctions().get(0).getMetricId();
+    long requestSliceStart = request.getStartTimeInclusive().getMillis();
+    long requestSliceEnd = request.getEndTimeExclusive().getMillis();
+
+    ThirdEyeResponse result;
+    MetricSlice slice;
```
xiaohui-sun: Is it possible that the missing slice is in the middle?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
```diff
@@ -0,0 +1,62 @@
...
+    c.update(metricUrn.getBytes());
+    return String.valueOf(c.getValue());
+  }
+
+  public static JsonObject buildDocumentStructure(TimeSeriesDataPoint point) {
+    JsonObject body = JsonObject.create()
+        .put("time", point.getTimestamp())
+        .put("metricId", point.getMetricId())
+        .put(point.getMetricUrnHash(),
+            (point.getDataValue() == null || point.getDataValue().equals("null")) ? "0" : point.getDataValue());
```
vincentchenjl: What about "NULL"? how do we handle null value here?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/ThirdEyeCacheResponse.java
```diff
@@ -0,0 +1,87 @@
...
+    if (this.hasNoRows()) {
+      return true;
+    }
+
+    // note: sliceEnd is exclusive so we don't need equality check.
+    long timeGranularity = request.getRequest().getGroupByTimeGranularity().toMillis();
+    return sliceEnd > getLastTimestamp() + timeGranularity;
+  }
+
+  public void mergeSliceIntoRows(ThirdEyeResponse slice, MergeSliceType type) {
```
xiaohui-sun: Why use merge type?
It is simply a mergeSort based on timestamp. Isn't it?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,154 @@
...
+    if (cacheResponse.isMissingSlice(sliceStart.getMillis(), sliceEnd.getMillis())) {
+      fetchMissingSlices(cacheResponse);
+    }
+
+    return buildThirdEyeResponseFromCacheResponse(cacheResponse);
+  }
+
+  // fetches missing time-series slices and merges them into the cacheResponse object.
+  private void fetchMissingSlices(ThirdEyeCacheResponse cacheResponse) throws Exception {
+
```
vincentchenjl: Probably log some metrics here so that we can see what the hit percentage of our cache is.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
bxji: You are right, this was just a temporary stop-gap for now to make the code a little cleaner. Will add code to read these from config file.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
bxji: You're correct, will move these to config file asap

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: Yep, still writing tests right now and will add those in asap

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/ThirdEyeCacheResponse.java
bxji: yes, my bad. It's an old approach I was trying out that I forgot to rework. Will change this asap

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: Yes, it's possible but probably fairly rare. The case where this happens is when the documents for a specific timeseries are about to expire in the cache, and there is a cache fetch during the small slice of time when some of the documents have expired, but others have not. Also, it's difficult to know whether the missing documents in the middle are because there are actually missing documents, or if these are missing data points from the data source itself. Since this case is so complicated and fairly rare, I decided to ignore it and pretend it doesn't happen for now. It can be added, but may require us to refetch the entire timeseries from the data source to ensure data integrity, which we're trying to avoid as much as possible.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
bxji: I assumed no, since historical timeseries data will not change (unless it's from inGraph...). I think it's fine to keep this assumption (it will save us a write), unless you feel strongly otherwise.

thirdeye/thirdeye-pinot/pom.xml
```diff
@@ -239,6 +239,12 @@
       <artifactId>parsii</artifactId>
     </dependency>
 
+    <dependency>
```
kishoreg: we should not bring this dependency here. can we make this a runtime dependency

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: Good idea, will add this

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/TimeSeriesDataPoint.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public String getDocumentKey() {
+    return metricId + "_" + timestamp;
+  }
+
+  public String getMetricUrnHash() {
+    return CacheUtils.hashMetricUrn(metricUrn);
+  }
+
+  public static TimeSeriesDataPoint from(String[] dataPoint, String metricUrn) {
```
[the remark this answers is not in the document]
bxji: will be removing this code, it was from a previous rendition and is not used anymore

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/TimeSeriesDataPoint.java
```diff
@@ -0,0 +1,81 @@
...
+    long timestamp = Long.valueOf(dataPoint[dataPoint.length - 1]);
+    long metricId = MetricEntity.fromURN(metricUrn).getId();
+    String dataValue = dataPoint[1];
+    return new TimeSeriesDataPoint(metricUrn, timestamp, metricId, dataValue);
+  }
+
+  public static TimeSeriesDataPoint from(String[] dataPoint, long metricId, String metricUrn) {
+    long timestamp = Long.valueOf(dataPoint[dataPoint.length - 1]);
```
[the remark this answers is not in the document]
bxji: will be removing this code, it was from a previous rendition and is not used anymore

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
bxji: Yes, it's used to prefetch data for the detection and warm up the cache. It can be slow depending on the number of dimension combinations we need to make.

thirdeye/thirdeye-pinot/pom.xml
akshayrai: @kishoreg, what is the rationale behind this?

kishoreg: Its a bit pre-mature but not abstracting these right now will result in a classpath nightmare. We are seeing this in Pinot and I feel its better to break optional stuff into interface and implementation. And then move the implementation to a separate module and load them at runtime. Once we have this, it makes it easy to add multiple implementations. For e.g. we might use some other system for caching.

thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/detection/cache/CacheUtilTests.java
```diff
@@ -0,0 +1,86 @@
+package org.apache.pinot.thirdeye.detection.cache;
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/detection/cache/CacheUtilTests.java
```diff
@@ -0,0 +1,86 @@
...
+import com.couchbase.client.java.document.json.JsonObject;
+import org.apache.pinot.thirdeye.util.CacheUtils;
+import org.mockito.Mockito;
+import org.testng.Assert;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class CacheUtilTests {
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package org.apache.pinot.thirdeye.util;
+
+import com.couchbase.client.java.document.json.JsonObject;
+import java.util.zip.CRC32;
+import org.apache.pinot.thirdeye.detection.cache.TimeSeriesDataPoint;
+
+
+public class CacheUtils {
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/TimeSeriesDataPoint.java
```diff
@@ -0,0 +1,81 @@
...
+ * under the License.
+ */
+
+package org.apache.pinot.thirdeye.detection.cache;
+
+import org.apache.pinot.thirdeye.rootcause.impl.MetricEntity;
+import org.apache.pinot.thirdeye.util.CacheUtils;
+
+
+public class TimeSeriesDataPoint {
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
```diff
@@ -0,0 +1,106 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(CouchbaseCacheDAO.class);
+
+  private Bucket bucket;
+
+  public CouchbaseCacheDAO() {
+    this.createDataStoreConnection();
+  }
+
+  private void createDataStoreConnection() {
+    if (CacheConfig.useCentralizedCache()) {
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/TimeSeriesDataPoint.java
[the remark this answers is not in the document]
bxji: code is now removed -- for context, the existing code for RelationalThirdEyeResponse puts all the returned data into a List<String[]>, where each String[] represents a single data point. I wanted to give some support if someone wanted to build a data point from the raw representation.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
bxji: yes, added some code to log the errors (there may be more than 1 error).

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
bxji: fixed this and added it to .getDataValue() method. Now, we check whether dataValue.toLowerCase() is equal to "null" rather than direct comparison to "null".

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/MergeSliceType.java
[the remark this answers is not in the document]
bxji: ended up removing this whole class while refactoring.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public static String hashMetricUrn(String metricUrn) {
+    CRC32 c = new CRC32();
```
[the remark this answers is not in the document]
bxji: Yes, but not sure if having to call CRC32.reset() after computing the hash each time is more overhead than just creating a new CRC32 object. Added it as static, the difference should be extremely small in either case.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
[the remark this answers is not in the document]
bxji: added more comments, and added in some examples.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,154 @@
...
+  private CouchbaseCacheDAO cacheDAO = new CouchbaseCacheDAO();
+
+  public DefaultTimeSeriesCache(MetricConfigManager metricDAO, DatasetConfigManager datasetDAO, QueryCache queryCache) {
+    this.metricDAO = metricDAO;
+    this.datasetDAO = datasetDAO;
+    this.queryCache = queryCache;
+  }
+
+  public ThirdEyeResponse fetchTimeSeries(ThirdEyeRequest thirdEyeRequest) throws Exception {
+    //LOG.info("trying to fetch data from cache...");
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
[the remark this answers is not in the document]
bxji: added

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/TimeSeriesDataPoint.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private String metricUrn;
+  private long timestamp;
+  private long metricId;
+  private String dataValue;
```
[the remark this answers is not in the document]
bxji: currently the RelationalThirdEyeResponse stores all values as strings. I can make it double instead, but we will just be converting back and forth each time we write/read from the cache.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/MergeSliceType.java
bxji: class removed as part of a refactor

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
bxji: I think it's highly unlikely for us to ever reach that point; we set a cap on the number of dimensions which we could feasibly have (I think somewhere around 1000). Couchbase has a built-in document size limitation of 20MB. One of our existing customer alerts does dimension exploration on 3 dimensions, and has a document size of 2.7KB per. For 1000 dimensions, we will end up with ~1MB per document, which will absolutely fit in our 20 MB limit.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
```diff
@@ -0,0 +1,62 @@
...
+    JsonObject body = JsonObject.create()
+        .put("time", point.getTimestamp())
+        .put("metricId", point.getMetricId())
+        .put(point.getMetricUrnHash(),
+            (point.getDataValue() == null || point.getDataValue().equals("null")) ? "0" : point.getDataValue());
+    return body;
+  }
+
+  public static String buildQuery(JsonObject parameters) {
+    StringBuilder sb = new StringBuilder("SELECT time, `")
```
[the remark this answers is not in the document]
bxji: done, used String.format() to make a little more clear and added an example query in comments. Eventually, the goal is to use the parameterized query that Couchbase Java SDK offers, but this will work for now.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: It's not overly slow, but we still save a couple seconds by parallelizing this. since each data point has no dependency on other data points (as in order of insertion doesn't matter), I figured this would be an easy optimization.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
bxji: will add this to config file

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CouchbaseCacheDAO.java
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/ThirdEyeCacheRequest.java
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: added

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: Currently we are doing parallel insertion, not sure if batch upsert will be faster or not. Will get back to you on this one

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: we can use fixedThreadPool instead, which will let us set a limit on the number of threads, but maybe using batch upsert is more efficient than parallel inserts. Will look more into this and get back to you

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/utils/ThirdeyeMetricsUtil.java
```diff
@@ -110,6 +110,15 @@ public Integer value() {
   public static final Counter datasourceExceptionCounter =
       metricsRegistry.newCounter(ThirdeyeMetricsUtil.class, "datasourceExceptionCounter");
 
+  public static final Counter couchbaseCallCounter =
```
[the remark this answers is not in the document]
bxji: I'm not sure how we can get current cache capacity, as this seems like a metric that whoever is operating and maintaining the cluster should emit. Maybe we can heartbeat the cluster every now and then and ask how it's doing, but I would need to look more into this. Cache hit ratio and cache miss ratio should be doable if we can get derived metrics (total cache hits in a day/total cache calls in a day, total cache misses in a day/total cache calls in a day, etc.), and I will also look into getting partial cache hit ratio.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: done, and number of threads is configurable in config file

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,194 @@
...
+    long metricId = request.getMetricFunctions().get(0).getMetricId();
+    long requestSliceStart = request.getStartTimeInclusive().getMillis();
+    long requestSliceEnd = request.getEndTimeExclusive().getMillis();
+
+    ThirdEyeResponse result;
+    MetricSlice slice;
+
+    if (cacheResponse.hasNoRows()) {
+      result = this.queryCache.getQueryResult(cacheResponse.getCacheRequest().getRequest());
+      insertTimeSeriesIntoCache(result);
```
jihaozh: How about we make all `insertTimeSeriesIntoCache()` asynchronous? so that it won't impact the performance even in the case of a cache miss.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
```diff
@@ -0,0 +1,194 @@
...
+      cacheResponse.mergeSliceIntoRows(result);
+    }
+
+    if (cacheResponse.isMissingStartSlice(requestSliceStart)) {
```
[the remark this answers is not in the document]
bxji: It's possible for the data in the cache to be missing some data from the beginning of the timeseries and also missing some data from the end of the timeseries at the same time, I think? It's a pretty rare case, though, since we need to fetch 28/90 days training data, but can still happen due to TTL/expiration in the cache if older documents don't have expiration refreshed and thus expire.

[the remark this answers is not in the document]
bxji: oh, I see your point. I will add this, thanks. Otherwise if we are doing detection from 10/01 - 11/01 but dataset only has data until 10/31, we will try to fetch last days data again even though it doesn't exist.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
xiaohui-sun: Can we add a sanity check to make sure there is no whole in the middle.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/RelationalThirdEyeResponse.java
```diff
@@ -80,6 +82,7 @@ public int getNumRowsFor(MetricFunction metricFunction) {
       rowMap.put(groupByKey, rowValues[i]);
     }
     rowMap.put(metricFunction.toString(), rowValues[metricFuncToIdMapping.get(metricFunction)]);
+    rowMap.put(CacheConstants.TIMESTAMP, rowValues[rowValues.length - 1]);
```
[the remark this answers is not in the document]
bxji: No, timestamp isn't a group by key for fetching timeseries, so we need to manually add it in.

thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
```diff
@@ -0,0 +1,20 @@
+
+# please use at most one cache at a time, either the in-memory cache or the centralized cache.
+useInMemoryCache: true
```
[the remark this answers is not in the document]
bxji: I'm not sure I understand your question; if both useInMemoryCache and useCentralizedCache are set to false, then no caching is enabled.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/ThirdEyeCacheRegistry.java
```diff
@@ -90,14 +98,42 @@ public static void initDataSources(ThirdEyeConfiguration thirdeyeConfig) {
...
+      if (INSTANCE.getTimeSeriesCache() == null) {
+        TimeSeriesCache timeSeriesCache = new DefaultTimeSeriesCache(DAO_REGISTRY.getMetricConfigDAO(), DAO_REGISTRY.getDatasetConfigDAO(),
+            ThirdEyeCacheRegistry.getInstance().getQueryCache(),
+            new CouchbaseCacheDAO(),
+            Executors.newFixedThreadPool(CacheConfig.getCentralizedCacheSettings().getMaxParallelInserts()));
+
+        ThirdEyeCacheRegistry.getInstance().registerTimeSeriesCache(timeSeriesCache);
+      }
+    } catch (Exception e) {
+      LOGGER.info("Caught exception while initializing centralized cache", e);
```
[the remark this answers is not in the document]
bxji: exception is probably better. When we make queries to the centralized cache, the code as is throws an exception if a fatal error occurs, so no detections will complete if the connection is not properly initialized.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/TimeSeriesDataPoint.java
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * We use the appended metricId and timestamp together
+   * as the document key for the data point in the cache.
+   * This might look something like:
+   * 1351840_1185783260000
+   * @return document key
+   */
+  public String getDocumentKey() {
+    return metricId + "_" + timestamp;
```
abagshaw:
```suggestion
    return String.format("%s_%s", metricId, timestamp);
```

thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/util/CacheUtilTest.java
```diff
@@ -0,0 +1,104 @@
...
+  @Test
+  public void testBuildQuery() {
+    jsonObject.put(BUCKET, bucketName)
+        .put(METRIC_ID, metricId)
+        .put(DIMENSION_KEY, CacheUtils.hashMetricUrn(metricUrn))
+        .put(START, 100)
+        .put(END, 200);
+
+    String query = CacheUtils.buildQuery(jsonObject);
+    String expectedQuery = "SELECT time, `624972944` FROM `TestBucket` WHERE metricId = 1 AND `624972944` IS NOT MISSING AND time BETWEEN 100 AND 200 ORDER BY time ASC";
```
abagshaw: This is pretty brittle - if the formatting of the query changes in any way your test will fail (even if the query itself is still valid). It may not be worth fixing this, but it would be worth noting at the top of the test that it is fairly brittle incase someone makes a change to the query in the future and is confused why the test is failing.

Ideally you would run a small test instance of the DB running that you could test the query against and then verify the results are as expected (less brittle). But that might be a pain to set up.

thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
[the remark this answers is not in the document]
bxji: changed the code to use in-memory by default if cache config could not be read

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConfig.java
```diff
@@ -0,0 +1,73 @@
...
+  // left blank
+  public CacheConfig() {}
+
+  public static CacheConfig getInstance() { return instance; }
+
+  public static boolean useCentralizedCache() { return useCentralizedCache; }
+  public static boolean useInMemoryCache() { return useInMemoryCache; }
+  public static CentralizedCacheConfig getCentralizedCacheSettings() { return centralizedCacheSettings; }
+
+  public static String getHost() { return host; }
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/ThirdEyeCacheRegistry.java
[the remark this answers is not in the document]
bxji: changed the code so that if exception occurs, we exit (since all queries to couchbase will fail anyways), but if the issue is that the config could not be read, we default to using in-memory cache.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/common/ThirdEyeConfiguration.java
```diff
@@ -77,7 +78,23 @@ public URL getDataSourcesAsUrl() {
       URL rootUrl = new URL(String.format("file:%s/", this.rootDir));
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheConstants.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.pinot.thirdeye.detection.cache;
+
+/*
+  Constants used for centralized-cache related code.
+ */
+public class CacheConstants {
+  public static final String TIMESTAMP = "timestamp";
+
+  // couchbase document field names
+  public static final String BUCKET = "bucket";
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/common/ThirdEyeConfiguration.java
```diff
@@ -77,7 +78,23 @@ public URL getDataSourcesAsUrl() {
...
       return new URL(rootUrl, this.dataSources);
     } catch (MalformedURLException e) {
-      throw new IllegalArgumentException(String.format("Could not parse relative path for rootDir '%s' and datSources '%s'", this.rootDir, this.dataSources));
+      throw new IllegalArgumentException(String.format("Could not parse relative path for rootDir '%s' and dataSources '%s'", this.rootDir, this.dataSources));
+    }
+  }
+
+  // same as above but for cache config
+  public URL getCacheConfigAsUrl() {
```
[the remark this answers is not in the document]
bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/algorithm/DimensionWrapper.java
```diff
@@ -238,6 +244,17 @@ public DetectionPipelineResult run() throws Exception {
           this.config.getId(), nestedMetrics.size(), MAX_DIMENSION_COMBINATIONS));
     }
 
+    // don't use in-memory cache for dimension exploration, it will cause thrashing
+    // we add a "duplicate" condition so that tests can run. will fix tests later on
+    if (CacheConfig.useCentralizedCache() && !CacheConfig.useInMemoryCache()) {
+      if (this.cachingPeriodLookback >= 0) {
```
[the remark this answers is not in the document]
bxji: fetchTimeseries logs messages about fetching a slice, nothing specifically for cache warmup/pre-fetching. do you feel that we need log messages specifically for cache warmup?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: added, it will log any missing timeseries points in the middle of the timeseries pulled from the cache. it will produce noise if the original time-series is missing data points in the middle of the time-series, the only way to fix this is to insert extra documents with "null" as the value, and skip those data points when putting the cached data back together.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
```diff
@@ -0,0 +1,88 @@
...
+   * SELECT time, `71252492` FROM `te-cache-bucket`
+   *    WHERE metricId = 1351714
+   *      AND `71252492` IS NOT MISSING
+   *        AND time BETWEEN 100000000000 AND 200000000000
+   *          ORDER BY time ASC
+   * @param parameters JsonObject containing the required data to build the query.
+   * @return query string
+   */
+  public static String buildQuery(JsonObject parameters) {
+    return String.format("SELECT time, `%s` FROM `%s` WHERE metricId = %d AND `%s` IS NOT MISSING AND time BETWEEN %d AND %d ORDER BY time ASC",
```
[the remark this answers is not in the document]
bxji: I had some issues with getting the parameterized query to work when I originally wrote the code, I will take another look at it once the other issues are fixed.

thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/util/CacheUtilTest.java
bxji: I think we have some setup for this already (or at least I saw some logs that other tests are doing this). I'll look into it, thanks!

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/TimeSeriesDataPoint.java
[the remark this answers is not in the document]
bxji: done @jihaozh

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/ThirdEyeCacheRequest.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public static ThirdEyeCacheRequest from(ThirdEyeRequest request) {
+
+    long metricId = request.getMetricFunctions().get(0).getMetricId();
+    String metricUrn = MetricEntity.fromMetric(request.getFilterSet().asMap(), metricId).getUrn();
+    long startTime = request.getStartTimeInclusive().getMillis();
+    long endTime = request.getEndTimeExclusive().getMillis();
+
+    return new ThirdEyeCacheRequest(request, metricId, metricUrn, startTime, endTime);
```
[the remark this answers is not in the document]
bxji: (sort of) done. Made it a wrapper minus two fields (metricUrn, metricUrnHash) since they are expensive to generate each time.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/ThirdEyeCacheRegistry.java
```diff
@@ -90,14 +98,42 @@ public static void initDataSources(ThirdEyeConfiguration thirdeyeConfig) {
...
+
+      cacheConfig.setHost(dataSource.getHost());
+      cacheConfig.setAuthUsername(dataSource.getAuthUsername());
+      cacheConfig.setAuthPassword(dataSource.getAuthPassword());
+      cacheConfig.setBucketName(dataSource.getBucketName());
+
+      if (INSTANCE.getTimeSeriesCache() == null) {
+        TimeSeriesCache timeSeriesCache = new DefaultTimeSeriesCache(DAO_REGISTRY.getMetricConfigDAO(), DAO_REGISTRY.getDatasetConfigDAO(),
+            ThirdEyeCacheRegistry.getInstance().getQueryCache(),
+            new CouchbaseCacheDAO(),
```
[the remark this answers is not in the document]
bxji: added todo

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
```diff
@@ -0,0 +1,88 @@
...
+   *   "metricId": 1351714
+   *   "71252492": "100.0"
+   * }
+   * @param point some TimeSeriesDataPoint
+   * @return JsonObject with base document.
+   */
+  public static JsonObject buildDocumentStructure(TimeSeriesDataPoint point) {
+    JsonObject body = JsonObject.create()
+        .put("time", point.getTimestamp())
+        .put("metricId", point.getMetricId())
```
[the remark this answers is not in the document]
bxji: done, forgot to actually add the constant variables in to this piece of code! also switched to using "timestamp"

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/ThirdEyeCacheResponse.java
```diff
@@ -0,0 +1,144 @@
...
+
+public class ThirdEyeCacheResponse {
+  private final ThirdEyeCacheRequest request;
+  private List<TimeSeriesDataPoint> rows;
+  private long firstTimestamp;
+  private long lastTimestamp;
+
+  public ThirdEyeCacheResponse(ThirdEyeCacheRequest request, List<TimeSeriesDataPoint> rows) {
+    this.request = request;
+    this.rows = rows;
```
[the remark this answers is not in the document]
bxji: done, though maybe still not 100% clear...

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/ThirdEyeCacheRequest.java
```diff
@@ -0,0 +1,77 @@
...
+public class ThirdEyeCacheRequest {
+
+  ThirdEyeRequest request;
+  private long metricId;
+  private String metricUrn;
+  private long startTimeInclusive;
+  private long endTimeExclusive;
+  private String dimensionKey;
+
+  public ThirdEyeCacheRequest(ThirdEyeRequest request, long metricId, String metricUrn, long startTimeInclusive, long endTimeExclusive) {
```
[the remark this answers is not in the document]
bxji: changed this class to be a wrapper class

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/DefaultTimeSeriesCache.java
bxji: Previously I was seeing that running it synchronously had slightly better performance, but I will take another look

PR #4763, opened 2019-10-30
[TE] endpoints - harleyjj/alert-activation - make endpoint for toggli…

harleyjj:
…ng alert activation

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
@@ -36,9 +36,11 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import javax.validation.constraints.NotNull;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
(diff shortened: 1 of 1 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
```diff
@@ -540,6 +541,41 @@ public Response createUserAnomaly(
     return Response.ok(anomaly.getId()).build();
   }
 
+  /**
+   * Toggle active/inactive for given detection
+   *
+   * @param detectionId detection config id (must exist)
+   */
+  @PUT
+  @Path("/activation/{id}")
```
xiaohui-sun: Can we explicitly setting active/inactive?
By using toggle() users need to know whether it is active/in-active first which is an extra work.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
```diff
@@ -540,6 +541,41 @@ public Response createUserAnomaly(
...
+    try {
+      DetectionConfigDTO config = this.configDAO.findById(detectionId);
+      if (config == null) {
+        throw new IllegalArgumentException(String.format("Cannot find config %d", detectionId));
+      }
+
+      // grab for logs
+      toggledTo = !config.isActive();
+      // update state
+      config.setActive(!config.isActive());
```
xiaohui-sun: config.setActive(toggledTo) is better.
Let's avoid duplicating logic, although it is very small logic.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
```diff
@@ -540,6 +541,41 @@ public Response createUserAnomaly(
...
+   *
+   * @param detectionId detection config id (must exist)
+   */
+  @PUT
+  @Path("/activation/{id}")
+  @ApiOperation("Make detection active or inactive, given id")
+  public Response toggleActivation(
+      @PathParam("id") long detectionId) throws Exception {
+    Map<String, String> responseMessage = new HashMap<>();
+    Boolean toggledTo = false;
```
xiaohui-sun: 1. Why use Boolean instead of "boolean"?
It requires box/unbox.
2. Default value is false so no need to set explicitly.
3. This is only a temporarily used variable. Declare a variable to its minimum scope.  
No need to declare  before  the "try" block.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/DetectionResource.java
```diff
@@ -540,6 +541,41 @@ public Response createUserAnomaly(
     return Response.ok(anomaly.getId()).build();
   }
 
+  /**
+   * Toggle active/inactive for given detection
+   *
+   * @param detectionId detection config id (must exist)
```
xiaohui-sun: If it is must exist then we should add "@NotNull" annotation but not only in the comment.

xiaohui-sun: Please add swagger annotations for this API.

harleyjj:
Thank you for the really valuable feedback @xiaohui-sun.  I made all the changes you noted.

PR #4764, opened 2019-10-31
Fix the bug of releasing the segment when there are still threads working on it

Jackie-Jiang:
Issue:
When we execute a query, the main thread holds the segment reference,
and avoid the segment getting released during the query execution.
When the query timed out, the main thread will return the exception
and release the segment reference without waiting for all the worker
threads finish the work. This can cause severe issue when segment get
released due to segment deletion/refresh while worker threads are
still accessing it, even JVM crash.

Solution:
Use a Phaser for the query execution to ensure when the main thread
returns the result, all the worker threads are not accessing the
segment (not scheduled or done). To shorten the period of waiting,
check the interrupt flag when the operator starts processing the next
block, and interrupt the execution if the thread is interrupted.

Applied the solution to both planning phase and execution phase, and
also added tests to check the behavior.

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/CombineGroupByOperator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/CombineGroupByOperator.java
@@ -28,6 +28,7 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.Phaser;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -102,8 +103,6 @@ public class CombineGroupByOperator extends BaseOperator<IntermediateResultsBloc
    */
   @Override
   protected IntermediateResultsBlock getNextBlock() {
-    int numOperators = _operators.size();
-    CountDownLatch operatorLatch = new CountDownLatch(numOperators);
     ConcurrentHashMap<String, Object[]> resultsMap = new ConcurrentHashMap<>();
     AtomicInteger numGroups = new AtomicInteger();
     ConcurrentLinkedQueue<ProcessingException> mergedProcessingExceptions = new ConcurrentLinkedQueue<>();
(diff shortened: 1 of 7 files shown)


pinot-core/src/main/java/org/apache/pinot/core/operator/CombineGroupByOperator.java
```diff
@@ -224,6 +241,8 @@ public void runJob() {
           future.cancel(true);
         }
       }
+      // Deregister the main thread and wait for all threads done
```
mcvsubbu: In line 179, don't you need to issue a cacel on the futures, and then jump down to awaitAndDeregister()?

mcvsubbu: OK, I see that the phaser. awaitAdvance() should cover this case. Thanks

pinot-core/src/main/java/org/apache/pinot/core/plan/CombinePlanNode.java
```diff
@@ -92,27 +94,42 @@ public Operator run() {
       // Calculate the time out timestamp
       long endTime = System.currentTimeMillis() + TIME_OUT_IN_MILLISECONDS_FOR_PARALLEL_RUN;
 
-      int threads = Math.min(numPlanNodes / MIN_TASKS_PER_THREAD + ((numPlanNodes % MIN_TASKS_PER_THREAD == 0) ? 0 : 1),
-          // ceil without using double arithmetic
-          MAX_PLAN_THREADS);
-      int opsPerThread = Math.max(numPlanNodes / threads + ((numPlanNodes % threads == 0) ? 0 : 1),
-          // ceil without using double arithmetic
-          MIN_TASKS_PER_THREAD);
+      int numThreads = Math.min((numPlanNodes + NUM_PLANS_PER_THREAD - 1) / NUM_PLANS_PER_THREAD, MAX_PLAN_THREADS);
```
mcvsubbu: while this math seems correct to me, let us avoid making changes unrelated to the PR.

Jackie-Jiang: Sure, will make another PR for the math change

pinot-core/src/test/java/org/apache/pinot/core/operator/CombineSlowOperatorsTest.java
```diff
@@ -0,0 +1,150 @@
...
+        Thread.sleep(3_600_000L);
+      } catch (InterruptedException e) {
+        // Thread should be interrupted
+        throw new RuntimeException(e);
+      } finally {
+        // Wait for 100 milliseconds before marking the operation done
+        try {
+          Thread.sleep(100L);
+        } catch (InterruptedException e) {
+          // Ignore
```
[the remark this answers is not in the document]
Jackie-Jiang: Modified the test so that it will fail if the thread is interrupted twice

PR #4766, opened 2019-10-31
[TE] Extending Jira Merge capabilities for supporting dimensional alerter

akshayrai:
Changes:
* Support for merging jira tickets when dimensions_alerter_pipeline is enabled.
* Injects dimensions and subsId as jira labels.
* Pass the dimensions info in addition to the recipients to the alerter.

Other changes:
* Truncate Jira Summary field which has a static limit of 255 characters max.
* Optm: Query Jira using mergeGap rather than fetching all and filtering in-memory.
* Deprecate DimensionMap and use Multimap

Testing:
* added unit tests
* tested end-to-end by deploying locally

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/DetectionAlertFilterNotification.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/DetectionAlertFilterNotification.java
@@ -19,6 +19,8 @@
 
 package org.apache.pinot.thirdeye.detection.alert;
 
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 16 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/DetectionAlertFilterNotification.java
```diff
@@ -27,19 +28,33 @@
...
     return notificationSchemeProps;
   }
 
-  public DetectionAlertFilterNotification setNotificationSchemeProps(Map<String, Object> notificationSchemeProps) {
+  public void setNotificationSchemeProps(Map<String, Object> notificationSchemeProps) {
     this.notificationSchemeProps = notificationSchemeProps;
-    return this;
+  }
+
+  public Map<String, String> getDimensionFilters() {
```
jihaozh: Is this `dimensionFilters`  the same with the `DimensionMap` in the anomaly? shall we use MultiMap instead of Map?

akshayrai: it is not necessary but probably better to support it. I will update the changes.
Thanks!

PR #4767, opened 2019-10-31
Introduce a mode in the server to disable queries

mcvsubbu:
When we have problems with pinot servers in production, and need to isolate the
server by not sending queries to it, we use the shutdownInProgress field in the
instance config. This does not work if the server is restarted (due to deployment
or by other automated systems) because the upon restart, the server automatically
sets shutdownInProgress to "false". We would still like to isolate the server to
find out what the real issue may be.

This commit introduces another field called "queriesDisabled" in the instance config
that the admin can set.  The broker pays attention to this field and excludes the
server from routing table computations (thus not routing queries to it).

This behavior continues until such time as the value of this field is set to false.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedRouting.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedRouting.java
@@ -222,9 +222,14 @@ public class HelixExternalViewBasedRouting implements ClusterChangeHandler, Rout
           previousInstanceConfig.getRecord().getSimpleField(CommonConstants.Helix.IS_SHUTDOWN_IN_PROGRESS);
       String isShuttingDown =
           currentInstanceConfig.getRecord().getSimpleField(CommonConstants.Helix.IS_SHUTDOWN_IN_PROGRESS);
+      String wasQueriesDisabled =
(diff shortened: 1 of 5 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/routing/HelixExternalViewBasedRouting.java
```diff
@@ -222,9 +222,14 @@ private boolean isRoutingTableRebuildRequired(String tableName, ExternalView ext
...
+      String isQueriesDisabled =
+          currentInstanceConfig.getRecord().getSimpleField(CommonConstants.Helix.QUERIES_DISABLED);
 
       boolean instancesChanged =
-          !EqualityUtils.isEqual(wasEnabled, isEnabled) || !EqualityUtils.isEqual(wasShuttingDown, isShuttingDown);
+          !EqualityUtils.isEqual(wasEnabled, isEnabled) || !EqualityUtils.isEqual(wasShuttingDown, isShuttingDown) ||
+          !EqualityUtils.isEqual(wasQueriesDisabled, isQueriesDisabled);
 
       if (instancesChanged) {
         LOGGER.info(
```
Jackie-Jiang: You should update this info log as well

mcvsubbu: Done

PR #4768, opened 2019-10-31
[TE] deactivate metrics instead of deleting

jihaozh:
Deactivate pinot metrics/datasets instead of deleting. Because some metrics can be removed and re-added later into the database with a different id. It will break existing detections. This PR changes the behavior to deactivate instead of deleting, which will maintain the id of the metrics and datasets.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/AutoOnboardPinotMetadataSource.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/AutoOnboardPinotMetadataSource.java
@@ -146,7 +146,7 @@ public class AutoOnboardPinotMetadataSource extends AutoOnboard {
       List<MetricConfigDTO> metrics = metricDAO.findByDataset(datasetConfigDTO.getDataset());
       int metricCount = metrics.size();
       for (MetricConfigDTO metric : metrics) {
-        if (!metric.isDerived()) {
+        if (!metric.isDerived() && !metric.getName().equals(ROW_COUNT)) {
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/AutoOnboardPinotMetadataSource.java
```diff
@@ -315,16 +305,24 @@ private void checkMetricChanges(String dataset, DatasetConfigDTO datasetConfig,
...
 
-    // remove deleted metrics from ThirdEye
+    // audit existing metrics in ThirdEye
     for (MetricConfigDTO metricConfig : datasetMetricConfigs) {
-      if (!metricConfig.isDerived()) {
-        if (!schemaMetricNames.contains(getColumnName(metricConfig))) {
-          LOG.info("Deleting metric {} in {}", metricConfig.getName(), dataset);
-          DAO_REGISTRY.getMetricConfigDAO().delete(metricConfig);
+      if (!schemaMetricNames.contains(getColumnName(metricConfig))) {
+        if (!metricConfig.isDerived() && !metricConfig.getName().equals(ROW_COUNT)) {
```
akshayrai: What abt derived metrics? Can we activate/de-activate them as well?

jihaozh: Actually, since derived metrics can be created across datasets, it's safer to not de-activate them when one dataset is missing. Also, it's hard to check when to re-activate a derived metric.

PR #4773, opened 2019-11-01
Enhance CombineOperator

Jackie-Jiang:
- Fix the constant calculation
  - Current logic always have MIN_THREADS_PER_QUERY >= MAX_THREADS_PER_QUERY, which will always lead to numGroups = Math.min(numOperators, MIN_THREADS_PER_QUERY)
  - Fix the logic by using at most 10 or half of the processors threads for each query
- Simplify the logic by removing the pre-calculation of the groups, directly using index to access operators
- Early terminate the step 2 merge when the query timed out
- Remove the redundant debug logs

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/operator/CombineOperator.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/operator/CombineOperator.java
@@ -18,7 +18,6 @@
  */
 package org.apache.pinot.core.operator;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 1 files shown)


pinot-core/src/main/java/org/apache/pinot/core/operator/CombineOperator.java
```diff
@@ -47,22 +45,16 @@
   private static final Logger LOGGER = LoggerFactory.getLogger(CombineOperator.class);
   private static final String OPERATOR_NAME = "CombineOperator";
 
+  // Use at most 10 or half of the processors threads for each query.
+  // If there are less than 2 processors, use 1 thread.
+  // Runtime.getRuntime().availableProcessors() may return value < 2 in container based environment, e.g. Kubernetes.
+  private static final int MAX_NUM_THREADS_PER_QUERY =
+      Math.max(1, Math.min(10, Runtime.getRuntime().availableProcessors() / 2));
```
mayankshriv: Will `Runtime.getRuntime().availableProcessors()` still work accurately on cloud VM's (say Azure)? Also, what about when there's cgroups limits imposed?

Jackie-Jiang: That I'm not sure and we need to test that out. Based on Xiang's experience, Kubernetes might return 1 for this, and that is the reason why we ensure this value is at least 1

PR #4774, opened 2019-11-02
Refactor virtualColumn

haibow:
Refactoring VirtualColumn
- Allow virtualColumnContext to support arbitrary fieldSpec
- Consolidate SegmentNameVirtualColumnProvider and HostNameVirtualColumnProvider to SingleStringExtraColumnProvider
- Add virtual columns to consuming segments and fix tests around it.
- Other data types (INT, DOUBLE, etc) and multi-value field support will be added in a separate diff.


pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -409,4 +409,10 @@ public ServerType getServerType() {
     public static final String RETRY_WAITIME_MS = "retry.wait.ms";
     public static final int RETRY_WAITIME_MS_DEFAULT = 100;
   }
+
```
npawar: is Schema the right place for these? Something like Segment.VirtualColumns would be better I think

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/immutable/ImmutableSegmentLoader.java
```diff
@@ -39,13 +41,13 @@
...
 import org.apache.pinot.core.segment.store.SegmentDirectory;
 import org.apache.pinot.core.segment.store.SegmentDirectoryPaths;
-import org.apache.pinot.core.segment.virtualcolumn.VirtualColumnContext;
-import org.apache.pinot.core.segment.virtualcolumn.VirtualColumnProvider;
-import org.apache.pinot.core.segment.virtualcolumn.VirtualColumnProviderFactory;
 import org.apache.pinot.core.startree.v2.store.StarTreeIndexContainer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.pinot.common.utils.CommonConstants.Schema.VIRTUAL_COLUMN_HOSTNAME;
```
npawar: Let's remove the static imports from all the files for the sake of readability. Instead use `Schema.VIRTUAL_COLUMN_HOSTNAME` etc, or better `VirtualColumn.HOSTNAME, VirtualColumn.SEGMENT_NAME` etc.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/immutable/ImmutableSegmentLoader.java
```diff
@@ -119,16 +120,19 @@ public static ImmutableSegment load(File indexDir, IndexLoadingConfig indexLoadi
...
     // Instantiate virtual columns
     for (FieldSpec fieldSpec : schema.getAllFieldSpecs()) {
       if (fieldSpec.isVirtualColumn()) {
         String columnName = fieldSpec.getName();
-        VirtualColumnProvider provider =
-            VirtualColumnProviderFactory.buildProvider(fieldSpec.getVirtualColumnProvider());
-        VirtualColumnContext context = new VirtualColumnContext(NetUtil.getHostnameOrAddress(), segmentName, columnName,
-            segmentMetadata.getTotalDocs());
+        ExtraColumnProvider provider = ExtraColumnProviderFactory.buildProvider(fieldSpec);
+        if (columnName.equals(VirtualColumn.HOSTNAME)) {
```
mcvsubbu: Please move these lines into VirtualColumnProviderFactory.addBuiltInVirtualColumnsToSchema() method

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -472,12 +474,15 @@ public SegmentMetadata getSegmentMetadata() {
   public ColumnDataSource getDataSource(String columnName) {
     FieldSpec fieldSpec = _schema.getFieldSpecFor(columnName);
     if (fieldSpec.isVirtualColumn()) {
-      VirtualColumnContext virtualColumnContext =
-          new VirtualColumnContext(NetUtil.getHostnameOrAddress(), getSegmentName(), columnName, _numDocsIndexed + 1);
-      VirtualColumnProvider virtualColumnProvider =
-          VirtualColumnProviderFactory.buildProvider(_schema.getFieldSpecFor(columnName).getVirtualColumnProvider());
-      return new ColumnDataSource(virtualColumnProvider.buildColumnIndexContainer(virtualColumnContext),
-          virtualColumnProvider.buildMetadata(virtualColumnContext));
+      if (columnName.equals(VirtualColumn.HOSTNAME)) {
```
mcvsubbu: same. Move it to VirtualColumnProviderFactory

pinot-core/src/main/java/org/apache/pinot/core/segment/extracolumn/ExtraColumnContext.java
```diff
@@ -16,14 +16,28 @@
...
 
 /**
- * Virtual column provider that contains the current segment name.
+ * Context used to pass arguments to the extra column provider.
  */
-public class SegmentNameVirtualColumnProvider extends SingleStringVirtualColumnProvider {
-  @Override
-  protected String getValue(VirtualColumnContext context) {
-    return context.getSegmentName();
+public class ExtraColumnContext {
```
mcvsubbu: SegmentNameVirtualColumnProvider got renamed to ExtraColumnContext?

Jackie-Jiang:
The current definition of virtual column is the column backed by the VirtualColumnProvider class. VirtualColumnProvider can store physical resources/access remote resources if necessary. If this is not changed, I don't think we need to rename the virtual column to extra column. IMO, extra column is not as intuitive as virtual column.

haibow:
> The current definition of virtual column is the column backed by the VirtualColumnProvider class. VirtualColumnProvider can store physical resources/access remote resources if necessary. If this is not changed, I don't think we need to rename the virtual column to extra column. IMO, extra column is not as intuitive as virtual column.

@Jackie-Jiang Another implication about virtual column is that it is not included in the user query result. It "does not exist in the record", so only physical (non-virtual) columns are used in query processing layers (e,g, select * query). I'm changing the class to extraColumnProvider, so in the near future it can also provide default values to the new columns that are added to the consuming segment at query time.

One other option is to keep the name virtual column as is, but drop the semantic that it is always excluded from the query result. Instead, only exclude columns that start with "$", as used in
https://github.com/apache/incubator-pinot/blob/master/pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/BaseClusterIntegrationTestSet.java#L218-L219

Thoughts?

Jackie-Jiang:
@haibow Why would renaming virtual column to extra column solve this problem? Or are you suggesting besides virtual column, add another type of column which is backed by a columnProvider but can be selected by `select *`? BTW, the only query which treats virtual column differently is `select *`

pinot-core/src/main/java/org/apache/pinot/core/segment/extracolumn/ExtraColumnContext.java
haibow: Github somehow misaligned the files.. 
- [VirtualColumnContext](https://github.com/apache/incubator-pinot/blob/master/pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/VirtualColumnContext.java) -> ExtraColumnContext
- SegmentNameVirtualColumnProvider and HostNameVirtualColumnProvider are removed, and merged to SingleStringExtraColumnProvider

haibow:
> @haibow Why would renaming virtual column to extra column solve this problem? Or are you suggesting besides virtual column, add another type of column which is backed by a columnProvider but can be selected by `select *`? BTW, the only query which treats virtual column differently is `select *`

Right, extraColumnProvider keeps the concept of providing custom data at query time and not persisting data to disk. It includes virtual column that cannot not answer any user query and another type of column (I don't have a new name for it yet) that can. 

Renaming basically broadens the notion of this column provider so some columns can be included in select * queries, will not be pruned by schema pruner, etc

haibow:
I can keep the virtualColumn as is in this diff since nothing new is introduced here. I can defer the renaming to future diffs where the usage of virtualColumn is expanded - hopefully then we can discuss with more context.

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -109,6 +109,7 @@ public ServerType getServerType() {
         return ServerType.REALTIME;
       }
     }
+
```
haibow: auto formatting

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -337,7 +338,8 @@ public ServerType getServerType() {
        * During realtime segment completion, the value of this enum decides how  non-winner servers should replace  the completed segment.
        */
       public enum CompletionMode {
-        DEFAULT, // default behavior - if the in memory segment in the non-winner server is equivalent to the committed segment, then build and replace, else download
+        DEFAULT,
```
haibow: auto formatting

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -148,6 +148,10 @@ public long getLatestIngestionTimestamp() {
     _segmentPartitionConfig = config.getSegmentPartitionConfig();
     _nullHandlingEnabled = config.isNullHandlingEnabled();
 
+    // Ensure that the schema has the virtual columns added
```
haibow: Now virtual columns are only added to immutable segments, although in mutable segments there are logics in `getDataSource` for virtual columns. So adding virtual columns to mutable segments too.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -472,10 +476,8 @@ public SegmentMetadata getSegmentMetadata() {
   public ColumnDataSource getDataSource(String columnName) {
     FieldSpec fieldSpec = _schema.getFieldSpecFor(columnName);
     if (fieldSpec.isVirtualColumn()) {
-      VirtualColumnContext virtualColumnContext =
-          new VirtualColumnContext(NetUtil.getHostnameOrAddress(), getSegmentName(), columnName, _numDocsIndexed + 1);
-      VirtualColumnProvider virtualColumnProvider =
-          VirtualColumnProviderFactory.buildProvider(_schema.getFieldSpecFor(columnName).getVirtualColumnProvider());
+      VirtualColumnContext virtualColumnContext = new VirtualColumnContext(fieldSpec, _numDocsIndexed);
```
haibow: I'm not sure why `_numDocsIndexed + 1` is used - it's causing `MutableSegmentImplTest` to fail after adding virtual columns to mutable segments.

mcvsubbu:
> I have updated the diff to keep the virtualColumn as is since nothing new is introduced here. I will defer the renaming to future diffs where the usage of virtualColumn is expanded - hopefully then we can discuss with more context.

I was going to suggest this. It is is also easier to review

haibow:
> > I have updated the diff to keep the virtualColumn as is since nothing new is introduced here. I will defer the renaming to future diffs where the usage of virtualColumn is expanded - hopefully then we can discuss with more context.
> 
> I was going to suggest this. It is is also easier to review

I have done the change. Mind taking another look?

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
```diff
@@ -386,6 +388,12 @@ public ServerType getServerType() {
       public static String REPLICA_GROUP_SEGMENT_ASSIGNMENT_STRATEGY = "ReplicaGroupSegmentAssignmentStrategy";
     }
 
+    public static class VirtualColumn {
```
mcvsubbu:
```suggestion
    public static class BuiltInVirtualColumns {
```

mcvsubbu: IIRC you can add a virtual column in the schema with its own provider.
I am not sure if there is a restriction that virtual columns have to start with the "$" sign, but we can discuss if we can introduce that now.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
mcvsubbu: Hmm.. Does it work before your change? Can you find out why?

haibow: Previously virtual columns were not added to the mutable segments. I added them in this diff.

In `MutableSegmentImplTest` there is a test validating the total number of documents for each column (physical and virtual) in the mutable segments and immutable segment that both read from the same avro file. It was only checking physical columns before. After I add the virtual columns to the mutable segment, the test failed due to one extra row in the mutable segment, likely caused by this ` + 1`

haibow:
Left a reply regarding the latest comment in https://github.com/apache/incubator-pinot/pull/4774#discussion_r344419210. Somehow it's not showing under that comment.

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
Jackie-Jiang: Virtual column is already added to the schema in RealtimeTableDataManager

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
Jackie-Jiang: I think this `+1` is a bug. (numDocs is misused as index)

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/SingleStringVirtualColumnProvider.java
```diff
@@ -52,8 +53,7 @@ public Dictionary buildDictionary(VirtualColumnContext context) {
   public ColumnMetadata buildMetadata(VirtualColumnContext context) {
     ColumnMetadata.Builder columnMetadataBuilder = super.getColumnMetadataBuilder(context);
 
-    columnMetadataBuilder.setCardinality(1).setHasDictionary(true).setHasInvertedIndex(true)
-        .setFieldType(FieldSpec.FieldType.DIMENSION).setDataType(FieldSpec.DataType.STRING).setSingleValue(true)
+    columnMetadataBuilder.setCardinality(1).setHasDictionary(true).setHasInvertedIndex(true).setSingleValue(true)
```
Jackie-Jiang: Field type and data type is not required?

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/SingleStringVirtualColumnProvider.java
```diff
@@ -34,9 +33,11 @@
 /**
  * Virtual column provider for a virtual column that contains a single string.
  */
-public abstract class SingleStringVirtualColumnProvider extends BaseVirtualColumnProvider {
+public class SingleStringVirtualColumnProvider extends BaseVirtualColumnProvider {
 
-  protected abstract String getValue(VirtualColumnContext context);
+  protected String getValue(VirtualColumnContext context) {
+    return context.getFieldSpec().getDefaultNullValue().toString();
```
Jackie-Jiang: This does not seem right to me

haibow:
> I don't fully understand what problem this PR is trying to solve. Is there a link to the big picture of the problem you want to solve?
> As for the `SELECT *`, I think it make sense to only skip the column starts with '$'

Detailed design/usage is in https://github.com/apache/incubator-pinot/issues/4225#issuecomment-548173120. Currently we have 3 builtin virtual columns. If you look at the current `Virtualcolumncontext`, hostname and segmentname are manually included, and we need separate column providers for hostname and segmentname which simply provide a different single value. This only works for predefined virtual columns. In the future I would like to add virtual columns dynamically (for any new column added to the consuming segment) with custom column name, default value and other specs like single/multi-value, data type, etc.

The new `Virtualcolumncontext` will use fieldSpec which will contain all those specs. I've also merged `SegmentNameVirtualColumnProvider` and `HostNameVirtualColumnProvider` to `SingleStringVirtualColumnProvider` which will provide a single string for any single-value, string-type virtual column.

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/SingleStringVirtualColumnProvider.java
haibow: Common metadata are moved to `BaseVirtualColumnProvider` (`setFieldType(fieldSpec.getFieldType()).setDataType(fieldSpec.getDataType()`)

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/SingleStringVirtualColumnProvider.java
Jackie-Jiang: I know it works, but always returning default null value for SingleStringVirtualColumnProvider is quite confusing.
I think you want to support a virtual column type for default null value, you can add a new class DefaultNullValueVirtualColumnProvider for that purpose.

haibow:
> I don't think using VirtualColumnContext to pass values are good design in general. VirtualColumnProvider can be plugged in, while VirtualColumnContext is not pluggable. To me, a better design would be putting everything into the VirtualColumnProvider. What's your opinion?

I agree VirtualColumnContext was not pluggable enough. Previously all built in virtual columns were included, and now we have a better abstraction by using fieldSpec to support adding virtual columns dynamically. I lean towards keeping each refactoring incremental based on needs. In the future when there are other use cases of virtual columns, more refactoring can be done based on the actual pluggability requirement.

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/SingleStringVirtualColumnProvider.java
haibow: Added a new class DefaultNullValueSingleStringVirtualColumnProvider

pinot-common/src/main/java/org/apache/pinot/common/utils/CommonConstants.java
Jackie-Jiang: Can you move the comments for DEFAULT before the value for readability?

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
```diff
@@ -52,13 +52,13 @@
 import org.apache.pinot.core.realtime.impl.invertedindex.RealtimeInvertedIndexReader;
 import org.apache.pinot.core.realtime.impl.nullvalue.RealtimeNullValueVectorReaderWriter;
 import org.apache.pinot.core.segment.creator.impl.V1Constants;
+import org.apache.pinot.core.segment.virtualcolumn.VirtualColumnContext;
```
Jackie-Jiang: The import order change seems not correct

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/VirtualColumnProviderFactory.java
```diff
@@ -18,36 +18,47 @@
...
 
+import static org.apache.pinot.common.utils.CommonConstants.Segment.BuiltInVirtualColumn;
+
 
 /**
  * Factory for virtual column providers.
  */
 public class VirtualColumnProviderFactory {
-  public static VirtualColumnProvider buildProvider(String virtualColumnProvider) {
+  public static VirtualColumnProvider buildProvider(FieldSpec fieldSpec) {
```
Jackie-Jiang: This method should be reverted

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/DefaultNullValueSingleStringVirtualColumnProvider.java
```diff
@@ -19,11 +19,10 @@
 package org.apache.pinot.core.segment.virtualcolumn;
 
 /**
- * Virtual column provider that returns the current host name.
+ * Provide the default null value as a single string.
  */
-public class HostNameVirtualColumnProvider extends SingleStringVirtualColumnProvider {
-  @Override
+public class DefaultNullValueSingleStringVirtualColumnProvider extends SingleStringVirtualColumnProvider {
```
Jackie-Jiang: This class is good for this PR. In your following PRs, you might want to change this class into a general DefaultNullValueVirtualColumnProvider

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplTest.java
```diff
@@ -67,17 +69,20 @@ public void setUp()
     Assert.assertNotNull(resourceUrl);
     File avroFile = new File(resourceUrl.getFile());
 
-    SegmentGeneratorConfig config =
-        SegmentTestUtils.getSegmentGeneratorConfigWithoutTimeColumn(avroFile, TEMP_DIR, "testTable");
+    // Set the same default segment name for mutable and immutable segments
+    SegmentGeneratorConfig config = SegmentTestUtils
+        .getSegmentGeneratorConfigWithoutTimeColumn(avroFile, TEMP_DIR, MutableSegmentImplTestUtils.SEGMENT_NAME);
```
Jackie-Jiang: Thy changing table name to segment name?

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplTestUtils.java
```diff
@@ -35,8 +35,8 @@
   private MutableSegmentImplTestUtils() {
   }
 
-  private static final String SEGMENT_NAME = "testSegment";
   private static final String STEAM_NAME = "testStream";
+  public static final String SEGMENT_NAME = "testSegment";
```
Jackie-Jiang: Revert the change in this file?

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplTest.java
haibow: After adding virtual columns to the mutable segment (which caught the [bug with +1](https://github.com/apache/incubator-pinot/pull/4774/files#r344978233)), the test complained the default segment name did not match between the mutable and immutable segment. So I added the hack to make them share the same segment name. Now I've change the test to allow the default segment name to be different.

pinot-core/src/test/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImplTest.java
```diff
@@ -139,7 +145,9 @@ public void testDataSourceForSVColumns() {
 
           int actualDictId = actualSVIterator.nextIntVal();
           int expectedDictId = expectedSVIterator.nextIntVal();
-          Assert.assertEquals(actualDictionary.get(actualDictId), expectedDictionary.get(expectedDictId));
+          // Only allow the default segment name to be different
+          if (!column.equals(CommonConstants.Segment.BuiltInVirtualColumn.SEGMENTNAME))
```
Jackie-Jiang: (nit) add braces for if statement

Jackie-Jiang: Do you still need this after fixing the segment name?

haibow: added the braces. I abandoned the segment name change. Hope this will make it a bit easier to read.

pinot-core/src/main/java/org/apache/pinot/core/data/manager/realtime/RealtimeTableDataManager.java
```diff
@@ -226,7 +227,7 @@ public void addSegment(String segmentName, TableConfig tableConfig, IndexLoading
         _logger.error("Not adding segment {}", segmentName);
         throw new RuntimeException("Mismatching schema/table config for " + _tableNameWithType);
       }
-      VirtualColumnProviderFactory.addBuiltInVirtualColumnsToSchema(schema);
+      VirtualColumnProviderFactory.addBuiltInVirtualColumnsToSegmentSchema(schema, NetUtil.getHostnameOrAddress(), segmentName);
```
mcvsubbu:
```suggestion
      VirtualColumnProviderFactory.addBuiltInVirtualColumnsToSegmentSchema(schema, segmentName);
```
Given that this is a static function, why not let VirtualColumnProviderFactory determine the hostName (and perhaps anything else that may be needed as a part of being in cloud, for example)? It can do so on a static basis as well, instead of doing it for every segment.

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/DefaultNullValueSingleStringVirtualColumnProvider.java
mcvsubbu: +1

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/VirtualColumnContext.java
```diff
@@ -18,33 +18,23 @@
...
+
+
 /**
  * Context used to pass arguments to the virtual column provider.
  */
 public class VirtualColumnContext {
-  private String _hostname;
-  private String _segmentName;
-  private String _columnName;
+  private FieldSpec _fieldSpec;
```
mcvsubbu: This is not related to your PR, but since you are touching it and have some context, some more comments on what VirtualColumnContext does will be useful. To me, the class seems to be a miscellaneous class of all things needed for a virtual column that cannot be determined in other places? If you can add a better class description that will be awesome, thank you.

pinot-core/src/main/java/org/apache/pinot/core/segment/virtualcolumn/VirtualColumnProviderFactory.java
```diff
@@ -35,19 +37,20 @@ public static VirtualColumnProvider buildProvider(String virtualColumnProvider)
     }
   }
 
-  public static void addBuiltInVirtualColumnsToSchema(Schema schema) {
-    if (!schema.hasColumn("$docId")) {
-      schema.addField(new DimensionFieldSpec("$docId", FieldSpec.DataType.INT, true, DocIdVirtualColumnProvider.class));
+  public static void addBuiltInVirtualColumnsToSegmentSchema(Schema schema, String hostName, String segmentName) {
```
mcvsubbu: remove hostname from the argument, and determine the hostname statically in this class.

haibow:
@mcvsubbu could you check whether the revision looks good to you?

pinot-core/src/main/java/org/apache/pinot/core/indexsegment/mutable/MutableSegmentImpl.java
mcvsubbu: Ok, all the usage seems to refer to document count, which indeed should be numDocsIndexed and not numDocsIndexed+1. I think if we had more tests that actually look up inverted index, etc, on doc Id, it would have revealed the earlier bug. 
I did a quick test of running LLCClusterIntegrationTest and ran some queries against it 
SELECT count(*) from mytable where $segmentName = "provided the last consuming segment here"
SELECT max($docId) from mytable where $segmentName = "provided the last consuming segment here"

I did this on current master as well as this fork. It looks like the responses come out correctly.

As an aside, we don't get virtual columns  on any realtime segments other than consuming ones. It feels strange. I think we stopped adding virtual column to completed data sometime ago...

PR #4777, opened 2019-11-02
[TE] add event driven scheduler

vincentchenjl:
This PR is second one for event driven scheduling (data availability scheduling). It mainly includes the scheduler that runs in certain frequency to schedule tasks based on the event information updated by the event listener (in the first PR).

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
@@ -1,0 +1,217 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.pinot.thirdeye.anomaly.detection.trigger;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import org.apache.pinot.thirdeye.anomaly.task.TaskConstants;
+import org.apache.pinot.thirdeye.anomaly.task.TaskInfoFactory;
+import org.apache.pinot.thirdeye.anomaly.utils.ThirdeyeMetricsUtil;
+import org.apache.pinot.thirdeye.datalayer.bao.DatasetConfigManager;
+import org.apache.pinot.thirdeye.datalayer.bao.DetectionConfigManager;
+import org.apache.pinot.thirdeye.datalayer.bao.TaskManager;
+import org.apache.pinot.thirdeye.datalayer.dto.DatasetConfigDTO;
+import org.apache.pinot.thirdeye.datalayer.dto.DetectionConfigDTO;
+import org.apache.pinot.thirdeye.datalayer.dto.TaskDTO;
+import org.apache.pinot.thirdeye.datalayer.pojo.DetectionConfigBean;
+import org.apache.pinot.thirdeye.datasource.DAORegistry;
+import org.apache.pinot.thirdeye.detection.DetectionPipelineTaskInfo;
+import org.apache.pinot.thirdeye.formatter.DetectionConfigFormatter;
+import org.apache.pinot.thirdeye.rootcause.impl.MetricEntity;
+import org.apache.pinot.thirdeye.util.ThirdEyeUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is to schedule detection tasks based on data availability events.
+ */
+public class DataAvailabilityTaskScheduler implements Runnable {
+  private static final Logger LOG = LoggerFactory.getLogger(DataAvailabilityTaskScheduler.class);
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private ScheduledExecutorService executorService;
+  private long delayInSec;
+  private long fallBackTimeInSec;
+  private Map<Long, Long> taskLastCreateTimeMap;
+  private TaskManager taskDAO;
+  private DetectionConfigManager detectionConfigDAO;
+  private DatasetConfigManager datasetConfigDAO;
+  /**
+   * Construct an instance of {@link DataAvailabilityTaskScheduler}
+   * @param delayInSec delay after each run to avoid polling the database too often
+   * @param fallBackTimeInSec global threshold for fallback if detection level one is not set
+   */
+  public DataAvailabilityTaskScheduler(long delayInSec, long fallBackTimeInSec) {
+    this.delayInSec = delayInSec;
+    this.fallBackTimeInSec = fallBackTimeInSec;
+    this.taskLastCreateTimeMap = new HashMap<>();
+    this.executorService = Executors.newSingleThreadScheduledExecutor();
+    this.taskDAO = DAORegistry.getInstance().getTaskDAO();
+    this.detectionConfigDAO = DAORegistry.getInstance().getDetectionConfigManager();
+    this.datasetConfigDAO = DAORegistry.getInstance().getDatasetConfigDAO();
+  }
+
+  @Override
+  public void run() {
+    Map<DetectionConfigDTO, Set<String>> detection2DatasetMap = new HashMap<>();
+    Map<String, Long> dataset2RefreshTimeMap = new HashMap<>();
+    populateDetectionMapAndDataset2RefreshTimeMap(detection2DatasetMap, dataset2RefreshTimeMap);
+    Map<Long, TaskDTO> runningDetection = retrieveRunningDetectionTasks();
+    int taskCount = 0;
+    for (DetectionConfigDTO detectionConfig : detection2DatasetMap.keySet()) {
+      try {
+        long detectionConfigId = detectionConfig.getId();
+        if (!runningDetection.containsKey(detectionConfigId)) {
+          if (isAllDatasetUpdated(detectionConfig, detection2DatasetMap.get(detectionConfig), dataset2RefreshTimeMap)) {
+            //TODO: additional check is required if detection is based on aggregated value across multiple data points
+            createDetectionTask(detectionConfig);
+            ThirdeyeMetricsUtil.eventScheduledTaskCounter.inc();
+            taskCount++;
+          } else if (needFallback(detectionConfig)) {
+            LOG.info("Scheduling a task for detection {} due to the fallback mechanism", detectionConfigId);
+            createDetectionTask(detectionConfig);
+            ThirdeyeMetricsUtil.eventScheduledTaskFallbackCounter.inc();
+            taskCount++;
+          }
+        } else {
+          LOG.info("Skipping creating detection task for detection {} because task {} is not finished.",
+              detectionConfigId, runningDetection.get(detectionConfigId));
+        }
+      } catch (Exception e) {
+        LOG.error("Error in scheduling a detection...", e);
+      }
+    }
+    LOG.info("Schedule {} tasks in this run...", taskCount);
+  }
+
+  public void start() {
+    executorService.scheduleWithFixedDelay(this, 0, delayInSec, TimeUnit.SECONDS);
+  }
+
+  public void close() {
+    executorService.shutdownNow();
+  }
+
+  private void populateDetectionMapAndDataset2RefreshTimeMap(
+      Map<DetectionConfigDTO, Set<String>> dataset2DetectionMap, Map<String, Long> dataset2RefreshTimeMap) {
+    Map<Long, Set<String>> metricCache = new HashMap<>();
+    List<DetectionConfigDTO> detectionConfigs = detectionConfigDAO.findAllActive()
+        .stream().filter(DetectionConfigBean::isDataAvailabilitySchedule).collect(Collectors.toList());
+    for (DetectionConfigDTO detectionConfig : detectionConfigs) {
+      List<String> metricUrns = DetectionConfigFormatter
+          .extractMetricUrnsFromProperties(detectionConfig.getProperties());
+      Set<String> datasets = new HashSet<>();
+      for (String urn : metricUrns) {
+        MetricEntity me = MetricEntity.fromURN(urn);
+        if (!metricCache.containsKey(me.getId())) {
+          datasets.addAll(ThirdEyeUtils.getDatasetConfigsFromMetricUrn(urn)
+              .stream().map(DatasetConfigDTO::getName).collect(Collectors.toList()));
+          // cache the mapping in memory to avoid duplicate retrieval
+          metricCache.put(me.getId(), datasets);
+        } else {
+          // retrieve dataset mapping from memory
+          datasets.addAll(metricCache.get(me.getId()));
+        }
+      }
+      if (datasets.isEmpty()) {
+        LOG.error("No valid dataset is found for detection {}", detectionConfig.getId());
+        continue;
+      }
+      dataset2DetectionMap.put(detectionConfig, datasets);
+      for (String dataset : datasets) {
+        if (!dataset2RefreshTimeMap.containsKey(dataset)) {
+          DatasetConfigDTO datasetConfig = datasetConfigDAO.findByDataset(dataset);
+          dataset2RefreshTimeMap.put(dataset, datasetConfig.getLastRefreshTime());
+        }
+      }
+    }
+  }
+
+  private Map<Long, TaskDTO> retrieveRunningDetectionTasks() {
+    List<TaskConstants.TaskStatus> statusList = new ArrayList<>();
+    statusList.add(TaskConstants.TaskStatus.WAITING);
+    statusList.add(TaskConstants.TaskStatus.RUNNING);
+    List<TaskDTO> tasks = taskDAO.findByStatusesAndTypeWithinDays(statusList, TaskConstants.TaskType.DETECTION,
+        (int) TimeUnit.MILLISECONDS.toDays(ThirdEyeUtils.DETECTION_TASK_MAX_LOOKBACK_WINDOW));
+    Map<Long, TaskDTO> res = new HashMap<>(tasks.size());
+    for (TaskDTO task : tasks) {
+      res.put(ThirdEyeUtils.getDetectionIdFromJobName(task.getJobName()), task);
+    }
+    return res;
+  }
+
+  private long createDetectionTask(DetectionConfigDTO detectionConfig) throws JsonProcessingException {
+    long detectionConfigId = detectionConfig.getId();
+    // Make sure start time is not out of DETECTION_TASK_MAX_LOOKBACK_WINDOW
+    long end = System.currentTimeMillis();
+    long start = Math.max(detectionConfig.getLastTimestamp(),
+        end - ThirdEyeUtils.DETECTION_TASK_MAX_LOOKBACK_WINDOW);
+    DetectionPipelineTaskInfo taskInfo = new DetectionPipelineTaskInfo(detectionConfigId, start, end);
+    String taskInfoJson = OBJECT_MAPPER.writeValueAsString(taskInfo);
+    TaskDTO taskDTO = new TaskDTO();
+    taskDTO.setTaskType(TaskConstants.TaskType.DETECTION);
+    taskDTO.setJobName(TaskConstants.TaskType.DETECTION.toString() + "_" + detectionConfigId);
+    taskDTO.setStatus(TaskConstants.TaskStatus.WAITING);
+    taskDTO.setTaskInfo(taskInfoJson);
+    long taskId = taskDAO.save(taskDTO);
+    LOG.info("Created detection pipeline task {} with taskId {}", taskDTO, taskId);
+    taskLastCreateTimeMap.put(detectionConfigId, end);
+    return taskId;
+  }
+
+  private void loadLatestTaskCreateTime(DetectionConfigDTO detectionConfig) throws Exception {
+    long detectionConfigId = detectionConfig.getId();
+    List<TaskDTO> tasks = taskDAO.findByNameOrderByCreateTime(TaskConstants.TaskType.DETECTION.toString() +
+        "_" + detectionConfigId, 1,false);
+    if (tasks.size() == 0) {
+      taskLastCreateTimeMap.put(detectionConfigId, 0L);
+    } else {
+      DetectionPipelineTaskInfo taskInfo = (DetectionPipelineTaskInfo) TaskInfoFactory.getTaskInfoFromTaskType(
+          TaskConstants.TaskType.DETECTION, tasks.get(0).getTaskInfo());
+      taskLastCreateTimeMap.put(detectionConfigId, taskInfo.getEnd());
+    }
+  }
+
+  private boolean isAllDatasetUpdated(DetectionConfigDTO detectionConfig, Set<String> datasets,
+      Map<String, Long> dataset2RefreshTimeMap) {
+    long lastTimestamp = detectionConfig.getLastTimestamp();
+    return datasets.stream().allMatch(d -> dataset2RefreshTimeMap.get(d) >= lastTimestamp);
+  }
+
+  private boolean needFallback(DetectionConfigDTO detectionConfig) throws Exception {
+    long detectionConfigId = detectionConfig.getId();
+    long notRunThreshold = ((detectionConfig.getTaskTriggerFallBackTimeInSec() == 0) ?
+        fallBackTimeInSec : detectionConfig.getTaskTriggerFallBackTimeInSec()) * 1000;
+    if (!taskLastCreateTimeMap.containsKey(detectionConfigId)) {
+      loadLatestTaskCreateTime(detectionConfig);
+    }
+    long lastRunTime = taskLastCreateTimeMap.get(detectionConfigId);
+    return (System.currentTimeMillis() - lastRunTime >= notRunThreshold);
+  }
+}
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/utils/DataAvailabilitySchedulingConfiguration.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/utils/DataAvailabilitySchedulingConfiguration.java
@@ -21,11 +21,10 @@ package org.apache.pinot.thirdeye.anomaly.detection.trigger.utils;
 
 import java.util.List;
 
-
 /**
  * Configuration class for DataAvailabilityListener.
  */
-public class DataAvailabilityListenerConfiguration {
+public class DataAvailabilitySchedulingConfiguration {
   private String consumerClass;
   private String kafkaBootstrapServers;
   private String kafkaTopic;
@@ -37,6 +36,10 @@ public class DataAvailabilityListenerConfiguration {
   private long consumerPollTimeInMilli = 5_000; // consumer wait 5 secs by default for the buffer to be filled
   private List<String> dataSourceWhitelist;
   private List<String> filterClassList;
+  // delay time after each run for the scheduler to reduce DB polling
+  private long schedulerDelayInSec = 300;
+  // default threshold if detection level threshold is not set
+  private long taskTriggerFallBackTimeInSec = 24 * 60;
 
   public String getConsumerClass() {
     return consumerClass;
@@ -124,5 +127,21 @@ public class DataAvailabilityListenerConfiguration {
 
   public void setFilterClassList(List<String> filterClassList) {
     this.filterClassList = filterClassList;
+  }
+
+  public long getSchedulerDelayInSec() {
+    return schedulerDelayInSec;
+  }
+
+  public void setSchedulerDelayInSec(long schedulerDelayInSec) {
+    this.schedulerDelayInSec = schedulerDelayInSec;
+  }
+
+  public long getTaskTriggerFallBackTimeInSec() {
+    return taskTriggerFallBackTimeInSec;
+  }
+
+  public void setTaskTriggerFallBackTimeInSec(long taskTriggerFallBackTimeInSec) {
+    this.taskTriggerFallBackTimeInSec = taskTriggerFallBackTimeInSec;
   }
 }
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/task/TaskDriver.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/task/TaskDriver.java
@@ -123,7 +123,6 @@ public class TaskDriver {
                 TaskType taskType = anomalyTaskSpec.getTaskType();
                 TaskRunner taskRunner = TaskRunnerFactory.getTaskRunnerFromTaskType(taskType);
                 TaskInfo taskInfo = TaskInfoFactory.getTaskInfoFromTaskType(taskType, anomalyTaskSpec.getTaskInfo());
-                updateTaskStartTime(anomalyTaskSpec.getId());
                 Future<List<TaskResult>> future = taskExecutorService.submit(new Callable<List<TaskResult>>() {
                   @Override
                   public List<TaskResult> call() throws Exception {
@@ -211,9 +210,8 @@ public class TaskDriver {
 
           boolean success = false;
           try {
-            success = taskDAO
-                .updateStatusAndWorkerId(workerId, anomalyTaskSpec.getId(), allowedOldTaskStatus,
-                    TaskStatus.RUNNING, anomalyTaskSpec.getVersion());
+            success = taskDAO.updateStatusAndWorkerId(workerId, anomalyTaskSpec.getId(),
+                    allowedOldTaskStatus, anomalyTaskSpec.getVersion());
             LOG.info("Trying to acquire task id [{}], success status: [{}] with version [{}]",
                 anomalyTaskSpec.getId(), success, anomalyTaskSpec.getVersion());
           } catch (Exception e) {
@@ -247,17 +245,6 @@ public class TaskDriver {
       }
     }
     return null;
-  }
-
-  private void updateTaskStartTime(long taskId) {
-    LOG.info("Starting updateTaskStartTime for task id {}", taskId);
-    try {
-      long startTime = System.currentTimeMillis();
-      taskDAO.updateTaskStartTime(taskId, startTime);
-      LOG.info("Updated task start time {}", startTime);
-    } catch (Exception e) {
-      LOG.error("Exception in updating task start time", e);
-    }
   }
 
   private void updateStatusAndTaskEndTime(long taskId, TaskStatus oldStatus, TaskStatus newStatus, String message) {
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/bao/jdbc/TaskManagerImpl.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/bao/jdbc/TaskManagerImpl.java
@@ -52,6 +52,12 @@ public class TaskManagerImpl extends AbstractManagerImpl<TaskDTO> implements Tas
   private static final String FIND_BY_STATUS_ORDER_BY_CREATE_TIME_DESC =
       " WHERE status = :status order by startTime desc limit 10";
 
+  private static final String FIND_BY_NAME_ORDER_BY_CREATE_TIME_ASC =
+      " WHERE name = :name order by createTime asc limit ";
+
+  private static final String FIND_BY_NAME_ORDER_BY_CREATE_TIME_DESC =
+      " WHERE name = :name order by createTime desc limit ";
+
   private static final String COUNT_WAITING_TASKS =
       "SELECT COUNT(*) FROM task_index WHERE status = 'WAITING'";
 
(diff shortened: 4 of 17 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+            datasets.add(metricConfig.getDataset());
+          }
+          // cache the mapping in memory to avoid duplicate retrieval
+          metricCache.put(me.getId(), datasets);
+        } else {
+          // retrieve dataset mapping from memory
+          if (metricCache.get(me.getId()) != null)
+            datasets.addAll(metricCache.get(me.getId()));
+        }
+      }
```
xiaohui-sun: This logic is duplicated. Can we put it in a helper function and get it reused?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+
+  public void start() {
+    executorService.scheduleWithFixedDelay(this, 0, delayInSec, TimeUnit.SECONDS);
+  }
+
+  public void close() {
+    executorService.shutdownNow();
+  }
+
+  private void populateRequiredMetadata(Multimap<Set<String>, DetectionConfigDTO> dataset2DetectionMap,
```
xiaohui-sun: the name "populateRequiredMetadata" is too vague.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+    this.executorService = new ScheduledThreadPoolExecutor(1, r -> {
+      Thread t = Executors.defaultThreadFactory().newThread(r);
+      t.setDaemon(true);
+      return t;
+    });
+  }
+
+  @Override
+  public void run() {
+    Multimap<Set<String>, DetectionConfigDTO> dataset2DetectionMap = ArrayListMultimap.create();
```
xiaohui-sun: Instead of using Multimap<Set<String>, DetectionConfigDTO>, why not just using Map<DetectionConfigDTO, Set<String>> which is easier to understand and use?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+            if (!runningDetection.containsKey(detectionConfigId)) {
+              if (!lastTaskCreateTimeMap.containsKey(detectionConfigId)) loadLatestTaskCreateTime(detectionConfig);
+              long lastRunTime = lastTaskCreateTimeMap.get(detectionConfigId);
+              if (System.currentTimeMillis() - lastRunTime >= TimeUnit.MINUTES.toMillis(notRunThresholdInMin)) {
+                LOG.info("Scheduling a task for detection {}, since the create time of last task is {}",
+                    detectionConfigId, lastRunTime);
+                createDetectionTask(detectionConfig);
+              }
+            }
+          } catch (Exception e) {
```
xiaohui-sun: Suggest to refactor this code block. Split that into sub-methods which is easier to understand.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+          } else {
+            LOG.info("Skipping creating detection task for detection {} because task {} is not finished.",
+                detectionConfigId, runningDetection.get(detectionConfigId));
+          }
+        } else {
+          try {
+            if (!runningDetection.containsKey(detectionConfigId)) {
+              if (!lastTaskCreateTimeMap.containsKey(detectionConfigId)) loadLatestTaskCreateTime(detectionConfig);
+              long lastRunTime = lastTaskCreateTimeMap.get(detectionConfigId);
+              if (System.currentTimeMillis() - lastRunTime >= TimeUnit.MINUTES.toMillis(notRunThresholdInMin)) {
```
xiaohui-sun: This "notRunThresholdInMin" should be pulled from cron in each detectionConfig. It should not be a global setting.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+  private ScheduledThreadPoolExecutor executorService;
+  private long delayInSec;
+  private int notRunThresholdInMin;
+  private Map<Long, Long> lastTaskCreateTimeMap;
+
+  public DataAvailabilityTaskScheduler(long delayInSec, int notRunThresholdInMin ) {
+    this.delayInSec = delayInSec;
+    this.notRunThresholdInMin = notRunThresholdInMin;
+    this.lastTaskCreateTimeMap = new HashMap<>();
+    this.executorService = new ScheduledThreadPoolExecutor(1, r -> {
```
jihaozh: Use `Executors.newSingleThreadScheduledExecutor()`

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/pojo/DetectionConfigBean.java
```diff
@@ -125,6 +127,14 @@ public void setActive(boolean active) {
     this.active = active;
   }
 
+  public List<String> getLixFlags() {
```
jihaozh: Why this is called `LixFlags`? This seems wired to me. `Lix` is an internal concept and shouldn't be exposed to the opensource project. 

Also, looks like we are using it only for availability schedule for now, why not use a boolean value and call it something like`isEventDrivenSchedule`, which is more self-explanatory.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+  private void populateRequiredMetadata(Multimap<Set<String>, DetectionConfigDTO> dataset2DetectionMap,
+      Map<String, Long> datasetRefreshTimeMap) {
+    DetectionConfigManager detectionConfigManager = DAORegistry.getInstance().getDetectionConfigManager();
+    MetricConfigManager metricConfigManager = DAORegistry.getInstance().getMetricConfigDAO();
+    DatasetConfigManager datasetConfigManager = DAORegistry.getInstance().getDatasetConfigDAO();
+    Map<Long, Set<String>> metricCache = new HashMap<>();
+    List<DetectionConfigDTO> detectionConfigs = detectionConfigManager.findAll();
+    for (DetectionConfigDTO detectionConfig : detectionConfigs) {
+      if (!detectionConfig.isActive()) continue; // skip inactive detection
+      if (detectionConfig.getLixFlags().stream().filter(x -> x.equals(LIX_FLAG_VALUE)).count() < 1) continue;
```
jihaozh: Don't use `lixFlags`. Consider putting an explicit field called `isDataAvalabilitySchedule `

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+  public void run() {
+    Multimap<Set<String>, DetectionConfigDTO> dataset2DetectionMap = ArrayListMultimap.create();
+    Map<String, Long> datasetRefreshTimeMap = new HashMap<>();
+    populateRequiredMetadata(dataset2DetectionMap, datasetRefreshTimeMap);
+    Map<Long, TaskDTO> runningDetection = retrieveRunningDetection();
+    for (Set<String> datasets : dataset2DetectionMap.keySet()) {
+      for (DetectionConfigDTO detectionConfig : dataset2DetectionMap.get(datasets)) {
+        long detectionConfigId = detectionConfig.getId();
+        long lastTimestamp = detectionConfig.getLastTimestamp();
+        boolean allUpdated = datasets.stream().map(d -> datasetRefreshTimeMap.get(d) >= lastTimestamp)
```
jihaozh: Need to consider multi-day granularity here. For example, detecting on the weekly aggregated value  on daily data.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/task/TaskDriver.java
```diff
@@ -123,7 +123,6 @@ public void start() throws Exception {
                 TaskType taskType = anomalyTaskSpec.getTaskType();
                 TaskRunner taskRunner = TaskRunnerFactory.getTaskRunnerFromTaskType(taskType);
                 TaskInfo taskInfo = TaskInfoFactory.getTaskInfoFromTaskType(taskType, anomalyTaskSpec.getTaskInfo());
-                updateTaskStartTime(anomalyTaskSpec.getId());
```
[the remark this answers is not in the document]
vincentchenjl: I have it update the start time when it updates the worker id. https://github.com/apache/incubator-pinot/pull/4777/files/33c0ae859250442aa9585a6a6867f87114860323#diff-6f7e9ed6a4d01cdcba67ed4cfcd69cdaR125

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
vincentchenjl: Refactored part of it into `ThirdEyeUtils.getDatasetConfigsFromMetricUrn()`

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
vincentchenjl: Good point!

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
vincentchenjl: Added TODO, since we don't tackle this in this iteration.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/utils/DataAvailabilitySchedulingConfiguration.java
```diff
@@ -37,6 +37,8 @@
   private long consumerPollTimeInMilli = 5_000; // consumer wait 5 secs by default for the buffer to be filled
   private List<String> dataSourceWhitelist;
   private List<String> filterClassList;
+  private long schedulerDelayPerRunInSec = 300;
+  private int notRunThreshold = 60 * 24; // 1 day by default
```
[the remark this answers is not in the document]
vincentchenjl: Added comments.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.pinot.thirdeye.anomaly.detection.trigger;
```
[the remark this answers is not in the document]
vincentchenjl: I would say that we do it in the refactor PR, since there are still a lot of classes in use under `org.apache.pinot.thirdeye.anomaly`

thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskSchedulerTest.java
```diff
@@ -0,0 +1,214 @@
...
+    testDAOProvider = DAOTestBase.getInstance();
+    MetricConfigManager metricConfigManager = DAORegistry.getInstance().getMetricConfigDAO();
+    final String TEST_METRIC_PREFIX = "metric_trigger_scheduler_";
+
+    MetricConfigDTO metric1 = new MetricConfigDTO();
+    metric1.setDataset(TEST_DATASET_PREFIX + 1);
+    metric1.setName(TEST_METRIC_PREFIX + 1);
+    metric1.setActive(true);
+    metric1.setDerived(false);
+    metric1.setAlias("");
```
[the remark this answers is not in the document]
vincentchenjl: Does it matter here? This method are only used in this test class and it sets all the required field for tests to run.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+    statuses.add(TaskConstants.TaskStatus.WAITING);
+    statuses.add(TaskConstants.TaskStatus.RUNNING);
+    List<TaskDTO> tasks = taskManager.findByStatusesAndTypeWithinDays(statuses, TaskConstants.TaskType.DETECTION,
+        (int) TimeUnit.MILLISECONDS.toDays(ThirdEyeUtils.DETECTION_TASK_MAX_LOOKBACK_WINDOW));
+    Map<Long, TaskDTO> res = new HashMap<>(tasks.size());
+    for (TaskDTO task : tasks) {
+      String[] parts = task.getJobName().split("_");
+      if (parts.length < 2) {
+        LOG.error("Invalid name found {} for task id {}", task.getJobName(), task.getId());
+        continue;
```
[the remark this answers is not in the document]
vincentchenjl: I would say that keeping it relaxed is better because we may append more information to the `JobName` in the future. If that's the case, we don't need to change this line.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,229 @@
...
+    TaskManager taskManager = DAORegistry.getInstance().getTaskDAO();
+    List<TaskConstants.TaskStatus> statuses = new ArrayList<>();
+    statuses.add(TaskConstants.TaskStatus.WAITING);
+    statuses.add(TaskConstants.TaskStatus.RUNNING);
+    List<TaskDTO> tasks = taskManager.findByStatusesAndTypeWithinDays(statuses, TaskConstants.TaskType.DETECTION,
+        (int) TimeUnit.MILLISECONDS.toDays(ThirdEyeUtils.DETECTION_TASK_MAX_LOOKBACK_WINDOW));
+    Map<Long, TaskDTO> res = new HashMap<>(tasks.size());
+    for (TaskDTO task : tasks) {
+      String[] parts = task.getJobName().split("_");
+      if (parts.length < 2) {
```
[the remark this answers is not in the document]
vincentchenjl: I would say that keeping it relaxed is better because we may append more information to the JobName in the future. If that's the case, we don't need to change this line.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,216 @@
...
+      DetectionPipelineTaskInfo taskInfo = (DetectionPipelineTaskInfo) TaskInfoFactory.getTaskInfoFromTaskType(
+          TaskConstants.TaskType.DETECTION, tasks.get(0).getTaskInfo());
+      lastTaskCreateTimeMap.put(detectionConfigId, taskInfo.getEnd());
+    }
+  }
+
+  private boolean isAllDatasetUpdated(DetectionConfigDTO detectionConfig, Set<String> datasets,
+      Map<String, Long> dataset2RefreshTimeMap) {
+    long lastTimestamp = detectionConfig.getLastTimestamp();
+    return datasets.stream().map(d -> dataset2RefreshTimeMap.get(d) >= lastTimestamp)
```
jihaozh: Maybe use `datasets.stream().allMatch(d -> dataset2RefreshTimeMap.get(d) >= lastTimestamp)` is cleaner

vincentchenjl: Good point!

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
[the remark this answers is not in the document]
akshayrai: You need not move the existing classes. It you can move the newly added classes in this PR under the new package, it will be easier to clean up later

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+  public void start() {
+    executorService.scheduleWithFixedDelay(this, 0, delayInSec, TimeUnit.SECONDS);
+  }
+
+  public void close() {
+    executorService.shutdownNow();
+  }
+
+  private void populateDetectionMapAndDataset2RefreshTimeMap(
+      Map<DetectionConfigDTO, Set<String>> dataset2DetectionMap, Map<String, Long> dataset2RefreshTimeMap) {
```
xiaohui-sun: Can you refine the naming "dataset2DetectionMap" and "dataset2RefreshTimeMap" to 
"detectionDatasetMap" and "datasetRefreshTimeMap"?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+
+  public void start() {
+    executorService.scheduleWithFixedDelay(this, 0, delayInSec, TimeUnit.SECONDS);
+  }
+
+  public void close() {
+    executorService.shutdownNow();
+  }
+
+  private void populateDetectionMapAndDataset2RefreshTimeMap(
```
xiaohui-sun: The name is too long...
Can we split it into two functions?
1. initDetectionDatasetMap()
2. initDatasetRefreshTimeMap()

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+        if (!dataset2RefreshTimeMap.containsKey(dataset)) {
+          DatasetConfigDTO datasetConfig = datasetConfigDAO.findByDataset(dataset);
+          dataset2RefreshTimeMap.put(dataset, datasetConfig.getLastRefreshTime());
+        }
+      }
+    }
+  }
+
+  private Map<Long, TaskDTO> retrieveRunningDetection() {
+    List<TaskConstants.TaskStatus> statuses = new ArrayList<>();
```
xiaohui-sun: statuses -> statusList

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+      for (String dataset : datasets) {
+        if (!dataset2RefreshTimeMap.containsKey(dataset)) {
+          DatasetConfigDTO datasetConfig = datasetConfigDAO.findByDataset(dataset);
+          dataset2RefreshTimeMap.put(dataset, datasetConfig.getLastRefreshTime());
+        }
+      }
+    }
+  }
+
+  private Map<Long, TaskDTO> retrieveRunningDetection() {
```
xiaohui-sun: retrieveRunningDetectionTasks

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+    List<TaskConstants.TaskStatus> statuses = new ArrayList<>();
+    statuses.add(TaskConstants.TaskStatus.WAITING);
+    statuses.add(TaskConstants.TaskStatus.RUNNING);
+    List<TaskDTO> tasks = taskDAO.findByStatusesAndTypeWithinDays(statuses, TaskConstants.TaskType.DETECTION,
+        (int) TimeUnit.MILLISECONDS.toDays(ThirdEyeUtils.DETECTION_TASK_MAX_LOOKBACK_WINDOW));
+    Map<Long, TaskDTO> res = new HashMap<>(tasks.size());
+    for (TaskDTO task : tasks) {
+      String[] parts = task.getJobName().split("_");
```
xiaohui-sun: The hard-coded part is scaring.
This logic should be in the utils class so that could be shared.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+    taskDTO.setJobName(TaskConstants.TaskType.DETECTION.toString() + "_" + detectionConfigId);
+    taskDTO.setStatus(TaskConstants.TaskStatus.WAITING);
+    taskDTO.setTaskInfo(taskInfoJson);
+    long taskId = taskDAO.save(taskDTO);
+    LOG.info("Created detection pipeline task {} with taskId {}", taskDTO, taskId);
+    lastTaskCreateTimeMap.put(detectionConfigId, end);
+    return taskId;
+  }
+
+  private void loadLatestTaskCreateTime (DetectionConfigDTO detectionConfig) throws Exception {
```
xiaohui-sun: Remove space before (

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+      lastTaskCreateTimeMap.put(detectionConfigId, 0L);
+    } else {
+      DetectionPipelineTaskInfo taskInfo = (DetectionPipelineTaskInfo) TaskInfoFactory.getTaskInfoFromTaskType(
+          TaskConstants.TaskType.DETECTION, tasks.get(0).getTaskInfo());
+      lastTaskCreateTimeMap.put(detectionConfigId, taskInfo.getEnd());
+    }
+  }
+
+  private boolean isAllDatasetUpdated(DetectionConfigDTO detectionConfig, Set<String> datasets,
+      Map<String, Long> dataset2RefreshTimeMap) {
```
xiaohui-sun: dataset2RefreshTimeMap -> datasetRefreshTimeMap

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+    long lastTimestamp = detectionConfig.getLastTimestamp();
+    return datasets.stream().allMatch(d -> dataset2RefreshTimeMap.get(d) >= lastTimestamp);
+  }
+
+  private boolean isTimeForFallback(DetectionConfigDTO detectionConfig) throws Exception {
```
xiaohui-sun: isTimeForFallback -> needFallback

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/utils/DataAvailabilitySchedulingConfiguration.java
```diff
@@ -37,6 +36,8 @@
   private long consumerPollTimeInMilli = 5_000; // consumer wait 5 secs by default for the buffer to be filled
   private List<String> dataSourceWhitelist;
   private List<String> filterClassList;
+  private long schedulerDelayPerRunInSec = 300;
+  private int globalMaxNotRunThresholdInMin = 24 * 60;
```
xiaohui-sun: 1. Add comment on what it is.
2. rename to "taskTriggerFallBackTime"
3. Let's align everything in second

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/utils/DataAvailabilitySchedulingConfiguration.java
```diff
@@ -37,6 +36,8 @@
   private long consumerPollTimeInMilli = 5_000; // consumer wait 5 secs by default for the buffer to be filled
   private List<String> dataSourceWhitelist;
   private List<String> filterClassList;
+  private long schedulerDelayPerRunInSec = 300;
```
xiaohui-sun: schedulerDelayPerRunInSec -> schedulerDelay

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ * This class is to schedule detection tasks based on data availability events.
+ */
+public class DataAvailabilityTaskScheduler implements Runnable {
+  private static final Logger LOG = LoggerFactory.getLogger(DataAvailabilityTaskScheduler.class);
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private ScheduledExecutorService executorService;
+  private long delayInSec; // delay time after each run for the scheduler to reduce DB polling
+  private int maxNotRunThresholdInMin; // default threshold if detection level threshold is not set
```
xiaohui-sun: Can you put comments in separate lines?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+  private Map<Long, Long> lastTaskCreateTimeMap;
```
xiaohui-sun: lastTaskCreateTimeMap ->  taskLastCreateTimeMap

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+  private boolean isAllDatasetUpdated(DetectionConfigDTO detectionConfig, Set<String> datasets,
+      Map<String, Long> dataset2RefreshTimeMap) {
+    long lastTimestamp = detectionConfig.getLastTimestamp();
+    return datasets.stream().allMatch(d -> dataset2RefreshTimeMap.get(d) >= lastTimestamp);
+  }
+
+  private boolean isTimeForFallback(DetectionConfigDTO detectionConfig) throws Exception {
+    long detectionConfigId = detectionConfig.getId();
+    long notRunThreshold = ((detectionConfig.getMaxNotRunThresholdInMin() == 0) ?
+        maxNotRunThresholdInMin : detectionConfig.getMaxNotRunThresholdInMin()) * 60_000;
```
xiaohui-sun: No need to multiply 60_000 if we use seconds everywhere.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+        long detectionConfigId = detectionConfig.getId();
+        if (!runningDetection.containsKey(detectionConfigId)) {
+          if (isAllDatasetUpdated(detectionConfig, detection2DatasetMap.get(detectionConfig), dataset2RefreshTimeMap)) {
+            //TODO: additional check is required if detection is based on aggregated value across multiple data points
+            createDetectionTask(detectionConfig);
+            taskCount++;
+          } else if (isTimeForFallback(detectionConfig)) {
+            LOG.info("Scheduling a task for detection {} due to the fallback mechanism", detectionConfigId);
+            createDetectionTask(detectionConfig);
+            taskCount++;
```
xiaohui-sun: Can you track the two types of scheduled tasks separately?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
```diff
@@ -0,0 +1,222 @@
...
+          metricCache.put(me.getId(), datasets);
+        } else {
+          // retrieve dataset mapping from memory
+          datasets.addAll(metricCache.get(me.getId()));
+        }
+      }
+      if (datasets.isEmpty()) {
+        LOG.error("No valid dataset is found for detection {}", detectionConfig.getId());
+        continue;
+      }
```
xiaohui-sun: Let's not having the error check code everywhere.
When the datasets will be empty?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/bao/jdbc/TaskManagerImpl.java
```diff
@@ -103,6 +123,7 @@ public boolean updateStatusAndWorkerId(Long workerId, Long id, Set<TaskStatus> p
     if (permittedOldStatus.contains(task.getStatus())) {
       task.setStatus(newStatus);
       task.setWorkerId(workerId);
+      task.setStartTime(System.currentTimeMillis());
```
xiaohui-sun: Why setting startTime to current?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/pojo/DetectionConfigBean.java
```diff
@@ -44,6 +43,8 @@
   Map<String, Object> componentSpecs;
   long lastTuningTimestamp;
   List<String> owners;
+  boolean isDataAvailabilitySchedule;
```
xiaohui-sun: isEventBasedTriggering

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/ThirdEyeUtils.java
```diff
@@ -370,9 +372,19 @@ public static DatasetConfigDTO getDatasetConfigFromName(String dataset) {
...
 
-  public static String getDatasetFromMetricFunction(MetricFunction metricFunction) {
-    MetricConfigDTO metricConfig = getMetricConfigFromId(metricFunction.getMetricId());
-    return metricConfig.getDataset();
+  public static List<DatasetConfigDTO> getDatasetConfigsFromMetricUrn(String metricUrn) {
+    DatasetConfigManager datasetConfigManager = DAORegistry.getInstance().getDatasetConfigDAO();
+    MetricEntity me = MetricEntity.fromURN(metricUrn);
+    MetricConfigDTO metricConfig = DAORegistry.getInstance().getMetricConfigDAO().findById(me.getId());
+    if (metricConfig == null) return new ArrayList<>();
+    if (!metricConfig.isDerived()) {
```
xiaohui-sun: reverse if else.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/ThirdEyeUtils.java
```diff
@@ -370,9 +372,19 @@ public static DatasetConfigDTO getDatasetConfigFromName(String dataset) {
...
+      return Collections.singletonList(datasetConfigManager.findByDataset(metricConfig.getDataset()));
+    } else {
```
xiaohui-sun: no need else

xiaohui-sun:
@vincentchenjl Could you please check my comments?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
vincentchenjl: I think we should name them either `dataset2DetectionMap` or `datasetToDetectionMap` to avoid confusion.

vincentchenjl: I can rename it to `datasetToDetectionMap` if you prefer that.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/pojo/DetectionConfigBean.java
vincentchenjl: I will prefer the more specific name of the flag, because we might have other triggers in the future.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/ThirdEyeUtils.java
vincentchenjl: If the metrics is derived, we need to get all the datasets from the expression and skip using the value under `dataset` field in the `metricConfig`. If the metrics is not derived, we still need to add the `dataset` field in the `metricConfig` into the list. I am not sure that I can remove either the if-block or else-block.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/ThirdEyeUtils.java
vincentchenjl: Please see the comments below.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
vincentchenjl: Please see the first comment.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datalayer/bao/jdbc/TaskManagerImpl.java
vincentchenjl: Because this method is only used when workers pick up the task, we can update the start time within the same SQL statement. By doing so, we can ensure that all tasks picked up by workers have startTime and avoid additional DB call to update the start time.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
vincentchenjl: It is possible for the `datasets` to be empty. Since some of our metrics used by active detections are no longer valid, there is no associated dataset with those invalid metrics.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/utils/DataAvailabilitySchedulingConfiguration.java
vincentchenjl: I keep the unit naming convention, naming it to `schedulerDelayInSec`.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityTaskScheduler.java
vincentchenjl: We can split them but it only moves last six lines to another methods. Since the initialization of  `dataset2RefreshTimeMap` depends on the initialization of `detection2DatasetMap`, should we put them into the same method?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/ThirdEyeUtils.java
xiaohui-sun: Minor: No need to add "else" after "return".
if () {
return()
}
// The logic for else

PR #4778, opened 2019-11-02
Decoupling Server instance name from physical host and port

icefury71:
This PR modifies the way host and port information is discovered for a given Server. The original implementation assumes that instance name contains the host and port information. This is not ideal for cloud deployments. This PR instead uses the instance name to look up the host and port using a cache (populated using Helix). 

This will resolve the first problem referenced in this issue: https://github.com/apache/incubator-pinot/issues/4525

### DIFF
--- a/pinot-common/src/main/java/org/apache/pinot/common/helix/HelixInstanceConfigCache.java
+++ b/pinot-common/src/main/java/org/apache/pinot/common/helix/HelixInstanceConfigCache.java
@@ -65,7 +65,9 @@ public class HelixInstanceConfigCache implements ClusterChangeHandler {
 
   @Override
   public void init(HelixManager helixManager) {
-    Preconditions.checkState(_helixManager == null, "HelixInstanceConfigCache is already initialized");
+    if (_initialized) {
+      return;
+    }
     _helixManager = helixManager;
     _instanceNameToConfigMap = new HashMap<>();
     _initialized = true;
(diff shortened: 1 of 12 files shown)


pinot-common/src/main/java/org/apache/pinot/common/helix/HelixInstanceConfigCache.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public void init(HelixManager helixManager) {
+    Preconditions.checkState(_helixManager == null, "HelixInstanceConfigCache is already initialized");
+    _helixManager = helixManager;
+    _instanceNameToConfigMap = new HashMap<>();
+    _initialized = true;
+  }
+
+  @Override
+  public void processClusterChange(HelixConstants.ChangeType changeType) {
```
kishoreg: This might not be efficient when the size of the cluster is large. We already do this in an efficient way in routing table handler.

icefury71:
> I thought we could modify the existing routing table listener/handler. Let me check the code.

Yes we could (modify LiveInstanceChangeHandler). I did not in my first draft since I needed to make too many changes (create a singleton so that its easy to access in different places and have to move it to pinot-common).

kishoreg:
@icefury71 Better to change the getRoutingTable to returntype from Map<String, List<String>> to Map<ServerInstance, List<String>>. There is already a cache maintained in RoutingTable.

kishoreg:
@Jackie-Jiang fixed this

PR #4779, opened 2019-11-02
Adding superset demo example

fx19880617:
- Refactor docker images directory and adding superset images
- Adding superset demo example for AirlineStats table 

Sample Dashboard:
image


docker/images/superset/Dockerfile
```diff
@@ -0,0 +1,112 @@
...
+
+ARG NODE_VERSION=latest
+ARG PYTHON_VERSION=latest
+
+# --- Build assets with NodeJS
+
+FROM node:${NODE_VERSION} AS build
+
+# Superset version to build
+ARG SUPERSET_VERSION=master
```
kishoreg: maybe set this to the latest released version or use ENV variable

fx19880617: There is an ENV variable. Please see `README.md` for that.

PR #4780, opened 2019-11-03
[TE] initiate docker for ThirdEye 

nadcharin:


xiaohui-sun:
LGTM. Thanks for adding docker support for TE! @nadcharin

PR #4786, opened 2019-11-04
Add query options from json to existing query options, instead of overriding

npawar:
Query options can be set in 2 ways
1. Json payload with key "queryOptions"
2. Using ` option("key"="value")` at the end of the query.

If both are provided, we need to create a union of the options. We had a bug where the options from 1 would override 2 if both were provided.

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -25,6 +25,7 @@ import com.google.common.util.concurrent.RateLimiter;
 import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -231,9 +231,15 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
     }
 
     if (request.has(Broker.Request.QUERY_OPTIONS)) {
-      Map<String, String> queryOptions = getOptionsFromRequest(request, Broker.Request.QUERY_OPTIONS);
-      LOGGER.debug("Query options are set to: {} for request {}: {}", queryOptions, requestId, query);
-      brokerRequest.setQueryOptions(queryOptions);
+      Map<String, String> queryOptionsFromRequest = getOptionsFromRequest(request, Broker.Request.QUERY_OPTIONS);
```
kishoreg: its probably better to create another map queryOptions.

queryOptions = new HashMap();

if(queryOptionsFromRequest != null) {
 queryOptions.addAll(queryOptionsFromRequest)
}
if(queryOptionsFromSQL != null) {
 queryOptions.addAll(queryOptionsFromSQL)
}

brokerRequest.setQueryOptions(queryOptions);

npawar: refactored

npawar:
> Do we have a test case already?

Couldn't find any tests. Added.

PR #4794, opened 2019-11-06
[TE] frontend - harleyjj/homepage - homepage facelift and default to …

harleyjj:
…application tier0-tier1 if none chosen

1) On homepage, default will be application tier0-tier1 if none chosen, to prevent blank landing page
2) Repairs bug of edge case where anomaly endTime is at timestamp index of 0
3) Turns pill selectors into dropdown, except for custom time range selector
4) Design updates to home page
5) Implements spinner on homepage to notify user when waiting for anomaly results after selections

### DIFF
--- a/thirdeye/thirdeye-frontend/app/pods/home/index/route.js
+++ b/thirdeye/thirdeye-frontend/app/pods/home/index/route.js
@@ -52,6 +52,7 @@ export default Route.extend(AuthenticatedRouteMixin, {
     const { appName, startDate, endDate, duration, feedbackType, subGroup } = params;//check params
     const applications = await this.get('anomaliesApiService').queryApplications();// Get all applicatons available
     const subscriptionGroups = await this.get('anomaliesApiService').querySubscriptionGroups(); // Get all subscription groups available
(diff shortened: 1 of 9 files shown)


thirdeye/thirdeye-frontend/app/pods/home/index/route.js
```diff
@@ -61,18 +62,28 @@ export default Route.extend(AuthenticatedRouteMixin, {
...
+    let startDate = Number(model.startDate) || this.get('startDate');//TODO: we can use ember transform here
+    let endDate = Number(model.endDate) || this.get('endDate');
+    let duration = model.duration || this.get('duration');
     const feedbackType = model.feedbackType || this.get('feedbackType');
     const subGroup = model.subGroup || null;
+
+    // if there are no selections made, set to tier0-tier1 application by default
+    if (!appName && !subGroup) {
+      appName = 'tier0-tier1';
+      startDate = moment().startOf('day').subtract(1, 'week').utc().valueOf();
```
xiaohui-sun: Are we using utc()? I remember we pinned pst for frontend.

Issue #4525, opened 2019-08-13
Pinot server replacement in Cloud where we have no control over hostname


chenboat:
Pinot server instance ids are formatted as Server_<host>_<port>. These ids are then used in places like table external_view and idealstate. When Pinot brokers construct the routing tables for a query, they use the above instance ids to reconstruct the host/port info to send the Pinot query via HTTP protocol. 

Right now we try to deploy Pinot in cloud here in Uber. When we replace a machine, we can not get the same hostname of the replaced one. By the current impl, the server instance id will also change due to the changed hostname. That implies the new machine can not be a replacement for the old instance even it has all the data copied to it.

We wonder how machine replacement is done in a cloud environment. One proposal we think of is to decouple the server instance id with the underlying communication protocol (e.g., HTTP used for now). In this design, the broker should NOT use the server instance id directly and instead using it to look up from Zookeeper to find out the actual server address. In fact right now these are stored in zk as HELIX_HOST and HELIX_PORT but they are apparently not used in routing table construction. 

@fx19880617 @mcvsubbu

chenboat:
(1) Pinot broker should use the server instance id to look up the server address from Helix/Zk instead of parsing the id string like the following:

https://github.com/apache/incubator-pinot/blob/6e209e4e0ef138f8c21d597aeb20d1993fd1af2b/pinot-core/src/main/java/org/apache/pinot/core/transport/Server.java#L49-L53

(2) Pinot server should decouple the configuration of server Id with server host/port. Right now they are coupled:
https://github.com/apache/incubator-pinot/blob/3236e40c2f8d3d603c82e706db41bf2c0c2eaf68/pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java#L142-L151

chenboat:
Currently when a Pinot server starts and joins the Helix cluster, it calls HelixManager's connect() method:
https://github.com/apache/incubator-pinot/blob/3236e40c2f8d3d603c82e706db41bf2c0c2eaf68/pinot-server/src/main/java/org/apache/pinot/server/starter/helix/HelixServerStarter.java#L160

The above method in its turn call ParticipantManager.joinCluster()

https://github.com/apache/helix/blob/7ac3a6a1762bc74869a8f7ea0a9c7feed8b3ba0a/helix-core/src/main/java/org/apache/helix/manager/zk/ParticipantManager.java#L144-L156

[...]

PR #4795, opened 2019-11-06
Enhance RoutingTable to use ServerInstance  

kishoreg:
Pre-cursor to #4525.

Currently, RoutingTable is represented as Map<String, List<String>>. This is a map of serverInstanceName to list of segments on that server. The rest of the code derives the host and port by parsing serverInstanceName assuming that it's of the Server_host_port. 

This PR replaces serverInstanceName with a concrete ServerInstance object - no change in functionality.

In the next PR, we will construct the ServerInstance using the real host and port from the Helix Instance Config.

Testing: Ran quick start, offline, realtime and hybrid integration tests

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
@@ -53,6 +53,7 @@ import org.apache.pinot.common.request.FilterOperator;
 import org.apache.pinot.common.request.FilterQuery;
 import org.apache.pinot.common.request.FilterQueryMap;
 import org.apache.pinot.common.response.BrokerResponse;
+import org.apache.pinot.common.response.ServerInstance;
 import org.apache.pinot.common.response.broker.BrokerResponseNative;
 import org.apache.pinot.common.utils.CommonConstants;
 import org.apache.pinot.common.utils.CommonConstants.Broker;
@@ -261,8 +262,8 @@ public abstract class BaseBrokerRequestHandler implements BrokerRequestHandler {
 
     // Calculate routing table for the query
     long routingStartTimeNs = System.nanoTime();
-    Map<String, List<String>> offlineRoutingTable = null;
-    Map<String, List<String>> realtimeRoutingTable = null;
+    Map<ServerInstance, List<String>> offlineRoutingTable = null;
+    Map<ServerInstance, List<String>> realtimeRoutingTable = null;
     if (offlineBrokerRequest != null) {
       offlineRoutingTable = _routingTable.getRoutingTable(new RoutingTableLookupRequest(offlineBrokerRequest));
       if (offlineRoutingTable.isEmpty()) {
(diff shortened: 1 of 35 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/BaseBrokerRequestHandler.java
```diff
@@ -261,8 +262,8 @@ public BrokerResponse handleRequest(JsonNode request, @Nullable RequesterIdentit
 
     // Calculate routing table for the query
     long routingStartTimeNs = System.nanoTime();
-    Map<String, List<String>> offlineRoutingTable = null;
-    Map<String, List<String>> realtimeRoutingTable = null;
+    Map<ServerInstance, List<String>> offlineRoutingTable = null;
```
mcvsubbu: IIRC this used to be a map of ServerInstance to segments, and we changed it to be String to segments. Not sure what prompted the change. Perhaps @Jackie-Jiang may recall (memory usage?) DNS resolution? (although I see that it is done once in the static map now).

pinot-core/src/main/java/org/apache/pinot/core/transport/QueryRouter.java
```diff
@@ -55,25 +56,27 @@ public QueryRouter(String brokerId, BrokerMetrics brokerMetrics) {
...
       long timeoutMs) {
     assert offlineBrokerRequest != null || realtimeBrokerRequest != null;
 
     // Build map from server to request based on the routing table
     Map<Server, InstanceRequest> requestMap = new HashMap<>();
     if (offlineBrokerRequest != null) {
       assert offlineRoutingTable != null;
-      for (Map.Entry<String, List<String>> entry : offlineRoutingTable.entrySet()) {
-        Server server = new Server(entry.getKey(), TableType.OFFLINE);
+      for (Map.Entry<ServerInstance, List<String>> entry : offlineRoutingTable.entrySet()) {
```
mcvsubbu: You are probably planning to do that in the next step, but does it make sense to get rid of the Server object, given that ServerInstance has everything that Server has (and calls resolve by host name everytime it needs to log ?)

kishoreg: yes, that will be the next step. Any idea on why we have seq in ServerInstance?

mcvsubbu: seq predates me. You approved it :)

```
commit e29933810a701e2e8c4690465ccb8774a0324180
Author: Xiang Fu <xiafu@linkedin.com>
Date:   Sun Mar 29 00:36:11 2015 -0700

    Bug fixing for federated broker test.
    1. fix logic for RealtimeColumnDataSource set range predicate as for
    realtime dictionary, dictId is from 1 to dictionary size.
    2. in MutableDictionaryReader, getString is also used for other data
    type dictionary, so change it to toString().
    3. fixing  offline ColumnDataSourceImpl would set filteredBitMap to null
    in Range query, which actually should be an empty bitMap.
    This is the root cause of DummyBlock get called.
    4. fixing NPE in HelixExternalViewBasedTimeBoundaryService.
    5. fixing time filter attching logic in BrokerRequestHandler for no
    filter request.
    6. changing attached time filters' ids to -1 and -2 to avoid potiential
    duplicated.
    
    RB=453859
    R=xiafu
    A=kgopalak,jfim
```

Jackie-Jiang: We need seqId to differentiate OFFLINE table response and REALTIME table response, which is very hard to understand.. And that is the main reason why I made a new class and use TableType instead. Also, we don't need to calculate the shortHostName when routing the queries

Jackie-Jiang:
Replaced with #4823

PR #4798, opened 2019-11-07
Decouple Key from Record

npawar:
Separating out the `Key` from the `Record`.
Record now has just `Object[] _columns`, which contains all the columns, with no distinction between keys/values. This is an attempt to make the Record more generic, and remove the aggregation/group by specific behavior.

One by product of removing `Key` from `Record` is that the upsert function in the `IndexedTable` now has to extract the key on its own.
Introduced a new `upsert(Key, Record)`, which will mainly be used by Group By. This is to avoid extracting the `Key` twice (we already extract Key in CombineGroupByOperator. If we simply call`upsert(Record)`, we have to extract Key again.
We still need `upsert(Record)`, for 
1) even group by cannot always have the Key e.g. merge calls upsert, broker side reduce
2) other implementations of Table

The Interface hierarchy now looks like:
Table -> BaseTable -> IndexedTable
where `IndexedTable` is used for a key/value style table. 
We will soon add other tables for Distinct, Selection, Aggregations. On a high level, we need the following flavors of Table:
1) Map Based - needed for Group By. This is the current` IndexedTable`
2) Set Based - needed for Distinct
3) List Based - needed for Selection
4) Single Record - needed for Aggregations
We could perhaps name them as such, once we get more implementations.


siddharthteotia:
LGTM

So all the specific implementation can extend BaseTable and go from there.

One thing that remains to be seen is if it is a good idea to pollute TableResizer with overloaded methods to work for both Map<Key,Record> (for group by) or Set<Record> (for distinct) or should we subclass it (to borrow maximum resizing functionality) or write another version of it.

Since resizer will be used only by group by and distinct code, probably adding multiple methods that take different data structures should be fine.  I can refactor that as part of my pr that adds order by support for distinct -- https://github.com/apache/incubator-pinot/pull/4790

pinot-core/src/main/java/org/apache/pinot/core/data/table/Record.java
```diff
@@ -22,33 +22,24 @@
 
 
 /**
- * Defines a single record in Pinot comprising of keys and values
+ * Defines a single record in Pinot
  */
 public class Record {
-  private Key _key;
-  private Object[] _values;
+  private Object[] _columns;
```
Jackie-Jiang: Recommend renaming this to `_values`, `_columns` sounds like column names

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -18,75 +18,55 @@
...
-  private DataSchema _dataSchema;
-
-  // the capacity we need to trim to
-  int _capacity;
-  // the capacity with added buffer, in order to collect more records than capacity for better precision
-  int _maxCapacity;
-
-  boolean _isOrderBy;
-  IndexedTableResizer _indexedTableResizer;
+  private KeyExtractor _keyExtractor;
```
Jackie-Jiang: final

pinot-core/src/main/java/org/apache/pinot/core/data/table/BaseTable.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.pinot.core.query.aggregation.function.AggregationFunction;
+import org.apache.pinot.core.query.aggregation.function.AggregationFunctionUtils;
+
+
+/**
+ * Base abstract implementation of Table
+ */
+public abstract class BaseTable implements Table {
+
+  AggregationFunction[] _aggregationFunctions;
```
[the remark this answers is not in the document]
npawar: fixed all the final

pinot-core/src/main/java/org/apache/pinot/core/data/table/Record.java
npawar: done

pinot-core/src/main/java/org/apache/pinot/core/data/table/BaseTable.java
```diff
@@ -0,0 +1,93 @@
...
+  // the capacity with added buffer, in order to collect more records than capacity for better precision
+  int _maxCapacity;
+
+  boolean _isOrderBy;
+  TableResizer _tableResizer;
+
+  /**
+   * Initializes the variables and comparators needed for the table
+   */
+  BaseTable(DataSchema dataSchema, List<AggregationInfo> aggregationInfos, List<SelectionSort> orderBy, int capacity) {
```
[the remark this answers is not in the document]
npawar: Every Table implementation will extend BaseTable. IndexedTable is the one which will be used only by AggregationGroupBy. We can rename that. But I would like that to be done in the future changes, once we have at least one more implementation (which should be soon in @siddharthteotia 's DISTINCT related changes).
At that time, we could revisit the Constructor of BaseTable, to remove irrelevant things (potentially keep only dataSchema and capacity).

pinot-core/src/main/java/org/apache/pinot/core/data/table/BaseTable.java
```diff
@@ -0,0 +1,93 @@
...
+      _maxCapacity = capacity;
+    }
+    _capacity = capacity;
+  }
+
+  @Override
+  public boolean merge(Table table) {
+    Iterator<Record> iterator = table.iterator();
+    while (iterator.hasNext()) {
+      upsert(iterator.next());
```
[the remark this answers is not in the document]
npawar: I don't see how that can be done.. We will have to make use of the KeyExtractor in any case.

npawar:
> LGTM
> 
> So all the specific implementation can extend BaseTable and go from there.
> 
> One thing that remains to be seen is if it is a good idea to pollute TableResizer with overloaded methods to work for both Map<Key,Record> (for group by) or Set (for distinct) or should we subclass it (to borrow maximum resizing functionality) or write another version of it.
> 
> Since resizer will be used only by group by and distinct code, probably adding multiple methods that take different data structures should be fine. I can refactor that as part of my pr that adds order by support for distinct -- #4790

Adding a `resizeRecordsSet` to the `TableResizer` sounds fine to me as a starting point. We can revisit that when we begin to have other operations needing resize (selection), or if adding `resizeRecordsSet` looks very messy.
One thing that will be good to do in your follow-up PR - rename the `IndexedTable` to say `MapBasedTable` (or AggregationGroupByTable), and the new one you add can be `SetBasedTable`. 
Another point as noted by Jackie - the constructor for `BaseIndexedTable` takes in things which won't be relevant to all its implementations. It will be good to clean that up as well, when we have another implementation.

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
```diff
@@ -18,75 +18,55 @@
...
+
+    KeyExtractor(int keyIndexes) {
+      _keyIndexes = keyIndexes;
+    }
+
+    /**
+     * Returns the Key from the Record
+     */
+    Key extractKey(Record record) {
+      Object[] keys = Arrays.copyOf(record.getValues(), _keyIndexes);
```
kishoreg: can we not make Key a lightweight wrapper around record? not needed for this PR but something to revisit later

pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
```diff
@@ -39,13 +39,13 @@
 /**
  * Helper class for trimming and sorting records in the IndexedTable, based on the order by information
  */
-class IndexedTableResizer {
+class TableResizer {
```
kishoreg: will moving this into a separate class outside IndexTable make it harder to optimize later?

pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
```diff
@@ -112,12 +112,12 @@
    * For aggregation values in the order by, the final result is extracted if the intermediate result is non-comparable
    */
   @VisibleForTesting
-  IntermediateRecord getIntermediateRecord(Record record) {
+  IntermediateRecord getIntermediateRecord(Key key, Record record) {
```
kishoreg: why do we need this method who will invoke it.

npawar: while doing the resize, the Record gets converted to IntermediateRecord. IntermediateRecord contains only the columns necessary for order by. It helps prevent array index lookup during order by.
Now that we have gotten rid of Key from Record, we might be able to do this using just the Record. Can revisit it after this

pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
npawar: In this PR, the changes are simply renaming IndexTableResizer to TableResizer.
Pulling out the resize logic to a separate class was done previously. It actually helped in being able to reuse the same logic for both of the IndexedTable implementations. This same class will be used by the Table implementation for distinct. As of now I don't see why it will become harder to optimize. What were you thinking of in particular?

pinot-core/src/main/java/org/apache/pinot/core/data/table/ConcurrentIndexedTable.java
```diff
@@ -68,17 +68,17 @@ public ConcurrentIndexedTable(DataSchema dataSchema, List<AggregationInfo> aggre
...
-    Key key = newRecord.getKey();
     Preconditions.checkNotNull(key, "Cannot upsert record with null keys");
 
     if (_noMoreNewRecords.get()) { // allow only existing record updates
       _lookupMap.computeIfPresent(key, (k, v) -> {
         Object[] existingValues = v.getValues();
         Object[] newValues = newRecord.getValues();
-        for (int i = 0; i < _numAggregations; i++) {
-          existingValues[i] = _aggregationFunctions[i].merge(existingValues[i], newValues[i]);
+        int aggNum =0;
```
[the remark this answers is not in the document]
npawar: i think it appears so because of the deleted lines.
I reformatted to check.

pinot-core/src/main/java/org/apache/pinot/core/data/table/IndexedTable.java
npawar: yes we could do that. But we don't want Key to be only constructed from Record right. FOr example, in CombineGroupByOperator, we already have the Object[] keys array, and we can just directly create Key(Object[] keys)

pinot-core/src/main/java/org/apache/pinot/core/data/table/BaseTable.java
```diff
@@ -0,0 +1,94 @@
...
+    _isOrderBy = CollectionUtils.isNotEmpty(orderBy);
+    if (_isOrderBy) {
+      _tableResizer = new TableResizer(dataSchema, aggregationInfos, orderBy);
+
+      // TODO: tune these numbers
+      // Based on the capacity and maxCapacity, the resizer will smartly choose to evict/retain recors from the PQ
+      if (capacity <= 100_000) { // Capacity is small, make a very large buffer. Make PQ of records to retain, during resize
+        _maxCapacity = 1_000_000;
+      } else { // Capacity is large, make buffer only slightly bigger. Make PQ of records to evict, during resize
+        _maxCapacity = (int) (capacity * 1.2);
```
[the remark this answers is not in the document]
npawar: yes.. it's pretty naive right now. Will be revisiting this to come up with a better formula.
This code is actually not new to this PR, just moved from IndexedTable. I have a TODO in place for this part

pinot-core/src/main/java/org/apache/pinot/core/data/table/BaseTable.java
```diff
@@ -0,0 +1,94 @@
...
+  // the capacity with added buffer, in order to collect more records than capacity for better precision
+  final int _maxCapacity;
+
+  final boolean _isOrderBy;
+  final TableResizer _tableResizer;
+
+  /**
+   * Initializes the variables and comparators needed for the table
+   */
+  BaseTable(DataSchema dataSchema, List<AggregationInfo> aggregationInfos, List<SelectionSort> orderBy, int capacity) {
```
mayankshriv: If we plan to use BaseTable for selection as well, then we won't have aggregationInfo there? In that case, we may need to revisit all places where we are assuming presence of aggregationInfo in this PR.

pinot-core/src/main/java/org/apache/pinot/core/data/table/BaseTable.java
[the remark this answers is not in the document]
mayankshriv: Could you create a github issue and link it here?

pinot-core/src/main/java/org/apache/pinot/core/data/table/BaseTable.java
npawar: Yes, that's right. We discussed on similar lines in this comment: https://github.com/apache/incubator-pinot/pull/4798#discussion_r343405515
This revisit and cleanup can be done as soon as we get another implementation for the table i.e. in the Distinct PR

pinot-core/src/main/java/org/apache/pinot/core/data/table/BaseTable.java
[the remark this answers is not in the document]
npawar: Done

pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
npawar: Merging this PR, to unblock the Distinct Order By PR.
Will address any specific concerns about this point in separate PRs if needed

npawar:

PR #4803, opened 2019-11-08
[TE] Set non additive flag for Pinot dataset

jihaozh:
The nonadditive flag was not set for Pinot non-additive dataset when importing the metric, which can cause the value to be wrong in ThirdEye. This PR fixes the issue.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/ConfigGenerator.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/ConfigGenerator.java
@@ -41,6 +41,8 @@ public class ConfigGenerator {
 
   private static final String PDT_TIMEZONE = "US/Pacific";
   private static final String BYTES_STRING = "BYTES";
+  private static final String NON_ADDITIVE = "non_additive";
+  private static final String PINOT_PRE_AGGREGATED_KEYWORD = "*";
 
   public static void setTimeSpecs(DatasetConfigDTO datasetConfigDTO, TimeGranularitySpec timeSpec) {
     datasetConfigDTO.setTimeColumn(timeSpec.getName());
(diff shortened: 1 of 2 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/auto/onboard/ConfigGenerator.java
```diff
@@ -92,6 +95,16 @@ private static TimeGranularity getExpectedDelayFromTimeunit(TimeUnit timeUnit) {
     return expectedDelay;
   }
 
+  /**
+   * Check if the dataset is non-additive. If it is, set the additive flag to false and set the pre-aggregated keyword.
+   * @param dataset the dataset DTO to check
+   */
+  static void checkNonAdditive(DatasetConfigDTO dataset) {
+    if (dataset.isAdditive() && dataset.getDataset().endsWith(NON_ADDITIVE)) {
```
xiaohui-sun: I don't understand the logic here.
So by default the dataset is additive()?
If so then why we need this dataset.isAdditive() check?
If it is one-time fix suggest not to put it into the onboard logic.

jihaozh: Actually, it's not a one-time fix. The logic a bit tricky here.

When creating a `new DatasetConfigDTO()`, the additive is `true`. Then we check if it's a non_additive dataset, we set the flag to be `false` and set the keyword to be `*`.

When checking an existing  `DatasetConfigDTO`, if additive = `true` for a non_additive dataset, we set the flag to be `false` and set the keyword to be `*`.  But if additive = `false` for a non_additive dataset, we leave the flag and the keyword unchanged. If remove the `dataset.isAdditive()` check, we might override the `PreAggregatedKeyword`. For some old dataset, the keyword is set to `ALL` instead of `*`.

PR #4806, opened 2019-11-09
Refactor segments admin rest APIs

Jackie-Jiang:
- Segment toggle APIs are removed (backward-incompatible change)
- List all segments in cluster API is removed (backward-incompatible change)
- Move the APIs under the Segment API tag
- Refactor the APIs to match the uri semantic

GET requests:
- "/segments/{tableName}": get the name of all segments
- "/segments/{tableName}/servers": get a map from server to segments hosted by the server
- "/segments/{tableName}/crc": get a map from segment to CRC of the segment (OFFLINE table only)
- "/segments/{tableName}/{segmentName}/metadata: get the metadata for a segment

POST requests:
- "/segments/{tableName}/{segmentName}/reload": reload a segment
- "/segments/{tableName}/reload": reload all segments

DELETE requests:
- "/segments/{tableName}/{segmentName}": delete a segment
- "/segments/{tableName}: delete all segments

Deprecated APIs:
- "GET /tables/{tableName}/segments"
- "GET /tables/{tableName}/segments/metadata"
- "GET /tables/{tableName}/segments/crc"
- "GET /tables/{tableName}/segments/{segmentName}"
- "GET /tables/{tableName}/segments/{segmentName}/metadata"
- "GET /tables/{tableName}/segments/{segmentName}/reload"
- "POST /tables/{tableName}/segments/{segmentName}/reload"
- "GET /tables/{tableName}/segments/reload"
- "POST /tables/{tableName}/segments/reload"

Deprecated APIs are subject to be removed in the next release
All APIs are manually tested

Design doc: https://cwiki.apache.org/confluence/display/PINOT/Segment+Admin+Rest+APIs


pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
-      @ApiParam(value = "enable|disable|drop", required = false) @QueryParam("state") String stateStr,
-      @ApiParam(value = "realtime|offline", required = false) @QueryParam("type") String tableTypeStr) {
-    CommonConstants.Helix.TableType tableType = Constants.validateTableType(tableTypeStr);
-    StateType state = Constants.validateState(stateStr);
+      @ApiParam(value = "OFFLINE|REALTIME") @QueryParam("type") String tableTypeStr) {
+    List<String> tableNamesWithType =
+        getExistingTableNamesWithType(tableName, Constants.validateTableType(tableTypeStr));
+    List<Map<TableType, List<String>>> resultList = new ArrayList<>(tableNamesWithType.size());
+    for (String tableNameWithType : tableNamesWithType) {
+      TableType tableType = TableNameBuilder.getTableTypeFromTableName(tableNameWithType);
```
mcvsubbu: If the table type is specified in the input, then we should list only the segments of that table type

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+    }
+    return resultList;
+  }
 
-    if (stateStr == null) {
-      return getInstanceToSegmentsMap(tableName, tableType);
+  @GET
+  @Path("segments/{tableName}/servers")
+  @Produces(MediaType.APPLICATION_JSON)
+  @ApiOperation(value = "Get a map from server to segments", notes = "Get a map from server to segments")
```
mcvsubbu:
```suggestion
  @ApiOperation(value = "Get a list of segments hosted by each server", notes = "Get a list of segments hosted by each server")
```

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+  public List<Map<String, Object>> getServerToSegmentsMap(
+      @ApiParam(value = "Name of the table", required = true) @PathParam("tableName") String tableName,
+      @ApiParam(value = "OFFLINE|REALTIME") @QueryParam("type") String tableTypeStr) {
```
mcvsubbu: May be useful to take an additional optional server name as an argument, and then filter by server name in the output.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: `getExistingTableNamesWithType` will only return the existing tables that match the given table name and type

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: This API is a replacement of the deprecated APIs. I'll add more APIs in the following PRs, and if we find it is useful to filter by server name, we can add another API and put serverName as path parameter.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: Changed to "Get a map from server to segments hosted by the server". Want to emphasize that the result for each table is a map.

Jackie-Jiang:
@mcvsubbu 

> Did you re-order the methods? I see signatures changed, but marked as deprecated.

Several methods are moved from `PinotSegmentUploadRestletResource` to `PinotSegmentRestletResource`. The current method name does not match the actual semantic of the method, so I changed the method name to match the actual semantic to make them easier to understand.

> (1) Since this is a backward incompatible change (removing some APIs, right?) please be sure to flag it in the commit message so that we can add it correctly in the release notes.

I didn't remove any API, but only mark them deprecated. The only backward incompatible change is that we no longer allow toggling segment state (including drop, drop should be done by DELETE instead of GET)

> (2) Since this is an API change, please open a cwiki page with the proposed new APIs (and changes and deprecations) and let the community comment on it for sometime before merging this change. Be sure to add a link to that page in the commit message.

Will do

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
-        List<String> offlineSegments =
-            _pinotHelixResourceManager.getSegmentsFor(TableNameBuilder.OFFLINE.tableNameWithType(tableName));
-        if (realtimeSegments.contains(segmentName)) {
-          tableType = CommonConstants.Helix.TableType.REALTIME;
-        } else if (offlineSegments.contains(segmentName)) {
-          tableType = CommonConstants.Helix.TableType.OFFLINE;
-        }
-      }
-      return toggleStateInternal(tableName, stateType, tableType, segmentName, _pinotHelixResourceManager);
+      @ApiParam(value = "MUST be null") @QueryParam("state") String stateStr,
```
mcvsubbu: required = false?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+    return getSegmentToCrcMap(tableName);
+  }
+
+  @GET
+  @Path("segments/{tableName}/{segmentName}/metadata")
   @Produces(MediaType.APPLICATION_JSON)
-  @ApiOperation(value = "Lists one segment's metadata", notes = "Lists one segment's metadata")
-  public String listMetadataForOneSegment(
+  @ApiOperation(value = "Get the metadata for a segment", notes = "Get the metadata for a segment")
+  public List<Map<String, Object>> getSegmentMetadata(
```
mcvsubbu: Not sure why we are returning a list of maps. A segment should only belong to one type of table. If the user specifies the table type, we can use that. Alternatively, we can derive the table type from the segment name and remove the table type argument altogether.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+      RealtimeSegmentZKMetadata realtimeSegmentZKMetadata =
+          ZKMetadataProvider.getRealtimeSegmentZKMetadata(propertyStore, tableNameWithType, segmentName);
+      return realtimeSegmentZKMetadata != null ? realtimeSegmentZKMetadata.toMap() : null;
+    }
+  }
 
-    ArrayNode result = JsonUtils.newArrayNode();
+  @Deprecated
+  @GET
+  @Path("tables/{tableName}/segments/{segmentName}/metadata")
```
mcvsubbu: Incompatible with existing API. Currently, this API returns a String. Please keep the api the same.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+  @Deprecated
   @GET
   @Path("tables/{tableName}/segments/metadata")
   @Produces(MediaType.APPLICATION_JSON)
-  @ApiOperation(value = "Lists metadata all segments of table", notes = "Lists segment metadata")
-  // TODO: 1. stateStr is not used
-  // TODO: 2. this method returns the map from server to segments instead of segment metadata
-  public String listMetadataForAllSegments(
+  @ApiOperation(value = "Get a map from server to segments hosted by the server (deprecated, use 'GET /segments/{tableName}/servers' instead)", notes = "Get a map from server to segments hosted by the server (deprecated, use 'GET /segments/{tableName}/servers' instead)")
+  public List<Map<String, String>> getServerToSegmentsMapDeprecated2(
```
mcvsubbu: This API currently returns a String. please keep the return type the same, we can change the types in the new APIs that you have added.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+    String offlineTableName = getExistingTableNamesWithType(tableName, TableType.OFFLINE).get(0);
+    return _pinotHelixResourceManager.getSegmentsCrcForTable(offlineTableName);
+  }
+
+  @Deprecated
+  @GET
+  @Path("tables/{tableName}/segments/crc")
+  @Produces(MediaType.APPLICATION_JSON)
+  @ApiOperation(value = "Get a map from segment to CRC of the segment (deprecated, use 'GET /segments/{tableName}/crc' instead)", notes = "Get a map from segment to CRC of the segment (deprecated, use 'GET /segments/{tableName}/crc' instead)")
+  public Map<String, String> getSegmentToCrcMapDeprecated(
```
mcvsubbu: Please keep current return type.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
-      if (tableType != null) {
-        errorMessage += " of type: " + tableType;
-      }
-      throw new ControllerApplicationException(LOGGER, errorMessage, Response.Status.NOT_FOUND);
+  @Deprecated
+  @GET
+  @Path("tables/{tableName}/segments/{segmentName}")
+  @Produces(MediaType.APPLICATION_JSON)
+  @ApiOperation(value = "Get the metadata for a segment (deprecated, use 'GET /segments/{tableName}/{segmentName}/metadata' instead)", notes = "Get the metadata for a segment (deprecated, use 'GET /segments/{tableName}/{segmentName}/metadata' instead)")
+  public List<List<Map<String, Object>>> getSegmentMetadataDeprecated2(
```
mcvsubbu: return type should be String

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
   @Produces(MediaType.APPLICATION_JSON)
-  @ApiOperation(value = "Reloads one segment", notes = "Reloads one segment")
-  public SuccessResponse reloadOneSegment(
+  @ApiOperation(value = "Reload a segment", notes = "Reload a segment")
+  public SuccessResponse reloadSegment(
       @ApiParam(value = "Name of the table", required = true) @PathParam("tableName") String tableName,
-      @ApiParam(value = "Name of segment", required = true) @PathParam("segmentName") @Encoded String segmentName,
-      @ApiParam(value = "realtime|offline") @QueryParam("type") String tableTypeStr) {
+      @ApiParam(value = "Name of the segment", required = true) @PathParam("segmentName") @Encoded String segmentName,
+      @ApiParam(value = "OFFLINE|REALTIME") @QueryParam("type") String tableTypeStr) {
```
mcvsubbu: Why do we need table type? Segment name ought to sufficce.

Also, we should verify that there is a segment of that name in the table before we issue a reload. Otherwise we may see other type of exceptions on the server.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+  public SuccessResponse reloadAllSegments(
+      @ApiParam(value = "Name of the table", required = true) @PathParam("tableName") String tableName,
+      @ApiParam(value = "OFFLINE|REALTIME") @QueryParam("type") String tableTypeStr) {
+    List<String> tableNamesWithType =
+        getExistingTableNamesWithType(tableName, Constants.validateTableType(tableTypeStr));
+    int numReloadMessagesSent = 0;
+    for (String tableNameWithType : tableNamesWithType) {
+      numReloadMessagesSent += _pinotHelixResourceManager.reloadAllSegments(tableNameWithType);
+    }
+    return new SuccessResponse("Sent " + numReloadMessagesSent + " reload messages");
```
mcvsubbu: Will be good if we can separate the number of messages between offline and realtime in the response.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+  @DELETE
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("/segments/{tableName}/{segmentName}")
+  @ApiOperation(value = "Delete a segment", notes = "Delete a segment")
+  public SuccessResponse deleteSegment(
+      @ApiParam(value = "Name of the table", required = true) @PathParam("tableName") String tableName,
+      @ApiParam(value = "Name of the segment", required = true) @PathParam("segmentName") @Encoded String segmentName,
+      @ApiParam(value = "OFFLINE|REALTIME") @QueryParam("type") String tableTypeStr) {
+    segmentName = URIUtils.decode(segmentName);
+    List<String> tableNamesWithType =
```
mcvsubbu: again, segment name should indicate table type

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -19,255 +19,409 @@
...
+    return new SuccessResponse("Segment deleted");
+  }
+
+  @DELETE
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("/segments/{tableName}")
+  @ApiOperation(value = "Delete all segments", notes = "Delete all segments")
+  public SuccessResponse deleteAllSegments(
+      @ApiParam(value = "Name of the table", required = true) @PathParam("tableName") String tableName,
+      @ApiParam(value = "OFFLINE|REALTIME") @QueryParam("type") String tableTypeStr) {
```
mcvsubbu: required=true. Please keep API compatible.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: required is false by default (IDE will even warn about explicit required = false)

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: The string for the existing API is a JSON Object string for `List<Map<String, String>>`, which is the same as the current API (the actual return will also be a JSON Object string).

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: Make sense, will remove the type parameter for APIs with segment name included.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: Good catch

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: Done

Jackie-Jiang:
@mcvsubbu Addressed the comments. It is a little bit hard to split it into two commits because the new APIs share some helper methods with the deprecated ones, so they have to be in the same file.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -1812,7 +1813,7 @@ private boolean updateExistedSegment(String tableNameWithType, SegmentZKMetadata
    * Returns a map from server instance to list of segments it serves for the given table.
    */
   public Map<String, List<String>> getServerToSegmentsMap(String tableNameWithType) {
-    Map<String, List<String>> serverToSegmentsMap = new HashMap<>();
+    Map<String, List<String>> serverToSegmentsMap = new TreeMap<>();
```
mcvsubbu: why change this to TreeMap? This method is also used by table size reader, and when servers have tens of thousands of segments, it will add significant overhead.
If necessary, the API to get the server map can sort it (for disaplay purpose)

mcvsubbu: ah, the sort is on the servers, so ignore this

PR #4807, opened 2019-11-09
Include dataSchema in the ResultTable to retain column type information

npawar:
1. In order to retain column type information in the ResultTable, adding dataSchema. 
2. Since DataSchema already has columnNames, removing _columns from ResultTable.
3. Changing the _rows to Object[] instead of Serializable[], so that we don't have to convert non-serializable columns to string (we do that currently for BYTES)

Also did a few lines of cleanup, related to queryOptions


kishoreg:
Does this mean the response will now contain data types. do u have a sample response?

npawar:
> Does this mean the response will now contain data types. do u have a sample response?

Here's a sample
Screen Shot 2019-11-11 at 2 09 06 PM

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
```diff
@@ -220,9 +221,9 @@ public BrokerResponseNative reduceOnDataTable(BrokerRequest brokerRequest,
 
     // Parse the option from request whether to preserve the type
     Map<String, String> queryOptions = brokerRequest.getQueryOptions();
-    String preserveTypeString =
-        (queryOptions == null) ? "false" : queryOptions.getOrDefault(QueryOptionKey.PRESERVE_TYPE, "false");
-    boolean preserveType = Boolean.valueOf(preserveTypeString);
+    boolean preserveType = QueryOptionsUtils.isPreserveType(queryOptions);
```
mayankshriv: This seems like cleanup items that are not really needed for using DataSchema in ResultTable. If so, just call it out in the PR description.

pinot-core/src/main/java/org/apache/pinot/core/util/QueryOptionsUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.Map;
+
+import static org.apache.pinot.common.utils.CommonConstants.Broker.Request.QueryOptionKey;
+
+
+/**
+ * Helper methods for reading query options
+ */
+public final class QueryOptionsUtils {
```
mayankshriv: Any specific reason for having a utils class as opposed to a QueryOptions class, that is generated from Map<String, String> queryOptions?

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/BrokerReduceService.java
npawar: Done

pinot-core/src/main/java/org/apache/pinot/core/util/QueryOptionsUtils.java
npawar: Methods were moved from GroupByUtils, hence ended up putting them in another Utils.
On second thought, I took your suggestion and make it into a QueryOptions class.

npawar:

PR #4757, opened 2019-10-30
Fix controller download segment api on non-local PinotFS.

chenboat:
Fix #3847. This PR replaces #3849 which is too old for github update.

The controller's segment download rest api now uses a local file object and does not work for a non local Pinot FS. This fix uses PinotFS interface instead. For local file system, It avoids copy for local file.

@mcvsubbu @npawar

PR #4808, opened 2019-11-09
Fix the controller rest API for non local FS.

chenboat:
Second try to fix the rest API for segment download. The previous PR #4757 failed build. The root cause was that java File and URI class treats special characters like space, % differently. In this PR, we just separate them: local FS will use File and remote FS will use URI.

@npawar @jenniferdai @mcvsubbu


jenniferdai:
@chenboat can you resolve conflicts? after that, I can help you merge if you don't have permission

chenboat:
> @chenboat can you resolve conflicts? after that, I can help you merge if you don't have permission

@jenniferdai Thanks. Rebase my branch to solve the conflicts and kick off the tests again.

PR #4811, opened 2019-11-12
[TE] Fail alert task if unable to notify recipients with exception of dimensions alerter

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/alert/scheme/DetectionAlertScheme.java
```diff
@@ -80,4 +80,17 @@ public static BaseNotificationContent buildNotificationContent(Properties alertS
...
+
+  /**
+   * Fail the alert task if unable to notify owner. However, in case of dimensions recipient alerter,
+   * do not fail the alert if a subset of recipients are invalid.
+   */
+  void handleAlertFailure(int size, IllegalArgumentException e) {
+    if (this.result.getResult().size() == 1) {
+      throw e;
+    } else {
+      LOG.warn("Skipping! Found illegal arguments while sending {} anomalies for alert {}." + " Exception message: ",
```
jihaozh: shall we log which recipient failed?

akshayrai: It should be part of the exception message.

jihaozh: I see

PR #4813, opened 2019-11-12
Increase timeout for flaky test

npawar:
The test has multiple expensive operations, serialized bytes and is a MV group by. It has very frequently been timing out in out internal test infrastructure (which typically has less resources than a real cluster setup).
Increasing timeout from 10s to DEFAULT_QUERY_EXECUTOR_TIMEOUT which is 15s.

### DIFF
--- a/pinot-core/src/test/java/org/apache/pinot/queries/BaseQueriesTest.java
+++ b/pinot-core/src/test/java/org/apache/pinot/queries/BaseQueriesTest.java
@@ -26,6 +26,7 @@ import java.util.concurrent.Executors;
 import org.apache.pinot.common.request.BrokerRequest;
 import org.apache.pinot.common.response.ServerInstance;
 import org.apache.pinot.common.response.broker.BrokerResponseNative;
+import org.apache.pinot.common.utils.CommonConstants;
 import org.apache.pinot.common.utils.DataTable;
 import org.apache.pinot.core.common.Operator;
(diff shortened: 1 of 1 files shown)


pinot-core/src/test/java/org/apache/pinot/queries/BaseQueriesTest.java
```diff
@@ -97,12 +98,12 @@ protected BrokerResponseNative getBrokerResponseForQuery(String query, PlanMaker
    * @return broker response.
    */
   private BrokerResponseNative getBrokerResponseForQuery(String query, PlanMaker planMaker,
-      Map<String, String> queryOptions) {
+      Map<String, String> queryOptions, long timeout) {
     BrokerRequest brokerRequest = COMPILER.compileToBrokerRequest(query);
     brokerRequest.setQueryOptions(queryOptions);
 
     // Server side.
-    Plan plan = planMaker.makeInterSegmentPlan(getSegmentDataManagers(), brokerRequest, EXECUTOR_SERVICE, 10_000);
```
Jackie-Jiang: By default, this should be `CommonConstants.Server.DEFAULT_QUERY_EXECUTOR_TIMEOUT_MS`

npawar: done

PR #4814, opened 2019-11-12
[TE][notification] Minor cleaning up of the jira related code

akshayrai:
* Reorganized the code without any logic change
* Modified the behavior of jira lookback.
** Value of 0 should always create new anomalies and not merge.
** Negative values should always merge tickets
* Updated code for attaching snapshot to jira which will be tested in the subsequent PR.

lookbackdays - 
```
=0 - means look back for 0 days which essentially means do not look back. Keeping creating new jira tickets and do not merge.
>0 - means look back for specific number of days.
<0 - Any negative value is used to denote always merge (infinite lookback). This is the default setting.
```

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/notification/commons/ThirdEyeJiraClient.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/notification/commons/ThirdEyeJiraClient.java
@@ -61,30 +61,36 @@ public class ThirdEyeJiraClient {
         jiraAdminConfig.getJiraPassword());
   }
 
+  private String buildQueryOnCreatedBy(long lookBackMillis) {
+    long lookBackDays = TimeUnit.MILLISECONDS.toDays(lookBackMillis);
+    String createdByQuery = "created";
+    if (lookBackDays < 0) {
+      createdByQuery += "<= -0m";
+    } else {
+      createdByQuery += ">= -" + lookBackDays + "d";
+    }
+
+    return createdByQuery;
+  }
+
+  private String buildQueryOnLabels(List<String> labels) {
+    return labels.stream()
+        .map(label -> "labels = \"" + label + "\"")
+        .collect(Collectors.joining(" and "));
+  }
+
   /**
    * Search for all the existing jira tickets based on the filters
    */
   public List<Issue> getIssues(String project, List<String> labels, String reporter, long lookBackMillis) {
     List<Issue> issues = new ArrayList<>();
 
-    long lookBackDays = TimeUnit.MILLISECONDS.toDays(lookBackMillis);
-    String createdByQuery = "created";
-    if (lookBackDays <= 0) {
-      createdByQuery += "<= -0m";
-    } else {
-      createdByQuery += ">= -" + lookBackDays + "d";
-    }
-
-    String andQueryOnLabels = labels.stream()
-        .map(label -> "labels = \"" + label + "\"")
-        .collect(Collectors.joining(" and "));
-
     StringBuilder jiraQuery = new StringBuilder();
     // Query by project first as a jira optimization
     jiraQuery.append("project=").append(project);
     jiraQuery.append(" and ").append("reporter IN (").append(reporter).append(")");
-    jiraQuery.append(" and ").append(andQueryOnLabels);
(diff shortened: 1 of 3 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/notification/formatter/channels/JiraContentFormatter.java
```diff
@@ -95,15 +95,13 @@ private void validateJiraConfigs(JiraConfiguration jiraAdminConfig) {
   public JiraEntity getJiraEntity(Multimap<String, String> dimensionFilters, Collection<AnomalyResult> anomalies) {
```
vincentchenjl: Better add javadoc for public methods.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/notification/commons/ThirdEyeJiraClient.java
```diff
@@ -61,30 +61,36 @@ private JiraRestClient createJiraRestClient(JiraConfiguration jiraAdminConfig) {
...
-   * Search for all the existing jira tickets based on the filters
-   */
-  public List<Issue> getIssues(String project, List<String> labels, String reporter, long lookBackMillis) {
-    List<Issue> issues = new ArrayList<>();
-
+  private String buildQueryOnCreatedBy(long lookBackMillis) {
     long lookBackDays = TimeUnit.MILLISECONDS.toDays(lookBackMillis);
     String createdByQuery = "created";
-    if (lookBackDays <= 0) {
+    if (lookBackDays < 0) {
```
vincentchenjl: Can `lookbackDays` be negative? In what case it is negative?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/notification/formatter/channels/JiraContentFormatter.java
```diff
@@ -156,13 +135,43 @@ private JiraEntity buildJiraEntity(String jiraTemplate, Map<String, Object> temp
...
       template.process(templateValues, out);
 
-      String alertEmailHtml = new String(baos.toByteArray(), CHARSET);
+      description = new String(baos.toByteArray(), CHARSET);
+    } catch (Exception e) {
+      description = "Found an exception while constructing the description content. Pls report & reach out"
+          + " to the Thirdeye team. Exception = " + e.getMessage();
+    }
+
+    return description;
```
vincentchenjl: When an exception is thrown, should `description` be `null` or empty, as I see that you assign empty string for snapshot if there is an error.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/notification/commons/ThirdEyeJiraClient.java
akshayrai: lookBackDays is the number of days beyond which you shouldn't merger (reopen and comment) on jira tickets.
```
=0 - means look back for 0 days which essentially means do not look back. Keeping creating new jira tickets and do not merge.
>0 - means look back for specific number of days.
<0 - Any negative value is used to denote always merge (infinite lookback). This is the default setting.
```

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/notification/formatter/channels/JiraContentFormatter.java
akshayrai: When an exception is thrown in this method, the description is neither null or empty. It will store and print an error message in the jira ticket.

Reg. Snapshot- this is not enabled right now. I will test and enable this in a subsequent PR.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/notification/formatter/channels/JiraContentFormatter.java
akshayrai: added

PR #4817, opened 2019-11-13
Replace ServerInstance with ServerRoutingInstance for Broker scatter-gather

Jackie-Jiang:
ServerInstance class is subject to be deprecated, replace it with the routing instance introduced in the async broker routing
Rename Server to ServerRoutingInstance to better reflect the content of the class
Cache the short host name in ServerRoutingInstance
Make BrokerReduceService work on ServerRoutingInstance


pinot-core/src/main/java/org/apache/pinot/core/transport/ServerRoutingInstance.java
```diff
@@ -19,42 +19,48 @@
...
+ * The {@code ServerRoutingInstance} class represents the routing target instance which contains the information of
+ * hostname, port, and table type it serves.
  * <p>Different table types on same host and port are counted as different instances. Therefore, one single Pinot Server
  * might be treated as two different routing target instances based on the types of table it serves.
  */
 @ThreadSafe
-public class Server {
-  private static final String NAME_PORT_DELIMITER = "_";
+public class ServerRoutingInstance {
+  private static final String HOSTNAME_PORT_DELIMITER = "_";
```
mayankshriv: Nit: HOST_PORT_DELIMITER?

pinot-core/src/main/java/org/apache/pinot/core/transport/ServerRoutingInstance.java
```diff
@@ -19,42 +19,48 @@
...
-  private final String _hostName;
+  private final String _hostname;
   private final int _port;
   private final TableType _tableType;
 
-  private transient String _shortName;
-
   /**
-   * NOTE: server instance name is of format: {@code Server_<hostName>_<port>}, e.g. {@code Server_localhost_12345}.
+   * NOTE: server instance name is of format: {@code Server_<hostname>_<port>}, e.g. {@code Server_localhost_12345}.
```
mayankshriv: Nit: Why the change to non-camel case?

pinot-core/src/main/java/org/apache/pinot/core/transport/ServerRoutingInstance.java
Jackie-Jiang: Based on wikipedia, hostname is usually used as a single word, so will keep it this way :-P

pinot-core/src/main/java/org/apache/pinot/core/transport/ServerRoutingInstance.java
Jackie-Jiang: Same here

PR #4823, opened 2019-11-14
Enhance RoutingTable with ServerInstance

Jackie-Jiang:
Currently, RoutingTable is represented as `Map<String, List>`, a map from serverInstanceName to list of segments hosted by the server. The rest of the code derives the hostname and port by parsing serverInstanceName assuming it's of the format: `Server_<hostname>_<port>`.
This PR replaces serverInstanceName with a concrete ServerInstance Object initiated from the InstanceConfig.
NOTE: this ServerInstance is not the same as the existing ServerInstance class which tight to the connection-pool routing and subject to be cleaned up.
This will allow user to de-couple the instance name from hostname and port for the instance.

Issue link #4525
Based on #4817
Replace #4795

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/ConnectionPoolBrokerRequestHandler.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/ConnectionPoolBrokerRequestHandler.java
@@ -50,6 +50,7 @@ import org.apache.pinot.common.response.BrokerResponse;
 import org.apache.pinot.common.response.ProcessingException;
 import org.apache.pinot.common.response.ServerInstance;
 import org.apache.pinot.common.response.broker.BrokerResponseNative;
+import org.apache.pinot.common.utils.CommonConstants.Helix;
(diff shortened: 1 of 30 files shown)


pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/ConnectionPoolBrokerRequestHandler.java
```diff
@@ -232,6 +237,19 @@ protected BrokerResponse processBrokerRequest(long requestId, BrokerRequest orig
     return brokerResponse;
   }
 
+  // For backward-compatible
+  private Map<String, List<String>> convertToOldRoutingTable(
```
kishoreg: why do we need this?

Jackie-Jiang: The connection-pool routing is already deprecated, and is subject to be removed, thus here we just keep it backward-compatible.

PR #4824, opened 2019-11-14
Enhance table rebalance rest API argument check

Jackie-Jiang:
Mark tableName and type as required
Reject request if type is not provided
Extend external view stabilization max wait to 1 hour to handle large scale rebalance

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
@@ -68,8 +68,9 @@ import org.slf4j.LoggerFactory;
 public class TableRebalancer {
   private static final Logger LOGGER = LoggerFactory.getLogger(TableRebalancer.class);
 
+  // TODO: make them configurable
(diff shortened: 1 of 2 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/rebalance/TableRebalancer.java
```diff
@@ -68,8 +68,9 @@
 public class TableRebalancer {
   private static final Logger LOGGER = LoggerFactory.getLogger(TableRebalancer.class);
 
+  // TODO: make them configurable
   private static final long EXTERNAL_VIEW_CHECK_INTERVAL_MS = 1_000L; // 1 second
-  private static final long EXTERNAL_VIEW_STABILIZATION_MAX_WAIT_MS = 20 * 60_000L; // 20 minutes
+  private static final long EXTERNAL_VIEW_STABILIZATION_MAX_WAIT_MS = 60 * 60_000L; // 1 hour
```
mayankshriv: What's the reasoning behind increasing the wait time from 20 min to 1 hour? Did we find that 20 minutes might not be good enough? Would be good to document.

Jackie-Jiang: Updated the PR description

PR #4825, opened 2019-11-14
Bug Fix: Fix segment download url in SegmentZkMetadata.

mayankshriv:
It turns out that for the case when segment data directory is
deep-storage, but the segment upload method is with segment payload
being puhsed (as opposed to deep-store URI), the code had a bug where it
was generating incorrect download URL in the SegmentZkMetadata.

This PR fixes the issue with minimal change, by making the download URL
the same as the final location of the segment in deep-store. Having said
that, this part of the code seems quite complex with too many variable
controls (segment upload type, segment moving or not to final location,
etc), it would be good to clean this up at some point.

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
@@ -352,10 +352,10 @@ public class PinotSegmentUploadRestletResource {
     if (baseDataDirURI.getScheme().equalsIgnoreCase(CommonConstants.Segment.LOCAL_SEGMENT_SCHEME)) {
       return URIUtils.constructDownloadUrl(provider.getVip(), rawTableName, segmentName);
     } else {
-      // Receiving .tar.gz segment upload for pluggable storage
-      LOGGER.info("Using configured data dir {} for segment {} of table {}", _controllerConf.getDataDir(), segmentName,
-          rawTableName);
-      return URIUtils.constructDownloadUrl(baseDataDirURI.toString(), rawTableName, segmentName);
+      // Receiving .tar.gz segment upload for pluggable storage. Download URI is the same as final segment location.
+      String downloadUri = URIUtils.getPath(baseDataDirURI.toString(), rawTableName, URIUtils.encode(segmentName));
+      LOGGER.info("Using download uri: {} for segment: {} of table {}", downloadUri, segmentName, rawTableName);
+      return downloadUri;
     }
   }
 


pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -352,10 +352,11 @@ private String getZkDownloadURIForSegmentUpload(FileUploadPathProvider provider,
     if (baseDataDirURI.getScheme().equalsIgnoreCase(CommonConstants.Segment.LOCAL_SEGMENT_SCHEME)) {
       return URIUtils.constructDownloadUrl(provider.getVip(), rawTableName, segmentName);
     } else {
-      // Receiving .tar.gz segment upload for pluggable storage
-      LOGGER.info("Using configured data dir {} for segment {} of table {}", _controllerConf.getDataDir(), segmentName,
-          rawTableName);
-      return URIUtils.constructDownloadUrl(baseDataDirURI.toString(), rawTableName, segmentName);
+      // Receiving .tar.gz segment upload for pluggable storage. Download URI is the same as final segment location.
+      String downloadUri = URIUtils.getUri(baseDataDirURI.toString(), rawTableName, URIUtils.encode(segmentName)).toString();
```
Jackie-Jiang: Use URIUtils.getPath(...)

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -352,10 +352,11 @@ private String getZkDownloadURIForSegmentUpload(FileUploadPathProvider provider,
...
+      LOGGER.info("Using configured data dir {} for segment {} of table {}, Download URI: {}", _controllerConf.getDataDir(),
```
Jackie-Jiang: Please revise the log info, the data dir in log is not in sync with the behavior.
Suggest changing it to "Using download URI: {} for segment: {} of table: {}"

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
mayankshriv: :+1:

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
mayankshriv: :+1:

mcvsubbu:
> Will this break the case in which we have a special character in the segment in local FS? You may want to run at least a manual test with special character, and make sure that it works right on the download side as well.
> The correct fix may be to check if it is local file system, or introduce APIs in PinotFS to generate the right URI (so, a local FS can do the right thing of not encoding special characters with %xx)

@mayankshriv thanks for pointing out that the if statement before covers the local file system. My comment does not hold in that case.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -352,10 +352,10 @@ private String getZkDownloadURIForSegmentUpload(FileUploadPathProvider provider,
     if (baseDataDirURI.getScheme().equalsIgnoreCase(CommonConstants.Segment.LOCAL_SEGMENT_SCHEME)) {
       return URIUtils.constructDownloadUrl(provider.getVip(), rawTableName, segmentName);
     } else {
-      // Receiving .tar.gz segment upload for pluggable storage
-      LOGGER.info("Using configured data dir {} for segment {} of table {}", _controllerConf.getDataDir(), segmentName,
-          rawTableName);
-      return URIUtils.constructDownloadUrl(baseDataDirURI.toString(), rawTableName, segmentName);
+      // Receiving .tar.gz segment upload for pluggable storage. Download URI is the same as final segment location.
```
chenboat: final segment location -> deep store URI or segment store URI to make it clearer? The word final sounds a bit vague to me.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentUploadRestletResource.java
```diff
@@ -352,10 +352,10 @@ private String getZkDownloadURIForSegmentUpload(FileUploadPathProvider provider,
...
+      String downloadUri = URIUtils.getPath(baseDataDirURI.toString(), rawTableName, URIUtils.encode(segmentName));
```
chenboat: So it looks like URIUtils.constructDownloadUrl() is a bit misleading. There is a need to fix that method too? At least update its documentation?

mayankshriv: Sorry, missed your comments before merging. Will file a separate PR for that.

PR #4826, opened 2019-11-14
[TE] Enable local debugging of CSV data

akshayrai:


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/sql/SqlResponseCacheLoader.java
```diff
@@ -209,22 +213,32 @@ public SqlResponseCacheLoader(Map<String, Object> properties) throws Exception {
...
-                  String line = scanner.nextLine();
-                  String[] items = line.split(",");
-                  DateTime dateTime = DateTime.parse(items[0], fmt);
-                  if (dateTime.isAfter(maxDateTime)) {
-                    maxDateTime = dateTime;
+
+                // For demo datasets, we want to show some sample plots for recent time period.
+                // We do this by sliding the historical static data stored in [daily|hourly].csv
+                // to the current timestamp.
+                if (isDemoDataset(tableName)) {
```
jihaozh: Shall we put this logic into `MockThirdEyeDataSource` and use that instead? That one is built for demo purposes. Feels like a hack to put it in the `SqlResponseCacheLoader`

akshayrai: That's a good idea. I will see how I can move it under the MockDataSource.

PR #4828, opened 2019-11-16
Add segment batch deletion rest API

Jackie-Jiang:
In order to batch delete segments: "POST /segments/{tableName}/delete" with a JSON array payload of segments to be deleted
Table type can be derived from the segment name provided
All segments must be from the same table in order to be deleted (mix of OFFLINE and REALTIME segments are not allowed)

Manually tested, will add unit test for all rest APIs later

### DIFF
--- a/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
+++ b/pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.Encoded;
 import javax.ws.rs.GET;
@@ -415,7 +416,7 @@ public class PinotSegmentRestletResource {
     segmentName = URIUtils.decode(segmentName);
     TableType tableType = SegmentName.isRealtimeSegmentName(segmentName) ? TableType.REALTIME : TableType.OFFLINE;
     String tableNameWithType = getExistingTableNamesWithType(tableName, tableType).get(0);
-    deleteSegments(tableNameWithType, Collections.singletonList(segmentName));
+    deleteSegmentsInternal(tableNameWithType, Collections.singletonList(segmentName));
     return new SuccessResponse("Segment deleted");
   }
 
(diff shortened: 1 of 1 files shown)


pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -431,11 +432,36 @@ public SuccessResponse deleteAllSegments(
...
+    for (int i = 1; i < numSegments; i++) {
+      if (SegmentName.isRealtimeSegmentName(segments.get(i)) != realtimeSegment) {
+        throw new ControllerApplicationException(LOGGER, "All segments must be of the same type (OFFLINE|REALTIME)",
+            Status.BAD_REQUEST);
+      }
+    }
+    TableType tableType = realtimeSegment ? TableType.REALTIME : TableType.OFFLINE;
+    String tableNameWithType = getExistingTableNamesWithType(tableName, tableType).get(0);
+    deleteSegmentsInternal(tableNameWithType, segments);
+    return new SuccessResponse("Deleted segments: " + segments + " from table: " + tableNameWithType);
```
mcvsubbu: Let us limit the number of segment names we return in the rerurn status. If there is more than a certaiun number ( suggest 5) just return the numnber of segments deleted.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
```diff
@@ -431,11 +432,36 @@ public SuccessResponse deleteAllSegments(
...
+  @Path("/segments/{tableName}/delete")
+  @ApiOperation(value = "Delete the segments in the JSON array payload", notes = "Delete the segments in the JSON array payload")
+  public SuccessResponse deleteSegments(
+      @ApiParam(value = "Name of the table", required = true) @PathParam("tableName") String tableName,
+      List<String> segments) {
+    int numSegments = segments.size();
+    if (numSegments == 0) {
+      throw new ControllerApplicationException(LOGGER, "Segments must be provided", Status.BAD_REQUEST);
+    }
+    boolean realtimeSegment = SegmentName.isRealtimeSegmentName(segments.get(0));
```
haibow: nit: s/realtimeSegment/isRealtimeSegment

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: Done

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSegmentRestletResource.java
Jackie-Jiang: Done

Jackie-Jiang:
@mcvsubbu The API is manually tested, and it is calling an existing tested method `PinotHelixResourceManager.deleteSegments()`. Will add a general test for rest API wires in a separate PR.

PR #4830, opened 2019-11-18
[TE] change default time zone to pacific time zone

jihaozh:
Change default time zone to pacific time zone in anomaly detection. Previously it was using UTC as the default time zone and does not take daylight savings time into consideration.


jihaozh:
> LGTM! UTC is also hard-coded in a lot of other places like for example in the Moving Window detection, TimeSpec etc. Do you plan to move all of them?

The moving window should be fixed. There is too much logic are depending on the default TimeSpec, I'll keep it as is. But after this fix, the anomaly detection should be using pacific time as default.

PR #4832, opened 2019-11-19
[TE] fix mismatch in dataset names in data availability listener

vincentchenjl:
This PR is to fix misuse of `datasetConfigDTO.getName()` mainly and also include miscellaneous changes.

### DIFF
--- a/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityEventListener.java
+++ b/thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityEventListener.java
@@ -74,18 +74,18 @@ public class DataAvailabilityEventListener implements Runnable {
     List<DataAvailabilityEvent> events = consumer.poll(pollTimeInMilli);
     ThirdeyeMetricsUtil.triggerEventCounter.inc(events.size());
     for (DataAvailabilityEvent event : events) {
-      try {
-        if (checkAllFiltersPassed(event)) {
+      if (checkAllFiltersPassed(event)) {
+        try {
           LOG.info("Processing event: " + event.getDatasetName() + " with watermark " + event.getHighWatermark());
           String dataset = event.getDatasetName();
           datasetTriggerInfoRepo.setLastUpdateTimestamp(dataset, event.getHighWatermark());
           //Note: Batch update the timestamps of dataset if the event traffic spike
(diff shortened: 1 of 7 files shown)


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityEventListener.java
```diff
@@ -74,14 +74,18 @@ void processOneBatch() throws InterruptedException {
...
-        LOG.info("Finished processing event: " + event.getDatasetName());
+      try {
+        if (checkAllFiltersPassed(event)) {
+          LOG.info("Processing event: " + event.getDatasetName() + " with watermark " + event.getHighWatermark());
+          String dataset = event.getDatasetName();
+          datasetTriggerInfoRepo.setLastUpdateTimestamp(dataset, event.getHighWatermark());
+          //Note: Batch update the timestamps of dataset if the event traffic spike
+          datasetConfigManager.updateLastRefreshTime(dataset, event.getHighWatermark());
+          ThirdeyeMetricsUtil.processedTriggerEventCounter.inc();
+          LOG.info("Finished processing event: " + event.getDatasetName());
```
akshayrai: Is this log required? can we remove it or turn it to debug level.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/detection/trigger/DataAvailabilityEventListener.java
```diff
@@ -74,14 +74,18 @@ void processOneBatch() throws InterruptedException {
...
+        }
+      } catch (Exception e) {
+        LOG.error("Error in processing event for {}, so skipping...", event.getDatasetName(), e);
```
akshayrai: Would this generate a lot of logs? If there is an issue while filtering the events or some schema change in wimd, then this will log pretty much the same info for every WIMD event thousands of times.

- How abt moving the try catch inside the `if-condition`? log the error only if it passes the filter.
- If the filter itself fails, then the success metric will go down and we can track it on ingraph.
- You can still keep this log, but turn it to debug level so that it doesn't clutter other info logs.

vincentchenjl: Put it inside the if-condition, which should not generate too many logs.

PR #4833, opened 2019-11-19
Use ResultTable for all queries

npawar:
We introduced a new data structure ResultTable - to represent the final results in a tabular sql-like structure. In previous PRs, we used the ResultTable for sql mode group by.
This PR uses ResultTable for all queries. It is controlled by queryOption `responseFormat=sql`.

In order to construct the final DataSchema for the ResultTable, It was necessary to introduce a new API `getFinalResultType` in the `AggregationFunction`.


pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/AggregationFunction.java
```diff
@@ -107,6 +108,12 @@ void aggregateGroupByMV(int length, int[][] groupKeysArray, GroupByResultHolder
    */
   ColumnDataType getIntermediateResultColumnType();
 
+  /**
+   * Returns the {@link ColumnDataType} of the final result.
+   * <p>This column data type is used in the {@link ResultTable::_dataSchema}
```
Jackie-Jiang: (nit) remove the link to ResultTable. The javadoc should not access private fields.
Can be replaced with `This column data type is used for constructing the result table`

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctAggregationFunction.java
```diff
@@ -128,6 +128,11 @@ public ColumnDataType getIntermediateResultColumnType() {
     return ColumnDataType.OBJECT;
   }
 
+  @Override
+  public ColumnDataType getFinalResultColumnType() {
+    return ColumnDataType.OBJECT;
```
Jackie-Jiang: Throw UnsupportedOperationException here as this method should not be called

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctCountRawHLLAggregationFunction.java
```diff
@@ -105,6 +105,11 @@ public ColumnDataType getIntermediateResultColumnType() {
     return _distinctCountHLLAggregationFunction.getIntermediateResultColumnType();
   }
 
+  @Override
+  public ColumnDataType getFinalResultColumnType() {
+    return ColumnDataType.OBJECT;
```
Jackie-Jiang: I think this should be STRING

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/AggregationResultSetter.java
```diff
@@ -0,0 +1,253 @@
...
+import org.apache.pinot.core.query.aggregation.function.AggregationFunction;
+import org.apache.pinot.core.query.aggregation.function.AggregationFunctionUtils;
+import org.apache.pinot.core.transport.ServerRoutingInstance;
+
+
+/**
+ * Helper class to set Aggregation results into the BrokerResponseNative
+ * Also handles Distinct results
+ */
+public class AggregationResultSetter extends ResultSetter {
```
mayankshriv: The ResultSetter class hierarchy seems a bit confusing. The base class just seems to have a constructor (i.e. no methods to overwrite). Also, there's a special casing for Distinct inside the setter. Will the following work:
1. Create an interface ResultSetter with api `setResults`. We can also create an abstract base class if there's enough common code.
2. Create implementations of this interface for Selection and Aggregation.
3. Create a factory that returns the right result setter based on aggregation function and response format.
4. For Distinct, the factory can return the selection setter.

This way, all of the conditional logic is inside the factory, and the appropriate setter is hooked up to do its job.

How about if we have a ResultSetterFactory that returns the appropriate result setter based on response format and aggregation function. This also allows to handle the Distinct case to be modeled as selection result.

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/AggregationResultSetter.java
```diff
@@ -0,0 +1,253 @@
...
+    Iterator<Key> iterator = distinctTable.getIterator();
+
+    while (iterator.hasNext()) {
+      Key key = iterator.next();
+      Object[] columns = key.getColumns();
+      Preconditions.checkState(columns.length == columnNames.length,
+          "Error: unexpected number of columns in RecordHolder for DISTINCT");
+      resultSet.add(columns);
+    }
+    // FIXME: how do we get the correct data types?
```
mayankshriv: Let's create an issue and tag here, so it can be tracked.

npawar:
Closing this PR. Will split this into 3 PRs, so that it's easy to review:
1. Introducing ResultSetter
2. AggregationFunction change - getFinalResultType
3. ResultTable for all

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/AggregationResultSetter.java
npawar: I took this suggestion. Starting in a fresh PR with smlaller scope so that it's easy to review: https://github.com/apache/incubator-pinot/pull/4851

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/DistinctCountAggregationFunction.java
```diff
@@ -213,6 +213,11 @@ public ColumnDataType getIntermediateResultColumnType() {
     return ColumnDataType.OBJECT;
   }
 
+  @Override
+  public ColumnDataType getFinalResultColumnType() {
+    return ColumnDataType.INT;
```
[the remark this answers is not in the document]
npawar: This method is simply returning the data type based on the class definition
```
public class DistinctCountAggregationFunction implements AggregationFunction<IntOpenHashSet, Integer> {
```

PR #4834, opened 2019-11-19
Replace NettyTCPServer with QueryServer for server side query handling

Jackie-Jiang:
QueryServer is a light-weight version of NettyTCPServer which has identical functionalities
This refactor will make future server routing related changes much easier
Remove the debug API of swapping QueryScheduler on the fly as it can cause resource leak

Metrics changes (backward-incompatible)
- The old NettyServerMetrics is not registered under the correct MetricsRegistry (passed in registry is always null)
- The new metrics are registered under the server metrics:
  - Meter: REQUEST_FETCH_EXCEPTIONS, NETTY_CONNECTION_BYTES_RECEIVED, NETTY_CONNECTION_RESPONSES_SENT, NETTY_CONNECTION_BYTES_SENT
  - Timer: NETTY_CONNECTION_SEND_RESPONSE_LATENCY


pinot-core/src/main/java/org/apache/pinot/core/transport/InstanceRequestHandler.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.pinot.common.utils.BytesUtils;
+import org.apache.pinot.core.query.request.ServerQueryRequest;
+import org.apache.pinot.core.query.scheduler.QueryScheduler;
+import org.apache.thrift.TDeserializer;
+import org.apache.thrift.protocol.TCompactProtocol;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class InstanceRequestHandler extends SimpleChannelInboundHandler<ByteBuf> {
```
siddharthteotia: some class description would be helpful. If I understand it correctly, this will be a new channel handler in this directory in addition to what we already have today -- DataTableHandler. DataTableHandler takes the inbound query response (data table) at the broker from server, and this new class is for the server's inbound query request from the server. If this is indeed correct, can you please add some Javadoc accordingly?

pinot-core/src/main/java/org/apache/pinot/core/transport/InstanceRequestHandler.java
```diff
@@ -0,0 +1,122 @@
...
+    long queryArrivalTimeMs = System.currentTimeMillis();
+    _serverMetrics.addMeteredGlobalValue(ServerMeter.QUERIES, 1);
+    int requestSize = msg.readableBytes();
+    _serverMetrics.addMeteredGlobalValue(ServerMeter.NETTY_CONNECTION_BYTES_RECEIVED, requestSize);
+    byte[] requestBytes = new byte[requestSize];
+    msg.readBytes(requestBytes);
+
+    InstanceRequest instanceRequest = new InstanceRequest();
+    try {
+      _deserializer.deserialize(instanceRequest, requestBytes);
```
siddharthteotia: I think we should offload the netty thread asap. Doing the deserialization inside the event loop thread will limit the concurrency/scalability. Limited, quick work should be done inside it. May be we should decouple this and pass over the burden of building the query request to the thread on server that actually processes the request?

siddharthteotia:
A high level comment in addition to the ones I have on changes: 

I am not sure if I follow the changes here completely. The PR indicates it replaces existing NettyTCPServer with QueryServer and I don't see that happening in the code. NettyTCPServer class is untouched. Do we intend to delete it? I see we have started using the QueryServer instead of NettyServer in ServerResponse.

pinot-core/src/main/java/org/apache/pinot/core/transport/InstanceRequestHandler.java
Jackie-Jiang: Correct. Added javadoc for both DataTableHandler and InstanceRequestHandler

pinot-core/src/main/java/org/apache/pinot/core/transport/InstanceRequestHandler.java
Jackie-Jiang: This is the same behavior as the current code (check ScheduledRequestHandler), but with lower overhead (no need to construct an immediate future).
If we found this become a bottleneck in the future, we can make it async.

Jackie-Jiang:
@siddharthteotia Addressed the comments. This PR just adds the new implementation. There will be a separate PR cleaning up all the deprecated/unused classes under package pinot-transport.

siddharthteotia:
> @siddharthteotia Addressed the comments. This PR just adds the new implementation. There will be a separate PR cleaning up all the deprecated/unused classes under package pinot-transport.

Thank you. Let's merge once the build passes.

PR #4837, opened 2019-11-20
[TE] clean up some legacy code

akshayrai:
Changes:
* Deprecated and refactored the `getDimension` method in Merged Anomaly
* Clean up all the related and unused legacy classes


thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/integration/AnomalyApplicationEndToEndTest.java
```diff
@@ -1,435 +0,0 @@
-/**
```
jihaozh: Why delete this test? Looks like it's still relevant.

thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/detection/algorithm/LegacyAnomalyFunctionAlgorithmTest.java
```diff
@@ -1,98 +0,0 @@
...
-import java.util.Map;
-import java.util.concurrent.TimeUnit;
-import org.testng.Assert;
-import org.testng.annotations.BeforeMethod;
-import org.testng.annotations.Test;
-
-import static org.apache.pinot.thirdeye.dataframe.util.DataFrameUtils.*;
-
-
-public class LegacyAnomalyFunctionAlgorithmTest {
```
jihaozh: Could you also delete the `LegacyAnomalyFuncitonAlgorithm` that is associated with the test? same with other classes.

akshayrai: That class has many dependencies. I will remove it in a different PR.

thirdeye/thirdeye-pinot/src/test/java/org/apache/pinot/thirdeye/integration/AnomalyApplicationEndToEndTest.java
akshayrai: This test is still relying on the legacy classes like DetectionJobScheduler and AlertJobSchedulerV2. It needs to be rewritten.

Issue #4225, opened 2019-05-20
Make Pinot schema evolution easier


icefury71:
This has been referenced in a few issues already:
https://github.com/apache/incubator-pinot/issues/74
https://github.com/apache/incubator-pinot/issues/4029

But I'm creating a new issue to highlight the end-end problem.

Here are the current steps to perform schema evolution in Pinot:
1) Create a new schema with default values for new columns being added
2) Update schema using Controller API
3) Rolling restart the Pinot servers to "reload" the segments to reflect the default value.

In general, restarting a Pinot server is very expensive (can take anywhere between 5 to 30 mins depending on number of segments). If we need to evolve the schema frequently, this becomes a huge operational overhead.

An alternate way to resolve this issue is to backfill the old segments but this is an expensive process as well.

**A better approach:**
a) Segments which have been committed / ONLINE
We can try to lookup the new schema "on the fly" during query processing using some technique (for eg: @mayankshriv suggested using virtual columns for old segments populated with default value). That way we don't depend on any server restart.

b) Segments which are currently open / CONSUMING
We need to solve this issue: https://github.com/apache/incubator-pinot/issues/151  (how to reflect new schema in an open segment).

mayankshriv:
We could also explore dynamically adding virtual columns to provide default values, to the consuming segments as well. Is there a fundamental reason why that is not possible/hard for consuming segments, that I couldn't gather from #151?

@Jackie-Jiang What do you think about the virtual columns approach? Or if you have any other ideas?

mayankshriv:
Just to clarify, I don't mean the virtual column semantics as they are defined today, but leverage parts of that solution to dynamically add a column that provides default values without having to generate one entry per row of the same value.

Jackie-Jiang:
@icefury71 Have you tried the reload table option? You don't need to restart the server, and it allows you to reload only one single table without impacting other tables.

icefury71:
@Jackie-Jiang We haven't benchmarked that yet - will do that. The initial suspicion was reloading segments will also take a non-trivial amount of time. But yes, will update this thread with some numbers.

Jackie-Jiang:
@icefury71 The reload should be quite fast without impacting other tables. Please give it a try. I would not recommend generating the columns on the fly as all the columns in the schema are supposed to exist physically.

haibow:
@Jackie-Jiang I tried the reload option
1. update schema for a table
2. reload all segments for the table (without restarting servers)
[...]

PR #4838, opened 2019-11-20
Update schema API to reload table on schema change when applicable

haibow:
As part of efforts in #4225, add the option to reload tables automatically when a backward-compatible schema is updated.

Updated API:
`/schemas/{schemaName}`:
- add QueryParam("shouldAutoReload") with default "false" to keep the behavior same with before.
- when shouldAutoReload=true, if the new schema is backward compatible with the old one, tables using the schema will be automatically reloaded.

`/schemas`
- add QueryParam("shouldOverride") with default "true" to keep the behavior same with before.
- when shouldOverride=false, if the schema with the same name already exists, the add schema action would be rejected (to address the previous TODO "This should not update if the schema already exists")


haibow:
@Jackie-Jiang could you take a look?

pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
```diff
@@ -113,6 +114,26 @@ public static String extractRawTableName(@Nonnull String tableName) {
...
   }
 
+  /**
+   * Get the table name after adding the given type suffix.
+   *
+   * @param tableNameOptType Table name with or without type suffix
+   * @param tableType Table type suffix
+   * @return Table name with type suffix
+   */
+  public static String getTableNameWithType(String tableNameOptType, @Nullable CommonConstants.Helix.TableType tableType) {
```
Jackie-Jiang: You can use `TableNameBuilder.forType(tableType).tableNameWithType(tableName)` which IMO is more readable

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -2203,6 +2228,58 @@ public boolean isSingleTenantCluster() {
...
+   */
+  public List<String> getExistingTableNamesWithType(String tableName, @Nullable TableType tableType) {
+    List<String> tableNamesWithType = new ArrayList<>(2);
+
+    TableType tableTypeFromTableName = TableNameBuilder.getTableTypeFromTableName(tableName);
+    if (tableTypeFromTableName != null) {
+      // Table name has type suffix
+
+      if (tableType != null && tableType != tableTypeFromTableName) {
+        throw new ControllerApplicationException(LOGGER,
```
Jackie-Jiang: This exception is not supposed to be thrown by PinotHelixResourceManager. You can add a Util for the rest resources for this method

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -959,12 +961,35 @@ public PinotResourceManagerResponse deleteBrokerTenantFor(String tenantName) {
   /**
    * Schema APIs
    */
-  public void addOrUpdateSchema(Schema schema) {
+  public void addOrUpdateSchema(Schema schema, boolean shouldAutoReload, boolean shouldOverride) {
```
Jackie-Jiang: I would recommend changing this to two method:
`addSchema(Schema schema, boolean override)`
`updateSchema(Schema schema, boolean reload)`

For `updateSchema`, throw exception when schema does not exist, or the change is not backward-compatible

mcvsubbu:
You should commit this only after implementing the reload for consuming segments, right?

Jackie-Jiang:
> You should commit this only after implementing the reload for consuming segments, right?

@mcvsubbu I think the reload already works for realtime table. The only issue is the consuming segments won't get the updates immediately, but when they finish, they will get the new columns.

mcvsubbu:
> > You should commit this only after implementing the reload for consuming segments, right?
> 
> @mcvsubbu I think the reload already works for realtime table. The only issue is the consuming segments won't get the updates immediately, but when they finish, they will get the new columns.

But then there will be a schema mismatch between consuming and other segments. Isnt that what we are trying to avoid here? (basically, provide an SLA to apply the new schema, not just that it will be available once all consuming segments have completed)

Jackie-Jiang:
> > > You should commit this only after implementing the reload for consuming segments, right?
> > 
> > 
> > @mcvsubbu I think the reload already works for realtime table. The only issue is the consuming segments won't get the updates immediately, but when they finish, they will get the new columns.
> 
> But then there will be a schema mismatch between consuming and other segments. Isnt that what we are trying to avoid here? (basically, provide an SLA to apply the new schema, not just that it will be available once all consuming segments have completed)

What SLA are you referring to? Right now we don't have the feature support for applying schema change to the consuming segments, but that should not be a blocker for this automatic reload. Also, the reload is off by default, so IMO it should be okay.

pinot-common/src/main/java/org/apache/pinot/common/config/TableNameBuilder.java
haibow: Thanks for the suggestion. I saw it in TableViews.java and moved it here. I was going to use it but later found other alternatives.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
haibow: What backward-compatible behavior are you referring to? Currently the update schema endpoint (PUT /schemas/{schemaName}) would create the new schema if it did not exist

mcvsubbu:
> > > > You should commit this only after implementing the reload for consuming segments, right?
> > > 
> > > 
> > > @mcvsubbu I think the reload already works for realtime table. The only issue is the consuming segments won't get the updates immediately, but when they finish, they will get the new columns.
> > 
> > 
> > But then there will be a schema mismatch between consuming and other segments. Isnt that what we are trying to avoid here? (basically, provide an SLA to apply the new schema, not just that it will be available once all consuming segments have completed)
> 
> What SLA are you referring to? Right now we don't have the feature support for applying schema change to the consuming segments, but that should not be a blocker for this automatic reload. Also, the reload is off by default, so IMO it should be okay.

Issue https://github.com/apache/incubator-pinot/issues/4225

The idea is for us to be able to support (in a multi-tenant environment) -- schema evolution without requiring restart of all realtime consuming hosts. Further, we need to provide some specific SLA (say, 1 hour) after which the new schema can be queried by users. This is what I understand as the requirement. @haibow  has designed towards this goal. The SLA itself is arbitrary time, but at that time, a query with new column should not return schema incompat exception

haibow:
> > > > You should commit this only after implementing the reload for consuming segments, right?
> > > 
> > > 
> > > @mcvsubbu I think the reload already works for realtime table. The only issue is the consuming segments won't get the updates immediately, but when they finish, they will get the new columns.
> > 
> > 
> > But then there will be a schema mismatch between consuming and other segments. Isnt that what we are trying to avoid here? (basically, provide an SLA to apply the new schema, not just that it will be available once all consuming segments have completed)
> 
> What SLA are you referring to? Right now we don't have the feature support for applying schema change to the consuming segments, but that should not be a blocker for this automatic reload. Also, the reload is off by default, so IMO it should be okay.

Right, reload is disabled by default, and this option would help offline table for now. Let me know if you prefer to explicitly restrict reloading realtime tables before the solution is ready.

wrt SLA for the schema update, do you mean delaying updating the schema and reloading the table until the consuming segment is committed? It makes some sense from the user's point of view - when they check the table, if they see the old schema, it means the new fields are not ready for consuming yet; if they see the new schema, it means we've started consuming the new fields (as opposed to updating the schema immediately but only returning default null values for the new fields for some unknown time).

mcvsubbu:
> > > > > You should commit this only after implementing the reload for consuming segments, right?
> > > > 
> > > > 
> > > > @mcvsubbu I think the reload already works for realtime table. The only issue is the consuming segments won't get the updates immediately, but when they finish, they will get the new columns.
> > > 
> > > 
> > > But then there will be a schema mismatch between consuming and other segments. Isnt that what we are trying to avoid here? (basically, provide an SLA to apply the new schema, not just that it will be available once all consuming segments have completed)
> > 
> > 
> > What SLA are you referring to? Right now we don't have the feature support for applying schema change to the consuming segments, but that should not be a blocker for this automatic reload. Also, the reload is off by default, so IMO it should be okay.
> 
> Right, reload is disabled by default, and this option would help offline table for now. Let me know if you prefer to explicitly restrict reloading realtime tables before the solution is ready.
> 
> wrt SLA for the schema update, do you mean delaying updating the schema and reloading the table until the consuming segment is committed? It makes some sense from the user's point of view - when they check the table, if they see the old schema, it means the new fields are not ready for consuming yet; if they see the new schema, it means we've started consuming the new fields (as opposed to updating the schema immediately but only returning default null values for the new fields for some unknown time).

@haibow IIRC your requirement (when we started this project) was that your customers should be able to change the schema, and expect that the new fields show up in queries without resulting in schema errors. This "show up" has to happen within a specified  time (what I called SLA). We did not discuss the support of starting to ingest new values right away, but we did talk about dropping the consuming segment and restarting consumption with the new schema (and the downsides of that approach). So, in terms of new values appearing, at least for now, we will leave it that they will show up only when the current consuming segments are done, and new ones are started. Meanwhile, if the customer queries the new field, we don't want a schema mismatch error (which is what you are facing now).

Correct me if I am wrong.

mcvsubbu:
On another note, (and thinking more about this PR) I would rather not have two operations combined into one. If there is a failure (e.g. timeout in response from Pinot) we will not know what failed or why.

It is better to expose as minimal an API as possible, and let the caller build the functionality they need out of that. I vote for NOT doing auto-reload.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
Jackie-Jiang: Let me rephrase, when calling `updateSchema()`, if the schema does not exist, or the schema change is not backward-compatible (removing columns, changing data types etc.), throw exception

mcvsubbu:
OK, so we all agree that this commit can hold off until the reload is implemented for CONSUMING segments?

haibow:
> OK, so we all agree that this commit can hold off until the reload is implemented for CONSUMING segments?

I don't think this should be blocked by implementing reload for CONSUMING segments (I will prepare a diff after #4774 is merged). This diff just provides an option to merge two existing endpoints to one for easier operation, and doesn't change the default behavior. In fact without this auto-reload option in schema update endpoint, some users might not even know they need to reload the table for the new schema to be effective on committed segments.

pinot-spi/src/main/java/org/apache/pinot/spi/data/Schema.java
```diff
@@ -617,6 +617,28 @@ public boolean equals(Object o) {
...
+  public boolean isBackwardCompatibleWith(Schema oldSchema) {
+    if (oldSchema == null) {
+      return true;
+    }
+
+    if (!EqualityUtils.isEqual(_timeFieldSpec, oldSchema.getTimeFieldSpec()) || !EqualityUtils
+        .isEqual(_dateTimeFieldSpecs, oldSchema.getDateTimeFieldSpecs())) {
+      return false;
+    }
+    for (String columnName : oldSchema.getColumnNames()) {
```
Jackie-Jiang: Performance can be optimized by directly comparing the `_fieldSpecMap`

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -940,12 +941,55 @@ public PinotResourceManagerResponse deleteBrokerTenantFor(String tenantName) {
...
+
+  public void updateSchema(Schema schema, boolean shouldAutoReload)
+      throws TableNotFoundException {
     ZNRecord record = SchemaUtils.toZNRecord(schema);
-    String name = schema.getSchemaName();
+    String schemaName = schema.getSchemaName();
+    Schema oldSchema = ZKMetadataProvider.getSchema(_propertyStore, schemaName);
+
+    if (oldSchema == null) {
+      throw new RuntimeException("Schema did not exist.");
```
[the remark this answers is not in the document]
haibow: Created a SchemaNotFoundException

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -940,12 +941,55 @@ public PinotResourceManagerResponse deleteBrokerTenantFor(String tenantName) {
...
     PinotHelixPropertyStoreZnRecordProvider propertyStoreHelper =
         PinotHelixPropertyStoreZnRecordProvider.forSchema(_propertyStore);
-    propertyStoreHelper.set(name, record);
+    propertyStoreHelper.set(schemaName, record);
+
+    if (oldSchema != null && shouldAutoReload && schema.isBackwardCompatibleWith(oldSchema)) {
+      // Automatically reload all tables that have this schema
+      LOGGER.info("Reloading all tables with schema name: {}", schemaName);
+      for (String tableName : getAllTables()) {
+        if (schemaName.equals(getTableSchema(tableName).getSchemaName())) {
```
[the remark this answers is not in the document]
haibow: We actually have some legacy tables with schema names that are different from table names. We could do a cleanup, but it may happen for other people since it's technically allowed.

pinot-spi/src/main/java/org/apache/pinot/spi/data/Schema.java
```diff
@@ -617,6 +617,28 @@ public boolean equals(Object o) {
         .isEqualIgnoreOrder(_dateTimeFieldSpecs, that._dateTimeFieldSpecs);
   }
 
+  public boolean isBackwardCompatibleWith(Schema oldSchema) {
+    if (oldSchema == null) {
```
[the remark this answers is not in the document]
haibow: In the current code path it's not null. Still feeling as a generic util in `Schema` class, it's good to always check this.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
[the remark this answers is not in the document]
Jackie-Jiang: We log warnings for such tables for a long time. We need to figure out a way to enforce table name to be the same as schema, and enforce OFFLINE tables to have a schema. Multiple features rely on this.
We have thousands of tables in the same cluster, and this can cause quite big performance issue...

pinot-spi/src/main/java/org/apache/pinot/spi/data/Schema.java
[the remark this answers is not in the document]
Jackie-Jiang: IMO, this should throw NPE (similar to `compareTo()`). User should check whether the schema is null before calling this method. If user code has some bug and accidentally call this method with null, we should terminate the operation instead of letting them put a new schema.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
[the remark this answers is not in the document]
haibow: Agree some enforcement should be in place on the stuff you mentioned. One slight optimization is, after retrieving schemaZNRecord, we just return schema name by ZNRecord id, instead of parsing the json and get schema name from the constructed schema. Or, we just reload by schema name, and declare we do not support auto reload for those "invalid" tables..

pinot-spi/src/main/java/org/apache/pinot/spi/data/Schema.java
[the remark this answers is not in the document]
haibow: Makes sense. Updated.

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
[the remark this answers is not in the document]
Jackie-Jiang: Calling `getTableSchema()` will fetch the table config if schema cannot be found, which is not acceptable for big cluster.
We can just reload the tables with raw name the same as schema name. For the "invalid" tables, auto-reload won't have any effect.

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -117,8 +119,9 @@ public SuccessResponse deleteSchema(
   @ApiResponses(value = {@ApiResponse(code = 200, message = "Successfully updated schema"), @ApiResponse(code = 404, message = "Schema not found"), @ApiResponse(code = 400, message = "Missing or invalid request body"), @ApiResponse(code = 500, message = "Internal error")})
   public SuccessResponse updateSchema(
       @ApiParam(value = "Name of the schema", required = true) @PathParam("schemaName") String schemaName,
+      @ApiParam(value = "Whether to reload the table if the new schema is backward compatible") @DefaultValue("false") @QueryParam("shouldAutoReload") boolean shouldAutoReload,
```
Jackie-Jiang: Change to `reload`

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -117,8 +119,9 @@ public SuccessResponse deleteSchema(
...
       FormDataMultiPart multiPart) {
-    return addOrUpdateSchema(schemaName, multiPart);
+    return updateSchema(schemaName, multiPart, shouldAutoReload);
```
Jackie-Jiang: Prefer inline `getSchemaFromMultiPart(multiPart)` for readability, your choice

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -127,30 +130,34 @@ public SuccessResponse updateSchema(
...
   @Produces(MediaType.APPLICATION_JSON)
   @Path("/schemas")
   @ApiOperation(value = "Add a new schema", notes = "Adds a new schema")
   @ApiResponses(value = {@ApiResponse(code = 200, message = "Successfully created schema"), @ApiResponse(code = 404, message = "Schema not found"), @ApiResponse(code = 400, message = "Missing or invalid request body"), @ApiResponse(code = 500, message = "Internal error")})
-  public SuccessResponse addSchema(FormDataMultiPart multiPart) {
-    return addOrUpdateSchema(null, multiPart);
+  public SuccessResponse addSchema(
+      @ApiParam(value = "Whether to override the schema if the schema exists") @DefaultValue("true") @QueryParam("override") boolean override,
+      FormDataMultiPart multiPart) {
+    return addSchema(multiPart, override);
```
Jackie-Jiang: Prefer inline `getSchemaFromMultiPart(multiPart)` for readability, your choice

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -940,12 +942,56 @@ public PinotResourceManagerResponse deleteBrokerTenantFor(String tenantName) {
...
+
+    boolean isNewSchemaBackwardCompatible = schema.isBackwardCompatibleWith(oldSchema);
+    if (!isNewSchemaBackwardCompatible) {
+      LOGGER.warn(String.format("New schema %s is not backward compatible", schemaName));
+    }
+
+    if (reload && isNewSchemaBackwardCompatible) {
+      // Automatically reload all tables that have this schema
+      LOGGER.info("Reloading all tables with schema name: {}", schemaName);
+      for (String tableNameWithType : getExistingTableNamesWithType(schemaName, null)) {
```
Jackie-Jiang: Catch table not found exception here (for "invalid" tables) and log a warning

pinot-controller/src/main/java/org/apache/pinot/controller/helix/core/PinotHelixResourceManager.java
```diff
@@ -940,12 +942,56 @@ public PinotResourceManagerResponse deleteBrokerTenantFor(String tenantName) {
...
         PinotHelixPropertyStoreZnRecordProvider.forSchema(_propertyStore);
-    propertyStoreHelper.set(name, record);
+    propertyStoreHelper.set(schemaName, record);
+
+    boolean isNewSchemaBackwardCompatible = schema.isBackwardCompatibleWith(oldSchema);
+    if (!isNewSchemaBackwardCompatible) {
+      LOGGER.warn(String.format("New schema %s is not backward compatible", schemaName));
+    }
+
+    if (reload && isNewSchemaBackwardCompatible) {
```
Jackie-Jiang: (nit) Put this into `else`?

pinot-controller/src/main/java/org/apache/pinot/controller/api/resources/PinotSchemaRestletResource.java
```diff
@@ -215,22 +240,19 @@ private SuccessResponse addOrUpdateSchema(@Nullable String schemaName, Schema sc
...
-
+      _pinotHelixResourceManager.updateSchema(schema, reload);
       // Best effort notification. If controller fails at this point, no notification is given.
-      LOGGER.info("Metadata change notification for schema {}, type {}", schema, eventType);
-      _metadataEventNotifierFactory.create().notifyOnSchemaEvents(schema, eventType);
-
+      LOGGER.info("Notifying metadata event for updating schema: {}", schemaName);
+      _metadataEventNotifierFactory.create().notifyOnSchemaEvents(schema, SchemaEventType.UPDATE);
       return new SuccessResponse(schema.getSchemaName() + " successfully added");
+    } catch (SchemaNotFoundException | TableNotFoundException e) {
```
Jackie-Jiang: Can we handle these 2 exceptions separately and return different error message?
For SchemaNotFoundException: "Failed to find schema %s"
For TableNotFoundException: "Failed to find table %s to reload"

PR #4839, opened 2019-11-20
Fix documentation on LLC configs

mcvsubbu:
Fixed other warnings that had showed up over time.

### DIFF
--- a/docs/pluggable_streams.rst
+++ b/docs/pluggable_streams.rst
@@ -121,7 +121,7 @@ All values should be strings. For example:
 
     "streamType" : "foo",
     "stream.foo.topic.name" : "SomeTopic",
-    "stream.foo.consumer.type": "lowlevel",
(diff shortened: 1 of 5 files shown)


docs/pluggable_streams.rst
```diff
@@ -205,7 +211,7 @@ Below is a sample table config used to create a realtime table with Kafka Partit
       "loadMode": "MMAP",
       "streamConfigs": {
         "streamType": "kafka",
-        "stream.kafka.consumer.type": "simple",
+        "stream.kafka.consumer.type": "LowLevel",
```
haibow: Change line 124 from `lowlevel` to `LowLevel` as well? 

It doesn't hurt as is since ultimately they are all converted to uppercase, but would be nice to make the configs more consistent and avoid confusion.

mcvsubbu: Agreed, changed.

PR #4841, opened 2019-11-20
Remove the support of Hocon based config

Jackie-Jiang:
The Hocon based config was introduced in #2746
We decide to stick with Json based config and remove Hocon based config for the following reasons:
- We don't have full support on nested config for Hocon
- No one is maintaining the code for Hocon based config
- No one is using the Hocon based config


kishoreg:
Bittersweet but the right thing to do.

PR #4842, opened 2019-11-20
[TE] add detection health for the alerts list

jihaozh:
- Add detection health in the alert list endpoint.
- Store the detection of health in detection config DTO and periodically update it in the monitoring tasks. So that when a user opens the page, it will avoid the time to generate the detection health.
- Minor change to the frontend to move away from the legacy endpoints.


thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/cache/metadata/MetaDataCollectionCache.java
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Create a Metadata collection cache
+   * @param cacheLoader the cache loader for the meta data collection
+   * @param reloadPeriod the period of the reloading frequency
+   * @param reloadUnit the unit for the reload period
+   */
+  public MetaDataCollectionCache(MetadataCacheLoader<T> cacheLoader, int reloadPeriod, TimeUnit reloadUnit) {
+    this.loadingCache = CacheBuilder.newBuilder()
+        .expireAfterWrite(10, TimeUnit.MINUTES)
```
xiaohui-sun: This looks dangerous.
What if user disabled/enabled the config, and then refresh the page?
Will the user see the updated config?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/monitor/MonitorTaskRunner.java
```diff
@@ -113,11 +117,35 @@ private void executeMonitorUpdate(MonitorTaskInfo monitorTaskInfo) {
...
+      // update detection health status
+      try {
+        DateTime healthStatusWindowEnd = DateTime.now();
+        config.setHealth(new DetectionHealth.Builder(config.getId(), healthStatusWindowEnd.minusDays(30).getMillis(),
+            healthStatusWindowEnd.getMillis()).addRegressionStatus(DAO_REGISTRY.getEvaluationManager())
+            .addAnomalyCoverageStatus(DAO_REGISTRY.getMergedAnomalyResultDAO())
+            .addDetectionTaskStatus(DAO_REGISTRY.getTaskDAO())
+            .addOverallHealth()
+            .build());
+        detectionDAO.update(config);
```
xiaohui-sun: How long does the previous build cost?
It may have racing-condition with other update modules.
Suggest to get the latest config again before actually updating it.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/formatter/DetectionConfigFormatter.java
```diff
@@ -154,6 +170,20 @@ public DetectionConfigFormatter(MetricConfigManager metricDAO, DatasetConfigMana
     return metricUrns;
   }
 
+  private DetectionHealth getDetectionHealth(DetectionConfigDTO config) {
+    // Return the stored detection health when it is available in the DetectionConfig.
+    if (!Objects.isNull(config.getHealth())) {
+      return config.getHealth();
+    }
+    DateTime end = DateTime.now();
```
xiaohui-sun: No need to calculate here.
We can wait for the background task to update the status.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/health/DetectionHealth.java
```diff
@@ -142,6 +142,17 @@ public Builder addDetectionTaskStatus(TaskManager taskDAO, long limit) {
...
     }
 
+    /**
+     * Add the detection task health status in the health report built by the builder. Do not return any task details
+     * @param taskDAO the task dao
+     * @return the builder
+     */
+    public Builder addDetectionTaskStatus(TaskManager taskDAO) {
+      this.taskDAO = taskDAO;
+      this.taskLimit = 0;
```
xiaohui-sun: Why init taskLimit here?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/cache/metadata/MetaDataCollectionCache.java
jihaozh: Good point. Yes, it should. It will trigger a cache refresh whenever the user changes the config.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/anomaly/monitor/MonitorTaskRunner.java
jihaozh: Good point. I've updated it to read again before writing the config.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/formatter/DetectionConfigFormatter.java
jihaozh: Sure. I've changed it to return an Unknown status if it's not generated yet.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/health/DetectionHealth.java
jihaozh: This is the number of taskDTO returned in the health report. Primarily used for debugging purposes. By default set it to 0 means don't return any task details.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/cache/metadata/MetaDataCollectionCache.java
xiaohui-sun: Config could be changed in many places.
Are you sure the cache is always updated?

jihaozh: It should be. It will trigger a cache refresh whenever 1. Detection or Subscription group YAML change. 2. Toggle the active button.  Also, there is a thread that refreshes the cache every 3 minutes.  The cache is kept very fresh. 

The goal of the cache is to pre-load the list into memory so that users don't need to wait for the results from the database when opening the page. In my testing, this reduces the page latency from around 3s to around 0.3s.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/cache/metadata/MetaDataCollectionCache.java
```diff
@@ -0,0 +1,91 @@
...
+    scheduledCacheRefreshExecutor.scheduleAtFixedRate(this::refresh, 0, reloadPeriod, reloadUnit);
+    adHocRefreshExecutor = Executors.newSingleThreadExecutor();
+  }
+
+  /**
+   * Create a Metadata collection cache. By default reloads every 3 minutes
+   * @param cacheLoader the cache loader for the meta data collection
+   */
+  public MetaDataCollectionCache(MetadataCacheLoader<T> cacheLoader) {
+    this(cacheLoader, 3, TimeUnit.MINUTES);
```
akshayrai: Would this offer more load on mysql? In our current model, we query the db only when users load the alerts page which is very rare. Refreshing every 3 minutes would result in a lot of unnecessary queries to mysql right.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/cache/metadata/MetaDataCollectionCache.java
akshayrai: Although this looks great, we need to keep in mind that whenever we update the metadata in the future, we need to explicitly make a call to cache refresh which is not very trivial and if we miss this, then it can be a pain while debugging.

Can we make the cache refresh call part of the data access layer?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/cache/metadata/MetaDataCollectionCache.java
jihaozh: Yes, it will query the DB more times. For our current usage pattern, this is called around 100 times a day, but mostly during the work hours. Keeping the cache fresh will require more queries, but will be more spread out to the whole day. The QPS doesn't vary much. As the number of the page view grows, it will benefit us more.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/cache/metadata/MetaDataCollectionCache.java
jihaozh: Actually, this cache is only used in the UI endpoint to show the list of alerts. Most of the time, it's not necessary to update the cache whenever we update the metadata because that page doesn't show that information and the consistency requirement is not very high. The most important places to refresh it are creating alerts, activate/deactivate the alerts, create/update subscription groups, where users expect to see the changes in UI. Otherwise, I think an eventual consistency model would be enough, it will be refreshed in 3 minutes

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/datasource/cache/metadata/MetaDataCollectionCache.java
xiaohui-sun: Keep cache consistent is a potential issue as @akshayrai pointed out. Actually mysql has cache itself and I don't think this is a big issue for now.

jihaozh: okay, I'll remove it for now

PR #4846, opened 2019-11-22
[TE] Refactor the fetch anomalies interface to incorporate detection id within the anomaly slice

akshayrai:


bxji:
Can we add a comment somewhere noting that configId and detectionId are referring to the same thing? We are using them interchangeably in the code, but it would not be clear to someone new to the codebase that they are the same.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/AnomalySlice.java
```diff
@@ -62,28 +83,40 @@ public long getEnd() {
...
 
+  public boolean getIsTaggedAsChild() {
+    return isTaggedAsChild;
+  }
+
   public Multimap<String, String> getFilters() {
     return filters;
   }
 
+  public List<String> getDetectionCompNames() {
```
bxji: minor nitpick, but is there a reason we are using DetectionCompNames rather than the full DetectionComponentNames?

akshayrai:
> Can we add a comment somewhere noting that configId and detectionId are referring to the same thing? We are using them interchangeably in the code, but it would not be clear to someone new to the codebase that they are the same.

@bxji, good point. I have made sure any new variable instantiation as part of this PR already follows this convention. Updating existing references will require touching other pieces of the class and could be taken up in a separate PR.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/spi/model/AnomalySlice.java
akshayrai: it shouldn't really matter as long as it is not conflicting or unclear.

PR #4847, opened 2019-11-22
Rewrite single predicate IN as EQUALS

siddharthteotia:
Rewrite single value IN predicate as EQUALITY. Also add support for doing the predicate rewriting
for nested tree in MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer
    
Added plenty of more tests covering:
    
(1) tests for cases where query need not be rewritten (the existing code didn't handle this well)
    
(2) nested query rewrite
    
(3) tests for both AND, OR rooted trees

### DIFF
--- a/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
+++ b/pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
@@ -20,148 +20,312 @@ package org.apache.pinot.broker.requesthandler;
 
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.Iterator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeSet;
 import org.apache.pinot.common.request.FilterOperator;
 import org.apache.pinot.common.utils.request.FilterQueryTree;
 
 
 /**
- * Optimizer that collapses multiple OR clauses to IN clauses. For example, <code>a = 1 OR a = 2 OR a =
- * 3</code> gets turned to <code>a IN (1, 2, 3)</code>.
+ * The optimizer looks at the FilterQueryTree to convert
+ *
+ * (1) multiple OR equality predicates into a single IN predicate
+ * e.g COL = 1 OR COl = 2 OR COL = 3 is rewritten as COL IN (1, 2, 3)
+ *
+ * (2) single value IN predicate into equality predicate
+ * e.g COL IN (1) is rewritten as COL = 1
+ *
+ * (3) Deduplicates the predicate values for IN.
+ *
+ * These rewrites are also done for child subtrees (handling nested nature)
+ *
+ * NOTE: This optimizer is dependent on {@link FlattenNestedPredicatesFilterQueryTreeOptimizer}
+ * first flattening the tree and removing redundant operators.
  */
 public class MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer extends FilterQueryTreeOptimizer {
+
   @Override
   public FilterQueryTree optimize(FilterQueryOptimizerRequest request) {
-    return optimize(request.getFilterQueryTree(), null);
+    return optimize(request.getFilterQueryTree());
   }
 
-  private FilterQueryTree optimize(FilterQueryTree filterQueryTree, FilterQueryTree parent) {
+  /**
+   * Optimize a FilterQueryTree. When we look at the root operator,
+   * there are 3 possible cases:
+   *
+   * (1) root is OR
+   * (2) root is AND
+   * (3) root itself is a leaf level operator: EQUALITY, RANGE, IN etc
+   *
+   * For (1) and (2), we look at the subtrees (children). For (3), currently
+   * the only case where rewrite is needed is for single predicate IN
+   *
+   * @param filterQueryTree root of tree to optimize/rewrite
+   * @return root of optimized tree
+   */
+  private FilterQueryTree optimize(FilterQueryTree filterQueryTree) {
     if (filterQueryTree.getOperator() == FilterOperator.OR) {
-      Map<String, Set<String>> columnToValues = new HashMap<>();
-      List<FilterQueryTree> nonEqualityOperators = new ArrayList<>();
-
-      // Collect all equality/in values and non-equality operators
-      boolean containsDuplicates = collectChildOperators(filterQueryTree, columnToValues, nonEqualityOperators);
-
-      // If we have at least one column to return
-      if (!columnToValues.isEmpty()) {
-        // We can eliminate the OR node if there is only one column with multiple values
-        if (columnToValues.size() == 1 && nonEqualityOperators.isEmpty()) {
-          Map.Entry<String, Set<String>> columnAndValues = columnToValues.entrySet().iterator().next();
-
-          return buildFilterQueryTreeForColumnAndValues(columnAndValues);
-        }
-
-        // Check if we need to rebuild the predicate
-        boolean rebuildRequired = isRebuildRequired(columnToValues, containsDuplicates);
-
-        if (!rebuildRequired) {
-          // No mutation needed, so just return the same tree
-          return filterQueryTree;
+      // CASE 1: Root operator is OR
+      return optimizeTreeRootedAtOR(filterQueryTree);
+    } else if (filterQueryTree.getOperator() == FilterOperator.AND) {
+      // CASE 2: Root operator is AND
+      return optimizeTreeRootedAtAND(filterQueryTree);
+    } else {
+      // CASE 3:
+      // if we are here, it means the query has a single predicate
+      // (just one column with one or more predicate values)
+      // a = 1
+      // a IN (1)
+      // a > 7
+      // a IN (1, 2, 3, 4)
+      // .... anything single predicate
+      if (filterQueryTree.getOperator() == FilterOperator.IN) {
+        // for single predicate rewriting, we currently handle the following scenario:
+        // for IN , dedup the values and rewrite it as IN/EQ based on
+        // whether the number of predicate values is more than 1 or 1
+        List<String> values = filterQueryTree.getValue();
+        if (values.size() > 1) {
+          // if more than 1 value, dedup them
+          // rewrite as IN if more than 1 value after deduping
+          // rewrite as EQ if 1 value after deduping
+          Set<String> deduped = new HashSet<>(values);
+          values = new ArrayList<>(deduped);
+          if (deduped.size() > 1) {
+            // e.g a IN (1, 2, 3, 4, 2, 3) -> a IN (1, 2, 3, 4)
+            return new FilterQueryTree(filterQueryTree.getColumn(), values, FilterOperator.IN, null);
+          }
+          // e.g a IN (1, 1, 1) -> a = 1
+          return new FilterQueryTree(filterQueryTree.getColumn(), values, FilterOperator.EQUALITY, null);
         } else {
-          // Rebuild the predicates
-          return rebuildFilterPredicate(columnToValues, nonEqualityOperators);
+          // e.g a IN (1) -> a = 1
+          return new FilterQueryTree(filterQueryTree.getColumn(), values, FilterOperator.EQUALITY, null);
         }
       }
-    } else if (filterQueryTree.getChildren() != null) {
-      // Optimize the child nodes, if any
-      applyOptimizationToChildNodes(filterQueryTree);
+      return filterQueryTree;
+    }
+  }
+
+  /**
+   * Optimize FilterQueryTree rooted at OR. We look at the immediate children
+   * of OR from two different scenarios:
+   *
+   * (1) The immediate child could be a IN or an EQUALITY operator
+   * (2) The immediate child could be other operators -- AND, NOT, RANGE, BETWEEN etc
+   *
+   * We first go over the immediate children and collect the operators
+   * in two separate data structures.
+   *
+   * (1) is maintained in a map of column name to unique set of 1 or more
+   * predicate values. For this we also track in a local variable whether
+   * the predicates in the map need to be rewritten or not. This decision
+   * is made by {@link #collectChildOperatorsOfRootOROperator(FilterQueryTree,
+   * Map, List)} method as it builds the data structure
+   *
+   * (2) is maintained in a simple list of FilterQueryTrees representing the
+   * non EQUALITY/IN operators. Also, as {@link #collectChildOperatorsOfRootOROperator
+   * (FilterQueryTree, Map, List)} builds this list, it also optimizes these set of
+   * FilterQueryTrees.
+   *
+   * We first deal with a very special case where the entire filter just has one column
+   * and there aren't any non EQUALITY/IN operators. Few examples to represent this
+   * scenario:
+   *
+   * A = 100 OR A = 200 OR A = 300
+   * A IN (100) OR A IN (200) OR A IN (300)
+   * A = 100 OR A = 200 OR A = 300 OR A IN (200, 300, 400, 500, 600)
+   *
+   * We simply remove OR and return a FilterQueryTree with leaf operator
+   * -- if the number of values is 1, use EQUALITY
+   * -- else use IN
+   *
+   * If this scenario is not applicable, we only need to look if EQUALITY/IN set
+   * of child operators need to be rewritten. This is already tracked in the
+   * boolean variable so we just need to check that.
+   *
+   * If the  variable is false, then we are done and return.
+   * Else, we go over each predicate in map, rewrite it and in the end append the
+   * list of non EQUALITY/IN operators (note hat these were already optimized as
+   * part of collect method). Finally, we return the root of this optimized tree
+   *
+   * Examples:
+   *
+   * a IN (1) OR b IN (2) -- two IN operators and zero other operators
+   * both IN operators can be rewritten to EQUALITY
+   *
+   * a IN (100) OR b = 300 -- one IN and one EQUALITY operator,
+   * zero other operators, IN can be rewritten
+   *
+   * a = 100 OR b = 300 -- two equality and zero other operators
+   * nothing needs to be rewritten
+   *
+   * a = 1 OR b = 2 OR c > 20 -- two equality and one other operator
+   * nothing needs to be rewritten
+
+   * a IN (1) OR b IN (2) OR c > 20 -- two IN operators and one other operators
+   * both IN operators can be rewritten to EQUALITY and other operator need not
+   * be rewritten
+   *
+   * a IN (1) OR b IN (2) OR (c > 20 AND a > 200)
+   * two IN operators and AND other operator. both IN can be rewritten
+   * and for AND we can't decide unless we drill down so we recurse.
+   *
+   * a = 1 OR b = 2 OR (c > 20 AND a > 200)
+   * two EQUALITY operators and AND other operator.
+   * both EQUALITY need not be rewritten
+   * and for AND we can't decide unless we drill down so we recurse
+   *
+   * a = 1 OR b = 2 OR (c IN (20) AND a > 200)
+   * two EQUALITY operators and AND other operator.
+   * both EQUALITY need not be rewritten
+   * and for AND we can't decide unless we drill down so we recurse --
+   * the recursion will return the AND tree with the inner IN rewritten
+   *
+   * @param filterQueryTree root of tree to optimize
+   * @return root of optimized tree
+   */
+  private FilterQueryTree optimizeTreeRootedAtOR(FilterQueryTree filterQueryTree) {
+    // TODO: evaluate if we can use TreeMap and TreeSet here
+    Map<String, Set<String>> columnToInEqPredicateValues = new HashMap<>();
+    List<FilterQueryTree> nonEqInOperators = new ArrayList<>();
+
+    // Collect all equality/in operators and non-equality operators
+    // these are the immediate children of root OR operator
+    boolean rewriteINEQChildrenOfORRootOperator = collectChildOperatorsOfRootOROperator(filterQueryTree, columnToInEqPredicateValues, nonEqInOperators);
+
+    if (columnToInEqPredicateValues.size() == 1 && nonEqInOperators.isEmpty()) {
+      // We can eliminate the OR root node if there is exactly one unique column with one or more
+      // predicate values(could be duplicate) using IN/OR. Rewritten as follows:
+      // -- if there is one predicate (after deduping), use EQUALITY
+      // -- else use IN
+      // e.g:
+      // a = 1 OR a = 1 --> can be rewritten as a = 1
+      // a = 1 OR a = 2 OR a = 3 --> can be rewritten as a IN (1, 2, 3)
+      Map.Entry<String, Set<String>> columnAndValues = columnToInEqPredicateValues.entrySet().iterator().next();
+      return buildFilterQueryTreeForColumnAndPredicateInfo(columnAndValues.getKey(), columnAndValues.getValue());
     }
 
+    if (!rewriteINEQChildrenOfORRootOperator) {
+      // No predicate rewriting, so just return the same tree
+      return filterQueryTree;
+    }
+
+    // Rewrite the IN/EQUALITY child operators
+    FilterQueryTree optimizedRoot = rebuildFilterPredicate(columnToInEqPredicateValues);
+    // append all non EQ/IN child operators
+    optimizedRoot.getChildren().addAll(nonEqInOperators);
+
+    return optimizedRoot;
+  }
+
+  /**
+   * Optimize FilterQueryTree rooted at AND. We recurse for each
+   * subtree by calling optimize
+   * The root (AND) is kept same and optimize is invoked on
+   * each subtree and updated in the children list of root
+   * @param filterQueryTree root of tree to optimize
+   * @return root of optimized tree
+   *
+   * Examples:
+   *
+   * a IN (1) AND b = 3 -- rewritten to a = 1 AND b = 3
+   * a IN (1) AND b IN (3) -- rewritten to a = 1 AND b = 3
+   * (a = 1 OR a = 2) AND b = 3 -- rewritten to (a IN (1, 2)) AND b = 3
+   * a > 20 AND a < 100 -- not rewritten
+   * a = 100 AND b = 300 -- not rewritten
+   * (a IN (1, 2, 3) OR b IN (4, 5, 6)) AND (a < 20 OR b > 2) -- not rewritten
+   */
+  private FilterQueryTree optimizeTreeRootedAtAND(FilterQueryTree filterQueryTree) {
+    List<FilterQueryTree> children = filterQueryTree.getChildren();
+    int numChildren = children.size();
+    for (int i = 0; i < numChildren; i++) {
+      FilterQueryTree childQueryTree = children.get(i);
+      children.set(i, optimize(childQueryTree));
+    }
     return filterQueryTree;
   }
 
-  private void applyOptimizationToChildNodes(FilterQueryTree filterQueryTree) {
-    Iterator<FilterQueryTree> childTreeIterator = filterQueryTree.getChildren().iterator();
-    List<FilterQueryTree> childrenToAdd = null;
-
-    while (childTreeIterator.hasNext()) {
-      FilterQueryTree childQueryTree = childTreeIterator.next();
-      FilterQueryTree optimizedChildQueryTree = optimize(childQueryTree, filterQueryTree);
-      if (childQueryTree != optimizedChildQueryTree) {
-        childTreeIterator.remove();
-        if (childrenToAdd == null) {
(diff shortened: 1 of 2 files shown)


siddharthteotia:
cc @Jackie-Jiang

siddharthteotia:
> Can we do this after the compilation instead of at compile time? You should do this in the BrokerRequestOptimizer

@Jackie-Jiang , Initially my thought was similar to do this as part of separate optimization pass after query compilation (in BrokerRequestOptimizer). The reason to do this during compilation is to avoid the performance overhead of traversing down the filter expression (which could be large) and finding that particular IN expression(s) that have a single predicate and rewriting it. 

Having said that, for each query we already spend time doing optimization (range merge etc) but in case other optimizations (range-merge, flattening) are a NO-OP then traversing down the filter query tree only for rewriting IN could potentially be an overhead that can be avoided by doing the rewrite at compile time. 

What do you think?

kishoreg:
Can we do what Jackie suggested but avoid the overhead?Are we not already traversing the filter tree to add time filter and do other optimizations

siddharthteotia:
> Can we do what Jackie suggested but avoid the overhead? Are we not already traversing the filter tree to add time filter and do other optimizations

Let me take a stab at doing this in an optimization pass.

siddharthteotia:
While working on this I realized that there is room for optimization in FilterQueryTreeOptimizer. I have addressed that in PR https://github.com/apache/incubator-pinot/pull/4901
It also subsumes the work planned for this one.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -18,163 +18,221 @@
...
+      return buildFilterQueryTreeForColumnAndPredicateInfo(columnAndValues.getKey(), columnAndValues.getValue());
+    }
 
-        if (!rebuildRequired) {
-          // No mutation needed, so just return the same tree
-          return filterQueryTree;
-        } else {
-          // Rebuild the predicates
-          return rebuildFilterPredicate(columnToValues, nonEqualityOperators);
+    boolean rewriteRequiredForInEqOperators = containsDuplicates || isRebuildRequiredForInEqPredicates(columnToInEqPredicateInfo);
```
siddharthteotia: The logic is more or less similar to old but some additional work was needed to handle nested better. Secondly, the new code handles the no-rewrite cases better.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -18,163 +18,221 @@
...
   private void applyOptimizationToChildNodes(FilterQueryTree filterQueryTree) {
-    Iterator<FilterQueryTree> childTreeIterator = filterQueryTree.getChildren().iterator();
-    List<FilterQueryTree> childrenToAdd = null;
-
-    while (childTreeIterator.hasNext()) {
-      FilterQueryTree childQueryTree = childTreeIterator.next();
-      FilterQueryTree optimizedChildQueryTree = optimize(childQueryTree, filterQueryTree);
+    List<FilterQueryTree> children = filterQueryTree.getChildren();
+    for (int i = 0; i < children.size(); i++) {
+      FilterQueryTree childQueryTree = children.get(i);
```
siddharthteotia: This logic has been simplified. No need to use iterator and remove()

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -18,163 +18,221 @@
...
-            containsDuplicates = true;
-          }
+      FilterOperator operator = childQueryTree.getOperator();
+      if (operator == FilterOperator.EQUALITY || operator == FilterOperator.IN) {
+        List<String> childValues = childQueryTree.getValue();
+        String column = childQueryTree.getColumn();
+        PredicateInfo predicateInfo = columnToInEqPredicateInfo.get(column);
+        if (predicateInfo == null) {
+          predicateInfo = new PredicateInfo(new HashSet<>(), false, operator == FilterOperator.IN);
+          columnToInEqPredicateInfo.put(column, predicateInfo);
```
siddharthteotia: This logic has been simplified and now also collects the position of each non IN/EQUALITY child operator

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -18,163 +18,221 @@
...
-        if (columnToValues.size() == 1 && nonEqualityOperators.isEmpty()) {
-          Map.Entry<String, Set<String>> columnAndValues = columnToValues.entrySet().iterator().next();
+  private static class PredicateInfo {
+    Set<String> _predicateValues;
+    boolean _multipleOccurences;
+    boolean _inPredicate;
 
-          return buildFilterQueryTreeForColumnAndValues(columnAndValues);
-        }
+    PredicateInfo(Set<String> predicateValues, boolean multipleOccurences, boolean inPredicate) {
```
Jackie-Jiang: Why do we need these extra info besides predicateValues?

siddharthteotia:
@Jackie-Jiang , made changes based on our offline discussion

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: In addition to predicateValues, we need a boolean rewritePredicate to indicate if that predicate needs to be rewritten

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -18,163 +18,221 @@
...
+      _multipleOccurences = multipleOccurences;
+      _inPredicate = inPredicate;
+    }
+  }
 
-        // Check if we need to rebuild the predicate
-        boolean rebuildRequired = isRebuildRequired(columnToValues, containsDuplicates);
+  private FilterQueryTree optimizeTreeRootedAtOR(FilterQueryTree filterQueryTree) {
+    Map<String, PredicateInfo> columnToInEqPredicateInfo = new HashMap<>();
+    Map<FilterQueryTree, Integer> nonEqInOperators = new HashMap<>();
```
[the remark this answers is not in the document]
siddharthteotia: Based on the latest changes, this is no longer needed. Reverted back to list. I was initially using the map to collect their respective positions of the non EQUALITY / IN operators in the children list of root OR operator.

Now, the collect() function returns an array of positions and the optimizeTreeRootedAtOR() method explains how the returned array is used.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: All the information is now self-contained in this data structure

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
+   * two EQUALITY operators and AND other operator.
+   * both EQUALITY need not be rewritten
+   * and for AND we can't decide unless we drill down so we recurse --
+   * the recursion will return the AND tree with the inner IN rewritten
+   *
+   * @param filterQueryTree root of tree to optimize
+   * @return root of optimized tree
+   */
+  private FilterQueryTree optimizeTreeRootedAtOR(FilterQueryTree filterQueryTree) {
+    Map<String, Set<String>> columnToInEqPredicateValues = new HashMap<>();
```
Jackie-Jiang: TreeMap might be better as numColumns should be very limited?
Also order will be deterministic

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
+   * a IN (1) AND b = 3 -- rewritten to a = 1 AND b = 3
+   * a IN (1) AND b IN (3) -- rewritten to a = 1 AND b = 3
+   * (a = 1 OR a = 2) AND b = 3 -- rewritten to (a IN (1, 2)) AND b = 3
+   * a > 20 AND a < 100 -- not rewritten
+   * a = 100 AND b = 300 -- not rewritten
+   * (a IN (1, 2, 3) OR b IN (4, 5, 6)) AND (a < 20 OR b > 2) -- not rewritten
+   */
+  private FilterQueryTree optimizeTreeRootedAtAND(FilterQueryTree filterQueryTree) {
+    List<FilterQueryTree> children = filterQueryTree.getChildren();
+    for (int i = 0; i < children.size(); i++) {
```
Jackie-Jiang: cache the size of children

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
+      FilterQueryTree childQueryTree = children.get(i);
+      FilterQueryTree optimizedChildQueryTree = optimize(childQueryTree);
       if (childQueryTree != optimizedChildQueryTree) {
-        childTreeIterator.remove();
-        if (childrenToAdd == null) {
-          childrenToAdd = new ArrayList<>();
-        }
-        childrenToAdd.add(optimizedChildQueryTree);
+        children.set(i, optimizedChildQueryTree);
```
Jackie-Jiang: Directly call `children.set(i, optimize(children.get(i))` to avoid this branching

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
-      for (Set<String> columnValues : columnToValues.values()) {
-        if (1 < columnValues.size()) {
-          rebuildRequired = true;
-          break;
+        if (!rewriteINEQChildrenOfORRootOperator && operator == FilterOperator.IN && predicateValues.size() == 1) {
+          // single predicate value IN should be rewritten
+          rewriteINEQChildrenOfORRootOperator = true;
+        }
+      } else {
+        // the immediate child of root operator is <, >, AND, <=, >=, BETWEEN ...
```
Jackie-Jiang: AND, RANGE ...

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
+        if (childQueryTree.getOperator() == FilterOperator.AND) {
+          nonEqInOperators.add(optimize(childQueryTree));
```
Jackie-Jiang: Call `optimizeAnd`

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
-          if (!containsDuplicates && value.size() != childValues.size()) {
-            containsDuplicates = true;
+        String column = childQueryTree.getColumn();
+        Set<String> predicateValues = columnToInEqPredicateValues.get(column);
+        if (predicateValues == null) {
+          // we didn't see this column before
+          predicateValues = new HashSet<>(childValues);
+          // if after adding to the set, the number of predicate values become different (less),
+          // it implies that values got deduplicated and so we need to rewrite this predicate later
+          if (predicateValues.size() != childValues.size()) {
```
Jackie-Jiang:
```
int numChildren = childValues.size();
if (operator == FilterOperator.IN && (numChildren == 1 || numChildren != predicateValues.size())) {
```
And remove line 282-285

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
   }
 
-  private FilterQueryTree rebuildFilterPredicate(Map<String, Set<String>> columnToValues,
-      List<FilterQueryTree> nonEqualityOperators) {
+  private FilterQueryTree rebuildFilterPredicate(Map<String, Set<String>> columnToInEqPredicateValues) {
     ArrayList<FilterQueryTree> newChildren = new ArrayList<>();
-    for (Map.Entry<String, Set<String>> columnAndValues : columnToValues.entrySet()) {
-      newChildren.add(buildFilterQueryTreeForColumnAndValues(columnAndValues));
+    for (Map.Entry<String, Set<String>> columnAndPredicate : columnToInEqPredicateValues.entrySet()) {
+      newChildren.add(buildFilterQueryTreeForColumnAndPredicateInfo(columnAndPredicate.getKey(), columnAndPredicate.getValue()));
```
Jackie-Jiang: Inline this

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
+      // if we are here, it means the query has a single predicate
+      // (just one column with one or more predicate values)
+      // a = 1
+      // a IN (1)
+      // a > 7
+      // a IN (1, 2, 3, 4)
+      // .... anything single predicate
+      // rewrite single predicate IN as EQUALITY, for remaining cases this optimizer is a NO-OP
+      if (filterQueryTree.getOperator() == FilterOperator.IN && filterQueryTree.getValue().size() == 1) {
+        filterQueryTree.setOperator(FilterOperator.EQUALITY);
```
Jackie-Jiang: Reconstruct and keep FilterQueryTree immutable

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: This is reusable -- called for the when there are multiple predicates on single column only and we can remove OR root all-together. So buildFilterQueryTreeForColumnAndPredicateInfo helps by creating the new tree with IN/EQ operator.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Yes probably. However, since amortized O(1) v/s O(logN) might impact the runtime of this algorithm a little bit (and not sure how much if there is a pathological query with deep tree and several columns). So I'd say we can do that in a follow-up PR accompanied by a benchmark. We can also use TreeSet for the inner predicate values.

I have added a TODO

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
+   * in two separate data structures.
+   *
+   * (1) is maintained in a map of column name to predicate info having the
+   * unique set of 1 or more predicate values along with information on whether
+   * we should rewrite this predicate or not
+   *
+   * (2) is maintained in a simple list of FilterQueryTrees representing the
+   * non EQUALITY/IN operators
+   *
+   * We first deal with a very special case where the entire filter just has one column
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
+   * Optimize FilterQueryTree rooted at OR. We look at the immediate children
+   * of OR from two different scenarios:
+   *
+   * (1) The immediate child could be a IN or an EQUALITY operator
+   * (2) The immediate child could be other operators -- AND, <, >, BETWEEN etc
+   *
+   * We first go over the immediate children and collect the operators
+   * in two separate data structures.
+   *
+   * (1) is maintained in a map of column name to predicate info having the
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,296 @@
...
+  }
 
-      // Collect all equality/in values and non-equality operators
-      boolean containsDuplicates = collectChildOperators(filterQueryTree, columnToValues, nonEqualityOperators);
+  /**
+   * Optimize FilterQueryTree rooted at OR. We look at the immediate children
+   * of OR from two different scenarios:
+   *
+   * (1) The immediate child could be a IN or an EQUALITY operator
+   * (2) The immediate child could be other operators -- AND, <, >, BETWEEN etc
```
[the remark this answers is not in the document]
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

siddharthteotia:
@Jackie-Jiang , I have addressed the review comments

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,291 @@
...
+      // if we are here, it means the query has a single predicate
+      // (just one column with one or more predicate values)
+      // a = 1
+      // a IN (1)
+      // a > 7
+      // a IN (1, 2, 3, 4)
+      // .... anything single predicate
+      // rewrite single predicate IN as EQUALITY, for remaining cases this optimizer is a NO-OP
+      if (filterQueryTree.getOperator() == FilterOperator.IN && filterQueryTree.getValue().size() == 1) {
+        return new FilterQueryTree(filterQueryTree.getColumn(), filterQueryTree.getValue(), FilterOperator.EQUALITY, null);
```
Jackie-Jiang: Here you also need to dedup the values for IN clause

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,291 @@
...
   }
 
-  private FilterQueryTree optimize(FilterQueryTree filterQueryTree, FilterQueryTree parent) {
+  /**
+   * Optimize a FilterQueryTree. When we look at the root operator,
+   * there are 3 possible cases:
+   *
+   * (1) root is OR
+   * (2) root is AND
+   * (3) root itself is a leaf level operator like =, <, >, BETWEEN, IN
```
Jackie-Jiang: (nit) EQUALITY, RANGE, IN etc.

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,291 @@
...
+  }
 
-      // Collect all equality/in values and non-equality operators
-      boolean containsDuplicates = collectChildOperators(filterQueryTree, columnToValues, nonEqualityOperators);
+  /**
+   * Optimize FilterQueryTree rooted at OR. We look at the immediate children
+   * of OR from two different scenarios:
+   *
+   * (1) The immediate child could be a IN or an EQUALITY operator
+   * (2) The immediate child could be other operators -- AND, NOT, RANGe, <, >,BETWEEN etc
```
Jackie-Jiang: (nit) RANGE
<, >, BETWEEN are all classified as RANGE

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,291 @@
...
-          if (!containsDuplicates && value.size() != childValues.size()) {
-            containsDuplicates = true;
+        String column = childQueryTree.getColumn();
+        Set<String> predicateValues = columnToInEqPredicateValues.get(column);
+        if (predicateValues == null) {
+          // we didn't see this column before
+          predicateValues = new HashSet<>(childValues);
+          // if after adding to the set, the number of predicate values become different (less),
+          // it implies that values got deduplicated and so we need to rewrite this predicate later
+	  // similarly, if this is a single value IN predicate, we need to rewrite it later to equality
```
Jackie-Jiang: (nit) indentation

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
```diff
@@ -20,148 +20,291 @@
...
-    // We need to rebuild the predicate if there were duplicate values detected (eg. a = 1 OR a = 1) or if there is
-    // more than one value for a column (eg. a = 1 OR a = 2)
-    boolean rebuildRequired = containsDuplicates;
-
-    if (!rebuildRequired) {
-      for (Set<String> columnValues : columnToValues.values()) {
-        if (1 < columnValues.size()) {
-          rebuildRequired = true;
-          break;
+        // the immediate child of root operator is <, >, AND, RANGE, <=, >=, BETWEEN ...
```
Jackie-Jiang: (nit) all the comparisons are classified as RANGE

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
Jackie-Jiang: You can inline rebuildFilterPredicate() and pre-allocate the size of newChildren. Your choice

pinot-common/src/main/java/org/apache/pinot/common/utils/request/FilterQueryTree.java
```diff
@@ -27,7 +27,7 @@
   private final String column;
   private final TransformExpressionTree _expression;
   private final List<String> value;
-  private final FilterOperator operator;
+  private FilterOperator operator;
```
Jackie-Jiang: Revert this

siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

pinot-broker/src/main/java/org/apache/pinot/broker/requesthandler/MultipleOrEqualitiesToInClauseFilterQueryTreeOptimizer.java
siddharthteotia: Done

PR #4848, opened 2019-11-22
Remove the redundant testSqlQueriesFromQueryFile() in integration tests

Jackie-Jiang:
This test is calling the testQueriesFromQueryFile(), which is already called in the test
To fix this test, we need to generate a SQL version of the query file


pinot-integration-tests/src/test/java/org/apache/pinot/integration/tests/BaseClusterIntegrationTestSet.java
```diff
@@ -262,8 +262,7 @@ public void testQueriesFromQueryFile()
 
   /**
    * Test random SQL queries from the query file.
-   *
-   * @throws Exception
+   * TODO: fix this teat by adding the SQL query file
```
jackjlli: typo: test

Jackie-Jiang: Haha, fixed

PR #4851, opened 2019-11-22
Split BrokerReduceService code into ResultSetters

npawar:
Splitting the code which sets results into BrokerResponseNative, in BrokerReduceService, into multiple ResultSetters. The work in this PR is mostly copy pasting the methods from BrokerReduceService into the right Setter, without any logic change. The only logic added, is the handling of empty dataTableMap into each setter. It was done upfront in the earlier implementation, before selecting the path based on the query.

This is necessary because for the sql compliance work, we will be making each query type generate results into ResultTable. Currently this exists only for group by with order by. As we add this code for each query (selection, aggregation, distinct, group by), the BrokerReduceService reduce method will get extremely long and windy. This was attempted in https://github.com/apache/incubator-pinot/pull/4833 and it became difficult to review. Hence decided to split that PR into multiple steps
1. ResultSetters (this PR)
2. AggregationFunction changes
3. ResultTable for all


npawar:

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/resultsetter/DistinctResultSetter.java
```diff
@@ -0,0 +1,116 @@
...
+import org.apache.pinot.common.utils.DataTable;
+import org.apache.pinot.core.data.table.Record;
+import org.apache.pinot.core.query.aggregation.DistinctTable;
+import org.apache.pinot.core.query.aggregation.function.AggregationFunction;
+import org.apache.pinot.core.query.aggregation.function.AggregationFunctionUtils;
+import org.apache.pinot.core.transport.ServerRoutingInstance;
+import org.apache.pinot.core.util.QueryOptions;
+
+
+public class DistinctResultSetter implements ResultSetter {
```
siddharthteotia: Javadoc might be helpful

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/resultsetter/DistinctResultSetter.java
```diff
@@ -0,0 +1,116 @@
...
+    _brokerRequest = brokerRequest;
+    _aggregationFunction = aggregationFunction;
+  }
+
+  @Override
+  public void setResults(String tableName, DataSchema dataSchema, Map<ServerRoutingInstance, DataTable> dataTableMap,
+      BrokerResponseNative brokerResponseNative, BrokerMetrics brokerMetrics) {
+
+    if (dataTableMap.isEmpty()) {
+      return;
```
siddharthteotia: I think we should send an empty non-null selection response here just like SelectionResultSetter does.

npawar: Done

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/resultsetter/DistinctResultSetter.java
```diff
@@ -0,0 +1,128 @@
...
+    // Up until now, we have treated DISTINCT similar to another aggregation function even in terms
+    // of the result from function and merging results.
+    // However, the DISTINCT query is just another SELECTION style query from the user's point
+    // of view and will return one or records in the result table for the column(s) selected and so
+    // for that reason, response from broker should be a selection query result.
+    brokerResponseNative.setSelectionResults((new SelectionResults(Arrays.asList(columnNames), resultSet)));
+  }
+
+  private String[] getDistinctColumns() {
+    return _brokerRequest.getAggregationsInfo().get(0).getAggregationParams().get("column").split(":");
```
siddharthteotia: (nit) I think it is better to use the constants for ":" and and "column". These constant are already defined in FunctionCallAstNode and are used in multiple places wherever we need to get the set of columns that are arguments to DISTINCT.

siddharthteotia:
LGTM. Just one minor comment on using constants from FunctionCallAstNode.

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/resultsetter/DistinctResultSetter.java
npawar: fixed.

pinot-core/src/main/java/org/apache/pinot/core/query/reduce/ResultReducer.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.pinot.common.response.broker.BrokerResponseNative;
+import org.apache.pinot.common.utils.DataSchema;
+import org.apache.pinot.common.utils.DataTable;
+import org.apache.pinot.core.transport.ServerRoutingInstance;
+
+
+/**
+ * Interface for result reducers of query results
+ */
+public interface ResultReducer {
```
Jackie-Jiang: Recommend changing it to DataTableReducer, which can reflect the functionality better

PR #4852, opened 2019-11-22
Add getFinalResultColumnType to be used in constructing ResultTable::dataSchema

npawar:
This is necessary because for the sql compliance work, we will be making each query type generate results into ResultTable. ResultTable needs to set the dataSchema of the final results, and we need a way to get the data type of the final extracted aggregation results.


pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/AggregationFunction.java
```diff
@@ -107,6 +107,13 @@ void aggregateGroupByMV(int length, int[][] groupKeysArray, GroupByResultHolder
    */
   ColumnDataType getIntermediateResultColumnType();
 
+  /**
+   * Returns the {@link ColumnDataType} of the final result.
+   * <p>This column data type is used for constructing the result table</p>
```
Jackie-Jiang: (nit) Follow the same convention for the javadoc (period at end; no return)

PR #4859, opened 2019-11-26
Enhance TagNameUtils to handle the TagOverrideConfig

Jackie-Jiang:
Also remove the following config classes (not really config):
- TagConfig
- OfflineTagConfig
- RealtimeTagConfig

NOTE: tenant should never have suffix; tag should always have suffix
Remove the support of mixed use of tenant/tag to prevent attaching wrong suffix


pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
```diff
@@ -27,78 +28,93 @@ private TagNameUtils() {
...
   }
 
   public static boolean isRealtimeServerTag(String tagName) {
-    return tagName.endsWith(ServerType.REALTIME.toString());
+    return tagName.endsWith(REALTIME_SERVER_TAG_SUFFIX);
   }
 
-  public static boolean isBrokerTag(String tagName) {
-    return tagName.endsWith(TenantRole.BROKER.toString());
+  public static String getTenantFromTag(String tagName) {
```
jackjlli: You can call `isOfflineServerTag`, `isRealtimeServerTag`, and `isBrokerTag` to reduce code redundancy.

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
```diff
@@ -27,78 +28,93 @@ private TagNameUtils() {
...
     }
   }
 
-  public static String getOfflineTagForTenant(String tenantName) {
-    if (tenantName == null) {
-      return TagNameUtils.getOfflineTagForTenant(DEFAULT_TENANT_NAME);
-    }
-    if (tenantName.endsWith(ServerType.OFFLINE.toString())) {
-      return tenantName;
+  public static String extractBrokerTag(TenantConfig tenantConfig) {
```
jackjlli: Can `tenantConfig` be null?

Jackie-Jiang: It won't be (everything not annotated with nullable are non-null)

pinot-common/src/main/java/org/apache/pinot/common/config/TagNameUtils.java
Jackie-Jiang: Simplified

PR #4862, opened 2019-11-26
[TE] Generify centralized cache DAO and add certificate-based authentication support for Couchbase

bxji:
Made Cache DAO an interface and generified config code to allow users to use their own data store of choice.

Also adds support for authenticating to Couchbase using certificates. This is a Couchbase Enterprise Edition feature, but some users with enterprise Couchbase setups may find this feature useful.

### DIFF
--- a/thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
+++ b/thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
@@ -12,7 +12,7 @@ centralizedCacheSettings:
     couchbase:
       className: org.apache.pinot.thirdeye.detection.cache.CouchbaseCacheDAO
       config:
-        useCertificateBasedAuthentication: falses
+        useCertificateBasedAuthentication: false
         # at least 1 host needed
         hosts:
           - 'host1' # ex. http://localhost:8091
--- a/thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
+++ b/thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
@@ -23,7 +23,7 @@ centralizedCacheSettings:
         # if using certificate-based authentication, authUsername and authPassword values don't matter and won't be used
         authUsername: 'your_bucket_user_username'
         authPassword: 'your_bucket_user_password'
-        enableDnsSrv: true
+        enableDnsSrv: false
         # certificate based authentication is only available in Couchbase enterprise edition. 
         keyStoreFilePath: 'key/store/path/keystore_file' # e.g. '/var/identity.p12'
         # if your keystore has a password, enter it here. by default, Java uses 'work_around_jdk-6879539' to enable empty passwords for certificates.
--- a/thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
+++ b/thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
@@ -12,7 +12,7 @@ centralizedCacheSettings:
     couchbase:
       className: org.apache.pinot.thirdeye.detection.cache.CouchbaseCacheDAO
       config:
-        useCertificateBasedAuthentication: false
+        useCertificateBasedAuthentication: falses
         # at least 1 host needed
         hosts:
           - 'host1' # ex. http://localhost:8091
@@ -23,6 +23,7 @@ centralizedCacheSettings:
         # if using certificate-based authentication, authUsername and authPassword values don't matter and won't be used
         authUsername: 'your_bucket_user_username'
         authPassword: 'your_bucket_user_password'
+        enableDnsSrv: true
         # certificate based authentication is only available in Couchbase enterprise edition. 
         keyStoreFilePath: 'key/store/path/keystore_file' # e.g. '/var/identity.p12'
         # by default, Java uses this password to enable empty passwords for certificates.
--- a/thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
+++ b/thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
@@ -26,9 +26,9 @@ centralizedCacheSettings:
         enableDnsSrv: true
         # certificate based authentication is only available in Couchbase enterprise edition. 
         keyStoreFilePath: 'key/store/path/keystore_file' # e.g. '/var/identity.p12'
-        # by default, Java uses this password to enable empty passwords for certificates.
+        # if your keystore has a password, enter it here. by default, Java uses 'work_around_jdk-6879539' to enable empty passwords for certificates.
         keyStorePassword: 'work_around_jdk-6879539'
         trustStoreFilePath: 'trust/store/path/truststore_file' # e.g. '/etc/riddler/cacerts'
-        trustStorePassword: '' 
+        trustStorePassword: ''
     # add your store of choice here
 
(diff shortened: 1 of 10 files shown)


thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
```diff
@@ -9,12 +7,28 @@ centralizedCacheSettings:
...
+          - 'host4' # and so on...
+        bucketName: 'your_bucket_name'
+        # if using certificate-based authentication, authUsername and authPassword values don't matter and won't be used
+        authUsername: 'your_bucket_user_username'
+        authPassword: 'your_bucket_user_password'
+        enableDnsSrv: true
+        # certificate based authentication is only available in Couchbase enterprise edition. 
+        keyStoreFilePath: 'key/store/path/keystore_file' # e.g. '/var/identity.p12'
+        # by default, Java uses this password to enable empty passwords for certificates.
+        keyStorePassword: 'work_around_jdk-6879539'
```
xiaohui-sun: Is it a true password?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheDataSource.java
```diff
@@ -19,30 +19,36 @@
...
-  public String getHost() { return host; }
-  public String getAuthUsername() { return authUsername; }
-  public String getAuthPassword() { return authPassword; }
-  public String getBucketName() { return bucketName; }
+  public String getClassName() { return className; }
+  public Map<String, Object> getConfig() { return config; }
+
+  public Object getField(String fieldName) { return config.get(fieldName); }
+  public String getFieldAsString(String fieldName) { return String.valueOf(config.get(fieldName)); }
+  public Boolean getFieldAsBoolean(String fieldName) { return Boolean.parseBoolean(this.getFieldAsString(fieldName)); }
```
xiaohui-sun: Why not using Configuration interface which has all the helper functions built in?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
```diff
@@ -86,4 +93,30 @@ public static String buildQuery(JsonObject parameters) {
         parameters.getLong("start"),
         parameters.getLong("end"));
   }
+
+  /**
+   * Convert list of strings to their proper URL hosts. For each
+   * string, it will parse out the proper host for the URI.
+   * Example: "http://localhost:8091" -> "localhost"
+   * @param bootstrapUris
+   * @return
```
xiaohui-sun: What is return value?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
```diff
@@ -86,4 +93,30 @@ public static String buildQuery(JsonObject parameters) {
...
+  public static List<String> getBootstrapHosts(List<String> bootstrapUris) {
+    List<String> bootstrapHosts = new ArrayList<>(bootstrapUris.size());
+    for (String bootstrapUri : bootstrapUris) {
+      try {
+        URI uri = new URI(bootstrapUri);
+        // The next workaround is to correctly parse urls that don't include any schema
+        if (uri.getHost() == null) {
+          uri = new URI("http://" + bootstrapUri);
+        }
+        String host = uri.getHost();
```
xiaohui-sun: Is it possible the host is still null?

thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
bxji: No, it is used as a placeholder to designate empty/not used password. See https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6879539

If there is a password associated with the keystore file, user can set it here.

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
bxji: oops, forgot to add that in. It returns a list of strings, which are the hosts (e.g. [ "host1", "host2", "host3", "host4"]).

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
bxji: Good catch. Yes, but Couchbase will throw an exception upon trying to connect to "null". I will add a check. Do you think it would be better to stop execution if there was an invalid host, or just warn the user and skip that particular host?

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheDataSource.java
bxji: I'm not sure which Configuration interface you're referring to. I switched to using MapUtils for now.

thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
```diff
@@ -9,12 +7,28 @@ centralizedCacheSettings:
...
-      # host can just be localhost if you're using this locally.
-      host: your_host
-      authUsername: your_username
-      authPassword: your_password
-      bucketName: your_bucket_name
-    # add your store of choice here
\ No newline at end of file
+      className: org.apache.pinot.thirdeye.detection.cache.CouchbaseCacheDAO
+      config:
+        useCertificateBasedAuthentication: falses
```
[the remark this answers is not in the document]
bxji: good catch, fixed

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheDataSource.java
xiaohui-sun: https://commons.apache.org/proper/commons-configuration/apidocs/org/apache/commons/configuration2/Configuration.html

thirdeye/thirdeye-pinot/config/data-sources/cache-config.yml
xiaohui-sun: Thanks for the explanation.

bxji: I added some more context in a comment, since it's not very intuitive. Thanks for pointing that out

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/util/CacheUtils.java
akshayrai: warn and skip

bxji: done

thirdeye/thirdeye-pinot/src/main/java/org/apache/pinot/thirdeye/detection/cache/CacheDataSource.java
bxji: Seems like Jackson doesn't like the Configuration interface, and isn't able to deserialize into it directly. I think keeping Map<String, Object> is fine if we reuse methods from MapUtils.

PR #4863, opened 2019-11-26
Query results in ResultTable if responseFormat = sql

npawar:
This falls under the sql compliance work. We are moving to ResultTable, which will serve as the single tabular sql like structure to return results from all types of queries.
This is currently turned off by default. This feature can be enabled by passing `responseFormat=sql` in queryOptions.

PS: this is the 3rd step of the breakdown of PR https://github.com/apache/incubator-pinot/pull/4833


npawar:

Issue #4731, opened 2019-10-21
Simplify Pinot data ingestion


kishoreg:
Getting data into Pinot is non-trivial. Most users end up writing a custom hadoop/spark job for data ingestion. There is no standard way to write these jobs. The goal of this issue is to come up with a design that makes it easy to ingest data into Pinot.

Before getting to the solution, let's look at all the variables
1. **Formats**: JSON, CSV, AVRO, Parquet, Thrift, Protobuf, ORC
2. **Batch Datasources**: HDFS, S3, ADLS, GCS
3. **Streaming Datasources**: Stream: Kafka, EventHub
4. **Execution Frameworks**: Hadoop, Spark, Flink, Samza, etc

As of now, we have RecordReader interface to ingest batch data.
```  
  GenericRow next(GenericRow reuse)     throws IOException;
```

And the real-time uses a ```StreamMessageDecoder```  interface.
```
  GenericRow decode(T payload, GenericRow reuse);
```
The RecordReader implementations are spread all over pinot-core code which brings in lot of external dependencies. We have moved pinot-parquet and pinot-orc to separate packages.

For real-time, we need one implementation per datasource, dataFormat. e.g. KafkaAvroMessageDecoder.

Any thoughts on how to standardize Data Ingestion in Pinot?

kishoreg:
One suggestion is to first refactor Pinot modules to something like this
Pinot-batch-reader (Implements RecordReader)
- parquet
- csv
- json
- avro
- thrift

Pinot-stream-decoders (StreamMessageDecoder)
- json
- avro
- protobuf
- Thrift

Pinot-connectors (For real-time StreamConsumerFactory etc) (PinotFS for batch)
- Kafka
- Hdfs
- S3
- ADLS
- GCS

Pinot-ingestion-job (no interface here) 
- Hadoop
- Spark
- Flink
- Samza

mayankshriv:
Great idea, and also agree with the initial refactoring.

mcvsubbu:
+1

Perhaps we can split PinotFS and Realtime ingestion connectors.

kishoreg:
More detailed proposal. https://docs.google.com/document/d/1160Wz-694-SDuFEKrZ2rW_iIswcTqk4hJ3k-BNp0BaI/edit?usp=sharing

PR #4866, opened 2019-11-27
Removing segment generation config from RecordReader interface

kishoreg:
Related to #4731 
refactoring the existing code to prepare for moving RecordReader to pinot-spi module. 

As of today, RecordReader takes in SegmentGenerationConfig which makes it very hard to extract RecordReader to pinot-spi. 

current interface
`void init(SegmentGenerationConfig config)`

new interface
`void init(String inputPath, Schema schema, ReaderConfig readerConfig)`

Next step:
- Move record reader interface to pinot-spi module
- Move Schema and related classes
- Instantiate record readers dynamically during segment generation (this is optional and might do it later depending on how easy/hard). We don't need it until we last steps

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReader.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReader.java
@@ -19,6 +19,7 @@
 package org.apache.pinot.core.data.readers;
 
 import java.io.Closeable;
+import java.io.File;
 import java.io.IOException;
 import javax.annotation.Nullable;
 import org.apache.pinot.common.data.Schema;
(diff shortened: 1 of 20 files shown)


pinot-core/src/main/java/org/apache/pinot/core/data/readers/RecordReader.java
```diff
@@ -34,10 +34,16 @@
...
+   * The implementation can chose to ignore one or more of these parameters and handle null gracefully <br/>
+   *
+   * @param inputPath absolute path to the file/directory
+   * @param schema Pinot Schema associated with the table
+   * @param recordReaderConfig config for the reader specific to the format. e.g. delimiter for csv format etc
+   * @throws Exception if the arguments are invalid
    */
-  void init(SegmentGeneratorConfig segmentGeneratorConfig)
-      throws IOException;
+  void init(@Nullable String inputPath, @Nullable Schema schema, @Nullable RecordReaderConfig recordReaderConfig)
```
mcvsubbu: Why not move input path and schema into RecordReaderConfig? That way, everything that the record reader needs is there in the config.

Jackie-Jiang: This config is for record readers with extra settings (e.g. CSV delimiter), which can be null. Input path and schema are required for all record readers

Jackie-Jiang: Should we make inputPath and schema mandatory?

kishoreg: I thought about it. Not all RecordReader need that. Ideally, we want to pass a PinotFS interface wrapper but want to avoid making too many changes in this PR. My first goal is to decouple the modules without adding too much functionality/changing existing classes.  We will have a better idea when we move the interfaces to spi module.

kishoreg: > This config is for record readers with extra settings (e.g. CSV delimiter), which can be null. Input path and schema are required for all record readers

Not really, some of them take GenericRows. There are too many implementations for recordReader: used in testing, segment conversion, Minion etc. Most of the readers get what they want in their constructors today.

kishoreg: I could not conclude that looking at the existing implementations.

Jackie-Jiang: When you call `init()`, you will always provide the input path and schema right?

kishoreg: only for Parquet and ORC readers. There are some record readers which dont need schema as of today.

Jackie-Jiang: I understand most record readers don't need schema as of now, but when you call the `init()`, you are not supposed to pass `null` as the schema

kishoreg:
#4869 captures the changes in this PR and incorporates the feedback. Closing this PR

PR #4869, opened 2019-11-27
Adding pinot-spi module and moving record reader interface

kishoreg:
Added pinot-spi module. I based this on the changes made it #4866. It mostly involves moving the classes around and changing the import. 

- Schema and related classes moved to pinot-spi along with this.  
- I had to move some utils that we are using in Schema and FieldSpec. The utils are bringing additional jars(commons, jackson,etc) for now.
- We can clean that up in the next version.

It was not as invasive as I had thought. Please review.

If this looks good we can drop the changes #4866 since this already includes the changes in that.


kishoreg:
Merging so that others can push changes.

kishoreg:
Jackie had also approved this earlier and we were waiting for build to pass

PR #4870, opened 2019-11-27
Add getResultColumnName API to AggregationFunction to fetch sql compliant result name

npawar:
PQL converts column name as "AVG(foo)" -> "avg_foo". In the SQL style, this would have been "avg(foo)". Adding a new API to get the new result name.
Enhanced tests to check the right name is received at the ResultTable in the broker results

### DIFF
--- a/pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
+++ b/pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
@@ -57,17 +57,16 @@ public class TableResizer {
     int numAggregations = aggregationInfos.size();
     int numKeyColumns = numColumns - numAggregations;
 
-    int aggIdx = 0;
     Map<String, Integer> columnIndexMap = new HashMap<>();
     Map<String, AggregationFunction> aggregationColumnToFunction = new HashMap<>();
     for (int i = 0; i < numColumns; i++) {
-      String columnName = dataSchema.getColumnName(i).toLowerCase();
+      String columnName = dataSchema.getColumnName(i);
       columnIndexMap.put(columnName, i);
       if (i >= numKeyColumns) {
+        AggregationInfo aggregationInfo = aggregationInfos.get(i - numKeyColumns);
         AggregationFunction aggregationFunction =
-            AggregationFunctionUtils.getAggregationFunctionContext(aggregationInfos.get(aggIdx)).getAggregationFunction();
+            AggregationFunctionUtils.getAggregationFunctionContext(aggregationInfo).getAggregationFunction();
         aggregationColumnToFunction.put(columnName, aggregationFunction);
-        aggIdx++;
       }
     }
 
@@ -78,7 +77,7 @@ public class TableResizer {
     if (numKeyColumns < numColumns) {
       for (int orderByIdx = 0; orderByIdx < _numOrderBy; orderByIdx++) {
         SelectionSort selectionSort = orderBy.get(orderByIdx);
-        String column = selectionSort.getColumn().toLowerCase();
+        String column = selectionSort.getColumn();
 
         if (columnIndexMap.containsKey(column)) {
           int index = columnIndexMap.get(column);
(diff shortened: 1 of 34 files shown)


kishoreg:
This is good to have for now since we don't support aliasing. But this should be typically solved at a very high level using a mapping table. Let's make sure that this will not make it difficult to implement aliasing later.

kishoreg:
what functions in selections and group by.

pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
```diff
@@ -57,15 +57,18 @@
...
+    Map<String, AggregationFunction> aggregationColumnToFunction = new HashMap<>();
     for (int i = 0; i < numColumns; i++) {
-      columnIndexMap.put(dataSchema.getColumnName(i), i);
-    }
-
-    Map<String, AggregationInfo> aggregationColumnToInfo = new HashMap<>();
-    for (AggregationInfo aggregationInfo : aggregationInfos) {
-      String aggregationColumn = AggregationFunctionUtils.getAggregationColumnName(aggregationInfo);
-      aggregationColumnToInfo.put(aggregationColumn, aggregationInfo);
+      String columnName = dataSchema.getColumnName(i).toLowerCase();
```
Jackie-Jiang: This is not correct as `column` and `COLUMN` should be treated as two different columns (case sensitive)

pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
```diff
@@ -57,15 +57,18 @@
...
+      columnIndexMap.put(columnName, i);
+      if (i >= numKeyColumns) {
+        AggregationFunction aggregationFunction =
```
Jackie-Jiang: Please add some comments here (key columns are in front of aggregation columns).
Also, `aggIdx === i - numKeyColumns`, so you don't need to maintain another index.

npawar:
> This is good to have for now since we don't support aliasing. But this should be typically solved at a very high level using a mapping table. Let's make sure that this will not make it difficult to implement aliasing later.

I feel this is required. We have no other way to fetch the name like `avg(foo)`. Order by cannot understand `avg_foo`, as the TransformExpressionTree converts it to `avg(foo)`.
I don't think this should affect ability to handle aliases later

pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
npawar: There's a comment at the beginning of this method (few lines above this code change).
Fixed the other suggestion

pinot-core/src/main/java/org/apache/pinot/core/data/table/TableResizer.java
npawar: Changed this. I ended up making function name lower case in the AggregationFunction::getResultColumnName, because TransformExpressionTree converts function names to lower case when building the broker request. Hence order by only has all lower case function names

pinot-core/src/main/java/org/apache/pinot/core/query/aggregation/function/MinAggregationFunction.java
```diff
@@ -40,6 +40,11 @@ public String getColumnName(String column) {
     return AggregationFunctionType.MIN.getName() + "_" + column;
   }
 
+  @Override
+  public String getResultColumnName(String column) {
```
mayankshriv: Curious if we can have this method in the base class. If the base class calls getFunctionName() that returns the concrete class's getFunctionName(), then we won't need to duplicate this logic in all classes (we would need getFunctionName() though).

npawar: Yes can be done. Moved the common logic to the default method implementation

npawar:
> > This is good to have for now since we don't support aliasing. But this should be typically solved at a very high level using a mapping table. Let's make sure that this will not make it difficult to implement aliasing later.
> 
> +1 to ^^. Would this be a throw away change when we support aliasing? Or will we just need a map at top level to do result column name translation?

I don't think it will be a throw away change. The need for getting the resultName in this brackets format arose primarily because the SelectionSort has the column name in the brackets format vs aggregations have the name in the underscore format. Only the aggregation functions are converted to underscore format. Other functions (udfs), are in brackets format even today. 

In the absence of the new method, either the the order by clause will need to be rewritten to underscore format, or the column will have to be translated to underscore when applying order by (which is in 3 places).

PR #4872, opened 2019-11-30
Refactor pinot record reader dependency

fx19880617:
1. Refactor to make `pinot-orc` only depends on `pinot-spi`
2. Refactor to make `pinot-parquet` only depends on `pinot-spi`
3. Refactor to move `RecordReaderUtils` from `org.apache.pinot.core.data.readers.RecordReaderUtils` to `org.apache.pinot.spi.data.readers.RecordReaderUtils`
4. Adding `pinot-avro` module and move `org.apache.pinot.core.data.readers.AvroRecordReader` to it.
5. Refactor `{Avro/Parquet/Csv/Json}RecordReaderTest` to extend `org.apache.pinot.spi.data.readers.AbstractRecordReaderTest`.


kishoreg:
Looks good. Nothing to do with this PR but we should clean up RecordReaderUtils. Looks like it has some methods related to IOUtils and DataType utils.

kishoreg:
Nice job extracting AbstractRecordReaderTest. We can make it better in the next iteration.

kishoreg:
Also, once we have the ability to dynamically load record-reader impls, we can remove pinot-avro dependency from pinot-core right?

fx19880617:
> Also, once we have the ability to dynamically load record-reader impls, we can remove pinot-avro dependency from pinot-core right?

Not really, RecordReader part is ok, but there is also many dependencies on `org.apache.pinot.avro.data.readers.AvroUtils`

PR #4873, opened 2019-12-01
Adding pinotFS.open(URI) method to pinotFS

kishoreg:
This will allow us to make pinot-ingestion-job independent of hadoop libraries.  DefaultControllerRestApi.pushSegments relies on getting an inputstream.  With this change, we can  rewrite pinot-ingestion-common to use pinot-fs api's instead of relying on hadoop apis.

See line 114
`         try (InputStream inputStream = fileSystem.open(tarFilePath)) {`


kishoreg:
I will take care of documenting/annotating once we have the plug-ins work end to end. I don't think anyone has implemented these interfaces outside as of now. For all practical purposes, these interfaces are experimental for now.

